PR #332, opened 2017-04-02
Updated readme with new install instructions.

illume:
FYI, I've updated the appveyor, travisci and launchpad builders to point to github.


illume:
Seems like I clicked a little button and the repo got rewritten with a force push. (Matching an author up.). So I won't do that again.

PR #345, opened 2017-08-13
Compile pypm.pyx with current cython

stuaxo:
Update pypm.c to be compiled with cython 0.20.0.

This is the latest version of cython that compiles this file, this stops pypy complaining about pypm.


stuaxo:
Updated to current cython (0.26) following help from their mailing list.

stuaxo:
@takluyver this passes everything except OS/X which I'm attempting to fix in another pull request.

stuaxo:
Don't know how to fix the OSX build. but don't think it is this patch is what broke it.

takluyver:
No, I think it was probably broken by changes to Travis' infrastructure. We were deliberately using the oldest OSX image available, because the binaries it compiles can be used on newer OSX versions but not older.

takluyver:
I think you asked on the Cython mailing list and figured out how to fix it so it works with newer versions of Cython.

If you've got a bit more time to spend on it, I think it would be best to add Cython to the build process, and remove the generated .c file from the repository. It's a bit more effort, but it's the 'right' way of doing it, IMO.

stuaxo:
Hi,
   I had a bit of a look at this, but free time is really tight.   What needs to happen is build_ext in setup.py calls cythonize for the relevant files.

I got stuck as the setup.py is a little complex (like any project the age of pygame).

A slight complication is that cythonize will only be available when cython is installed, so it will need to be in setup_requires, and I guess have a stub cythonize the rest of the time.

S

takluyver:
Thanks! I think it's OK if building from source fails without Cython - it already requires several other things, and Cython isn't particularly tricky to install.

I'm not really sure how the different build bits in pygame fit together, but it looks like you could add the cythonize results to the `extensions` list created here:

https://github.com/pygame/pygame/blob/d8cef0d436b0c0632e5c819f60d3424c7b4859cc/setup.py#L165

illume:
Thanks!

Would rather not add cython dependency to 1.9 series. As all the upstream packages would need to be updated too. Compiling with latest cython can happen in a separate commit. Mergey merge.

stuaxo:
Cheers - I'm pretty slow at getting back round to things, so thanks for merging.

illume:
@stuaxo you're welcome. I guess you keep up with pypy things... have there been any other movements there? Those key issues still existing? I've got some time in the next days to look into things, and debug stuff if needed.

stuaxo:
Pypys support of the Python C API has definitely been getting better, at the same time newer versions of Cython seem to support pypy better.

PyCairo just had some similar patches to the one I attempted on math.c - moving from direct field access to using the Py* defines gets you quite far.

If you hit cases that don't work, the pypy guys are usually pretty helpful.

CFFI remains the most efficient option when you are in Pypy, so in the end I'd hope everything using the python C API will move to this, but we shall see.

Every now and then I like to test libraries like pygame, panda3d and others on pypy...   pygame and pycairo working with pypy should be quite a big thing.

I has happily suprised that Cocos2D started working on pypy in the last year or so.


Not sure if it's useful but I have a github gist to grab pypy

https://gist.github.com/stuaxo/5d4c0363317a875617271a6d424abbba/

Often new things work in the nightly build first.


Definitely just test and see what works, and if things don't work try the newest versions of the libraries concerned.

illume:
Cheers, thanks for the info. I'll be using that gist shortly :)

stuaxo:
Fixed a bug on it that affected running it for the first time.

PR #348, opened 2017-08-23
Attempt to stop travis build failing

stuaxo:
Change to libX11-dev to libx11-dev so build stops failing.


takluyver:
Seems to be working!

PR #357, opened 2017-09-21
add UYVY support

adicarlo:
Consider a camera like this:

```
Driver Info (not using libv4l2):
	Driver name   : stk1160
	Card type     : stk1160
	Bus info      : usb-0000:00:14.0-10.7
	Driver version: 4.12.6
	Capabilities  : 0x85200001
		Video Capture
		Read/Write
		Streaming
		Extended Pix Format
		Device Capabilities
	Device Caps   : 0x05200001
		Video Capture
		Read/Write
		Streaming
		Extended Pix Format
Priority: 2
Video input : 0 (Composite0: ok)
Video Standard = 0x00001000
	NTSC-M
Format Video Capture:
	Width/Height      : 640/480
	Pixel Format      : 'UYVY'
	Field             : Interlaced
	Bytes per Line    : 1280
	Size Image        : 614400
	Colorspace        : SMPTE 170M
	Transfer Function : Default
	YCbCr/HSV Encoding: Default
	Quantization      : Default
	Flags             : 
Streaming Parameters Video Capture:
	Frames per second: 29.970 (30000/1001)
	Read buffers     : 2

User Controls

                     brightness (int)    : min=0 max=255 step=1 default=128 value=128 flags=slider
                       contrast (int)    : min=0 max=127 step=1 default=64 value=64 flags=slider
                     saturation (int)    : min=0 max=127 step=1 default=64 value=64 flags=slider
                            hue (int)    : min=-128 max=127 step=1 default=0 value=0 flags=slider
                     chroma_agc (bool)   : default=1 value=1 flags=update
                    chroma_gain (int)    : min=0 max=127 step=1 default=40 value=47 flags=inactive, slider, volatile
```

Note the pixel format here is UYVY.  Attempting to start capture with pygame.camera will give you an error like this:

```
    cam.start()
SystemError: ioctl(VIDIOC_S_FMT) failure: no supported formats
```

The attached patch allows pygame to make use of this sort of camera.

I have tested this and I get good output in RGB, and, I *think*, expected output in YUV, its hard to say, since it doens't look normal like RGB, but its consistent with what I get with other camers in YUV mode.

### DIFF
--- a/src/_camera.c
+++ b/src/_camera.c
@@ -939,6 +939,142 @@ void yuyv_to_yuv (const void* src, void* dst, int length, SDL_PixelFormat* forma
     }
 }
 
+/* cribbed from above, but modified for uyvy ordering */
+void uyvy_to_rgb (const void* src, void* dst, int length, SDL_PixelFormat* format) {
+    Uint8 *s, *d8;
+    Uint16 *d16;
+    Uint32 *d32;
+    int i;
+    int r1, g1, b1, r2, b2, g2;
+    int rshift, gshift, bshift, rloss, gloss, bloss, y1, y2, u, v, u1, rg, v1;
+
+    rshift = format->Rshift;
+    gshift = format->Gshift;
+    bshift = format->Bshift;
+    rloss = format->Rloss;
+    gloss = format->Gloss;
+    bloss = format->Bloss;
+
+    d8 = (Uint8 *) dst;
+    d16 = (Uint16 *) dst;
+    d32 = (Uint32 *) dst;
+    i = length >> 1;
+    s = (Uint8 *) src;
+
+    /* yuyv packs 2 pixels into every 4 bytes, sharing the u and v color
+       terms between the 2, with each pixel having a unique y luminance term.
+       Thus, we will operate on 2 pixels at a time. */
+    while (i--) {
+        u = *s++;
+        y1 = *s++;
+        v = *s++;
+        y2 = *s++;
+
+        /* The lines from here to the switch statement are from libv4l */
+        u1 = (((u - 128) << 7) +  (u - 128)) >> 6;
+        rg = (((u - 128) << 1) +  (u - 128) + ((v - 128) << 2) +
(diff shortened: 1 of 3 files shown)


adicarlo:
The test failures seem to be spurious.

aszlig:
Damn, just found this PR **after** creating my own [patch](https://gist.github.com/aszlig/79d2cac15ed3e2841902eab9aab2ea21).

@adicarlo: It might make sense for your implementation to make the pixel format conversion functions a bit more DRY instead of just copying the `yuyv_to_*` functions (see my patch above, although even in my patch it would probably make sense to rename the `yuyv_to_*` functions to something more general).

adicarlo:
@aszlig You raise a good point about DRY, but speaking of DRY, why is it that pygame needs to include these fundamental pixel format conversions at all?  Surely there is an efficient C library version, perhaps in a library we're already using, with an implementation we can use.  That would mean less code, easier maintenance, and we'd also pick up a lot of additional unimplemented pixel conversions.

This more minimal approach would benefit pygame more, I think, but its beyond what I wanted to embark upon.

takluyver:
Closing/reopening to re-run the tests with fixes in master.

adicarlo:
@takluyver any issues or way I can help ?  Or was this already merged?  Github seems to say "no".

takluyver:
The tests are passing now. :-) I'd like someone else to have a quick review of it, because I'm not very familiar with C code.

aszlig:
@takluyver: Apart from the comment mentioned above, LGTM.

takluyver:
Thanks @aszlig !

In terms of clarity, it seems better to me to have separate functions than to have a `yuyv_to_rgb` function handling both `yuyv` and `uyvy`. Though maybe the solution is to give it a clearer name. @adicarlo , what do you think?

illume:
Either are good if it makes it work on more cameras :)

👍 from me.

takluyver:
OK, let's press the button - the code arrangement can always be improved in a separate PR.

PR #360, opened 2017-09-27
Update math.c

stuaxo:
Replace ob_item access in math.c with macros so pygame compiles in pypy.

(Thanks to Armin Rigo of Pypy for suggesting how to do this).

With this change, pygame seems to work in pypy.

I tried all the examples, the only problems seemed to be:

Errors on any keyboard events e.g.

```
  File "aliens.py", line 321, in <module>
    if __name__ == '__main__': main()
  File "aliens.py", line 257, in main
    (event.type == KEYDOWN and event.key == K_ESCAPE):
AttributeError: 'Event' object has no attribute 'key'
```

The fastevents example seemed to be frozen, all the examples displayed graphics and played sounds.

### DIFF
--- a/src/math.c
+++ b/src/math.c
@@ -801,7 +801,6 @@ vector_GetSlice(PyVector *self, Py_ssize_t ilow, Py_ssize_t ihigh)
     /* some code was taken from the CPython source listobject.c */
     PyListObject *slice;
     Py_ssize_t i, len;
-    PyObject **dest;
 
     /* make sure boundaries are sane */
     if (ilow < 0)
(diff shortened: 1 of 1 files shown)


stuaxo:
One failure is unrelated -

AssertionError: Failed to access the SoundFont /usr/share/sounds/sf2/FluidR3_GM.sf2


I'm not sure why the segfault in the math tests is happening, can't reproduce it myself right now.

stuaxo:
Have reproduced it..  I guess not knowing the CPython API doesn't help here..

stuaxo:
My lack of C knowledge showing there.

Now, all the math tests pass under pypy except 

```
.......................................................F...........................
FAIL: testSwizzle (__main__.Vector3TypeTest)
  File "test/math_test.py", line 1331, in testSwizzle
    self.assertRaises(AttributeError, invalidSwizzleW)
AssertionError: AttributeError not raised

Ran 83 tests in 0.020s

FAILED (failures=1)

```

NOTE - you need to first apply the cython patch in my other pull request for this to work.

illume:
PyList_SET_ITEM seems fine, since it's a new list.

It's interesting that it gets all the way to `self.assertRaises(AttributeError, invalidSwizzleW)` Since the tests just before that seem to do pretty much the same thing.

stuaxo:
Cheers.

PR #361, opened 2017-10-10
add unit test for some key functions

### LINKED ISSUE #309: SDL2: Function to get a key from name is needed.
Originally reported by: **Miloslav Číž (Bitbucket: [tastyfish](https://bitbucket.org/tastyfish), GitHub: [tastyfish](https://github.com/tastyfish))**

In key module, there is a function pygame.key.name() that gets the name from a key code, but there is no function that does the opposite, i.e. get the key code from name. It would be easy to implement, just call SDL_GetKeyFromName.

It is needed because when for example someone uses the name() function to store key mappings in readable format, it is hard to load the key mapping back.

- Bitbucket: https://bitbucket.org/pygame/pygame/issue/309

mweb:
Add unit tests for:

key.get_mods
key.set_mods
key.get_repeat
key.set_repeat
key.name

I wanted to resolve #309 but since the SDL_GetKeyFromName is not available within SDL 1.2 I just added the unit tests that I started when I looked into the issues. I had no idea how to implement the tests for the other functions.

### DIFF
--- a/test/key_test.py
+++ b/test/key_test.py
@@ -11,13 +11,21 @@ else:
     is_pygame_pkg = __name__.startswith('pygame.tests.')
 
 import unittest
+import pygame
+import pygame.key
 
 
 class KeyModuleTest(unittest.TestCase):
+    def setUp(self):
+        pygame.init()
+
+    def tearDown(self):
+        pygame.quit()
+
     def test_import(self):
         'does it import'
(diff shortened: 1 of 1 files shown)


takluyver:
Reopening to re-run tests.

takluyver:
Thanks!

PR #366, opened 2017-11-29
Update to latest terryfy commit

takluyver:
Trying to fix Mac builds on Travis

### DIFF
--- a/.travis_osx.sh
+++ b/.travis_osx.sh
@@ -4,7 +4,7 @@ set -e
 git clone https://github.com/MacPython/terryfy.git
 cd terryfy
 # Work with a specific commit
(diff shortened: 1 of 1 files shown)


takluyver:
Recording what's going on:

- The `beta-xcode6.1` images we were using no longer exist, and now appear to use XCode 8.3. We want to use the oldest version of OS X available, because the version used to compile pygame is the oldest version it can be used on.
- Next I tried the `xcode6.4` image, the oldest listed [in Travis' docs](https://docs.travis-ci.com/user/reference/osx/#OS-X-Version). However, homebrew decides to build a lot of stuff from source (including cmake and python), and the build reaches the 45 minute timeout before finishing.
- I'm currently trying the `xcode7.3` image, the next one listed, which is also Travis' default.

PR #381, opened 2018-01-25
Update rect.rst

raymonshansen:
Adds: Extended comment regarding behavior of rect.inflate methods.

Spent two hours trying to figure out why my rects behaved strangely when resizing by -1.

### DIFF
--- a/docs/reST/ref/rect.rst
+++ b/docs/reST/ref/rect.rst
@@ -110,7 +110,8 @@
 
       Returns a new rectangle with the size changed by the given offset. The
       rectangle remains centered around its current center. Negative values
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks - that seems like a useful thing for the docs to note. Could I ask you to have another look at the wording, though? I think it could be clearer. 

I see you're from Norway, so English probably isn't your first language - if you're having trouble finding the right words, let me know and I can have a go.

illume:
Thanks! 👍 It's nice to document this oddity for others.

The behaviour really does seem like a bug we should fix (one day... lol).

PR #384, opened 2018-02-06
Fix repetition of `current_h` attribute in documentation

hjpotter92:
The docs for `pygame.display.Info`s `VideoInfo` object's attributes
listed `current_h` twice, where one of them should be `current_w`
instead.

Signed-off-by: hjpotter92 <hjpotter92+github@gmail.com>


illume:
Thanks 🎉

Issue #372, opened 2017-12-16
surface_test.AllTestCases times out on sh4


Natureshadow:
From https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=861256

```
ERROR: all_tests_for (pygame.tests.surface_test.AllTestCases)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/pygame.tests.surface_test.py", line 1, in all_tests_for
subprocess completely failed with return code of "Process timed out (time_out = 120 secs) and was successfully terminated"
cmd:          ['/usr/bin/python3.5', '-m', 'pygame.tests.test_utils.test_runner', 'pygame.tests.surface_test', '--exclude', 'opengl,interactive,subprocess_ignore,python3_ignore']
test_env:     [...]
working_dir:  /tmp/tmpqaf65cf7
return (first 10 and last 10 lines):
WARNING: You are using the SDL disk writer audio driver!
 Writing to file [sdlaudio.raw].
WARNING: You are using the SDL disk writer audio driver!
 Writing to file [sdlaudio.raw].
WARNING: You are using the SDL disk writer audio driver!
 Writing to file [sdlaudio.raw].
WARNING: You are using the SDL disk writer audio driver!
 Writing to file [sdlaudio.raw].
WARNING: You are using the SDL disk writer audio driver!
 Writing to file [sdlaudio.raw].
...
WARNING: You are using the SDL disk writer audio driver!
 Writing to file [sdlaudio.raw].
WARNING: You are using the SDL disk writer audio driver!
 Writing to file [sdlaudio.raw].
WARNING: You are using the SDL disk writer audio driver!
 Writing to file [sdlaudio.raw].
/usr/lib/python3.5/runpy.py:125: RuntimeWarning: 'pygame.tests.test_utils.test_runner' found in sys.modules after import of package 'pygame.tests.test_utils', but prior to execution of 'pygame.tests.test_utils.test_runner'; this may result in unpredictable behaviour
  warn(RuntimeWarning(msg))
loading pygame.tests.surface_test
<--!! TES
```

illume:
According to here: https://buildd.debian.org/status/package.php?p=pygame
It looks like this one has been built successfully now with the 1.9.3+dfsg-2 package on sh4.

Issue #371, opened 2017-12-16
BufferProxyLegacyTest.test_write fails on 64-bit big endian


Natureshadow:
From https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=861253

```
  ERROR: test_write (pygame.tests.bufferproxy_test.BufferProxyLegacyTest)
  ----------------------------------------------------------------------
  Traceback (most recent call last):
    File "/<<BUILDDIR>>/pygame-1.9.3+dfsg/.pybuild/pythonX.Y_2.7/build/pygame/tests/bufferproxy_test.py", line 469, in test_write
      bp.write(data[:3], 2)
  IndexError: 'offset' is out of range
```

illume:
Note: `Version: 1.9.1release+dfsg-10+b2` in the failure linked email. I can't reproduce it on ppc64 with current git master.

Issue #370, opened 2017-12-16
some color-handling tests fail on ppc64el


Natureshadow:
From https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=861252

```
  ======================================================================
  FAIL: test_save_colorkey (pygame.tests.image_test.ImageModuleTest)
  make sure the color key is not changed when saving.
  ----------------------------------------------------------------------
  Traceback (most recent call last):
    File "/<<BUILDDIR>>/pygame-1.9.3+dfsg/.pybuild/pythonX.Y_2.7/build/pygame/tests/image_test.py", line 283, in test_save_colorkey
      self.assertEqual(p1, s2.get_at((0,0)))
  AssertionError: (23, 23, 23, 255) != (23, 23, 255, 23)
  
  ======================================================================
  FAIL: test_cross (pygame.tests.math_test.Vector2TypeTest)
  [as per #861249]
  
  ======================================================================
  FAIL: test_blanket_alpha (pygame.tests.surface_test.SurfaceSelfBlitTest)
  ----------------------------------------------------------------------
  Traceback (most recent call last):
    File "/<<BUILDDIR>>/pygame-1.9.3+dfsg/.pybuild/pythonX.Y_2.7/build/pygame/tests/surface_test.py", line 2331, in test_blanket_alpha
      self._assert_same(surf, comp)
    File "/<<BUILDDIR>>/pygame-1.9.3+dfsg/.pybuild/pythonX.Y_2.7/build/pygame/tests/surface_test.py", line 2252, in _assert_same
      a.get_bitsize())))
  AssertionError: (255, 0, 0, 255) != (0, 255, 170, 255), bpp: 8
  
  ======================================================================
  FAIL: test_colorkey (pygame.tests.surface_test.SurfaceSelfBlitTest)
  ----------------------------------------------------------------------
  Traceback (most recent call last):
    File "/<<BUILDDIR>>/pygame-1.9.3+dfsg/.pybuild/pythonX.Y_2.7/build/pygame/tests/surface_test.py", line 2311, in test_colorkey
      self._assert_same(surf, comp)
    File "/<<BUILDDIR>>/pygame-1.9.3+dfsg/.pybuild/pythonX.Y_2.7/build/pygame/tests/surface_test.py", line 2252, in _assert_same
      a.get_bitsize())))
  AssertionError: (255, 0, 0, 255) != (0, 255, 0, 255), bpp: 8
```

illume:
[...]

Issue #369, opened 2017-12-16
test_get_bounding_rect fails on big-endian


Natureshadow:
From https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=861250

```
  FAIL: test_get_bounding_rect (pygame.tests.surface_test.SurfaceTypeTest)
  ----------------------------------------------------------------------
  Traceback (most recent call last):
    File "/«BUILDDIR»/pygame-1.9.3+dfsg/.pybuild/pythonX.Y_2.7/build/pygame/tests/surface_test.py", line 154, in test_get_bounding_rect
      self.assertEqual(bound_rect.width, 0)
  AssertionError: 1 != 0
```

PR #391, opened 2018-02-10
ppc64le and other Debian fixes

illume:
- [x] [math_test.Vector2TypeTest.test_cross fails](https://github.com/pygame/pygame/issues/368)
- [x] [color surf tests on little endian](https://github.com/pygame/pygame/issues/370) (skip failing tests on ppc64le, SDL issue)
- [x] [get_bounding_rect test on big endian](https://github.com/pygame/pygame/issues/369)
- [x] [BufferProxyLegacyTest.test_write fails on 64-bit big endian](https://github.com/pygame/pygame/issues/371) (looks like a 1.9.1 bug, can't reproduce in current tree, with ppc64).
- [x] [surface_test.AllTestCases times out on sh4](https://github.com/pygame/pygame/issues/372) ([1.9.3 package now builds](https://buildd.debian.org/status/package.php?p=pygame) on Debian build infrastructure)

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -30,6 +30,7 @@ from pygame.locals import *
 from pygame.compat import xrange_, as_bytes, as_unicode
 from pygame.bufferproxy import BufferProxy
 
+import platform
 import gc
 import weakref
 import ctypes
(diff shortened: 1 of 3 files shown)


test/surface_test.py
```diff
@@ -2292,6 +2293,10 @@ def test_overlap_check(self):
     def test_colorkey(self):
         # Check a workaround for an SDL 1.2.13 surface self-blit problem
         # (MotherHamster Bugzilla bug 19).
+
+        if 'ppc64le' in platform.uname():
+            # skip https://github.com/pygame/pygame/issues/370#issuecomment-364625291
+            return
```
takluyver: You can explicitly skip a test, so that it's clear in the output that it was skipped: https://docs.python.org/3/library/unittest.html#unittest.TestCase.skipTest

takluyver:
👍

PR #403, opened 2018-02-24
pygame.math.Vector2/3 not experimental

illume:
- pickle/unpickle
- swizzle on by default
- speed up swizzle access.
- swizzle returns a Vector2 or Vector3 if of length 2 or 3.
- more tests, removed some DeprecationWarning in tests.
- removed experimental notice.

### DIFF
--- a/docs/py-modindex.html
+++ b/docs/py-modindex.html
@@ -4,7 +4,7 @@
 
 <html xmlns="http://www.w3.org/1999/xhtml">
   <head>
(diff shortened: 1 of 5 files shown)


illume:
I won't move pygame.math.Vector2/3 into pygame or pygame.locals namespace for now. That's potentially breaking things which already have their own Vector2/3. This type of API change will wait for a pygame2.

Narrative documentation is missing, but that will come in a separate PR.

bitcraft:
Thanks for all your work!

PR #408, opened 2018-03-01
pygame.transform.threashold tests, keyword arguments, docs.

illume:
"Threshold does not take keyword arguments"
https://www.reddit.com/r/pygame/comments/80zyqg/threshold_does_not_take_keyword_arguments/

Identifies a couple of problems:
- [x] no test for inverse argument.
- [x] argument names are super confusing and the documentation could be better.
- [x] no test for keyword arguments.
- [x] keyword arguments aren't supported.
- [x] Using the 'inverse' argument with the last surface argument set to None is not possible.
- [x] lots of invalid argument combinations were possible.
- [x] unit tests made more readable (using kwargs).
- [x] new documentation.

### DIFF
--- a/docs/reST/ref/transform.rst
+++ b/docs/reST/ref/transform.rst
@@ -199,6 +199,24 @@ Instead, always begin with the original image and scale to the desired size.)
 
 .. function:: threshold
 
+
+   TODO: new names and a better explanation.
+        # dest_surf                  # Surface we are changing. See 'set_behavior'.
+        #                            # None - if counting (set_behavior is 0), don't need 'dest_surf'.
+        # surf                       # Surface we are looking at.
+        # search_color               # Color we are searching for.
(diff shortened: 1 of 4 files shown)


illume:
Was thinking of constants for this...

pygame.transform.THRESHOLD_USE_SET_COLOR == 1
pygame.transform.THRESHOLD_JUST_COUNT == 0 
pygame.transform.THRESHOLD_USE_SURFACE == 2

```
+        # set_behavior = 1           # 1 - pixels in dest_surface will be changed to 'set_color'.
+        #                            # 0 - we do not change 'dest_surf', just count. Make dest_surf=None.
+        #                            # 2 - pixels set in 'dest_surf' will be from 'surface'. 
```

But then I thought just using keyword arguments would be cleaner.
```
use_set_color = True
just_count = True
use_surface_colors = True
```

Unfortunately we have to keep backwards compatibility. So, no extra keyword arguments.

PR #414, opened 2018-03-05
Mac Travis homebrew fix

illume:
Homebrew updated the pythons. Both python 2 and python 3 builds needed updating.

See:
- https://github.com/Homebrew/brew/pull/3877 
- Sent a PR to https://github.com/MacPython/terryfy/issues/16

cc @takluyver


takluyver:
Thanks for taking care of it!

PR #444, opened 2018-04-13
Add message for compilation and asking for help.

illume:


Brambor:
Didn't work it only annoys people:
https://stackoverflow.com/questions/51464455/why-when-import-pygame-it-prints-the-version-and-welcome-message-how-delete-it

MyreMylar:
I disagree, it works fine, and helps many people come to contribute to pygame. The message is also optional and in recent versions also includes much useful version information which frequently helps us with bug reports from inexperienced users - but thank you for your feedback.

Brambor:
Maybe it could link how to disable itself?

MyreMylar:
There is an issue requesting documentation for the pygame and SDL environment variables here that I think is a preferable solution:

https://github.com/pygame/pygame/issues/1364

If you are motivated to help perhaps this would be a good issue for you to contribute on?

PR #460, opened 2018-06-09
Updated framework intro

orangudan:
Updated framework intro snippet so it works out of the box if you paste the code and the ball image onto your computer.

Otherwise its a bit jarring for any novices thinking of using pygame?


takluyver:
Looks right,  thanks.

Issue #390, opened 2018-02-10
pygame 1.9.4 release


illume:
Please leave a comment if you have done some testing, and what you tested.

- [x] go through the issue tracker, and pull requests to include as much as we can.
- [x] announce intention to release on mailing list.
- [x] look at [distro bug trackers](http://www.pygame.org/wiki/patchesandbugs#OS/Distribution%20bug%20trackers%20for%20pygame) for bugs, and patches (debian, fedora, etc).
  - [x] address debian issues
  - [x] look through ubuntu issues
  - [x] [fedora issues](https://bugz.fedoraproject.org/pygame)
- [x] docs are building and ok.
- [x] make raspberry wheels: https://github.com/bennuttall/piwheels/issues/52
- [x] ~~high cpu sound issue~~ issue identified, reported upstream https://github.com/pygame/pygame/issues/331
- [ ] timidity config paths issue https://github.com/pygame/pygame/issues/343
- [x] Two [source alpha](https://bitbucket.org/pygame/pygame/issues/340/surfaceconvert-does-not-remove-the), [issues](https://bitbucket.org/pygame/pygame/issues/341/surfaceset_alpha-sets-the-srcalpha-flag) 
- [x] add a thank you note to website for people who made commits in this release.
- [x] manual testing on linux
- [x] manual testing on mac
- [x] manual testing on windows
- [x] run a few old games you know to see if things work (solar wolf, zanthor, ...).
- [x] prepare the release notes (aka, WHATSNEW).
- [x] skip known pypy failing tests on pypy.
- [x] add python 3.7 builds on 
  - [x] appveyor (windows py3.7 builds)
  - [x] travisci (mac, linux py3.7 builds)
- [x] prepare website news post.
- [x] improve 'contribute' web page.
- [x] release candidate, post to mailing list, give people a week.
- [ ] pre-release wheels, dev builds, release candidates. https://github.com/pygame/pygame/issues/409
- [x] ask for people to test from test.pypi.org
- [x] upload release wheels, and source to pypi.
- [x] upload source release to pygame.org/ftp
- [x] update `downloads.shtml` page on the website with new source release.
- [x] announce on pygame website news section.
- [x] announce on mailing lists (pygame-users, python announce).
- [x] announce on pygame reddit.

illume:
[...]

PR #469, opened 2018-07-16
Updated docker vagrant install instructions for manylinux build.

illume:
Updating instructions for manylinux build as I go. So far docker instructions changed since last year.
- [x] freetype pkg-config does not work, fallback to freetype-config.
- [x] update docker install instructions on vagrant.
- [x] upgrade manylinux to have python3.7 (cp37-cp37m)

re: https://github.com/pygame/pygame/issues/390

### DIFF
--- a/config_darwin.py
+++ b/config_darwin.py
@@ -98,6 +98,25 @@ class DependencyPython:
         else:
             print (self.name + '        '[len(self.name):] + ': not found')
 
+def find_freetype():
+    """ modern freetype uses pkg-config. However, some older systems don't have that.
+    """
+    pkg_config = DependencyProg(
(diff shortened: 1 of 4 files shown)


illume:
Looks like this freetype pkg-config change broke the manylinux build: https://github.com/pygame/pygame/pull/442
```
    WARNING: "pkg-config freetype2" failed!
```

illume:
@takluyver do you remember how to upload an updated image to the docker hub? I had to do a `make pull` for the manylinux python 3.7

takluyver:
There's a `make push` target in the makefile which should do it. The actual commands it uses look like `docker push pygame/manylinux1_base_x86_64` .

illume:
Ah ha! Thanks. I also had to do `docker login`, and the `make push` ran. Looks like https://hub.docker.com/u/pygame/ has the new images now.

Issue #473, opened 2018-07-20
FreeBSD build issue


ndowens:
Using UNIX configuration...

Hunting dependencies...
SDL     : found 1.2.15
FONT    : found
IMAGE   : found
MIXER   : found
PNG     : found
JPEG    : found
SCRAP   : found
Traceback (most recent call last):
  File "config.py", line 173, in <module>
    if __name__ == '__main__': main()
  File "config.py", line 161, in main
    deps = CFG.main()
  File "/wrkdirs/usr/ports/devel/py-pygame/work-py36/pygame-1.9.4/config_unix.py", line 232, in main
    d.configure(incdirs, libdirs)
AttributeError: 'NoneType' object has no attribute 'configure'
*** Error code 1

Stop.

illume:
Thanks for the report!

Sorry, I didn't get a chance to try 1.9.4 with freebsd before the release. We don't have a freebsd maintainer anymore. Also we don't have any build robots checking things in Continuous integration for freebsd (Travis CI and Appveyor don't support it).

cheers,

ndowens:
Any idea how to fix it?  What I could change to make a patch for it. I’m not a programmer so I’m not sure how to fix issue

ndowens:
All I know is that someone value of None is being returned

ndowens:
I believe this to be a possible issue on other platforms than just FreeBSD. For some reason configure seems to be returning None or d is returning/receieving None

ndowens:
I have fixed it. Will link to patches. Though in summary it’s about the code to check if portaudio is portmidi part

ndowens:
https://pastebin.com/PwFBbeT7 
https://pastebin.com/UJzFGetL

ndowens:
I was thinking, maybe you could add if os=freebsd skip that section I had to delete?

illume:
Thanks for that :) Ah, it seems that portmidi does not exist for freebsd (from my limited searching). I've made a patch. Can you please try that?

ndowens:
Oh another I had to fix was I think in config where it says if linux import extra file that contains camera because it has import <linux dir/file.h>

ndowens:
**src/_camera.c:213:36: error: use of undeclared identifier 'V4L2_CID_HFLIP'
    if (v4l2_get_control(self->fd, V4L2_CID_HFLIP, &value))
                                   ^
src/_camera.c:216:36: error: use of undeclared identifier 'V4L2_CID_VFLIP'
[...]

PR #478, opened 2018-07-25
video4linux header for linux only. freebsd no have.

illume:
I don't have freebsd to test this... but apparently there is no video4linux header on there. Was there once one? The comment in the file seems to suggest so. A bit of searching I can't find it.

For https://github.com/pygame/pygame/issues/473


illume:
This package is needed: `sudo pkg install v4l_compat-1.6.3_1`

Issue #411, opened 2018-03-03
Pygame crashes when scaling a zero area Surface


Yatima1460:
- [x] turn the code below into a unit test in tests/surface_test.py (**good first issue**).
- [ ] check the arguments for transform.scale in src_c/transform.c and raise an exception with a good error message.

---
```
import pygame

pygame.init()
pygame.display.set_mode((1280, 720))
tmp_surface = pygame.transform.scale(pygame.Surface((128, 128)), (0, 0))
tmp_surface = pygame.transform.scale(tmp_surface, (128, 128))
```

Shorter version:
```
import pygame
pygame.transform.scale(pygame.Surface((0, 0)), (1, 1))

```
Environment:
Pygame 1.9.3
Python 3.6.4 (v3.6.4:d48eceb, Dec 19 2017, 06:04:45) [MSC v.1900 32 bit (Intel)] on win32

augusto2112:
@illume  Hi, I wrote a test for this. Currently the test crashes with a segmentation fault (because of the underlying issue). Is that ok?

illume:
Yes, that's perfect! Well done for crashing it :)

Yatima1460:
I'm actually surprised no one found about this before.
I found this because I was writing a UI library for my game and progressbars would degenerate to 0 when the player had 0 health.
Good thing this was fixed, thanks.

PR #489, opened 2018-08-16
transform.scale does not crash for zero sized surfaces.

illume:
Thanks for the unit test @augusto2112 

- This fixes https://github.com/pygame/pygame/issues/411
- I made a mistake in the issue description for 411, as the transform.scale test should have been put into test/transform_test.py So I moved the test in there.

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -547,7 +547,7 @@ surf_scale (PyObject* self, PyObject* arg)
         return RAISE (PyExc_ValueError,
                       "Source and destination surfaces need the same format.");
 
-    if (width && height)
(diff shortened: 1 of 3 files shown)


illume:
Oh, sorry about that! Did you get things to compile in the end? The instructions here should help. https://www.pygame.org/wiki/MacCompile Let me know if you have any issues, or if you want to get stuck into another C issue, let me know, and I can look for one for you?

augusto2112:
No need to apologize!

I didn't manage to get it to compile yet, but I'm at work now so I'll continue trying tonight.

After I get it to run I'll let you know if I can't find an issue.

Thanks for all the help :smile:

Issue #491, opened 2018-08-17
mask.from_surface() on a Surface whose width or height is 0


illume:
Steps to fix this problem.
- [x] convert the example code below into a unit test inside [tests/mask_test.py](https://github.com/pygame/pygame/blob/master/tests/mask_test.py) (**good first issue**)
- [x] fix from_surface in [src_c/mask.c](https://github.com/pygame/pygame/blob/master/src_c/mask.c)

Documentation for [pygame.mask.from_surface](https://www.pygame.org/docs/ref/mask.html#pygame.mask.from_surface)

---

mask.from_surface() on a Surface whose width or height is 0, you get a scary looking error:
```
python(2730,0x7fff96248380) malloc:  mach_vm_map(size=4611686018427392000) failed (error code=3)
 error: can't allocate region
*** set a breakpoint in malloc_error_break to debug
```

Reproduce with:
```python
import pygame; pygame.init()
pygame.mask.from_surface(pygame.Surface((0, 100)))
```

shakirck:
can i work on this

illume:
@shakirck this one is taken by @augusto2112 

There's some more [good first issue](https://github.com/pygame/pygame/labels/good%20first%20issue) ones though.

PR #493, opened 2018-08-17
0 sized masks are now possible

augusto2112:
Issue: https://github.com/pygame/pygame/issues/491

### DIFF
--- a/src_c/bitmask.c
+++ b/src_c/bitmask.c
@@ -977,11 +977,6 @@ bitmask_t *bitmask_scale(const bitmask_t *m, int w, int h)
     return 0;
   }
 
-  if (!m->w || !m->h || !w || !h) {
-    nm = bitmask_create(0, 0);
-    return nm;
-  }
-
   nm = bitmask_create(w,h);
 
   if (!nm)
--- a/src_c/bitmask.c
+++ b/src_c/bitmask.c
@@ -88,9 +88,18 @@ static INLINE unsigned int bitcount(BITMASK_W n)
 bitmask_t *bitmask_create(int w, int h)
 {
   bitmask_t *temp;
-  temp = malloc(offsetof(bitmask_t,bits) + h*((w - 1)/BITMASK_W_LEN + 1)*sizeof(BITMASK_W));
-  if (!temp)
+  size_t size;
+
+  size = offsetof(bitmask_t,bits);
+  if (w && h) {
+    size += h * ((w - 1) / BITMASK_W_LEN + 1) * sizeof(BITMASK_W);
+  }
+
+  temp = malloc(size);
+
+  if (!temp) {
     return 0;
+  }
   temp->w = w;
   temp->h = h;
   bitmask_clear(temp);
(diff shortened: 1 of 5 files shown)


illume:
Looks like a few builds are failing. You can see them by clicking the "Details" link next to Appveyor and TravisCI.

Here's one issue on a 32bit python: https://ci.appveyor.com/project/pygame/pygame/build/1.0.584/job/wi2o2m3smiuim2wq
```
ERROR: test_zero_size_from_surface (pygame.tests.mask_test.MaskModuleTest)
  File "C:\Python27\lib\site-packages\pygame\tests\mask_test.py", line 460, in test_zero_size_from_surface
    zero_w_mask = pygame.mask.from_surface(pygame.Surface((0, 100)))
ValueError: Cannot create mask with negative size
```

src_c/mask.c
```diff
@@ -467,6 +470,11 @@ static PyObject* mask_from_surface(PyObject* self, PyObject* args)
         return NULL;
     }
 
+    if (surf->w < 0 || surf->h < 0) {
```
illume: This check should be below `surf =`.

src_c/bitmask.c
```diff
@@ -88,9 +88,18 @@ static INLINE unsigned int bitcount(BITMASK_W n)
 bitmask_t *bitmask_create(int w, int h)
 {
   bitmask_t *temp;
-  temp = malloc(offsetof(bitmask_t,bits) + h*((w - 1)/BITMASK_W_LEN + 1)*sizeof(BITMASK_W));
-  if (!temp)
+  size_t size;
+
+  size = offsetof(bitmask_t,bits);
+  if (w && h) {
```
illume: For the case where w<=0, or h<=0 this check would pass and the size would be a weird amount.  But I guess we check that in the mask.c.

augusto2112:
@illume I'll do it tomorrow though because today I won't have enough time. Or you prefer, you can accept it and tomorrow I submit a new PR with the tests.

augusto2112:
@illume I added a bunch of tests. I'm not sure what should happen when trying to scale a (0, 100) mask by (0, 2) for example. Currently it stays (0, 100).

test/mask_test.py
```diff
@@ -452,10 +458,167 @@ def test_from_threshold(self):
...
+            mask = pygame.mask.Mask(size)
+            mask.invert()
+            self.assertEqual(mask.count(), 0)
 
+    def test_zero_mask_scale(self):
+        sizes = ((100, 0), (0, 100), (0, 0))
+
+        for size in sizes:
+            mask = pygame.mask.Mask(size)
+            mask.scale((2, 3))
```
illume: mask.scale should return a new mask with the requested size.
https://www.pygame.org/docs/ref/mask.html#pygame.mask.Mask.scale

So the test should be something like...
```python
scaled_mask = mask.scale((2, 3))
self.assertEqual(scaled_mask.get_size(), (2, 3))
```

Issue #410, opened 2018-03-02
Mask.overlap_mask broken


Rybec:
Mask.overlap_mask is not working properly in some cases.

Steps to fix this issue.

- [x] convert the example code below into a unit test inside [tests/mask_test.py](https://github.com/pygame/pygame/blob/master/tests/mask_test.py) (**good first issue**)
- [ ] fix overlap_mask in [src_c/mask.c](https://github.com/pygame/pygame/blob/master/src_c/mask.c)

Documentation for [pygame.mask.Mask.overlap_mask](https://www.pygame.org/docs/ref/mask.html#pygame.mask.Mask.overlap_mask)

---

https://groups.google.com/forum/#!topic/pygame-mirror-on-google-groups/8RGH0HIg2o8

I described the issue in more detail than the original poster at the end of that thread.

I'll see if I can get the student experiencing the problem to post some code here and describe the issue.

Rybec:
There is a possibility that this is not a problem with masks.  I may be an issue with dynamically altering image surfaces in Pygame.  Unfortunately, it is impossible to render a mask directly, so the program in question is using Surface.set_at() to change pixels.  According to some sources, this is known to be buggy.

I'll try to write up a program to demonstrate the issue in the next day or so.

Rybec:
Alright, code:

```
import pygame

pygame.display.init()

screen = pygame.display.set_mode((400, 300))

background = pygame.Surface((400, 300), flags=pygame.SRCALPHA)
image = pygame.Surface((50, 50), flags=pygame.SRCALPHA)


#####
# Dynamically creating some images, so we don't have to deal with image files.
background.fill((0, 0, 0, 0))
image.fill((0, 0, 0, 0))

pa_background = pygame.PixelArray(background)
pa_image = pygame.PixelArray(image)

for y in xrange(100, 200):
	for x in xrange(100, 300):
		if (((x - 100) / 2) + (y - 100)) >= 100:
			pa_background[x, y] = (255, 255, 0)

for y in xrange(5, 45):
	for x in xrange(5, 45):
		pa_image[x, y] = (255, 127, 0)

del(pa_background)
del(pa_image)
#
#####


# Make the masks
bg_mask = pygame.mask.from_surface(background)
im_mask = pygame.mask.from_surface(image)


pygame.mouse.set_visible(False)

run = True
while run:
	for e in pygame.event.get():
		if e.type == pygame.QUIT:
			run = False
	
	mouse = pygame.mouse.get_pos()
[...]

PR #500, opened 2018-08-22
Test for overlap mask 

augusto2112:
Added a test for overlap mask that demonstrates at least one case where overlap mask currently fails.


augusto2112:
Issue: https://github.com/pygame/pygame/issues/410

illume:
Because it's a known failing test, can we please mark it with an ignore tag?
```python
        """ |tags: ignore| 
        """ 
```

This way it still fails if we run python test/mask_test.py but it will pass on CI. Because it could take some time for the fix for the issue to be committed, we still want to be able to have the CI do builds and only fail if tests we know should pass fail.

illume:
Thanks @augusto2112 :)

Issue #233, opened 2015-02-15
pygame.draw.ellipse does not draw ellipses with odd sizes


illume:
Steps to fix this issue of "pygame.draw.ellipse not being able to draw an odd sized ellipse".

- [x] ~~write a unit test that shows the problem of odd sizes not being possible inside [tests/draw_test.py](https://github.com/pygame/pygame/blob/master/tests/draw_test.py)~~ (**good first issue** DONE)
- [x] fix draw ellipse in [src_c/draw.c](https://github.com/pygame/pygame/blob/master/src_c/draw.c) (harder, not a good first issue)

https://www.pygame.org/docs/ref/draw.html#pygame.draw.ellipse

---

Originally reported by: **donLorenzo (Bitbucket: [donLorenzo](https://bitbucket.org/donLorenzo), GitHub: [donLorenzo](https://github.com/donLorenzo))**

----------------------------------------

Reported by Florian Krause on the [pygame mailing list](http://article.gmane.org/gmane.comp.python.pygame/25539):

```
#!plain

I was actually wrong. It seems even worse. Pygama is not capable to draw ellipses with odd sizes at all. It will always be the same ellipse as x-1, y-1.
E.g. an elipse with size 3,3 will result in the same ellipse as the one with size 2,2.
Can anyone reproduce this?
```

Code to reproduce:

```
#!python

import pygame
pygame.init()
screen = pygame.display.set_mode((100, 100))
pygame.draw.ellipse(screen, (255,0,0), (0,0,5,5))
pygame.display.update()
```
The resulting ellipse will have a 4x4 size

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/233

Earlo:
Did #511 fix this? Should this be closed?

charlesej:
Yes and yes. This issue has been resolved by PR #511.

PR #501, opened 2018-08-22
Add draw_ellipse unit test #233

amosbastian:
The unit test creates a surface of a certain width and height, then creates an ellipse with the same width and height on the surface. It then creates four lists containing the values of the top, left, right and bottom borders respectively. Finally it checks if each of these lists contains the colour of the ellipse, and if they don't then the test fails.

### DIFF
--- a/test/draw_test.py
+++ b/test/draw_test.py
@@ -240,21 +240,48 @@ class DrawModuleTest(unittest.TestCase):
           # filled.
           # 
 
-        surface = pygame.display.set_mode((201, 401))
+        sizes = [(4, 4), (5, 4), (4, 5), (5, 5)]
+        color = (1, 13, 24, 255)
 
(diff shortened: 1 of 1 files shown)


amosbastian:
I removed the `display.set_mode` and added another loop so it checks ellipses with a border width of 1, but also filled ellipses. Now both `draw_ellipse` and `draw_fillellipse` are covered by this unit test.

illume:
Nice one :) Thanks.

Aha! I see what the problem is. If you click on the travis ci ones that fail you can see the error.

The print statement needs to be removed. The tests shouldn't have print in them.
The fstrings it uses is in python 3.6+ only.

illume:
Oh. I forgot to mention...

People usually put a link to the issue in the description of the PR. So that people looking at the PR can more easily click on the issue.

Issue #496, opened 2018-08-21
Test suite : documentation and cleanup ?


e1000:
Hello,

I tried to write some tests, but I found the test suite a bit confusing ;

First, the test code seems to follow a bit the principle of the *broken window* ;-) : 

[edit : transformed this list into a check lst]

- [x] some 15 occurences of the strange pattern `if 1: <code_block>` 
- [ ] cleanup many deprecated legacy `assert*` and `fail*` methods
- [x] cleanup *sndarray_test* : `arraytype` is always `= "numpy"` (l. 20), but many there are many occurences of `if [not] arraytype: ...`
- [ ] *image_test.py* : use the **tempfile** module instead of `finally: os.remove` 
- [x] *mixer_test.py* : use `assertRaises` instead of `try/except/else`
- [x] use `setUp` and `tearDown` instead of `try/finally` like in *midi_test.py* (maybe splitting some *TestCases* in different classes)
- [ ] Need of big PEP8 cleanup (white space, unused imports, commented code that seems to have no meaning, etc )

[edit] ~~If~~ we support only Python >= 2.7 , so we ~~could~~ can

- [x] use  `unittest.skip` and  `unittest.skipIf` decorators instead of indentation (cf *test/surflock_test.py*) etc
- [x] use of `setUpClass` and `tearDownClass` where it makes sense
- [x] use `with self.assertRaises` 

I'd also like some documentation of how the suite works, specially for following points:
 
* where does "*pygame.tests*" package come from and what's the meaning of *test* vs *pygame.tests* import ? Related to that, the need and meaning of **is_pygame_pkg** ?
 Why can't we use the standard `python -m unittest test_module.TestCase` ?
* why do we copy-paste the docstring of the methods into the test suite ?  (the *docstring* of the methods might evolve, while these commmented docstrings are 10 years old; 
  copying each time seems strange & the developer can look it up anyway...)
* how does the *sth_tags.py* work ?
* The *midi_test.py* uses in test method *docstrings* sth like `|tags: interactive|` ; what's the
  meaning of it?
* what is the meaning of `test/util/build_page/` ? 

I could do the standard code cleanup, but it would be very noisy commits, possibly generating annoying merge conflicts. Which cleanups should be done or not ? One big test-cleanup Pull request ok ?

e1000:
[...]

PR #502, opened 2018-08-23
Cleanup some test methods & test/sndarray_test.py

e1000:
Hello,

here some first cleanup branch on tests : 

* replace all occurences of `failIf` with `assertFalse`
* replace some deprecated `failUnlessEqual` with standard `assertEqual`
* rm obsolete `if not arraytype ...` in **test/sndarray_test.py** & rm a test on import

I first fancied, I could cleanup all tests methods, but that would be way to noisy.
2 tests of **gfxdraw_test.py** fail, they fail also on master, right ?

Once merged, we can check the *test/sndarray_test.py* in #496


illume:
Thanks @e1000 

Yeah, those are disabled so that the tests can pass on the CI.

PR #504, opened 2018-08-24
Cleanup mixer test

e1000:
Some cleanup of the `mixer_test.py`  . Merge this only if/after dropping support for **Python < 2.7** !

* cleanup deprecated test method aliases
* some PEP8 cleanup
* use Contectmanager 'with self.assertRaises as ...' ( /!\ this implies **Python >= 2.7** /!\ )

We can then check the point in #496 ; later on , we might write some `setUp` and `tearDown` methods to clean the code further, but it would need some better understanding of what is tested and  what can be removed...


illume:
💥 Thanks! 💥

I asked in the discord contributing channel if anyone else wants to do a code review for practice at that.

e1000:
Hello,

what do you mean by "code review for practice at that" ? About dropping Python2.7 ? Or adding `setUp` and `tearDown` methods ?

illume:
Sorry for being unclear.

I meant that I asked someone to do a code review of this pull request so that they could get practice at doing pull requests :)

test/mixer_test.py
```diff
@@ -169,88 +162,65 @@ def get_bytes(snd):
...
-            try:
+
+            with self.assertRaises(TypeError) as cm:
                 mixer.Sound()
-            except TypeError:
-                self.assertEqual(str(geterror()), arg_emsg)
-            else:
-                self.fail("no exception")
-            try:
+                self.assertEqual(cm.exception, arg_emsg)
```
e1000: oups, I just realized, there is a mistake here (we never enter the `assertEqual`). It should be one indentation lower, and then it needs to be spelled like : 
```python
self.assertEqual(cm.exception.message, arg_emsg)
```
 (or `str(cm.exception)` ...)

illume:
Hello,

it looks like there's an issue on python 3s.

```python
ERROR: test_sound_args (pygame.tests.mixer_test.MixerModuleTest)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/pygame-1.9.5.dev0-py3.5-linux-x86_64.egg/pygame/tests/mixer_test.py", line 175, in test_sound_args
    self.assertEqual(cm.exception.message, arg_emsg)
AttributeError: 'TypeError' object has no attribute 'message'
```

e1000:
should be fixed by now...

PR #509, opened 2018-08-27
Midi tests : add, refactor, cleanup

e1000:
Hi, many new tests for midi: 

* the PR is a bit noisy, since I created two more test case classes **MidiInputTest** and **MidiOutputTest**, moving code around
* add a lot of tests (no more `def todo_test_sth` left)
* tried to get rid of the `Exception: b"PortMidi: Invalid MIDI message Data'"` messages during test run with *setUp* and *tearDown* (worked only partially)
* cleanup using **Python >= 2.7**
* also, I removed the `__doc__ as of 2009 ...` docstrings (noisy last commit to be possibly reverted)

I do not understand so much midi, and specially not understanding `midi.midis2events`  ...

Note: I also tested `pg.midi.Output.pitch_bend` while going through source code, but it is not mentionned in https://www.pygame.org/docs/ref/midi.html ; ommitted or intended ?

### DIFF
--- a/test/midi_test.py
+++ b/test/midi_test.py
@@ -72,6 +72,9 @@ class MidiInputTest(MidiTestBase):
             self.assertEqual(read, [])
             # TODO fake some  incoming data
 
+    def todo_test_close(self):
+        pass
+
 
 class MidiOutputTest(MidiTestBase):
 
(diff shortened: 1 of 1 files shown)


illume:
This is awesome :) Thank you!

- midis2events is for turning midi events into pygame events.
- oh yes, there is no documentation for pg.midi.Output.pitch_bend. This would need to be added in docs/reST/ref/midi.rst

Issue #395, opened 2018-02-14
blue lines drawn with aalines washed out


benjaminbradley:
ARGB Surfaces didn't work nicely with `draw.aaline`.

What's left to do:
- [x] a unit test or two in [test/draw_test.py](https://github.com/pygame/pygame/blob/master/test/draw_test.py#L166). Check ARGB and RGBA surfaces, with and without SRC_ALPHA. (**good first issue**)
- [x] check other platforms (the unit tests can do that for us mostly).
- [x] fix draw.c DRAWPIX32 to work with not RGBA format Surfaces.

---

Lines drawn with the pygame.draw.aalines() function seem to be washed out of the blue color. Here is a simple demo program to compare the color of lines drawn with pygame.draw.line() and pygame.draw.aalines(), both with the same color code, and a screencap of the difference on my monitor.

```
import pygame, colors, sys
image = pygame.display.set_mode((200,200))
image.fill((0,0,0))
pygame.draw.aalines(image, colors.BLUE, False, [(0,0), (100,100), (100, 200)])
pygame.draw.line(image, colors.BLUE, (0,100), (100,200), 2)
pygame.display.update()
while True:
	for event in pygame.event.get():
		if event.type == pygame.KEYDOWN and event.key == pygame.K_ESCAPE: sys.exit()
	pygame.time.Clock().tick(30)
```

![screen shot 2018-02-13 at 6 02 07 pm](https://user-images.githubusercontent.com/38849/36181099-e1abfe0e-10e8-11e8-9756-2940ce288660.png)

I'm running Python 3.6.2 and pygame 1.9.3 on Mac OS 10.12.6

illume:
Confirmed on 10.11.x. The blue component is getting drawn as black.

```
import pygame, sys
screen = pygame.display.set_mode((200,200))
offscreen = screen.convert()
going = True

while going:
    for event in pygame.event.get():
        if ((event.type == pygame.KEYDOWN and event.key == pygame.K_ESCAPE) or
            event.type == pygame.QUIT):
            going = False

    screen.fill((0,0,0))
    offscreen.fill((0,0,0))
    if 0:
        pygame.draw.aalines(screen, (0, 0, 255), False, [(0,0), (100,100), (100, 200)])
        pygame.draw.line(screen, (0, 0, 255), (0,100), (100,200), 2)
    else:
        pygame.draw.aalines(offscreen, (255, 255, 255), False, [(0,0), (100,100), (100, 200)])
[...]

PR #510, opened 2018-08-27
Update unit tests draw.ellipse, add unit tests draw.(aa)line(s)

amosbastian:
This pull request contains the following:

### Ellipse

I updated the draw.ellipse unit test so that it now also includes a test that checks if at least two sides of an ellipse, that has a width and height that is 1 pixel smaller than the surface, is touching the surface's border.

It also cleans up the code slightly, including a better way of retrieving the border's values and adds reusability in the form of additional functions.

See: #233

### Lines

It also adds unit tests for draw.aaline, draw.line, draw.aalines and draw.lines.

For draw.aaline and draw.line it now draws a line of differing colours on all types of surfaces and checks if the colour stays the same at position (0, 0). There is also a test that checks if the line contains any gaps, as I noticed aaline(s) sometimes does (or is not drawn at all when close to a surface's border).

The same is done for draw.aalines and draw.lines, except these tests draw a line around the border of the given surface then check if all pixels are of the given colour, or if they contain gaps.

See: #395

### DIFF
--- a/test/draw_test.py
+++ b/test/draw_test.py
@@ -1,6 +1,8 @@
-import unittest
+#################################### IMPORTS ###################################
 
 import pygame
+import unittest
+
 from pygame import draw
 from pygame.locals import SRCALPHA
 
@@ -8,6 +10,18 @@ if is_pygame_pkg:
     from pygame.tests import test_utils
 else:
     from test import test_utils
+
+if __name__ == '__main__':
+    import sys
+    import os
+    pkg_dir = os.path.split(os.path.abspath(__file__))[0]
+    parent_dir, pkg_name = os.path.split(pkg_dir)
+    is_pygame_pkg = (pkg_name == 'tests' and
+                     os.path.split(parent_dir)[1] == 'pygame')
+    if not is_pygame_pkg:
+        sys.path.insert(0, parent_dir)
+else:
+    is_pygame_pkg = __name__.startswith('pygame.tests.')
 
 
 def get_border_values(surface, width, height):
@@ -23,6 +37,68 @@ def get_border_values(surface, width, height):
         surface.get_at((x, height - 1)) for x in range(width)]
 
     return [border_top, border_left, border_right, border_bottom]
+
+
+class DrawEllipseTest(unittest.TestCase):
+    """
+    Class for testing ellipse().
+    """
+    def test_ellipse(self):
+        """ |tags: ignore|
+        Draws ellipses of differing sizes on surfaces of differing sizes and
+        checks to see if the number of sides touching the border of the surface
+        is correct.
+        """
+        # __doc__ (as of 2008-08-02) for pygame.draw.ellipse:
+
+          # pygame.draw.ellipse(Surface, color, Rect, width=0): return Rect
+          # draw a round shape inside a rectangle
+          # 
+          # Draws an elliptical shape on the Surface. The given rectangle is the
+          # area that the circle will fill. The width argument is the thickness
+          # to draw the outer edge. If width is zero then the ellipse will be
+          # filled.
+          # 
+
+        left_top = [(0, 0), (1, 0), (0, 1), (1, 1)]
+        sizes = [(5, 4)]
+        color = (1, 13, 24, 255)
+
+        def same_size(width, height, border_width):
+            """
+            Test for ellipses with the same size as the surface.
+            """
+            surface = pygame.Surface((width, height))
+
+            draw.ellipse(
+                surface, color, (0, 0, height, width), border_width)            
+
+            # For each of the four borders check if it contains the color
+            borders = get_border_values(surface, width, height)
+            for border in borders:
+                self.assertTrue(color in border)
+
+        def not_same_size(width, height, border_width, left, top):
+            """
+            Test for ellipses that aren't the same size as the surface.
+            """
+            surface = pygame.Surface((width, height))
+
+            draw.ellipse(surface, color, (left, top, width - 1, height - 1),
+                         border_width)
+
+            borders = get_border_values(surface, width, height)
+
+            # Check if two sides of the ellipse are touching the border
+            sides_touching = [
+                color in border for border in borders].count(True)
+            self.assertEqual(sides_touching, 2)
+
+        for width, height in sizes:
+            for border_width in (0, 1):
+                same_size(width, height, border_width)
+                for left, top in left_top:
+                    not_same_size(width, height, border_width, left, top)
 
 
 def lines_are_color(surface, color, line_type="lines"):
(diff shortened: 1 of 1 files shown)


test/draw_test.py
```diff
@@ -1,7 +1,325 @@
-import unittest
-from pygame.tests import test_utils
+#################################### IMPORTS ###################################
+
+if __name__ == '__main__':
```
illume: This is_pygame_pkg stuff should be removed.

The imports should be something like:
```python
import unittest

import pygame
from pygame import draw
from pygame.locals import SRCALPHA
from pygame.tests import test_utils
```

test/draw_test.py
```diff
@@ -1,7 +1,309 @@
...
+                         border_width)
+
+            borders = get_border_values(surface, width, height)
+
+            # Check if two sides of the ellipse are touching the border
+            sides_touching = [
+                color in border for border in borders].count(True)
+            self.assertEqual(sides_touching, 2)
+
+        for width, height in sizes:
```
e1000: a loop over `sizes` which contains only one element ? bit strange...

amosbastian: Thanks for pointing that out! I forgot I had removed the other sizes... added them back in.

e1000:
Personnally, I have some issues about *line_is_color* ,*line_has_gaps* and *lines_have_gaps* : 

1. all repeat the  `if line_type == "aalines":` , hence it seems to be repeated code
2. it mixes doing something and testing something , imho, that's bit confusing, it looks bit as you were testing your own test functions ;)
3. ,*line_has_gaps* and *lines_have_gaps* : with a quick look I can not understand what the gaps are ? And why does the equalities at the end mean, there is a gap (or are gaps) ?

So , for me, these helper functions contain a bit too much logic / factorisation ?

test/draw_test.py
```diff
@@ -1,7 +1,309 @@
...
+    # __doc__ (as of 2008-08-02) for pygame.draw.aalines:
+
+      # pygame.draw.aalines(Surface, color, closed, pointlist, blend=1): return Rect
+      # 
+      # Draws a sequence on a surface. You must pass at least two points in
+      # the sequence of points. The closed argument is a simple boolean and
+      # if true, a line will be draw between the first and last points. The
+      # boolean blend argument set to true will blend the shades with
+      # existing shades instead of overwriting them. This function accepts
+      # floating point values for the end points.
```
e1000: @illume : should we keep or should we delete  copies of 2008 docstrings ?

illume:
@e1000 some good suggestions. Thank you.

1. `line_is_color` and friends could be improved by taking a draw function as the argument instead of a string. Maybe call the argument `draw_line` and then pass in `draw.line` or `draw.aaline`.

2. Then when using line_is_color, the draw_line can be just another parameter in the one test function(test_line_color). This would reduce the number of tests you have too. Also the implementation of `line_is_color` could be moved inside that one test(because now there are not two). So it would be easier for the person reading it to see what it is doing. Because then `line_is_color` and `test_line_color` would fit in one screen.

These two changes would reduce the amount of repeated code (DRY) https://en.wikipedia.org/wiki/Don%27t_repeat_yourself And I think it would bring related test code closer together in the file, making it easier to understand.

3. Yeah, a different explanation of what gaps are would improve this. It would also help readers of the code to put a link to the issue in the docstring. So they could see the image of the issue themselves.

illume:
On Travis CI tests I see this error come up:

```
ERROR: test_line (pygame.tests.draw_test.DrawModuleTest)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/pygame-1.9.5.dev0-py2.7-linux-x86_64.egg/pygame/tests/draw_test.py", line 312, in setUp
    self.surf = pygame.Surface(self.surf_size, pygame.SRCALPHA)
ValueError: no standard masks exist for given bitdepth with alpha
```

So maybe we need to specify the bit depth for when the tests run like this.

amosbastian:
Thanks for the feedback @e1000 @illume, I just refactored everything and updated the `lines_set_up()` function so it creates surfaces of an even width/height and odd width/height, with differing depths (16 and 32). If there is anything else left to be improved then please let me know.

test/draw_test.py
```diff
@@ -1,7 +1,204 @@
...
+
+        def line_has_gaps(surface, draw_line):
+            """Returns True if the line drawn on the surface contains gaps.
+            """
+            width = surface.get_width()
+            color = (255, 255, 255)
+
+            draw_line(surface, color, (0, 0), (width - 1, 0))
+
+            colors = [surface.get_at((width, 0)) for width in range(width)]
```
e1000: Hello, thank you for modifications ! Some more remarks : 

1. it is not so nice to give same name for loop variable `width` and the outer `width` ; better use a single letter for loop variable in list-comprehension : 
```python
colors = [surface.get_at((x, 0)) for x in range(width)]
```
2. In the cases below, you can write `in_border.count(True) == 4` as `all(in_border)` which seems clearer to me.

3. Also, I still don't understand : for me reading the code, *line_has_gaps(* checks that there is *no* gap since _all_ are the same color ... (same remark for below *lines_have_gaps* ) ; could you please explain what the "gaps" are , or am I missing something ?

test/draw_test.py
```diff
@@ -1,7 +1,204 @@
...
+            """
+            width = surface.get_width()
+            height = surface.get_height()
+            points = [(0, 0), (width - 1, 0), (width - 1, height - 1),
+                      (0, height - 1)]
+
+            draw_lines(surface, color, True, points)
+
+            borders = get_border_values(surface, width, height)
+            return [color in border for border in borders]
```
e1000: Here also little question : `color in border` tests that the color appears at least one time on the border ;  if I use *any* (he is a good friend of *all*), it is equivalent to : 

```python
return  [any(c == color for c in border) for border in borders]

# but shouldn't it be ?
return  [all(c == color for c in border) for border in borders]
```

test/draw_test.py
```diff
@@ -1,7 +1,204 @@
...
+            width = surface.get_width()
+            height = surface.get_height()
+            color = (255, 255, 255)
+            points = [(0, 0), (width - 1, 0), (width - 1, height - 1),
+                      (0, height - 1)]
+
+            draw_lines(surface, color, True, points)
+
+            borders = get_border_values(surface, width, height)
+            return [len(border) == border.count(color) for border in borders]
```
e1000: I suggest to replace `len(border) == border.count(color)` with the pattern

```python
all(c == color for c in border)
```

test/draw_test.py
amosbastian: 1. Will update that.
2. Will update that.
3. Since it wasn't clear before I already created #512 showing what I meant.

test/draw_test.py
amosbastian: I see what you mean, I will also update that.

test/draw_test.py
amosbastian: Will also update this, thanks.

illume:
Wonderful! 🎈 🎉

PR #511, opened 2018-08-28
Fix odd-width ellipses not drawing correctly

LambdaGoblin:
The solid and hollow ellipse functions are combined into a single
function that takes a center point, width and height instead of
two radii.  The function is then able to draw to the appropriate
size.

For  #233


illume:
Nice one. We can wait until https://github.com/pygame/pygame/pull/510 is merged in, and then see how this one goes with all the build machines testing it.

Issue #313, opened 2016-09-06
"pygame.draw.polygon" fails to draw polygon on all edges of shape.


illume:
Originally reported by: **User_e (Bitbucket: [User_e](https://bitbucket.org/User_e), GitHub: Unknown)**

----------------------------------------

As shown in the attached image, when drawing a Pygame **polygon** (using the
```
#!python
pygame.draw.polygon
```
function) the polygon is **mostly filled correctly apart from one edge**. Switching the drawing function to `lines` drew all the edges correctly (but had no polygon fill which I want). To see the problem better, I **added code that drew red dots** in the Pygame windows **on the vertices** of the polygon edge (with and without fill) and a **black alignment square which should have 1 white pixel (gap)** between it's top left sides and the shape. In the attached image there is also the relevant code (which I have added, simplified, below). It seems quite clear that this is a Pygame problem (as using the `lines` function worked) so **please take a look** and see if you can fix it.  
Additional case information: Pygame version: `1.9.2a0`, Python version: `3.5.1`.
![Pygame Python module polygon vertices to content drawing issue.png](https://bitbucket.org/repo/aAbk8/images/1255951821-Pygame%20Python%20module%20polygon%20vertices%20to%20content%20drawing%20issue.png)
Simplified example code (full, to run as a file with Python 3.5.1 and Pygame 1.9.2):
```
#!python
import pygame, sys
pygame.init()
surface = pygame.display.set_mode((335, 80))
pygame.display.set_caption("Demonstration of Pygame bug")
border, rect = {"width": 5}, pygame.Rect((0,0,60,60)) # Defined here for debug.
while True:
    surface.fill((200,200,200))
    surface_2 = pygame.Surface((60,60))
    surface_2.fill((255,255,255)) # White is contrasting with black shape.
    for index in range(1):#2):
        path_data = [({0:0,1:rect.width-1}[index],{0:0,1:rect.height
            -1}[index]),(rect.width-1,0), (rect.width-border[
            "width"],border["width"]-1),({0:border["width"]-1,1:
            rect.width-border["width"]}[index],{0:border["width"]-1,
            1:rect.height-border["width"]}[index]),(border["width"]-
            1,rect.height-border["width"]), (0, rect.height-1)]
        #pygame.draw.lines(surface_2,(0,0,0),True,path_data) # Normal function.
        pygame.draw.polygon(surface_2,(0,0,0),path_data) # Function with bug.
        for dot in path_data: # Draw vertices for debugging.
[...]

Issue #234, opened 2015-02-15
pygame.draw.polygon handles x and y direction inconsistently


illume:
Originally reported by: **donLorenzo (Bitbucket: [donLorenzo](https://bitbucket.org/donLorenzo), GitHub: [donLorenzo](https://github.com/donLorenzo))**

----------------------------------------

Reported by Florian Krause on the [pygame mailing list](http://article.gmane.org/gmane.comp.python.pygame/25542):

```
#!plain

Hi there,

I think I found another bug. When drawing a cross, defined as:

pygame.draw.polygon(screen, [255,0,0], [[11, 0], [11, 9], [20, 9], [20, 11], [11, 11], [11, 20], [9, 20], [9, 11], [0, 11], [0, 9], [9, 9], [9, 0]], 0)

the result will show a cross with 2 pixel thickness in the vertical direction, but 3 pixels thickness in the horizontal direction. Which one is "correct" is probably debatable, but Pygame should at least treat both directions equal, shouldn't it?
```

note that the simple example:
```
#!python

import pygame
pygame.init()
screen = pygame.display.set_mode((100,100))
pygame.draw.polygon(screen, [255,0,0], [[0, 0], [3, 0], [3, 3], [0, 3]], 0)
pygame.display.update()
```
yields a 4x4 rect

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/234

e1000:
For me, it should be consistent with `draw.lines`, just `polygon` be the filled version of it. (In fact, when we call `polygon` with `width>0` , we call directly *lines* in the code.

Also it seems difficult to be consistent with `range` when we think of triangle, diamond, ... all shapes where some sides are neither parallel to x-axis nor to y-axis

PR #515, opened 2018-08-30
Draw polygon : tests & code cleanup

e1000:
Progress on #234 and #313 

I did not succeed in solving the bug(s), but I hope I worked out the problem a bit through the tests
which then adds two more failing tests:

* add *draw_test.DrawPolygonTest* class
* this includes tests for the issues #234 & #313 (which both fail as expected)
* some *polygon* C-code simplifications
* remove duplicated equality testing in  *drawhorzlineclip* / *drawhorzline* C-code

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1687,10 +1687,8 @@ draw_fillpoly(SDL_Surface *dst, int *vx, int *vy, int n, Uint32 color)
             else {
                 continue;
             }
-            if ((y >= y1) && (y < y2)) {
-                polyints[ints++] = (y - y1) * (x2 - x1) / (y2 - y1) + x1;
-            }
-            else if ((y == maxy) && (y > y1) && (y <= y2)) {
+            if ( ((y >= y1) && (y < y2)) ||
+                  ((y == maxy) && (y <= y2)) ) {
                 polyints[ints++] = (y - y1) * (x2 - x1) / (y2 - y1) + x1;
             }
         }
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1646,6 +1646,22 @@ draw_fillpoly(SDL_Surface *dst, int *vx, int *vy, int n, Uint32 color)
         maxy = MAX(maxy, vy[i]);
     }
 
+    if (miny == maxy) {
+        /* Special case: polygon only 1 pixel high. */
+        int minx, maxx;
+
+        /* Determine X bounds */
+        minx = vx[0];
+        maxx = vx[0];
+        for (j = 1; (j < n); j++) {
+            minx = MIN(minx, vx[j]);
+            maxx = MAX(maxx, vx[j]);
+        }
+        drawhorzlineclip(dst, color, minx, miny, maxx);
+        PyMem_Free(polyints);
+        return;
+    }
+
     /* Draw, scanning y */
     for (y = miny; (y <= maxy); y++) {
         ints = 0;
@@ -1667,23 +1683,6 @@ draw_fillpoly(SDL_Surface *dst, int *vx, int *vy, int n, Uint32 color)
                 y1 = vy[i];
                 x2 = vx[j];
                 x1 = vx[i];
-            }
-            else if (miny == maxy) {
-                /* Special case: polygon only 1 pixel high. */
-                int minx, maxx;
-
-                /* Determine X bounds */
-                minx = vx[0];
-                maxx = vx[0];
-                for (j = 1; (j < n); j++) {
-                    minx = MIN(minx, vx[j]);
-                    maxx = MAX(maxx, vx[j]);
-                }
-
-                /* Just a line from minimum to maximum X */
-                polyints[ints++] = minx;
-                polyints[ints++] = maxx;
-                break;
             }
             else {
                 continue;
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1312,10 +1312,7 @@ drawhorzlineclip(SDL_Surface *surf, Uint32 color, int x1, int y1, int x2)
     if (x2 < surf->clip_rect.x || x1 >= surf->clip_rect.x + surf->clip_rect.w)
         return;
 
-    if (x1 == x2)
-        set_at(surf, x1, y1, color);
-    else
-        drawhorzline(surf, color, x1, y1, x2);
+    drawhorzline(surf, color, x1, y1, x2);
 }
 
 static void
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1627,12 +1627,10 @@ compare_int(const void *a, const void *b)
 static void
 draw_fillpoly(SDL_Surface *dst, int *vx, int *vy, int n, Uint32 color)
 {
-    int i;
-    int y;
+    int i, j, y;
     int miny, maxy;
     int x1, y1;
     int x2, y2;
-    int ind1, ind2;
     int ints;
     int *polyints = PyMem_New(int, n);
     if (polyints == NULL) {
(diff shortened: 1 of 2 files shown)


test/draw_test.py
```diff
@@ -374,21 +374,144 @@ def todo_test_circle(self):
 
         self.fail()
 
-    def todo_test_polygon(self):
 
-        # __doc__ (as of 2008-08-02) for pygame.draw.polygon:
+RED = (255, 0, 0)
```
illume: `pygame.Color('red')` is a good alternative to these color constants.

test/draw_test.py
```diff
@@ -374,21 +374,144 @@ def todo_test_circle(self):
...
+        pygame.draw.rect(self.surface, RED, (0, 0, 10, 10), 0)
+        # 4. one-pixel-wide, not filled
+        pygame.draw.polygon(self.surface, GREEN, [(4, y) for x, y in CROSS], 1)
+        for y in range(7):
+            self.assertEqual(self.surface.get_at((3, y)), RED)
+            self.assertEqual(self.surface.get_at((4, y)), GREEN)
+            self.assertEqual(self.surface.get_at((5, y)), RED)
+
+    def test_draw_symetric_cross(self):
+        # issue #234 : the result is/was different wether we fill or not
```
illume: This comment could be a docstring.

test/draw_test.py
```diff
@@ -374,21 +374,144 @@ def todo_test_circle(self):
...
+            for y in range(10):
+                if (x in range(2, 5) and y <7) or (y in range(2, 5) and x < 7):
+                    # we are on the border of the cross:
+                    self.assertEqual(self.surface.get_at((x, y)), GREEN, msg=str((x, y)))
+                else:
+                    # we are outside
+                    self.assertEqual(self.surface.get_at((x, y)), RED)
+
+    def test_illumine_shape(self):
+        # Extracted from bug #313
```
illume: A docstring here of what is being tested would be useful.

test/draw_test.py
```diff
@@ -374,21 +374,144 @@ def todo_test_circle(self):
...
+            self.assertEqual(self.surface.get_at((x, y)), GREEN, msg=str((x, y)))
+        for x in range(2, 5):
+            for y in range(2, 5):
+                self.assertEqual(self.surface.get_at((x, y)), GREEN)
+
+    def test_1_pixel_high_or_wide_shapes(self):
+        # 1. one-pixel-high, filled
+        pygame.draw.rect(self.surface, RED, (0, 0, 10, 10), 0)
+        pygame.draw.polygon(self.surface, GREEN, [(x, 2) for x, y in CROSS], 0)
+        for x in range(7):
```
illume: 7 could be put into a variable name.

test/draw_test.py
```diff
@@ -374,21 +374,144 @@ def todo_test_circle(self):
...
+
+
+class DrawPolygonTest(unittest.TestCase):
+
+    def setUp(self):
+        self.surface = pygame.Surface((20, 20))
+
+    def test_draw_square(self):
+        pygame.draw.polygon(self.surface, RED, SQUARE, 0)
+        # note : there is a discussion (#234) if draw.polygon should include or
```
illume: This comment could go in the doc string too.

src_c/draw.c
```diff
@@ -1580,55 +1575,54 @@ draw_fillpoly(SDL_Surface *dst, int *vx, int *vy, int n, Uint32 color)
...
+        for (j = 1; (j < n); j++) {
+            minx = MIN(minx, vx[j]);
+            maxx = MAX(maxx, vx[j]);
+        }
+        drawhorzlineclip(dst, color, minx, miny, maxx);
+        PyMem_Free(polyints);
+        return;
+    }
+
+    /* special case : horizontal border lines with miny < y < maxy */
```
illume: I don't understand this part. What is this special case about exactly?

illume:
Very nice set of tests and bug fixes :)

One possible improvement to this function, would be improving the naming of the variables.
```
int *vx, int *vy
```

src_c/draw.c
```diff
@@ -1580,55 +1575,54 @@ draw_fillpoly(SDL_Surface *dst, int *vx, int *vy, int n, Uint32 color)
...
-            }
-            else {
+            else { // y1 == y2 : handled above
                 continue;
             }
-            if ((y >= y1) && (y < y2)) {
-                polyints[ints++] = (y - y1) * (x2 - x1) / (y2 - y1) + x1;
-            }
-            else if ((y == maxy) && (y > y1) && (y <= y2)) {
+            if ( ((y >= y1) && (y < y2)) ||
```
illume: What is being checked here?

illume:
It looks like the C code needs to be run through clang-format.

e1000:
better now or too much comments ? I guess it's a standard algorithm, buy anyway

src_c/draw.c
```diff
@@ -1557,88 +1554,111 @@ compare_int(const void *a, const void *b)
...
-            else {
+            else { // y1 == y2 : has to be handled as special case (below)
                 continue;
             }
-            if ((y >= y1) && (y < y2)) {
-                polyints[ints++] = (y - y1) * (x2 - x1) / (y2 - y1) + x1;
-            }
-            else if ((y == maxy) && (y > y1) && (y <= y2)) {
-                polyints[ints++] = (y - y1) * (x2 - x1) / (y2 - y1) + x1;
+            if ( ((y >= y1) && (y < y2)) ||
```
illume: One way to make expressions a bit smaller is to give names to their components.
```c
bool y_crosses_edge = ((y >= y1) && (y < y2))
bool on_lowest_line = ((y == maxy) && (y2 == maxy))
if (y_crosses_edge || on_lowest_line) {
```
However, I wonder if C compilers are smart enough to opitimze this well... ?

illume:
Thanks for that. It's waaaaaay more readable now.

Yeah definitely, the whole draw.c could use some improvements. Since you had that function in your head, I thought you would be able to make it easier to understand for the next person.

illume:
I asked in the Discord chat room if someone is available to read the cleaned up function.

src_c/draw.c
```diff
@@ -1557,88 +1554,114 @@ compare_int(const void *a, const void *b)
...
+        maxy = MAX(maxy, point_y[i]);
     }
 
-    /* Draw, scanning y */
+    if (miny == maxy) {
+        /* Special case: polygon only 1 pixel high. */
+        int minx, maxx;
+
+        /* Determine X bounds */
+        minx = point_x[0];
```
illume: I think these types can be declared on the same line as they are assigned.
```c
int minx = point_x[0];
int maxx = point_x[0];
```

illume:
Looks like it broke in appyveyor Visual C at some point a while ago.

e1000:
Sorry, I do not understand the broken `appyveyor Visual C` ; what is "a while ago", before this branch?

test/draw_test.py
e1000: done

test/draw_test.py
e1000: done

e1000:
python-code : Not sure what a "gist" is. 
We have `./test/test_utils/png.py` , (python implementation of *png* , and it's only used in tests...) maybe that's the right place for that kind of stuff?

illume:
A gist is a paste bin type place on github. You can put code there that is also versioned.
https://gist.github.com/

illume:
If you look at the commits tab of this PR, you can see each time appveyor compiled and ran the tests. So, somewhere between the last time it passed, and the first time it failed is where the error was introduced.

illume:
It looks like the old visual c is having problems with the type declaration here:
```
src_c/draw.c(1610) : error C2143: syntax error : missing ';' before 'type'
src_c/draw.c(1612) : error C2065: 'n_intersections' : undeclared identifier
src_c/draw.c(1634) : error C2065: 'n_intersections' : undeclared identifier
src_c/draw.c(1638) : error C2065: 'n_intersections' : undeclared identifier
src_c/draw.c(1640) : error C2065: 'n_intersections' : undeclared identifier
```
https://github.com/pygame/pygame/blob/80911a3b521164fdd34f2ef1a528fcd4a0edf835/src_c/draw.c#L1610

https://stackoverflow.com/questions/18912912/syntax-error-missing-before-type-in-c

illume:
CPython has a policy now of having python versions of all new C modules. It's a very useful thing to have for testing, and debugging... as you've shown. So perhaps it should go in src_py/draw.py

e1000:
alright, I like also the idea to have it in `src_py` ; no risk of confusion with other python modules? maybe a subfolder?

I've put `int n_intersections = 0;` inside the loop; I hope it makes **appveyor** happy...

illume:
If it goes in src_py/ then it can't be called draw.py. Because then there could be a confusion on which module to load... the draw.so/dll or the draw.py. There is a convention that the 'c' module is called something like _draw.so, with the leading underscore. But in this case we can not have the draw.py be there because it does not do imports from the _draw.so. So calling it something like draw_py.py could be a good temporary name. In the future perhaps the draw.py would import from _draw.c the functions it needs.

PR #534, opened 2018-09-06
More Test cleanup

### LINKED ISSUE #496: Test suite : documentation and cleanup ?

e1000:
Some more test cleanups; it's an infinite job, but I want to believe ( that we might close #496 this year :) )

But this time, when I run all tests, I get  : 
```
  File "/usr/lib/python3.5/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "test/__main__.py", line 131, in <module>
    run_and_exit(*args, **kwds)
  File "/home/eea/src/pygame/test/test_utils/run_tests.py", line 339, in run_and_exit
    total, fails = run(*args, **kwargs)
  File "/home/eea/src/pygame/test/test_utils/run_tests.py", line 299, in run
    assert total == untrusty_total
AssertionError
```
which I don't understand. I checked the commit where I introduced it (917f3645), but I still do not understand ...

Other thing, many tests depend on `pygame.HAVE_NEWBUF`, but I could find no documentation on this variable, where and when it's introduced, and if it can be *false* ...

### DIFF
--- a/test/base_test.py
+++ b/test/base_test.py
@@ -207,32 +207,22 @@ class BaseModuleTest(unittest.TestCase):
 
     if pygame.HAVE_NEWBUF:
         from pygame.tests.test_utils import buftools
-        def test_newbuf(self):
-            self.NEWBUF_test_newbuf()
-        def test_PgDict_AsBuffer_PyBUF_flags(self):
-            self.NEWBUF_test_PgDict_AsBuffer_PyBUF_flags()
-        def test_PgObject_AsBuffer_PyBUF_flags(self):
-            self.NEWBUF_test_PgObject_AsBuffer_PyBUF_flags()
-        def test_bad_format(self):
-            self.NEWBUF_test_bad_format()
 
     def NEWBUF_assertSame(self, proxy, exp):
         buftools = self.buftools
         Importer = buftools.Importer
         self.assertEqual(proxy.length, exp.len)
         imp = Importer(proxy, buftools.PyBUF_RECORDS_RO)
-        try:
-            self.assertEqual(imp.readonly, exp.readonly)
-            self.assertEqual(imp.format, exp.format)
-            self.assertEqual(imp.itemsize, exp.itemsize)
-            self.assertEqual(imp.ndim, exp.ndim)
-            self.assertEqual(imp.shape, exp.shape)
-            self.assertEqual(imp.strides, exp.strides)
-            self.assertTrue(imp.suboffsets is None)
-        finally:
-            imp = None
+        self.assertEqual(imp.readonly, exp.readonly)
+        self.assertEqual(imp.format, exp.format)
+        self.assertEqual(imp.itemsize, exp.itemsize)
+        self.assertEqual(imp.ndim, exp.ndim)
+        self.assertEqual(imp.shape, exp.shape)
+        self.assertEqual(imp.strides, exp.strides)
+        self.assertTrue(imp.suboffsets is None)
 
-    def NEWBUF_test_newbuf(self):
+    @unittest.skipIf(not pygame.HAVE_NEWBUF, 'newbuf not implemented')
+    def test_newbuf(self):
         from pygame.bufferproxy import BufferProxy
 
         Exporter = self.buftools.Exporter
@@ -250,7 +240,8 @@ class BaseModuleTest(unittest.TestCase):
             v = BufferProxy(o)
             self.NEWBUF_assertSame(v, o)
 
-    def NEWBUF_test_bad_format(self):
+    @unittest.skipIf(not pygame.HAVE_NEWBUF, 'newbuf not implemented')
+    def test_bad_format(self):
         from pygame.bufferproxy import BufferProxy
         from pygame.newbuffer import BufferMixin
         from ctypes import create_string_buffer, addressof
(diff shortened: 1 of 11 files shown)


test/base_test.py
```diff
@@ -81,14 +81,11 @@ def __init__(self, shape, typechar, itemsize):
...
         self.assertEqual(proxy.length, obj.size)
-        d = proxy.__array_interface__
-        try:
-            self.assertEqual(d['typestr'], obj.typestr)
-            self.assertEqual(d['shape'], obj.shape)
-            self.assertEqual(d['strides'], obj.strides)
-            self.assertEqual(d['data'], obj.data)
-        finally:
-            d = None
+        iface = proxy.__array_interface__
```
illume: I wonder why you moved this out of a try finally block?

Probably the reason it assigns that to None is so that the destructor is called immediately.

test/base_test.py
```diff
@@ -207,32 +204,22 @@ def get__array_struct__(self):
...
-        try:
-            self.assertEqual(imp.readonly, exp.readonly)
-            self.assertEqual(imp.format, exp.format)
-            self.assertEqual(imp.itemsize, exp.itemsize)
-            self.assertEqual(imp.ndim, exp.ndim)
-            self.assertEqual(imp.shape, exp.shape)
-            self.assertEqual(imp.strides, exp.strides)
-            self.assertTrue(imp.suboffsets is None)
-        finally:
-            imp = None
```
illume: Similar to previous comment, this `imp = None` part is probably needed (I'm guessing).

test/base_test.py
e1000: ok, I just wondered about all these `del sth`, I might have been going too far... 
But why should the destructor be called immediately? After all, the tests pass without...

test/base_test.py
e1000: Same question as above: why would it be needed? Do you think, it could be related to the `AssertionError` crash above ? I'll try it out..

e1000:
I don't know about `PG_ENABLE_NEWBUF` and `PG_ENABLE_NEWBUF` , but I can do the cleanup if you are sure !

As about the above *crash*, do you have any hint or documentation ? I came across before, but I don't remember what solved the problem. A more explaining Error message would be nice, I think...

test/base_test.py
e1000: no, my remark was non sens, it can not be related since the crash happens as of commit (917f364) where as the `del` are removed on my last commit :(

e1000:
got it for the crash ! it was the test machinery not accepting letters 's' (skipping) and 'x' (expected failure);

(at least I hope so, because it seemed to behave randomly, but the mistake was problably in front of the computer ...)

e1000:
should I reintroduce the try/finally for immediate destructor calling ?

test/base_test.py
illume: I guess this will cause issues on pypy, which doesn't call destructors when something leaves scope.

illume:
Thanks 🎉 

ps. Again... sorry for the delay.

Issue #522, opened 2018-09-04
What is draw.aaline supposed to draw with float point coordinates?


e1000:
Hello,

I'm trying to understand the code around *draw.aaline* (and hence *draw.aalines*). There are ~~many~~ several bugs, it seems:

* `blend` seems completely ignored
* black artefacts here and there at the end or the border of some lines
* float point coordinates are handled inconsistently 

Here is a code snippet that generates two bitmaps with short aa-lines in different directions : 

```python
import pygame
pygame.init()

SIDE = 50

BG_COLOR = (255, 0, 0)
BG_COLOR2 = (0, 0, 255)
DRAW_COLOR = (255, 255, 0)

def draw_short_lines(half_int):
    surface = pygame.display.set_mode((SIDE, SIDE))
    # add two background colors
    pygame.draw.rect(surface, BG_COLOR, (0, 0, 10, SIDE), 0)
    pygame.draw.rect(surface, BG_COLOR2, (10, 0, 40, SIDE), 0)

    for x in range(-1, 3):
        for y in range(-1, 3):

            x_start = 10 * (x + 2)
            y_start = 10 * (y + 2)
            if half_int:
                x_start += 0.5
                y_start += 0.5
            pygame.draw.aaline(surface, DRAW_COLOR, (x_start, y_start),
                               (x_start + x * 2, y_start + y * 2))
    pygame.display.update()
    return surface


pygame.image.save(draw_short_lines(False), 'aaline-integer.bmp')
pygame.image.save(draw_short_lines(True), 'aaline-int-n-half.bmp')
```

As for me, I don't understand the meaning of *floating point coordinates* for this algorithm? I don't even know what I should put in unit tests...

So I watched out the web: 

- The reference seems to be [Xiaolin_Wu's_line_algorithm](https://en.wikipedia.org/wiki/Xiaolin_Wu's_line_algorithm), this wikipedia page doesn't mention floating points, but the *gif* uses the center of the pixel. 
- This (implementation)[https://www.geeksforgeeks.org/anti-aliased-line-xiaolin-wus-algorithm/] considers only integer startpoints
- Same for [codeproject](https://www.codeproject.com/articles/13360/antialiasing-wu-algorithm) which adds also the point of view that when we are on horizontal or vertical lines, or on exact diagonal, there should be no antialiasing (different to our current implementation).

My suggestion would be : 

* We should only consider integer points for start and end point.
[...]

PR #535, opened 2018-09-09
draw_py for Python algos & aaline testing

e1000:
* introduce *src_py/draw_py.py* for Python implementation of drawing algorithms
* implement and test Python version for `draw.polygon`
* tests for `aaline` conforming to the suggestions made in #522
* [edit] some *draw.c* cleanup

Remark: it seems that, within the *src_py/draw_py.py* module,
the following code does not work in *Python2*  : 
```python
from __future__ import relative_imports
from math import floor, ceil, trunc
```

Instead, **math** is interpreted as *pygame.math* and Python does not find the
functions ... But I would like to use those functions for an `aaline` algorithm.
(I ll try to investigate this).


illume:
I guess `pygame.math` was a bad naming choice considering there is a `math` builtin module.

src_py/draw_py.py
```diff
@@ -0,0 +1,337 @@
...
+
+    drawvertline(surf, color, x, y_from, y_to)
+
+
+LEFT_EDGE = 0x1
+RIGHT_EDGE = 0x2
+BOTTOM_EDGE = 0x4
+TOP_EDGE = 0x8
+
+def encode(x, y, left, top, right, bottom):
```
illume: encode could use documentation

src_py/draw_py.py
```diff
@@ -0,0 +1,337 @@
...
+            (y < top) * TOP_EDGE +
+            (y > bottom) * BOTTOM_EDGE)
+
+
+INSIDE = lambda a: not a
+ACCEPT = lambda a, b: not (a or b)
+REJECT = lambda a, b: a and b
+
+
+def clip_line(line, left, top, right, bottom):
```
illume: needs docs

src_py/draw_py.py
```diff
@@ -0,0 +1,337 @@
...
+            if x2 != x1:
+                x1 += int((bottom - y1) / m)
+            y1 = bottom
+        elif code1 & TOP_EDGE:
+            if x2 != x1:
+                x1 += int((top - y1) / m)
+            y1 = top
+
+
+def clip_and_draw_line(surf, rect, color, pts):
```
illume: needs docs

src_py/draw_py.py
```diff
@@ -0,0 +1,337 @@
...
+    if pts[1] == pts[3]:
+        drawhorzline(surf, color, pts[0], pts[1], pts[2])
+    elif pts[0] == pts[2]:
+        drawvertline(surf, color, pts[0], pts[1], pts[3])
+    else:
+        draw_line(surf, color, pts[0], pts[1], pts[2], pts[3])
+    return 1
+
+
+# Variant of https://en.wikipedia.org/wiki/Bresenham's_line_algorithm
```
illume: Needs docs.

illume:
Documentation improvements could go into a future PR. Going to merge this now.

Thanks :)

Issue #494, opened 2018-08-20
Cleanup / wiki / dev-documentation ?


e1000:
Hello,

I wanted to get involved in Pygame development process, but found it not so easy : 
 
* I could find nowhere which Python and SDL versions are supported
* A lot of code (eg in `src/*c` and `src/_pygame.h`) refers to Python < 2.7 or SDL < 1.2.4
*  It is refered to  unittest of Python3 in [lets-write-unit-test](http://renesd.blogspot.com/2018/07/draft-of-lets-write-unit-test.html) , which could be misleading if we support *Python2.7* (a lot of new stuff has been introduced since then)
* It seems to me, it would be nice to note somewhere why we don't use *ctypes*, *Cython* or *CFFI* ? 
* Isn't there somewhere a numpy dependency ? which version ?
* Maybe we could move some old stuff in the https://www.pygame.org/wiki/
  (Py3 migration, 2010 ..., etc) in a subfolder *Archive* ? 

* more code cleanup ? (specially the `test/` folder and files need some attention it seems to me ; eg  `test/util/svn_log_to_whatsnew.py` , also what is the meaning of `test/util/build_page/index.php`) 
* SDL2 is in a single ticket #174, can it really be done in  one step, or should we split it in several tickets ?
* Maybe be can drop Python2 support for Pygame2? 
* I tried to run the tests, but get two crashes (ok I should run `setup.py` I guess, but then I'm confused, what's the meaning of `installed pygame.tests` referred in the [wiki/Hacking](https://www.pygame.org/wiki/Hacking))
```bash
$ python3 -m pygame.tests
/usr/bin/python3: Error while finding module specification for 'pygame.tests' (ImportError: No module named 'pygame')

# and the other way : 

$ python3 test/__main__.py 
Traceback (most recent call last):
  File "test/__main__.py", line 32, in <module>
    from test.test_utils.run_tests import run_and_exit
  File "/home/eea/src/pygame/test/__init__.py", line 40, in <module>
    from test.test_utils.run_tests import run
  File "/home/eea/src/pygame/test/test_utils/__init__.py", line 5, in <module>
    import tempfile, sys, pygame, time, os
ImportError: No module named 'pygame'
```
* Oh yes, and I found many very old open branches in github, maybe we can close them ?
 
Feel free to close this ticket though, if you think it's inappropriate or obsolete...

illume:
Thanks for this list. It's very valuable to address each of these things in a way that will also be helpful for others later.

[...]

PR #536, opened 2018-09-11
dependency version into README & Cleanup SDL1.2

e1000:
Here are my best guesses for the dependency versions that I wrote into REAMDE.rst :  
* Python >= 2.7
* SDL >= 1.2.15 (last version; released 2012, https://en.wikipedia.org/wiki/Simple_DirectMedia_Layer)
* SDL_mixer >= 1.2.13 (last version; released 2012, see source code)
* NumPy >= 1.6.2 (just as example : of Debian Wheezy (old-old-stable), released 2013: 
   https://www.debian.org/releases/wheezy/, https://packages.debian.org/wheezy/python-numpy)
* SDL_image (version?)
* SDL_tff (version ?)
* SDL_gfx (optional, version?)

Then I cleaned up "mixer.c" and "music.c" ...

At merge, we can check one more point in check list #494 !

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -126,8 +126,6 @@ Credits
      Also a big thanks to Roger Dingledine and the crew at SEUL.ORG for our
      excellent hosting.
 
-
-
 Dependencies
 ============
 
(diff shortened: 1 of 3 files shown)


illume:
Sorry for the long delay...

Looks good to me :) The other versions could be filled in another time.

Issue #179, opened 2013-10-04
Need to update system font lists in sysfont.py (2013)


illume:
The system font lists for Windows are out-of-date.

Docs: https://www.pygame.org/docs/ref/font.html#pygame.font.SysFont
Code: https://github.com/pygame/pygame/blob/master/lib/sysfont.py
Test: https://github.com/pygame/pygame/blob/master/test/sysfont_test.py

- [x] update windows support using some more modern font lists, or investigate a way to programmatically list fonts on various windows systems.

Perhaps see the matplotlib implementation for windows font finder: https://matplotlib.org/_modules/matplotlib/font_manager.html#win32InstalledFonts

- Windows 10 fonts: https://docs.microsoft.com/en-us/typography/fonts/windows_10_font_list
- Windows 7 fonts: https://docs.microsoft.com/en-us/typography/fonts/windows_7_font_list
- https://en.wikipedia.org/wiki/List_of_typefaces_included_with_Microsoft_Windows
- http://www.fixarna.com/frontpage/windowsfonts.htm

---

- [x] update mac support using `system_profiler` instead of fc-list.

`fc-list` on Mac is not installed by default any more(High Sierra), and relies on X. Also, it can run for 20 seconds or more if the user has never used it before.

On mac, `system_profiler SPFontsDataType -xml` is available, which dumps lots of font info in xml format to stdout. It takes 2.1 seconds on a 2013 SSD macbook pro. It works at least from Mac OSX 10.10.

OS X font locations: http://support.apple.com/kb/HT2435
Mac OS X 10.3 list: http://support.apple.com/kb/HT2444
Mac OS X 10.4 list: http://support.apple.com/kb/HT1538
Mac OS X 10.5 list: http://support.apple.com/kb/HT1642
Mac OS X 10.6 list: http://support.apple.com/kb/ht5154
Mac OS X 10.7 list: http://support.apple.com/kb/HT5098
OS X 10.8 list: http://support.apple.com/kb/ht5379
OS X 10.9 list: http://support.apple.com/kb/HT5944
OS X 10.10 list: ??
OS X 10.11 list: ??
macOS Sierra 10.12 list: https://support.apple.com/en-us/HT206872
macOS High Sierra 10.13 https://support.apple.com/en-us/HT207962

----------------------------------------
Originally reported by: **Jason Marshall (Bitbucket: [jmm0](https://bitbucket.org/jmm0), GitHub: [jmm0](https://github.com/jmm0))**
[...]

PR #538, opened 2018-09-16
Mac fonts support using system_profiler

Cerealdragon:
Replaced fc-list with system_profiler as requested in this issue: https://github.com/pygame/pygame/issues/179

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -24,6 +24,7 @@ import os
 import sys
 from pygame.compat import xrange_, PY_MAJOR_VERSION
 from os.path import basename, dirname, exists, join, splitext
+import xml.etree.ElementTree as ET
 
 
 OpenType_extensions = frozenset(('.ttf', '.ttc', '.otf'))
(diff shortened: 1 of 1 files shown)


src_py/sysfont.py
```diff
@@ -138,19 +139,80 @@ def initsysfonts_win32():
...
-    # disc
-    elif exists("/usr/X11R6/bin/fc-list"):
-        fonts = initsysfonts_unix("/usr/X11R6/bin/fc-list")
-    else:
-        fonts = {}
+def fetch_elements_from_xmlString(xml_string, Xpath):
+    """Gets the specified element iterable"""
+
+    #Get the ElementTree object from the xml string
+    mac_fonts_xml_tree = ET.fromstring(xml_string)
```
e1000: why not just one line ? like : 
```python
def fetch_elements_from_xmlString(xml_string, Xpath):
    return ET.fromstring(xml_string).iterfind(Xpath)
```

src_py/sysfont.py
```diff
@@ -138,19 +139,80 @@ def initsysfonts_win32():
...
+def textContentList_from_xmlElement(element):
+    """Gets the content of tags (returned as a list) from the specified element and its subelements"""
+    #Gets all the text within each tag in the node
+    tag_text = element.itertext()
+
+    #Cleans the list of any whitespace items
+    tag_content_list = [content for content in tag_text if content.strip() != ""]
+
+    return tag_content_list
+
```
e1000: maybe we can also do this in one line ?
```python
def textContentList_from_xmlElement(element)
     return [c for c in element.itertext() if c.strip()]
```

Cerealdragon: @e1000  
Thank you for the suggestions, I made the adjustments :D

Issue #364, opened 2017-11-18
DirtySprite doesn't allow you to set _layer attribute before adding to groups


pathunstrom:
In `DirtySprite.__init__` at line [259 of sprite.py](https://github.com/pygame/pygame/blob/8aeb2f72047218043b2353dae4ccb288eaa5d814/lib/sprite.py#L253) `DirtySprite` sets `self._layer` to 0, preventing the ability to subclass cleanly.

Example:

    class MySprite(pygame.sprite.DirtySprite):
        def __init__(*groups):
            self._layer = 1
            super().__init__(*groups)

If you instantiate this class and pass it a few `Layered` groups `MySprite(group1, group2)` instances are set to Layer 0 instead of layer 1.

Moving the `self._layer` after the super makes `MySprite._layer` become 1, but it's 0 when it's added to the groups.

illume:
Good point.

I guess adding something like this to the DirtySprite initialiser would work:
`self._layer = getattr(self, '_layer', 0)`

pathunstrom:
At Pycon, can get likely get a PR in today.

amosbastian:
Did this ever get fixed? @pathunstrom

pathunstrom:
@amosbastian I wasn't able to get Pygame to build from source locally so I couldn't properly test any fix I made. So, no, not fixed.

illume:
@pathunstrom sorry for the poor state of the developer documentation. There's been some improvement recently, but it still lacks in various areas that I'll try and address shortly.

@dtschirmer 's fix was merged in, which should make it into the next release.

PR #540, opened 2018-09-18
DirtySprite subclass allow other _layer values

dtschirmer:
See: https://github.com/pygame/pygame/issues/364
When subclassing DirtySprite _layer can be set to different values.

### DIFF
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -245,8 +245,8 @@ class DirtySprite(Sprite):
         the screen.)
 
     _layer = 0
-        A READ ONLY value, it is read when adding it to the LayeredUpdates
-        group. For details see documentation of sprite.LayeredUpdates.
+        0 is the default value but this is able to be set differently 
+        when subclassing.
 
     """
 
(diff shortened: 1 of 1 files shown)


src_py/sprite.py
```diff
@@ -256,7 +256,8 @@ def __init__(self, *groups):
         self.blendmode = 0  # pygame 1.8, referred to as special_flags in
                             # the documentation of Surface.blit
         self._visible = 1
-        self._layer = 0     # READ ONLY by LayeredUpdates or LayeredDirty
+        self._layer = getattr(self, '_layer', 0)    # Default 0 unless 
+                                                    # initialized differently.
```
e1000: If it should be possible to define `_layer` by the user at instantiation of the **DirtySprit** (or any subclass), then imho, we should do : 

```python

class DirtySprite(Sprite):

    def __init__(self, *groups, layer=0):
        ...
        self._layer = layer
```
And update the documentation of the constructor.
So it clearer, saves one line, and some nice cases, you won't even need an extra subclass...

illume:
Thanks for the fix :)

PR #545, opened 2018-10-02
update system font lists in sysfont.py

andreyx86:
mac os fix for issue: #179

i've done this today in a couple of hours and didn't notice that there is another pull request for this issue.

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -173,7 +173,7 @@ def _add_sys_font(current_font, multiple_fonts, fonts):
         for font_item in multiple_fonts:
             _add_sys_font_inner(font_item, fonts)
 
-    if current_font is not None:
+    if len(current_font) > 0:
         _add_sys_font_inner(current_font, fonts)
 
 
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -148,8 +148,76 @@ def initsysfonts_darwin():
     # disc
     elif exists("/usr/X11R6/bin/fc-list"):
         fonts = initsysfonts_unix("/usr/X11R6/bin/fc-list")
+    elif exists("/usr/sbin/system_profiler":
+        fonts = initsysfonts_macos("/usr/sbin/system_profiler")
     else:
         fonts = {}
+
+    return fonts
+
+
+# read the fonts using system_profiler on macOS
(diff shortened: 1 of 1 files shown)


andreydev:
fixed some problems in code, seems to be working good now.

src_py/sysfont.py
```diff
@@ -148,12 +148,112 @@ def initsysfonts_darwin():
...
+    _addfont(
+        _simplename(current_font['full name']), bold, italic or oblique, current_font['path'],
+        fonts)
+
+
+def _add_sys_font(current_font, multiple_fonts, fonts):
+
+    if len(multiple_fonts) > 0:
+        for font_item in multiple_fonts:
+            _add_sys_font_inner(font_item, fonts)
```
e1000: no need of the *if*-test, those three lines can just be written like this : 
```python
        for font_item in multiple_fonts:
            _add_sys_font_inner(font_item, fonts)
```

src_py/sysfont.py
```diff
@@ -148,12 +148,112 @@ def initsysfonts_darwin():
...
+
+    if len(current_font) > 0:
+        _add_sys_font_inner(current_font, fonts)
```
e1000: it's more "pythonic" to just write: 
```python
if current_font:
    ...
```

src_py/sysfont.py
```diff
@@ -148,12 +148,112 @@ def initsysfonts_darwin():
...
+    parsing_font = False
+    current_font = {}
+
+    multiple_fonts = []
+
+    lines = entries.split('\n')
+    lines_len = len(lines);
+                
+    try:
+        for idx in range(0, lines_len):
```
e1000: no need of variable `lines_len` and `idx` , just loop over `lines` : 

```python
for line in lines:
  key, value = line.strip().split(':', 1)
  ...
```
also , if you do the `lower()` here : 
```python
for line in lines:
  key, value = line.strip().lower().split(':', 1)
  ...
```
then, you can remove all the other calls to `lower()` which would more clean imho ...

andreydev: can't use ```lower()``` on the whole line because mac's  file system is case sensitive and this line can contain a path to the font file, i fixed it for ```key```

PR #546, opened 2018-10-02
Improve message on failed imports & cleanup (fix #541)

### LINKED ISSUE #541: No proper error message showing up when there's an import error
I was trying to use pygame.surfarray module without knowing that it requires numpy, and got this error:
```
  File "/Users/jeong/Dev/python/miniconda2/envs/temp/lib/python3.6/site-packages/pygame/__init__.py", line 108, in __getattr__
    raise NotImplementedError(MissingPygameModule)
NotImplementedError: surfarray module not available
```

So I looked up which part is causing this problem, and this seems to be a cause.

https://github.com/pygame/pygame/blob/d5cb29b1900cab646bc8891b78aa50d12911c104/src_py/__init__.py#L419-L424

The get_error() function produces a proper error message inside the except block like:
```
No module named 'numpy'
```

but returns None after the exception block ends, making MissingModule.info None.

Am I correct, or is there anything I'm missing?

e1000:
Fix and cleanup failing pygame modules (edit : fixes #541) : 
* Capture the import exception at `MissingModule.__init__`
* `_failed_import` variable usage
* cleanup `MissingModule` class

### DIFF
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -121,7 +121,7 @@ from pygame.base import *
 from pygame.constants import *
 from pygame.version import *
 from pygame.rect import Rect
-from pygame.compat import geterror, PY_MAJOR_VERSION
+from pygame.compat import PY_MAJOR_VERSION
 from pygame.rwobject import encode_string, encode_file_path
 import pygame.surflock
 import pygame.color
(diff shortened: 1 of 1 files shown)


illume:
Very nice :)

bendavis78:
This PR breaks https://github.com/pygame/pygame/pull/413.

PR #547, opened 2018-10-03
Set resolution

andreyx86:
couldn't find an option to change resolution, with this code it's possible to change the resolution by calling ```pygame.display.set_resolution((x, y))```, it preserves the old depth and flags.

tested with:
```
import pygame
import random

pygame.init()
screen = pygame.display.set_mode((640, 480))
clock = pygame.time.Clock()
done = False

screen.fill((255, 255, 255))
pygame.display.flip()

green = (0,255,0)

print(pygame.get_sdl_version())

while not done:
        if event.type == pygame.QUIT:
            done = True
            done = True
        if event.type == pygame.KEYDOWN and event.key == pygame.K_SPACE:
            pygame.display.set_resolution((random.randint(32, 512), random.randint(32, 512)))

    screen.fill((255, 255, 255))

    pygame.draw.circle(screen, green, (screen.get_width() / 2, screen.get_height() / 2), min(screen.get_width() , screen.get_height()) / 2, 5)

    pygame.display.flip()

    clock.tick(60)
```

works fine on ```macOS Mojave``` and ```Ubuntu 18.04``` with SDL

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -682,7 +682,7 @@ pg_set_resolution(PyObject *self, PyObject *arg)
     int w = 0;
     int h = 0;
 
-    if (!PyArg_ParseTuple(arg, "|(ii)ii", &w, &h))
+    if (!PyArg_ParseTuple(arg, "|(ii)", &w, &h))
         return NULL;
 
     if (w < 0 || h < 0)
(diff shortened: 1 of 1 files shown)


illume:
Did you set display.set_mode ?

andreyx86:
yea, i figured out what set_mode is doing today (still learning the API and the code)

PR #550, opened 2018-10-05
updated for python 3

DavidCaughell:
https://stackoverflow.com/questions/436198/what-is-an-alternative-to-execfile-in-python-3


illume:
Oh! That's why it wasn't working on python3.

Thanks :)

Issue #548, opened 2018-10-03
Freezes when playing music from file objects


dlon:
It usually occurs when I play sound effects at the same time.
```python
import pygame


pygame.mixer.pre_init(buffer=512)
pygame.init()

displaySurface = pygame.display.set_mode((320, 240))

h = open('music.ogg', 'rb')
pygame.mixer.music.load(h)
pygame.mixer.music.play()
s = pygame.mixer.Sound('soundeffect.wav')

clock = pygame.time.Clock()
running = True

while running:
	for event in pygame.event.get():
		if event.type == pygame.QUIT or\
			(event.type == pygame.KEYDOWN and\
			event.key == pygame.K_ESCAPE):
			running = False
			break
		elif event.type == pygame.KEYDOWN:
			s.play()
	pygame.display.flip()
	clock.tick(60)
```

When I spam any key while running the above example, it eventually locks. However, if I use a string to load the music, it doesn't occur.

I'm using Windows 10, pygame 1.9.5-dev0, SDL 1.2.15, SDL-mixer 1.2.12, and Python 3.7. It also happens using pygame 1.9.4 (PyPI package).

dlon:
This is caused by SDL_LockAudio() somehow. Editing SDL_sysmutex.c, line 76, from
`if ( WaitForSingleObject(mutex->id, INFINITE) == WAIT_FAILED ) {`
to 
`if ( WaitForSingleObject(mutex->id, 5000) == WAIT_FAILED ) {`
causes it to freeze for only 5 seconds before continuing.

PR #551, opened 2018-10-07
Fixed freezing while playing music from file objects

dlon:
#548

The problem is primarily `_pg_rw_read_th`. Calling fileobj.read() (and possibly Py_DECREF) makes Python switch to the other thread and often causes the other thread to wait forever to acquire the mixer mutex.

My solution was to store the file descriptor and use this instead of the Python method, if it's available. Are there any drawbacks to this?

### DIFF
--- a/src_c/rwobject.c
+++ b/src_c/rwobject.c
@@ -37,6 +37,7 @@ typedef struct {
     PyObject *seek;
     PyObject *tell;
     PyObject *close;
+    int fileno;
 } pgRWHelper;
 
 /*static const char pg_default_encoding[] = "unicode_escape";*/
@@ -275,6 +276,7 @@ pgRWopsFromFileObject(PyObject *obj)
         PyMem_Del(helper);
         return (SDL_RWops *)PyErr_NoMemory();
     }
+    helper->fileno = PyObject_AsFileDescriptor(obj);
     fetch_object_methods(helper, obj);
     rw->hidden.unknown.data1 = (void *)helper;
 #if IS_SDLv2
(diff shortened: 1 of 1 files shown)


src_c/rwobject.c
```diff
@@ -398,6 +402,17 @@ _pg_rw_seek(SDL_RWops *context, Sint64 offset, int whence)
     Sint64 retval;
 #endif
 
+    if (helper->fileno != -1) {
+        if (!(offset == 0 &&
```
illume: Is the logic here correct?

It seems lseek could be called twice here in the case where offset != 0 && whence != SEEK_CUR, and lseek doesn't return -1?

illume:
Thanks for this.

- I wonder if we have a unit test for this already? 
- I guess this leave the bug there for file objects that don't have file descriptors.

src_c/rwobject.c
```diff
@@ -435,6 +450,15 @@ _pg_rw_read(SDL_RWops *context, void *ptr, size_t size, size_t maxnum)
     size_t retval;
 #endif /* IS_SDLv2 */
 
+    if (helper->fileno != -1) {
+        retval = read(helper->fileno, ptr, size * maxnum);
+        if (retval == -1) {
+            return -1;
+        }
+        retval /= size;
```
illume: I guess it's possible that read could return less than a multiple of size. It's possible read could return 1 byte for example 4 times in a row. I'm not sure how to handle this case really, and it's also quite unlikely.

src_c/rwobject.c
dlon: True. It did do the right thing but that extra lseek was unnecessary.

src_c/rwobject.c
dlon: Seems to be true of the other call as well, so it shouldn't be any worse at least.

PR #557, opened 2018-10-14
Update typecheck to allow subtypes

### LINKED ISSUE #552: subclassing the math.Vector2 class results in Fatal Error
When I make a subclass of pygame.math.Vector2 and do a scalar multiplication, it results in a crash and a Fatal Error
```python
import pygame

class Vector(pygame.math.Vector2):
    pass

# this stuff works fine
vec_a = Vector(2, 0)
vec_b = Vector(0, 1)

print(vec_a + vec_b)
print(vec_a.length())
print(vec_a.normalize())

vec_c = vec_a.rotate(90)
print(vec_c)

# but this raises the Error
vec_a *= 2
```

The Error message I get is 
> Fatal Python error: (pygame parachute) Segmentation Fault

Reproducability: I only tried this on two Windows 7 PCs with python 3.6 and in Spyder. Pygame Version is 1.9.3

Also someone on reddit said they got the same error 
https://www.reddit.com/r/learnpython/comments/9mnu4b/making_an_existing_class_hashable_pygame_vector/

sw00:
Fixes #552 .

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -592,10 +592,14 @@ vector_generic_math(PyObject *o1, PyObject *o2, int op)
         vec = (pgVector *)o1;
         other = o2;
     }
-    else {
+    else if (pgVector_Check(o2)) {
         vec = (pgVector *)o2;
         other = o1;
         op |= OP_ARG_REVERSE;
+    }
+    else {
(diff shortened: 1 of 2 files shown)


illume:
Thanks :)

PR #558, opened 2018-10-14
Raise TypeError on invalid point arguments (fix #521)

### LINKED ISSUE #521: known bug : wrong arguments in *draw.(aa)lines* and *draw.polygon* are ignored
* in *draw.c* **(aa)lines** and **polygon** Python interface functions, we have `/*note, we silently skip over bad points :[ */`
* It seems to me, it should be checked before any drawing and raise some *ValueError*
* It is kind of "low hanging fruit" to fix this bug (just loop over points before any drawing function is called).

Anyway, it's not documented...

Nunu-Willump:
Raises TypeError and not ValueError as said in issue to be consistent with the error that is raised by the python algorithm.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -249,11 +249,17 @@ aalines(PyObject *self, PyObject *arg)
         return RAISE(PyExc_ValueError,
                      "points argument must contain more than 1 points");
 
+    for(loop = 0; loop < length; ++loop) {
+        item = PySequence_GetItem(points, loop);
+        result = pg_TwoIntsFromObj(item, &x, &y);
+        Py_DECREF(item);
+        if(!result)
+            return RAISE(PyExc_TypeError, "all points must be number pairs");
+    }
+
     item = PySequence_GetItem(points, 0);
     result = pg_TwoFloatsFromObj(item, &x, &y);
     Py_DECREF(item);
-    if (!result)
-        return RAISE(PyExc_TypeError, "points must be number pairs");
 
     startx = pts[0] = x;
     starty = pts[1] = y;
@@ -268,8 +274,6 @@ aalines(PyObject *self, PyObject *arg)
         item = PySequence_GetItem(points, loop);
         result = pg_TwoFloatsFromObj(item, &x, &y);
         Py_DECREF(item);
-        if (!result)
-            continue; /*note, we silently skip over bad points :[ */
         ++drawn;
         pts[0] = startx;
         pts[1] = starty;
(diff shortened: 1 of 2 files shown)


illume:
Thanks for this :)

One improvement could be to make the repeated code into a function so it's not repeated many times.

I'm a bit worried that these extra checks could cause performance issues because we have to iterate through the lines multiple times. But I guess it's better to be correct.

illume:
I just noticed that this fixes line and polygon drawing. However there's only one test added. Would you be able to please add a test for these too?

Nunu-Willump:
Oh wow, until I checked it just now I thought that variable sized arrays on the stack weren't a thing in C and that malloc/free would be more expensive than this solution. But since I was wrong at least about the first one, I guess I'll change that code up since that seems to be a solution that is probably a lot faster, and I doubt that a few python objects will blow the stack up. Should I make a new PR for this?

illume:
Sorry, I'm a bit confused about what you mean about the stack?

I guess the logic would be quite complicated if instead of looping over the points twice, the error checking is done in line? Then you'd have weird half drawn shapes I guess... not good.

Nunu-Willump:
No, what I mean is having something like an `int x_coods[n];`, where the values for x are stored during the check and then later accessed from instead of getting them again from the python object. Of course, if someone passes too much points at once to this, this will consume a lot of memory on the stack, but I don't think it would cause a stack overflow, modern systems usually have at least a few MB of stack size by default.

illume:
Ah ok. Thanks for the explanation. Sounds like you've found a bug there with huge amounts of points.

So coming up with a list of potential improvements...

To get this merged in right away...
- [ ] add similar unit test for other functions that were modified.

Nice to have (perhaps in another PR)...
- [ ] time functions with 100, and 1000 points to compare the speed.
- [ ] look at speeding things up, iff it is slow.

Nunu-Willump:
Huh, seems like git decided to use my old account after I updated it yesterday. I'll look into why that might be and try to fix it, but does it matter?

illume:
Only if it matters to you.

It’s probably because you need to change your git email address to the one
this account uses.

Maybe best to set your git email up again, and do a new branch with the
commits. See
https://help.github.com/articles/setting-your-commit-email-address-in-git/

Up to you of course.

Nunu-Willump:
Thanks for telling me. Changed my ~/.gitconfig to use the newer one now, but as it doesn't really matter to me I'm OK with it being merged like it is now :p

e1000:
Hello,

sorry, I just realized I did a duplicate PR ... 

* I've used everywhere the `xlist` and `ylist` variables, so I can avoid this block : 
```C
    item = PySequence_GetItem(points, 0);
    result = pg_TwoFloatsFromObj(item, &x, &y);
    Py_DECREF(item);
```
I'm not sure about the performance of doing the loop two times. I thought, the steps done for each separate lines (and much more drawing the whole polygon) are much longer than repeating the loop. 
 
I thought to a macro for this repeated code, but it's tricky to say the least since `aalines` uses float coordinates (at least now : see #522)  ...

Also, I added a macro for checking the colors.

Nunu-Willump:
Closing since PR #567 does the job better

illume:
I think what we'll do is merge this PR first, and then merge the other one on top of it.

e1000:
I agree to merge this first.  I basically just save data and then do some cleanup after it. Plus, he was there before :-)

Nunu-Willump:
Maybe I'm missing something obvious here, but why are we waiting with merging these 2 PRs?

illume:
This was merged in here: https://github.com/pygame/pygame/pull/578

Issue #542, opened 2018-09-27
pygame infects stdout


metulburr:
```
>>> import pygame
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
>>>
```

Anything written which uses pygame, infects stdout with this. Unacceptable behavior for a library.

ase1590:
just remove it from the [__init__.py file](https://github.com/pygame/pygame/blob/master/src_py/__init__.py#L495) if you have such an issue with it.

metulburr:
you can remove it but the point is users should not have to

illume:
'Infects' is maybe the wrong word, but I understand where you are coming from.

- If a user wants to remove that for other people in their pygame using app, they can wrap stdout whilst they import pygame.
- Locally, people can comment out the print line in their pygame install at `pygame/__init__.py` file. `import pygame;print(pygame.__file__)`
- Or, they could help pygame 2 get done - which is when this will be considered for removal next.

Why is it there? We(and I) needed (and need) help. Because pygame needs contributors. It has millions of users, and less than 0.001% contribute anything to it. Of course putting a link there doesn't actually attract contributors, spending months cleaning up pygame, writing docs, and helping people contribute does.

Julian-O:
[PygameSilent](https://pypi.org/project/pygamesilent/) is a new shim for PyGame that applies the #559 workaround more easily than directly applying environment variables or overriding stdout at every import.

illume:
Opinions are like... you.

illume:
@Julian-O stop spamming repos. Your package is buggy and pointless. People can just use PYGAME_HIDE_SUPPORT_PROMPT. Rather than depending on a potentially dangerous random package from someone who has done nothing noteworthy.

metulburr:
Nice. Im glad you guys added this. 
PYGAME_HIDE_SUPPORT_PROMPT
Thanks.

monokal:
I'm having the same issue now with pygame spamming my project's STDOUT. Setting `os.environ['PYGAME_HIDE_SUPPORT_PROMPT'] = "hide"` isn't disabling it either for some reason (Python 3.8.1, pygame 1.9.6, pipenv venv).

monokal:
Just a note to anyone who may come across this via Google, etc. You must set the environment variable BEFORE importing pygame. E.g:

```python
os.environ['PYGAME_HIDE_SUPPORT_PROMPT'] = '1'
import pygame
```

[...]

PR #559, opened 2018-10-15
Allow clean silencing of support prompt

Bottersnike:
As nice as the prompt is, there are some situations where it's really not appropriate. At current, the best way to silence it is something like
```py
import io, sys
stdout = sys.stdout
sys.stdout = io.StringIO()
import pygame
sys.stdout = stdout
del stdout, io, sys
```
This patch would allow the user to define the `PYGAME_HIDE_SUPPORT_PROMPT` environment variable in which case importing pygame wouldn't print the extra two lines.

### DIFF
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -392,9 +392,11 @@ def __color_reduce(c):
 copy_reg.pickle(Color, __color_reduce, __color_constructor)
 
 
+# Thanks for supporting pygame. Without support now, there won't be pygame later.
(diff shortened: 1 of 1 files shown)


illume:
No.

https://github.com/pygame/pygame/issues/542

illume:
Thanks, and sorry.

Julian-O:
[PygameSilent](https://pypi.org/project/pygamesilent/) is a new shim for PyGame that applies this workaround more easily than directly applying environment variables or overriding stdout at every point.

illume:
Stop spamming our repo.

PR #560, opened 2018-10-15
Added math.Vector2 subclass test for issue #552

cmtrapp02:
Added new function that tests a subclassed math.Vector2 for issue #552.
This is my first contribution on Github, so any feedback if necessary would be greatly appreciated!

### DIFF
--- a/test/math_test.py
+++ b/test/math_test.py
@@ -1577,6 +1577,13 @@ class Vector3TypeTest(unittest.TestCase):
         self.assertEqual(pickle.loads(pickle.dumps(v2)), v2)
         self.assertEqual(pickle.loads(pickle.dumps(v3)), v3)
 
+    def testVector2Subclass(self):
+        class Vector(pygame.math.Vector2):
+            pass
+        v = Vector(2.0,2.0)
+        v *= 2
+        self.assertEqual(v, (4.0,4.0))
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/test/math_test.py
+++ b/test/math_test.py
@@ -1576,13 +1576,14 @@ class Vector3TypeTest(unittest.TestCase):
         v3 = Vector3(1, 2, 3)
         self.assertEqual(pickle.loads(pickle.dumps(v2)), v2)
         self.assertEqual(pickle.loads(pickle.dumps(v3)), v3)
-
+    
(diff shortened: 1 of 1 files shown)


illume:
Thank you!

test/math_test.py
```diff
@@ -1577,6 +1577,13 @@ def test_pickle(self):
         self.assertEqual(pickle.loads(pickle.dumps(v2)), v2)
         self.assertEqual(pickle.loads(pickle.dumps(v3)), v3)
 
+    def testVector2Subclass(self):
+        class Vector(pygame.math.Vector2):
+            pass
+        v = Vector(2.0,2.0)
```
illume: It's usual to put a space after arguments. For example: `v = Vector(2.0, 2.0)`

cmtrapp02: Okay thanks, I'll be sure to do that next time.

test/math_test.py
```diff
@@ -1577,6 +1577,13 @@ def test_pickle(self):
         self.assertEqual(pickle.loads(pickle.dumps(v2)), v2)
         self.assertEqual(pickle.loads(pickle.dumps(v3)), v3)
 
+    def testVector2Subclass(self):
```
illume: Can you please add a decorator to show that this is an expected failure?
Please see the documentation here on how to do that: https://docs.python.org/3/library/unittest.html#unittest.expectedFailure

This way the tests will pass on TravisCI and Appveyor.

test/math_test.py
illume: The pygame code isn't the cleanest yet, or using modern conventions everywhere. But eventually we will move to using [pylint](https://www.pylint.org/) to help catch these types of issues.

test/math_test.py
cmtrapp02: Yup, I'll do that as soon as I can. Thanks for the help

Nunu-Willump:
Wouldn't it be desirable to test this in a child process? The way it is now the test doesn't really fail, but instead the process just terminates with only the message "Segmentation fault (core dumped)"

illume:
Wonderful. Thanks :)

illume:
Ah yes. The CI (Appveyor and Travis) are failing because it crashes python. I'll merge this in now anyway, and merge the other one right away.

PR #563, opened 2018-10-16
Improve sysfont on mac

illume:
A mixture of branches from @Cerealdragon and @andreyx86

- a simple unit test to see it doesn't crash.
- runs on python2 and python3.
- uses fc-list first if it exists (because it is faster usually)

@Cerealdragon and @andreyx86 could you please review? Does this work for you? Is there anything that could be improved?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,6 +37,7 @@ matrix:
       language: generic
       env:
         - PY_VERSION=3
+        - PY_VERSION_=3.7
         - MAKEFLAGS=-j8
     - os: osx
       osx_image: xcode7.3
--- a/README.rst
+++ b/README.rst
@@ -126,8 +126,6 @@ Credits
      Also a big thanks to Roger Dingledine and the crew at SEUL.ORG for our
      excellent hosting.
 
-
-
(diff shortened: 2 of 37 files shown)


Cerealdragon:
@illume Can confirm that so far the latest commit works on my mac High Sierra (Version: 10.13.6), I've also noticed that this message gets printed out in the stdout:

**Python[2412:14597] ApplePersistenceIgnoreState: Existing state will not be touched. New state will be written to (null)**

This error can be fixed by executing _defaults write org.python.python ApplePersistenceIgnoreState NO_ in the terminal.

illume:
@Cerealdragon thanks. Weirdly, I don't get that message at all. Are you sure it's related to these changes?

I had to change the test to only initialise pygame.font, because it was causing failures on windows.

illume:
From testing with a few people, and looking about the interwebs, this looks like a python related thing. I'm going to merge now, but if you still think it's related to this change please let me know?

Cerealdragon:
@illume I doubt it's related to the change, looks more like something to do with Mac version to me

PR #567, opened 2018-10-17
Draw: data validate

### LINKED ISSUE #521: known bug : wrong arguments in *draw.(aa)lines* and *draw.polygon* are ignored

e1000:
* a macro for validating colours 
* `aaline` and `aalines` now also accept *integer* as color (like the other drawing functions)
* `aalines`, `lines` and `polygon` : full validation of the path points instead of skipping points
* tests for color and path validation for all drawing functions

[edit] : this closes #521

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -569,7 +569,7 @@ polygon(PyObject *self, PyObject *arg)
     SDL_Surface *surf;
     Uint8 rgba[4];
     Uint32 color;
-    int width = 0, length, loop, numpoints;
+    int width = 0, length, loop;
     int *xlist, *ylist;
     int x, y, top, left, bottom, right, result;
 
@@ -603,27 +603,23 @@ polygon(PyObject *self, PyObject *arg)
         return RAISE(PyExc_ValueError,
                      "points argument must contain more than 2 points");
 
-    item = PySequence_GetItem(points, 0);
-    result = pg_TwoIntsFromObj(item, &x, &y);
-    Py_DECREF(item);
-    if (!result)
-        return RAISE(PyExc_TypeError, "points must be number pairs");
-    left = right = x;
-    top = bottom = y;
+    left = top = 10000;
+    right = bottom = -10000;
 
     xlist = PyMem_New(int, length);
     ylist = PyMem_New(int, length);
 
-    numpoints = 0;
     for (loop = 0; loop < length; ++loop) {
         item = PySequence_GetItem(points, loop);
         result = pg_TwoIntsFromObj(item, &x, &y);
         Py_DECREF(item);
-        if (!result)
-            continue; /*note, we silently skip over bad points :[ */
-        xlist[numpoints] = x;
-        ylist[numpoints] = y;
-        ++numpoints;
+        if (!result) {
+            PyMem_Del(xlist);
+            PyMem_Del(ylist);
+            return RAISE(PyExc_TypeError, "points must be number pairs");
+        }
+        xlist[loop] = x;
+        ylist[loop] = y;
         left = MIN(x, left);
         top = MIN(y, top);
         right = MAX(x, right);
@@ -636,7 +632,7 @@ polygon(PyObject *self, PyObject *arg)
         return NULL;
     }
 
-    draw_fillpoly(surf, xlist, ylist, numpoints, color);
+    draw_fillpoly(surf, xlist, ylist, length, color);
 
     PyMem_Del(xlist);
     PyMem_Del(ylist);
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -218,13 +218,14 @@ aalines(PyObject *self, PyObject *arg)
     PyObject *surfobj, *colorobj, *closedobj, *points, *item;
     SDL_Surface *surf;
     float x, y;
-    int top, left, bottom, right;
+    float top, left, bottom, right;
     float pts[4];
     Uint8 rgba[4];
     Uint32 color;
     int closed, blend;
     int result, loop, length, drawn;
     float startx, starty;
+    float *xlist, *ylist;
 
     /*get all the arguments*/
     if (!PyArg_ParseTuple(arg, "O!OOO|i", &pgSurface_Type, &surfobj, &colorobj,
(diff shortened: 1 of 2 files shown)


e1000:
This is similar to [PR 558](https://github.com/pygame/pygame/pull/558/) , but a bit different ;)

Nunu-Willump:
Ok, this seems a somewhat better than mine since it didn't invest time optimizing out the double lookup of every coordinate, so I'll close mine...

Just out of curiosity (still pretty new to OSS, the other one was my first real PR); why is PyMem_New() used for allocating memory for the lists? They're pretty short-lived, and since I don't think someone would pass multiple hundred thousands of points at once (which could cause an stack overflow), I'd save them on the stack, possibly avoiding an expensive malloc() call by python. Though maybe I'm just overthinking stuff that isn't worth optimizing :p

e1000:
Hm, me I'm very new to **C** and **C-python**, I'm not sure about `PyMem_New()`, I just copied what has been done in `polygon` ; 

how do I "save them on the stack" ? If you tell me, I could just do the code...

and since I'm asking questions, would it be more expensive to have a list of some "Point" struct with just *x* and *y* ? Because it would be much easier to read, specially if we would use *Point* also for the *pts* list...

Nunu-Willump:
Unless the compiler doesn't optimize very well, using a struct for points should, to my limited knowledge, not impact performance. In fact, if you allocate space with something like `malloc()` it might even be faster since you're calling it just 1 time instead of 2.

You'd allocate memory for an array of such points on the stack by doing something like `Point points[count];` with `Point` being typedef'd to your point struct.

This is probably faster, but in this project a downside would be that we're declaring all the variables in the first few lines of the function, but the declaration I wrote only works if you already know the length at the time you're declaring this variable. I think I'll run a quick benchmark comparing the two, post the results, and let someone who is more familiar with this project judge if the speedup (if I even get a noticeable one) is worth it.

e1000:
I did not understand why we can't use count/length (that was calculated before...) do just

```C
Point points[count];
``` 
just before we need this `points` variable? 

like : 

```C
    int xlist[length];
    int ylist[length];
```
 ?

And if I do it that way (either with int or `Point`)  , do I need to "deallocate" ? If yes, how?
Or is this what you called  "points on the stack "?

Sorry for my questions, I'm real C-beginner, I wonder how it's even possible to do PRs :)

Nunu-Willump:
Pretty interesting results. Actually, using the stack is not as fast as I though for larger stuff.

(Drawing 1000 lines 10000 times, run 10 times or so)
| type | avg | lowest | highest|
|:------|:------|:--------|:----------|
| Your code without modification |`2.0s` | `1.86s` | `2.7s` |
| Your code with points as structs | `2.05s` | `2.0s` | `2.1s` |
| Your code with x/y on stack | `2.5s` | `1.84s` | `3.3s` |
| Your code with points as structs on stack | `2.5s` | `1.79s` | `3.6s` |

The versions using the stack were very slow the first 2-3 runs, but got very fast afterwards. It probably has to do with the kernel having to resize the stack, which seems to take longer than allocating with pythons build-in functions, but being able to predict that the process needs a lot of stack space later. When using the stack, after 5 runs there were no runs with more than 2.1s, and the fastest times were always in the last 3 runs.

It got even funnier when I dropped all caches of the system before my 10 runs; for all tests, I got results ranging from 2s-6s, sometimes getting better, sometimes getting worse over time.

___

Using the stack might be better for average FPS in the long run, but to avoid sudden lag spikes I think how you've written it might be the best solution (especially since idk how windows performs with this). Nice to learn something new about C, lol

Nunu-Willump:
You can use the length as calculated before, but it is good style to declare all variables at the start of the function, even if they're initialized later, which afaik can't be done without allocating without some function like PyMem_New()

The nice thing is that you don't have to deallocate, since you're not really "allocating" something, you're just changing the stack pointer (don't worry if you don't know what that is). But as said in my previous comment, using the stack for stuff larger than 1kb doesn't seem to be a good idea anyway :/

e1000:
Ah, interesting!

Which did you test? *lines* or **alines** ?  And how long are your lines?

Maybe you can some benchmarks for the older code?
I'd like to know if removing the buffer variables `startx, starty` and the `drawn` variable was sensible improvement...

Nunu-Willump:
I only tested `lines`, and I also benchmarked the older code and yours doesn't seem to really take longer (unlike the code from my PR which takes around 20% longer :p)

illume:
Nice work on the benchmarking and investigations into the different methods.

In python you can play around with the stack size. https://docs.python.org/3.7/library/_thread.html?highlight=thread#_thread.stack_size

It's interesting to see how python and the stack behaves when you have a bunch of threads. If you're careful with the code you run, you can create threads with pretty small amount of stack memory used.  Much lower than the default. Of course, if you allocate a lot of data on the stack then the program might crash.

There's some pretty good reading describing memory in C, and the stack is a very common part of it you will come across.
- https://craftofcoding.wordpress.com/2015/12/07/memory-in-c-the-stack-the-heap-and-static/
- https://www.gribblelab.org/CBootCamp/7_Memory_Stack_vs_Heap.html
- https://en.wikipedia.org/wiki/Call_stack (sort of hilarious that this page uses DrawLine as an example)
- https://www.geeksforgeeks.org/memory-layout-of-c-program/

On Linux you can change the stack with `setrlimit` https://linux.die.net/man/2/setrlimit and `ulimit`.

The stackless python project(and pypy compiled specially, and the greenlet extension) took advantage of using very little C stack in order to do large amounts of concurrent work at once with very low switching overhead compared to threads or processes.

With regards to malloc in python, often they cache areas of memory. So if you free and request the same size multiple times, then the malloc implementation will return much more quickly. This is sometimes known as a "memory pool". Definitely avoiding malloc is a good idea in general, but especially in real time applications where it can introduce variable latencies. These days memory IO is one of the major slow downs in most systems.

I highly recommend reading this short document on "Latency Numbers Every Programmer Should Know" if you haven't already.
- https://gist.github.com/hellerbarde/2843375

If data is in L1, or even L2 cache, processing multiple times over it isn't all that bad. Modern systems often have at least 32KB L1 cache. And most "cache lines" are 64 bytes. If your data is already in cache, you can almost do extra processing for free on faster CPUs (relatively speaking). Since the memory read latency is so long compared to the actual processing.

illume:
Going to merge these commits in here: https://github.com/pygame/pygame/pull/578

PR #569, opened 2018-10-18
Fixed freezing while playing music from file-like objects without file descriptors

dlon:
Same as #551 but general. Always releases the GIL before acquiring the audio mutex.

This will freeze in the current version:

```python
import pygame
import io

pygame.mixer.pre_init(buffer=512)
pygame.init()

displaySurface = pygame.display.set_mode((320, 240))

with open('music.ogg', 'rb') as f:
	h = io.BytesIO(f.read())
	pygame.mixer.music.load(h)
	pygame.mixer.music.play()

s = pygame.mixer.Sound('sound.wav')

clock = pygame.time.Clock()
running = True
while running:
		if event.type == pygame.QUIT or\
			(event.type == pygame.KEYDOWN and\
			event.key == pygame.K_ESCAPE):
			running = False
			break
			s.play()
	pygame.display.flip()
	clock.tick(60)
```

### DIFF
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -272,7 +272,9 @@ pgMixer_AutoQuit(void)
 {
     int i;
     if (SDL_WasInit(SDL_INIT_AUDIO)) {
+        Py_BEGIN_ALLOW_THREADS
         Mix_HaltMusic();
+        Py_END_ALLOW_THREADS
 
         if (channeldata) {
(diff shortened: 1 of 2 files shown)


illume:
Thank you.

Yes, this makes sense. I guess each of these SDL functions can call back into python, which can lead to a deadlock.

PR #570, opened 2018-10-18
draw_py.py needs docs

e1000:
add some documentation to src_py/draw_py.py according to comments on PR 535

(for some reason , I could not push some additional commits directly into the other PR, so I'm creating a new one & close the other)


illume:
Cool, thanks for this :)

PR #573, opened 2018-10-20
Fixed FreeType memory leak

### LINKED ISSUE #572: On windows os, this file handle will never be closed, this may cause python "IOError: [Errno 24] Too many open files" if fonts are created multiple times.
https://github.com/pygame/pygame/blob/f85590933bfa986300a2b38bd300eb42c05496df/src_py/freetype.py#L58

dlon:
When the font was unloaded, the wrong struct was being freed. (Does not fix #572.)


illume:
cheers!

PR #577, opened 2018-10-23
Fix FreeType SDL_RWops leak

### LINKED ISSUE #572: On windows os, this file handle will never be closed, this may cause python "IOError: [Errno 24] Too many open files" if fonts are created multiple times.

dlon:
Fixes #572. Not sure whether the first or second commit is cleaner. Added 2nd as the user may not expect pygame to close a stream that they passed to it? At least in CPython, files normally close when the ref count hits 0, so either works.

### DIFF
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -706,7 +706,7 @@ _ftfont_dealloc(pgFontObject *self)
     _PGFT_UnloadFont(self->freetype, self);
 #ifdef HAVE_PYGAME_SDL_RWOPS
     if (src) {
-        SDL_RWclose(src);
+        pgRWopsFreeFromObject(src);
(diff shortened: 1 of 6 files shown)


illume:
Is adding a unit test for this possible?

illume:
I guess your second commit would be better? Because it cleans up all the resources, whereas closing the file will just make sure the file is closed but will leave the objects around?

illume:
@dlon thanks. So, do you think this is ok to merge now?

I guess that test will fail with pypy.

dlon:
Yes

Issue #575, opened 2018-10-21
`draw.aaline` not blending with background


artifactz:
Doc says:

> `aaline(Surface, color, startpos, endpos, blend=1) -> Rect`
> Draws an anti-aliased line on a surface. This will respect the clipping rectangle. A bounding box of the affected area is returned as a rectangle. If blend is true, the shades will be be **blended with existing pixel shades instead of overwriting them**. This function accepts floating point values for the end points.

I don't get the point of `blend`, because who wouldn't want to blend the line with already present pixels?
Anyway, whichever value is handed over to `blend`, the line color is always blended with the color *black*.

Example:

    import pygame as pg

    pg.init()
    surface = pg.display.set_mode((800, 600))
    clock = pg.time.Clock()
    while True:
        for event in pg.event.get():
            if event.type == pg.KEYDOWN:
                if event.key == pg.K_q or event.key == pg.K_ESCAPE:
                    exit(0)
            elif event.type == pg.QUIT:
                exit(0)

        surface.fill(pg.Color(0, 0, 0))

        pg.draw.aaline(surface, pg.Color(255, 255, 255), (25, 20), (780, 580), 1)
        pg.draw.aaline(surface, pg.Color(255, 255, 255), (20, 80), (780, 20), 1)

        pg.display.flip()
        clock.tick(60)

Result (zoomed):

![pg-test](https://user-images.githubusercontent.com/4330620/47265496-08094200-d529-11e8-968c-3142b5a285de.png)

Note that the gray values of the second line's anti-aliasing **overwrite** the first one instead of blending.
Drawing black lines on a white background makes this even more obvious.
Same bevahior on a surface with `SRCALPHA`.

Python 3.5.2 / pygame 1.9.4 / Ubuntu 16.04

Also mentioned here: https://github.com/pygame/pygame/issues/512#issuecomment-416935045

e1000:
yes, it's known bug. see also #522

e1000:
The problems lie in the macro [DRAWPIX32](https://github.com/pygame/pygame/blob/42778d045fbf4f39d18cfbdd171b262b5aa27af9/src_c/draw.c#L986) : 

one obvious thing is that we forget to define `nbr` , we need some line like 

```diff
+        nbr = 1 - br;        \
```

But the other problem is the use of `SDL_GetRGBA` : 
```C
        SDL_GetRGBA(*pixel, surf->format, &pixel_r, &pixel_g, &pixel_b,     \
                    &pixel_a);        
[...]

PR #582, opened 2018-10-27
draw.aaline: blend correctly

dlon:
#575 

Difference:

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -32,7 +32,7 @@
 #include <math.h>
 
 /* Many C libraries seem to lack the trunc call (added in C99) */
-#define trunc(d) (((d) >= 0.0) ? (floor(d)) : (ceil(d)))
+#define trunc(d) ((int)(d))
 #define FRAC(z) ((z)-trunc(z))
 #define INVFRAC(z) (1 - FRAC(z))
 
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -102,10 +102,6 @@ aaline(PyObject *self, PyObject *arg)
         return NULL;
     surf = pgSurface_AsSurface(surfobj);
 
-    if (surf->format->BytesPerPixel != 3 && surf->format->BytesPerPixel != 4)
-        return RAISE(
-            PyExc_ValueError,
-            "unsupported bit depth for aaline draw (supports 32 & 24 bit)");
 
     CHECK_LOAD_COLOR(colorobj)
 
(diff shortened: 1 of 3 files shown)


src_c/draw.c
```diff
@@ -102,10 +102,6 @@ aaline(PyObject *self, PyObject *arg)
         return NULL;
     surf = pgSurface_AsSurface(surfobj);
 
-    if (surf->format->BytesPerPixel != 3 && surf->format->BytesPerPixel != 4)
```
illume: oh cool. I guess this means it supports more than 32/24 bit now.

src_c/draw.c
```diff
@@ -983,27 +975,75 @@ set_at(SDL_Surface *surf, int x, int y, Uint32 color)
...
+        case 1:
+            return *pixels;
+    }
+    return 0;
+}
+
+static void
+set_pixel_32(Uint8 *pixels, SDL_PixelFormat *format, Uint32 pixel)
+{
+    switch (format->BytesPerPixel) {
```
e1000: Cool! It looks similar to the code in **drawline** now, but why it is much more simple here? Can we use this code in *drawline* ?

src_c/draw.c
```diff
@@ -983,27 +975,75 @@ set_at(SDL_Surface *surf, int x, int y, Uint32 color)
...
+    if (blend) \
+        draw_pixel_blended_32(pixels, colorptr, br, surf->format); \
+    else {\
+        set_pixel_32(pixels, surf->format, \
+                     SDL_MapRGBA(surf->format, \
+                                 br * colorptr[0], \
+                                 br * colorptr[1], \
+                                 br * colorptr[2], \
+                                 br * colorptr[3])); \
+    }}
```
e1000: maybe @illume would like to apply some *clang-format* ? ;)

dlon:
Some measurements:

Not inlined:
  18985 calls/s
Inlined:
  18854 calls/s
Replaced trunc
  23240 calls/s
Replaced trunc + inlined:
  22826 calls/s

Inlining seems to make little difference when I try it, but replacing trunc() made a big difference.

Also, the code for big endian might be wrong.

src_c/draw.c
dlon: I think it's always between 1 and 4, but 2 and 1 also work now

src_c/draw.c
dlon: In drawline, the line is drawn for each case inside the switch statement, probably because it might be faster.

e1000:
* `x` is for expected Failure,
* `u` is for unexpected Success,

so it seems, no unexpected success :-/ I will investigate that "ASAP"...

PR #583, opened 2018-10-27
SDL2: Compile on Windows

dlon:
Had to make this change to compile (using MSVC) with SDL2. I also had to replace the #warning directives, but I'm not sure whether I should add that patch.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -71,8 +71,11 @@ void *
 _alloca(size_t size);
 #define alloca _alloca
 #endif
-
 #endif
+
+#define PG_STRINGIZE_HELPER(x) #x
+#define PG_STRINGIZE(x) PG_STRINGIZE_HELPER(x)
+#define PG_WARN(desc) message(__FILE__ "(" PG_STRINGIZE(__LINE__) "): WARNING: " #desc)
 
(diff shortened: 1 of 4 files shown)


illume:
Cool, thanks :)

I guess we'll need to:
- use a portable warning macro. https://stackoverflow.com/questions/171435/portability-of-warning-preprocessor-directive
- or, comment out those warnings.

I think it's more important to have windows compiling than have the warnings appear at the moment.

So... let's comment out the warnings until someone has time to make a warning macro?

illume:
There's a section on the wiki for how to compile ubuntu and mac with sdl2.

So probably we also need to add a section on the wiki with instructions here:
https://www.pygame.org/wiki/CompileWindows

dlon:
Maybe [this](https://stackoverflow.com/a/11244789/1239986) for the warning macro?

```c
#define PG_STRINGIZE_HELPER(x) #x
#define PG_STRINGIZE(x) PG_STRINGIZE_HELPER(x)
#define PG_WARN(desc) message(__FILE__ "(" PG_STRINGIZE(__LINE__) "): WARNING: " #desc)
```

illume:
Nice one :)

PR #591, opened 2018-10-30
Undefined default parameter value

dlon:
Calling aalines() without setting blend had undefined behavior. Could potentially equal 0


e1000:
After merging PR #585, maybe we should just repeat the test with **blend** not passed to the function `draw` ?

illume:
Thanks again :)

PR #596, opened 2018-11-03
Clean aaline

e1000:
* much of variable renaming in `aaline` to get things clearer
* a `aaline` algorithm in Python 
* little `draw_test.py` cleanup
* make 2 `aaline` tests pass for Python

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -51,12 +51,12 @@ clip_and_draw_line_width(SDL_Surface *surf, SDL_Rect *rect, Uint32 color,
 static int
 clipline(int *pts, int left, int top, int right, int bottom);
 static int
-clipaaline(float *pts, int left, int top, int right, int bottom);
+clip_aaline(float *pts, int left, int top, int right, int bottom);
 static void
 drawline(SDL_Surface *surf, Uint32 color, int startx, int starty, int endx,
          int endy);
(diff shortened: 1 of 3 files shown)


e1000:
I don't know if it's considered a good practice in C to return inside a loop like a did in `draw.c` around lines 838-842 and 899-902 : did it because it saves one variable definition ...

e1000:
More remarks: 
* I dropped the half-integers in the Py-algo
* I have to add `+ 1` at the end of the line to get anything reasonable ... 
* going through the algorithm, I don't understand the usage of `trunc`, it seems to me, it should be `ceil` or `floor` (cf my Python-version of the algo)
* I'll have to fix some more tests (cf #522)

illume:
- trunc always rounds towards zero, so for negative numbers it's different to ceil or floor. dlon tested before that perhaps trunc is slower than a simple int() cast.
- Usually it's good to only return in one place in a function... if possible. The reasoning is that otherwise it's easy to miss the places where the function could return. However in this case, they are very close to each other.

e1000:
Can somebody help with the failing tests? it's strange messages from apple, and I'm more about piguins...

illume:
Looks like the mac build bots were having issues. Restarted the jobs, and they don't fail now.

PR #598, opened 2018-11-04
Create GL context in SDL2.

dlon:
I can run examples/glcube.py with this patch. There should probably be some tests for OpenGL as well.


illume:
Nice one :)

This reminds me... we have an issue when using an OpenGL renderer in SDL2. https://github.com/pygame/pygame/issues/601 Not sure how to do this one really.

PR #604, opened 2018-11-05
Adding multi touch and finger events for SDL2 builds.

illume:
https://wiki.libsdl.org/SDL_TouchFingerEvent
https://wiki.libsdl.org/SDL_MultiGestureEvent
https://github.com/SDL-mirror/SDL/blob/master/test/testgesture.c

---

### open questions?

- [x] What do we do in the case of SDL1 builds? It would perhaps be good to have constants available for the events, even though they are not used. On the other hand, these event constant shouldn't be actually used to post events and such.
- [x] How do we handle documentation? I guess adding in the new events and putting a 'new in pygame 2' note in there would be fine. Also a note on how it needs an SDL2 build.

You can play with it and see them with the event list example.
`python -m pygame.example.eventlist`


illume:
For now I'm going to define the constants when compiled with SDL1 as an event type of -1. This way it would be possible to write an app that can handle the events, but they just never would come in.

Now I'm working on documentation and an example.

PR #607, opened 2018-11-06
Draw aaline : fix tests, docs, new C algo

### LINKED ISSUE #522: What is draw.aaline supposed to draw with float point coordinates?

e1000:
Aaline cleanup
==============

* aaline : rewrite tests for float endpoints
* add some documentation in the Python-algo for `_draw_aaline`
* fix Py-algo
* rewrite **C** algo sticking most closely to Py-algo, hopefully a bit faster
* With this branch, we can now close #522 .
* I had some trouble with macros and int/float casting, so don't hesitate if something can be simplified


src_c/draw.c
```diff
@@ -1079,48 +1101,37 @@ draw_aaline(SDL_Surface *surf, Uint32 color, float from_x, float from_y, float t
...
-        xgap = FRAC(to_x); /* this also differs from Hugo's description. */
-        ito_x = (int)pt_x;
-        ito_y = (int)pt_y;
-        brightness1 = INVFRAC(pt_y) * xgap;
-        brightness2 = FRAC(pt_y) * xgap;
-        pixel = surf_pmap + pixx * ito_x + pixy * ito_y;
-        DRAWPIX32(pixel, colorptr, brightness1, blend)
-        pixel += pixy;
-        DRAWPIX32(pixel, colorptr, brightness2, blend)
+        G_x = ceil(from_x);
```
dlon: Nice. Still not sure about this: Why use ceil rather than round (or trunc(x+0.5))? Won't that always round to the right or bottom pixel instead of the nearest pixel?

e1000:
I added a commit to fix the case of identical float-coordinates in Py-algo + tests : if start end points are identical, we just take the integer part and draw one pixel. 
We could do some blending here, but it seems overshoot I think..

e1000:
I start again the whole **aaline** process, bisect the C-code to find the reason of crash.

Hence closing this PR.

PR #611, opened 2018-11-08
Border Aaline : fix *aaline* on the border of the clip-rect (closes #512)

### LINKED ISSUE #512: Lines drawn with aaline() and aalines() contain gaps or aren't drawn
Using `draw.aaline(s)` to draw lines sometimes doesn't work at all (when close to the border of the surface) or contains gaps.

What's left to do:

- [x] Example showing error
- [x] Write a unit test in `test/draw_test.py`
- [ ] Create a fix in [src_c/draw.c](https://github.com/pygame/pygame/blob/master/src_c/draw.c)

---

For some reason `draw.aaline` does not draw anything when it's on the border of a surface (not sure if it's intended).

```python
import pygame
from pygame.locals import QUIT
pygame.init()
surface = pygame.display.set_mode((50, 50))
pygame.draw.aaline(surface, (0, 0, 255), (0, 0), (50, 50))

print(surface.get_at((0, 0)))
going = True
while going:
    pygame.display.update()
        if event.type == QUIT:
            going = False
```

gives the following image:

image

Same thing happens when drawing lines around the border with `aalines()`. Even when drawing it just inside the border of the surface it gives lines that contain gaps, as seen below:

```python
pygame.draw.aalines(surface, (0, 0, 255), True, [(1, 1), (48, 1), (48, 48), (1, 48)])
```
image

Even this

```
pygame.draw.aalines(surface, (0, 0, 255), True, [(1, 1), (49, 0), (49, 49), (0, 49)])
```

only draws 1 pixel at `(1, 1)`.

e1000:
* This branch can only be merged after **aaline-tests-docs** because it just follows that branch
* we need to use a wider clipping rectangle and test explicitely for each point if it is in the 
  clipping rectangle, like in *set_at* in `draw.c`
* no more expected failures in `draw_test.py`
* closes #512 

We could avoid testing *each* point by seperating border areas and inner areas, but only on the price on making the algorithm more complicated...


illume:
Seems to be crashing on windows and linux build bots, but not on mac.

e1000:
As far as I understand the logs, the tests are even not run? Unfortunately, I see nothing that I could relate to my PR; except maybe that **trunc** has become unused?

e1000:
So yes, I want some help ;)

src_c/draw.c
```diff
@@ -1020,53 +1020,58 @@ set_pixel_32(Uint8 *pixels, SDL_PixelFormat *format, Uint32 pixel)
...
-                            br *colorptr[2], br *colorptr[3]));             \
-        }                                                                   \
+    if (x < surf->clip_rect.x || x >= surf->clip_rect.x + surf->clip_rect.w ||
+        y < surf->clip_rect.y || y >= surf->clip_rect.y + surf->clip_rect.h) {
+        // the (x, y) can happen to be a bit over the border because we need
+        // the extra space to make draw_aaline work correctly on the border
+        return;
     }
-
+    SDL_PixelFormat* format = surf->format;
```
dlon: Some old C compilers require that variables be declared at the start of a bracket/block, so at the start of the function in this case.
That's probably why the first compilation failed. I don't know why the other tests failed.

e1000: Until now, no test failed because no test was executed :) I fixed the variable declaration. Let's see if it changes something.

e1000: Well, it didn't change anything...
Unfortunately, we can not do some *bisect* because all the C code is in one commit, and it's a complete re-write. It must _somehow_ have to do with the automated-testing framsworks because on my Debian **stretch** it all works and all tests pass!

```bash
$ uname -a
Linux terbium 4.9.0-8-amd64 #1 SMP Debian 4.9.130-2 (2018-10-27) x86_64 GNU/Linux

$ python
Python 3.5.3 (default, Jan 19 2017, 14:11:04) 
[GCC 6.3.0 20170118] on linux
```

illume:
Perhaps the issue will show up locally when using different video/audio drivers.

This is how the tests are run on the CI servers.
```
SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=disk python -m pygame.tests.__main__ --exclude opengl --time_out 300
```

illume:
I merged master into here. Also I changed the base to master to make it easier to read the changes in the PR.

illume:
I was able to reproduce the segfault on Ubuntu with:
```python
SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=disk python test/__main__.py
```

But not on Mac.

illume:
I could find which tests were failing with this...
```python
SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=disk python test/draw_test.py -v
```
The -v option is verbose (run it with --help to see all options).

These are the two stalling tests.
- test_lines_color
- test_lines_gaps

I had to ctrl+z and then kill the test runner with...
```
kill -9 `jobs -p`
```

illume:
Then I removed the `build` and `dist` folders and rebuilt with:
```
python setup.py build --debug
```
So it has debug symbols in it.

The real issue was being hidden. Because there was a memory error, it told me there was an issue inside set_mode. So I changed the unit test to init/quit for each test inside LineMixin.

```
$ SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=disk gdb python
...
(gdb) run test/__main__.py
...
(gdb) backtrace
#0  0x00007ffff626f399 in SDL_GetRGBA () from /usr/lib/x86_64-linux-gnu/libSDL-1.2.so.0
#1  0x00007ffff04fa80f in draw_aaline (blend=1, to_y=1, to_x=1, from_y=<optimized out>, 
    from_x=<optimized out>, color=123456, surf=0xed50b0) at src_c/draw.c:1098
#2  clip_and_draw_aaline (surf=0xed50b0, rect=<optimized out>, color=123456, 
    pts=<optimized out>, blend=1) at src_c/draw.c:689
#3  0x00007ffff04fb747 in aaline (self=<optimized out>, arg=<optimized out>)
    at src_c/draw.c:128
...
```

There were also some failures inside pygame.display.set_mode.
Perhaps in lines_set_up should not use set_mode over and over again in a loop.

Also, I was getting failures within surf_init. So I'm pretty sure it's writing all over memory and causing these random failures.

illume:
Inside test_lines_color, and test_lines_gaps I was able to comment out:

```python
            # draw_lines(surface, color, True, points)
```
And the tests would run without segfault.

illume:
It appears `test_lines_gaps` when I disable `draw.aalines` is the issue.

When I comment out `draw.aalines` it seems to run all the way through.
```
class DrawLineTest(LineMixin, unittest.TestCase):
    '''Test draw functions "aaline", "line", "aalines" and "lines".'''

    single_line = [draw.aaline, draw.line]
    # multi_line = [draw.aalines, draw.lines]
    multi_line = [draw.lines]
```

illume:
Using a gdb breakpoint I was able to get close to the crash.
```
break src_c/draw.c:1098 if blend=1 && to_y=1.0 && to_x=1.0
run test/__main__.py
cont 680

Thread 1 "python" hit Breakpoint 1, draw_aaline (surf=0xf00720, color=4294967295, from_x=0, 
    from_y=49, to_x=0, to_y=0, blend=1) at src_c/draw.c:1098
```

And then put a couple of printf's in there too.
```c
static void
draw_aaline(SDL_Surface *surf, Uint32 color, float from_x, float from_y, float to_x,
           float to_y, int blend)
{
    float slope, dx, dy, swaptmp;
    int x, y;
    float G_x, G_y, S_x, S_y, pt_x, pt_y, rest, brightness;

    int pixx, pixy;
    Uint8 colorptr[4];

    Uint8 *pixel;
    Uint8 *surf_pmap = (Uint8 *)surf->pixels;
    printf("draw_aaline... from_x:%f, from_y:%f, to_x:%f, to_y:%f, blend:%d\n", from_x, from_y, to_x, to_y, blend);
    printf("surf->clip_rect.x:%d surf->clip_rect.y:%d surf->clip_rect.w:%d surf->clip_rect.h:%d\n", surf->clip_rect.x, surf->clip_rect.y, surf->clip_rect.w, surf->clip_rect.h);
    fflush(stdout);


And it crashes with these values.
```
draw_aaline... from_x:0.000000, from_y:49.000000, to_x:0.000000, to_y:0.000000, blend:1
surf->clip_rect.x:0 surf->clip_rect.y:0 surf->clip_rect.w:50 surf->clip_rect.h:50
```

**Comment by illume** [comment]
See here where the rect returned is (0,0,51,51)? It shouldn't be bigger than the input surface.

I guess the ceil() call or something like that is making it bigger than 50.

```python
>>> import pygame
>>> surface = pygame.Surface((50,50), pygame.SRCALPHA, 32)
>>> pygame.draw.aalines(surface, (255, 255, 255), True, [(0, 0), (49, 0), (49, 49), (0, 49)])
draw_aaline... from_x:0.000000, from_y:0.000000, to_x:49.000000, to_y:0.000000, blend:1
surf->clip_rect.x:0 surf->clip_rect.y:0 surf->clip_rect.w:50 surf->clip_rect.h:50
draw_aaline... from_x:49.000000, from_y:0.000000, to_x:49.000000, to_y:49.000000, blend:1
surf->clip_rect.x:0 surf->clip_rect.y:0 surf->clip_rect.w:50 surf->clip_rect.h:50
draw_aaline... from_x:49.000000, from_y:49.000000, to_x:0.000000, to_y:49.000000, blend:1
surf->clip_rect.x:0 surf->clip_rect.y:0 surf->clip_rect.w:50 surf->clip_rect.h:50
draw_aaline... from_x:0.000000, from_y:49.000000, to_x:0.000000, to_y:0.000000, blend:1
surf->clip_rect.x:0 surf->clip_rect.y:0 surf->clip_rect.w:50 surf->clip_rect.h:50
<rect(0, 0, 51, 51)>
```

illume:
@e1000 ok, back to you.

e1000:
ok I could reproduce almost exactly the behavior on Debian which @illume posted for Ubuntu; the same tests crashing etc. I believe I solved some issues with last commit, including too large *Rect*. 

But crashes continue to occur: currently it segaults at `SDL_GetRGBA` (draw.c l.1108) which I don't understand?

When I remove the color `123456` in *test_color_validation*, it crashes at `test_lines_color` , couldn't find out for which combination yet...

dlon:
This seems to have been fixed by #802. Closing

PR #612, opened 2018-11-08
SDL2: Implement key repeats

### LINKED ISSUE #527: SDL2 implement pg_EnableKeyRepeat event.c
https://www.pygame.org/docs/ref/key.html#pygame.key.get_repeat

dlon:
Fixes #527 

~~Is a 2 KB array too large?
```SDL_TimerID _scancode_to_timer[SDL_NUM_SCANCODES]```~~

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -247,6 +247,7 @@ typedef enum {
     PGE_EVENTBEGIN = SDL_NUMEVENTS,
     SDL_VIDEORESIZE,
     SDL_VIDEOEXPOSE,
+    PGE_KEYREPEAT,
     PGE_EVENTEND
 } PygameEventCode;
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by illume** [comment]
Cool :)

I got around to trying it... but couldn't get it working.

With `examples/scroll.py` I get this error:
```python
  File "examples/scroll.py", line 195, in <module>
    main(image_file)
  File "examples/scroll.py", line 116, in main
    pygame.key.set_repeat (500, 30)
pygame.error: Couldn't open /etc/timidity/freepats.cfg
```

I guess that error is just the last SDL error before the real error.
Tried with python2.7, and python3.7 on mac.

dlon:
Works for me. How about this?
```python3
import pygame as pg

pg.display.init()
surf = pg.display.set_mode((123,123))
surf.fill((123,123,123))

pg.key.set_repeat(500, 100)

running = True

while running:
    for event in pg.event.get():
        if event.type in (pg.KEYDOWN, pg.KEYUP):
            print(event)
            if event.key == pg.K_ESCAPE:
                running = False
```

Issue #454, opened 2018-05-25
Invalid path for Sound() not detected 


artag:
If I open a file with Sound(), and the file doesn't open correctly, no error is reported.

I initially noted this when trying to open  a file as ~/sounds2/file.wav. If the file is played, it sounds just a click. If play(-1) is used, it generates many clicks in the form of a screech.

1. It appears that ~ is not a valid path. Only a cwd-relative or absolute path is workable.
2. If some other error occurs (such as a .wav file that's in some incorrect format), again, no error is reported and an empty buffer is created.

illume:
Related issue: https://github.com/pygame/pygame/issues/417

PR #615, opened 2018-11-09
SDL2: Adding float 32bit sound support when using SDL2.

illume:
- [x] add some more example wav files in as tests (there are a few 32bit tests for mixer, no 32bit wav files)
- [x] add `pygame.mixer.pre_init(44100, 32, 2, 1024)` in examples/aliens.py when sdl2 is used. So this can be use to test if setting pre init to 32 bit works.
- [x] remove `pre_init` in `__init__.py` and let SDL2 do it's thing.
- [x] making a Sound from a ">f" format numpy array fails
- [x] fix for Sound('some-wrong-filename.wav') #454 on python2.

### DIFF
--- a/docs/reST/ref/mixer.rst
+++ b/docs/reST/ref/mixer.rst
@@ -54,6 +54,8 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
    values mean unsigned audio samples will be used. An invalid value raises an
    exception.
 
+   New in pygame 2 compiled with SDL2 size can be 32, which means 32bit floats.
+
    The channels argument is used to specify whether to use mono or stereo. 1
    for mono and 2 for stereo. No other values are supported (negative values
    are treated as 1, values greater than 2 as 2).
--- a/docs/reST/ref/mixer.rst
+++ b/docs/reST/ref/mixer.rst
@@ -54,7 +54,7 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
    values mean unsigned audio samples will be used. An invalid value raises an
    exception.
 
-   New in pygame 2 compiled with SDL2 size can be 32, which means 32bit floats.
+   New in pygame 2(when compiled with SDL2) - size can be 32 (32bit floats).
 
    The channels argument is used to specify whether to use mono or stereo. 1
    for mono and 2 for stereo. No other values are supported (negative values
--- a/examples/aliens.py
+++ b/examples/aliens.py
@@ -178,6 +178,8 @@ class Score(pygame.sprite.Sprite):
 
 def main(winstyle = 0):
     # Initialize pygame
+    if pygame.get_sdl_version()[0] == 2:
+        pygame.mixer.pre_init(44100, 32, 2, 1024)
     pygame.init()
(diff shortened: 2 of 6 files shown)


illume:
There is also a fix in here for when sounds finish playing, whilst other python threads are running.

illume:
I tried the deadlock example from: https://github.com/pygame/pygame/pull/569
It does not deadlock with the threading change in this PR (for me).
```python
import pygame
import io

pygame.mixer.pre_init(buffer=512)
pygame.init()

displaySurface = pygame.display.set_mode((320, 240))

with open('examples/data/house_lo.ogg', 'rb') as f:
	h = io.BytesIO(f.read())
	pygame.mixer.music.load(h)
	pygame.mixer.music.play()

s = pygame.mixer.Sound('sound.wav')

clock = pygame.time.Clock()
running = True
while running:
		if event.type == pygame.QUIT or\
			(event.type == pygame.KEYDOWN and\
			event.key == pygame.K_ESCAPE):
			running = False
			break
			s.play()
	pygame.display.flip()
	clock.tick(60)
```

docs/reST/ref/mixer.rst
```diff
@@ -54,6 +54,8 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
    values mean unsigned audio samples will be used. An invalid value raises an
    exception.
 
+   New in pygame 2 compiled with SDL2 size can be 32, which means 32bit floats.
```
Zireael07: 'New' should be 'now'?

illume:
@dlon Oh right. Yes, that seems to make sense.

illume:
@dlon I tried acquiring the GIL that endsound_callback. However, then I got some crashes. I put the allow threads call around Mix_FreeChunk as well, and that also had issues.

The error I get is:
```
Fatal Python error: PyEval_SaveThread: NULL tstate
```
How it is now, I tried various thread using sound programs and couldn't reproduce an issue anymore.

dlon:
I meant like this:
```c
        if (channeldata[channel].queue) {
            PyGILState_STATE gstate = PyGILState_Ensure();
            int channelnum;
            Mix_Chunk *sound = pgSound_AsChunk(channeldata[channel].queue);
            Py_XDECREF(channeldata[channel].sound);
            channeldata[channel].sound = channeldata[channel].queue;
            channeldata[channel].queue = NULL;
            PyGILState_Release(gstate);
            channelnum = Mix_PlayChannelTimed(channel, sound, 0, -1);
            if (channelnum != -1)
                Mix_GroupChannel(channelnum, (intptr_t)sound);
        }
        else {
            PyGILState_STATE gstate = PyGILState_Ensure();
            Py_XDECREF(channeldata[channel].sound);
            channeldata[channel].sound = NULL;
            PyGILState_Release(gstate);
        }
```
Not sure if it's related to the first crash, but it seems to be unsafe in other places, eg channel.stop(), too.

docs/reST/ref/mixer.rst
illume: I updated the wording.

PR #620, opened 2018-11-11
Implement key.name for SDL2

### LINKED ISSUE #533: SDL2 FAIL: test_name (test.key_test.KeyModuleTest)
```
FAIL: test_name (test.key_test.KeyModuleTest)
  File "/Users/rene/dev/pygame/pygame/test/key_test.py", line 54, in test_name
    self.assertEqual(pygame.key.name(pygame.K_RETURN), "return")
AssertionError: '(' != 'return'
```

dlon:
* Switch from using scan codes to (mostly) key codes.
* Implement key.name() for SDL2.
* For backward compatibility, use old strings (from SDL1) where possible. It may not cover unusual cases (some constants were removed).

Closes #533.

## Questions
- [ ] Is there any reason not to use key codes instead of scan codes?
- [x] ~~Should the scan code constants added to .constants be kept?~~


BastiHz:
I admit I don't fully understand the difference between keycodes and scancodes. But doesn't removing scancodes make it more annying/harder to write games for non-us keyboard layouts?

dlon:
@BastiHz It might, but key events still have a scancode field. Did pygame use scancodes before? If so, I wonder why key events had a "scancode" field as well as a "key" field.

illume:
The SDL migration guide has a section on input:
https://wiki.libsdl.org/MigrationGuide#Input

There are these sections in the wiki.
https://wiki.libsdl.org/SDL_KeyboardEvent
https://wiki.libsdl.org/SDL_TextInputEvent
https://wiki.libsdl.org/SDL_TextEditingEvent

There is also this event: SDL_KEYMAPCHANGED which "`keymap changed due to a system event such as an input language or keyboard layout change (>= SDL 2.0.4)`".

What do you think about starting to add relevant documentation URLs to each of the modules or functions?

Related to this is this issue about SDL_TEXTINPUT:
https://github.com/pygame/pygame/issues/623

illume:
Scancodes are useful when you want to use the position on the keyboard. But not what the letter is.
eg. Where Z is on a US keyboard, is where Y is on a German one.

Scancode == z, means on a German keyboard it will press where the Y key is. (theoretically, there are some weird things on some keyboards... but mostly this will be true).

To answer the question: Using key codes seems the best way to go for backwards compatibility. (there is the scancode on the event for people who want to use that) Using SDL_TEXTINPUT seems the best way going forward, to be able to handle all text input properly.

PR #624, opened 2018-11-15
SDL2: Improve unicode field (take into account modifiers, etc.)

### LINKED ISSUE #621: SDL2: unicode field of KEYDOWN event ignores most modifiers
Using Python 3.7 and Windows 10.

dlon:
I used SDL_TextInputEvent instead. Keys that don't emit an input event and accents didn't work properly before. Now if you press ` and then o, for example, it will output ò.

One problem is that I'm not sure whether SDL_TEXTINPUT is guaranteed to come after SDL_KEYDOWN instead of before.

Related to #621.

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -102,6 +102,8 @@ pgEvent_AutoInit(PyObject *self, PyObject *args)
     Py_RETURN_NONE;
 }
 
+static char _pg_last_unicode_char[5] = { 1, 0, 0, 0, 0 };
+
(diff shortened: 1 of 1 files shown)


illume:
I'm not sure about if "SDL_TEXTINPUT is guaranteed to come after SDL_KEYDOWN", or know how to find out. But we can probably test it out to see how it works on different platforms. And it's easier for people to test this when it's merged in :)

illume:
For me the `unicode` field is coming in as the empty string.
`Mac pygame 2.0.0.dev3 (SDL 2.0.10, python 3.7.4)`

PR #625, opened 2018-11-15
Constants: Fix macro expansion for macro constants

dlon:
Constants that were defined as macros were expanded before being stringized.

E.g., DEC_CONSTN and DEC_CONSTS were defined as
```c
#define DEC_CONSTS(x, y)                             \
    if (PyModule_AddIntConstant(module, #x, (int)y)) \
#define DEC_CONSTN(x) DEC_CONSTS(x, x)
```

So `DEC_CONSTN(KMOD_CTRL)` evaluated to `DEC_CONSTS(0, 0)`.


illume:
Great. Thanks for adding tests for this as well :)

ps. Looks like one of the travis mac hosts had a temporary network error. I restarted the job.

Issue #629, opened 2018-11-20
freetype rotation rendering bug


CAIQT:
When rendering with customized rotation using freetype, I found the following problem:
Increment the degree from 0 to 90, the rendered character appears to be disordered when the degree exceeds 63. The code to reproduce is in the following:

```python
import pygame
import pygame.freetype
from pygame.locals import *
if __name__ == "__main__":
    pygame.init()
    font = pygame.freetype.Font(None, 36)
    screen = pygame.display.set_mode((800, 600))
    screen.fill (pygame.Color(200, 200, 200))
    MM=16
    for i in range(0, 6*MM):
        font.render_to(screen, (100+(i%MM)*40, 200+i//MM*36), "|", pygame.Color('black' if i<64 else 'red'), size=36, rotation=i)
    pygame.display.flip()

    while 1:
        if pygame.event.wait().type in (QUIT, KEYDOWN, MOUSEBUTTONDOWN):
            break
```
The result image (note problems with the *RED* chars)
![untitled](https://user-images.githubusercontent.com/3953402/48759373-db8a4680-ecdd-11e8-9ba9-a4bf7251f184.png)

illume:
Hello. Thanks again for the really well done bug report :)

For anyone else reading... there's a related issue here: https://github.com/pygame/pygame/issues/565

CAIQT:
A pull request #630 to fix this bug

illume:
Thanks for the fix :)

PR #630, opened 2018-11-20
freetype rotation rendering bugfix

CAIQT:
The bug #629  is caused by the mistaken cache key calculation.


src_c/freetype/ft_cache.c
```diff
@@ -63,7 +63,7 @@ set_node_key(NodeKey *key, GlyphIndex_t id, const FontRenderMode *mode)
     KeyFields *fields = &key->fields;
     const FT_UInt16 style_mask = ~(FT_STYLE_UNDERLINE);
     const FT_UInt16 rflag_mask = ~(FT_RFLAG_VERTICAL | FT_RFLAG_KERNING);
-    unsigned short rot = (unsigned short)FX6_TRUNC(mode->rotation_angle);
+    unsigned short rot = (unsigned short)(((unsigned int)(mode->rotation_angle))>>16);
```
dlon: Might be cleaner or more consistent to use a macro? `#define FX16_TRUNC(x) ((x) >> 16)`

illume:
For the large angles, perhaps we could use a mod function on the angle first.
```
>>> -10810 % 360
350
```

Issue #603, opened 2018-11-05
SDL2 set_color_key chimp.py not working


illume:
You can see the issue with the examples/chimp.py.
The fist and the monkey do not have transparency.

Somehow the color key transparency is not working.

MacOS 10.13.6, python2.

illume:
Seems the chimp still has issues. I'm debugging it now.

markph0204:
Seeing transparency -- and quite a funny test. :)

py37;SDL209;pygame195dev

illume:
Thanks @markph0204. closing.

Issue #600, opened 2018-11-04
SDL2 surface tests fail


dlon:
```
loading test.surface_test
FE.E.....E...........F...................s.......s.F..............
======================================================================
ERROR: test_blit_blend (test.surface_test.SurfaceBlendTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\David\Documents\dev\pygame\test\surface_test.py", line 1940, in test_blit_blend
    [masks[1], masks[2], masks[0], masks[3]])
ValueError: Invalid mask values

======================================================================
ERROR: test_blit_blend_rgba (test.surface_test.SurfaceBlendTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\David\Documents\dev\pygame\test\surface_test.py", line 2031, in test_blit_blend_rgba
    (masks[1], masks[2], masks[3], masks[0]))
ValueError: Invalid mask values

======================================================================
ERROR: test_array_interface_masks (test.surface_test.SurfaceGetBufferTest)
Test non-default color byte orders on 3D views
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\David\Documents\dev\pygame\test\surface_test.py", line 1518, in test_array_interface_masks
    pygame.Surface(sz, 0, 24, masks).get_view, '3')
ValueError: Invalid mask values

======================================================================
FAIL: test_GET_PIXELVALS (test.surface_test.SurfaceBlendTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\David\Documents\dev\pygame\test\surface_test.py", line 2098, in test_GET_PIXELVALS
    self.assertEqual(dst.get_at((0, 0)), (0, 0, 0, 255))
AssertionError: (0, 0, 0, 128) != (0, 0, 0, 255)

======================================================================
FAIL: test_overlap_check (test.surface_test.SurfaceSelfBlitTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\David\Documents\dev\pygame\test\surface_test.py", line 2241, in test_overlap_check
    self.assertEqual(surf.get_at(test_posn), rectc_right)
AssertionError: (128, 64, 32, 255) != (255, 255, 255, 255)

======================================================================
FAIL: test_image_convert_bug_131 (test.surface_test.SurfaceTypeTest)
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #633, opened 2018-11-21
Blit to surfaces properly with alpha blending or color keys

dlon:
The code looked for the flags SDL_SRCALPHA and SDL_SRCCOLORKEY even though they're only used in SDL 1.

A couple of tests from #600 no longer fail: `test_colorkey` and `test_blanket_alpha`.

examples/chimp.py seems to work as well. #603


illume:
Oh! Very good :)

PR #634, opened 2018-11-23
WIP: SDL2 audio devices, and _sdl2 done with cython.

illume:
This starts a new _sdl2 module using Cython (https://cython.org/).
The idea being that new SDL2 functionality can live in there for the moment in 'experimental' mode.

The code in this PR is pretty close to being done.
- AudioDevice - for recording and playback of audio.
- get_num_audio_devices and get_audio_device_name - for getting audio devices.

Two new event types (For when devices are added and removed)
- AUDIODEVICEADDED 
- AUDIODEVICEREMOVED

There is an examples/audiorecording.py included, which records some sound and then plays it back. I've also used it in another real program to give it some good testing.

## why Cython?

I think we had some sort of consensus on the pygame mailing list that Cython could be used for new SDL2 functionality(and even porting some older functionality would be ok). It's maybe even easier to wrap functions than ctypes.

Note that I added the cython sdl2.c file, like we have done with the portmidi pypm.pyx file. This way people do not need cython installed to compile pygame.

### how to compile with Cython?

```
cython src_c/_sdl2.pyx && python setup.py install
```

## Open questions?
- [ ] I'm not sure if we can define the documentation in our .pyx file and have sphinx use it. This would be much nicer than having to maintain two sets of documentation.
- [ ] Not sure where to put it. Putting it in pygame.mixer seems sort of sensible, except it can be used without requiring SDL_mixer. So perhaps pygame.audio or just pygame.base.AudioDevice...
- [ ] I think an _sdl2 module would be nice for SDL2 stuff, even keeping as much to the API as possible (but making it OO where appropriate like in the case of AudioDevice).
- [ ] @dlon I think after merging this, I can merge your Renderer/Window branch and convert your ctypes code into Cython.

Also...

- [ ] Needs tests.


illume:
Unit tests for this can come in a separate branch.

PR #635, opened 2018-11-23
Fix a couple of warnings, and get color key blit working on SDL2.

illume:
The chimp example wasn't working still for me.
https://github.com/pygame/pygame/issues/603

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -181,12 +181,15 @@ SoftBlitPyGame (SDL_Surface * src, SDL_Rect * srcrect, SDL_Surface * dst,
                 alphablit_solid (&info);
             break;
 #else /* IS_SDLv2 */
-            if (SDL_ISPIXELFORMAT_ALPHA (src->format->format))
-                alphablit_alpha (&info);
-            else if (info.src_has_colorkey)
+            if (info.src_has_colorkey) {
                 alphablit_colorkey (&info);
-            else
+            }
+            else if (SDL_ISPIXELFORMAT_ALPHA (src->format->format)) {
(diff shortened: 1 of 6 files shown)


src_c/alphablit.c
```diff
@@ -181,12 +181,15 @@ SoftBlitPyGame (SDL_Surface * src, SDL_Rect * srcrect, SDL_Surface * dst,
...
             break;
 #else /* IS_SDLv2 */
-            if (SDL_ISPIXELFORMAT_ALPHA (src->format->format))
-                alphablit_alpha (&info);
-            else if (info.src_has_colorkey)
+            if (info.src_has_colorkey) {
                 alphablit_colorkey (&info);
-            else
+            }
+            else if (SDL_ISPIXELFORMAT_ALPHA (src->format->format)) {
```
dlon: Should it check whether blending is enabled here?

dlon: I'm not sure but I think it should check whether SDL_BLENDMODE_BLEND is set with SDL_GetSurfaceBlendMode(). The tests in #639 should cover it.

illume:
I tested [solarwolf](https://github.com/pygame/solarwolf) and noticed a few other issues with color key transparency.

- I fixed an issue with the pygame.transform functions not respecting RLE, color keys.
- Also, now get_colorkey is working better when an image is loaded with a color key, and then surface.convert() ed.

However there's still a few issues with solarwolf. It seems the images are not blending properly, and also there are a few weird colors in some of the images.

illume:
Further fixes and cleanups for this can happen in separate PRs.

PR #637, opened 2018-11-24
Fix SDL_SYSWMEVENT on non-Windows platforms. 

### LINKED ISSUE #636: SDL_SYSWMEVENT broken in master on non-Win32 platforms
PR #583 seems to break `SDL_SYSWMEVENT` event handling on non-Win32 platforms, because it moves the `case SDL_SYSWMEVENT` inside the `#define WIN32` block:

https://github.com/pygame/pygame/commit/ef85a79fc3a0ffecaa305ff26ab00aa65c03eccd#diff-ebe2adb6529fbf190c471b8cb3aa9e04L464

Before this patch:

https://github.com/pygame/pygame/blob/42778d045fbf4f39d18cfbdd171b262b5aa27af9/src_c/event.c#L464-L465

After this patch:

https://github.com/pygame/pygame/blob/ef85a79fc3a0ffecaa305ff26ab00aa65c03eccd/src_c/event.c#L471-L472

All this code becomes unreachable:

https://github.com/pygame/pygame/blob/ef85a79fc3a0ffecaa305ff26ab00aa65c03eccd/src_c/event.c#L480-L501

This should be straightforward to address -- just reverse the order of those lines.

micolous:
Fixes #636.

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -524,8 +524,8 @@ dict_from_event(SDL_Event *event)
             _pg_insobj(dict, "w", PyInt_FromLong(event->resize.w));
             _pg_insobj(dict, "h", PyInt_FromLong(event->resize.h));
             break;
-#ifdef WIN32
         case SDL_SYSWMEVENT:
+#ifdef WIN32
             _pg_insobj(dict, "hwnd",
                    PyInt_FromLong((long)(event->syswm.msg->hwnd)));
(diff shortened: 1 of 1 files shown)


micolous:
I found another occurrence of the issue, which I've addressed in amended commit https://github.com/pygame/pygame/commit/2f81390172cfbccadf6f9d4eaed689066218c5d9.

dlon:
Thanks

PR #640, opened 2018-11-24
Cleaned up dict_from_event()

dlon:
Due to #637. I think this is a lot more readable.


dlon:
Does this work for you, @micolous?

micolous:
Thanks, that's much easier to read!

PR #647, opened 2018-11-28
added scale2xraw = raw version of scale2x, without dithering

nhpackard:
Title says it all.

For dynamical systems applications of pygame (for 2d spatial systems where the state is a 2d grid of values), it is convenient to have raw 2x scaling without the cool dithering that is so good for 2x scaling of sprites etc.

This is purely an addition.  Does not affect any existing functionality.

### DIFF
--- a/src_c/doc/transform_doc.h
+++ b/src_c/doc/transform_doc.h
@@ -9,6 +9,8 @@
 
 #define DOC_PYGAMETRANSFORMROTOZOOM "rotozoom(Surface, angle, scale) -> Surface\nfiltered scale and rotation"
 
+#define DOC_PYGAMETRANSFORMSCALE2XRAW "scale2x(Surface, DestSurface = None) -> Surface\nspecialized image doubler"
+
 #define DOC_PYGAMETRANSFORMSCALE2X "scale2x(Surface, DestSurface = None) -> Surface\nspecialized image doubler"
 
 #define DOC_PYGAMETRANSFORMSMOOTHSCALE "smoothscale(Surface, (width, height), DestSurface = None) -> Surface\nscale a surface to an arbitrary size smoothly"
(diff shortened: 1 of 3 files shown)


illume:
Thanks for that!

How does this differ (not dither... lol) from transform.scale?

nhpackard:
Note: your question holds for transform.scale2x as well as transform.scale2xraw.

transform.scale scales to arbitrary surface sizes (relevant code is in transform.stretch)
transform.scale2x is optimized specifically for the 2x scaling, and so is faster.
My transform.scale2xraw uses the same optimization as transform.scale2x.

illume:
Thanks for the explanation.

Would it be possible to call this function as a paramater for scale2x? The benefits would be to reduce the amount of code, and to reduce the amount of scale functions in that module (we already have quite a lot!). What do you think? Something like `transform.scale2x(src, dest=None, raw=False)`?

The {module}_doc.h files are generated files.
More details here:
http://www.pygame.org/wiki/Hacking#Generating%20docs

It would be helpful if there were unit tests for this. Probably some existing tests could be reused inside test/transform_test.py

Quadruple word score bonus points if you can provide a python implementation.
You might be interested in the `LOOP_UNROLLED4` and `REPEAT_4` macros in surface.h and used in alphablit.c. I suspect they could be used for a fairly easy extra speedup if you need it.

nhpackard:
Good idea to add arg to scale2x.  Will do.
Will also check out docs and tests.
Not sure about the extra credit... :)  
Python will never be as fast as C, right?
But I'll take a look at alphablit.c

illume:
Thanks @nhpackard 

@robertpfeiffer made some changes in https://github.com/pygame/pygame/pull/1183 so your optimized code is called with scale() when the height and width is 2x.

Issue #196, opened 2014-03-26
Font file name containing unicode error


illume:
Originally reported by: **matham (Bitbucket: [matham](https://bitbucket.org/matham), GitHub: [matham](https://github.com/matham))**

----------------------------------------

Hi,

I have an issue where when a filename contains unicode chars on windows 7, pygmae fails to read it. Following is code demonstrating the issue:

```
#!python

>>> import pygame
>>> pygame.font.init()
>>> filenameu = u'कीवी.ttf'
>>> filenameb = u'कीवी.ttf'.encode('utf-8')
>>> print os.path.exists(filenameu)
True
>>> print os.path.exists(filenameb)
False
```

Now:

```
#!python

pygame.font.Font(filenameu, 15)
```

results in:

```
#!python

Traceback (most recent call last):
  File "C:\playground.py", line 8, in <module>
    pygame.font.Font(filenameu, 15)
IOError: unable to read font file '????.ttf'
```

and:

```
#!python

pygame.font.Font(filenameb, 15)
```

results in:

```
#!python

Traceback (most recent call last):
  File "C:\playground.py", line 9, in <module>
    pygame.font.Font(filenameb, 15)
IOError: unable to read font file 'कीवी.ttf'
```

Conclusion is that pygame doesn't support unicode filenames on windows. I also tried various encodings instead of utf8, but none worked.

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/196

PR #649, opened 2018-11-29
Fix unicode paths for Font() and saved images

### LINKED ISSUE #645: fopen() fails on Windows for non-ANSI characters
I use windows 10 with a Chinese username '运城国际学校', the desktop dir will display as "C:\Users\运城国~1\Desktop" in command line or use `os.getcwd()`.

While I try to load a font file which store in this path by:

```
root = os.getcwd() # C:\Users\运城国~1\Desktop\Test
fontfile = os.path.join(root, 'test.ttf')
pygame.font.Font(fontfile, 32)
```

I got this error [OSError: unable to read font file 'C:\Users\运城国~1\Desktop\Test\test.ttf']

I can use `os.listdir()` to location this font file, and it works well if I move this file to `c:\test`. So the font file have no problem at all. It must be something wrong to location this file in `pygame.font.Font`.

Since I can load images, audios from the same UTF-8 path, except fonts. Maybe there are different ways to location file between load image / audios and load fonts?

I have to use this UTF-8 path, because I want to package my game with Pyinstaller. It will automaticly extract everything to Temp directory which is in user's home directory.

dlon:
`fopen()` doesn't work for unicode paths on Windows. As a result, saving surfaces to images and loading fonts could fail.

Fixes #645 and #196.

- [x] test unicode paths
- [x] ftfont: accept unicode paths too


illume:
Nice one :)

Here is a test...

```python
>>> import pygame
>>> pygame.font.init()
>>> from pygame.tests.test_utils import example_path                                          
>>> import os
>>> import tempfile                                                                           
>>> import shutil
>>> tmpdirname = tempfile.mkdtemp()
>>> try:
...     newfontpath = os.path.join(tmpdirname, u"运城国际学校.ttf")
...     pgfont = example_path('data/sans.ttf')  
...     shutil.copy(pgfont, newfontpath)  
...     pygame.font.Font(newfontpath, 20)  
... finally:
...     shutil.rmtree(tmpdirname)
```

I guess we can't use the python io.open for this?

dlon:
I guess it's possible: `open()` -> `PyObject_AsFileDescriptor()` -> `fdopen()`. But it's probably less overhead the way it is now.

Also, there's this in CPython:

```c
/* Open a file. Call _wfopen() on Windows, or encode the path to the filesystem
   encoding and call fopen() otherwise.

   Return the new file object on success. Raise an exception and return NULL
   on error.

   The file descriptor is created non-inheritable.

   When interrupted by a signal (open() fails with EINTR), retry the syscall,
   except if the Python signal handler raises an exception.

   Release the GIL to call _wfopen() or fopen(). The caller must hold
   the GIL. */
FILE*
_Py_fopen_obj(PyObject *path, const char *mode)
```

https://github.com/python/cpython/blob/master/Python/fileutils.c

dlon:
There's an issue with the CI: `error: can't copy 'test\fixtures\fonts\?????????.bdf': doesn't exist or not a regular file`

illume:
Hrmm. I guess that is distutils Windows code not able to handle that filename.
It works locally for you? Might be easiest to work around that bug.

Seems to pass on travis Linux, and Mac.

PR #643, opened 2018-11-27
Update build config + don't link to libpng or libjpeg for SDL2 + config SDL2 prebuilts for Windows

dlon:
The binaries (development libraries) for SDL 2, mixer, etc. work. Also updated it so that the config can be run with `python -m buildconfig -sdl2`.

Still need to:

- [x] Make it work for the remaining DLLs.
- [x] Add FreeType.
- [ ] Write instructions? https://www.pygame.org/wiki/CompileWindows
- [x] imageext: IMG_SaveJPG() and IMG_SavePNG()? Don't link directly.
- [x] Copy all required DLLs.

### Other build configs

- [ ] Update to not look link directly to libpng and libjpeg.

PR #651, opened 2018-11-29
WIP: Dlon sdl2 prebuilts win

illume:
Some code to download zip dependencies, unzip them and move them into place.

Based on PR: https://github.com/pygame/pygame/pull/643

@dlon Just commit be967080af035decbfe262b15c391aa64867cae2 id needed.


illume:
I'm not really clear what you mean by `-prebuilts`, but that sounds ok. One other option would be on Windows it could try and download them if there is not a Setup file already?

I think this is merged into the other PR now. ... closing.

PR #656, opened 2018-12-03
Handle Unicode objects properly

### LINKED ISSUE #37: pygame.mixer.music.load doesn't work with unicode symbols in path
Originally reported by: **René Dudfield (Bitbucket: [illume](https://bitbucket.org/illume), GitHub: [illume](https://github.com/illume))**

== Vasiliy, 2009-09-01 11:32:23 -0700

    
    This method doesn't work with unicode symbols in os path.
    For example:
    audiofile = u'D:/Work/Projects/Programming/iTutor/src/iTutor/dictionaries/Казахо - Русский/3.mp3'
    pygame.mixer.music.load(audiofile)
    
    return error:
    pygame.error: Couldn't read from 'D:/Work/Projects/Programming/iTutor/src/iTutor/dictionaries/Казахо - Русский/3.mp3'
    
    if i'm delete 'Казахо - Русский/' from audiofile variable then method working normally.
    

----

- Bitbucket: https://bitbucket.org/pygame/pygame/issue/37

### LINKED ISSUE #645: fopen() fails on Windows for non-ANSI characters

dlon:
Unicode objects now work for Sound(), music.load(), .image.load_extended()/load(), image.save_extended()/save(), ftfont, and font.

Changes: SDL generally expects UTF-8 strings, so I switched to using that in a lot of places. `fopen()` in Windows cannot handle UTF-8, so I added workarounds for that.

I had to disable `test_save_bad_filename()` and `test_load_bad_filename()` for now. SDL simply strips out "bad" characters instead of returning an error.

Close #37 
Close #645 

Related PR: #649

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -10,7 +10,8 @@ environment:
     PYGAME_USE_PREBUILT: "1"
     SDL_VIDEODRIVER: "dummy"
     SDL_AUDIODRIVER: "disk"
- 
+    PYTHONIOENCODING: "UTF-8"
+
   matrix:
     # - PYTHON: "pypy.exe"
     #   PYTHON_VERSION: "2.7.13"
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -188,6 +188,44 @@ pg_SetDefaultWindowSurface(PyObject *);
 #endif /* IS_SDLv2 */
 
 
+#ifdef WIN32
+static FILE*
+pg_Fopen(const char *filename, const char *mode) {
+    FILE *fp = NULL;
+    static wchar_t modebuf[24];
+    size_t nameInputSize = strlen(filename) + 1;
+    size_t namebuf_chars = MultiByteToWideChar(CP_UTF8, 0, filename, nameInputSize, 0, 0);
+    wchar_t* namebuf;
+    if (namebuf_chars == 0)
+        return NULL;
+    namebuf = malloc(namebuf_chars * sizeof(wchar_t));
+
+    if (!namebuf)
+        goto leave;
+    if (MultiByteToWideChar(CP_UTF8, 0, filename, nameInputSize,
+                            namebuf, namebuf_chars) == 0) {
+        PyErr_Format(PyExc_OSError, "error code %u", GetLastError());
+        goto leave;
+    }
+    if (MultiByteToWideChar(CP_UTF8, 0, mode, -1,
+                            modebuf, sizeof(modebuf) / sizeof(wchar_t)) == 0) {
+        PyErr_Format(PyExc_OSError, "error code %u", GetLastError());
+        goto leave;
+    }
+
+    fp = _wfopen(namebuf, modebuf);
+
+leave:
+    free(namebuf);
+    return fp;
+}
+#else /* !WIN32 */
+static FILE*
+pg_Fopen(const char *filename, const char *mode) {
+    return fopen(filename, mode);
+}
+#endif /* !WIN32 */
+
 static int
 pg_CheckSDLVersions(void) /*compare compiled to linked*/
 {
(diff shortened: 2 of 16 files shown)


.appveyor.yml
```diff
@@ -1,118 +1,119 @@
...
-  - "mkdir empty_folder"
-  - "cd empty_folder"
-  - "%PYTHON% -m pygame.tests --exclude opengl"
-
-  # Move back to the project folder
-  - "cd .."
-
-artifacts:
-  - path: 'dist\*'
+environment:
```
illume: Hrmm. What changed in this file? It seems all the lines were changed for some reason, but I can't see what the changes were.

src_c/base.c
```diff
@@ -188,6 +188,44 @@ pg_SetDefaultWindowSurface(PyObject *);
...
+    fp = _wfopen(namebuf, modebuf);
+
+leave:
+    free(namebuf);
+    return fp;
+}
+#else /* !WIN32 */
+static FILE*
+pg_Fopen(const char *filename, const char *mode) {
+    return fopen(filename, mode);
```
illume: I wonder what we have to do about the GIL here? Does it need to be wrapped in Py_BEGIN_ALLOW_THREADS; ?

.appveyor.yml
e1000: the file in **master** has `\r`n` at end of each line, maybe that?

illume:
Very good :)

I left two questions for bits I didn't understand.

src_c/base.c
dlon: If SDL is used for error handling, then the caller can safely use Py_BEGIN_ALLOW_THREADS. It shouldn't matter if the GIL is held by the thread, but it doesn't have to be held?
Maybe imageext should be rewritten to use `pgRWopsFromObjectThreaded()` or `SDL_RWFromFile()` at some point. Then `pg_Fopen()` won't be needed anymore.

.appveyor.yml
dlon: I only added `PYTHONIOENCODING: "UTF-8"`. AppVeyor reroutes stdout, and it caused some issue. But I'm not sure if it's needed anymore.

src_c/imageext.c
```diff
@@ -775,15 +775,9 @@ image_save_ext(PyObject *self, PyObject *arg)
               (name[namelen - 2] == 'p' || name[namelen - 2] == 'P') &&
               (name[namelen - 3] == 'j' || name[namelen - 3] == 'J')))) {
 #ifdef JPEGLIB_H
-            /* jpg save functions seem *NOT* thread safe at least on windows.
-             */
-            /*
-            Py_BEGIN_ALLOW_THREADS;
-            */
+            Py_BEGIN_ALLOW_THREADS; /* safe as long as Python is never used */
```
illume: Quite some years ago (10yrs) jpeg saving multiple jpegs at once had problems in threads when releasing the GIL. I never investigated why. It could have been a libjpeg thing at the time... not sure.

.appveyor.yml
illume: Ah, ok. I guess the editor (on windows) changed all the line endings.

src_c/imageext.c
dlon: Holding the GIL blocks multiple threads from using SaveJPEG() simultaneously, so I guess libjpeg wasn't thread safe.

dlon:
I guess it can be merged now?

PR #658, opened 2018-12-03
Change surface docs to use defaults instead of repeated calls

Bottersnike:
When IDEs like IntelliJ IDEA attempt for figure out the function signatures for C extensions, they check the doc strings. Especially IDEA will stop when it finds the first one in a `__doc__`. This means that `.convert()` and `.convert_alpha()` result in a lot of `Parameter 'Surface' unfilled more... (Ctrl+F1)`s scattered around the code even though it's an optional parameter. By switching to `=None`, it solves this problem.


illume:
Thanks for that. The `{module}_doc.h` files are generated files.
More details here:
http://www.pygame.org/wiki/Hacking#Generating%20docs

(a side note... Probably all the parameters should be using type definitions and the sphinx supported `param:` scheme.)

markph0204:
@Bottersnike can you review my new PR incorporating suggested change by illume?

- new change will remain intact throughout doc generation following
- [ ] please verify it provides your IDE with the proper detailed information needed

Following you/we can take on other RST files.

PR for your review and validation:  https://github.com/pygame/pygame/pull/809

PR #663, opened 2018-12-04
Compile for PyPy 3 and PyPy fixes

dlon:
I also enabled FreeType (but had to avoid using the module state). Rect seems to be broken.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -139,12 +139,12 @@ Dependencies
      numpy package for its multidimensional numeric arrays.
      Dependency versions:
 
-     * Python >= 2.7
+     * Python >= 2.7 or PyPy >= 6.0.0 (and pypy3)
      * SDL >= 1.2.15
      * SDL_mixer >= 1.2.13
-     * SDL_image (version?)
-     * SDL_tff (version ?)
-     * SDL_gfx (optional, version?)
+     * SDL_image >= 1.2.12
+     * SDL_tff >= 2.0.11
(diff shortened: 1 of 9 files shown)


illume:
Very nice :)

Hi. I tried it with pypy 5.10.0 from homebrew mac, and got this error.
```
    import pygame.font
ImportError: unable to load extension module '/usr/local/Cellar/pypy/5.10.0_1/libexec/site-packages/pygame/font.pypy-41.so': dlopen(/usr/local/Cellar/pypy/5.10.0_1/libexec/site-packages/pygame/font.pypy-41.so, 6): Symbol not found: _PyUnicode_AsUTF16String
```

However, because 5.10.0 has quite a few errors in CPython API extension support it's ok to require pypy 6.0+ If you agree, let's merge this in? (also added supported pypy version to README).

pypy - 1 failure, 8 errors in the rect module.

On pypy3 I get these 
```
loading test.mouse_test
loading test.overlay_test
loading test.pixelarray_test
pypy3(97323,0x7fff898c4380) malloc: *** error for object 0x7fc4de0bb3c0: incorrect checksum for freed object - object was probably modified after being freed.
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

I skipped pixelarray_test:PixelArrayTypeTest.test_contains and the test suite runs with the tests again same as pypy (failures=1, errors=8).

dlon:
I agree with that.

PR #672, opened 2018-12-12
SDL2: Partially implement display.list_modes()

dlon:
The flags are ignored.


illume:
Nice one.

I added a very basic test, and made the function use keyword arguments.

I'm not sure how to support the flags argument with SDL2.

dlon:
I think the depth doesn't include an alpha channel right now, since it uses the depth of the display.

illume:
I guess this is ready to be merged in now?

Not sure I understand your comment about the depth not containing an alpha channel. Probably no displays have an alpha channel, because they are the final destination.

dlon:
You're right, but I think `mode_ok` and `list_modes` are referring to the depth of the surface in SDL1:

```
>>> pygame.display.mode_ok((320,240), pygame.FULLSCREEN, 24)
32
```

The SDL2 version returns 24 instead.

illume:
I guess returning 24 for SDL2 here is ok. Because it's supposed to return the best match.
https://www.pygame.org/docs/ref/display.html#pygame.display.mode_ok

PR #675, opened 2018-12-13
display.get_num_displays()

dlon:
Not sure how to document this since it's dependent on SDL2.


illume:
For forwards compatibility people would have to:
```python
if getattr(display, "get_num_displays"):
```

A SDL 1 using get_num_displays could probably just return 1?

illume:
Very good :) Multiple display support in SDL2 is a really nice enhancement.

PR #677, opened 2018-12-14
joystick.get_name(): Use locale encoding. Fix #674

### LINKED ISSUE #674: joystick.get_name() can give an exception "invalid utf-8 character"
Happens with "Logitech® G27 Shifter" and others that have ® in the name.  Sample code:
```
    pygame.init()
    self.num_controllers = pygame.joystick.get_count()
    for j in range(self.num_controllers):
      _j = pygame.joystick.Joystick(j)
      _j.get_name()
```
Gives UnicodeError

pygame 1.9.4  Python 3.6 on Windows 10 64bit

dlon:
SDL1 doesn't use UTF-8 here. Closes #674.

I'm not able to test this, so I'm not sure if it solves the problem.

### DIFF
--- a/src_c/joystick.c
+++ b/src_c/joystick.c
@@ -168,7 +168,7 @@ joy_get_name(PyObject *self)
     int joy_id = pgJoystick_AsID(self);
     JOYSTICK_INIT_CHECK();
 #if IS_SDLv1
-    return Text_FromUTF8(SDL_JoystickName(joy_id));
+    return Text_FromLocale(SDL_JoystickName(joy_id));
 #else  /* IS_SDLv2 */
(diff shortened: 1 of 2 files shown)


illume:
Sounds like this fix works? We can merge it in? cc @TonyWhitley @dlon  

If it's not easy to test this, I can test it by putting the failing text in there instead of "SDL_JoystickName(joy_id)".

TonyWhitley:
Apologies for the slow reply, yes it fixed the issue (but just kicked it
down the road into a later call to aother library).

Thank you,
-- 
Tony

illume:
Thanks Tony :)

Issue #668, opened 2018-12-06
Python type stubs for pygame


rchen152:
I'd like to contribute some type stubs for pygame to https://github.com/python/typeshed, a repository of Python type information that is used by type checkers. Per https://github.com/python/typeshed/blob/master/CONTRIBUTING.md#adding-a-new-library, I'm first opening this issue to request permission from the maintainers of pygame.

illume:
Great :) Thank you.

Some thoughts and questions below...

---

It would be preferable to have them inside of pygame, because they would also be useful for documentation and maintenance would be easier. Does this sound ok? I wonder if there are some advantages to adding them into the typeshed?

We are a python2, and python3 single code base, with a mixture of python/C/Cython - so type stubs would maybe make sense (rather than specifying them inline).

The files would probably live in src_py/ So for example, src_py/display.pyi ?

The docs currently have a form of argument and type specification which is not machine checked, and not entirely consistent.

Maybe you know of a good workflow which uses the types in documentation? Also, perhaps [sphinx-autodoc-typehints](https://github.com/agronholm/sphinx-autodoc-typehints) can be used, or modified to allow using the type definitions in documentation. It's also possible to defer the task of using the types in documentation to another time, if there is no known good workflow for this.

Perhaps we need to modify some sort of sphinx autodoc tool for this. We already generate .h files containing documentation for our C modules from the restructured text docs.

The test suite doesn't have 100% coverage, but may be useful to generate type hints by recording. There are also many examples/ both included in pygame and elsewhere. The chimp.py, and aliens.py examples are probably a good place to start for deciding which modules to type first. Core parts are things like Rect, Surface, display.set_mode, sprite, image.load, and draw. So these would perhaps be good for the first modules to do? Would it be good to do one or two parts at a time completely?

We use appveyor, and travisci for continuous integration. The buildconfig folder is where we are trying to keep as much config as possible if any needs to be added.

Links to our developer guide and such can be found here: https://www.pygame.org/wiki/Contribute

illume:
pep 561 is relevant: https://www.python.org/dev/peps/pep-0561/

illume:
numpy is making a `numpy-stubs` package available.
https://github.com/numpy/numpy-stubs

illume:
I asked for some feedback on the typing-sig mailing list.
https://mail.python.org/archives/list/typing-sig@python.org/thread/QPJC7PM6V46GTM24DEXY2BTZSEH623X6/

[...]

PR #678, opened 2018-12-15
Add some type stubs in buildconfig/pygame-stubs.

rchen152:
Per https://github.com/pygame/pygame/issues/668, adds complete
stubs for bufferproxy, color, event, joystick, rect, and surface,
and an incomplete stub for __init__. The stubs are based on
https://www.pygame.org/docs/ as of March 2018.

The stubs are flake8-clean, using the flake8 settings from
https://github.com/python/typeshed which I've copied over, and
parse correctly under both `mypy` and `pytype-single --parse-pyi`
(modulo import errors, since the immediate directory they're in is
not called pygame).


illume:
Thanks :)

PR #680, opened 2018-12-16
SDL2: Select which screen to appear on with set_mode(display=0)

dlon:
This lets you to select a display to use with `set_mode()`. Keyword arguments were also added.

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -104,7 +104,7 @@ required).
 .. function:: set_mode
 
    | :sl:`Initialize a window or screen for display`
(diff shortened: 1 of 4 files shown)


illume:
Works for me :)

illume:
As a side note...
I wonder if instead of "resolution", we perhaps use "size" instead?
Because size is shorter than resolution, and the term is already used in Rect and Surface.

I know it's already in the documentation as "resolution", but no one would be using that in code yet because it's not usable with keywords.

illume:
I added a basic test for set_mode keyword arguments.

Issue #605, opened 2018-11-05
Use of SDL_Renderer


dlon:
Since surfaces are software rendered in SDL2, should a new module be created for SDL_Renderer and textures? One thing that SDL_Renderer lacks (unlike SDL_gpu) is support for shaders, but that could perhaps be added to pygame.

dlon:
I tried the suggestion in SDL's wiki (https://wiki.libsdl.org/MigrationGuide#If_your_game_wants_to_do_both), ie updating a texture using the screen surface. This is backward compatible but seems to be slower than pure software rendering.

illume:
There is also this overview of how the renderer, textures, and window fits together: https://github.com/pygame/pygame/wiki/SDL-2-Overview

dlon:
That seems straightforward. The typical usage might be this:

```python3
window = pygame.Window(width, height)
renderer = pygame.render.Renderer(window)
tex = pygame.render.Texture(renderer, pygame.image.load('image.png'))

while True:
	renderer.clear()
	renderer.copy(tex)
	renderer.present()
```

And the new types:

```python3
class Window:
	__init__(width, height, flags=0)

	# low-priority stuff for surface-based windows:
	get_surface() : Surface
	update_surface(rects=[])

class Renderer:
	__init__(window, driver=some_default, flags=ACCELERATED)

	clear()
	copy(texture, srcrect=None, destrect=None)
	present()
	draw_line()
	fill_rect(rect=None)
	color = (123, 123, 123)
	blend_mode = BLENDMODE_BLEND
	...

class Texture:
	__init__(renderer, surface)
	__init__(renderer, file) # maybe for convenience
	__init__(renderer, w, h, format=0, flags=0)
	lock()
	unlock()
	...
```

A new module (maybe called "render") seems sensible since most of pygame.display could probably be deprecated. "Window" could be separate since it's not dependent on renderers or textures, whereas textures and renderers are dependent on each other. You might just want create a window and use Vulkan or something directly.

The equivalent to the old approach would be this:

```python3
window = pygame.Window(width, height, flags)
display_surface = window.get_surface() # instead of display.set_mode()

while True:
	window.update_surface() # instead of display.flip()
```

illume:
Very cool :)

I'm not sure if making a top level 'pygame.window', or 'pygame.renderer' is needed. Because there is probably nothing else to go in there.

They probably should just live in the top?

pygame.Window, pygame.Texture, pygame.Renderer

[...]

PR #686, opened 2018-12-19
Port Renderer, Window, and Texture to Cython (from ctypes)

dlon:
The earlier ctypes version: #608.

An example exists in `python -m example.video`.

Related: #684, #605.

### To do

- [x] Move constants from class scope to module scope for consistency.
- [ ] Tests.
- [ ] More features.
- [x] Fix `src_c/_sdl2\audio.pxd: cannot find cimported module 'sdl2'` (even though it works anyway).
- [ ] Allow window position to be set like `win.position.x = 10`, not just as a tuple.


illume:
Awesome :)

illume:
Seems the mac builders on Travis stalled building the boost dependency from source. I might have to look at changing some stuff there to reduce the build time. It seems homebrew is having much less stuff pre-compiled than before.

illume:
Instead of the FLAG CONSTANTS, what do you think of using key word arguments instead?

So you could do this:
```python
renderer = pg.video.Renderer(software=1, accelerated=1)
```

Instead of this:
```python
renderer = pg.video.Renderer(flags=pg.video.RENDERER_SOFTWARE | pg.video.RENDERER_ACCELERATED)
```

dlon:
Agreed. Maybe just "accelerated"? Software and accelerated should be mutually exclusive.

dlon:
Should `Window` also use keyword arguments?

E.g.
```python3
args = {
    size=(320, 240),
    position=WINDOWPOS_UNDEFINED,
    hidden=False,
    maximized=True,
}
Window(**args)
```

src_c/_sdl2/video.pyx
```diff
@@ -0,0 +1,302 @@
...
+cdef class Renderer:
+    def __init__(self, Window window, int index=-1,
+                 int accelerated=-1, bint vsync=False,
+                 bint target_texture=False):
+        """ Create a 2D rendering context for a window.
+
+        :param window Window: where rendering is displayed.
+        :param index int: index of rendering driver to initialize,
+                          or -1 to init the first supporting requested options.
+        :param accelerated int: if 1, the renderer uses hardware acceleration.
```
illume: I like the accelerated handling of flags! (wasn't sure about it at first).
```python
Renderer(accelerated=0)
Renderer(accelerated=1)
Renderer()
```

Reading those I can tell what accelerated=0 means, and accelerated=1 means. The default of -1 probably won't be specified usually, so I don't think it matters that it is not immediately understandable.

Nice!

illume:
> Also, many of the "X11 only" arguments actually work on Windows too.

That's good. I guess no one has updated the wiki yet.

src_c/_sdl2/video.pyx
```diff
@@ -0,0 +1,415 @@
...
+    def __init__(self, title='pygame',
+                 size=DEFAULT_SIZE,
+                 position=WINDOWPOS_UNDEFINED,
+                 int fullscreen=0, **kwargs):
+        """ Create a window with the specified position, dimensions, and flags.
+
+        :param title str: the title of the window, in UTF-8 encoding
+        :param size tuple: the size of the window, in screen coordinates (width, height)
+        :param position: a tuple specifying the window position, WINDOWPOS_CENTERED, or WINDOWPOS_UNDEFINED.
+        :param fullscreen int: 0: windowed mode
```
illume: Instead of Window(fullscreen=2), how about Window(fullscreen_desktop=True) ?
Window(fullscreen=1) and Window(fullscreen=1) would still exist.

dlon: It seems a bit confusing to have separate parameters for mutually exclusive options. Maybe fullscreen=pg.FULLSCREEN, fullscreen=pg.FULLSCREEN_DESKTOP, fullscreen=0?

illume:
Hi,

I wonder why this was merged in?

PR #688, opened 2018-12-21
SDL2 messagebox()

dlon:
This implements [SDL_ShowMessageBox](https://wiki.libsdl.org/SDL_ShowMessageBox) in the _sdl2 package.

### Function signature
```cython
def messagebox(message, title="",
               type=MESSAGEBOX_INFORMATION,
               Window window=None, buttons=None,
               return_button=-1,
               escape_button=-1)
```

### Example usage

This causes the program to quit if the user presses escape or "no":

```python3
from pygame._sdl2 import (
    messagebox,
    MESSAGEBOX_WARNING
)

import random
answer = messagebox("Continue?", "Title!", type=MESSAGEBOX_WARNING,
                    buttons=('Yes', 'No'),
                    return_button=0, escape_button=1)
if answer == 1:
    import sys
    sys.exit(0)
```

The default button (if `buttons` is empty or evaluates to false) is "OK".

### DIFF
--- a/examples/video.py
+++ b/examples/video.py
@@ -14,7 +14,6 @@ from pygame._sdl2 import (
     Renderer,
     get_drivers,
     messagebox,
-    MESSAGEBOX_WARNING
 )
 
 def load_img(file):
@@ -27,7 +26,7 @@ for driver in get_drivers():
     print(driver)
 
 import random
-answer = messagebox("I will open two windows! Continue?", "Hello!", type=MESSAGEBOX_WARNING,
+answer = messagebox("I will open two windows! Continue?", "Hello!", info=True,
                     buttons=('Yes', 'No', 'Chance'),
                     return_button=0, escape_button=1)
 if answer == 1 or (answer == 2 and random.random() < .5):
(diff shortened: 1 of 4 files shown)


illume:
This is very useful functionality. There's a very old implementation of something like this here: https://github.com/pygame/solarwolf/blob/master/solarwolf/errorbox.py

I'm not sure if we should make these arguments instead of using the flags?
```python
warn=True, error=False, info=False
```

return_button and error_button are a bit confusing, but I'm not sure of a better solution.

Is this really the most common way people will use it?
```python
if messagebox("Continue?", "Title!"):
    do_something()
```

In that case, then the defaults could be `return_button=0` and `error_button=1`, and `buttons=('Yes', 'No')`?

The other implementation from solarwolf writes info to sdtout if the message box fails. This is useful on platforms that don't actually have windows (embedded systems for example). (except it probably should write to stderr). Also, the other implementation falls back to using SDL itself to show the error if none of the other implementations are available. Probably this is stuff to consider later.

dlon:
It seems a bit confusing to have separate parameters for mutually exclusive options. It makes `warn=True, info=True, error=True` look valid?

dlon:
The default case should probably just be "OK", as it is in SDL? That's probably the most common usage.

illume:
Ok, OK sounds good. 👍

Issue #682, opened 2018-12-18
SDL2: Mouse wheel events


husano896:
https://wiki.libsdl.org/MigrationGuide#Input
https://wiki.libsdl.org/SDL_MouseWheelEvent
https://wiki.libsdl.org/SDL_MouseButtonEvent
> Now, for mouse input.
The first change, simply enough, is that the mousewheel is no longer a button. This was a mistake of history, and we've corrected it in SDL 2.0. Look for SDL_MOUSEWHEEL events. We support both vertical and horizontal wheels, and some platforms can treat two-finger scrolling on a trackpad as wheel input, too. You will no longer receive SDL_BUTTONDOWN events for mouse wheels, and buttons 4 and 5 are > real mouse buttons now.

Should we follow the new behavior (SDL_MOUSEWHEEL / pygame.MOUSEWHEEL)

or keep old behavior by creating MOUSEBUTTONDOWN events?

    #if IS_SDLv2

        case SDL_MOUSEWHEEL:
            _pg_insobj(dict, "x", PyInt_FromSsize_t(event->wheel.x));
            _pg_insobj(dict, "y", PyInt_FromSsize_t(event->wheel.y));
            break;

    #endif

This does break some back compatibility for old games.

illume:
Hi,

thanks for writing up the issue. I look forward to our MOUSEWHEEL future. 🐭 🐁 🖱 

The general plan for SDL2/pygame 2 is:
- to keep backwards compatibility with existing pygame API as much as possible. So old programs work when running under SDL2.
- to keep forwards compatibility for new SDL2 only features. Such that you can write a pygame program that takes advantage of SDL2 features but your app won't break when using a SDL1 pygame.

So for this MOUSEWHEEL one:
- MOUSEWHEEL constant should be -1 in pygame+SDL1.
- we should generate the appropriate events when we get a SDL_MOUSEWHEEL, so old pygame apps don't break.
- it should be documented how MOUSEWHEEL behaves in different programs.

cheese,

🧀 🐁

PR #691, opened 2018-12-24
SDL2 MOUSEWHEEL event, and emulate SDL1 wheel events

husano896:
Added MOUSEWHEEL event for #682 

access the event by **event.pos** and **event.direction**

I'm not good at writing document, so I wrote a code example instead :/

<details><summary>Code Example</summary>

        if event.type == MOUSEWHEEL:
            mov_x, mov_y = event.pos
            if (mov_x > 0): #scroll right
                pass
            elif (mov_x < 0): #scroll left
                pass
            if (mov_y > 0): #scroll up
                pass
            elif (mov_y <0): #scroll down
                pass
            
			
</details>

<details><summary>Testbench</summary>

    import pygame
    from pygame.locals import *

    pygame.init()

    screen = pygame.display.set_mode((640,480))
    clock = pygame.time.Clock()
    def main():
        while True:
                if event.type == QUIT:
                    pygame.quit()
                    return
                elif event.type == MOUSEWHEEL:
                    print("Wheel event")
                    print(event)
                elif event.type == MOUSEBUTTONDOWN:
                    print("Button down event")
                    print(event)
            clock.tick(60)
    main()
</details>

<details><summary>Outputs</summary>

	pygame 1.9.5.dev0
	Button down event
	<Event(1025-MouseButtonDown {'pos': (383, 280), 'button': 4})>
	Wheel event
	<Event(1027-MouseWheel {'pos': (0, 1), 'direction': 0})>
	Button down event
	<Event(1025-MouseButtonDown {'pos': (384, 278), 'button': 5})>
	Wheel event
	<Event(1027-MouseWheel {'pos': (0, -1), 'direction': 0})>
</details>

🧀 🖱🐁


src_c/event.c
```diff
@@ -173,7 +173,36 @@ pg_event_filter(void *_, SDL_Event *event)
...
+        
+        /* 4/5 for roll up/down
+		 TODO: Wheel for x axis?
+		if (event->wheel.x > 0)
+            newevent.button.button = 4;
+        else if (event->wheel.x < 0)
+            newevent.button.button = 5;
+        */
+        if (event->wheel.y > 0)
+            newevent.button.button = 4;
```
dlon: > You will no longer receive SDL_BUTTONDOWN events for mouse wheels, and buttons 4 and 5 are real mouse buttons now.

https://wiki.libsdl.org/MigrationGuide

How might one disambiguate these buttons from scrolling? A boolean "wheel" attribute in the event dict?

illume:
Hello!

For repr(event) and str(event) there needs to be something like in this PR:
https://github.com/pygame/pygame/pull/604/files#diff-ebe2adb6529fbf190c471b8cb3aa9e04

```python
python -m pygame.examples.eventlist
```

husano896:
just to make sure,

so wheel in dict is to distinguish whell x/y?

dlon:
Hi. No, it's to distinguish between button presses and any wheel. Buttons 4 and 5 will be recognized as wheels in old games, so maybe it's not the best solution, though

husano896:
Understood, thanks for explaining!

husano896:
Okay, so my way is to use a non-existing button in event.button,

and if a non-existing button detected(event->button.button > 96), convert it back.

not a best solution but a working one I guess.

src_c/event.c
```diff
@@ -174,7 +174,34 @@ pg_event_filter(void *_, SDL_Event *event)
...
+		
+        newevent.button.state = SDL_PRESSED;
+        newevent.button.clicks = 1;
+        
+        if (event->wheel.y != 0) {
+            newevent.button.button = (event->wheel.y > 0) ? 4 : 5;
+        }
+        else if (event->wheel.x != 0) {
+            newevent.button.button = (event->wheel.x > 0) ? 4 : 5;
+        }
```
dlon: I get a bunch of mouse button events with button == 80 when I scroll. Apparently pos can be (0, 0), at least on my touchpad. So it shouldn't generate a mouse button event in that case.

src_c/event.c
```diff
@@ -573,6 +603,12 @@ dict_from_event(SDL_Event *event)
             _pg_insobj(dict, "pinched", PyFloat_FromDouble(event->mgesture.dDist));
             _pg_insobj(dict, "num_fingers", PyInt_FromLong(event->mgesture.numFingers));
             break;
+        case SDL_MOUSEWHEEL:
+            /* https://wiki.libsdl.org/SDL_MouseWheelEvent */
+            obj = Py_BuildValue("(ii)", event->wheel.x, event->wheel.y);
+            _pg_insobj(dict, "pos", obj);
```
dlon: `pos` is actually the amount scrolled, not the position. It should probably have a different name

src_c/event.c
```diff
@@ -573,6 +603,12 @@ dict_from_event(SDL_Event *event)
...
+            _pg_insobj(dict, "direction", PyInt_FromLong(event->wheel.direction));
```
dlon: Should constants for this be added, ie MOUSEWHEEL_NORMAL and MOUSEWHEEL_FLIPPED?

Cf. https://wiki.libsdl.org/SDL_MouseWheelEvent

dlon: Should the `which` field be added?

> which may be SDL_TOUCH_MOUSEID, for events that were generated by a touch input device, and not a real mouse. You might want to ignore such events, if your application already handles SDL_TouchFingerEvent.

If not, I think those events must be filtered, or else they will be received in pygame anyhow.

src_c/event.c
```diff
@@ -518,7 +547,8 @@ dict_from_event(SDL_Event *event)
         case SDL_MOUSEBUTTONUP:
             obj = Py_BuildValue("(ii)", event->button.x, event->button.y);
             _pg_insobj(dict, "pos", obj);
-            _pg_insobj(dict, "button", PyInt_FromLong(event->button.button));
+            _pg_insobj(dict, "button", PyInt_FromLong((event->button.button > 96) ? event->button.button - 96 : event->button.button));
```
dlon: For SDL1, this should be true if `button` is 4 or 5

src_c/event.c
husano896: > I wonder if this might a better idea for the button events: Add 2 if button id >= 4 (and < 96)? > So the buttons become 6, 7... instead of 4, 5...
> SDL_BUTTON_LEFT
> SDL_BUTTON_MIDDLE
> SDL_BUTTON_RIGHT
> SDL_BUTTON_X1
> SDL_BUTTON_X2
> I suggest we add these constants and also add constants for BUTTON_WHEELUP and BUTTON_WHEELDOWN (also for SDL 1).

Using 6/7 instead of 4/5 still break compatibility, but that's needed I think.

BUTTON_WHEELUP / BUTTON_WHEELDOWN should be 4/5 in SDL1 and 6/7 in SDL2 in this way,

but some apps are still using if event.button == 4 or 5, so taking changes in old apps are necessary.

husano896: > I get a bunch of mouse button events with button == 80 when I scroll. Apparently pos can be (0, 0), at least on my touchpad. So it shouldn't generate a mouse button event in that case.

if it generates button == 80 and button == 4/5 at the same time, 

then the button==80 event can be safely filtered I think?

dlon:
I pushed a change that keeps the same values for the constants in SDL2 as in SDL1

husano896:
If that case, we should update documentation in [pygame.mouse](https://www.pygame.org/docs/ref/mouse.html)

> The button will be set to 4 when the wheel is rolled up, and to button 5 when the wheel is rolled down.

by using pygame.BUTTON_WHEELUP & pygame.BUTTON_WHEELDOWN?

(Although some projects are using 4/5)

dlon:
This can be merged, I think. The numbers are the same now. But I agree that the docs should use constants instead of literals.

Issue #687, opened 2018-12-21
mac travis build speedups


illume:
Mac travis builds are timing out.
https://travis-ci.org/pygame/pygame/jobs/470762914

Try and cache the build from source homebrew results using these techniques:
- https://stackoverflow.com/questions/39930171/cache-brew-builds-with-travis-ci
- https://github.com/droe/sslsplit/blob/1a8894175ebef042cb1cacb4e3fe91e63a81f310/.travis.yml

pradyunsg:
Is this issue resolved?

PR #692, opened 2018-12-24
travis+macOS: Save local bottles

dlon:
Partly implements #687.

It will still stall while building boost.

### To do

- [ ] In `check_local_bottles`, check if the local bottles are still appropriate (compare versions).
- [ ] `before_cache`: First backup local bottles (hard links?), run `brew cleanup`, and then restore deleted bottles.

### DIFF
--- a/buildconfig/ci/travis/.travis_osx_before_install.sh
+++ b/buildconfig/ci/travis/.travis_osx_before_install.sh
@@ -83,15 +83,103 @@ function retry {
 }
 
 function install_or_upgrade {
-	set +e
-    if brew ls --versions "$1" >/dev/null; then
-        echo "package already installed: $@"
+  local deps=""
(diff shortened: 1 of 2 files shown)


buildconfig/ci/travis/.travis_osx_before_install.sh
```diff
@@ -83,15 +83,105 @@ function retry {
...
+    echo "$1 is already installed and up to date."
+  else
+    if (brew outdated | grep "$1" >/dev/null); then
+      echo "$1 is installed but outdated."
+      if (brew info "$1" | grep "(bottled)" >/dev/null); then
+        echo "$1: Found bottle."
+        retry brew upgrade "$1"
+        return 0
+      else
+        brew uninstall --ignore-dependencies "$1"
```
dlon: I'm unsure about this especially. Can't you upgrade and create a bottle at the same time?

PR #695, opened 2018-12-26
Travis: MacOS upload script

dlon:
This is useful for creating bottles that we could add to https://github.com/pygame/homebrew-portmidi in order to reduce test times, if they're not available at https://bintray.com/homebrew. For example, older Python versions or bottles for old OS X versions. Or they could be hosted on bintray (https://bintray.com/signup/oss)?

See https://github.com/pygame/homebrew-portmidi/pull/1.

Close if you disagree, @illume.

### DIFF
--- a/buildconfig/ci/travis/.travis_osx_upload_bottles.sh
+++ b/buildconfig/ci/travis/.travis_osx_upload_bottles.sh
@@ -1,0 +1,168 @@
+set -e
+
+# Work around https://github.com/travis-ci/travis-ci/issues/8703 :-@
+# Travis overrides cd to do something with Ruby. Revert to the default.
+unset -f cd
+shell_session_update() { :; }
+
+echo -en 'travis_fold:start:brew.update\\r'
+echo "Updating Homebrew listings..."
+brew update
+echo -en 'travis_fold:end:brew.update\\r'
+export HOMEBREW_NO_AUTO_UPDATE=1
+
+brew uninstall --force --ignore-dependencies pkg-config
+brew install pkg-config
+
+if [[ ${BUILD_UNIVERSAL} == "1" ]]; then
+  UNIVERSAL_FLAG='--universal'
+  echo "Using --universal option for builds"
+else
+  UNIVERSAL_FLAG=''
+  echo "Not using --universal option for builds"
+fi
+
+
+function fail {
+  echo $1 >&2
+  exit 1
+}
+
+function retry {
+  local n=1
+  local max=5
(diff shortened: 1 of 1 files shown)


buildconfig/ci/travis/.travis_osx_upload_bottles.sh
```diff
@@ -0,0 +1,168 @@
...
+  #local cachefile=$(brew --cache $1)
+
+  mkdir -p Formula
+  brew cat $1 > "Formula/$1.rb"
+  mkdir -p "bottles"
+  mv "$bottlefile" "bottles/"
+  #tar cfzv "$1-formula.tar.gz" "$1.rb" "$bottlefile"
+
+  # upload the package
+  #curl -T "$1-formula.tar.gz" "https://transfer.sh/$archive"
```
illume: Could add before the curl upload so we can confirm the file built is the one downloaded:
```bash
shasum -a 256 "$1-formula.tar.gz"
```

buildconfig/ci/travis/.travis_osx_upload_bottles.sh
```diff
@@ -0,0 +1,168 @@
...
+done
+
+# archive all packages upload the archive
+echo "Creating brew-packages.tar.gz..."
+tar cfzv "brew-packages.tar.gz" "Formula" "bottles"
+
+set -e
+
+echo "Uploading brew-packages.tar.gz..."
+curl -vs --upload-file "./brew-packages.tar.gz" "https://transfer.sh/" &> /dev/stdout
```
illume: Could add before the curl upload so we can confirm the file built is the one downloaded:
```bash
shasum -a 256 "./brew-packages.tar.gz"
```

illume:
This is great!
I added the 'pygame' team with write access to pygame/homebrew-portmidi repo.
So you should be able to write there now (sorry about that).

Or should I start a shared pygame bintray account?

illume:
Looking at this more...

Is the plan to use that 'homebrew-portmidi' repo to host the bottles as releases? This seems to make sense.

How would building bottles, and saving them here be done in the future?

Would moving some of the bottle building code into the 'homebrew-portmidi' repo make things easier? The reason is then we can more easily have that repo upload the bottle files as github releases. See the "deploy" config in here as an example of uploading files into github releases: https://github.com/viblo/pymunk/pull/149/files

dlon:
I guess I'll move the entire thing there now. It's not dependent on pygame.

PR #700, opened 2018-12-30
Build commits once on travis, and appveyor.

illume:
So that commits are only built for a PR, master branch, or version number tags.

Before both travis and appveyor were building for both the branch and PR.


illume:
By the way: very good work on fixing the mac builds on travis. I know how annoying and hard that can be.

PR #703, opened 2018-12-30
Clean up C API for rwobject.

dlon:
Removes `pgRWopsFromFileObjectThreaded`, `pgRWopsFromObjectThreaded`, and `pgRWopsCheckObjectThreaded`. Instead, always acquire the GIL if threading is supported (and use `pgRWopsFromFileObject`, `pgRWopsFromObject`, and `pgRWopsCheckObject`). It is safe:

    Ensure that the current thread is ready to call the Python C API regardless of the current state of Python, or of the global interpreter lock. This may be called as many times as desired by a thread as long as each call is matched with a call to PyGILState_Release().

https://docs.python.org/2/c-api/init.html#c.PyGILState_Ensure


illume:
Cool. Seems like a nice cleanup. 

I worry that we don't have tests which use threads, so I opened an issue here: https://github.com/pygame/pygame/issues/704

Issue #704, opened 2018-12-31
Tests using threads for rwops file objects


illume:
We don't have any tests for file objects which use threads (that I'm aware of).

- [ ] loading images in separate threads
- [ ] using python file objects with pygame.mixer_music.

illume:
It still needs tests using file objects, and music still needs threaded tests.

The issue with gif loading in threads might need some attention (like adding a lock around image loading when a gif is being loaded).

PR #707, opened 2019-01-01
Thread tests

dlon:
Test: Load image on multiple threads.
Test: Load music from file-like objects.

Loading music from threads (whether file objects or paths) seems to fail.

#704


dlon:
It's SDL_image 1.x + GIFs

PR #708, opened 2019-01-01
SDL2: Render targets, viewports, empty textures

dlon:
* Renderer: Adds [get_viewport](https://wiki.libsdl.org/SDL_RenderGetViewport)/[set_viewport](https://wiki.libsdl.org/SDL_RenderSetViewport). I used `get_viewport` and `set_viewport` instead of a setter and getter, because get_viewport doesn't return `None` if set_viewport is set to `None` (it returns the entire area instead). That's counter-intuitive for a variable (but I also don't think it should return 
`None`).
* Adds `Renderer.target` for rendering to textures. I.e. you set r.target = sometexture or r.target=None.
* Adds `Texture.create_empty(renderer, size, depth=0, static=False, streaming=False, target=False)`. I assume you might prefer to have one constructor instead. Wouldn't `Texture.from_surface()` work also?

I set all of the "texture access" flags to False by default, because otherwise you'd have to use it like this:
```python3
tex2 = Texture.create_empty(renderer, (32,32), static=False, target=True)
```
Instead of just
```python3
tex2 = Texture.create_empty(renderer, (32,32), target=True)
```

## To do

- [ ] SDL_RenderReadPixels

For reading pixels (https://wiki.libsdl.org/SDL_RenderReadPixels), I'm considering using `renderer.read_pixels() -> Surface`. (A function rather than a property to emphasize its slowness.)

Related: #605


illume:
Very cool. Sorry I haven't had time to think about this properly.

I like that they are using more common/current/standard terms for things like viewport, rather than clip.

Yeah, maybe a single constructor would be better? Perhaps static methods are a bit too exotic for some folks? Not sure about this.

Playing around with it a bit to see what some options look like...

```python

tex = Texture(renderer, surface=surface)
tex = Texture(renderer, (20, 20), target=True)

tex = renderer.from_surface(surface)
tex = renderer.texture(surface=surface)
empty_tex = renderer.texture(size=(20, 20))
```

dlon:
Went with `Texture(renderer, size=(20,20))` and `Texture.from_surface(renderer, somesurface)` for now

llindstrom:
I did some experimenting with a [Cython view module](https://github.com/llindstrom/pygame_SDL2/blob/SDL2_sandbox/src/pygame.video.pyx) for pygame. It is somewhat out of data as it pre-dates the updates to the pygame C api. My approach to Texture creation was to have Renderer methods return new, open, Texture instances. The Texture class itself returned a closed, unusabled instance. An example using the experimental view module to create a surface texture is [here](https://github.com/llindstrom/pygame_SDL2/blob/SDL2_sandbox/examples/window_render.py).

dlon:
That seems cleaner. Also, since `renderer.copy(tex)` is redundant, should `texture.copy()` be used instead? I find the former slightly more intuitive, though.

PR #714, opened 2019-01-05
SDL2: Fix key mods test.

### LINKED ISSUE #532: SDL2 test_set_and_get_mods fails.
```
ERROR: test_set_and_get_mods (test.key_test.KeyModuleTest)
  File "/Users/rene/dev/pygame/pygame/test/key_test.py", line 60, in test_set_and_get_mods
    pygame.key.set_mods(pygame.KMOD_CTRL)
AttributeError: 'module' object has no attribute 'KMOD_CTRL'
```

dlon:
key_test now fails with this exception:

```
loading test.key_test
...F.
FAIL: test_set_and_get_mods (test.key_test.KeyModuleTest)
  File "C:\Users\David\Documents\dev\pygame\test\key_test.py", line 41, in test_set_and_get_mods
    self.assertEqual(pygame.key.get_mods(), pygame.KMOD_NONE)
AssertionError: 4096 != 0
```

The test's assumptions are wrong. The initial state might not be KMOD_NONE, since numlock or capslock might be active, so I just removed the assertion.

I don't know why, but it doesn't fail with SDL1, even though it should.

Should close #532

### DIFF
--- a/test/key_test.py
+++ b/test/key_test.py
@@ -38,8 +38,6 @@ class KeyModuleTest(unittest.TestCase):
         self.assertEqual(pygame.key.name(pygame.K_SPACE), "space")
 
     def test_set_and_get_mods(self):
-        self.assertEqual(pygame.key.get_mods(), pygame.KMOD_NONE)
-
         pygame.key.set_mods(pygame.KMOD_CTRL)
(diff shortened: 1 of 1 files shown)


illume:
Thanks :)

PR #715, opened 2019-01-05
SDL2: convert_alpha() used wrong pixel format. + Fix mask example

### LINKED ISSUE #713: SDL2: mask example fails
python -m examples.mask examples/data/alien1.gif examples/data/asprite.bmp

`surface.get_at` expects only integers?

```
  File "C:\Python37\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Python37\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\David\Documents\dev\pygame\examples\mask.py", line 172, in <module>
    main(*sys.argv[1:])
  File "C:\Users\David\Documents\dev\pygame\examples\mask.py", line 114, in main
    masks.append(maskFromSurface(images[-1]))
  File "C:\Users\David\Documents\dev\pygame\examples\mask.py", line 22, in maskFromSurface
    if surface.get_at((x+0.1,y+0.1)) != key:
TypeError: integer argument expected, got float
```

Windows, Python 3.7.

dlon:
Fix #713.

The first commit makes the `mask` example work for BMPs.

The second commit fixes the `mask` example for GIFs and PNGs. Loading GIFs sets the color key in SDL2, but not in SDL1. For some reason, the example uses floats in that case. That's why the TypeError is only raised with SDL2.

It seems like it might be a problem that color keys behave differently with SDL2?

### DIFF
--- a/examples/mask.py
+++ b/examples/mask.py
@@ -19,7 +19,7 @@ def maskFromSurface(surface, threshold = 127):
     if key:
         for y in range(surface.get_height()):
             for x in range(surface.get_width()):
-                if surface.get_at((x+0.1,y+0.1)) != key:
+                if surface.get_at((x,y)) != key:
                     mask.set_at((x,y),1)
     else:
         for y in range(surface.get_height()):
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1822,10 +1822,6 @@ surf_convert_alpha(PyObject *self, PyObject *args)
     PyObject *final;
     pgSurfaceObject *srcsurf = NULL;
     SDL_Surface *newsurf, *src;
-#if IS_SDLv2
-    PyObject *pg_winsurf;
-    SDL_Surface *winsurf;
-#endif
 
     if (!SDL_WasInit(SDL_INIT_VIDEO))
         return RAISE(pgExc_SDLError,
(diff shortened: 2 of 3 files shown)


src_c/surface.c
```diff
@@ -1830,6 +1834,16 @@ surf_convert_alpha(PyObject *self, PyObject *args)
     if (!PyArg_ParseTuple(args, "|O!", &pgSurface_Type, &srcsurf))
         return NULL;
 
+#if IS_SDLv2
```
illume: I think pygame SDL1 also needs set_mode to be called first for this, and raises an exception.

surf.convert() probably also needs this check

illume:
I think perhaps the examples/testsprite.py issue is related to this. https://github.com/pygame/pygame/issues/717

illume:
This is the error message with SDL1.
```python
>>> import pygame
pygame 1.9.5.dev0
>>> s=pygame.image.load('alien1.png')
>>> s.convert()
pygame.error: cannot convert without pygame.display initialized
>>> s.convert_alpha()
pygame.error: cannot convert without pygame.display initialized
```

illume:
- [ ] convert() and convert_alpha() should use the same message as SDL1. `pygame.error: cannot convert without pygame.display initialized`
- [ ] surf.convert() needs same fix as surf.convert_alpha().
- [ ] the image used in examples/testsprite.py could be used in a test with surf.convert()
- [ ] a test which makes sure the exception is raised when set_mode is not called should be made in test/image_test.py.

illume:
This potentially will fix some more graphical glitches with solarwolf as well. Perhaps the gif difference is worth investigating and testing on there too.

dlon:
There are a couple of exceptions:
1. no display.init(): "cannot convert without pygame.display initialized".
2. display.init() without set_mode(): "no video mode set".

It behaves the same way in SDL2 now.

examples/mask.py
```diff
@@ -19,7 +19,7 @@ def maskFromSurface(surface, threshold = 127):
     if key:
         for y in range(surface.get_height()):
             for x in range(surface.get_width()):
-                if surface.get_at((x+0.1,y+0.1)) != key:
+                if surface.get_at((x,y)) != key:
```
illume: This is worth further investigation imho. Does `Surface.get_at work` with floats on SDL1 but not on SDL2?

illume: Oh. Interesting. I'm confused why floats were used there. Oh well. It will remain a mystery.

PR #716, opened 2019-01-06
SDL2: Let set_mode() update existing window's fullscreen state and other flags.

### LINKED ISSUE #712: SDL2: fullscreen after starting in windowed mode doesn't work.
Thanks to blubberquark for providing this issue report.

dlon:
Should close #712 but needs more testing.

### DIFF
--- a/examples/aliens.py
+++ b/examples/aliens.py
@@ -185,6 +185,7 @@ def main(winstyle = 0):
         print ('Warning, no sound')
         pygame.mixer = None
 
+    fullscreen = False
     # Set the display mode
     winstyle = 0  # |FULLSCREEN
     bestdepth = pygame.display.mode_ok(SCREENRECT.size, winstyle, 32)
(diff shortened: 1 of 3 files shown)


illume:
I confirm that the test from @markph0204 works as well on my machine. I also see the graphical glitch when it changes modes.

However, I couldn't get glcube.py to work to change to fullscreen (I modified it to switch with the f key).

illume:
Ok. Now examples/glcube.py can toggle fullscreen.

I also added fullscreen toggling to examples/aliens.py with K_f. Fullscreen toggles, however I can't see any graphics after a toggle.

markph0204:
pygame1.9.5.dev0 | SDL2.0.9 | Mac 10.13.6 | py3.7.2

- glcube fullscreen toggle works
- aliens fullscreen toggle works; playable; when in fullscreen game is square with the left/right ends clipped black

illume:
I can't get fullscreen without-gl working properly on my mac machine... but it works on other machines, and it's still better than before. So I merged that in.

PR #722, opened 2019-01-07
Add version information macros.

dlon:
Adds the following macros:

* PG_MAJOR_VERSION
* PG_MINOR_VERSION
* PG_PATCH_VERSION
* PG_VERSIONNUM(MAJOR, MINOR, PATCH) (1000*(MAJOR) + 100*(MINOR) + (PATCH))
* PG_VERSION_ATLEAST(MAJOR, MINOR, PATCH)

Analogous macros exist in SDL.

setup.py makes sure the version info is consistent between the version module, setup script, and header.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,7 @@ Credits
 
      Ulf Ekström for his pixel perfect collision detection code.
 
-     Pete Shinners - orginal author.
+     Pete Shinners - original author.
 
      David Clark - for filling the right-hand-man position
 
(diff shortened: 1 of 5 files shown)


illume:
Very nice.

illume:
Great. Your new version macro will come in handy if we have some experimental features we don't want people to rely on.

By the way... Not sure if I told you or just wrote in the Discord...

I started labeling the issues with milestone 2.1 for things which are new SDL2 additions, and left he 2 milestone for things that are compatibility issues. This way it's a bit easier for us to see when pygame 2 is ready. However, I think it's fine for us to include SDL2 features in pygame 2 if they are ready.

PR #723, opened 2019-01-08
Add textinput example

husano896:
Use python -m pygame.examples.textinput should work.

Adding 'showevent' in argv like python -m pygame.examples.textinput showevent

will show events from [ KEYDOWN, TEXTEDITING, TEXTINPUT ].

Note: exception will be raised if SDL version < 2.0

TODO:
- Complete documentation for example
- Add more fonts for multi-language support

### DIFF
--- a/examples/textinput.py
+++ b/examples/textinput.py
@@ -1,0 +1,151 @@
+import pygame, sys
+import pygame.freetype
+
+###CONSTS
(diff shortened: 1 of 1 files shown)


illume:
Thanks. I merged it in. I know there are some TODO notes there in your description, but those can come in separate pull requests.

PR #737, opened 2019-01-15
re-apply color key after convert

### LINKED ISSUE #717: SDL2: examples/testsprite.py and examples/aliens.py alpha does not work
It seems the color key alpha does not work once .convert() is called on sprites.

robertpfeiffer:
In order to set the colour key of the converted surface, get the colour
key of the old surface and map to RGB, then map back to the format of
the converted surface for applying colour key there.
This is necessary to convert 8-bit palette-mapped surfaces correctly.
The colour key of an 8-bit surface is an 8-bit palette index (usually
the first palette entry, value 0), which can not be used as colour key
when the target surface uses a different palette or RGB(A) format.
Unfortunately both are represented as 32-bit unsigned ints.

this should fix #717

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1605,7 +1605,9 @@ surf_convert(PyObject *self, PyObject *args)
 #if IS_SDLv2
     Uint8 alpha;
     Uint32 colorkey;
-    int ecode;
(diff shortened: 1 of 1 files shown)


dlon:
Works for me.

Issue #721, opened 2019-01-06
SDL2: solarwolf game should work with SDL2.


illume:
One old game that can be used for compatibility testing is solarwolf.
https://github.com/pygame/solarwolf

robertpfeiffer:
Solarwolf and the aliens example (#717) are both using indexed 8-bit images. Solarwolf has some indexed pngs with per-pixel alpha, i.e. palette colours are not RGB but RGBA. Additionally, solarwolf also uses per-sprite set_alpha() and set_colorkey(). If you convert the pngs to 32bit RGBA with imagemagick, the glitches in the main menu disappear, so it must be related to 8bit surfaces.

I think I have isolated the problem:
```python
import pygame
pygame.init()

screen=pygame.display.set_mode((200,150))
surf=pygame.image.load("solarwolf/data/menu_creds_off.png")
print(surf.get_colorkey())
print(surf.convert().get_colorkey())
```
output with old pygame:
```
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
(219, 10, 8, 255)
(219, 10, 8, 255)
```
output with pygame+SDL2
```
pygame 1.9.5.dev0
Hello from the pygame community. https://www.pygame.org/contribute.html
(219, 10, 8, 255)
(0, 0, 0, 0)
```

robertpfeiffer:
saving and re-applying the colorkey fixes the issue *and I have no idea why*
```
def my_convert():
    c=img.get_colorkey()
    img=img.convert()
    img.set_colorkey(c)
    return img
```

surf_convert surface.c already calls SDL_SetColorKey.

illume:
Thanks heaps for narrowing this down and coming up with the test case :)

robertpfeiffer:
so this probably means if RLEACCEL is set, then convert still breaks colourkey somehow.

robertpfeiffer:
minimal test case, again

```
import pygame
pygame.init()
screen=pygame.display.set_mode((200,150))
surf=pygame.image.load("data/menu_creds_off.png")

print(surf.get_colorkey())
print(surf.convert().get_colorkey())

surf.set_colorkey(surf.get_colorkey(), pygame.RLEACCEL)
print(surf.get_colorkey())
print(surf.convert().get_colorkey())
```
on pygame+SDL2 commit c474aa9
```
(219, 10, 8, 255)
(219, 10, 8, 255)
(75, 75, 75, 255)
(75, 75, 75, 255)
```
[...]

PR #739, opened 2019-01-15
SDL2: Fix MapRGB for surfaces with a color key and palette

dlon:
I don't know if this is expected behavior in SDL2, but SDL_MapRGB always assumes the alpha value is 255 for palettes, which means it doesn't work for the color key, whose alpha is 0. (This bug has been reported already: https://bugzilla.libsdl.org/show_bug.cgi?id=3827.)

Cf. #721


dlon:
#721 is fixed due to this and #737.

Issue #752, opened 2019-01-22
Clean up deprecated unittest assert method warnings


charlesej:
Some unittests are using deprecated assert methods. They do not cause tests to fail, but they do make it hard to read the test output. Here are links to information on the deprecated assert methods for python [2.7](https://docs.python.org/2.7/library/unittest.html#deprecated-aliases) and [3](https://docs.python.org/3/library/unittest.html#deprecated-aliases)

Example of deprecation warning.
```
py -3 -m pygame.tests.font_test
pygame 1.9.5.dev0
Hello from the pygame community. https://www.pygame.org/contribute.html
.......................C:\Program Files\Python37\Lib\site-packages\pygame-1.9.5.dev0-py3.7-win-amd64.egg\pygame\tests\font_test.py:246: DeprecationWarning: Please use assertTrue instead.
  self.assert_(um[0] is not None)
............
----------------------------------------------------------------------
Ran 35 tests in 3.167s

OK
```

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit)
- pygame: 1.9.5.dev0 (SDL: 1.2.15) at 269d6892fea8bf0aaacf5e13b55d7805e4937469
- numpy: 1.15.4

Here's a list of the offenders.
- [x] `failUnless` (5 files)
- [x] `assert_` (48 files)
- [x] `assertEquals` (8 files)
- [x] `assertAlmostEquals` (1 file)
- [x] `assertRaisesRegexp`(1 file) - the replacement `assertRaisesRegex` does not exist in python 2.7, so a python version flag will be needed for this one

PR #754, opened 2019-01-22
Clean up deprecated assertEquals() and assertAlmostEquals()

charlesej:
This update cleans up the use of the deprecated `unittest.TestCase.assertEquals()` and  `unittest.TestCase.assertAlmostEquals()` methods.

Overview of changes:
- Replace `assertEquals()`/`assertAlmostEquals()` with an appropriate `unittest.TestCase` assert method and alter the code to work with this new method.
- Some formatting for readability.
- Add the decorator `@unittest.expectedFailure` to the failing test method `test_overlap_mask()`. The test method was tagged with `|tags: ignore|`, but `unittest` should be informed of the expected the failure.
- Update `print` in `regexes.py` to make it python 3 compatible.

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit)
- pygame: 1.9.5.dev0 (SDL: 1.2.15) at 269d6892fea8bf0aaacf5e13b55d7805e4937469
- numpy: 1.15.4

Resolves the assertEquals and assertAlmostEquals items of #752.

### DIFF
--- a/test/color_test.py
+++ b/test/color_test.py
@@ -640,10 +640,10 @@ class ColorTypeTest (unittest.TestCase):
 
         t = c.normalize()
 
-        self.assertAlmostEquals(t[0], 0.800000, 5)
-        self.assertAlmostEquals(t[1], 0.149016, 5)
-        self.assertAlmostEquals(t[2], 0.760784, 5)
-        self.assertAlmostEquals(t[3], 0.215686, 5)
+        self.assertAlmostEqual(t[0], 0.800000, 5)
+        self.assertAlmostEqual(t[1], 0.149016, 5)
+        self.assertAlmostEqual(t[2], 0.760784, 5)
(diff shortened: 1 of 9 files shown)


charlesej:
Upon further review: Allowing `unittest` to fail `test_overlap_mask()` is intended. The `@unittest.expectedFailure` decorator has been removed.

Reference: issues #410 and #500.

illume:
Thanks @charlesej 

I guess @unittest.expectedFailure was a good idea there. For tests that fail, but don't segfault, it's a good idea to use expectedFailure. If the test doesn't segfault, then we can remove the ignore doc comment and use the expectedFailure.

Probably we can remove test/util/build_page/ folder entirely now. It hasn't been used for several years, and I don't think anyone will get around to using it again. If someone wants to it's always available in the git history.

charlesej:
Thanks for the comments. I'll open 2 new issues to track those items.

PR #787, opened 2019-02-02
Fix Solarwolf menu and RLE bug.

dlon:
See: #721.

This makes Solarwolf look right, but it disables RLE in too many cases. I think it should also be available if Amask=0 and a color key is set, but this screws up the blitting.

How to test it: In Solarwolf, press enter for any menu item, then go back. The buttons will be rendered incorrectly.

Edit: It's probably worth figuring out why it fails.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1419,9 +1419,14 @@ surf_set_colorkey(PyObject *self, PyObject *args)
          */
         result = SDL_SetColorKey(surf, SDL_FALSE, color);
     }
-    if (result == 0 && hascolor) {
-        result = SDL_SetSurfaceRLE(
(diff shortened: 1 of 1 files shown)


illume:
Yeah, I think this needs a unit test so we can check it's fixed on different platforms. I guess based off the code @dlon wrote in https://github.com/pygame/pygame/issues/721#issuecomment-459548722

illume:
Hrmm... it doesn't seem to fix it for me.

dlon:
Disabling RLE entirely probably fixes it, but I guess that's no solution.

illume:
Closing this to keep our PR list tidy.

Issue #792, opened 2019-02-03
Font files not being closed


charlesej:
In some cases of font creation, the font files are not being closed.

The following code run with the warnings set to default (`python -Wd font_resource_warning.py`), produces unclosed font file `ResourceWarnings`.

Notes:
- There does not appear to be any warnings for some of the `ftfont` calls.
- The output for pygame 1.9.5.dev0 has more warnings than for pygame 1.9.4. The latter only seems to produce warnings when `Font()` is given `None` as the filename.

```
# File font_resource_warning.py.
import pygame
import pygame.freetype
import pygame.ftfont

SPACER = '-' * 75

# Test font.
pygame.font.init()

print(SPACER)
print(r"pygame.font.Font(r'c:\windows\fonts\arial.ttf', 20)""\n")
pygame.font.Font(r'c:\windows\fonts\arial.ttf', 20)

print(SPACER)
print(r"pygame.font.SysFont('arial', 20)""\n")
pygame.font.SysFont('arial', 20)

print(SPACER)
print(r"pygame.font.Font(None, 20)""\n")
pygame.font.Font(None, 20)
pygame.font.quit()


# Test freetype.
pygame.freetype.init()

print(SPACER)
print(r"pygame.freetype.Font(r'c:\windows\fonts\arial.ttf', 20)""\n")
pygame.freetype.Font(r'c:\windows\fonts\arial.ttf', 20)

print(SPACER)
print(r"pygame.freetype.SysFont('arial', 20)""\n")
pygame.freetype.SysFont('arial', 20)

print(SPACER)
print(r"pygame.freetype.Font(None, 20)""\n")
pygame.freetype.Font(None, 20)
pygame.freetype.quit()


# Test ftfont.
pygame.ftfont.init()

print(SPACER)
print(r"pygame.ftfont.Font(r'c:\windows\fonts\arial.ttf', 20)""\n")
pygame.ftfont.Font(r'c:\windows\fonts\arial.ttf', 20)

print(SPACER)
print(r"pygame.ftfont.SysFont('arial', 20)""\n")
pygame.ftfont.SysFont('arial', 20)

print(SPACER)
print(r"pygame.ftfont.Font(None, 20)""\n")
pygame.ftfont.Font(None, 20)
print(SPACER)
pygame.ftfont.quit()
```

Using pygame 1.9.5.dev0 (at 735a0823f71b28f3379d2d01c45cadd05d245e05), the above code gives the following output.
```
pygame 1.9.5.dev0
[...]

PR #796, opened 2019-02-04
Close font files

charlesej:
This update closes some open font files.

Overview of changes:
- Closed the open file in `test_freetype_Font_dealloc()`.
- Closed the open file in the `Font` class when the object is deleted/destroyed.

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit)
- pygame: 1.9.5.dev0 (SDL: 1.2.15) at fe8e32d8c77d426fb18d30aaae97f8f81e12406c

Resolves part of #792.

### DIFF
--- a/src_py/freetype.py
+++ b/src_py/freetype.py
@@ -47,6 +47,8 @@ def SysFont(name, size, bold=0, italic=0, constructor=None):
 
 class Font(_Font):
     def __init__(self, file, *args, **kwargs):
+        self._file = None
+
         if sys.platform == 'win32':
             if isinstance(file, bytes):
                 # Windows paths are unicode...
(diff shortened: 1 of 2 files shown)


src_py/freetype.py
```diff
@@ -55,6 +57,15 @@ def __init__(self, file, *args, **kwargs):
             if isinstance(file, compat.unicode_):
                 # ...but we can't pass a unicode path to Freetype,
                 # so we'll open the file in Python and pass the handle instead.
-                file = open(file, 'rb')
```
dlon: It may be possible to remove this code for Windows altogether now. _freetype should work with Unicode paths as of 864888f1d7f2.

dlon: It may be possible to remove all of the code for Windows here. _freetype should work with Unicode paths as of 864888f.

charlesej: Thanks for the insight. I've updated the code.

Issue #794, opened 2019-02-04
PyGame application crashes on SIGABRT - xcb_io.c:165: dequeue_pending_request: Assertion `!xcb_xlib_unknown_req_in_deq' failed.


jirikeller:
This does not happen always, however is present on both latest raspbian and Debian.
The error is: 

[xcb] Unknown request in queue while dequeuing
[xcb] Most likely this is a multi-threaded client and XInitThreads has not been called
[xcb] Aborting, sorry about that.
python: ../../src/xcb_io.c:165: dequeue_pending_request: Assertion `!xcb_xlib_unknown_req_in_deq' failed.
Neúspěšně ukončen (SIGABRT)

The code is multithreaded, starting and stopping parallel (non-GUI) threads. Based on this: 
https://stackoverflow.com/questions/31952711/threading-pyqt-crashes-with-unknown-request-in-queue-while-dequeuing
the problem seems to be realated to PyQt pre-initialization phase. I do not import PyQt at all and rely to pygame code. Can you,please, add 

QtCore.QCoreApplication.setAttribute(QtCore.Qt.AA_X11InitThreads)

prior QApplication constructor ?

Otherwise I am happy with pygame and I would like to promote it in my scientific project ;-)

dlon:
Haven't checked but this may not be a problem if you build with SDL2. It does call `X11_XInitThreads()`.

Note that not everything in SDL is thread-safe, though, particularly the video functions. Another one is `SDL_PumpEvents()`. This is called by `event.get()`, among other functions. So you may have to run these on the main thread. Maybe some of this could change.

dlon:
Hello
If you were just listening to events in the keyboard thread, then that should be safe to do when #805 has been committed (using `pump=False`). I think you have to call `event.pump()` in the thread that initialized pygame, though.

jirikeller:
pump=False and git version of pygame solved the problem for me ! Thanks

PR #805, opened 2019-02-08
Add pump parameter and keyword arguments to some event functions

dlon:
* Adds keyword arguments: `event.clear(eventtype=None, pump=True)`, `event.clear(eventtype=None, pump=True)`, and `event.get(eventtype=None, pump=True)`.
* `pump` can be set to `False` to safely use these functions in other threads (w/ GIL). This works for both SDL1 and SDL2.

Related: #794.


docs/reST/ref/event.rst
```diff
@@ -151,6 +152,8 @@ attributes.
    If you are only taking specific events from the queue, be aware that the
    queue could eventually fill up with the events you are not interested.
 
+   If ``pump`` is ``True``, then :func:`pygame.event.pump()` will be called.
```
charlesej: For all the API changes, you might want to consider adding a version notification (e.g. `.. versionadded:: 1.9.5`) with a short description of what is new.

src_c/event.c
```diff
@@ -1617,9 +1703,9 @@ static PyMethodDef _event_methods[] = {
     {"pump", (PyCFunction)pg_event_pump, METH_NOARGS, DOC_PYGAMEEVENTPUMP},
     {"wait", (PyCFunction)pg_event_wait, METH_NOARGS, DOC_PYGAMEEVENTWAIT},
     {"poll", (PyCFunction)pg_event_poll, METH_NOARGS, DOC_PYGAMEEVENTPOLL},
-    {"clear", pg_event_clear, METH_VARARGS, DOC_PYGAMEEVENTCLEAR},
-    {"get", pg_event_get, METH_VARARGS, DOC_PYGAMEEVENTGET},
-    {"peek", pg_event_peek, METH_VARARGS, DOC_PYGAMEEVENTPEEK},
+    {"clear", pg_event_clear, METH_VARARGS | METH_KEYWORDS, DOC_PYGAMEEVENTCLEAR},
```
charlesej: You may need to cast these methods as `(PyCFunction)` to avoid compiler warnings.

Ref: https://docs.python.org/3/c-api/structures.html#c.PyMethodDef

docs/reST/ref/event.rst
dlon: Sure. I think the other documentation should be updated to use the directive instead of text also? If the text for ".. versionadded X" is changed to "New in pygame X".

illume:
Oops. I made some conflicts by merging in the other branch.

Issue #800, opened 2019-02-06
Improvements to pygame.mask


charlesej:
Here's a list of improvements that can be made to `pygame.mask` related areas.

- [x] Add customized `repr()`/`str()` output for `Mask` objects.

- [x] Add an argument to the `Mask()` constructor to provide users the option to have the mask filled on creation. Also, update the related documentation and add a test for this.

- [x] Provide consistent boundary checking in `bitmask.c`. The functions `bitmask_overlap()`, `bitmask_overlap_pos()`, `bitmask_overlap_area()`, `bitmask_overlap_mask()`, `bitmask_draw()`, and `bitmask_erase()` should all have the same boundary checking. Some are missing the `xoffset` check for the second mask.

- [x] Rearrange `mask_test.py`. Some of the test methods in the `MaskModuleTest()` class should be moved to the `MaskTypeTest()` class (i.e. `test_overlap_mask()` and all the `test_zero_mask_`* methods).

- [x] Add tests for some of the `Mask` class methods that are not being tested. Also, add test stubs (`todo_test_`) for any methods that do not have a test.

- [ ] Update the `pygame.mask` documentation. Some items of note.
  - Provide a better explanation for how the `overlap()` method finds its "point of intersection" (i.e. How does it search/compare the masks?).
  - Move the offsets ASCII art example, currently under the `overlap()` method, to the general description area (create one) for `pygame.mask.Mask`.

Adding a few more items:
- [x] Add support to allow the `Mask` class to be subclassed. Also, update the related documentation and add tests.
- [x] Consolidate the `Mask` creation code (in `mask.c`) to allow for consistent error checking and exception messages (e.g. some of the areas where `PyObject_New(pgMaskObject, &pgMask_Type)` and `bitmask_create(w, h)` are used).

PR #806, opened 2019-02-08
Add fill kwarg to the Mask constructor

charlesej:
This update adds a `fill` keyword parameter to the `Mask` constructor.

Overview of changes:
* Added the `fill` keyword parameter to the `Mask` constructor. The default value is `False` to maintain backward compatibility.
* Changed the `Mask` constructor to raise a `PyExc_MemoryError` if `bitmask_create()` fails to return a valid mask object.
* Updated the `pygame.mask` documentation.
* Added test methods to test `Mask` constructor parameters.

System details:
* os: windows 10 (64bit)
* pygame: 1.9.5.dev0 (SDL: 1.2.15) at 5de404e926f4b7563b90839fe4e91cda46dba4b3

Resolves the "Add an argument to the `Mask()` constructor..." item of #800.

### DIFF
--- a/docs/reST/ref/mask.rst
+++ b/docs/reST/ref/mask.rst
@@ -46,7 +46,7 @@ Starting from pygame 1.9.5 masks with width or height 0 are supported.
 .. class:: Mask
 
    | :sl:`pygame object for representing 2D bitmasks`
-   | :sg:`Mask((width, height)[, fill=False]) -> Mask`
+   | :sg:`Mask(size=(width, height)[, fill=False]) -> Mask`
 
    A ``Mask`` object is used to represent a 2D bitmask. Each bit in
    the mask represents a pixel. 1 is used to indicate a set bit and 0 is used
@@ -61,13 +61,15 @@ Starting from pygame 1.9.5 masks with width or height 0 are supported.
    be accessed using the :func:`pygame.mask.Mask.get_at()` and
    :func:`pygame.mask.Mask.set_at()` methods.
 
-   :param list|tuple (width, height): sequence of 2 ints indicating the width
+   :param list|tuple size: sequence of 2 ints indicating the width
       and height of the mask
    :param bool fill: create mask unfilled (``False`` - default) or filled
       (``True``)
    :rtype: ``Mask`` object
 
-   .. versionadded:: 1.9.5 The keyword parameter ``fill``.
+   .. versionadded:: 1.9.5 Named parameter ``size`` (previously it was an
+      unnamed positional parameter) and the optional keyword parameter
+      ``fill``.
 
    .. method:: get_size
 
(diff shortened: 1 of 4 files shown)


src_c/mask.c
```diff
@@ -1555,27 +1555,42 @@ static PyTypeObject pgMask_Type = {
...
 static PyObject *
-Mask(PyObject *self, PyObject *args)
+Mask(PyObject *self, PyObject *args, PyObject *kwargs)
 {
     bitmask_t *mask;
     int w, h;
+    int fill = 0; /* Default is false. */
     pgMaskObject *maskobj;
-    if (!PyArg_ParseTuple(args, "(ii)", &w, &h))
+    char *keywords[] = {"", "fill", NULL};
```
dlon: I think `size` could be named as well. Guess it's for backward compatibility?

charlesej: Thanks for the comment.

I was just going to leave it as an unnamed positional-only parameter, but giving it a name is a good idea I've made the change.

docs/reST/ref/mask.rst
```diff
@@ -45,8 +45,31 @@ Starting from pygame 1.9.5 masks with width or height 0 are supported.
 
 .. class:: Mask
 
-   | :sl:`pygame object for representing 2d bitmasks`
-   | :sg:`Mask((width, height)) -> Mask`
+   | :sl:`pygame object for representing 2D bitmasks`
+   | :sg:`Mask(size=(width, height)[, fill=False]) -> Mask`
```
dlon: Maybe this should show two signatures instead? I.e.

```
| :sg:`Mask(size=(width, height)) -> Mask`
| :sg:`Mask(size=(width, height), fill=False) -> Mask`
```

docs/reST/ref/mask.rst
```diff
@@ -45,8 +45,31 @@ Starting from pygame 1.9.5 masks with width or height 0 are supported.
...
+
+   A filled mask has all of its bits set to 1, conversely an unfilled/cleared
+   mask has all of its bits set to 0. Masks can be created unfilled (default)
+   or filled by using the ``fill`` parameter. Masks can also be cleared or
+   filled using the :func:`pygame.mask.Mask.clear()` and
+   :func:`pygame.mask.Mask.fill()` methods respectively. Individual bits can
+   be accessed using the :func:`pygame.mask.Mask.get_at()` and
+   :func:`pygame.mask.Mask.set_at()` methods.
+
+   :param list|tuple size: sequence of 2 ints indicating the width
```
dlon: I think the correct syntax is `:param list or tuple size:`, according to http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html

docs/reST/ref/mask.rst
charlesej: Yes it should. I've made the requested change.

docs/reST/ref/mask.rst
charlesej: Thanks for the suggestions. The `param` field doesn't seem to work using multiple words for the type, so I had to use the `type` field. Also, I decided to avoid using `sequence` because it doesn't just mean a tuple or list (e.g. string). I've updated the code accordingly.

PR #809, opened 2019-02-09
Updated one file based off PR #658

markph0204:
- here I am showing example of how to proceed with documentation/update if warranted to use type defs in params for docs
- this *should* improve readability/processing of some IDE's like IDEA

### DIFF
--- a/docs/reST/ref/surface.rst
+++ b/docs/reST/ref/surface.rst
@@ -148,10 +148,9 @@
    .. method:: convert
 
       | :sl:`change the pixel format of an image`
-      | :sg:`convert(Surface) -> Surface`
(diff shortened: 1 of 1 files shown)


markph0204:
I'm waiting for original PR requestor to validate this output before proceeding with updates across all docs.  This is just one file example.

NEW
image


ORIGINAL
image

docs/reST/ref/surface.rst
```diff
@@ -148,10 +148,9 @@
    .. method:: convert
 
       | :sl:`change the pixel format of an image`
-      | :sg:`convert(Surface) -> Surface`
+      | :sg:`convert(Surface=None) -> Surface`
```
Bottersnike: Looks good

Issue #815, opened 2019-02-12
Touch device functions


dlon:
The events are already implemented (#604), but there are also these functions which may be useful:

https://wiki.libsdl.org/SDL_GetNumTouchDevices
https://wiki.libsdl.org/SDL_GetNumTouchFingers
https://wiki.libsdl.org/SDL_GetTouchDevice
https://wiki.libsdl.org/SDL_GetTouchFinger

husano896:
For all touch device related functions, will 'pygame.touch' be a good idea?

Issue #814, opened 2019-02-12
Touch gestures


dlon:
Functions:

https://wiki.libsdl.org/SDL_RecordGesture
https://wiki.libsdl.org/SDL_SaveDollarTemplate
https://wiki.libsdl.org/SDL_SaveAllDollarTemplates
https://wiki.libsdl.org/SDL_LoadDollarTemplates
https://wiki.libsdl.org/SDL_DollarGestureEvent

PR #820, opened 2019-02-12
Touch module

dlon:
Re: #815

- [x] Might need to use long long for SDL finger IDs and touch IDs? The type is `Sint64`.
- [x] Needs documentation.
- [ ] `touch.get_finger()` returns a dict right now. Maybe it should return a class?
- [x] Needs testing on an actual touch device.

Also, gestures (#814) should probably be included here.

### DIFF
--- a/src_c/touch.c
+++ b/src_c/touch.c
@@ -20,6 +20,7 @@
 #include "pygame.h"
 #include "pgcompat.h"
 
+#include "doc/touch_doc.h"
 
 static PyObject *
 pg_touch_num_devices(PyObject *self, PyObject *args)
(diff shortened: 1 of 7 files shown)


src_c/touch.c
```diff
@@ -0,0 +1,160 @@
+/*
+  pygame - Python Game Library
+  Copyright (C) 2000-2001  Pete Shinners
+  Copyright (C) 2019  David L�nnhager
```
illume: You can just have your name here.

dlon:
Should this be moved into ._sdl2?

husano896:
> Should this be moved into ._sdl2?

Since almost all top-level modules will have at least some SDL1/2 support.

If a module can only work with SDL2, it's the best to move it into ._sdl2.

PR #821, opened 2019-02-13
experimental _sdl2.mixer module with post_mix()

illume:
Work in progress, please don't merge.

post_mix is useful for recording audio sent through the mixer, or visualising the audio.


markph0204:
I like your use of WIP: -- Gitlab recognizes that as a do not merge -- but surprised Github doesn't have similar.  I created a label for such items...applying here.

dlon:
It could probably be part of the top-level mixer module instead: https://www.libsdl.org/projects/SDL_mixer/docs/SDL_mixer_79.html

illume:
Yop, I think it should be part of the top level mixer module.
Good idea about Mix_RegisterEffect() too.

I'm going to merge this in however for now to do some testing of this function on different platforms.

PR #826, opened 2019-02-15
Add @unittest.skipIf() decorator to silently skipped tests

charlesej:
This update adds the `@unittest.skipIf()` decorator to some silently skipped tests.

Note:
* The `opengl` tag is being used to skip the `test_image_save_works_with_opengl_surfaces()` test in some instances. Using the decorator here makes the skipping explicit and allows the test to be skipped when testing without tag support.

System details:
* os: windows 10 (64bit)
* pygame: 1.9.5.dev0 (SDL: 1.2.15) at 4e42dd6ac1b050ca7cbe1e24c49942857f8304e3

### DIFF
--- a/test/image__save_gl_surface_test.py
+++ b/test/image__save_gl_surface_test.py
@@ -5,6 +5,9 @@ from pygame.tests import test_utils
 import pygame
 from pygame.locals import *
 
+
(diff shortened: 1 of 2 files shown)


markph0204:
Smart cleanup -- ty!

PR #832, opened 2019-02-18
SDL2: allow set_mode to "logical" screen size, scaled to actual screen

robertpfeiffer:
Feedback welcome, don't merge yet!

Following https://wiki.libsdl.org/MigrationGuide, create an SDL_Renderer
which renders the contents of the display surface to the screen.

pygame.display.set_mode((w,h), pygame.LOGICAL) will always create a full-screen window, but on the Python side it returns a surface of the requested size. Scaling the texture to screen size is done by SDL, scaling mouse coordinates down is done in pygame to make it seamless.

This lets you run a low-res game on your HD monitor without the slowness of pygame.transform.scale, and without additional coordinate wrangling for bigger screens.

I have no idea what the API should look like, if the name/value of pygame.LOGICAL are likely to
clash with something else in the future, or if we should just bail early when LOGICAL is used together with FULLSCREEN or OPENGL.


illume:
Some related documentation:

https://wiki.libsdl.org/SDL_RenderSetLogicalSize
https://wiki.libsdl.org/SDL_CreateWindow
https://wiki.libsdl.org/SDL_GetDesktopDisplayMode
https://wiki.libsdl.org/SDL_SetWindowFullscreen
https://wiki.libsdl.org/SDL_SetWindowPosition

src_c/display.c
```diff
@@ -755,32 +784,45 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
-                SDL_GetWindowPosition(win, &x, &y);
+            if (flags & PGS_LOGICAL && !(flags & PGS_FULLSCREEN)){
+                SDL_DisplayMode dm;
+                int scale=2;
+
+                if (SDL_GetDesktopDisplayMode(0, &dm) != 0) {
+                    return RAISE(pgExc_SDLError, SDL_GetError());
+                }
+
+                while(w*scale <= dm.w && h*scale <= dm.h){
```
charlesej: There might be a better way to do this scaling: https://stackoverflow.com/a/14731922

robertpfeiffer: Sure thing. This could be done in closed form with a couple of divisions, roundings and min() operations, if that makes the code clearer. I don't think it matters much in terms of performance though.

robertpfeiffer:
As per @illume's request, I'll explain why I think this should be merged now: If we just add the LOGICAL flag to pygame, it's completely backwards-compatible. If you don't use LOGICAL, everything should stay the same, and if you use SDL1.2, LOGICAL just gets ignored. FULLSCREEN|LOGICAL is interpreted just like FULLSCREEN. If we add the LOGICAL flag to pygame, games can use LOGICAL if and when it becomes available.

If we don't add the LOGICAL flag and related code to the mainline yet, pygame.display.set_mode((160,144), pygame.LOGICAL) will just throw an error, but I think we should add it now as experimental code.

The whole idea of this feature is to make screen scaling and HD display support transparent, i.e. invisible to the pygame user. You can do the same thing with GL and get_view, or by using pygame.transform.scale manually, but I hope we can piggyback on SDL2 to automatically pick the fastest available method, which might even be DirectX.

Right now, what needs to be done is to make LOGICAL either play nice with explicit use of SDL_Renderer and textures, or just forbid the usage of LOGICAL together with SDL_Renderer. It's already not allowed to use LOGICAL together with OPENGL.

pygame.display.update(list) could be implemented so that only the sub-regions that have changed will be updated, but I have a suspicion this might be slower in the end than just updating the whole buffer in one step.

src_c/mouse.c
```diff
@@ -55,6 +72,34 @@ mouse_get_pos(PyObject *self)
 
     VIDEO_INIT_CHECK();
     SDL_GetMouseState(&x, &y);
+
+#if IS_SDLv2
+    SDL_Window *sdlWindow = pg_GetDefaultWindow();
+    SDL_Renderer *sdlRenderer = SDL_GetRenderer(sdlWindow);
```
dlon: These must be declared at the beginning of a block due to that old VC compiler.

src_c/display.c
```diff
@@ -1067,8 +1125,16 @@ pg_flip(PyObject *self)
     if (state->using_gl) {
         SDL_GL_SwapWindow(win);
     }
-    else
+    else {
+        if (pg_renderer != NULL){
+            SDL_Surface *screen = pgSurface_AsSurface(pg_GetDefaultWindowSurface());
+            SDL_UpdateTexture(pg_texture, NULL, screen->pixels, screen->pitch);
```
dlon: Should we test whether this is faster than scaling up a surface in software, just to be sure? Streaming to a texture is pretty slow

src_c/display.c
```diff
@@ -827,7 +869,23 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                 SDL_GL_DeleteContext(state->gl_context);
                 state->gl_context = NULL;
             }
-            surf = SDL_GetWindowSurface(win);
+
+            if (flags & PGS_LOGICAL) {
+                if (pg_renderer==NULL){
+                    pg_renderer = SDL_CreateRenderer(win, -1, 0);
+                    SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "nearest");
```
dlon: I wonder if a linear filter is ever preferable?

src_c/display.c
```diff
@@ -755,32 +784,45 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+        int w_1=w;
+        int h_1=h;
+
+        if(win){
+            if (SDL_GetWindowDisplayIndex(win) == display) {
+                SDL_GetWindowPosition(win, &x, &y);
+            }
+            if (!(flags & PGS_OPENGL) !=
+                !(SDL_GetWindowFlags(win) & SDL_WINDOW_OPENGL)){
+                SDL_DestroyWindow(win);
```
dlon: There may be a memory leak for the placeholder surface here

src_c/_pygame.h
```diff
@@ -294,6 +294,8 @@ typedef enum {
     PGS_HWPALETTE = 0x20000000,
     PGS_DOUBLEBUF = 0x40000000,
     PGS_FULLSCREEN = 0x80000000,
+    PGS_LOGICAL =    0x00000200,
```
dlon: Maybe it should be called FULLSCREEN_DESKTOP?

src_c/display.c
```diff
@@ -827,7 +869,23 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+                    pg_renderer = SDL_CreateRenderer(win, -1, 0);
+                    SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "nearest");
+                    SDL_RenderSetLogicalSize(pg_renderer, w, h);
+
+                    pg_texture = SDL_CreateTexture(pg_renderer,
+                                                   SDL_PIXELFORMAT_ARGB8888,
+                                                   SDL_TEXTUREACCESS_STREAMING,
+                                                   w, h);
+                }
+                surf = SDL_CreateRGBSurface(SDL_SWSURFACE, w, h, 32, 0xff << 16,
```
dlon: I think `owner = 1` should be set further down

robertpfeiffer:
> A couple of things I noticed, testing it with the aliens example:
> 
>     * The window size isn't updated when you set LOGICAL for an existing window.
That's definitely a problem for future-me..., but it *is* a problem.

>     * I noticed that LOGICAL can be used without FULLSCREEN now. It runs much more smoothly than pure software rendering despite not scaling anything. Maybe we should consider using the texture streaming method by default? (Although in an earlier attempt, I concluded that it was slower, so I'm not sure.)
It depends on your screen/window size. In windowed mode, I wrote it deliberately so windowed scaling is always a whole number. Scaling a game by 1.5, whether linear, bi-linear, or nearest-neighbour, usually looks worse than not scaling at all.
My main use case here is very small resolutions like 160x144 (GameBoy) or 320x240 (QVGA). I make many QVGA games and scale them up.

src_c/display.c
robertpfeiffer: In an ideal world, SDL_Renderer magically uses the right way to scale things, which might even be software? I found that streaming to a GPU texture is faster than scaling up in software on PC (mesa, nvidia) and Android (nexus 7, ouya), but not on RaspberryPI, but that was when using pygame 1.9.4 and modernGL or PGS4A (pygame 1.9.1? and GLES).

src_c/display.c
robertpfeiffer: Shouldn't that same memory leak also apply to the fake 1x1 openGL placeholder surface then?

markph0204:
Capturing output here: pygame1.9.5.dev0+**robertpfeiffer@sdl2-logical**

**RESULTS**

Prep:
```
git checkout -b robertpfeiffer-sdl2-logical master
git pull https://github.com/robertpfeiffer/pygame.git sdl2-logical
python buildconfig/config.py -sdl2
```

Test 1 command:
```SDL_RENDER_DRIVER=software python examples/testsprite.py -logical```

Test 2 command:
```python examples/testsprite.py -logical```

_Each test was run for ~2 mins_

**SDL2.0.9 | Mac 10.13.6 (Sierra) ATI Radeon HD 5750 1024 MB | py3.7.2**

T1 output:
```
| => SDL_RENDER_DRIVER=software python examples/testsprite.py -logical
pygame 1.9.5.dev0
[640, 480]
renderer backend: software
FPS: 77.080630
```

T2 output:
```
| => python examples/testsprite.py -logical
pygame 1.9.5.dev0
[640, 480]
renderer backend: opengl
FPS: 583.986369
```

**SDL2.0.9 | Mac 10.14.3 (Mojave) Intel HD Graphics 5000 1536 MB | py3.7.2** 

T1 output:
```
| => SDL_RENDER_DRIVER=software python examples/testsprite.py -logical
pygame 1.9.5.dev0
[640, 480]
renderer backend: software
FPS: 204.716644
```

T2 output:
```
| => python examples/testsprite.py -logical
pygame 1.9.5.dev0
[640, 480]
renderer backend: opengl
FPS: 452.442495
```

I also have a 3rd MacBook Pro I can try tests on if you'd like/helpful that is also Mojave -- just newer.  But it takes more setup on that one...long story short.

File: src_c/display.c [hunk 8]
```diff
@@ -901,7 +944,28 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+            if (flags & PGS_LOGICAL) {
+                if (pg_renderer==NULL){
+                    SDL_RendererInfo info;
+
+                    SDL_SetHint(SDL_HINT_RENDER_SCALE_QUALITY, "nearest");
+                    pg_renderer = SDL_CreateRenderer(win, -1, 0);
+                    SDL_RenderSetLogicalSize(pg_renderer, w, h);
+
+                    SDL_GetRendererInfo(pg_renderer, &info);
+                    printf("renderer backend: %s\n", info.name);
```
illume: note, printf

src_c/mouse.c
```diff
@@ -66,6 +115,21 @@ mouse_get_rel(PyObject *self)
     VIDEO_INIT_CHECK();
 
     SDL_GetRelativeMouseState(&x, &y);
+
+/*
```
illume: This code should probably be removed if it's not being used?

illume:
Continued at https://github.com/pygame/pygame/pull/1039

PR #835, opened 2019-02-18
Add check for mixer init during Sound()'s creation

### LINKED ISSUE #417: MemoryError when attempting to create Sound object before pygame.init()
- [ ] convert the code below reproducing the problem into a unit test.
- [ ] fix the problem when creating a Sound before initialising the mixer. Raise an nice error in this case rather than crashing.
- [ ] improve error messages for when trying to create from file path.
- [x] discuss making a `pygame.sound.load` or something like that which is similar to `pygame.image.load`. It makes sense to have a similar name for loading images and sounds.
  - Moved to its own issue: #836

---

Currently if trying to create a Sound object from a buffer before pygame (or maybe mixer?) is initialized, Python will throw a MemoryError. It would be much more useful if we got an error like "Sound objects cannot be created before pygame is initialized"

I am using Python 2.7.14 with Pygame 1.9.3 on arch. I can confirm this also happens in Python 3.6.4 with Pygame 1.9.4pre0. 

Reproduce with:
```
>>> filething=open("/home/myuser/something.ogg","rb")
>>> filedata=filething.read()
>>> filething.close()
>>> asdf=pygame.mixer.Sound(buffer=filedata)
MemoryError
```

If trying to create from a file, the error is also not terribly useful:

```
>>> asdf=pygame.mixer.Sound("/home/myuser/something.ogg")
pygame.error: Unable to open file '/home/myuser/something.ogg'
```

Thanks for all the great work on pygame, coming up on 20 years of using it!

charlesej:
This update changes `Sound()`'s creation to raise an appropriate exception if the mixer is not initialized.

Overview of changes:
- Changed `Sound()`'s creation to raise an appropriate exception if mixer not initialized
- Made the error messages for the mixer not being initialized consistent
- Added creation tests for `Sound()` and `Channel()`

System details:
* os: windows 10 (64bit)
* pygame: 1.9.5.dev0 (SDL: 1.2.15) at 07b414848bc21476b1cabdf1ab5f7ce4157eacd5

Resolves #417.

### DIFF
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -1419,7 +1419,7 @@ _chunk_from_array(void *buf, PG_sample_format_t view_format, int ndim,
     Py_ssize_t loop1, loop2, step1, step2, length, length2 = 0;
 
     if (!Mix_QuerySpec(&freq, &format, &channels)) {
-        RAISE(pgExc_SDLError, "Mixer not initialized");
+        RAISE(pgExc_SDLError, "mixer not initialized");
         return -1;
     }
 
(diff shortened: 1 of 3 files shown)


markph0204:
Smart improvements to tests

PR #852, opened 2019-02-25
SDL2: Use valid masks in surface tests.

dlon:
Fixes the three test errors seen in #600. (Four assertions still fail.)

Related: #848


illume:
I wonder why they are invalid masks in SDL2? Or if/how it matters for compatibility.

Issue #292, opened 2016-02-29
PyGame.Mixer.init() format size = -8 and +16


illume:
Originally reported by: **Richard St-Aubin (Bitbucket: [richard_st_aubin](https://bitbucket.org/richard_st_aubin), GitHub: Unknown)**

----------------------------------------

Hi guys,
I use pygame.mixer version 1.9.2 in a raspberry pi 2 with uname -a
Linux RicPi 4.1.17-v7+ #838 SMP Tue Feb 9 13:15:09 GMT 2016 armv7l GNU/Linux
I also use a Tektronic Type 422 oscilloscope connected to my raspberry pi 2 audio jack.
When I run this program fg.py using python 2.7.9 program, I received:
>>> ================================ RESTART ================================
>>> 
8 8 -8 32776 16 16 -16 32784
Set (22050, 8, 1)
Get (22050, 8, 1)
Set (22050, -8, 1)
Get (22050, 8, 1)
Set (22050, 16, 1)
Get (22050, -16, 1)
Set (22050, -16, 1)
Get (22050, -16, 1)
Avg 32767.000000 Amp 32767.000000 Step 0.142800
Size 16 Itemsize 2
Set for play (22050, 16, 1)
Get for play (22050, -16, 1)
array('H', [32767, 37430, 41999, 46379, 50482, 54225, 57531, 60332, 62573, 64207, 65200, 65534, 65200, 64207, 62573, 60332, 57531, 54225, 50482, 46379, 41999, 37430, 32767, 28104, 23535, 19155, 15052, 11309, 8003, 5202, 2961, 1327, 334, 0, 334, 1327, 2961, 5202, 8003, 11309, 15052, 19155, 23535, 28104])
Stop?
>>> 

As you can see when I set  size=-8 and I get size=+8, and when I set size=16 I get size=-16.
Also when I look the signal coming from my audio jack using my oscilloscope, the signal is fine when the set value==get value, but when it is not equal, the signal is not good.

I think your program is trying to tell me that pygame.mixer is not supporting -8 and +16 size format. and you program is setting the best alternative value.  This behaivour is not documented anywhere.  I'm not sure if it is a bug or a if it is expected.

Richard

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/292

Issue #113, opened 2012-03-17
support surround sound (aka 5.1 / aka 3D sound / aka multi-channels)


illume:
Originally reported by: **pierriko (Bitbucket: [pierriko](https://bitbucket.org/pierriko), GitHub: [pierriko](https://github.com/pierriko))**

----------------------------------------

I found 2 source mentioning surround sound:

  * http://hg.pygame.org/pygame/src/d1cbb8c9d94b/test/playwave.py#cl-181
  * http://hg.pygame.org/pygame/src/9c6aa550da25/doc/src/openal_constants.rst#cl-69

but the mixer's doc says http://pygame.org/docs/ref/mixer.html :

//The channels argument is used to specify whether to use mono or stereo. 1 for mono and 2 for stereo. No other values are supported (negative values are treated as 1, values greater than 2 as 2).//

I'm using: http://packages.ubuntu.com/natty/python-pygame 

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/113

PR #861, opened 2019-02-28
SDL2: fix mixer errors, mixer.init(..., allowchanges=0)

### LINKED ISSUE #860: SDL2: mixer test failures
```
FAIL: test_array_interface (test.mixer_test.MixerModuleTest)
  File "C:\Users\David\Documents\dev\pygame\test\mixer_test.py", line 323, in test_array_interface
    self.assertEqual(d['shape'], (20,))
AssertionError: Tuples differ: (10, 2) != (20,)

First differing element 0:
10
20

First tuple contains 1 additional elements.
First extra element 1:
2

- (10, 2)
+ (20,)

FAIL: test_init__zero_values (test.mixer_test.MixerModuleTest)
  File "C:\Users\David\Documents\dev\pygame\test\mixer_test.py", line 83, in test_init__zero_values
    self.assertEqual(mixer.get_init(), (44100, 8, 1))
AssertionError: Tuples differ: (44100, 8, 2) != (44100, 8, 1)

First differing element 2:
2
1

- (44100, 8, 2)
?            ^

+ (44100, 8, 1)
?            ^


Ran 26 tests in 8.135s

FAILED (failures=2)
```

dlon:
Close #860


illume:
Ah. SDL 2 is required for the allowchanges stuff to work.
https://wiki.libsdl.org/SDL_OpenAudioDevice#allowed

- SDL_AUDIO_ALLOW_FREQUENCY_CHANGE
- SDL_AUDIO_ALLOW_FORMAT_CHANGE
- SDL_AUDIO_ALLOW_CHANNELS_CHANGE
- SDL_AUDIO_ALLOW_ANY_CHANGE

With SDL1 (and pygame), I think it defaulted to SDL_AUDIO_ALLOW_ANY_CHANGE behaviour.

src_c/mixer.c
```diff
@@ -343,20 +345,22 @@ _init(int freq, int size, int stereo, int chunk, char *devicename)
     if (!size) {
         size = request_size;
     }
-    if (!stereo) {
-        stereo = request_stereo;
+
+    if (!channels) {
+        channels = request_channels;
     }
+    channels = (channels >= 2) ? 2 : 1;
```
illume: I wonder if we should let `channels` to go through as a number more than 2? I guess none of the code is prepared for this really (sound array etc), and it would require testing. But SDL2 accepts more than 2.

https://wiki.libsdl.org/SDL_AudioSpec#Remarks
```
channels specifies the number of output channels. As of SDL 2.0, supported values are 1 (mono), 2 (stereo), 4 (quad), and 6 (5.1). 
```

illume: So, I guess this `channels >= 2` test should only be for SDL1.
For SDL2 it should check for 1, 2, 4, and 6.

dlon:
allowedchanges=0 doesn't cause it to fail. SDL converts the audio to and
from the requested format instead of using a similar one (in the
callbacks). Seems to be the default behavior in SDL 1

Den fre 1 mars 2019 06:14René Dudfield <notifications@github.com> skrev:

src_c/mixer.c
```diff
@@ -349,7 +349,27 @@ _init(int freq, int size, int channels, int chunk, char *devicename, int allowed
...
         channels = request_channels;
     }
-    channels = (channels >= 2) ? 2 : 1;
+#if IS_SDLv1
+    switch (channels) {
+        case 1:
+        case 2:
+            break;
+        default:
+            PyErr_SetString(PyExc_ValueError, "'channels' must be 1 (mono) or 2 (stereo)");
```
dlon: I made it check the arguments more strictly here. Do you think anyone relied on the old behavior?

dlon:
This also solved the issues in #850.

src_c/constants.c
```diff
@@ -317,6 +317,11 @@ MODINIT_DEFINE(constants)
     //Still to be decided
     DEC_CONSTS(BUTTON_WHEELUP, PGM_BUTTON_WHEELUP);
     DEC_CONSTS(BUTTON_WHEELDOWN, PGM_BUTTON_WHEELDOWN);
+    DEC_CONST(AUDIO_ALLOW_FREQUENCY_CHANGE);
+    DEC_CONST(AUDIO_ALLOW_FORMAT_CHANGE);
+    DEC_CONST(AUDIO_ALLOW_CHANNELS_CHANGE);
+    DEC_CONST(AUDIO_ALLOW_ANY_CHANGE);
```
charlesej: Do these constants need to be set to their corresponding SDL constants?

```
SDL_AUDIO_ALLOW_FREQUENCY_CHANGE
SDL_AUDIO_ALLOW_FORMAT_CHANGE
SDL_AUDIO_ALLOW_CHANNELS_CHANGE
SDL_AUDIO_ALLOW_ANY_CHANGE
```

src_c/mixer.c
```diff
@@ -343,20 +345,42 @@ _init(int freq, int size, int stereo, int chunk, char *devicename)
...
     if (!devicename) {
         devicename = request_devicename;
     }
-    if (stereo >= 2)
-        stereo = 2;
-    else
-        stereo = 1;
+    if (allowedchanges == -1) {
+        allowedchanges = request_allowedchanges;
+    }
```
charlesej: Should there be a check/conversion from `PYGAME_MIXER_DEFAULT_ALLOWEDCHANGES` to `AUDIO_ALLOW_ANY_CHANGE`/`SDL_AUDIO_ALLOW_ANY_CHANGE` here?

src_c/constants.c
dlon: The macro function does that:
```
#define DEC_CONSTS_(x, y)                           \
    if (PyModule_AddIntConstant(module, x, (int)y)) \
    ADD_ERROR
#define DEC_CONST(x) DEC_CONSTS_(#x, SDL_##x)
```

src_c/mixer.c
dlon: `PYGAME_MIXER_DEFAULT_ALLOWEDCHANGES` should be set to 0 or any combination of those flags, so it seems right

src_c/mixer.c
dlon: On the other hand, ``allowedchanges=0`` makes you expect no changes to the requested options

dlon: Specifically, unless you set `SDL_AUDIO_ALLOW_CHANNELS_CHANGE` (or any combination including it, like `SDL_AUDIO_ALLOW_ANY_CHANGE`), I think it should fail if the value isn't valid.

src_c/mixer.c
charlesej: You are right. For some reason I thought 0 wasn't an acceptable value because there was no SDL define for it, but it just means no changes allowed.

src_c/constants.c
charlesej: Ok, so it is being done.

src_c/mixer.c
charlesej: My understanding is that with SDL1.2 there is no way to enforce the no changes on the audio arguments (issue #292 is related to this regarding `format` values).

[Mix_QuerySpec](https://www.libsdl.org/projects/SDL_mixer/docs/SDL_mixer_15.html#SEC15) does not guarantee the values you get will be the ones you passed into [Mix_OpenAudio](https://www.libsdl.org/projects/SDL_mixer/docs/SDL_mixer_11.html).

> Get the actual audio format in use by the opened audio device. This may or may not match the parameters you passed to Mix_OpenAudio.

Instead of `0` as the default for `allowedchanges` maybe it should be `AUDIO_ALLOW_ANY_CHANGE` (with a note stating that is the only supported value in SDL1.2). Then the SDL1.2 code simple becomes `channels = (channels >= 2) ? 2 : 1;` (as you originally had it).

dlon:
From the [SDL 1.2.15 docs](https://www.libsdl.org/release/SDL-1.2.15/docs/html/sdlopenaudio.html):
> int SDL_OpenAudio(SDL_AudioSpec *desired, SDL_AudioSpec *obtained);

> This function opens the audio device with the desired parameters, and returns 0 if successful, placing the actual hardware parameters in the structure pointed to by obtained. If obtained is NULL, the audio data passed to the callback function will be guaranteed to be in the requested format, and will be automatically converted to the hardware audio format if necessary.

From SDL_mixer (1.*):

```c
/* Open the mixer with a certain desired audio format */
int Mix_OpenAudio(int frequency, Uint16 format, int nchannels, int chunksize)
{
	int i;
	SDL_AudioSpec desired;

...

	/* Set the desired format and frequency */
	desired.freq = frequency;
	desired.format = format;
	desired.channels = nchannels;
	desired.samples = chunksize;
	desired.callback = mix_channels;
	desired.userdata = NULL;

	/* Accept nearly any audio format */
	if ( SDL_OpenAudio(&desired, &mixer) < 0 ) {
		return(-1);
	}
...
}

*obtained* is not NULL (it's &mixer), so I guess allowedchanges=0 is the wrong default. But changing it causes a bunch of test failures, so I guess we update the tests instead.

File: src_c/mixer.c [hunk 5]
```diff
@@ -346,28 +351,36 @@ _init(int freq, int size, int channels, int chunk, char *devicename, int allowed
...
+            channels = 6;
+    } else {
+        switch (channels) {
+            case 1:
+            case 2:
+            case 4:
+            case 6:
+                break;
+            default:
+                PyErr_SetString(PyExc_ValueError, "'channels' must be 1, 2, 4, or 6");
```
husano896: #113 can be closed since channels > 2 are supported in SDL2.

Issue #864, opened 2019-03-01
SDL1-SDL2 blit discrepancy


dlon:
This outputs a slightly different color with SDL2 than with SDL1:

```python3
import pygame

pygame.display.init()

fg_color = (128, 97, 213, 60)

surf = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
surf.fill((0, 0, 0, 255))

fg = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
fg.set_at((0, 0), fg_color)

surf.blit(fg, (0, 0))

print("Color:", surf.get_at((0, 0)))
```

SDL2 output: (30, 22, 49, 253)
SDL1 output: (30, 23, 50, 255)

The color for SDL2 seems to be slightly off:
128\*60/255 = 30
97\*60/255 = 23
213\*60/255 = 50

**Related Docs:** https://www.pygame.org/docs/ref/surface.html#pygame.Surface.blit

imallett:
The less-rounded version is:
<pre>128 * 60/255 ≈ 30.12
 97 * 60/255 ≈ 22.82
213 * 60/255 ≈ 50.12</pre>

You can get the SDL1 results by rounding, and the SDL2 results by having the alpha be 60/256 and truncating. I don't know how they're actually computed, but I concur that the SDL1 results seem preferable.

Of-course, the "correct" thing would be to use sRGB+A or something:
<pre>lrgb_to_srgb( srgb_to_lrgb((128,97,213)) * (60/255) )
≈ ( 63.70, 46.73, 110.2 )</pre>

dlon:
Ah, that makes sense. I think it might be due to this code:

https://hg.libsdl.org/SDL/file/25998acc4810/src/video/SDL_blit_A.c#l479

MyreMylar:
May be related to this:

https://bugzilla.libsdl.org/show_bug.cgi?id=4761

MyreMylar:
Output now in dev22:

```
pygame 2.0.0.dev22 (SDL 2.0.12, python 3.8.5)
Hello from the pygame community. https://www.pygame.org/contribute.html
Color: (30, 23, 50, 255)
```

This is the same as SDL1 again. Closing.

Issue #859, opened 2019-02-28
SDL2: freetype test failures


dlon:
```
..............F...F.........F..............
======================================================================
FAIL: test_freetype_Font_render_mono (test.freetype_test.FreeTypeFontTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\David\Documents\dev\pygame\test\freetype_test.py", line 744, in test_freetype_Font_render_mono
    self.assertEqual(surf.get_alpha(), translucent_color.a)
AssertionError: None != 55

======================================================================
FAIL: test_freetype_Font_render_to_mono (test.freetype_test.FreeTypeFontTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\David\Documents\dev\pygame\test\freetype_test.py", line 804, in test_freetype_Font_render_to_mono
    self.assertEqual(surf.get_at(bottomright), r_fg_color)
AssertionError: (30, 23, 50, 255) != (30, 22, 49, 253)

======================================================================
FAIL: test_issue_144 (test.freetype_test.FreeTypeFontTest)
Issue #144: unable to render text
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\David\Documents\dev\pygame\test\freetype_test.py", line 1273, in test_issue_144
    self.assertRaises(pygame.error, font.render, s, (0, 0, 0))
AssertionError: error not raised by render

----------------------------------------------------------------------
Ran 43 tests in 3.325s

FAILED (failures=3)
```

**Related Docs:** 

https://www.pygame.org/docs/ref/freetype.html
https://www.pygame.org/docs/ref/surface.html#pygame.Surface.blit

dlon:
The issues are fixed but one test is still skipped due to #864.

illume:
@zoldalma999 removed a skip in #2351 because it now passes the test.

> Since Freetype tests dont run on CI

So probably they need to get added to CI.

These are the current failing test/freetype_test.py tests for me...
```
$ python test/freetype_test.py
pygame 2.0.1.dev1 (SDL 2.0.12, python 2.7.16)
Hello from the pygame community. https://www.pygame.org/contribute.html
.....................s.......E.Fs...............
======================================================================
ERROR: test_freetype_SysFont_name (__main__.FreeTypeFontTest)
that SysFont accepts names of various types
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #865, opened 2019-03-01
SDL2: freetype issues

dlon:
* Fix one freetype issue.

* Skip one test because it no longer applies (there's no hardcoded size-limit to surfaces anymore, it seems).

* Skip another test (for now) because it fails for an unrelated reason (#864).

Related: #859


illume:
Great :)

Issue #841, opened 2019-02-23
Calling pygame.mask.Mask.connected_component can result in a segmentation fault


charlesej:
Calling `pygame.mask.Mask.connected_component` with an index outside the mask's bounds can result in a segmentation fault.

Expected behavior: Raise an `IndexError` (like the `get_at`/`set_at` methods).
Actual behavior: Segmentation fault or silently accepting an out of bounds index.

This code shows both behaviors: silently accepting an out of bounds index (`mask1`) and a segmentation fault (`mask2`).
```
# Filename: mask_cc.py
import pygame

mask1 = pygame.mask.Mask((10, 10))
mask2 = pygame.mask.Mask((10, 11))

mask1.connected_component((0, -1))
mask2.connected_component((0, -1))  # Causes segmentation fault.
```

Code output using python 3.7.2.
```
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
Fatal Python error: (pygame parachute) Segmentation Fault

Current thread 0x00003e24 (most recent call first):
  File "mask_cc.py", line 8 in <module>
```

Code output using python 2.7.10.
```
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
Fatal Python error: (pygame parachute) Segmentation Fault

This application has requested the Runtime to terminate it in an unusual way.
Please contact the application's support team for more information.
```

System details:
* os: windows 10 (64bit)
* python: 3.7.2 (64bit) and 2.7.10 (64bit)
* pygame: 1.9.4 and 1.9.5.dev0 (SDL: 1.2.15) at 3a84be5e0e4b8ac9fecc130913f1d68e7484db55

PR #873, opened 2019-03-02
Fix mask index out of bounds segmentation fault

charlesej:
This update fixes the segmentation fault caused by calling `pygame.mask.Mask.connected_component` with an index outside the mask's bounds.

Overview of changes:
- `IndexError` raised when an index is outside of the mask's bounds
- Fixed some memory leaks on the fail paths
- Fixed an incorrect exception message
- Removed `@unittest.skip` decorator from the related test method and fixed a variable name error

System details:
- os: windows 10 (64bit)
- pygame: 1.9.4 and 1.9.5.dev0 (SDL: 1.2.15) at 1a4c7897f577ff89495e557e13f14968437c3420

### DIFF
--- a/src_c/mask.c
+++ b/src_c/mask.c
@@ -1443,26 +1443,39 @@ static PyObject *
 mask_connected_component(PyObject *self, PyObject *args)
 {
     bitmask_t *input = pgMask_AsBitmap(self);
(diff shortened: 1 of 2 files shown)


illume:
Nice one :) This closes: #841

PR #874, opened 2019-03-02
Add mask tests - part 4

charlesej:
This update adds some mask test methods.

Overview of changes:
- Added some mask test methods
- Moved `test_scale__negative_size` to be co-located with the other scale test

System details:
- os: windows 10 (64bit)
- pygame: 1.9.5.dev0 (SDL: 1.2.15) at 1a4c7897f577ff89495e557e13f14968437c3420

Partially resolves the "Add tests for some of the Mask class methods that are not being tested." item of #800.

### DIFF
--- a/test/mask_test.py
+++ b/test/mask_test.py
@@ -235,9 +235,92 @@ class MaskTypeTest(unittest.TestCase):
 
         self.fail()
 
-    def todo_test_overlap__offset(self):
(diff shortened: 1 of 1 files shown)


illume:
Very good :) The mask module is getting quite a lot better thanks to all this work you're doing.

PR #882, opened 2019-03-05
WIP: _sdl2.Texture & Renderer

husano896:
get_rect() & copy() for Texture, 

blit() for Renderer

Texture.get_rect(**kwargs)

>  Get the rectangular area of the texture.
        like surface.get_rect(), returns a new rectangle covering the entire surface. 
        This rectangle will always start at 0, 0 with a width. and height the same size as the texture.
        Passing arguments like (center=(100,100)) also works

Texture.copy(dstrect=None, srcrect=None)
> Like Renderer.copy(), but takes 1st argument as destination rect.
       Will be rendered by the renderer that creates the texture.

Renderer.blit(source, dest=None, area=None, special_flags = 0)
> Accepts Textures and Surfaces as source, and allows dest to be None(stretch to fullscreen)
       special_flags take no effect but for compatibility.
       using renderer for sprite.Group().draw(renderer) will work now.

#837

Issue #837, opened 2019-02-19
pygame.sprite vs Texture


husano896:
pygame.sprite only supports pygame.Surface at this moment.

Which is better ?

1. Add Texture support in pygame.sprite
2. Create a new class like pygame.texsprite (pygame._sdl2.sprite ?) and try to keep the same behavior

illume:
There was a little bit of discussion in a separate PR about Texture... https://github.com/pygame/pygame/pull/708

@llindstrom wrote:
> I did some experimenting with a [Cython view module](https://github.com/llindstrom/pygame_SDL2/blob/SDL2_sandbox/src/pygame.video.pyx) for pygame. It is somewhat out of data as it pre-dates the updates to the pygame C api. My approach to Texture creation was to have Renderer methods return new, open, Texture instances. The Texture class itself returned a closed, unusabled instance. An example using the experimental view module to create a surface texture is [here](https://github.com/llindstrom/pygame_SDL2/blob/SDL2_sandbox/examples/window_render.py).

illume:
@husano896 interesting question.

Why can't Texture implement the `.blit()` and `.get_rect()` used by the sprite module (and other pygame code)?

Perhaps it is worth thinking about how it could be made compatible enough so that such code would work. If we special case the 'screen' variable, where instead of being a screen surface, it is instead a renderer perhaps?

husano896:
Texture.blit() or Renderer.blit()?
Rendering to texture will have to set the target of Renderer to texture, and flags must be set.

Looks a little complex for me :(

dlon:
Should we have something like a subtexture class? Using a single texture for everything sort of defeats the purpose of batching

```python3
class TextureRegion:
	region (rect)
	texture

	def __init__(Texture, rect)
	def copy(self, dstrect=None, srcrect=None)
```

dlon:
There could be a third class for the attributes:

```
class Texture:
	def draw(self, srcrect=None, dstrect=None, angle=0, origin=None, flipX=False, flipY=False)

class TextureRegion:
	rect
	texture

	def __init__(self, Texture, srcrect=None)
	def draw(self, srcrect=None, dstrect=None, angle=0, origin=None, flipX=False, flipY=False)

class Sprite:
	tex
	color
	angle
	origin
	flipX, flipY
	rect

	def __init__(self, Texture tex)
		self.tex = tex
	def __init__(self, TextureRegion tex)
		self.tex = tex

	def draw(self)
[...]

PR #883, opened 2019-03-05
SDL2: _sdl2.Texture, Image & Renderer

dlon:
Continuing #882 on a shared branch.

* Texture.draw() instead of Texture.copy()
* New features: Texture.draw(self, dstrect=None, srcrect=None, angle=0, origin=None, bint flipX=False, bint flipY=False)
* Image(tex, srcrect=None). Defines a subarea of a Texture, and contains attributes angle, origin, flipX/flipY, etc.

Re: #837


src_c/_sdl2/video.pyx
```diff
@@ -764,3 +893,47 @@ cdef class Renderer:
...
+        :param source: A Texture or Surface.
+                        Surface will be converted by SDL_CreateTextureFromSurface.
+        :param area: Rect only, the portion of source texture.
+        :param special_flags: have no effect at this moment.
+        """
+        cdef SDL_Texture *tex
+        cdef Texture pyTex
+        if pgSurface_Check(source):
+            #Not recommended, creating texture every frame is expensive.
+            tex = SDL_CreateTextureFromSurface(self._renderer,
```
dlon: @husano896: If this has to be supported, maybe something like this?
```
__init__(self):   self.surfaceToTexture = {}


tex = self.surfaceToTexture.get(source)
if not tex:
    tex = SDL_CreateTextureFromSurface(self._renderer, pgSurface_AsSurface(source))
    self.surfaceToTexture[source] = tex
```

husano896: @dlon Do you mean caching Surfaces as Textures?

if that case, should we limit the cache size (if len(self.surfaceToTexture) > MAX_CACHE_SIZE)?

dlon: Yes, but maybe it should just raise an exception instead? Otherwise it either leaks memory or is very slow. I don't think it has to accept a Surface

husano896: Maybe accepting surfaces is not a good idea, since Renderer is a new thing compared to SDL1.

My first thought was: if someone add Surfaces and Textures to a sprite Group, and uses Renderer as target, but not a problem if Surfaces are converted at first.

dlon:
<details><summary>Image and pygame.sprite.Sprite example code</summary>

```python3
import pygame


if pygame.get_sdl_version()[0] < 2:
    raise SystemExit('This example requires pygame 2 and SDL2.')

import os
data_dir = os.path.join(os.path.split(os.path.abspath(__file__))[0],
                        'data')

from pygame._sdl2 import (
    Window,
    Texture,
    Image,
    Renderer
)

def load_img(file):
    return pygame.image.load(os.path.join(data_dir, file))

pygame.display.init()
pygame.key.set_repeat(10, 10)

win = Window('asdf', resizable=True)
renderer = Renderer(win)
tex = Texture.from_surface(renderer, load_img('alien1.gif'))

class Something(pygame.sprite.Sprite):
    def __init__(self, img):
        super().__init__()
        self.rect = img.get_rect()
        self.image = img

        self.rect.w *= 5
        self.rect.h *= 5

        img.origin = self.rect.w / 2, self.rect.h / 2

sprite = Something(Image(tex, (0, 0, tex.width/2, tex.height/2)))
sprite.rect.x = 250
sprite.rect.y = 50

#sprite2 = Something(Image(sprite.image))
sprite2 = Something(Image(tex))
sprite2.rect.x = 250
sprite2.rect.y = 250
sprite2.rect.w /= 2
sprite2.rect.h /= 2

group = pygame.sprite.Group()
group.add(sprite2)
group.add(sprite)

import math
t = 0
running = True
clock = pygame.time.Clock()
renderer.draw_color = (255, 0, 0, 255)

while running:
        if event.type == pygame.QUIT:
            running = False
            if event.key == pygame.K_ESCAPE:
                running = False
            elif event.key == pygame.K_LEFT:
                sprite.rect.x -= 5
            elif event.key == pygame.K_RIGHT:
                sprite.rect.x += 5
            elif event.key == pygame.K_DOWN:
                sprite.rect.y += 5
            elif event.key == pygame.K_UP:
                sprite.rect.y -= 5

    renderer.clear()
    t += 1

    img = sprite.image
    img.angle += 1
    img.flipX = t % 50 < 25
    img.flipY = t % 100 < 50
    img.color[0] = int(255.0 * (.5 + math.sin(.5 * t + 10.0) / 2.0))
    img.alpha = int(255.0 * (.5 + math.sin(.1 * t) / 2.0))
    #img.draw(dstrect=(x, y, 5 * img.srcrect['w'], 5 * img.srcrect['h']))

    group.draw(renderer)

    renderer.present()

    clock.tick(60)
    win.title = str('FPS: {}'.format(clock.get_fps()))
```

</details>

<br />

Image interface:

```
cdef class Image:
    cdef public float angle
    cdef public float origin[2]
    cdef public bint flipX
    cdef public bint flipY
    cdef public object color
    cdef public float alpha

    cdef public Texture texture
    cdef public SDL_Rect srcrect

    def __init__(self, textureOrImage, srcrect=None)
    def get_rect(self)

    def draw(self, srcrect=None, dstrect=None):
        """ Copy a portion of the image to the rendering target.

        :param srcrect: source rectangle specifying a sub-image, or None for the entire image.
        :param dstrect: destination rectangle on the render target, or None for entire target.
                        The image is stretched to fill dstrect.
        """
```

illume:
- fixed a few little issues (I hope correctly!) and added the example from dlon.
- we can merge this in but still change the APIs around. It will be good to test this out on different computers/platforms and get feedback from others on the API.

PR #899, opened 2019-03-11
bugfix for dirty sprite when using a source rect

dr0id:
fixes the bug reported by Inigo González <inigoini@gmail.com>:

[pygame] BUG: sprite refresh when using source_rect

from 2019.03.10

### DIFF
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -245,7 +245,7 @@ class DirtySprite(Sprite):
         the screen.)
 
     _layer = 0
-        0 is the default value but this is able to be set differently 
+        0 is the default value but this is able to be set differently
         when subclassing.
 
     """
@@ -256,7 +256,7 @@ class DirtySprite(Sprite):
         self.blendmode = 0  # pygame 1.8, referred to as special_flags in
                             # the documentation of Surface.blit
         self._visible = 1
-        self._layer = getattr(self, '_layer', 0)    # Default 0 unless 
+        self._layer = getattr(self, '_layer', 0)    # Default 0 unless
                                                     # initialized differently.
         self.source_rect = None
         Sprite.__init__(self, *groups)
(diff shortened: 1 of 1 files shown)


src_py/sprite.py
```diff
@@ -1099,17 +1099,14 @@ def draw(self, surface, bgd=None):
...
-                                             spr.source_rect.size)
                         _spr_rect_clip = _spr_rect.clip
                         for idx in _spr_rect.collidelistall(_update):
                             # clip
                             clip = _spr_rect_clip(_update[idx])
                             _surf_blit(spr.image,
                                        clip,
-                                       (clip[0] - _spr_rect[0],
-                                        clip[1] - _spr_rect[1],
+                                       (clip[0] - _spr_rect[0] + spr.source_rect[0],
```
charlesej: The default value for `source_rect` is `None`, so this will crash when a `source_rect` is not used.

src_py/sprite.py
```diff
@@ -1099,17 +1099,14 @@ def draw(self, surface, bgd=None):
                     if spr._visible:
                         # sprite not dirty; blit only the intersecting part
                         _spr_rect = spr.rect
-                        if spr.source_rect is not None:
```
charlesej: I think a `DirtySprite`'s `rect` and `source_rect` can be different sizes so this code is probably still required to ensure the proper image is drawn.

src_py/sprite.py
dr0id: Thanks for the review and feedback.

You are right, the source_rect can be None. I have considered it now in the second commit.

About the different sizes... it does not make much sense because its the source area of the blit method (not sure what will happen if you have a bigger source rect...... for a smaller source rect I guess you only the that part).

I still wonder if the clipping will work in all cases.

src_py/sprite.py
charlesej: My understanding is that `source_rect` is used to get the correct image. If you look above at lines 1063-1067 you will see that a similar thing is being done there.

dlon:
The test in #900 still seems to fail

charlesej:
The test is still failing because this update is removing code (lines [1102-1104](https://github.com/pygame/pygame/blob/03ffd527263cd86bfe0cc95c70dcf8b11c6ba1a5/src_py/sprite.py#L1102-L1104) in the original code) that is unrelated to fixing this issue. The removal of those lines breaks current usage and is inconsistent with other parts of the code (lines [1063-1067](https://github.com/pygame/pygame/blob/28e246017808334e8a7cc3bd03dbc8e3cd426f9e/src_py/sprite.py#L1063-L1067): `source_rect` usage).

The originator did a good job of fixing the original issue, but making an unrelated functionality breaking change (granted this is a minor one) would be best done via a separate update. It would also be useful to include some related tests and update the documentation stating that this functionality has been changed and may break existing code.

illume:
Continuing this PR in a pygame repo branch so we can all collaborate more easily...
https://github.com/pygame/pygame/pull/919

Issue #711, opened 2019-01-03
SDL2: game controller


illume:
Various SDL_GameController functions need to be implemented in _sdl2.
https://wiki.libsdl.org/CategoryGameController

husano896:
> Better interface for controller mapping in pygame (ask @husano896 what this is about)

https://wiki.libsdl.org/SDL_GameControllerAddMapping

At this moment, mapping uses a **really long** string as SDL2 like 

_"341a3608000000000000504944564944,Afterglow PS3 Controller,a:b1,b:b2,y:b3,x:b0,start:b9,guide:b12,back:b8,dpup:h0.1,dpleft:h0.8,dpdown:h0.4,dpright:h0.2,leftshoulder:b4,rightshoulder:b5,leftstick:b10,rightstick:b11,leftx:a0,lefty:a1,rightx:a2,righty:a3,lefttrigger:b6,righttrigger:b7"_

What I think is: 

since we can easily get guid and controller name, add mapping should use a easier parameter like 

    mapping = {
     "a" : "b1",
     "b" : "b2",
     "y" :  "b3",
     "x" : "b0"
     # etc...
    }
    controller.add_mapping(mapping)

and so as get_mapping.

    mapping = controller.get_mapping() # return a dict

PR #901, opened 2019-03-12
SDL2: starting _sdl2.controller API.

husano896:
Re: #711 

Bug:
- [x] For some reason controller events is not present.

Improve:
- [ ] Tests for controller are needed
- [ ] CONTROLLER_* consts look a bit long,  shorter names are needed I think?
- [ ] Better interface for controller mapping in pygame
- [ ] Controller docs

Controller constructors:

```
class Controller:
    def __init__(self, int index)
    @staticmethod
    def from_joystick(joy)
```

added init() quit() get_init() for both controller and controller.Controller


src_c/event.c
```diff
@@ -632,6 +644,24 @@ dict_from_event(SDL_Event *event)
             _pg_insobj(dict, "start", PyLong_FromLong(event->edit.start));
             _pg_insobj(dict, "length", PyLong_FromLong(event->edit.length));
             break;
+        case SDL_CONTROLLERAXISMOTION:
+            /* https://wiki.libsdl.org/SDL_ControllerAxisEvent */
+            _pg_insobj(dict, "joy", PyInt_FromLong(event->caxis.which));
```
dlon: Should probably use PyLong_FromLong() to guarantee that it may contain at least 32 bits.

src_c/event.c
```diff
@@ -632,6 +644,24 @@ dict_from_event(SDL_Event *event)
...
+            _pg_insobj(dict, "axis", PyLong_FromLong(event->caxis.axis));
+            _pg_insobj(dict, "value", PyLong_FromLong(event->caxis.value));
+            break;
+        case SDL_CONTROLLERBUTTONDOWN:
+        case SDL_CONTROLLERBUTTONUP:
+            /* https://wiki.libsdl.org/SDL_ControllerButtonEvent */
+            _pg_insobj(dict, "joy", PyInt_FromLong(event->cbutton.which));
```
dlon: Should probably use PyLong_FromLong() to guarantee that it may contain at least 32 bits.

src_c/event.c
```diff
@@ -632,6 +644,24 @@ dict_from_event(SDL_Event *event)
...
+            _pg_insobj(dict, "button", PyLong_FromLong(event->cbutton.button));
+            break;
+        case SDL_CONTROLLERDEVICEADDED:
+        case SDL_CONTROLLERDEVICEREMOVED:
+        case SDL_CONTROLLERDEVICEREMAPPED:
+            /* https://wiki.libsdl.org/SDL_ControllerDeviceEvent */
+            _pg_insobj(dict, "joy", PyInt_FromLong(event->cdevice.which));
```
dlon: Should probably use PyLong_FromLong() to guarantee that it may contain at least 32 bits.

src_c/constants.c
```diff
@@ -334,12 +369,46 @@ MODINIT_DEFINE(constants)
...
+	DEC_CONSTS(CONTROLLERDEVICEREMAPPED, -1);
+	
     DEC_CONST(BUTTON_WHEELUP);
     DEC_CONST(BUTTON_WHEELDOWN);
     DEC_CONSTS(AUDIO_ALLOW_FREQUENCY_CHANGE, -1);
     DEC_CONSTS(AUDIO_ALLOW_FORMAT_CHANGE, -1);
     DEC_CONSTS(AUDIO_ALLOW_CHANNELS_CHANGE, -1);
     DEC_CONSTS(AUDIO_ALLOW_ANY_CHANGE, -1);
+	
+	DEC_CONSTS(CONTROLLER_AXIS_INVALID, -1);
```
dlon: Using tabs instead of spaces here

src_c/_sdl2/controller.pyx
```diff
@@ -0,0 +1,129 @@
...
+    @staticmethod
+    def from_joystick(joy):
+        # https://wiki.libsdl.org/SDL_GameControllerFromInstanceID
+        """ Create a controller object from pygame.joystick.Joystick object.
+        
+        """
+        if not pgJoystick_Check(joy):
+            raise TypeError('should be a pygame.joystick.Joystick object.')
+
+        cdef Controller self = Controller.__new__(Controller)
```
dlon: Does `Controller(joy.get_id())` work?

src_c/_sdl2/controller.pyx
```diff
@@ -0,0 +1,129 @@
...
+    """
+    return SDL_IsGameController(index)
+
+def name_forindex(index):
+    """ Returns the name of controller,
+        or NULL if there's no name or the index is invalid.
+    """
+    return SDL_GameControllerNameForIndex(index)
+
+cdef class Controller:
```
dlon: I think you forgot to close it in a __dealloc__ method.

src_c/_sdl2/controller.pyx
```diff
@@ -0,0 +1,129 @@
...
+        
+    def get_button(self, SDL_GameControllerButton button):
+        # https://wiki.libsdl.org/SDL_GameControllerGetButton
+        return SDL_GameControllerGetButton(self._controller, button)
+        
+    def get_mapping(self):
+        #https://wiki.libsdl.org/SDL_GameControllerMapping
+        # TODO: mapping should be a readable dict instead of a string.
+        mapping = SDL_GameControllerMapping(self._controller)
+        SDL_free(mapping)
```
dlon: This seems wrong. Either it's trying to free a Python string, or it may be using a string after it's freed. Maybe do this (untested):

```cython
mapping = SDL_GameControllerMapping(self._controller)
try:
    return mapping.decode('utf8')
finally:
    SDL_free(mapping)
```

src_c/_sdl2/controller.pyx
husano896: For some reason,

```
def __dealloc__(self):
        if self._controller:
            SDL_GameControllerClose(self._controller)
```

caused weird behavior on exit(no error in Python Shell, but looks like some error occurred running cython code.

but closing controllers anytime before exit runs fine.

src_c/_sdl2/controller.pyx
husano896: joy.get_id() means joystick.Joystick.get_id, which is also a number from 0 to joystick.get_count()

src_c/_sdl2/controller.pyx
husano896: this causes unreachable code while compiling

I wonder if this works:
Convert it to a Python string and free the original mapping pointer.

        cdef char* mapping = SDL_GameControllerMapping(self._controller)
        
        try:
            #Convert the string to Python string?
            pyMapping = mapping.decode('UTF-8')
        except:
            #Returns bytes if decoding failed.
            pyMapping = mapping
        finally:
            SDL_free(mapping)
	        
        return pyMapping

illume:
I'm going to merge this one in. Further work can happen in other pull requests.

illume:
I added the checklist from here to https://github.com/pygame/pygame/issues/711

Issue #524, opened 2018-09-06
SDL2 display.c pygame.display.Info vidinfo for SDL2


illume:
https://www.pygame.org/docs/ref/display.html#pygame.display.Info

- [ ] a unit test calling it in test/display_test.py
- [ ] src_c/display.c is where it should be implemented.

markph0204:
I'm unsure what the issue is... with py195dev;sdl2;py37 I'm getting back...

```
<VideoInfo(hw = 0, wm = 1,video_mem = 0
         blit_hw = 0, blit_hw_CC = 0, blit_hw_A = 0,
         blit_sw = 0, blit_sw_CC = 0, blit_sw_A = 0,
         bitsize  = 32, bytesize = 4,
         masks =  (65280, 16711680, -16777216, 0),
         shifts = (8, 16, 24, 0),
         losses =  (0, 0, 0, 8),
         current_w = 320, current_h = 240
>
```

markph0204:
Seems like I need to take this back... latest 195dev is not supporting this.  Curious if recent change removed it? Or am I losing my mind switching between environments?

dlon:
> It may be worth considering the retirement of this. If you decide to introduce the Renderer, then it - itself has [info](https://wiki.libsdl.org/SDL_RendererInfo). The items found under renderer could be used here, for example: SDL_RENDERER_SOFTWARE, SDL_RENDERER_ACCELERATED. BUT would it further complicate the new Render object you might introduce and should belong there?

That already exists FWIW:
https://github.com/pygame/pygame/blob/15b1c1fd97682f5cdcbe6c24d55ab0c829a5c743/src_c/_sdl2/video.pyx#L37-L49

... But I'll create this thing quickly for backward compatibility. IIRC, no blitting is "hardware accelerated" anymore, so most of those fields should just be zero.

PR #908, opened 2019-03-13
SDL2: Implement display.Info()

dlon:
Re: #524.

Potential issues:
* Hardcoding wm_available = 1.
* Amount of video memory is always set to 0.


illume:
Nice one :)

The appveyor failure is unrelated to this PR.
The video memory is zero if unknown, as according to the docs.

PR #909, opened 2019-03-13
Update Surface docs.

dlon:
`surface.set_alpha()` can be used in conjunction with per-pixel alpha now.


illume:
Oh, that's pretty cool.

PR #910, opened 2019-03-13
WIP: Dollar gestures

dlon:
See #814.

There are two events used for single-stroke gestures: DOLLARRECORD and DOLLAREVENT. Every gesture is associated with a gesture id when recorded.

The events look like this:

    DOLLARRECORD: new Gesture object
    DOLLARGESTURE: Gesture object, x, y, error, num_fingers

Record a new gesture with `Gesture.record()`. The next DOLLARRECORD event returns a new Gesture instance.

```python3
class Gesture:
    readonly long gesture_id
    readonly long touch_id

    long last_touch_id
    int last_num_fingers
    float last_error
    float last_x, last_y

    @staticmethod
    def record(touch_id) # fires a DOLLARRECORD event after the next gesture

    def save(file) # saves this $ gesture to a file path or file object

    @staticmethod
    def save_all(file) # saves all gestures
    @staticmethod
    def load(file) # loads gestures from a file path or object
```

### int pgEvent_RegisterFilter(cb)

This takes a pointer to a callback function with the signature `pgEventObject* callback(pgEventObject *)`. The callback functions receive a `pgEventObject` before it's returned by pgEvent_New or pgEvent_New2.

Not sure if it's a good idea but it seems better than coupling `_sdl2.touch` and `event`.

# WIP

- [ ] Gesture.load
- [ ] Gesture.save
- [ ] Gesture.save_all
- [ ] Docs: pgEvent_RegisterFilter(), Gesture, events


illume:

PR #926, opened 2019-03-23
WIP: Start updating dependency versions for manylinux.

illume:
I started updating the dependencies for the manylinux builds.

- [x] work out sha512 sums for all the downloads.
- [x] make a new docker image and test it.
- [x] do some testing of whl files.


buildconfig/manylinux-build/docker_base/Dockerfile-x86_64
```diff
@@ -4,10 +4,10 @@ FROM quay.io/pypa/manylinux1_x86_64
 COPY RPM-GPG-KEY.dag.txt /tmp/
 RUN rpm --import /tmp/RPM-GPG-KEY.dag.txt
 
-ENV RPMFORGE_FILE "rpmforge-release-0.5.3-1.el5.rf.x86_64.rpm"
-ADD "https://repoforge.cu.be/redhat/el5/en/x86_64/dag/RPMS/${RPMFORGE_FILE}" /tmp/${RPMFORGE_FILE}
+#ENV RPMFORGE_FILE "rpmforge-release-0.5.3-1.el5.rf.x86_64.rpm"
+#ADD "https://repoforge.cu.be/redhat/el5/en/x86_64/dag/RPMS/${RPMFORGE_FILE}" /tmp/${RPMFORGE_FILE}
```
takluyver: Without repoforge, it looks like it's unable to find smpeg-devel (which SDL_mixer builds against), or pulseaudio-libs-devel (which fluidsynth and possibly SDL build against). Was that intentional?

Smpeg looks like it's unmaintained, so maybe it makes sense to drop it. Not supporting Pulseaudio might be more problematic, though I don't really know all the different pathways for sound on Linux. But I'm just looking at building with manylinux2010, and that seems to have pulseaudio-libs-devel anyway.

notpygame: I'm sort of surprised we didn't have a test for mp3 loading.

smpeg is pretty bad at what it does, and yes unmaintained outside of the various random distro patches. It was unintentional to leave it out.

@llindstrom made this patch which uses libmpg123 for sdl_mixer 1.2. https://bitbucket.org/llindstrom/sdl_mixer-1.2-patches/overview However, I guess we should use smpeg still for this just because we know how it's bad, and the other platforms are using smpeg anyway.

There were lots of issues with fluidsynth and alsa/pulse using 100% CPU. They haven't been investigated fully. It was also unintentional to leave out pulseaudio-libs-devel. It would be interesting to see if the 100% cpu use is still there in the released 1.9.5 binary wheels. Or if the new manylinux2010 build solves this.

PR #919, opened 2019-03-20
bugfix for dirty sprite when using a source rect, part 2.

illume:
Continuing on from https://github.com/pygame/pygame/pull/899

in dr0id-master branch.

PR #900, opened 2019-03-12
Add sprite tests

charlesej:
This update adds some sprite tests. One of the tests reproduces the behavior seen in issue #898.

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit) and 2.7.10 (64bit)
- pygame: 1.9.5.dev0 (SDL: 1.2.15) at 811eaf051e86a4c711d42a0d426dfc138668f384

Related to issue #898.

PR #928, opened 2019-03-23
Fix LayeredDirty's inconsistent use of its source rect

charlesej:
This update changes `LayeredDirty` so the use of its `source_rect` is consistent.

Overview of changes:
- Changed `LayeredDirty` so the use of its `source_rect` is consistent
- Removed `@unittest.expectedFailure` decorator from the now passing test

System details:
- os: windows 10 (64bit)
- pygame: 1.9.5.dev0 (SDL: 1.2.15) at 883dacca377160da2483db85e7482d27126b94b0

Related: #899, #900, #919.


illume:
Very good :)

PR #929, opened 2019-03-24
Use SDL_RWops instead of pg_FopenUTF8.

dlon:
This removes pg_FopenUTF8 in favor of SDL_RWFromFile. It was only used in imageext.c.

One issue is that the RWops API lacks a flush function, so I had to remove those. It seems to work fine without explicit flushing, though. SDL_image does more or less the same thing.

- [ ] Manually test the image saving functions.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -373,9 +373,9 @@ typedef struct {
 
 #define PYGAMEAPI_BASE_FIRSTSLOT 0
 #if IS_SDLv1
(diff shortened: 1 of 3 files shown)


illume:
Cool. It's nice to get rid of all that code.

Without flushing, and fsync things will mostly work ok on local file systems with machines that aren't very loaded. When the machine is very busy, and file systems are remote or busy, that's when the issues start to come.

I guess the way forward is to try to get a flush call into SDL?

PR #932, opened 2019-03-25
Flush while saving images

dlon:
The CI error is unrelated to the PR.

The SDL_RWflush() macro only works for stdio and Windows but this should be fine for the save functions since they only accept paths at the moment.


illume:
I guess this could be sent to the SDL folks :)

Probably we should call the macro something else, in case they include one in the future.

dlon:
I've sent them a patch

Issue #816, opened 2019-02-12
Drop events


dlon:
This event for drag-and-drop seems worth including:

- [x] https://wiki.libsdl.org/SDL_DropEvent
- [x] Documentation

PR #933, opened 2019-03-25
Drop event docs

dlon:
#816


notpygame:
Do we need the 'SDL' prefix?

dlon:
Oops

PR #934, opened 2019-03-25
Add Vector2.set() and Vector3.set()

dlon:
Changes:
* Add Vector.set(), which sets the coordinates in place and has the same parameters as the constructors. Returns self.
* Docs: Removed "constructors require 2-3 elements" notice. The constructors still don't require any argument.


illume:
Thanks. I asked about other names than set, which kind of reminds me of `set`.

Here were the alternatives...
```
v.set(1,1,0).rotate_ip(10)
v.mutate(1,1,0).rotate_ip(10)
v.fix(1,1,0).rotate_ip(10)
```

But I guess `v.set` is ok? We have a couple of days before the final 1.9.5 is out to change it.

PR #935, opened 2019-03-25
Vector: allow chaining of in-place methods

dlon:
Return `self` from in-place methods like rotate_ip, allowing you to chain methods, eg `v.rotate_ip(123).scale_to_length(30)`.


illume:
Very cool feature :)

PR #943, opened 2019-03-29
Alternative to "Rename vector.set() to vector.update()"

dlon:
Same as #942 but also reverts the chaining patch. Close that one if you agree that the change should be reverted.


illume:
I really like the idea of chaining... but I guess maybe we should think about it for longer.

PR #945, opened 2019-03-30
Vector: Explain slices and swizzling

dlon:


illume:
It definitely isn't obvious how to do this. Thanks.

PR #947, opened 2019-03-30
draw.circle(): accept Vector2

### LINKED ISSUE #579: Vector2 useless for Point Coordinates
Try to us Vector2(D) for Point Coordinates like:

`vec = pygame.math.vector2`  
`point1 = vec(100,40)`
`pygame.draw.circle(surfase, color, point1, radius)`

This throws an error like, only integer allowed for point1. 

I recommend to change the coordinate usages of draw to floating point in general. So there ist no confusing like draw.aalines = floating point coordinates, draw.rect = integer coordinates and so on.

dlon:
The pos parameter to `pygame.draw.circle` can be a sequence of floats as well as integers, so that it works with `pygame.Vector2`.

Closes #579.


dlon:
Limited it to draw.circle. I think #579 can be closed because the other functions already seem to work with floats.

Gravitar64:
Great work! Tried the Version 2.0.0 dev3 today. Works fine with np.arrays with float (needed for 3D/4D rotation matrix multiplications).

I'm very happy with that :-)

PR #978, opened 2019-04-12
Build 64-bit Linux wheels with manylinux2010 

takluyver:
This updates the manylinux build to use the newer manylinux2010 tag based on CentOS 6. The base images are only available for 64-bit architectures at the moment, so I've disabled the machinery to build 32-bit wheels, but left it there in case it's wanted again in the future.

The built packages currently fail some tests, however.

### DIFF
--- a/buildconfig/manylinux-build/Makefile
+++ b/buildconfig/manylinux-build/Makefile
@@ -1,7 +1,7 @@
 REPO_ROOT = $(abspath ../..)
 
 wheels-x64:
-	docker run --rm -v ${REPO_ROOT}:/io pygame/manylinux1_base_x86_64 /io/buildconfig/manylinux-build/build-wheels.sh
+	docker run --rm -v ${REPO_ROOT}:/io:z pygame/manylinux2010_base_x86_64 /io/buildconfig/manylinux-build/build-wheels.sh
 
 wheels-x86:
 	docker run --rm -v ${REPO_ROOT}:/io pygame/manylinux1_base_i686 /io/buildconfig/manylinux-build/build-wheels.sh
@@ -9,7 +9,7 @@ wheels-x86:
 wheels: wheels-x64 wheels-x86
 
 base-image-x64:
-	docker build -t pygame/manylinux1_base_x86_64 -f docker_base/Dockerfile-x86_64 docker_base
+	docker build -t pygame/manylinux2010_base_x86_64 -f docker_base/Dockerfile-x86_64 docker_base
 
 base-image-x86:
 	docker build -t pygame/manylinux1_base_i686 -f docker_base/Dockerfile-i686 docker_base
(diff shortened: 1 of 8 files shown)


takluyver:
Tests pass now. It seems that the tests don't run without fontconfig and at least one font installed. We might want to make an explicit error for that.

illume:
Very cool that the next manylinux is coming out.

I'd like to modify the current manylinux builds to work with smpeg(and perhaps pulsaudio if that was missing too). This would be done on a branch based on the 1.9.5 tag, as a `1.9.5.post1` release. This is because a few people have asked for this.

Additionally, there hopefully won't be another SDL1 based release. I want to put all my efforts after this smpeg fix onto an SDL2 based release.

For pygame, I wonder if it is worth losing 32bit support? The current manylinux builds sort of 'work'(but have various issues on different distros). For most people it's better compiling from source, or using platform packaged variants (that have patched SDL themselves with a big series of patches). Rather than try and do this work to fix the SDL1 based builds, I'd rather try to get an SDL2 based release which has further modern platform support.

It would be interesting to compare the new manylinux2010 builds to see if they fix anything (the 100% CPU core issue perhaps). But I would rather spend the time working SDL2 builds before doing lots of testing. Or do you see value testing these anyway?

How about:
Leaving the existing manylinux folder there, and moving your work into a new folder?

ps. Because mac, and linux both have mostly custom build processes now, it might be better for us to move to the conan C/C++ package manager, and use their packages. To take advantage of their maintenance work. Here is the SDL2 package: https://github.com/bincrafters/conan-sdl2

takluyver:
I can do another pull request just re-enabling repoforge, if you like; I think that should let things build against smpeg and pulseaudio again.

If you can stand losing 32-bit Linux packages, I'd migrate to manylinux2010 alongside the switch to SDL2. This doesn't need to mean dropping support for 32-bit Linux: pygame could still be built from source, or installed through other packaging systems. And of course, someone could make an image to build 32-bit manylinux2010 wheels - it's perfectly possible, it was just decided to focus on the 64-bit image to get something ready.

I'm inclined to stick with the canonical way to build the dependencies, rather than adding a layer of indirection through an unfamiliar package manager. But if you want to use another package manager, go for it.

illume:
I'd definitely like to test this out, and think we should merge this in.

But I'm wary of dropping support for peoples machines if there's not a big benefit right now. Even if it's only 2% of people. If it fixed some bad bugs that would be a different story.

We can always resurrect the existing manylinux builds with the magic of source control to make 32bit builds.

ps. do you know about the pypy situation with manylinux2010?
pps. conan is gaining a lot of momentum in the C/C++ world, and it seems their SDL2 packages are the best maintained of all SDL2 packages. The benefit of using only one package manager for windows, mac, linux, android, ... seems far to good to pass up. One small problem at the moment is configuration of the dependencies being different on each platform, leading to subtle differences. Having to change several build systems each time we want to update a dependency is quite hard. But trying conan can come later, and may still not turn out to be viable. Just to be clear, I'm not asking you to do this, I just wanted to let you know of a plan I've been thinking about.

illume:
I guess we can look at this in the future. For now manylinux is ok. Closing this to keep the PR list clean.

PR #981, opened 2019-04-13
Fix deadlock in music module

Programator2:
This PR fixes deadlock between two threads holding the GIL and `mixer_lock`.

There are a few functions in `music` module that didn't release the GIL when calling SDL_mixer functions that lock the `mixer_lock`. 

### An example
There are two threads:
**Thread 1**, created using `threading` module, it plays the background music using the `pygame.mixer.music` module.
**Thread 2** was created by an external library (SDL) to play a sound effect (via `pygame.mixer.Sound.play`).

Consider  the following succession of events:
1. Thread 2 locks `mixer_lock` in `SDL_RunAudio()`in preparation to play a sound.
2. Thread 1 (running the Python interpreter and thus holding the GIL) calls `music.get_busy()`(or other function fixed in the commit).
3. Thread 1 calls `SDL_LockAudio()`, but the `mixer_lock` is already acquired, so it has to wait
4. Thread 2 calls `_Mix_channel_done_playing()` which in turn  calls pygame callback `endsound_callback()` that tries to acquire the GIL, but the GIL is owned by Thread 1 which is waiting on `mixer_lock` => **Deadlock**

### DIFF
--- a/src_c/music.c
+++ b/src_c/music.c
@@ -110,8 +110,15 @@ music_play(PyObject *self, PyObject *args, PyObject *keywds)
 static PyObject *
 music_get_busy(PyObject *self, PyObject *args)
 {
+    int playing;
+
     MIXER_INIT_CHECK();
-    return PyInt_FromLong(Mix_PlayingMusic());
+
+    Py_BEGIN_ALLOW_THREADS
+    playing = Mix_PlayingMusic();
+    Py_END_ALLOW_THREADS
+
+    return PyInt_FromLong(playing);
 }
 
 static PyObject *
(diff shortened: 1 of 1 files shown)


illume:
Great, thank you!

Do you have a program that can be used to reproduce this?

Programator2:
I'm glad to help.

Yes, you can reproduce it in PySolFC, but it shows the problem just in one method and you have to play it for some time before a deadlock occurs.

I have pushed [some scripts](https://github.com/Programator2/pygame-deadlock-pr981) that reproduce the deadlock for all affected methods automatically.

illume:
Thanks again. I could reproduce easily with your scripts.

Issue #966, opened 2019-04-05
MP3 support missing in Pygame 1.9.5 wheels


lordmauve:
Pygame 1.9.5 seems not to be able to play MP3 files. I tried it and got an error about undecodeable file type.

Pygame 1.9.4 wheels contain libsmpeg.so while the 1.9.5 wheels do not:

```
$ unzip -l pygame-1.9.5-cp37-cp37m-manylinux1_x86_64.whl | grep /.libs/
  2582880  2019-03-31 09:23   pygame/.libs/libSDL-1-c39928c2.2.so.0.11.4
    59984  2019-03-31 09:23   pygame/.libs/libportmidi-7fcf6c23.so
    59024  2019-03-31 09:23   pygame/.libs/libvorbisfile-f207f3a6.so.3.3.7
    17000  2019-03-31 09:23   pygame/.libs/libcap-5fec18ee.so.1.10
  1724904  2019-03-31 09:23   pygame/.libs/libFLAC-6cefaab3.so.8.3.0
   703944  2019-03-31 09:23   pygame/.libs/libSDL_mixer-1-34ab1931.2.so.0.12.0
   142552  2019-03-31 09:23   pygame/.libs/libjpeg-bd53fca1.so.62.0.0
   236240  2019-03-31 09:23   pygame/.libs/libvorbis-373f4a71.so.0.4.8
  1723552  2019-03-31 09:23   pygame/.libs/libfluidsynth-e2f12a37.so.1.5.2
  3631832  2019-03-31 09:23   pygame/.libs/libwebp-ec096d5f.so.7.0.4
  5163104  2019-03-31 09:23   pygame/.libs/libasound-87b0bd31.so.2.0.0
   260624  2019-03-31 09:23   pygame/.libs/libdbus-1-30e56626.so.3.4.0
    87848  2019-03-31 09:23   pygame/.libs/libz-a147dcb0.so.1.2.3
   308648  2019-03-31 09:23   pygame/.libs/libmikmod-fabcac29.so.2.0.4
   328920  2019-03-31 09:23   pygame/.libs/libsndfile-d5ecee11.so.1.0.17
   389776  2019-03-31 09:23   pygame/.libs/libtiff-97e44e95.so.3.8.2
   280640  2019-03-31 09:23   pygame/.libs/libSDL_image-1-c6cf9ff4.2.so.0.8.4
   959776  2019-03-31 09:23   pygame/.libs/libpng16-e11200e8.so.16.36.0
[...]

PR #984, opened 2019-04-14
Build smpeg from source in the manylinux build.

illume:
for https://github.com/pygame/pygame/issues/966


buildconfig/manylinux-build/docker_base/sdl_libs/build-sdl-libs.sh
```diff
@@ -26,6 +26,16 @@ cd ..
 # Link sdl-config into /usr/bin so that smpeg-config can find it
 ln -s /usr/local/bin/sdl-config /usr/bin/
 
+# Build smpeg.
+svn co svn://svn.icculus.org/smpeg/tags/release_0_4_5
```
takluyver: Is the `svn://` protocol secured? A quick search suggests it's not, but maybe I've missed something.

I tried to check hashes of all the tarballs downloaded because this would be a very convenient pathway for malware: Man-In-The-Middle the downloads on the build system (or hijack an expiring domain), inject malware along with a functional copy of the source code, and get it built into a popular Python package. The C sources of the dependencies aren't included in the wheel, so well-designed malware in one of the bundled libraries could be really hard to spot.

So we should be careful that anything downloaded by the build process is validated somehow to ensure it's the thing we think it is.

illume: Thanks. I added a hash for that too.

PR #980, opened 2019-04-13
Create ISSUE_TEMPLATE.md

BastiHz:
With the increasing popularity of pygame it may be useful to have an issue template.
It contains:
* link to the website's repository so that issues can be created in the right place
* bug template asking for important information

PR #987, opened 2019-04-15
Add python version to startup printout

BastiHz:
Adds the python version to the startup printout (see #980).

tested in python 3.6.7 and 2.7.15
pygame: 2.0.0.dev0 with sdl 1.2.15

At first I tried `.format(ver, *get_sdl_version(), *sys.version_info[0:3]))` which works in python 3 but throws a syntax error in python 2. This is why I "+" the two version tuples before unpacking.


illume:
Awesome! Thank you :)

PR #990, opened 2019-04-17
Update header files, update API, add public headers

dlon:
## Overview

pygame module import:
* Moved module import code _pygame.h to pgimport.h. Main benefit: This allows one to include the `mask` module and `bufferproxy` without depending on SDL (or the base modules). *Update*: _pygame.h can also be included without SDL.h (but some functions won't work)
* Added macros PYGAMEAPI_DEFINE_SLOTS(), PYGAMEAPI_EXTERN_SLOTS(), PYGAMEAPI_GET_SLOT(). Reasons: (a) Easier search, (b) can define NO_PYGAME_C_API to cause them to be replaced with nothing, (c) improved readability. E.g. from pgbufferproxy.h
```c
PYGAMEAPI_DEFINE_SLOTS(PgBUFPROXY_C_API, PYGAMEAPI_BUFPROXY_NUMSLOTS);
```

Fixed bugs:
* `pgFont_Type()` referred to the wrong slot
* `pgBufproxy_Check()` checked against to the type value instead of its pointer
* Inclusion of "pygame.h" in freetype.h probably created a bunch of unneeded arrays?
* ftfont.render_raw() imported PG slots on each call

Added ~~include/~~ src_c/include/ directory for public header files that may hide unneeded data structures, etc. Advantages: (a) avoids giving public headers potentially conflicting names like "freetype.h", (b) allows hiding of internal stuff, (c) doesn't break existing code, (d) improved readability/reduced complexity.

Current directory structure:
```
src_c/include/bitmask.h
src_c/include/pgcompat.h
src_c/include/pgimport.h
src_c/include/pgplatform.h
src_c/include/pygame.h
src_c/include/pygame_bufferproxy.h
src_c/include/pygame_font.h
src_c/include/pygame_freetype.h
src_c/include/pygame_mask.h
src_c/include/pygame_mixer.h
src_c/include/_pygame.h
```

If the last couple of commits aren't worth including, the others probably still are.

General cleanup and other stuff:
* pgplatform.h for (most) platform/compiler stuff
* improved formatting
* share c_api arrays between modules instead of copying them


illume:
f' awesome :)

It's nice that you fixed some bugs in the cleanup.

I'm wondering about `src_c/include/` instead of `include`, because less files in the root ?

dlon:
Is it alright if _pygame.h doesn't include SDL.h at all? It will only cause a compiler error if you try to use functions that depend on its types

dlon:
Should a null check be added to `_IMPORT_PYGAME_MODULE`? So doing something like this doesn't access Python or cause it to update the array:
```
import_pygame_base();
import_pygame_base();
import_pygame_base();
```
And potentially one could do this:
```
#define import_pygame_color()          \
    import_pygame_base()               \
    IMPORT_PYGAME_MODULE(color, COLOR)
```

dlon:
Can I merge this?

Issue #967, opened 2019-04-05
SDL2: ImportError raised for touch tests with python 2.7


charlesej:
Running the touch tests with python 2.7 (`py -2 -m pygame.tests.touch_test`) raises the following error (works with python 3.7).

```
pygame 2.0.0.dev0
Hello from the pygame community. https://www.pygame.org/contribute.html
Traceback (most recent call last):
  File "C:\Program Files\Python27\lib\runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Program Files\Python27\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\Program Files\Python27\Lib\site-packages\pygame-2.0.0.dev0-py2.7-win-amd64.egg\pygame\tests\touch_test.py", line 3, in <module>
    from pygame._sdl2 import touch
ImportError: No module named _sdl2
```
\
One quick work around is to manually add an empty `__init__.py` file to the `pygame\_sdl2` directory.\
\
System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit) and 2.7.10 (64bit)
- pygame: 2.0.0.dev0 (SDL: 1.2.15 and SDL: 2.0.9) at 38779ec6ece565f013b6b3a64ecf57e0d1f05459

notpygame:
Fixed in #992

PR #992, opened 2019-04-17
Only build and test SDL2 based windows wheels on appveyor.

notpygame:
So we can start uploading binaries for testing 2.0.0.devN.


charlesej:
The 2.7 tests fail due to a missing module (#967).

PR #995, opened 2019-04-17
Add static assertions

dlon:
This branch derives from #990.

Additions:
* Static bounds check for `PYGAMEAPI_GET_SLOT()`. If the number is greater than the size of the array declared with `PYGAMEAPI_DEFINE_SLOTS()`/`PYGAMEAPI_EXTERN_SLOTS()` it causes a compiler error.
* Safer encapsulation of the c_api[] slot arrays. If size of the array doesn't match PYGAMEAPI_<MODULENAME>_NUMSLOTS, it causes a static assertion fail with the message "encapsulate_wrong_size".
* Static checks import_pygame_*(). If the imported slots can't fit inside the array declared with `PYGAMEAPI_DEFINE_SLOTS()`/`PYGAMEAPI_EXTERN_SLOTS()`, it causes a static assertion fail with the message "api_root_too_small".

Other stuff:
* Also, running import_pygame_xyz() more than once has no effect now.
* Moved encapsulation code to private pgimport.h.

Possible issues:
* Maybe overly complicated
* Maybe the import_pygame_*() checks are unnecessary since the encapsulation checks pretty much do the same thing (and not repeatedly).
* Not sure how portable the `PG_STATIC_ASSERT()` macro is. It definitely works for C99.
* Not entirely sure if `PG_STATIC_ASSERT_EXPR()` works as intended, but probably. It evaluates to a non-zero integer if the assertion succeeds, so for `(PG_STATIC_ASSERT_EXPR(...) ? dostuff() : NULL)`, any compiler should remove the if statement since it's always true.

### DIFF
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -2250,13 +2250,11 @@ MODINIT_DEFINE(_freetype)
     DEC_CONST(BBOX_PIXEL_GRIDFIT);
 
     /* export the c api */
-#if PYGAMEAPI_FREETYPE_NUMSLOTS != 2
-#error Mismatch between number of api slots and actual exports.
-#endif
     c_api[0] = &pgFont_Type;
     c_api[1] = &pgFont_New;
 
-    apiobj = encapsulate_api(c_api, "freetype");
+    _encapsulate_api_safe(&apiobj, c_api, freetype, FREETYPE,
+                          PgFREETYPE_C_API);
     if (!apiobj) {
         DECREF_MOD(module);
         MODINIT_ERROR;
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -178,6 +178,8 @@ struct pgSubSurface_Data {
     int offsetx, offsety;
 };
 
+#include "pgimport.h"
+
 /*
  * include public API
  */
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -2185,20 +2185,15 @@ MODINIT_DEFINE(base)
     c_api[17] = pgDict_AsBuffer;
     c_api[18] = pgExc_BufferError;
 #if IS_SDLv1
-#define FILLED_SLOTS 19
 #else /* IS_SDLv2 */
     c_api[19] = pg_GetDefaultWindow;
     c_api[20] = pg_SetDefaultWindow;
     c_api[21] = pg_GetDefaultWindowSurface;
     c_api[22] = pg_SetDefaultWindowSurface;
-#define FILLED_SLOTS 23
 #endif /* IS_SDLv2 */
 
-#if PYGAMEAPI_BASE_NUMSLOTS != FILLED_SLOTS
-#error export slot count mismatch
-#endif
 
-    apiobj = encapsulate_api(c_api, "base");
+    encapsulate_api_safe(&apiobj, c_api, base, BASE);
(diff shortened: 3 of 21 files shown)


dlon:
`_IMPORT_PYGAME_MODULE(module, MODULE, api_root)`

If the redundant "MODULE" parameter is removed, it should be possible to do something like this:

```c
#define PYGAMEAPI_GET_SLOT(api_root, index)    \
    PG_STATIC_ASSERT(                          \
        index < PYGAMEAPI_##api_root##_NUMSLOTS, \
        slot_out_of_bounds);                   \
    api_root[(index)]
```

Could also add a "MODULE" argument, but it seems unnecessary

notpygame:
I don't know if it makes sense statically when using modules? Because each module is a separate compilation unit.

dlon:
It should work. An example for PYGAMEAPI_GET_SLOT():
```c
#define pg_FloatFromObj \
    (*(int (*)(PyObject *, float *)) \
        PYGAMEAPI_GET_SLOT(PyGAME_C_API, PYGAMEAPI_BASE_FIRSTSLOT + 5))
```

The first argument is the array created (or referenced with extern) by the same header as the one defining the slot:
```c
#ifdef PYGAME_H
PYGAMEAPI_DEFINE_SLOTS( PyGAME_C_API, PYGAMEAPI_TOTALSLOTS );
#else /* ~PYGAME_H */
PYGAMEAPI_EXTERN_SLOTS( PyGAME_C_API, PYGAMEAPI_TOTALSLOTS );
#endif /* ~PYGAME_H */
```

Here are those macro functions:
```c
#define PYGAMEAPI_DEFINE_SLOTS(api_root, numslots) \
    void *api_root[(numslots)] = {NULL}
#define PYGAMEAPI_EXTERN_SLOTS(api_root, numslots) \
    extern void *api_root[(numslots)]
```

PYGAMEAPI_GET_SLOT() checks that the index doesn't exceed the number of elements in the array:
```c
#define PYGAMEAPI_NUM_SLOTS(api_root) \
    sizeof(api_root) / sizeof(void*)
#define PYGAMEAPI_GET_SLOT_UNSAFE(api_root, index) \
    api_root[(index)]
#define PYGAMEAPI_GET_SLOT(api_root, index)                           \
    (PG_STATIC_ASSERT_EXPR( (index) < PYGAMEAPI_NUM_SLOTS(api_root) ) \
        ? PYGAMEAPI_GET_SLOT_UNSAFE(api_root, index) : NULL)
```
sizeof() works here because PyGAME_C_API is an array whose size is known.

dlon:
But couldn't we remake the module import entirely so that it doesn't create a copy of the array in each module? Instead of having a bunch of macros, straight function pointers would work:

```c
struct {
    PyTypeObject* pgFont_Type;
    PyObject (*pgFont_New)(const char*, long);
} pgFreeTypeModule;

pgFreetypeModule *import_pygame_freetype();

int function() {
    pgFreetypeModule *mod = import_pygame_freetype();
    mod->pgFont_New("abc", 123);
}
```

It seems better than defining an array in every module header, IMO

notpygame:
Seems like a great idea. I wonder if that works ok with the different dll systems (and static linking)?

dlon:
Should be fine since the address of the array won't change in the module that owns it, and PyCapsule_GetPointer is already being used to read from it

PR #996, opened 2019-04-18
Tests to NULL check API slots

dlon:
Ensures that the "c_api" arrays are initialized. Eg if you forget to set a slot,
```c
c_api[0] = &pgSurface_Type;
c_api[2] = pgSurface_Blit;
```
then you get an error like this:

```
loading test.api_test
........F
FAIL: test_surface (test.api_test.BaseModuleEntryTest)
  File "C:\Users\David\Documents\dev\pygame\test\api_test.py", line 86, in ret
    BaseModuleEntryTest.check_module(test, mod)
  File "C:\Users\David\Documents\dev\pygame\test\api_test.py", line 59, in check_module
    'API slot {} is NULL'.format(entryind))
AssertionError: unexpectedly None : API slot 1 is NULL

Ran 9 tests in 0.037s

FAILED (failures=1)
```
I made it a test because it didn't have to be a runtime check.

### DIFF
--- a/test/api_test.py
+++ b/test/api_test.py
@@ -12,6 +12,10 @@ ctypes.pythonapi.PyCapsule_GetName.restype = ctypes.c_char_p
 ctypes.pythonapi.PyCapsule_GetPointer.restype = \
     ctypes.POINTER(ctypes.c_void_p)
 
+def moduleToNUMSLOTS(module):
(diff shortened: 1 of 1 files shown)


notpygame:
I guess this one, and https://github.com/pygame/pygame/pull/995 can be closed?

PR #1000, opened 2019-04-19
Share c_api array instead of copying it for each module

dlon:
Makes one of the changes mentioned earlier. Using a shared c_api[] array instead of copying it for each import.
* More readable headers.
* Uses slighly less memory. If there were about 10 slots on average per module and 30 modules, each slot taking up about 4 bytes (a void* pointer), it took up about 35 KB of memory. Now there is roughly one pointer to each other module and a single array for each, so it should use about 1/10 of that. Insignificant in practice, I guess.

The import_pygame_*() macros are backward compatible.

Error is due to #967

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -178,6 +178,13 @@ struct pgSubSurface_Data {
     int offsetx, offsety;
 };
 
+/*
+ * include public API
+ */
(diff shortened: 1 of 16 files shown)


notpygame:
Looks good... should this be merged in now?

Issue #946, opened 2019-03-30
Is posting events from a signal handler reliable?


wexi:
Hi fellows,

I am using Python 2.7.15 with Pygame 1.9.2 on a small buildroot made
Linux hardware.

I am now quite certain that pygame.event.post() and pygame.event.Event()
are unsafe to use from Python signal handler due to some deadlock
condition with the main thread pygame activities. It is very likely
that sdl-1.2.15 is to blame.

I solved my problem by making sure through signal.pause() that Python
signal handler and the main thread do not operate concurrently, but
that's "ugly".

Comments please.

Thanks, Enoch.

**Related Docs:** https://www.pygame.org/docs/ref/event.html#pygame.event.post

wexi:
SDL_mutexP is to blame. It should return an error when trying to lock a locked mutex if called from the signal handler, which is also the main thread of execution. Instead, it deadlocks... alas.

illume:
Perhaps pygame.fastevent will be your friend...

wexi:
Thanks, I already took a look at fastevents.c (as included in pygame sources).
Actually it incorporates the problematic SDL_PushEvent and I don't even understand what's fast about it.
It's clear that attention was given to multi-threading support and none at all to signal handlers integration.

illume:
It seems to work better when posting from different threads.

Another thing to try might be SDL2, which has quite an different event system.

wexi:
That's my thinking too.  Since Python signal handler can only be in the main thread I'll move video handling to a second thread. 

sdl-1.2.15 documentation though says:
"Don't call SDL video/event functions from separate threads" :-)
Will see.  My use of signal.pause() to avoid collision works for now.

pygame done not support SDL2 yet, doesn't it?

husano896:
> pygame done not support SDL2 yet, doesn't it?

The newest version of pygame supports SDL2 now,

you can follow the instructions here.

https://www.pygame.org/wiki/CompileWindows#SDL2%20instructions.

wexi:
As previously reported pygame.event.post is unsafe (may deadlock) if called from a signal handler.
Using signal.pause in pygame event loop (to avoid the main-thread and the signal-handler working concurrently) was my first (limited applicability) fix. Now it is time for a better solution. Attached is my proposed patch, pygame.event.apost(e) to post the event 'e' via a dedicated thread. Comments?
[async-event-post.zip](https://github.com/pygame/pygame/files/3074422/async-event-post.zip)

[...]

PR #1003, opened 2019-04-19
pygame.event.apost: async event posting, safe for main thread calling

wexi:
Calling pygame.event.post from Python signal handlers (which execute in the main thread) was found to be unsafe (can deadlock). This solution, which is using serialized threads to push the events, was tested on the following buildroot python-pygame (buildroot's latest):

PYTHON_PYGAME_VERSION = d61ea8eabd56
PYTHON_PYGAME_SOURCE = pygame-$(PYTHON_PYGAME_VERSION).tar.gz
PYTHON_PYGAME_SITE = https://bitbucket.org/pygame/pygame
PYTHON_PYGAME_SITE_METHOD = hg

The patch was adapted to the latest pygame code in the hope that it would find someone to test it in a different environment.

Thanks, Enoch.

### DIFF
--- a/src_c/doc/event_doc.h
+++ b/src_c/doc/event_doc.h
@@ -13,6 +13,9 @@
 #define DOC_PYGAMEEVENTSETGRAB "set_grab(bool) -> None\ncontrol the sharing of input devices with other applications"
 #define DOC_PYGAMEEVENTGETGRAB "get_grab() -> bool\ntest if the program is sharing input devices"
 #define DOC_PYGAMEEVENTPOST "post(Event) -> None\nplace a new event on the queue"
+
+#define DOC_PYGAMEEVENTAPOST "apost(Event) -> None\nplace a new event on the queue asynchronously (via SDL thread)"
+
 #define DOC_PYGAMEEVENTEVENT "Event(type, dict) -> EventType instance\nEvent(type, **attributes) -> EventType instance\ncreate a new event object"
 #define DOC_PYGAMEEVENTEVENTTYPE "pygame object for representing events"
 #define DOC_EVENTTYPETYPE "type -> int\nevent type identifier."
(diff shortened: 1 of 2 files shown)


dlon:
Just a guess, but have you tried wrapping `Py_BEGIN_ALLOW_THREADS` and `Py_END_ALLOW_THREADS` around SDL_PushEvent?

wexi:
I am not the only one encountering this deadlock issue. Let me quote from fastevent.c comments, it does SDL_PushEvent just the same (and it does prefix/suffix PushEvents with Py_BEGIN_ALLOW_THREADS!!!):

/*DOC*/ "Because pygame.fastevent.post() may have to wait for the queue\n"
/*DOC*/ "to empty, you can get into a dead lock if you try to append an\n"
/*DOC*/ "event on to a full queue from the thread that processes events.\n"
/*DOC*/ "For that reason I do not recommend using this function in the\n"
/*DOC*/ "main thread of an SDL program.\n"

I followed this advice in apost() and do PushEvent from a thread.
As I encountered SDL_CreteThread lack of resources occasionally (my signal handler interrupts 20 times per second) I serialized the threads via SDL_WaitEvent().

My pygame version (buildroot's approved) is circa V1.9.1. I hope that my updating to the latest standard is Ok. Right?

Thanks, Enoch.



On 4/19/19 5:10 PM, David Lönnhager wrote:

Just a guess, but have you tried wrapping Py_BEGIN_ALLOW_THREADS and Py_END_ALLOW_THREADS around SDL_PushEvent?

illume:
This is related to the issue: https://github.com/pygame/pygame/issues/946

illume:
@wexi Thank you very much for this.

How can I test this? Do you already have some code snippet I could run?

Something like this?
- setting up a signal handler which posts events 
- then get the signal handler to run
- watch it deadlock

robertpfeiffer:
I like this feature, and it is certainly much easier to do it this way than to use py3 asyncio. Why can't all events be posted like this? Is there a reason to keep the old version around as a separate function?

illume:

Issue #896, opened 2019-03-09
Improvements to pygame.draw


charlesej:
Here are some improvements that can be made to `pygame.draw` related areas.

Note: There are currently 2 draw modules `draw` (`src_c/draw.c`) and `draw_py` (`src_py/draw_py.py`). Any changes/tests should be applied to both modules.

- [x] Add keyword arguments to the draw functions (related to issue #808). Also add relevant tests and update the documentation with [sphinx info field lists](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists) for the parameters and return values.
  - [x] `pygame.draw.rect`
  - [x] `pygame.draw.polygon`
  - [x] `pygame.draw.circle`
  - [x] `pygame.draw.ellipse`
  - [x] `pygame.draw.arc`
  - [x] `pygame.draw.line`
  - [x] `pygame.draw.lines`
  - [x] `pygame.draw.aaline`
  - [x] `pygame.draw.aalines`

- [ ] Add tests to ensure correct bounding rects are being returned from all the draw functions. In some instances the bounding rects for `pygame.draw.line` and `pygame.draw.aaline` are incorrect (issue #895). Other draw functions could have similar issues.
  - [ ] `pygame.draw.rect` (related to issue #1305)
  - [ ] `pygame.draw.polygon` (related to issue #1306)
  - [x] `pygame.draw.circle` (related to issue #1122)
  - [x] `pygame.draw.ellipse` (related to issue #1241)
  - [ ] `pygame.draw.arc` (being tracked by issue #1143, related to issue #1212)
  - [x] `pygame.draw.line` (related to issue #895)
  - [x] `pygame.draw.lines` (related to issue #1117)
  - [x] `pygame.draw.aaline` (related to issue #895)
  - [x] `pygame.draw.aalines` (related to issue #1153)

- [x] Add tests to ensure all the draw functions respect a given surface's clip area.
  - [x] `pygame.draw.rect`
  - [x] `pygame.draw.polygon`
  - [x] `pygame.draw.circle`
  - [x] `pygame.draw.ellipse`
  - [x] `pygame.draw.arc`
  - [x] `pygame.draw.line`
  - [x] `pygame.draw.lines`
  - [x] `pygame.draw.aaline`
  - [x] `pygame.draw.aalines`

- [x] Update the `pygame.draw` documentation. Some items of note.
  - [x] Ensure proper code linking and highlighting.
[...]

Issue #839, opened 2019-02-21
Types, hints, issues and approach


markph0204:
@illume provided a good approach/plan I wanted to capture and mark off as we progress:

- [ ] document types for the html(sphinx) and IDE documentation reading
- [ ] allow types to be used for type checking tools(mypy, pylint, jedi, etc), and/or IDEs
- [ ] types to be used for completion in IDEs (and tools like jedi)
- [ ] somehow work with our python, C, Cython, and rst
- [ ] figure out how we can do it incrementally (eg, maybe just Surface and a few other modules to start with).
- [ ] hopefully getting something useful released in the near term for 'core' types and modules in (1.9.5 release perhaps).
- [ ] hopefully not too much duplication of types if possible (eg, specifying in .rst, .pyi, and .py files)
- [ ] hopefully don't increase load time too much

Also for reference [[types](https://github.com/pygame/pygame/issues?utf8=✓&q=is%3Aissue+is%3Aopen+label%3Atypes+)] marked issues.

markph0204:
ref: https://www.python.org/dev/peps/pep-0561/

"For package maintainers wishing to ship stub files containing all of their type information, it is preferred that the *.pyi stubs are alongside the corresponding *.py files."

markph0204:
PR: https://github.com/pygame/pygame/pull/871

Provides type hints in IDEA type editors (PyCharm)
![image](https://user-images.githubusercontent.com/210862/53681349-5c31e900-3cb6-11e9-8eee-2b3102f95684.png)

NotaSmartDev:
Hey, I really like the idea but I'm not using PyCharm. Might do one day but as of now it is not the case. The IDE I'm using is SublimeText and I wondered if it'd be possible (and not too hard) for you guys to do something alike it ? I have no idea if SublimeText supports this kind of hints or not though.

markph0204:
@NotaSmartDev Yes this will work with Sublime text as well. Just use the plugin called **SublimeJEDI**.

* Sublime Text (SublimeJEDI [ST2 + ST3], anaconda [only ST3])

ref: https://github.com/davidhalter/jedi

NotaSmartDev:
Thanks you very much ! I've just installed Anaconda and it works fine if a little slow at first. Would you recommend me to use SublimeJEDI instead ? It's a really nice feature btw. Allows me not to look at the docs just as often.

markph0204:
[...]

PR #1004, opened 2019-04-20
Add keyword arguments to draw.arc

charlesej:
Overview of changes:
- Added keyword arguments to `draw.arc`
- Added/updated some exception messages for `draw.arc`
- Added tests to check `draw.arc` args/kwargs
- Fixed test variable name
- Updated draw documentation

System details:
- os: windows 10 (64bit)
- pygame: 2.0.0.dev0 (SDL: 1.2.15) at d84a08a31f7682dfd704f8d46dfd578031d742cd

Resolves sub-item `pygame.draw.arc` of item "Add keyword arguments to the draw functions..." of #896.


docs/reST/ref/draw.rst
```diff
@@ -153,13 +153,57 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
 
 .. function:: arc
 
-   | :sl:`draw a partial section of an ellipse`
-   | :sg:`arc(Surface, color, Rect, start_angle, stop_angle, width=1) -> Rect`
+   | :sl:`draw an elliptical arc`
+   | :sg:`arc(surface=Surface, color=Color, rect=Rect, start_angle=radians, stop_angle=radians) -> Rect`
```
dlon: Shouldn't `=` be used for default values?

charlesej: I've been using it to indicate named parameters as well. Should it be like this instead?
`arc(surface, color, rect, start_angle, stop_angle) -> Rect`
`arc(surface, color, rect, start_angle, stop_angle, width=1) -> Rect`

dlon: IMO, that's better. Though I wonder if it would be useful to include some of the RST type info in the docstring somehow?

charlesej: I've made the requested changes. I also opened issue #1020 to track this for the other shape functions.

Regarding the type info, currently the `:param:` and `:type:` directives are used to indicate any type information. Also there is issue #839 which is working on a general type strategy for pygame.

PR #1005, opened 2019-04-22
Update mask documentation - part 1

charlesej:
Partially resolves the "Update the `pygame.mask` documentation." item of #800.

### DIFF
--- a/docs/reST/ref/mask.rst
+++ b/docs/reST/ref/mask.rst
@@ -152,7 +152,7 @@ Starting from pygame 1.9.5 masks with width or height 0 are supported.
       Sets all bits in the mask to 1.
 
       :returns: ``None``
-      :rtype: None
+      :rtype: NoneType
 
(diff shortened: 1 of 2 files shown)


docs/reST/ref/mask.rst
```diff
@@ -158,26 +161,41 @@ Starting from pygame 1.9.5 masks with width or height 0 are supported.
...
+      :returns: ``None``
+      :rtype: None
 
       .. ## Mask.clear ##
 
    .. method:: invert
 
-      | :sl:`Flips the bits in a Mask`
+      | :sl:`Flips all the bits`
       | :sg:`invert() -> None`
```
dlon: I wonder if it's bad that the return type/value is repeated?

charlesej: Good point, the type should be `NoneType`. I've updated the documentation to fix this.

PR #1006, opened 2019-04-23
Update mask documentation - part 2

charlesej:
Partially resolves the "Update the `pygame.mask` documentation." item of #800.


illume:
Awesome. The mask module is heaps better now :)

PR #1014, opened 2019-05-02
Fix mixer cleanup and init

### LINKED ISSUE #955: cannot delete an mp3 file played using pygame
I am trying to play and audio file synthsized by Amzon Polly and want to delete the same .mp3 file after it plays.
I am unable to remove a file using os.remove() using the following code,

`import pygame
from pygame.mixer import *
import os
from pygame import mixer

 with open('output.mp3', 'wb') as f:
            f.write(spoken_text['AudioStream'].read())
            f.close()

        mixer.init()
        mixer.music.load('output.mp3')
        mixer.music.play()
        pygame.quit()
        pygame.mixer.init()
        while pygame.mixer.music.get_busy(): 
            pygame.time.Clock().tick(10)

        os.remove('output.mp3')`

The following error is shown:

`PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: output3.mp3`

dlon:
Close #955. Obtain `current_music` and `queue_music` every time mixer.init() is called instead of only when the module is initialized, or else mixer.quit() would only unload music the first time it was called.

### DIFF
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -338,6 +338,17 @@ pgMixer_AutoQuit(void)
     }
 }
 
+static PyObject*
+import_music()
+{
+    PyObject *music = PyImport_ImportModule(IMPPREFIX "mixer_music");
+    if (music == NULL) {
+        PyErr_Clear();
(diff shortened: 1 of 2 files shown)


illume:
Nice fix.

PR #1019, opened 2019-05-03
[DOC]Changing buffersize to buffer

hugonun:
Quoting [Ian Mallett's comment](https://www.pygame.org/docs/ref/mixer.html#comment_pygame_mixer_pre_init).

    Should be "buffer", not "buffersize"

Not sure to which branch I should PR, let me know if this is the right one.


dlon:
Thanks!

PR #1025, opened 2019-05-05
Fix draw function signatures

### LINKED ISSUE #1020: Fix draw function signatures
Change the draw function signatures to only use '=' to indicate optional arguments.

This currently applies to `draw.rect` and `draw.circle`, but after PRs #977 and #1018 are merged, `draw.ellipse` and `draw.polygon` will also need changing.

Example for `draw.rect`.
From:
> rect(surface=Surface, color=Color, rect=Rect) -> Rect
> rect(surface=Surface, color=Color, rect=Rect, width=0) -> Rect

To:
> rect(surface, color, rect) -> Rect
> rect(surface, color, rect, width=0) -> Rect

charlesej:
Fixed the draw function signatures. Also added some text to circle's `center` parameter.

Resolves #1020.


docs/reST/ref/draw.rst
```diff
@@ -105,16 +105,16 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
...
 
    Draws an circle on the given surface.
 
    :param Surface surface: surface to draw on
    :param color: color to draw with, the alpha value is optional if using a
       tuple ``(RGB[A])``
    :type color: Color or int or tuple(int, int, int, [int])
-   :param center: center point of the circle
+   :param center: center point of the circle, (x, y)
    :type center: tuple(int, int) or list(int, int)
```
illume: center now takes floats as well as int. Also Vector2 works. Not sure how to express all this exactly.

charlesej: Thanks. I opened #1040 to track this.

PR #1041, opened 2019-05-15
pgRWops_FromObject: Fail if an object is not a file object

dlon:
- [x] Check if failed tests are due to bugs


illume:
Interesting.

dlon:
Code relied on `pgRWops_FromFileObject()` not failing for objects that weren't file objects. IMO, it should fail immediately, so I updated it.

PR #1046, opened 2019-05-15
Fix SCALED for existing windows and improve test

dlon:
* Makes SCALED set the right size for existing windows.
* Added `display.get_window_size()`. For SDL1 (in case there's ever another release...), it's equal to the display surface size. For SDL2, it equals the window size (which may differ for SCALED).
* Test SCALED using window size.

- [x] Test it to make sure it works before merging.

Possible issue: `display.get_window_size()` does not work using pygame 1.9.6.

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -486,4 +486,16 @@ required).
 
    .. ## pygame.display.get_num_displays ##
 
+.. function:: get_window_size
+
+   | :sl:`Return the size of the window or screen`
+   | :sg:`get_window_size() -> tuple`
+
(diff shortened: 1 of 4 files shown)


dlon:
`SDL_GetDesktopDisplayMode()` reports the wrong resolution for me due to DPI scaling. So the window often is not as large as it should be. This should probably be handled automatically?

dlon:
Here's what the aliens example looks like on my machine without SCALED:
https://i.imgur.com/AiwlRup.mp4
And with SCALED:
https://i.imgur.com/pRzLtq0.mp4
It runs much more smoothly with SDL_Renderer.

illume:
Great :)

PR #1053, opened 2019-05-16
Enable DPI awareness on Windows.

dlon:
This patch enables DPI awareness on Windows. Before, `SDL_GetDesktopDisplayMode()` reported my resolution to be 1536x864, but it's actually 1920x1080. This meant that `pygame.SCALED` didn't scale the window maximally.

Platform: Windows 10

Related: #1046

### DIFF
--- a/buildconfig/Setup.SDL1.in
+++ b/buildconfig/Setup.SDL1.in
@@ -17,6 +17,7 @@ SCRAP = -lX11
 PORTMIDI = -lportmidi
 PORTTIME = -lporttime
 FREETYPE = -lfreetype
+BASE =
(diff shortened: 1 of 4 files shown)


illume:
Nice one.

I found this open bug report about hidpi support in SDL2: https://bugzilla.libsdl.org/show_bug.cgi?id=3281

This probably this function might have issues on other platforms too.

dlon:
Is disabling DPI scaling even the correct approach? If so, this might work:
```
SDL_SetHint(SDL_HINT_VIDEO_HIGHDPI_DISABLED, "1");
```

illume:

Issue #1029, opened 2019-05-07
Request a  numpy array data exporting interface of the mask.


DeluxeYang:
I want to make one of the overlapped surfaces to be half transparent, and I use mask.

the demo code is in
https://stackoverflow.com/questions/52043957/how-to-calculate-overlap-pixel-of-2-surfaces

problem is, now I have the overlapped mask, but I can only access the mask data one by one, it's pretty slow.

if pygame could export the mask's numpy array, maybe I can just reach the target by operating the numpy array and surfarray, I think that would be faster

dlon:
From bitmask.h:
```c
#define BITMASK_W unsigned long int
#define BITMASK_W_LEN (sizeof(BITMASK_W)*CHAR_BIT)
#define BITMASK_W_MASK (BITMASK_W_LEN - 1)
#define BITMASK_N(n) ((BITMASK_W)1 << (n))

typedef struct bitmask
{
  int w,h;
  BITMASK_W bits[1];
} bitmask_t;

static INLINE int bitmask_getbit(const bitmask_t *m, int x, int y)
{
  return (m->bits[x/BITMASK_W_LEN*m->h + y] & BITMASK_N(x & BITMASK_W_MASK)) != 0;
}

/* Sets the bit at (x,y) */
static INLINE void bitmask_setbit(bitmask_t *m, int x, int y)
{
  m->bits[x/BITMASK_W_LEN*m->h + y] |= BITMASK_N(x & BITMASK_W_MASK);
}
```

I guess you'd have to obtain (x, y) using `view[x // BITMASK_W_LEN, y] & BITMASK_N(x & BITMASK_W_MASK)`?

DeluxeYang:
No, I mean the [pygame.mask](https://www.pygame.org/docs/ref/mask.html).

The thing is, now I have a mask, if I want to know which points have been set, the only thing I can do is use [pygame.mask.get_at](https://www.pygame.org/docs/ref/mask.html#pygame.mask.Mask.get_at) then get  the point status one by one. So I am hoping, can there be an interface to output the status of all points at once? Like what you do on [pygame.surfarray.array_alpha()](https://www.pygame.org/docs/ref/surfarray.html#pygame.surfarray.array_alpha)? That would be a great help.

dlon:
I created a function to copy the data into a bytes object: #1057. Can be used with numpy as follows (for now):

```python3
import numpy as np

maskdata = mask.get_array()

[...]

PR #1057, opened 2019-05-19
mask.get_array()

dlon:
This creates a copy of the mask data as a `bytes` object.

It's still slow to loop over all pixels, though. Here is my test:

```python3
import numpy as np
import pygame


mask = pygame.mask.Mask((1000, 1000))

def readallnormal():
    w, h = mask.get_size()
    for x in range(w):
        for y in range(h):
            px = mask.get_at((x, y))

def readallnumpy():
    maskdata = mask.get_array()

    array = np.frombuffer(maskdata, dtype=np.int8)
    array = array.view()
    array.shape = mask.get_size()

    w, h = mask.get_size()
    for x in range(w):
        for y in range(h):
            array[x, y]

import timeit

print("get_at:", timeit.timeit(readallnormal, number=10))
print("array:", timeit.timeit(readallnumpy, number=10))

```

Output:

```
normal: 3.246463215
array: 2.1993275400000005
```

Re: #1029.

### DIFF
--- a/src_c/mask.c
+++ b/src_c/mask.c
@@ -73,6 +73,28 @@ mask_get_size(PyObject *self, PyObject *args)
 }
 
 static PyObject *
+mask_array(PyObject *self, PyObject *args)
+{
+    bitmask_t *m = pgMask_AsBitmap(self);
+    int x, y;
+    PyObject *bytesobj;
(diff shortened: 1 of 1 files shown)


dlon:
Closing due to #1059

PR #1059, opened 2019-05-19
Implement buffer interface for pygame.Mask

dlon:
Due to how the data is stored in bitmask_t, it's not very intuitive. The pixels are stored in bits.

Example usage:

```python3
>>> m = pygame.Mask((128,128))
>>> view = memoryview(m)
>>> view.strides
(512, 4)
>>> view.shape
(4, 128)
>>> view[0, 1]
0
>>> m.set_at((0, 1))
>>> view[0, 1]
1
>>> m.set_at((100, 1))
>>> view[0, 1]
1
>>> view[3, 1] # 100 // 32 = 3
16
>>> view[0, 1] = 0
>>> m.get_at((0, 0))
0
```

The first index ranges between 0 and ``width // (8 * sizeof(long))`` (on Win64, ``8 * sizeof(long) = 32``). The second index is in the interval 0 to ``height-1``. (Both inclusive)

Think of each row between 0 and the height (128) as containing 4 long ints above (since the shape is (4, 128)). If the x coordinate is eg 50, then it must be in the second long int for that row since 32 <= x < 64. For the fifth row, that would be in the integer `view[1, 4]`.

### Why is ``view[3,1]`` equal to 16 above?

The long int that the pixel is in begins at the offset ``x = 3*32 = 96``. 100 = 96 + 4, which means that the fifth pixel/bit in that integer is set. 10000<sub>2</sub> = 2<sup>4</sup> = 16.

### Comparison to `get_array()` and `get_at()`

Looping over all pixels is still slow, but it's faster than using the other methods:

| Time (s) | Iterations | Method             |
|----------|------------|--------------------|
| 3.2      | 10         | get_at()           |
| 2.2      | 10         | get_array() (copy) |
| .92      | 10         | memoryview()       |

```python3
mask = pygame.mask.Mask((1000, 1000))

def readallpixels():
    bits1 = memoryview(mask)
    w, h = mask.get_size()
    cols = bits1.shape[0]
    intwidth = math.ceil(w/cols)

    for col in range(cols):
        for y in range(h):
            val = bits1[col, y]
            for x in range(intwidth):
                if val & (1 << x):
                    pass # pixel is set

print("array:", timeit.timeit(readallpixels, number=10))
```

### Issues

- [x] Ignores flags.
- [ ] Fix for Python 2.7.

### Related

#1057, #1029

### DIFF
--- a/src_c/mask.c
+++ b/src_c/mask.c
@@ -1744,6 +1744,9 @@ pgMask_GetBuffer(pgMaskObject *self, Py_buffer *view, int flags)
     view->internal = bufinfo;
     view->shape = bufinfo->shape;
     view->strides = bufinfo->strides;
+    if (flags & PyBUF_FORMAT) {
+        view->format = "L"; /* L = unsigned long */
+    }
 
     Py_INCREF(self);
     view->obj = self;
--- a/src_c/mask.c
+++ b/src_c/mask.c
@@ -1767,6 +1767,9 @@ pgMask_ReleaseBuffer(pgMaskObject *self, Py_buffer *view)
 }
 
 static PyBufferProcs pgMask_BufferProcs = {
+#if PY2
+    NULL, NULL, NULL, NULL,
+#endif
     (getbufferproc)pgMask_GetBuffer,
     (releasebufferproc)pgMask_ReleaseBuffer
 };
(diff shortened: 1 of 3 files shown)


src_c/mask.c
```diff
@@ -1708,6 +1708,72 @@ mask_init(PyObject *self, PyObject *args, PyObject *kwargs)
...
+} mask_bufinfo;
+
+static int
+pgMask_GetBuffer(pgMaskObject *self, Py_buffer *view, int flags)
+{
+    bitmask_t *m = self->mask;
+    mask_bufinfo *bufinfo = (mask_bufinfo*)self->bufdata;
+
+    if (bufinfo == NULL) {
+        bufinfo = PyMem_Malloc(sizeof(mask_bufinfo));
```
illume: I think there should be an error check here for PyMem_Malloc... something like this?
```c
if (bufinfo == NULL)
    return -1;
```

dlon:
Not sure how to use multidimensional memory views in Py 2.7, so skipping for now. Do we care about Python 2.7 anymore?

notpygame:
2.7 usage of pygame according to pypi is 13% (Django 25%, Twisted 60%)

For anything that takes extra work, it's probably ok to skip python2 support at this stage. It would be good to get pygame 2 out with python 2 support.

illume:
There's some crash on 3.5, and 3.7 python builds on windows. I'm rerunning the builds in case it was some unrelated thingo.

illume:
Cool... you fixed those windows issues :)

I'll merge this in now. If someone else wants to do a python2.7 version of this at some other time then they can.

PR #1061, opened 2019-05-21
SDL2: Renderer.draw*, and Renderer.read_surface. Texture.update(surface)

husano896:
Texture:

    # Update Texture with Surface using SDL_UpdateTexture
    # Only do typecheck now, the surface should be in the same format & size of texture
    def update(self, surface, area=None)

Renderer:

    # Drawing functions using Renderer
    def draw_line(self, p1, p2)
    def draw_point(self, point)
    def draw_rect(self, rect)
    def fill_rect(self, rect)

    # Read Pixels using SDL_RenderReadPixels from target texture / window and create a surface
    # This shouldn't frequently used according to the SDL docs
    def read_surface(self, depth=32)

<details><summary>Test Code</summary>

    import pygame as pg
    from pygame._sdl2 import Window, Renderer, Texture

    win = Window()
    renderer = Renderer(win)

    renderer.draw_color = (255,255,255, 255)

    # New drawing functions
    renderer.draw_line((0,0), (64,64))
    renderer.draw_line((64,64), (128,0))
    renderer.draw_point((72,32))
    renderer.draw_rect(pg.Rect(0, 64, 64, 64))
    renderer.fill_rect(pg.Rect(0, 128, 64, 64))

    surf = pg.Surface((64,64))
    surf.fill((0,255,0))

    # This should draw a green rect
    tex = Texture.from_surface(renderer, surf)
    tex.draw(None, pg.Rect(64, 64, 64, 64))

    # This should update the texture with a surface filled with red
    # Instead of creating texture every frame, use this will be less expensive
    surf.fill((255,0,0))
    tex.update(surf)
    tex.draw(None, pg.Rect(64, 128, 64, 64))

    renderer.present()

    # Read pixels into surface and save it, can be used like in-game screenshot
    readsurf = renderer.read_surface()
    pg.image.save(readsurf, "test.png")

</details>

<details><summary>Result image</summary>

image

</details>


src_c/_sdl2/video.pyx
```diff
@@ -903,3 +927,57 @@ cdef class Renderer:
...
+
+        cdef SDL_Texture* tex = SDL_GetRenderTarget(self._renderer)
+        cdef int width, height
+        if tex == NULL:
+            # NULL for default target (the window that created the renderer)
+            width, height = self._win.size
+        else:
+            SDL_QueryTexture(tex, &format, NULL, &width, &height)
+
+        cdef SDL_Surface *surface = SDL_CreateRGBSurfaceWithFormat(0, width, height, 32, format);
```
dlon: Shouldn't it be `depth` rather than 32?

src_c/_sdl2/video.pyx
```diff
@@ -686,6 +686,29 @@ cdef class Texture:
...
+
+        :param surface: source Surface.
+        """
+
+        if not pgSurface_Check(surface):
+            raise error("update source should be a Surface.")
+            
+        cdef SDL_Surface *surf = pgSurface_AsSurface(surface)
+
+        res = SDL_UpdateTexture(self._tex, &surf.clip_rect, surf.pixels, surf.pitch)
```
dlon: Is `surf.clip_rect` used when blitting *to* `surf`? If so, it should probably use an argument instead

src_c/_sdl2/video.pyx
husano896: fixed.

src_c/_sdl2/video.pyx
husano896: Yea, should have an 'area' argument.

illume:
Looks good :)

There's an issue on the mac builds. I'll try and reproduce it myself to fix.
```
^~~~~~~
src_c/_sdl2/video.c:15231:46: error: too many arguments provided to function-like macro invocation
  __pyx_t_7 = pgSurface_New(__pyx_v_surface, 1); if (unlikely(!__pyx_t_7)) __PYX_ERR(0, 985, __pyx_L1_error)
                                             ^
src_c/_sdl2/../include/_pygame.h:399:9: note: macro 'pgSurface_New' defined here
#define pgSurface_New(surface) pgSurface_New2((surface), 1)
        ^
src_c/_sdl2/video.c:15231:15: error: use of undeclared identifier 'pgSurface_New'
  __pyx_t_7 = pgSurface_New(__pyx_v_surface, 1); if (unlikely(!__pyx_t_7)) __PYX_ERR(0, 985, __pyx_L1_error)
              ^
```

illume:
What do you think about `renderer.line` instead of `renderer.draw_line`?

illume:
I wonder what does this line do? `tex.draw(None, pg.Rect(64, 128, 64, 64))`

dlon:
Alternatively, the draw module could be used for this? Eg `pygame.draw.line(renderer, color, start_pos, end_pos)`

husano896:
> What do you think about `renderer.line` instead of `renderer.draw_line`?

I would like to preserve the word from SDL_Render**Draw** or SDL_Render**Fill** actually.


> I wonder what does this line do? tex.draw(None, pg.Rect(64, 128, 64, 64))

This draws a rect at (64, 128) with size (64,64), you can see a red rect from the result image.

before drawing, I actually fill the surface with red color, and update it to the texture.

```
surf.fill((255,0,0))
tex.update(surf)
tex.draw(None, pg.Rect(64, 128, 64, 64))
```


> Alternatively, the draw module could be used for this? Eg pygame.draw.line(renderer, color, start_pos, end_pos)

That's a good idea if every functions in pygame.draw accept renderer.

At this time, pygame.draw uses algorithms to draw, but some can&can't be done with SDL_RenderDraw*.

illume:
For Texture.update, we might need to do some pixel format conversion if the two formats do not match. 

https://wiki.libsdl.org/SDL_UpdateTexture
> The pixel data must be in the pixel format of the texture. Use SDL_QueryTexture() to query the pixel format of the texture.

I guess we need to check this ourselves with SDL_QueryTexture.

illume:
I fixed the Mac compile issue in here: husano896-sdl2-renderer

Using pgSurface_New2 rather than pgSurface_New, because somehow the pre processor didn't like it.

dlon:
For `draw` and `gfxdraw`, we could probably just vendor SDL2_gfx? It seems to have everything.

illume:
I'm going to merge this into a separate branch, and then merge my fixes and this into master. Then we can figure out the conflicts with the "move cython files" pr.

Issue #1068, opened 2019-05-25
Cythonize pygame.sprite module


dlon:
This should speed up rendering of sprites.

Cf. #605

mcpalmer1980:
I have three suggestions for the new Sprite class, two of which are interdependent on each other: one may enhance performance for a common task and the other will make it easier for users to gather useful data about collisions and program side effects. I'll begin with the unrelated item.

**Sprite Sorting**

I don't actually derive my sprite class from pygame.Sprite. Each of my handlers for enemies, allies, shots, blood, etc maintain its own list of sprites. One of the main reasons I don't use pygame.Sprite is because I need to order their drawing calls, using their y coordinate like a z-buffer to make sure that 'closer' characters are drawn after farther ones.

A bit more information in case that didn't make sense: The first game I wrote using pygame was Cooties(mcpalmer1980.itch.io/Cooties). Its plays on a top-down map with several girls chasing around the main character. The perspective makes characters drawn higher on the screen seem farther away, so if two characters overlap but the one located higher on the screen is drawn over the lower one it would look inside-out. My method works kind of like this.

	GetInput = GetJoyInput if joystick else GetKeyMouseInput
	def ysort(spr):
		return spr.y

	ti = 0  #this will be used to track time since last frame for animation purposes
	while playing:
		events = pygame.event.get()
		for e in events:
			...

		move, pressed = GetInput(events)
		ylist = player.Handle(move, pressed, ti)  #handles user input and animation
		ylist += aliens.Handle(ti)  #handles AI and animation
		ylist += friends.Handle(ti)

		tilemap.Follow(player)
		for i in sorted(ylist, key=ysort):  #sort all characters in ascending y order
			i.Draw()  #and draw the 'closer' ones last, like a z-buffer
		splats.Draw(ti)
		shots.Draw(ti)  #I don't bother sorting these sort of things
		effects.Render()  #screen transitions must go last
		ti = clock.tick(30)  

Perhaps a sort method could be added to sprite groups, which would accept reference to a sorting fuction as its parameter. Another option might be to sort by an existing attribute. Either way our programs would need to maintain a sprite group that contains all sprites that need to be sorted.

```
	def ysort(spr):
		return spr.y

	player = Player()
	enemy_group = CreateEnemies()
	ally_group = CreateAllies()
	shots = ShotHandler()
	ordered_group = pygame.sprite.Group()
	ordered_group.add(player, enemy_group, ally_group)
	while playing:
		...game logic...
[...]

PR #1066, opened 2019-05-24
WIP: C version of _sdl2.video

dlon:
It seems to be slightly faster (~9.5%) than the Cython version when drawing and is much smaller.

Missing features:
- [ ] Renderer.target
- [ ] Image
- [ ] Window
- [ ] get_drivers()
- [ ] messagebox()
- [ ] Documentation

<details><summary>Example: Render lots of sprites using Texture</summary>

```python3
import pygame
import os
data_dir = os.path.join(os.path.split(os.path.abspath(__file__))[0],
                        'data')

from pygame._sdl2 import (
    Window,
    Texture,
    Image,
    Renderer
)

def load_img(file):
    return pygame.image.load(os.path.join(data_dir, file))

pygame.display.init()
pygame.key.set_repeat(10, 10)

win = Window('asdf', resizable=True)
renderer = Renderer(win)
tex = Texture.from_surface(renderer, load_img('alien1.gif'))

class Something(pygame.sprite.Sprite):
    def __init__(self, img):
        pygame.sprite.Sprite.__init__(self)

        self.rect = img.get_rect()
        self.image = img

        self.rect.w *= 5
        self.rect.h *= 5

        if isinstance(img, Image):
            img.origin = self.rect.w / 2, self.rect.h / 2

group = pygame.sprite.Group()

for i in range(20000):
    #sprite = Something(Image(tex))
    sprite = Something(tex)
    sprite.rect.x = 10 * i % win.size[0]
    sprite.rect.w /= 2
    sprite.rect.h /= 2
    sprite.rect.y = sprite.rect.h / 2 * (10 * i // win.size[0])
    group.add(sprite)

import math
t = 0
running = True
clock = pygame.time.Clock()
renderer.draw_color = (255, 0, 0, 255)


while running:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            running = False
        elif event.type == pygame.KEYDOWN:
            if event.key == pygame.K_ESCAPE:
                running = False
            elif event.key == pygame.K_LEFT:
                sprite.rect.x -= 5
            elif event.key == pygame.K_RIGHT:
                sprite.rect.x += 5
            elif event.key == pygame.K_DOWN:
                sprite.rect.y += 5
            elif event.key == pygame.K_UP:
                sprite.rect.y -= 5

    renderer.clear()
    group.draw(renderer)
    renderer.present()

    clock.tick(60)
    win.title = str('FPS: {}'.format(clock.get_fps()))
```

</details

PR #1069, opened 2019-05-25
Cythonized sprite module

dlon:
Re: #1068 

This replaces src_py/sprite.py with src_c/sprite.pyx.

## Comparison (drawing)

| Version                                     | C renderer, average time (s) | Cython renderer, average time (s) |
|---------------------------------------------|------------------------------|-----------------------------------|
| sprite.py                                   | 0.9766                       | 1.04                              |
| sprite.pyx (no changes)                     | 0.8083                       | 0.88                              |
| Sprite extension type Any Sprite.rect type  | 0.7894                       |                                   |
| Sprite extension type Rect-type Sprite.rect | 0.7384                       | 0.883 (maybe a fluke)             |
| AbstractGroup extension type                | 0.6853                       | 0.805                             |
| Update sprite.pyx                           |                              | 0.795                             |
| Optimize sprite rendering                           |                              | 0.565                             |

"C renderer" was based on the branch `sdl2-video-c` (#1066). "Cython renderer" is compared to master.

<details> <summary>Test used. I averaged about five measurements for each version.</summary>

```python3
import pygame
import os
data_dir = os.path.join(os.path.split(os.path.abspath(__file__))[0],
                        'data')

from pygame._sdl2 import (
    Window,
    Texture,
    Image,
    Renderer
)

def load_img(file):
    return pygame.image.load(os.path.join(data_dir, file))

pygame.display.init()
pygame.key.set_repeat(10, 10)

win = Window('asdf', resizable=True)
renderer = Renderer(win)
tex = Texture.from_surface(renderer, load_img('alien1.gif'))

class Something(pygame.sprite.Sprite):
    def __init__(self, img):
        pygame.sprite.Sprite.__init__(self)

        self.rect = img.get_rect()
        self.image = img

        self.rect.w *= 5
        self.rect.h *= 5

        if isinstance(img, Image):
            img.origin = self.rect.w / 2, self.rect.h / 2

group = pygame.sprite.Group()

for i in range(20000):
    #sprite = Something(Image(tex))
    sprite = Something(tex)
    sprite.rect.x = 10 * i % win.size[0]
    sprite.rect.w /= 2
    sprite.rect.h /= 2
    sprite.rect.y = sprite.rect.h / 2 * (10 * i // win.size[0])
    group.add(sprite)

import timeit
print('time:', timeit.timeit(lambda: group.draw(renderer), number=100))
```

</details>

100k objects in a group (same texture) are rendered at 30 FPS. It would likely be improved further by calling `Renderer.blit()` directly, as mentioned below.

## Issues

* sprite.rect can only be assigned a `Rect` now. That might cause issues in case someone's using a tuple.
* Probably a lot of stuff could be improved.
* Collision detection: Could it be sped up using quadtrees or some other method? Runtime for testing every pair is currently n^2.
- [ ] `Renderer.blit()` could be declared with `cdef` or `cpdef` and called directly in `sprite.AbstractGroup`, to avoid using `PyObject_CallFunctionObjArgs` for every iteration.
*I tried this*: It brought down the interval to 0.421 s, but there are issues using cimport when the root is src_c.
- [x] sprite.pyx -> _sprite.pyx
- [ ] Update other group types.


illume:
Very interesting results indeed!

I wonder what the issues with src_c are?

Collision detection can definitely be done in faster ways. Especially in game specific ways. A grid in a platformer for example is very good for dynamically updating and checking collisions for things that are on screen. Then there are more complex collision detection systems like pymunk. Perhaps a spatial hash can be dropped in there for the cases when there is a large Number of sprites to check.

The Layered group uses a heuristic to guess at an optimal way to draw, either full screen updates or dirty rect updates. Maybe this will need some investigation adjustment in a SDL2/sprite.pyx world.

The sprite code doesn't have the best test coverage, but there are quite a few users of it. So those could be used as some sort of test.

Having it as a _sprite.pyx might be a good idea. This would allow it to coexist with sprite.py until 

There's a benchmark written by the pyarcade author that might be useful:
http://arcade.academy/performance_tips.html#drawing-stationary-performance

illume:
The test appveyor failure is that annoying one the intermittently fails.
```
  File "C:\Python27-x64\lib\site-packages\pygame\tests\mixer_test.py", line 139, in test_sound_args
    self.assertEqual(get_bytes(mixer.Sound(wave_path)), snd_bytes)
```

dlon:
`Renderer` speed is disappointing:



I should compare .py to .pyx as well.

Renderer stress test: https://gist.github.com/dlon/85be7654d0781a1a70b14cc0f815ac33
(Somewhat unfair because it doesn't draw any text.)

dlon:
I accidentally loaded a texture for each coin in the last comment. It looks better with batching:

illume:
Nice one :) 

Is this with SDL 2.0.9?

dlon:
Yes. And here's using the OpenGL backend (the default is Direct3D):

illume:
Cool.

SDL 2.0.10 has batching, and is supposed to improve performance a lot.
> So, even on those desktop PCs with the forgiving GPUs, performance went up, across all backends, by 2x to 5x. 20,000 smileys on Direct3D 11 went from 70fps to 240. OpenGL went from 40fps to 120. Metal went from 80 to 190. And so on.

So with your work, I'm guessing it will be close to using OpenGL display lists(like how pyarcade does).

dlon:
Ah, I assumed it had it already. Probably VBOs are even faster than display lists too?

dlon:
I think squash and merge this if it's fine. More updates can come in separate PRs

PR #1075, opened 2019-05-27
Update .pyx directory structure

dlon:
* Move all the .pyx files into src_c/cython/pygame/ in order for cimport to work properly.
* Update setup.py to explicitly output the C files relative to src_c/ (not src_c/pygame/).
* The module names are now correctly recognized as `pygame._sdl2.video`, etc. by Cython.


setup.py
```diff
@@ -88,14 +88,50 @@ def compilation_help():
...
     # So you can `setup.py cython` or `setup.py cython install`
     try:
-        from Cython.Build import cythonize
+        from Cython.Build.Dependencies import cythonize_one
     except ImportError:
         print("You need cython. https://cython.org/, pip install cython --user")
+        sys.exit(1)
+
+    import glob
+    pyx_files = glob.glob(os.path.join('src_c', 'cython', 'pygame', '*.pyx')) + \
```
illume: Does this handle `.pxd` files changing as well?

dlon: Fixed. It would be useful if this also worked for the regular header files.

dlon:
It would be easiest to merge this after #1077.

PR #1076, opened 2019-05-28
Old raspberries... See if Mix_OpenAudioDevice is defined before using it.

illume:
Mix_OpenAudioDevice is not in the sdl2_mixer on the raspberry pi.

### DIFF
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -454,7 +454,7 @@ _init(int freq, int size, int channels, int chunk, char *devicename, int allowed
         if (SDL_InitSubSystem(SDL_INIT_AUDIO) == -1)
             return PyInt_FromLong(0);
 
-#if IS_SDLv2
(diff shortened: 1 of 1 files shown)


src_c/mixer.c
```diff
@@ -454,7 +454,7 @@ _init(int freq, int size, int channels, int chunk, char *devicename, int allowed
         if (SDL_InitSubSystem(SDL_INIT_AUDIO) == -1)
             return PyInt_FromLong(0);
 
-#if IS_SDLv2
+#ifdef Mix_OpenAudioDevice
```
dlon: Won't this always be false?

illume: Yes, oops. I fixed it in https://github.com/pygame/pygame/pull/1078/commits/1814c96bebc899febd44c6541b5f00a52ae30ccb

PR #1077, opened 2019-05-28
SDL2: Renderer.draw*, and Renderer.to_surface. Texture.update(surface) part 2.

illume:
Continuing on from https://github.com/pygame/pygame/pull/1061 in a pygame repo branch so we can more easily collaborate.


src_c/_sdl2/video.pyx
```diff
@@ -706,6 +706,31 @@ cdef class Texture:
...
+        for optimization reasons you may not get the pixels back if you lock the texture afterward.
+
+        :param surface: source Surface.
+        """
+
+        if not pgSurface_Check(surface):
+            raise error("update source should be a Surface.")
+
+
+        cdef SDL_Rect rect = pgRect_AsRect(area)
```
illume: pgRect_AsRect(area) could fail, need to return an error here if it does.

illume:
In the future pygame.image.save could take a Renderer.

src_c/_sdl2/video.pyx
```diff
@@ -940,3 +966,57 @@ cdef class Renderer:
...
+
+    def fill_rect(self, rect):
+        # https://wiki.libsdl.org/SDL_RenderFillRect
+        cdef SDL_Rect _rect = pgRect_AsRect(rect)
+        res = SDL_RenderFillRect(self._renderer, &_rect)
+
+        if res < 0:
+            raise error()
+
+    def get_surface(self, depth=32):
```
illume: - Maybe get_surface/read_surface should be to_surface() instead? We have from_surface in Texture.

- For the use case of saving the Renderer to disk/network every frame... Having it take a destsurface would be pretty fast, because it could allow you to reuse a surface and not malloc it every frame.

src_c/_sdl2/video.pyx
```diff
@@ -940,3 +966,57 @@ cdef class Renderer:
...
+        except ValueError as e:
+            raise e
+
+        cdef SDL_Texture* tex = SDL_GetRenderTarget(self._renderer)
+        cdef int width, height
+        if tex == NULL:
+            # NULL for default target (the window that created the renderer)
+            width, height = self._win.size
+        else:
+            SDL_QueryTexture(tex, &format, NULL, &width, &height)
```
illume: SDL_QueryTexture can fail

src_c/_sdl2/video.pyx
```diff
@@ -706,6 +706,31 @@ cdef class Texture:
...
+        """
+
+        if not pgSurface_Check(surface):
+            raise error("update source should be a Surface.")
+
+
+        cdef SDL_Rect rect = pgRect_AsRect(area)
+        cdef SDL_Surface *surf = pgSurface_AsSurface(surface)
+
+        res = SDL_UpdateTexture(self._tex, &rect, surf.pixels, surf.pitch)
```
dlon: > If the texture is intended to be updated often, it is preferred to create the texture as streaming and use the locking functions referenced below. While this function will work with streaming textures, for optimization reasons you may not get the pixels back if you lock the texture afterward.

https://wiki.libsdl.org/SDL_UpdateTexture

Maybe it's better to use `SDL_LockTexture()` here? At least if it's a streaming texture.

examples/video.py
```diff
@@ -88,12 +98,46 @@ def load_img(file):
...
+    # TODO: use this from_surface somehow.
+    # surf = pg.Surface((64,64))
+    # surf.fill((0,255,0))
+    # # This should draw a green rect
+    # tex = Texture.from_surface(renderer, surf)
+    # tex.draw(None, pg.Rect(64, 64, 64, 64))
+    # # This should update the texture with a surface filled with red
+    # # Instead of creating texture every frame, use this will be less expensive
+    # surf.fill((255,0,0))
+    # tex.update(surf)
```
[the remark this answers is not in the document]
dlon: Also, it would likely be a lot faster to create a separate texture for each surface before the loop starts than to update a single one.

src_c/_sdl2/video.pyx
dlon: I prefer either read_surface() or to_surface() to get_surface(), because the latter might make it sound like it's just returning a stored surface if you don't know how it works

dlon: `destsurface=None`?

husano896: I prefer read_surface() than get_surface() since it sounds like you can get the **surface reference** of the texture, like how `pygame.display.get_surface()` does.

`def read_surface(dest=None)`,`def read_surface(surf=None)`or`def read_surface(surface=None)`?

husano896: I think those drawing functions can be merged into pygame.draw or pygame.gfxdraw, too.

illume: Good point about the get_surface() perhaps seeming like it would return a reference. We also have pygame.image.tostring, frombuffer, fromstring.

dlon: I guess to_surface is a bit more consistent with Texture.from_surface

examples/video.py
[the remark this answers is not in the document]
husano896: Is every texture need to be locked before SDL_UpdateTexture, or only those with `streaming=True`?

Also yes, create separate texture for each surface will be faster, but the example is to show you can update a texture with surface.

src_c/_sdl2/video.pyx
husano896: Then `to_surface()` seems like the best at least moment.

src_c/_sdl2/video.pyx
husano896: This will be done with surface size & format check.

src_c/_sdl2/video.pyx
```diff
@@ -940,3 +966,57 @@ cdef class Renderer:
...
+    def draw_point(self, point):
+        # https://wiki.libsdl.org/SDL_RenderDrawPoint
+        res = SDL_RenderDrawPoint(self._renderer,
+                                  point[0], point[1])
+        if res < 0:
+            raise error()
+
+    def draw_rect(self, rect):
+        # https://wiki.libsdl.org/SDL_RenderDrawRect
+        cdef SDL_Rect _rect = pgRect_AsRect(rect)
```
dlon: Segfaults when `rect` is not a pgRectObject.

src_c/_sdl2/video.pyx
```diff
@@ -940,3 +966,57 @@ cdef class Renderer:
...
+        res = SDL_RenderDrawRect(self._renderer, &_rect)
+        if res < 0:
+            raise error()
+
+    def fill_rect(self, rect):
+        # https://wiki.libsdl.org/SDL_RenderFillRect
+        cdef SDL_Rect _rect = pgRect_AsRect(rect)
```
dlon: Segfaults when `rect` is not a pgRectObject.

src_c/_sdl2/video.pyx
dlon: Segfaults when `rect` is not a pgRectObject. It has to use `pgRect_FromObject`

examples/video.py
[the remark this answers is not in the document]
dlon: Locking would be used instead of `SDL_UpdateTexture()` for streaming textures. The latter might not update the pixel array properly in case we ever add direct pixel access to `Texture`, according to the wiki. But it can be done later?

dlon:
* Renamed read function to `Renderer.to_surface()`: `to_surface(self, surface=None, area=None)`
* Added a `surface` parameter to to_surface, so that one can read the pixels into an existing surface.
* Added an `area` parameter for reading only part of the viewport into the surface.

src_c/_sdl2/video.pyx
```diff
@@ -940,3 +974,98 @@ cdef class Renderer:
...
+            surface = pgSurface_New2(surf, 1)
+        elif pgSurface_Check(surface):
+            surf = pgSurface_AsSurface(surface)
+            if surf.w != rarea.w or surf.h != rarea.h:
+                raise ValueError('surface size must equal the area')
+            format = surf.format.format
+        else:
+            raise TypeError("'surface' must be a surface or None")
+
+        if SDL_RenderReadPixels(self._renderer,
```
illume: With [SDL_RenderReadPixels](https://wiki.libsdl.org/SDL_RenderReadPixels), I can confirm that it clips the area rect for us if it is too big.
https://github.com/SDL-mirror/SDL/blob/a9b1e5ee66144d414f342b54b6622590b730e68d/src/render/SDL_render.c#L3087

However, it returns 0 on success or a negative error code on failure. 

So it should be:
```python
if SDL_RenderReadPixels(...) < 0:
    raise error()
```

examples/video.py
```diff
@@ -88,12 +96,48 @@ def load_img(file):
                 else:
                     win.set_windowed()
                     full = 0
+            elif event.key == pygame.K_s:
+                readsurf = renderer.get_surface()
```
illume: The example needs updating to use `renderer.to_surface()`.

dlon:
Had to clip it manually to check the surface size. I've manually tested most cases:
* surface=None and surface set, with and without an area set
* surface is smaller than the area (ValueError)
* surface is larger than the area (the pixels are just in the top-left corner)

src_c/_sdl2/video.pyx
```diff
@@ -940,3 +977,110 @@ cdef class Renderer:
...
+        cdef SDL_Texture *targettex
+
+        # obtain area to use
+        if area is not None:
+            if pgRect_FromObject(area, &rarea) == NULL:
+                raise TypeError('area must be None or a rect')
+
+            # clip area
+            SDL_RenderGetViewport(self._renderer, &tempviewport)
+            SDL_IntersectRect(&rarea, &tempviewport, &rarea)
```
dlon: What should happen if there is no intersection?

dlon: The current behavior is probably right: Return a new 0x0 surface if no surface is passed, or else just return the passed-in surface.

PR #1097, opened 2019-06-07
Add copy() method to Mask class

### LINKED ISSUE #1086: Add copy() method to Mask class
Add support for a `copy()` method to the `Mask` class. It would return a new `Mask` object that is a copy (same dimensions and same bits set/unset) of the calling mask.

charlesej:
Overview of changes:
- Added a `copy()` method to the `Mask` class
- Added a `copy()` entry to the `Mask` documentation
- Added `copy()` tests
- Renamed a `mask.c` function to get rid of leading underscore
- Converted a `MaskTypeTest` method into a general function to allow other testcases to use it

System details:
- os: windows 10 (64bit)
- pygame: 2.0.0.dev1 (SDL: 1.2.15) at 90344c4c9c64f9ecdbe3c62837a41e17a43a1274

Resolves #1086.

### DIFF
--- a/docs/reST/ref/mask.rst
+++ b/docs/reST/ref/mask.rst
@@ -129,6 +129,9 @@ to store which parts collide.
    :returns: a newly created :class:`Mask` object
    :rtype: Mask
 
+   .. versionchanged:: 2.0.0
+      Shallow copy support added. The :class:`Mask` class supports the special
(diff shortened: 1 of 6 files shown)


charlesej:
@dlon Thanks for the review. I have updated the mask documentation based on your suggestion.

PR #1106, opened 2019-06-15
Compile with SDL2 by default.  Use SDL2 in manylinux builds.

illume:
- SDL2 is the default build for pygame now. You can use python setup.py install -sdl1 to build with SDL1.
- The `manylinux` builds use SDL2
- pgcompat.c for functions not available in less than SDL 2.0.5. Not sure how to ifdef in cython, so implement functions that fail(return -1/NULL) or do nothing.

closes https://github.com/pygame/pygame/issues/905


husano896:
instead of # ifdef everything,

how about define a minimum required SDL version?

    #if !(SDL_VERSION_ATLEAST(2, 0, 5))
    #pragma error "SDL 2.0.5 or above is required"
    #end

src_c/pgcompat.c
```diff
@@ -0,0 +1,43 @@
+#include <SDL.h>
+
+#if defined(SDL_VERSION_ATLEAST)
+#if !(SDL_VERSION_ATLEAST(2, 0, 5))
+/* this function is only defined in 2.0.5.
```
charlesej: Rewording`"this function is only defined in 2.0.5."` to something like `"These functions require SDL 2.0.5 or greater."` might be clearer.

src_c/pgcompat.c
```diff
@@ -0,0 +1,43 @@
...
+  https://wiki.libsdl.org/SDL_SetWindowResizable
+*/
+void SDL_SetWindowResizable(SDL_Window* window,
+                            SDL_bool    resizable) {
+
+}
+int SDL_GetWindowOpacity(SDL_Window* window,
+                         float*      opacity) {
+    SDL_SetError("SDL 2.0.5 required!");
```
charlesej: Changing the error text (and all the other similar ones) from `"SDL 2.0.5 required!"` to something like `"SDL 2.0.5 or greater required"` might be more accurate.

src_c/pgcompat.c
```diff
@@ -0,0 +1,43 @@
...
+    return -1;
+}
+SDL_Surface* SDL_CreateRGBSurfaceWithFormat(Uint32 flags,
+                                            int    width,
+                                            int    height,
+                                            int    depth,
+                                            Uint32 format)
+{
+    SDL_SetError("SDL 2.0.5 required!");
+	return NULL;
```
charlesej: You might want to run clang-format on this file.

src_c/include/_pygame.h
```diff
@@ -80,9 +80,31 @@
 #include "pgcompat.h"
 
 #if defined(SDL_VERSION_ATLEAST)
+
 #ifndef SDL_WINDOW_VULKAN
 #define SDL_WINDOW_VULKAN 0
 #endif
+
```
dlon: Could all of these defs be in src_c/_pygame.h or src_c/pgcompat.h?

illume:
> instead of # ifdef everything,
> 
> how about define a minimum required SDL version?
> 
> ```
> #if !(SDL_VERSION_ATLEAST(2, 0, 5))
> #pragma error "SDL 2.0.5 or above is required"
> #end
> ```

@husano896 Good idea. I'm not sure what the minimum version that works yet is. Also I'm not sure how widespread older SDL 2 versions are. I'll look into these later.

illume:
Thanks for the review notes :)

Issue #574, opened 2018-10-20
Addition of a lerp (linear interpolation) method to the Color class


Skorpio-X:
A lerp method (like that of the `Vector` classes) would be useful, so that colors can be interpolated in a cleaner and easier way.

Also, the Godot engine's `Color` has a `blend` method and some more which could be useful additions to pygame's Color class as well. 

http://docs.godotengine.org/ko/latest/classes/class_color.html
https://github.com/godotengine/godot/blob/master/core/color.h

illume:
Yeah, I think color blending could be useful in the Color class.
Do you have a specific use case in mind?

Skorpio-X:
I'd probably use that mainly for little interactive art programs (I don't think "demo" is the right term) or things like text that changes its color cyclically.

I would also find it very nice if I could change the hsla and hsva values of colors separately. At the moment you have to unpack them into separate variables (`h, s, v, a = color.hsva`), change one value and reassign them. `color.hue = 75` or `+= 1` would be much nicer.

illume:
That sounds nice too.

MyreMylar:
I think this needs adding to both the Color class, the Vector2 & Vector3 classes and also probably as a function in the math module for single float values. 

I find a lerp/blend/mix is one the most commonly recreated non-pygame functions I use when starting a from-scratch pygame project.

For additional support, game programmer Bruce Hill (formerly of Maxis) includes a mix() function as one of his '6 useful snippets' here:

https://blog.bruce-hill.com/6-useful-snippets

Which includes a simple graphical demo of it in action.

charlesoblack:
I wrote the following code snippet:

    static PyObject *
    color_lerp(pgColorObject *self, pgColorObject *color2, PyObject *amt)
    {
        Uint8 rgab[4];

        if (!PyObject_IsInstance(color2, (PyObject *)&pgColor_Type)) {
            Py_INCREF(Py_NotImplemented);
            return Py_NotImplemented;
        }

        if (amt < 0 || amt > 1) {
            PyErr_SetString(PyExc_ValueError,
                            "Argument 2 must be in range [0, 1]");
            return NULL;
        }

        rgba[0] = (Uint8)(self->data[0] * amt + color2->data[0] * (1 - amt))
        rgba[1] = (Uint8)(self->data[1] * amt + color2->data[1] * (1 - amt))
        rgba[2] = (Uint8)(self->data[2] * amt + color2->data[2] * (1 - amt))
[...]

PR #1115, opened 2019-06-23
Add Color.lerp() function for linear interpolation between two colors.

charlesoblack:
This PR adds a `Color.lerp()` function that works similarly to the `Vector2.lerp()` function, fixing issue #574. It should take tuples or Colors as its first argument (keyword `color`), and a float bounded by 0 and 1 inclusive as its second argument (keyword `amount`). The linear interpolation returns a new `Color` object that is identical to self if `amount` was 0, or identical to other if `amount` was 1.

I've added tests and updated the documentation. I set the version added to 2.0.1, in case we don't get it out for 2.0.0. As for the tests, they should verify whether the math works in the first place, whether the returned object is in fact `pygame.Color` type, whether extreme values for `amount` work (0, 0.01, 0.99, 1), whether the function does accept kwargs, and whether errors are raised properly with invalid input. The same tests are run for a subclass of `pygame.Color`.

Additionally, I had found a couple of typos in the `math.rst` documentation file, which I've fixed.

Please let me know if anything else should be done!


charlesoblack:
I'm not sure why the 32bit Python 2.7 build fails on AppVeyor. Could I get some guidance? The code looks fine to me, and the 64bit version passes, as well as the exact same version on Linux. :(

illume:
Thanks :)

charlesoblack:
Thank you!

PR #1118, opened 2019-06-25
Add links between gfxdraw and draw docs

charlesej:
Resolves sub-item "Add a link to `pygame.gfxdraw` as an alternative to `pygame.draw`." of item "Update the `pygame.draw` documentation. " of #896.


illume:
Good idea.

PR #1147, opened 2019-07-13
Support path like objects in image.load(), fixes #1129

### LINKED ISSUE #1129: pygame.image.load() does not work with pathlib.Path objects

- [x] unit test which fails to load pathlib.Path objects (**good first issue**). Mark with @unittest.expectedFailure decorator.
- [x] step 2 in a second PR: implement support for pathlib.Path objects in image.load.
- [ ] link to pathlib in docs? (see discussion at https://github.com/pygame/pygame/pull/2387)
- [ ] use param notation in docs for types and descriptions of params?

---
image.load() does not accept pathlib.Path objects, needs a plain string

**Related Docs:** https://www.pygame.org/docs/ref/image.html#pygame.image.load

isidentical:
Support path like objects in image.load(),

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -72,7 +72,7 @@ image_load_basic(PyObject *self, PyObject *arg)
         return NULL;
     }
 
-    oencoded = pg_EncodeString(obj, "UTF-8", NULL, pgExc_SDLError);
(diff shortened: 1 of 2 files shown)


illume:
It looks like PyOS_FSPath is new on 3.6 
https://docs.python.org/3/c-api/sys.html#c.PyOS_FSPath

Issue #477, opened 2018-07-24
User events clash


Dreamsorcerer:
- [x] Write an example of using the API as if it already existed. (good first issue)
- [x] write a unit test for new API and mark with expected failure.

--- 

When selecting a user event a developer currently has to select any value from USEREVENT to NUMEVENTS. When developing libraries, a developer may need to select a custom event to use for their library. A problem occurs when different libraries have selected the same event type or the game is using the same event type as a library, and they end up processing each others events.

To avoid this, it would be nice to have a mechanism do get a new event type from Pygame, without clashing. This could be something as simple as:
`CUSTOM_EVENT = iter(range(pygame.constants.USEREVENT, pygame.constants.NUMEVENTS))`
which can then be used in a game/library like:
`OUR_EVENT = next(CUSTOM_EVENT)`

If this is added for Pygame 2, then we can even consider deprecating the use of pygame.constants.USEREVENT and removing it from the documentation, ensuring that new developers all follow the new convention.

mar77i:
And then there was this: I keep redefining TIMER in each program I write.

illume:
Yop. Good idea.
Something like below for the API?

```python
ZOG_UP = pg.event.custom_type()
ZOG_DOWN = pg.event.custom_type()
pg.event.post(pg.Event(ZOG_DOWN, msg='hello'))
```

The other problem is that people run out of custom events when they use them this way.
Especially if they think they can do all logic for all game events.

So maybe we should think of an API that lets people do that. I usually just use another attribute to switch on. Something like this:
```python
CUSTOM = pg.event.custom_type()
pg.event.post(pg.Event(CUSTOM, say='hello'))
pg.event.post(pg.Event(CUSTOM, do='nothing'))

if hasattr(e, 'do'):
   print ('doing it')
elif hasattr(e, 'say'):
    print(e.say) 
```
But it's still awkward.

Maybe we can fix it underneath at the C level. So we have a map, and it translates it in and out of what is given to SDL. So at the python level we don't have that limit anymore.

Dreamsorcerer:
The second method is what I do as well (except using a single attribute as a subtype), and I think that's absolutely fine. e.g.

```
Event(CUSTOM, action="say", msg="hello")
Event(CUSTOM, action="do")

if e.action == "do":
  print("doing it")
elif e.action == "say":
  print(e.msg)
```

It's just the clash between different libraries/games that try to use the same event type that bothers me.

[...]

Issue #1056, opened 2019-05-19
There is not enough possible event types


Jajasek:
I love pygame and I am creating a library extension to it [(https://github.com/Jajasek/pygame_widgets)](https://github.com/Jajasek/pygame_widgets) and I have figured out that the Event.type can be only in range 0 to 31, and that is for my project really little. I think it seriously limites the size of the project, that can be created in pygame. I would like to extend it to at least 1 byte.

dlon:
Hi. If I recall correctly, pygame's currently limiting the no. of user events to 8 for compatibility reasons. I don't know if that might eventually change with pygame 2.

It's possible to work around it, though, eg by adding another event type field in your user event.

illume:
Yop. I think 8192 events work in SDL2. From my testing so far anyway.

pygame 2.0.0.dev3 will have this.

illume:
pygame 2.0.0.dev3 is out now with this.

PR #1151, opened 2019-07-13
pygame.event.custom_type to prevent user events clashing. More user events in SDL2.

illume:
For https://github.com/pygame/pygame/issues/477


charlesej:
If this PR is increasing the number of events it will address issue #1056.

PR #1168, opened 2019-07-16
Clarify that Color.lerp works in RGBA space

takluyver:


robertpfeiffer:
We could add parameters to control whether color lerp works in RGBA space and whether it does gamma correction, like `mycolor.lerp(othercolor, 0.6, gamma=1.4)` or  `mycolor.lerp(othercolor, 0.6, hsv_space=True)`

The doc change looks good to me, but I'm not sure documenting the current behaviour is the right long-term solution.

takluyver:
Yup, I suggested on the mailing list the possibility of adding support for interpolation in other colour spaces. I think it would be interesting to use one based on human perception, like CAM02-UCS which was used to design matplotlib's new colourmaps.

But I don't have the time to work on this myself, so for now I thought it useful to document what it does at present.

robertpfeiffer:
I wrote this: https://github.com/robertpfeiffer/pygame/tree/lerp-gamma
based on a YouTube video I watched a while ago: https://www.youtube.com/watch?v=LKnqECcg6Gw

But this is not exactly blending in CAM02-UCS. I had to google it. It seems to be an update of the CIE LAB colour space I heard about before.

illume:
@robertpfeiffer let's discuss further in discord?

takluyver:
I still think it would be worthwhile to document that `.lerp` currently works in RGBA, as this PR does.

illume:
Sorry, I didn't mean to close it.

illume:
(also... thanks for the PR :) )

PR #1175, opened 2019-07-19
simple scrap module based on SDL2 clipboard

robertpfeiffer:
Currently the master branch doesn't build the scrap module when you compile with SDL2. This is a simple replacement for the platform-specific scrap modules based on the cross-platform SDL2 clipboard functions. It can only handle utf-8 plaintext and nothing else.

Can somebody look at the code and tell me whether I have to used memcpy instead of strdup? I'm not sure if mixing null-terminated strings and length fields breaks anything besides wasting a single byte of memory.

### DIFF
--- a/src_c/scrap_sdl2.c
+++ b/src_c/scrap_sdl2.c
@@ -1,30 +1,29 @@
 #include "SDL.h"
 #include "SDL_clipboard.h"
 
-char* plaintext_type= "text/plain;charset=utf-8";
+char *plaintext_type = "text/plain;charset=utf-8";
 
 int
-pygame_scrap_contains(char *type){
-    return (strcmp(type, plaintext_type) == 0)
-        && SDL_HasClipboardText();
+pygame_scrap_contains(char *type)
+{
+    return (strcmp(type, plaintext_type) == 0) && SDL_HasClipboardText();
 }
 
-char*
-pygame_scrap_get(char *type, unsigned long *count){
+char *
+pygame_scrap_get(char *type, unsigned long *count)
+{
     char *retval = NULL;
     char *clipboard = NULL;
-
 
     if (!pygame_scrap_initialized()) {
         PyErr_SetString(pgExc_SDLError, "scrap system not initialized.");
         return NULL;
     }
 
-    if  (strcmp(type, plaintext_type) == 0){
-        printf("type OK\n");
+    if (strcmp(type, plaintext_type) == 0) {
         clipboard = SDL_GetClipboardText();
-        if (clipboard != NULL){
-            *count=strlen(clipboard);
+        if (clipboard != NULL) {
+            *count = strlen(clipboard);
             retval = strdup(clipboard);
             SDL_free(clipboard);
             return retval;
@@ -33,9 +32,10 @@ pygame_scrap_get(char *type, unsigned long *count){
     return NULL;
 }
 
-char**
-pygame_scrap_get_types(void){
-    char** types;
+char **
+pygame_scrap_get_types(void)
+{
+    char **types;
     types = malloc(sizeof(char *) * 2);
     if (!types)
         return NULL;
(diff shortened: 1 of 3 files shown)


robertpfeiffer:
The alternative would be ~~scrapping~~ deprecating `scrap` completely and writing a new module that does less, but in a more consistent and portable manner.

src_c/scrap_sdl2.c
```diff
@@ -0,0 +1,73 @@
...
+    char *clipboard = NULL;
+
+
+    if (!pygame_scrap_initialized()) {
+        PyErr_SetString(pgExc_SDLError, "scrap system not initialized.");
+        return NULL;
+    }
+
+    if  (strcmp(type, plaintext_type) == 0){
+        printf("type OK\n");
```
illume: printf is in here

src_c/scrap_sdl2.c
```diff
@@ -0,0 +1,73 @@
...
+        clipboard = SDL_GetClipboardText();
+        if (clipboard != NULL){
+            *count=strlen(clipboard);
```
illume: needs clang format run on the code

src_c/scrap_sdl2.c
```diff
@@ -0,0 +1,73 @@
+#include "SDL.h"
+#include "SDL_clipboard.h"
+
+char* plaintext_type= "text/plain;charset=utf-8";
```
illume: CONSTANT?

src_c/scrap_sdl2.c
```diff
@@ -0,0 +1,73 @@
...
+int
+pygame_scrap_init(void){
+    SDL_Init(SDL_INIT_VIDEO);
+    _scrapinitialized = 1;
+    return _scrapinitialized;
+}
+
+int
+pygame_scrap_lost(void){
+    return 1;
```
illume: How does [SDL_HasClipboardText](https://wiki.libsdl.org/SDL_HasClipboardText) relate to scrap_lost?

src_c/scrap_sdl2.c
```diff
@@ -0,0 +1,73 @@
...
+}
+
+int
+pygame_scrap_put(char *type, int srclen, char *src){
+    if (!pygame_scrap_initialized()) {
+        PyErr_SetString(pgExc_SDLError, "scrap system not initialized.");
+        return 0;
+    }
+    if (strcmp(type, plaintext_type) == 0){
+        if (SDL_SetClipboardText(src) == 0) {
```
illume: srclen is not used here. Should srclen be used?

src_c/scrap_sdl2.c
illume: Returning `1` means the system has always lost the clipboard.

robertpfeiffer: HasClipboardText means there is text in the clipboard at all, not that it's the text we had put in there.

src_c/scrap_sdl2.c
robertpfeiffer: whoops, my bad.

src_c/scrap_sdl2.c
dlon: Also, static

src_c/scrap_sdl2.c
dlon: AFAICT, `src` is not necessarily null-terminated (a bytes object).

src_c/scrap_sdl2.c
dlon: Might `SDL_HasClipboardText() == 0` still work better than 1, even though it's wrong?

dlon:
Aside: There seems to be a memory leak where `pygame_scrap_get()` and `pygame_scrap_get_types()` are used.

src_c/scrap_sdl2.c
robertpfeiffer: Man, you guys got me to copy and null-terminate the strings manually, but then I looked into the Python docs, and bytes objects are null-terminated according to https://docs.python.org/3/c-api/bytes.html

robertpfeiffer: should have read the docs first...

illume: bytes 'may contain null bytes'. Which means it's not a null terminated string exactly, because the NULLs inside the string make it not.

illume:
I agree with this point:
> Might SDL_HasClipboardText() == 0 still work better than 1, even though it's wrong?

Seems the srclen/null terminated issue needs some sort of fix.

src_c/scrap_sdl2.c
robertpfeiffer: Ok. I see your point here BUT for the record utf-8-encoded strings shouldn't contain null bytes anywhere, so it's only a problem if we allowed other binary data. SDL2 doesn't allow null bytes in the clipboard.

illume: Ok. Let's leave it as is.

src_c/scrap_sdl2.c
illume: I guess since we can't implement pygame_scrap_lost with the sdl clipboard API, then always returning 1 is the more correct thing to do.

PR #1176, opened 2019-07-19
Draw circles with the old and aesthetically pleasing Bresenham algorithm

robertpfeiffer:
In pygame 1.9.4, circle drawing looked like this:

img

Drawing a circle used the correct bresenham algorithm (light blue circle). Drawing a filled circle used a different one based on ellipse drawing (light red circle, concentric grey filled circles). Drawing a thick circle just drew multiple concentric circles, as you can see when you compare the thick green circle with the concentric red and black circles.

Then in pygame 2.0-dev1 the moire pattern created by drawing concentric circles was "fixed":
img
Now everything uses the ugly ellipse drawing primitive, and circles are drawn twice, offset by one pixel horizontally, to avoid the moire pattern/holes created by the old version. One-pixel circles look especially ugly, and different from the old circles for no good reason.

In this branch, I returned to drawing circles based on the old algorithm. The drawing of thick or filled circles is consistent with one-pixel circles in pygame 1.9.4. Thin, thick and filled circles all have the same shape without any distortion or moire patterns.
img


illume:
Looks nice from your screenshot. Any chance of a test case to prevent a regression?

robertpfeiffer:
Not sure what the tests should look like? Just comparing drawn circles to a reference image maybe? Could I use the one I posted as a test case?

illume:
Something which fails on the old one, and passes with the new one.
Something based on properties that should hold true.

Two differences I can see for things which could be tested...
- the light blue circle... is that supposed to be 1 pixel width? In the second one, it seems (but I am unsure) that there are 2 pixel width parts.
- In the final picture there is a circle with white, black and red. How is that drawn? Could you paste that code here? Are the white spots supposed to be there? They disappeared in the second image.

robertpfeiffer:
The red and black circle is just concentric circles with width 1, so it approximates the 1.9.4 drawing of thick circles. The grey circles are concentric filled circles drawn over one another.

the that produced the images is
```python
import pygame
pygame.init()
flags=0
try:
    flags|=pygame.SCALED
except:
    pass
screen=pygame.display.set_mode((160,144), flags)
screen.fill((255,255,255))
pygame.display.flip()

pygame.draw.circle(screen, (100,100,100), (30,30), 29, 2)
pygame.draw.circle(screen, (100,200,200), (60,30), 29, 1)
pygame.draw.circle(screen, (200,100,100), (90,30), 29, 0)
pygame.draw.circle(screen, (100,200,100), (120,30), 29, 15)

for i in range(29, 5, -1):
    pygame.draw.circle(screen, [(i*30)%256]*3, (30,90), i, 0)

for i in range(29, 5, -1):
    pygame.draw.circle(screen, ((i%2)*255, 0, 0), (90,90), i, 1)

pygame.display.flip()
pygame.time.wait(5000)
pygame.image.save(screen, "test-draw-circles-"+pygame.version.ver+".png")
```

illume:
So, the white dots are an artifact caused by 1 pixel NOT being filled in (for each segment of the circle). I don't think this matters so much.

PR #1183, opened 2019-07-20
scale 2x optimization, now with added tests and documentation

robertpfeiffer:
this supersedes #647.

### DIFF
--- a/docs/reST/ref/transform.rst
+++ b/docs/reST/ref/transform.rst
@@ -95,6 +95,17 @@ Instead, always begin with the original image and scale to the desired size.)
 
    .. ## pygame.transform.scale2x ##
 
+.. function:: scale2xraw
+
+   | :sl:`nearest-neighbour image doubler`
+   | :sg:`scale2xraw(Surface, DestSurface = None) -> Surface`
+
+   This will return a new image that is double the size of the original. It
+   does nothing fancy, it just scales up the image. This is an optimised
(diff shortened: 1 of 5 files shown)


robertpfeiffer:
I don't know why the build failed. It should have worked, because that last commit only changed the docs. Is the CI broken for mac OS builds?

robertpfeiffer:
something's wring with the CI

illume:
A few thoughts and worries I have...

- [SDL_BlitScaled](https://wiki.libsdl.org/SDL_BlitScaled) is available in SDL2, and maybe we can use that. Then we don't need to maintain all this code, and optimizations can go into SDL instead. Maybe it's already quite well optimized.

- We don't really have that many tests for scaling, so this bunch of C code worries me. If you've been watching how many tests that charlesj has been writing for things, I guess you'll probably realise how many possible problems there could be. Does it work with subsurfaces, all the bit depths, all the surface types, ... etc.

- probably this could be put in as an optimization for transform.scale rather than a separate method. I guess it's faster (but there's not any benchmarks showing so), and it's probably better to let everyone benefit from this (if it is actually faster). In the other PR I mentioned it's maybe better as a separate flag to scale2x. But after thinking for some time, I realized that it doesn't really need a function or a flag, just some code to detect 2x scaling.

robertpfeiffer:
I also looked into BlitScaled, but for some reason I couldn't get it to work. We should at the very least benchmark this, I agree.

PR #1189, opened 2019-07-22
The problem is in the code

Juan65Lira:
I found that on line 122 it tries to display the text by calling the function "display", however the actual function is called "tprint" (line 16); please excuse my poor English and have a good day


illume:
Hello,

Oh right. Yes, in `docs/reST/ref/code_examples/joystick_calls.py`.

However, it seems your commit changed the wrong file. Do you want to try again?

Gayxican:
Honestly I didn't even know I could edit the code directly, besides I've
been coding for the past six hours and drinking about a liter of coffee and
I need to get some rest before I want to rip my eyeballs out in order to
not see another 'self'

Issue #976, opened 2019-04-10
ValueError raised when drawing an ellipse/circle/arc with a width value greater than its radius


charlesej:
A `ValueError` is raised when drawing an ellipse/circle/arc with a width value greater than its radius.

**Expected result:** The shape should be drawn with its width capped at its radius value.

**Actual result:** A `ValueError` is raised.

---

The following program shows that a `ValueError` is raised when drawing an ellipse with a width value greater than its radius value. Similar results occur when drawing an arc or a circle.

```python
# Filename: draw_max_width.py
import pygame

surface = pygame.Surface((10, 10))
rect = pygame.Rect((1, 1), (6, 6))
pygame.draw.ellipse(surface, pygame.Color('red'), rect, rect.h // 2 + 1)
```

Output from the above program.
```
pygame 2.0.0.dev0 (SDL 1.2.15)
Hello from the pygame community. https://www.pygame.org/contribute.html
Traceback (most recent call last):
  File "draw_max_width.py", line 6, in <module>
    pygame.draw.ellipse(surface, pygame.Color('red'), rect, rect.h // 2 + 1)
ValueError: width greater than ellipse radius
```

---

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit)
- pygame: 2.0.0.dev0 (SDL: 1.2.15) at 9d55b5fd7d589cbfaf1def4925d41b077397ecc0

Related to issues #968 and #975.

charlesej:
This issue can be done in 3 parts.

- [ ] Cap width values to radius for `pygame.draw.arc` (also add relevant tests and update documentation)
- [ ] Cap width values to radius for `pygame.draw.circle` (also add relevant tests and update documentation)
- [ ] Cap width values to radius for `pygame.draw.ellipse` (also add relevant tests and update documentation)

thiagojobson:
@charlesej, I'd like to contribute to this issue. Should I open a new PR or add commits to #1190?

charlesej:
It is best if you open a new PR to keep the changes separate.

PR #1190, opened 2019-07-22
Drawing ellipse/circle/arc with negative width value

### LINKED ISSUE #975: ValueError raised when drawing an ellipse/circle/arc with a negative width value
A `ValueError` is raised when drawing an ellipse/circle/arc with a negative width value.

**Expected result:** Same as when drawing a rect with a negative width value (draw nothing and return a bounding rect with the position set to that of the rect argument and its width and height set to zero).

**Actual result:** A `ValueError` is raised.

---

The following program shows that a negative width/thickness value for drawing a rect is accepted, but when drawing an ellipse a `ValueError` is raised. Similar results occur when drawing an arc or a circle.

```python
# Filename: draw_neg_width.py
import pygame

surface = pygame.Surface((10, 10))
rect = pygame.Rect((1, 1), (5, 5))
color = pygame.Color('red')
width = -1

bounds = pygame.draw.rect(surface, color, rect, width)
print('rect bounds =', bounds)
bounds = pygame.draw.ellipse(surface, color, rect, width)
print('ellipse bounds =', bounds)
```

Output from the above program. 
```
pygame 2.0.0.dev0 (SDL 1.2.15)
rect bounds = <rect(1, 1, 0, 0)>
  File "draw_neg_width.py", line 11, in <module>
    bounds = pygame.draw.ellipse(surface, color, rect, width)
ValueError: negative width
```

---

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit)
- pygame: 2.0.0.dev0 (SDL: 1.2.15) at 9d55b5fd7d589cbfaf1def4925d41b077397ecc0

thiagojobson:
Closes #975 : drawing an ellipse/arc/circle with negative `width` value would raise `ValueError`. 

Now the behavior is the same as when drawing a rect with a negative width value (draw nothing and return a bounding rect with the position set to that of the `rect` argument and its width and height set to zero). 

This PR has 3 parts.

- [x] Accept negative width values for pygame.draw.arc (also add relevant tests and update documentation)  
- [x] Accept negative width values for pygame.draw.circle (also add relevant tests and update documentation)
- [x] Accept negative width values for pygame.draw.ellipse (also add relevant tests and update documentation)

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -613,7 +613,7 @@ ellipse(PyObject *self, PyObject *arg, PyObject *kwargs)
     CHECK_LOAD_COLOR(colorobj)
 
     if (width < 0) {
-        return RAISE(PyExc_ValueError, "negative width");
+        return pgRect_New4(rect->x, rect->y, 0, 0);
     }
 
     if (width > rect->w / 2 || width > rect->h / 2) {
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -543,7 +543,7 @@ arc(PyObject *self, PyObject *arg, PyObject *kwargs)
     CHECK_LOAD_COLOR(colorobj)
 
     if (width < 0) {
-        return RAISE(PyExc_ValueError, "negative width");
+        return pgRect_New4(rect->x, rect->y, 0, 0);
     }
 
     if (width > rect->w / 2 || width > rect->h / 2) {
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -695,7 +695,7 @@ circle(PyObject *self, PyObject *args, PyObject *kwargs)
     }
 
     if (width < 0) {
-        return pgRect_New4(posx - radius, posy - radius, 0, 0);
+        return pgRect_New4(posx, posy, 0, 0);
     }
 
     if (width > radius) {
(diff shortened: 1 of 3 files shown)


src_c/draw.c
```diff
@@ -695,7 +695,7 @@ circle(PyObject *self, PyObject *args, PyObject *kwargs)
     }
 
     if (width < 0) {
-        return RAISE(PyExc_ValueError, "negative width");
+        return pgRect_New4(posx - radius, posy - radius, 0, 0);
```
charlesej: I think in this case the rect's x, y values should just be the circle's center (posx, posy). You will also need to update the test case for this.

From the circle's documentation.
> if nothing is drawn the bounding rect's position will be the ``center`` parameter value (float values will be truncated) and its width and height will be 0

thiagojobson: Sure, makes sense. I'll update this.

thiagojobson: Added a commit to address this change.

thiagojobson:
I think the build failed because github was off.

illume:
@thiagojobson the build failed on one node because of some unrelated reason.

thiagojobson:
I see, I'm glad it worked now. 
If everything is ok with this PR, I'd like to contribute to #976 too.

charlesej:
@thiagojobson I'll assign you to that issue.

PR #1202, opened 2019-07-29
Do not import numpy until it is used. Faster pygame import, less memory used.

illume:
pygame would import numpy if it was installed, even though most pygame apps do not use it.

```python
>>> 'numpy' in sys.modules
False
>>> import pygame
>>> 'numpy' in sys.modules
True
```
So now we do a lazy import. This adds 6ns to each function call in the surfarray and sndarray modules however. With python 3 this function call overhead should be able to be removed entirely with a smarter lazyload system.

Before this PR:
```bash
$ time python3 -c "import pygame"
real    0m0.172s
user    0m0.341s
sys    0m0.094s
```
After this PR:
```bash
$ time python3 -c "import pygame"
real    0m0.079s
user    0m0.055s
sys    0m0.018s
```

Thanks to @robertpfeiffer for finding this one.

### DIFF
--- a/src_py/sndarray.py
+++ b/src_py/sndarray.py
@@ -57,9 +57,12 @@ def array (sound):
     pygame.mixer.get_init().
     """
     global numpysnd
-    if numpysnd is None:
+    try:
+        return numpysnd.array (sound)
+    except AttributeError:
         import pygame._numpysndarray as numpysnd
-    return numpysnd.array (sound)
+        return numpysnd.array (sound)
+
 
 def samples (sound):
     """pygame.sndarray.samples(Sound): return array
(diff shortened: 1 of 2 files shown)


illume:
Not entirely happy with it, because it's ugly as hell and makes calling those functions slower.

illume:
The try/except way is faster. The code is even uglier.

Probably in python 3 we could use module level getattr with some more fancy module lazy loading.

robertpfeiffer:
This looks like a lot of code for a simple use case.

I would suggest something like this:
```python
if not "PYGAME_EXPLICIT_NUMPY" in os.environ or "numpy" in sys.modules:
    try:
       import pygame.surfarray
        surfarray = MissingModule("surfarray", urgent=0)

    try:
       import pygame.sndarray
        sndarray = MissingModule("sndarray", urgent=0)

```
surfarray has to be imported explicitly, and other modules that rely on numpy are not broken. It's a fringe use case anyway, improving quality of life for developers and making scripts run faster, not for games at runtime...

illume:
I think that would have been a good idea on how to do it originally. But now we have code relying on it, I'd prefer not to break it.

Seems module level gettatr didn't come until python 3.7 :( https://www.python.org/dev/peps/pep-0562/

PR #1213, opened 2019-07-30
Fix missing exception type (TypeError) when assigning rect.width to an invalid param

### LINKED ISSUE #1205: pygame.Rect.width: different error from all other properties for invalid value assignment
**System info:**
pygame 1.9.3 on python 3.5.2 on 32 bit Lubuntu (Ubuntu 16.04.6 LTS).
pygame installed via "sudo pip3 install pygame" (pip3 is pip 18.0, python 3.5)

**The issue:**
The example code below is obviously expected to raise an error, as it does, because we are assigning an invalid value to the "width" property. Nothing wrong here.

from pygame import Rect
r = Rect(10, 20, 30, 40)
r.width = "a"

What draws the attention, however, is the fact that "width" is the only pygame.Rect property that raises a "SystemError: error return without exception set", while all the other properties will raise a "TypeError: invalid rect assignment"

It looks like the wrong behaviour and that's why I'm reporting. Thanks for your time.

thiagojobson:
Fixes a missing `TypeError` exception when assigning an invalid width param to a rect.

```
>>> import pygame
pygame 2.0.0.dev3 (SDL 2.0.9, python 3.7.3)
>>> r = pygame.Rect(0, 0, 0, 0)
>>> r.width = "a"
TypeError: invalid rect assignment
```

Closes #1205

### DIFF
--- a/src_c/rect.c
+++ b/src_c/rect.c
@@ -1251,6 +1251,7 @@ pg_rect_setwidth(pgRectObject *self, PyObject *value, void *closure)
     int val1;
 
     if (!pg_IntFromObj(value, &val1)) {
+        RAISE(PyExc_TypeError, "invalid rect assignment");
         return -1;
     }
     self->r.w = val1;


illume:
Thanks :)

Issue #808, opened 2019-02-09
Most functions should accept keyword arguments


dlon:

Bottersnike:
Looking at it, it should just require changes from `METH_VARGS` to `METH_KEYWORDS` then have https://docs.python.org/2/c-api/arg.html#c.PyArg_ParseTupleAndKeywords be used to parse the arguments.

charlesej:
Changes for the draw module to accept keyword arguments is being done via #896.

illume:
See `rect-keyword-args` branch if anyone wants to finish off the rect kwargs.

MyreMylar:
Having a look at this, now I have pygame _mostly_ building on windows.
Modelling on the work @charlesej already did, I thought I'd start by having a go at doing one simple looking function in the event module, you can see my first commit here:

https://github.com/pygame/pygame/compare/master...MyreMylar:event-keyword-args

So the eventual syntax here would be:

`name = pygame.event.event_name(type=pygame.MOUSEBUTTONDOWN)`

it seems to work, though it is a bit of a trivial case. Was a bit unsure about specifying the 'type' of the 'type' parameter as int. I mean it _is_ an int but most people interact with the events system through the defined constants, but that is consistent with what was added by @charlesej .

If what I've done seems OK I can go through the rest of the event module. Not sure how/if we can handle the functions with two overlapping signatures like:

`set_blocked(type)`
`set_blocked(typelist)`

But there are definitely a couple of others that could have keyword arguments added.

MyreMylar:
I didn't really get very far with this one, I kept getting distracted by other things and was unable to think of neat solutions to map some c functions to python signatures. In the end my main interest was getting better auto-complete and linting support, which seems to have been largely solved by the .pyi files.

So, I'll let it go and see if anyone else has more interest in closing it off in the future.

cwaldro:
I have been working on this task a bit with surface.c and would love to jump in as well.

jonotassia:
I made a few updates #3804 for event.c and display.c

Event.c
• pg_event_init(pgEventObject *self, PyObject *args, PyObject *kwargs)

Display.c
• pg_gl_set_attribute(PyObject *self, PyObject *arg, PyObject *kwarg)
• pg_gl_get_attribute(PyObject *self, PyObject *arg, PyObject *kwarg)
• pg_set_palette(PyObject *self, PyObject *args, PyObject *kwargs)
[...]

PR #1219, opened 2019-08-01
Add keyword arguments to rect module functions

thiagojobson:
Hey, I started some preliminary work on adding support to kwargs on Rect methods, but I need some insights from you, guys.

For example. what should be the signatures for:

- move
- move_ip
- inflate
- inflate_ip

Those methods expect both `x`  and `y` as args. I think that if we are going to add kwargs to them, they should have a default value, e.g:

```
Rect.move(x=0, y=0)

# so this will be valid
rect.move(x=1)
 
# or even
rect.move() 
```

I added some code on that, please let me know what you guys think :)

Edit: builds are all failing, I added this code just to discuss.


charlesej:
Adding kwargs to the Rect methods will be very useful. I think, in most cases, defaulting to 0 is what a user would expect.

illume:
- Agreed about 0 being a good default.
- x & y as the argument names are good.

illume:
This is for issue: https://github.com/pygame/pygame/issues/808

illume:
Merging this into a `rect-kwargs` branch to finish it off.

Issue #1200, opened 2019-07-27
segfault with surface manipulations


hpnok:
There's probably some useless manipulations but I can consistently reproduce a segfault with those steps. worked on 1.9.
```
>>>import pygame
pygame 2.0.0.dev3 (SDL 2.0.9, python 3.7.0)
Hello from the pygame community. https://www.pygame.org/contribute.html
>>>pygame.init()
(5, 1)
>>>screen = pygame.display.set_mode((40, 20))
>>>image = pygame.image.load("sprite/block/trig2R3x8x8.png").convert()
>>>image.set_colorkey((0, 0, 0))
>>>sub_img = image.subsurface(pygame.Rect(0, 0, 8, 8))
>>>flip_img = pygame.transform.flip(sub_img, True, False)
>>>flip_img._pixels_address
193841920
>>>screen.blit(flip_img, (0, 0))
<rect(0, 0, 8, 8)>
>>>flip_img._pixels_address
0
>>>my_segfault_cant_be_this_cute = flip_img.copy()
Fatal Python error: (pygame parachute) Segmentation Fault
```

**Related Docs:** https://www.pygame.org/docs/ref/surface.html#pygame.Surface.copy

The image used (24bits): ![trig2R3x8x8](https://user-images.githubusercontent.com/8284306/61999733-8df88180-b095-11e9-8b7b-c6e23141c71e.png)

Might be related to #1181

nthykier:
Work around: Removing the `.convert()` call avoids the issue (not very satisfying but at least it provides a clue to where the problem appears).

nthykier:
Ok, `flip_img._pixel_address` turning to `0` is "simply" because SDL2 (as a part of blit'ing) changes to RLE.  When using the RLE format, the pixels felt is NULL'ed to avoid people fiddling with them without locking them (so that part is "expected behaviour").

As for the crash: I ran it through a debugger and it dies as a part of "undoing" the RLE format when doing the copy (i.e. `SDL_ConvertSurface`).

The bt from gdb is:

```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
0x00007ffff77e9a21 in uncopy_32 (dst=0x234f358, src=src@entry=0x1321c80, n=n@entry=154, sfmt=sfmt@entry=0x1321c60, dfmt=dfmt@entry=0x104e4f0)
    at ./src/video/SDL_RLEaccel.c:1013
1013    ./src/video/SDL_RLEaccel.c: No such file or directory.
[...]

PR #1224, opened 2019-08-02
WIP, quickfix for blit segfault

robertpfeiffer:
I don't know why this works. I don't know if there's a bug in SDL2.
It's not a mustock issue, or a refcount issue, as I initially suspected.
If you add a couple of printfs to surface.c

```python3
    else {
        /* Py_BEGIN_ALLOW_THREADS */
        printf("src->pixels BEFORE: 0x%x\n", src->pixels);
        printf("SDL_BlitSurface()\n");
        result = SDL_BlitSurface(src, srcrect, dst, dstrect);
        printf("src->pixels AFTER: 0x%x\n", src->pixels);
        /* Py_END_ALLOW_THREADS */
    }
```
the code from #1200 prints something like this:

```
src->pixels BEFORE: 0x175b3a0
SDL_BlitSurface()
src->pixels AFTER: 0x0
```
So I replaced SDL_BlitSurface with pygame_Blit in that line, and the bug was gone.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -4026,7 +4026,8 @@ pgSurface_Blit(PyObject *dstobj, PyObject *srcobj, SDL_Rect *dstrect,
 #endif /* IS_SDLv2 */
     else {
         /* Py_BEGIN_ALLOW_THREADS */
-        result = SDL_BlitSurface(src, srcrect, dst, dstrect);
+        pygame_Blit(src, srcrect, dst, dstrect, 0);
+        //result = SDL_BlitSurface(src, srcrect, dst, dstrect);
         /* Py_END_ALLOW_THREADS */
     }
 


charlesej:
This [post](https://discourse.libsdl.org/t/pixels-getting-set-to-null/9811/10) states that having the `surface->pixels=0` is done by SDL to `SDL_RLEACCEL` surfaces and that the pixel pointer is only valid when the surface is locked.

The problem doesn't actually seem to be with the blitting (although it is responsible for changing the pixel pointer value). If you use your original code with the printfs and blit a few times in a row, you will see that the pixel pointer in now always 0, but the blitting still works. If you lock the surface before printing, you will see the correct pixel pointer value.
e.g.
```c
        /* Py_BEGIN_ALLOW_THREADS */
        printf("src->pixels BEFORE: 0x%x\n", src->pixels);
        result = SDL_BlitSurface(src, srcrect, dst, dstrect);
        printf("src->pixels AFTER: 0x%x\n", src->pixels);
        SDL_LockSurface(src);
        printf("src->pixels AFTER (locked): 0x%x\n", src->pixels);
        SDL_UnlockSurface(src);
        /* Py_END_ALLOW_THREADS */
```

For issue #1200 the segfault is coming from `SDL_ConvertSurface()` in `surf_copy()`. It doesn't seem to handle some `SDL_RLEACCEL` surfaces properly once `SDL_BlitSurface` has set the pixel pointer to 0.

hpnok:
From some simple tests, the `RLEACCEL` flag is set only after blitting a surface which was returned from `transform.flip()`.
```
# 2.0.0.dev3
>>>flip_img = pygame.transform.flip(sub_img, True, False)
>>>hex(flip_img.get_flags())
'0x1000'  # SRCCOLORKEY 
>>>screen.blit(flip_img, (0, 0))
<rect(0, 0, 8, 8)>
>>>hex(flip_img.get_flags())
'0x5000'  # SRCCOLORKEY + RLEACCEL

# 1.9.6
>>>flip_img = pygame.transform.flip(sub_img, True, False)
>>>hex(flip_img.get_flags())
'0x1000'
>>>screen.blit(flip_img, (0, 0))
<rect(0, 0, 8, 8)>
>>>hex(flip_img.get_flags())
'0x1000'
```
Note that for surface `sub_img` (the original not flipped image), blitting doesn't add the `RLEACCEL` on either pygame version. Hope this help!

PR #1237, opened 2019-08-10
toggle_fullscreen with SDL2

### LINKED ISSUE #712: SDL2: fullscreen after starting in windowed mode doesn't work.

### LINKED ISSUE #1220: SDL2: display.toggle_fullscreen() not working
Maybe related to: https://github.com/pygame/pygame/issues/712

Including on windows 10.

robertpfeiffer:
This should close #1220. Maybe it will also close #712. 

It works on my machine, for the most part, but my window manager/desktop environment really doesn't like when a game switches the physical resolution of the monitor. 

With SCALED, everything runs really smooth though.


robertpfeiffer:
I took some functionality out of toggle_fullscreen again. This is tricky. Switching from windowed to fullscreen and back invalidates the display surface.

src_c/display.c
```diff
@@ -825,12 +825,17 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
         if (flags & PGS_FULLSCREEN){
             if (flags & PGS_SCALED){
                 sdl_flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
+            } else if (w == 0 && h == 0) {
+                sdl_flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
```
illume: I don't understand why this happens? Does this mean that having a 0 sized window is possible? And why is SDL_WINDOW_FULLSCREEN_DESKTOP used when it is zero?

src_c/display.c
```diff
@@ -867,6 +872,7 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
             int y = SDL_WINDOWPOS_UNDEFINED_DISPLAY(display);
             int w_1, h_1;
             int scale = 1;
+            SDL_DisplayMode dm;
```
illume: `display_mode` instead of `dm` would be easier to read.

src_c/display.c
```diff
@@ -879,18 +885,25 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
-            if (flags & PGS_SCALED && !(flags & PGS_FULLSCREEN)) {
-                SDL_DisplayMode dm;
-                int xscale, yscale;
-                if (SDL_GetDesktopDisplayMode(display, &dm) != 0) {
-                    return RAISE(pgExc_SDLError, SDL_GetError());
+            if (SDL_GetDesktopDisplayMode(display, &dm) != 0) {
+                return RAISE(pgExc_SDLError, SDL_GetError());
+            }
+
+            if (flags & PGS_SCALED){
```
illume: I guess this source needs to be run through clang format.

illume:
Very good.

I'll do some testing on some machines I have... 
How should I test this on my machines?

src_c/display.c
```diff
@@ -2001,13 +2014,59 @@ pg_toggle_fullscreen(PyObject *self, PyObject *args)
...
+            PyObject *surface = pg_GetDefaultWindowSurface();
+            result = SDL_SetWindowFullscreen(win, 0);
+            pgSurface_AsSurface(surface) = SDL_GetWindowSurface(win);
+        }
+    } else {
+        if (pg_renderer != NULL) {
+            int w, h;
+            SDL_Rect rect;
+            SDL_RenderGetLogicalSize(pg_renderer, &w, &h);
+            result = SDL_SetWindowFullscreen(win, SDL_WINDOW_FULLSCREEN_DESKTOP);
```
illume: Here it uses `SDL_WINDOW_FULLSCREEN_DESKTOP` in the branch where `SDL_GetWindowFlags(win) & SDL_WINDOW_FULLSCREEN_DESKTOP` is false. Is this intended? I don't understand it well enough to know.

src_c/display.c
```diff
@@ -2001,13 +2014,59 @@ pg_toggle_fullscreen(PyObject *self, PyObject *args)
 {
     SDL_Window *win = pg_GetDefaultWindow();
     int result;
+    int ww, wh;
```
illume: `window_w`, and `window_h` would be more readable.

src_c/display.c
```diff
@@ -879,18 +885,25 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
-                    return RAISE(pgExc_SDLError, SDL_GetError());
+            if (SDL_GetDesktopDisplayMode(display, &dm) != 0) {
+                return RAISE(pgExc_SDLError, SDL_GetError());
+            }
+
+            if (flags & PGS_SCALED){
+                if (!(flags & PGS_FULLSCREEN)) {
+                    int xscale, yscale;
+
+                    xscale = dm.w / w;
```
illume: Is it true that here `w` and `h` will never be zero?

src_c/display.c
robertpfeiffer: size (0, 0) used to mean "Set to maximum resolution supported by the monitor. The user must query the size of the returned screen surface". We can exploit that behaviour and use the resolution of the desktop and use SDL_WINDOW_FULLSCREEN_DESKTOP instead of touching the physical resolution of the monitor.

src_c/display.c
robertpfeiffer: Yes. This is the behaviour of SCALED | FULLSCREEN. And in this branch, we have an SDL_Renderer, but the window is **not yet** in FULLSCREEN, so we toggle and set to FULLSCREEN.

src_c/display.c
robertpfeiffer: Right now, they can't be zero. Should we add a check now for the sake of future-proofing?

illume:
Some notes our discussion on discord...

- make issue for toggle_fullscreen (with test results in there)
- return from toggle_fullscreen if not one of the sucessfully tested modes. (eg, always return on Mac, return for the failing windows cases).
- write docs about toggle_fullscreen being platform-dependent? (for now)
- merge in PR (maybe with a few more comments on the tricky bits for our future selves)
- make issue about display.set_mode((0,0)) selecting biggest resolution mode like SDL1.
- have a 'call toggle_fullscreen` type test case. So at least the function is run on all the different CI.
- How should people toggle fullscreen? We should probably have an example which shows how to do it. For the longest time, the safest way is to basically reinit your app?
- How do we query if we are in fullscreen? I've kept that state in the app before. But really, it should query that from the system. Because on some systems you can send it to fullscreen from the OS. (like on Mac).

robertpfeiffer:
As of now (not on master, in this PR),
- display.set_mode((0,0)) **does** select biggest resolution mode like SDL1.
- we can't query if we are in fullscreen from Python
- toggle_fullscreen returns early on unsupported platforms 
- toggle_fullscreen is a hack.
- we need to keep some state in pygame.display
- we need to query some state from SDL2
- SDL2 is behaving inconsistently and I can't test on all OSes.

I'll check with wayland next.
Does the CI have a GUI/windowing environment?
Can you get a GL context on the CI?

If this won't get merged within this week, I could pull out the changes related to 
setting the resolution to (0,0) to make a separate PR, and rebase this branch after that has been merged.

illume:
I think you may be able to get a GL context on appveyor and travis Mac. Not sure though.

This seems to be:
- better than what we have
- introduces no regressions

Is that right? If so, I'll do some more testing and then merge it in.

src_c/display.c
```diff
@@ -2000,14 +2033,140 @@ static PyObject *
...
+    display_surface = pg_GetDefaultWindowSurface();
+
+    // could also take the size of the old display surface
+    SDL_GetWindowSize(win, &window_w, &window_h);
+    window_display = SDL_GetWindowDisplayIndex(win);
+    if (SDL_GetDesktopDisplayMode(window_display, &display_mode) != 0) {
+        return RAISE(pgExc_SDLError, SDL_GetError());
+    }
+
+    /*
```
illume: I guess this commented code can be removed?

src_c/display.c
```diff
@@ -2000,14 +2033,140 @@ static PyObject *
...
+      case SDL_SYSWM_WAYLAND:
+      case SDL_SYSWM_MIR:
+#endif
+#if SDL_VERSION_ATLEAST(2, 0, 4)
+      case SDL_SYSWM_ANDROID:
+#endif
+#if SDL_VERSION_ATLEAST(2, 0, 5)
+      case SDL_SYSWM_VIVANTE:
+#endif
+    // everything that doesn't work will fallthrough to here
```
illume: 👍

markph0204:
Used this to check out how this is progressing:
https://gist.github.com/markph0204/b0e38e261ec1bb6fc5f010216452a016
Used this branch with (SDL 2.0.10, python 3.7.3) | Mac 10.14.6

---

- Full screen switching with "f" key works nicely.
- exit "ESC" from both modes works well
- switching to fullscreen using OS green fullscreen button not supported (disabled)

robertpfeiffer:
>     * switching to fullscreen using OS green fullscreen button not supported (disabled)
That would be related to the resizable flag and videosize events. These should work already.

I wrote these test cases, https://gist.github.com/robertpfeiffer/4e6e91a9287044f19b334ec34639518f https://gist.github.com/illume/f7c48b49b99ea880a1e3d6c65c766941

src_c/display.c
```diff
@@ -2021,14 +2066,140 @@ static PyObject *
...
+                             state->toggle_windowed_h);
+                state->toggle_windowed_w = 0;
+                state->toggle_windowed_h = 0;
+            }
+        } else if (flags == SDL_WINDOW_FULLSCREEN_DESKTOP) {
+            result = SDL_SetWindowFullscreen(win, 0);
+            display_surface->surf = SDL_GetWindowSurface(win);
+        } else {
+            SDL_DestroyWindow(win);
+            win = SDL_CreateWindow(state->title,
```
dlon: Could the flags from the existing window be used here?

Also, if I use `SDL_SetWindowFullscreen` here on Windows, switching back to windowed mode works fine

src_c/display.c
```diff
@@ -2021,14 +2066,140 @@ static PyObject *
...
+    */
+    if (state->using_gl) {
+        p_glViewport = (GL_glViewport_Func)SDL_GL_GetProcAddress("glViewport");
+        SDL_GL_GetDrawableSize(win, &w, &h);
+    } else {
+        w = display_surface->surf->w;
+        h = display_surface->surf->h;
+    }
+
+    if (flags & SDL_WINDOW_FULLSCREEN) {
```
dlon: This won't include the case where SDL_WINDOW_FULLSCREEN_DESKTOP is used (unless SDL_WINDOW_FULLSCREEN is always set at the same time?)

src_c/display.c
```diff
@@ -2021,14 +2066,140 @@ static PyObject *
...
+            SDL_GL_MakeCurrent(win, state-> gl_context);
+            if (state->toggle_windowed_w>0
+                && state->toggle_windowed_h>0) {
+                p_glViewport(0, 0,
+                             state->toggle_windowed_w,
+                             state->toggle_windowed_h);
+                state->toggle_windowed_w = 0;
+                state->toggle_windowed_h = 0;
+            }
+        } else if (flags == SDL_WINDOW_FULLSCREEN_DESKTOP) {
```
dlon: This should probably say `flags & SDL_WINDOW_FULLSCREEN_DESKTOP` (if SDL_WINDOW_FULLSCREEN *can* be set at the same time)

src_c/display.c
robertpfeiffer: SDL_WINDOW_FULLSCREEN_DESKTOP & SDL_WINDOW_FULLSCREEN == SDL_WINDOW_FULLSCREEN

src_c/display.c
robertpfeiffer: if flags == SDL_WINDOW_FULLSCREEN
then flags & SDL_WINDOW_FULLSCREEN_DESKTOP == SDL_WINDOW_FULLSCREEN

src_c/display.c
```diff
@@ -2021,14 +2066,140 @@ static PyObject *
...
+            result = SDL_SetWindowFullscreen(win, 0);
+            display_surface->surf = SDL_GetWindowSurface(win);
+        } else {
+            SDL_DestroyWindow(win);
```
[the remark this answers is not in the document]
robertpfeiffer: do you mean this is a Double Free Bug(tm) or is it just superfluous?

src_c/display.c
robertpfeiffer: Maybe that's a bug in SDL2? You see this code destroys the old window and creates a new one with no fancy flags set. No fullscreen, no hidden, borderless, just a normal window. This is already a workaround for weird behaviour of SDL2.0.8 with X11 on my machine, where the display surface returned by SDL_GetWindowSurface is wrong after calls to SDL_SetWindowFullscreen or SDL_SetWindowSize that change the physical resolution.

robertpfeiffer: And when I switch to fullscreen from there with SDL_SetWindowFullscreen, the surface returned by SDL_GetWindowSurface is a valid display surface again, and when I switch back to windowed again, it isn't, and so on...

src_c/display.c
robertpfeiffer: This should probably say `flags & SDL_WINDOW_FULLSCREEN_DESKTOP == SDL_WINDOW_FULLSCREEN_DESKTOP` in that case

dlon: Sorry. I see now that SDL_WINDOW_FULLSCREEN_DESKTOP includes the flag SDL_WINDOW_FULLSCREEN

src_c/display.c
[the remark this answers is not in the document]
dlon: Double Free Bug(tm)! SDL_DestroyWindow is called twice (`win` itself is SDL_free()'d in SDL_DestroyWindow). It should be fine to just delete this line

src_c/display.c
dlon: Probably. Could preprocessing be used to decide whether to use the workaround instead of `SDL_SetWindowFullscreen`? Did you find issues with window systems other than X11?

dlon:
Here are some other issues I found.

(1) An apparent issue with fullscreen where the resolution was changed (fullscreen_toggle.py). The actual surface would only appear in the upper-left corner instead of being scaled properly. But I think that that was just caused by 160x120 not being supported by my monitor.

(2) Fullscreen with SCALED looks strange (fullscreen_toggle.py):
image

Using this code instead, it looks right to me (though stretched):
```c
        if (pg_renderer != NULL) {
            result =
                SDL_SetWindowFullscreen(win, SDL_WINDOW_FULLSCREEN_DESKTOP);
            //SDL_RenderSetLogicalSize(pg_renderer, w, h);
        }
```

image

src_c/display.c
```diff
@@ -2021,14 +2066,140 @@ static PyObject *
...
+    if (SDL_GetDesktopDisplayMode(window_display, &display_mode) != 0) {
+        return RAISE(pgExc_SDLError, SDL_GetError());
+    }
+
+    /*
+      if (pg_renderer != NULL) {
+        SDL_RenderGetLogicalSize(pg_renderer, &w, &h);
+    }
+    */
+    if (state->using_gl) {
```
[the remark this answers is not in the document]
illume: From R. > renderer sometimes creates its own GL context

Let's look into this in a separate PR :)

src_c/display.c
illume: Let's clean this comment up in a separate PR.

src_c/display.c
```diff
@@ -2028,17 +2082,281 @@ static PyObject *
...
+      case SDL_SYSWM_DIRECTFB:
+      case SDL_SYSWM_UIKIT: // iOS currently not supported by pygame
+#if SDL_VERSION_ATLEAST(2, 0, 4)
+      case SDL_SYSWM_ANDROID: // currently not supported by pygame
+#endif
+          return PyInt_FromLong(-1);
+
+      // Untested and unsupported platforms
+#if SDL_VERSION_ATLEAST(2, 0, 2)
+    case SDL_SYSWM_MIR: //nobody uses mir any more, wayland has won
```
illume: The indentation of the `case` is wrong here. Run clang-format again?

src_c/display.c
```diff
@@ -2028,17 +2082,281 @@ static PyObject *
...
+#endif
+#if SDL_VERSION_ATLEAST(2, 0, 5)
+    case SDL_SYSWM_VIVANTE:
+#endif
+    case SDL_SYSWM_UNKNOWN:
+    default:
+	   return RAISE(pgExc_SDLError, "Unsupported platform");
```
illume: Raising an error here breaks backwards compatibility. Before you could call toggle_fullscreen and it would just do nothing. Instead, `return PyInt_FromLong(-1);` as above.

src_c/display.c
illume: @robertpfeiffer this one looks like it has been addressed?

illume: "Resolving" this comment, because SDL_DestroyWindow is not used anymore.

src_c/display.c
robertpfeiffer: I know how it's supposed to look. Will do when I rearrage he platforms.

src_c/display.c
robertpfeiffer: I put this here mostly for testing purposes. This will be changed once the code has been tested on the supported platforms.

robertpfeiffer:
This is obviously in a WIP state, and I'd like some feedback on 

1. whether it currently works on your machine
2. what you expect the behaviour to be
3. whether the API for resizeable SCALED windows should be kept

_before_ I

1. add any more user-facing documentation
2. clean up formatting and delete commented out code
3. remove the testing env-vars stuff (can be a separate branch if you like it)
4. make the code return early on any unsupported platforms
5. cherry pick stuff into other branches

illume:
I tested on Mac, and all good as far as I see. Thanks @robertpfeiffer.

PR #1225, opened 2019-08-02
Fix SDL_WINDOWEVENT event identified as 'Unknown'

charlesej:
Overview of changes:
- Fixed the `SDL_WINDOWEVENT` event name to be 'WindowEvent'
- Changed `test_event_name()` to test all the different event names

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit) and 2.7.10 (64bit)
- pygame: 2.0.0.dev3 (SDL: 2.0.10) at 7a5ed0b864645e7c325d80aa8866e837559c7a1c

Resolves #1222.

PR #1239, opened 2019-08-11
WIP: Test on old timey linux for old timey SDL2.

illume:
pygame should be able to work with older raspberian and older Debian.
We need a CI bot to test this for us, so one of the travis builders has been switched.

Related: https://github.com/pygame/pygame/pull/1225

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ matrix:
   include:
     - os: linux
       sudo: required
-      dist: xenial
+      dist: precise
       python: 2.7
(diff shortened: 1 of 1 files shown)


illume:
It seems precise doesn't have libsdl2 at all.

My plan is to make a Debian Jessie based docker build. Similar to the manylinux build.

robertpfeiffer:
I the steam runtime chroot environment at all relevant or useful here? https://github.com/ValveSoftware/steam-runtime

Issue #1230, opened 2019-08-05
SDL2: Incorrect name and value for pygame.NUMEVENTS


charlesej:
**Expected results:**
- The event name for `pygame.NUMEVENTS` should be 'Unknown'
- All event values should be greater than `pygame.NOEVENT` and less than `pygame.NUMEVENTS` (`pygame.NOEVENT` < event < `pygame.NUMEVENTS`)

**Actual results:**
- The event name for `pygame.NUMEVENTS` is 'ActiveEvent'
- Some event values are greater than `pygame.NUMEVENTS`

---

This code is used to print out some event names and values.

```python
import pygame

EVENT_DATA = (
    ('NoEvent', pygame.NOEVENT),
    ('ActiveEvent', pygame.ACTIVEEVENT),
    ('VideoResize', pygame.VIDEORESIZE),
    ('VideoExpose', pygame.VIDEOEXPOSE),
    ('Unknown', pygame.NUMEVENTS),
)

for expected_name, event in EVENT_DATA:
    name = pygame.event.event_name(event)

    if name == expected_name:
        print('{}=0x{:X}'.format(name, event))
    else:
        print('{}=0x{:X} (expected_name={})'.format(name, event, expected_name))
```

---

Output using pygame 1.9.4. The expected name for `pygame.NUMEVENTS` is 'Unknown' and the given event values are between `pygame.NOEVENT` and `pygame.NUMEVENTS`.
```
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
NoEvent=0x0
ActiveEvent=0x1
VideoResize=0x10
VideoExpose=0x11
Unknown=0x20
```

---

Output using pygame 2.0.0.dev3. The name for `pygame.NUMEVENTS` is 'ActiveEvent' and the given event values are not between `pygame.NOEVENT` and `pygame.NUMEVENTS`.
```
pygame 2.0.0.dev3 (SDL 2.0.10, python 3.7.2)
Hello from the pygame community. https://www.pygame.org/contribute.html
NoEvent=0x0
ActiveEvent=0xA000
VideoResize=0xA001
VideoExpose=0xA002
ActiveEvent=0xA000 (expected_name=Unknown)
```

---

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit)
- pygame: 1.9.4 and 2.0.0.dev3 (SDL: 2.0.10) at 7a5ed0b864645e7c325d80aa8866e837559c7a1c

[...]

PR #1246, opened 2019-08-14
Fix pygame.NUMEVENTS name problem #1230

flaambe:


src_c/_pygame.h
```diff
@@ -80,11 +80,11 @@ typedef enum {
...
     SDL_NOEVENT = 0,
     /* SDL 1.2 allowed for 8 user defined events. */
-    SDL_NUMEVENTS = SDL_USEREVENT + (32768/4),
-    SDL_ACTIVEEVENT = SDL_NUMEVENTS,
-    PGE_EVENTBEGIN = SDL_NUMEVENTS,
+    SDL_ACTIVEEVENT = SDL_USEREVENT + 0x2000,
+    PGE_EVENTBEGIN = SDL_ACTIVEEVENT,
     SDL_VIDEORESIZE,
     SDL_VIDEOEXPOSE,
+    SDL_NUMEVENTS,
```
illume: I'm not sure what SDL_NUMEVENTS is correct here.

It should be the maximum number an event can be assigned.
See: http://sdl.beuc.net/sdl.wiki/SDL_UserEvent
> type is a value from SDL_USEREVENT to SDL_NUMEVENTS-1 

SDL_ACTIVEEVENT is an event that isn't in SDL2, so we use one of the user events for it.

charlesej:
This should address the issues raised in #1230, but I think it might create a new issue of user event ids overlapping with pygame ids.

The `pg_event_custom_type()` function uses `SDL_NUMEVENTS` as its maximum user event id, so all ids in the range of `SDL_USEREVENT` to `SDL_NUMEVENTS - 1` are user event ids. This also needs to be true to maintain backward compatibility ([event docs](https://www.pygame.org/docs/ref/event.html)).

> The event type identifier is in between the values of NOEVENT and NUMEVENTS. User defined events should have a value in the inclusive range of USEREVENT to NUMEVENTS - 1. It is recommended all user events follow this system.

From src_c/event.c:
```c
int _custom_event = SDL_USEREVENT + 1;
static PyObject *
pg_event_custom_type(PyObject *self, PyObject *args)
{
    int result = _custom_event + 1;
    if (result > SDL_NUMEVENTS) {
        return RAISE(pgExc_SDLError, "pygame.event.custom_type made too many event types.");
    }
    _custom_event++;
    return PyInt_FromLong(result);
}
```

The `PygameEventCode` enum will now have `SDL_ACTIVEEVENT`, `SDL_VIDEORESIZE`, and `SDL_VIDEOEXPOSE` in the user event range of `USEREVENT` to `NUMEVENTS - 1`.

```c
typedef enum {
    SDL_NOEVENT = 0,
    /* SDL 1.2 allowed for 8 user defined events. */
    SDL_ACTIVEEVENT = SDL_USEREVENT + 0x2000,
    PGE_EVENTBEGIN = SDL_ACTIVEEVENT,
    SDL_VIDEORESIZE,
    SDL_VIDEOEXPOSE,
    SDL_NUMEVENTS,
    PGE_KEYREPEAT,
    PGE_EVENTEND
} PygameEventCode;
```

One idea would be to have the user events after all the other events to prevent any id trampling. Maybe try something like this where the pygame events are put at the start of the user events (instead of at the end). `PGE_USEREVENT` now becomes the start of the user events that are made available to the users. `PGE_USEREVENT` also now needs to be the pygame.locals constant value for `USEREVENT`.

```c
typedef enum {
    /* Any SDL_* events here are for backward compatibility. */
    SDL_NOEVENT = 0,

    /* pygame events */
    PGE_EVENTBEGIN = SDL_USEREVENT, /* Not an event. Indicates start of pygame events. */
    SDL_ACTIVEEVENT = PGE_EVENTBEGIN,
    SDL_VIDEORESIZE,
    SDL_VIDEOEXPOSE,
    PGE_KEYREPEAT,
    PGE_EVENTEND, /* Not an event. Indicates end of pygame events. */

    /* User event range. */
    /* SDL 1.2 allowed for 8 user defined events. */
    PGE_USEREVENT = PGE_EVENTEND,
    SDL_NUMEVENTS = SDL_USEREVENT + 0x2000, /* Not an event. Indicates end of user events. */
} PygameEventCode;
```

flaambe:
Did i understand correct, that SDL_NUMENTS == SDL_LASTEVENT?

charlesej:
Currently the full range of user events is not being used. So `SDL_NUMEVENTS < SDL_LASTEVENT`.

The current pygame 2.0.0.dev3 values are:
```c
SDL_USEREVENT == 0x8000
SDL_NUMEVENTS == 0xa000
SDL_LASTEVENT == 0xffff
```

dlon:
> try something like this where the pygame events are put at the start of the user events (instead of at the end). `PGE_USEREVENT` now becomes the start of the user events that are made available to the users. `PGE_USEREVENT` also now needs to be the pygame.locals constant value for `USEREVENT`.

+1. Only some uses of SDL_USEREVENT within the source should and must be replaced.

flaambe:
Should I change SDL_USEREVENT to PGE_USEREVENT in the file [sdlmain_osx.m](https://github.com/pygame/pygame/blob/master/src_c/sdlmain_osx.m)? Thanks for idea @charlesej

dlon:
Use PGE_USEREVENT wherever it refers to the first pygame user event visible
to pygame users.
Use SDL_USEREVENT wherever it refers to the first of all SDL user events,
including the ones used internally by pygame.


Den fre 16 aug. 2019 09:38flaambe <notifications@github.com> skrev:

src_c/event.c
```diff
@@ -1863,7 +1864,7 @@ MODINIT_DEFINE(event)
             MODINIT_ERROR;
         }
 
-        if (SDL_RegisterEvents(PGE_NUMEVENTS) != PGE_EVENTBEGIN) {
+        if (SDL_RegisterEvents(PGE_NUMEVENTS) != SDL_NUMEVENTS) {
```
dlon: This second [`SDL_RegisterEvents`](https://wiki.libsdl.org/SDL_RegisterEvents) call should just be removed now, or else it must be called before the above `SDL_RegisterEvents` invocation (because its range now comes first). Also, `PGE_EVENTBEGIN` should still work here (returns first event).

But the easiest fix is to just remove this if statement altogether.

test/event_test.py
```diff
@@ -48,7 +48,7 @@
 
     ('UserEvent', pygame.USEREVENT),
     ('UserEvent', pygame.USEREVENT + 1),
-    ('UserEvent', pygame.ACTIVEEVENT - 1),
+    ('UserEvent', pygame.USEREVENT - 1),
```
dlon: This is probably outside the allowed range? Could be reverted with the change mentioned above.

src_c/_pygame.h
```diff
@@ -78,15 +78,21 @@ typedef enum {
...
     SDL_VIDEORESIZE,
     SDL_VIDEOEXPOSE,
     PGE_KEYREPEAT,
-    PGE_EVENTEND
+    PGE_EVENTEND, /* Not an event. Indicates end of pygame events. */
+
+    /* User event range. */
+    /* SDL 1.2 allowed for 8 user defined events. */
+    PGE_USEREVENT = PGE_EVENTEND,
+    SDL_NUMEVENTS = PGE_USEREVENT + 0x2000 /* Not an event. Indicates end of user events. */
```
dlon: It's probably best to set `SDL_ACTIVEEVENT = SDL_NUMEVENTS` here instead, rather than use the value above, so that it still equals `(the last user event + 1)`.

src_c/event.c
flaambe: Should I change 2xSDL_USERVENT to PGE_USEREVENT in this IS_SDLv2 if statements?

src_c/_pygame.h
flaambe: SDL_NUMEVENTS maximum number an event can be assigned, why SDL_ACTIVEEVENT = SDL_NUMEVENTS?

dlon: You're right. I was just confused by this line from before:

```python3
('UserEvent', pygame.ACTIVEEVENT - 1),
```

It seems wrong.

src_c/event.c
dlon: No. We're reserving the values in the interval [SDL_USEREVENT, PGE_USEREVENT) as well as in [PGE_USEREVENT, SDL_NUMEVENTS), so this should be enough:

```c
#if IS_SDLv2
    if (!have_registered_events) {
        int numevents = SDL_NUMEVENTS - SDL_USEREVENT;
        Uint32 user_event = SDL_RegisterEvents(numevents);

        if (user_event == (Uint32)-1) {
            PyErr_SetString(pgExc_SDLError, "unable to register user events");
            DECREF_MOD(module);
            MODINIT_ERROR;
        }
        if (user_event != SDL_USEREVENT) {
            PyErr_SetString(PyExc_ImportError,
                            "Unable to create another module instance");
            DECREF_MOD(module);
            MODINIT_ERROR;
        }

        have_registered_events = 1;
    }

    SDL_SetEventFilter(pg_event_filter, NULL);
#endif /* IS_SDLv2 */
```

illume:
🎉 thanks

PR #1255, opened 2019-08-17
set_mode: Wrong 'hidden' flag was used

dlon:


illume:
https://wiki.libsdl.org/SDL_WindowFlags

PR #1257, opened 2019-08-17
small typo fix in freetype docs

s0lst1ce:
replaced `an Python` with `a Python`. Unless I'm mistaken this is so in english because `a` precedes a word starting with a consonant and `an` one with a vowel.


dlon:
Thanks

PR #1265, opened 2019-08-26
add editorconfig, make docscomments greppable

robertpfeiffer:
I don't want to bikeshed too much.
I made this editorconfig so emacs and vim automatically uses spaces instead of tabs.
Pretty-printing docscomments is just a little quality-of-life thing, but `git grep` was totally unusable without the line breaks. I always got the whole contents of the whole file.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+root=true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# The settings for C (*.c and *.h) files are mirrored in src_c/.clang-format.
+# Keep them in sync.
+[*.{c,h}]
+indent_style = space
+indent_size = 4
+tab_width = 4
+trim_trailing_whitespace = true
+max_line_length = 79
+
+[*.{py,pyx}]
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,20 @@
...
+indent_size = 4
+tab_width = 4
+trim_trailing_whitespace = true
+max_line_length = 79
+
+[*.{py,pyx}]
+indent_style = space
+indent_size = 4
+charset = utf-8
+file_type_emacs = python
```
charlesej: It would be good to also have these lines in the `[*.{py,pyx}]` section.
```
trim_trailing_whitespace = true
max_line_length = 79
```

illume:
Cool.

I'm wondering if we can avoid having this file in the root of the repo, and having one in each of the sub folders? Would that work?

robertpfeiffer:
> I'm wondering if we can avoid having this file in the root of the repo, and having one in each of the sub folders? Would that work?

I don't see why not, but I also don't understand why.

PR #1287, opened 2019-09-06
WIP: linux-ppc64le and linux-ppc64be builds

illume:
I don't know where these are documented... but I saw haproxy uses linux-ppc64le. So... I'll give it a go.


illume:
linux-ppc64be boots as an x86.
linux-ppc64le gets stuck on 'booting virtual machine'.

PR #1288, opened 2019-09-06
Adding PowerPC linux-ppc64le, and python 3.8.dev travis builds.

illume:


illume:
I set them both to 'allowed to fail' for now because they both have failing things.

- ppc has a few failing tests: https://travis-ci.org/pygame/pygame/jobs/581695143
- python 3.8 fails with the cython build again: https://travis-ci.org/pygame/pygame/jobs/581695141

PR #1296, opened 2019-09-07
Ensure window is initially black

### LINKED ISSUE #1254: SDL2: Display defaults to white background
The display defaults to a white background. The default used to be black.

```pygame
import pygame
import time

pygame.display.init()
screen = pygame.display.set_mode((200, 200))
time.sleep(10)
pygame.quit()
```

---

Default display for pygame 2.0.0.dev3.
pygame_display_white

---

Default display for pygame 1.9.4.
pygame_display_black

---

System details:
- os: windows 10 (64bit)
- pygame: 2.0.0.dev3 (SDL: 2.0.10) at 5dac61141d49a74004e385ad32a647814ca8682f

dlon:
Fix #1254. The window surface actually is black initially, but `SDL_UpdateWindowSurface()` has to be called in order for it to show, as this example shows:

```python3
import pygame
import time

pygame.display.init()
screen = pygame.display.set_mode((200, 200))
pygame.display.flip() # white if this is removed
time.sleep(10)
pygame.quit()
```

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1252,12 +1252,12 @@ pg_flip_internal(_DisplayState *state)
             SDL_RenderPresent(pg_renderer);
         }
         else {
-            status = SDL_UpdateWindowSurface(win) == -1;
+            status = SDL_UpdateWindowSurface(win);
         }
     }
     Py_END_ALLOW_THREADS;
 
-    if (status == -1) {
+    if (status < 0) {
         RAISE(pgExc_SDLError, SDL_GetError());
         return -1;
     }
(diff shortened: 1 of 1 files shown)


src_c/display.c
```diff
@@ -1248,8 +1257,20 @@ pg_flip(PyObject *self, PyObject *args)
     }
     Py_END_ALLOW_THREADS;
 
-    if (status == -1)
-        return RAISE(pgExc_SDLError, SDL_GetError());
+    if (status == -1) {
```
charlesej: This isn't directly related to the changes you made, but I noticed this in the surrounding code while reviewing your update.

The `status` variable is initialized to 0 and is only ever set on line 1255:
`status = SDL_UpdateWindowSurface(win) == -1;`

Based on this, `status` can only be 0 or 1 and the check on line 1260 never evaluates to true. Also, any check of `SDL_UpdateWindowSurface()`'s return code should not just be checking `== -1`, it should be `<= -1` as the [SDL_UpdateWindowSurface's documentation](https://wiki.libsdl.org/SDL_UpdateWindowSurface) states:
> Returns 0 on success or a negative error code on failure

dlon: Good catch.

illume:
Before on Mac I get...
```python
>>> import pygame
>>> pygame.init()
>>> screen = pygame.display.set_mode((200, 200))
>>> screen.get_at((0,0))
(0, 0, 99, 0)
```

On this branch I get...
``` python
>>> screen.get_at((0,0))
(0, 0, 0, 0)
```

PR #1300, opened 2019-09-09
handle zero dimensions, infrastructure for toggle_fullscreen

robertpfeiffer:
this is basically #1237 without the actual toggling.


illume:
Merging in the #1237 PR instead.

Issue #1133, opened 2019-07-06
code quality: LGTM alerts


illume:
See here:
https://lgtm.com/projects/g/pygame/pygame/alerts/?mode=list

It finds various C and python issues.

charlesej:
LGTM can also be setup to automatically review code in pull requests:

- https://lgtm.com/projects/g/pygame/pygame/ci/
- https://lgtm.com/help/lgtm/about-automated-code-review

illume:
I've enabled the LGTM app on github... so it should be enabled on PRs now.

illume:
LGTM started throwing errors in the PRs. So I had to disable it. The logs didn't show anything. Probably needs a support request to be logged.

charlesej:
Found this in one of the [logs](https://lgtm.com/projects/g/pygame/pygame/logs/rev/pr-5db83ee99287bda27473158467112461113b4fb2/lang:python/stage:Build%20merge_32497f2cac2695ba9942c484f5e4851afad204bc) (under the Extraction tab).

```
[2019-10-08 21:25:58] [build] Configuration file:
[2019-10-08 21:25:59] [build] prepare:
[2019-10-08 21:25:59] [build] python_setup:
[2019-10-08 21:26:00] [build] Source root appears to be the real root.
[2019-10-08 21:26:00] [build] /opt/dist/tools/pybuild.zip/version.py:60: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
[2019-10-08 21:26:04] [build] New python executable in /opt/out/snapshot/workspace/venv/bin/python2
[2019-10-08 21:26:04] [build] Also creating executable in /opt/out/snapshot/workspace/venv/bin/python
[2019-10-08 21:26:04] [build] Installing setuptools, pip, wheel...
[2019-10-08 21:26:06] [build] done.
[2019-10-08 21:26:06] [build] WARNING, No "Setup" File Exists, Running "buildconfig/config.py"
[2019-10-08 21:26:06] [build] Using UNIX configuration...
[2019-10-08 21:26:06] [build] sh: 1: sdl2-config: not found
[2019-10-08 21:26:06] [build] sh: 1: sdl2-config: not found
[2019-10-08 21:26:06] [build] sh: 1: sdl2-config: not found
[2019-10-08 21:26:06] [build] Hunting dependencies...
[...]

PR #1352, opened 2019-09-29
Fix the LGTM alerts in setup_win_common.py

charlesej:
Overview of changes:
- Fixed the LGTM alerts in setup_win_common.py
  (LGTM info for pygame: https://lgtm.com/projects/g/pygame/pygame)

System details:
- os: windows 10 (64bit)
- python: 3.7.4 (64bit) and 2.7.10 (64bit)
- pygame: 2.0.0.dev3 (SDL: 2.0.10) at 1556f34474a6f70f2165968eebf9fd95e9737aa3

Resolves the setup_win_common.py alerts from the LGTM link in #1133.

### DIFF
--- a/buildconfig/setup_win_common.py
+++ b/buildconfig/setup_win_common.py
@@ -15,14 +15,12 @@ class Definition(object):
         self.name = name
         self.value = value
 
+
 def read():
     """Return the contents of the Windows Common Setup as a string"""
-
-    setup_in = open(PATH)
-    try:
+    with open(PATH) as setup_in:
         return setup_in.read()
-    finally:
-        setup_in.close()
+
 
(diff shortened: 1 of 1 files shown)


illume:
Thanks

charlesej:
Thanks for reviewing the changes.

This was addressing 2 LGTM alerts of type: "Instance of context-manager class file is closed in a finally block. Consider using 'with' statement."

In the `get_definitions()` function I thought moving that code (it doesn't require the file to be open) outside the `with` context manager was okay, since the `finally` block was only being used to close the file.

Probably should have included this information in the initial description of the changes.

illume:
Thanks for the explanation. Makes sense to me :)

Issue #1316, opened 2019-09-18
Compiler warnings should fail on CI.


illume:
This way we can try and reduce new warnings being introduced and we miss them in the PR review.

- `-Werror` can be used to turn warnings into errors
- `-Wno-error=unknown-pragmas`
- should run only in CI, and not in the build by default. See [werror-is-not-your-friend](https://embeddedartistry.com/blog/2017/5/3/-werror-is-not-your-friend).
- Use most recent gcc, and most recent clang (change two travis linux builders to do this).
- first have to fix all the warnings (and hopefully not break something in the process).
- consider adding extra warnings to gcc: `"-Wall -Wextra -Wpedantic"` See [C tooling](http://renesd.blogspot.com/2019/09/post-modern-c-tooling.html
)
- Also the Microsoft compiler MSVC on Appveyor can be configured to treat warnings as errors. The `/WX` argument option treats all warnings as errors. See [MSVC warning levels](https://docs.microsoft.com/en-us/cpp/build/reference/compiler-option-warning-level?view=vs-2019)
- For MSVC on Appveyor, `/wdnnnn` Suppresses the compiler warning that is specified by nnnn. For example, /wd4326 suppresses compiler warning C4326.

To build with warnings as errors(on gcc/clang) use the following:
```
CFLAGS="-Werror" python setup.py build
```

illume:
The warnings are down to these counts (including the pragma warn warnings):

- 40, MSVC appveyor 32bit python 3.7.
- 78, gcc travis python3.7 build, gcc "(Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609"
- 68, clang travis mac python3 build.

illume:
The warning pragmas should probably be replaced with github issues.
Maybe a link to the issue could be left in the source.

Issue #1031, opened 2019-05-08
SDL2: Mac OS X special character prompt on pygame2


galexandreg:
The special character pop-up that appears when you hold a key on Mac OS X is visible with Pygame2, even if there is no cursor and the program isn't used to enter text anywhere.

I guess this is simply because SDL2 is using more modern features of Mac OS. It can be disabled in the system preferences but it's not really practical and it would be useful to be able do disable it in Pygame when it isn't needed.

Tested on MacOS 10.14.4 with Pygame 2.0.0.dev1 (SDL 2.0.9) and Python 3.7.3 from the official installer.

For example, this is a screenshot of the pop-up menu that appears when holding "a":
![Capture d’écran, le 2019-05-07 à 19 42 28](https://user-images.githubusercontent.com/45190261/57340003-f886ca80-7101-11e9-87ea-296d58c2d463.png)

The apple support page describing the feature:
https://support.apple.com/en-us/HT201586

Edit:
I have no experience with the SDL itself and not much more with C, but this discussion seems to have some useful informations: https://discourse.libsdl.org/t/sdl-2-0-4-text-input-mode-on-osx/21731

This post by VectorStorm:
> After a bunch of testing with the kind help of Alex Szpakowski, we’ve pretty well established that SDL2 applications normally will never invoke the accent menu on OS X, regardless of build settings, library type (framework vs. shared libs vs. static), deployment targets, or varying build toolchains.
> 
> The problem turned out to have been that as I migrated my project from SDL1.2 to SDL2, I had left SDLMain.m in place for the OS X builds (the migration guide (https://wiki.libsdl.org/MigrationGuide#Moving_from_SDL_1.2_to_2.0) refers to SDL_main as “optional”). But in practice, the operations performed in that SDLMain.m source file from SDL1.2 accidentally enables the accent menu, under SDL2.

atopuzov:
I ran into the same issue. I managed to get around it by removing the call to sdlmain
```sdlmain_osx.InstallNSApplication(default_icon_data)```
https://github.com/pygame/pygame/blob/19ccbd4bf9bf2ad39ee4dd41f7e3c3c767f2573b/src_py/macosx.py#L19
Which basically does not call code defined here:
https://github.com/pygame/pygame/blob/19ccbd4bf9bf2ad39ee4dd41f7e3c3c767f2573b/src_c/sdlmain_osx.m#L234
[...]

PR #1367, opened 2019-10-03
Mac sdlmain cleanups.

illume:
- Removed the app delegate because of hold down key bringing up the selectors for unicode keys. See https://github.com/pygame/pygame/issues/1031
- Removed unneeded code for DROPFILE (in sdl2 now).
- Fixed deprecated compiler warnings, and stopped using deprecated functions: https://developer.apple.com/documentation/appkit/nseventmodifierflags?language=objc
- Cleaned up returning None (before it was potentially making None disappear crashing the interpreter, because it was reducing the reference count).

Compiler warnings PR: https://github.com/pygame/pygame/issues/1316

### DIFF
--- a/src_c/sdlmain_osx.m
+++ b/src_c/sdlmain_osx.m
@@ -150,24 +150,6 @@ _RunningFromBundleWithNSApplication(PyObject* self)
 }
 @end
 
-@interface PYGPYGSDLApplicationDelegate : NSObject
-@end
-@implementation PYGPYGSDLApplicationDelegate
-- (BOOL)application:(NSApplication *)theApplication openFile:(NSString *)filename
-{
-    int posted;
-
-    /* Post the event, if desired */
-    posted = 0;
-    SDL_Event event;
-    event.type = SDL_USEREVENT;
-    event.user.code = 0x1000;
-    event.user.data1 = SDL_strdup([filename UTF8String]);
-    posted = (SDL_PushEvent(&event) > 0);
-    return (BOOL)(posted);
-}
-@end
-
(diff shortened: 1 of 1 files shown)


src_c/sdlmain_osx.m
```diff
@@ -316,51 +307,56 @@ static void setupWindowMenu(void)
...
 _ScrapSetMode(PyObject *self, PyObject *args) {
 #if defined (PYGAME_MAC_SCRAP_OLD)
-	return Py_None;
+    Py_RETURN_NONE;
 #else
 	char *mode;
-	if (!PyArg_ParseTuple (args, "s", &mode))
-		return Py_None;
-	return Py_None;
+	if (!PyArg_ParseTuple (args, "s", &mode)) {
```
charlesej: There's a Coverity issue for this bit of code if you are interested in addressing it as well.

> CID 1486568 (# 1 of 1): Identical code for different branches (IDENTICAL_BRANCHES)
> identical_branches: The same code is executed when the condition !PyArg_ParseTuple(args, "s", &mode) is true or false, because the code in the if-then branch and after the if statement is identical. Should the if statement be removed?

illume: Ah. I guess these failing PyArg_ParseTuple ones should be `return NULL`.

charlesej: Yes, that is what is done in the other files.

illume: I put `return NULL` for all the PyArg_ParseTuple failed cases.

PR #1371, opened 2019-10-03
WIP: Try and get unicode field working for keydown.

illume:
For https://github.com/pygame/pygame/issues/621

test case:

See if the keydown event has the appropriate text in the `unicode` field(Also try modifiers like shift). If a key on the keyboard is pressed then there should be an `a` in the `unicode` field of the keydown event.
```
python3 -m pygame.examples.eventlist
```

TODO: does `unicode` field work on?
- [x]  Linux X11
- [x] Windows

(Also removes a compiler warning: https://github.com/pygame/pygame/issues/1316)

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -100,6 +100,8 @@ pgEvent_AutoInit(PyObject *self, PyObject *args)
 }
 
 static char _pg_last_unicode_char[32] = { 0 };
+static SDL_Event *_pg_last_keydown_event = NULL;
+
 
 /*SDL 2 to SDL 1.2 event mapping and SDL 1.2 key repeat emulation*/
 static int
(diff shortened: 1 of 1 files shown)


dlon:
Is it safe to assume that SDL_TEXTINPUT will be filtered before SDL_KEYDOWN is processed?

illume:
Have you seen this happen? On my Mac I bashed at my keyboard, and couldn't get this happening.
I haven't tried on Windows, Ubuntu, etc yet.

illume:
Ah, I think I understand now. I guess you can get KEYBOARD events separately from TEXTINPUT events from the queue too. This would mess things up as well.

MyreMylar:
I'm seeing sporadic blank unicode fields on KEYDOWN events in 2.0.0.dev4 on Windows. Sample output of me pounding the 'h' key a handful of times:

`Key code h unicode type: <class 'str'> contents: h
Key code h unicode type: <class 'str'> contents: 
Key code h unicode type: <class 'str'> contents: h
Key code h unicode type: <class 'str'> contents: 
Key code h unicode type: <class 'str'> contents: 
Key code h unicode type: <class 'str'> contents: 
Key code h unicode type: <class 'str'> contents: 
Key code h unicode type: <class 'str'> contents: h`

I didn't get this before in Windows on dev3.

Issue #1289, opened 2019-09-07
SDL 2: Per Pixel Alpha error.


Pololot64:
- [x] minimal working example to reproduce the issue
- [x] write unit test(s) (see [test_src_alpha_issue_1289](https://github.com/pygame/pygame/blob/master/test/surface_test.py#L980)) edit the file to unskip test, then `python3 tests/surface_test.py -k test_src_alpha_issue_1289`
- [x] add usable BLEND_PREMULTIPLIED special_flag to blit & document.
- [ ] find conditions for the (src_surf+dest_surf+blit opts) to use the new blit. (see [test_src_alpha_issue_1289](https://github.com/pygame/pygame/blob/master/test/surface_test.py#L980) to help debug this) [inside pgSurface_Blit in surface.c](https://github.com/pygame/pygame/blob/master/src_c/surface.c#L4198) (started... see below in this issue description)
- [x] implement a more comprehensive test case with more example values and edge cases.
- [ ] implement SDL1 compatible blitter for the specific case (similar to new BLEND_PREMULTIPLIED blitter) 'BLEND_SDL1_ALPHA'. This should be the default for backwards compatibility. Only with SSE2+ARM neon, not mmx (because 99.99% of x86 systems have SSE2 these days).
- [ ] the SDL2 blitter should be available as 'BLEND_SDL2_ALPHA'
- [ ] a function/option which makes 'BLEND_SDL2_ALPHA' the default.

Blend mode formula SDL2: https://hg.libsdl.org/SDL/file/3b03741c0095/include/SDL_blendmode.h#l37
```
         SDL_BLENDMODE_BLEND = 0x00000001,    /**< alpha blending
                                                   dstRGB = (srcRGB * srcA) + (dstRGB * (1-srcA))
                                                   dstA = srcA + (dstA * (1-srcA)) */
```

The BLEND_PREMULTIPLIED PRs for reference (the new blitter should be mostly the same as this)
- [Document BLEND_PREMULTIPLIED Surface.blit flag](https://github.com/pygame/pygame/pull/1706)
- [Adds MMX, SSE2 & Arm NEON optimised versions of blit_blend_premultiplied()](https://github.com/pygame/pygame/pull/1711)
- [Some tweaks to the pre-multiply alpha blending](https://github.com/pygame/pygame/pull/1834)

### find conditions for the (src_surf+dest_surf+blit opts) to use the new blit.

[...]

PR #1372, opened 2019-10-04
test case for SRC_ALPHA blit issue 1289

illume:
For https://github.com/pygame/pygame/issues/1289#issuecomment-538167187

- [x] test case


illume:
Note, there's a test case in this comment on the solarwolf issue: https://github.com/pygame/pygame/issues/721#issuecomment-459548722

Issue #1384, opened 2019-10-07
dev4: circles drawn a pixel wider or rects smaller?


mcpalmer1980:
Decided to try my games with dev4. My round_rect function worked in 1.9.1, 1.9.6, and dev3. Now it looks a little funny. At first I thought that the circles draw larger because the circles at each corner of the rounded rectange overlapped the rectangle by one pixel. I think what is actually happening is that the rect is drawing 1 pixel smaller. The circle only overlaps on the right side and bottom.

System is Linux Mint 18.3
pygame 2.0.0.dev4 (SDL 2.0.10, python 3.5.2)
pygame 2.0.0.dev4 (SDL 2.0.10, python 3.7.12)

![overlap](https://user-images.githubusercontent.com/47106920/66357338-58490200-e93c-11e9-9503-8a73d8fc4555.png)

```
def round_rect(surf, rect, rad, color, thick=0):
	trans = (255,255,1)
	if not rad:
		pygame.draw.rect(surf, color, rect, thick)
		return
	elif rad > rect.width / 2 or rad > rect.height / 2:
		rad = min(rect.width/2, rect.height/2)

	if thick > 0:
		r = rect.copy()
		x, r.x = r.x, 0
		y, r.y = r.y, 0
		buf = pygame.surface.Surface((rect.width, rect.height)).convert()
		buf.set_colorkey(trans)
		buf.fill(trans)
		round_rect(buf, r, rad, color, 0)
		r = r.inflate(-thick*2, -thick*2)
		round_rect(buf, r, rad, trans, 0)
		surf.blit(buf, (x,y))


	else:
		r  = rect.inflate(-rad * 2, -rad * 2)
		for corn in (r.topleft, r.topright, r.bottomleft, r.bottomright):
			pygame.draw.circle(surf, color, corn, int(rad))
		pygame.draw.rect(surf, color, r.inflate(rad*2, 0))
		pygame.draw.rect(surf, color, r.inflate(0, rad*2))
```

MightyJosip:
I am not sure if this is connected to the rect, but it could have something with the change when drawing circle. Before if you set radius to be 3 than the diameter would be 6, if you set radius to be 4 than the diameter would be 8 etc. But with the version 2.0.0dev4 if you set radius to be 3 than the diameter is 7, if you set the radius to be 4 than the diameter is 9 etc.

illume:
This issue should be fixed now. Thanks @MightyJosip

Issue #1122, opened 2019-06-27
Incorrect bounding rects returned from pygame.draw.circle


charlesej:
The bounding rects returned from `pygame.draw.circle` are incorrect in some instances. The following code gives an example of an incorrect bounding rect.

```python
import pygame

def create_bounding_rect(surface, start, surf_color):
    """Create a rect to bound all the pixels that don't match surf_color."""
    width, height = surface.get_clip().size
    xmin, ymin = width, height
    xmax, ymax = -1, -1

    for y in range(height):
        for x in range(width):
            if surface.get_at((x, y)) != surf_color:
                xmin = min(x, xmin)
                xmax = max(x, xmax)
                ymin = min(y, ymin)
                ymax = max(y, ymax)

    if -1 == xmax:
        # No points means 0 sized rect with the position at the start.
        return pygame.Rect(start, (0, 0))

    return pygame.Rect((xmin, ymin), (xmax - xmin + 1, ymax - ymin + 1))


surf_color = pygame.Color('black')
surface = pygame.display.set_mode((300, 300))
surface.fill(surf_color)
pos = surface.get_rect().center

bounding_rect = pygame.draw.circle(surface, pygame.Color('red'), pos, 21, 1)
actual_rect = create_bounding_rect(surface, pos, surf_color)

print('{} bounding rect'.format(bounding_rect))
print('{} actual bounding rect'.format(actual_rect))
```

The above code gives the following output showing the actual bounding rect is different than the bounding rect returned from `pygame.draw.circle`. The circle being drawn is bigger than what is being reported as drawn.

```
pygame 2.0.0.dev2 (SDL 2.0.9, python 3.7.2)
Hello from the pygame community. https://www.pygame.org/contribute.html
<rect(129, 129, 42, 42)> reported bounding rect
<rect(129, 129, 43, 42)> actual bounding rect
```

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit)
- pygame: 2.0.0.dev2 (SDL: 2.0.9) at f0de1695da4ba8ede907e09d415fce8d65f6dfc7

charlesej:
Transferring the discussion from #1121.

There are, however, a few items that need addressing:

- The change in the way circles are being drawn (now honoring the `center` parameter) should be noted in the draw documentation and maybe add an example for clarity.
[...]

PR #1386, opened 2019-10-08
Fixed pygame.draw.circle size

MightyJosip:
This should resolve issue #1384, if you need me to write the comments in code for this commit I will gladly do it. The problem was with the new algorithm we use to draw the circle we can only draw odd number of pixels (midpoint algorithm has the range of [-radius, radius] so that means it draws 2 * radius + 1 pixels). Algorithm we used before had the range of [-radius, radius - 1] so that means it used to draw 2 * radius pixels (one less). This had the effect if you run any code that called pygame.draw.circle() function. With this fix I still use the same algorithm, just move pixels based on the quadrant which makes circle have the diameter of 2 * radius (just like it used to be pre 2.0.0dev4).

illume said that eventually we should do sub pixel rendering, but until than something like this should work

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1753,24 +1753,31 @@ draw_arc(SDL_Surface *dst, int x, int y, int radius1, int radius2,
     }
 }
 
-
+/* This function is used just for circle drawing because Bresenham Circle Algorithm
+ * draws the circle with the odd diameter and in the original pygame it had even diameter
+ * It does scaling by that one pixel (direction based on relative quadrant to the center
+ * to keep symmetry). In any other drawing function use normal set_at function
+ */
 static void
 draw_circle_pixel(SDL_Surface *dst, int x0, int y0, int x1, int y1, Uint32 color)
 {
+    // (x0, y0) -> center of the circle, (x1, y1) -> pixel to bve drawn
+    // leading_bit -> used for fast quadrant calculation, 1 for negative and 0 for positive
+    // number (number is "vector" from center to the pixel)
     int x_leading_bit = ((x1-x0) > 0) - ((x1-x0) < 0) == 1 ? 0 : 1;
     int y_leading_bit = ((y1-y0) > 0) - ((y1-y0) < 0) == 1 ? 0 : 1;
     int quadrant = (x_leading_bit != y_leading_bit) + y_leading_bit + y_leading_bit + 1;
     if (quadrant == 1) {
-        set_at(dst, x1 - 1, y1 - 1, color);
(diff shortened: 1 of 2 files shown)


robertpfeiffer:
can you run this:
```python
import pygame
pygame.init()

flags|=pygame.SCALED
screen=pygame.display.set_mode((160,144), flags)
screen.fill((255,255,255))
pygame.display.flip()

pygame.draw.circle(screen, (100,100,100), (30,30), 29, 2)
pygame.draw.circle(screen, (100,200,200), (60,30), 29, 1)
pygame.draw.circle(screen, (200,100,100), (90,30), 29, 0)
pygame.draw.circle(screen, (100,200,100), (120,30), 29, 15)

for i in range(29, 5, -1):
    pygame.draw.circle(screen, [(i*30)%256]*3, (30,90), i, 0)

for i in range(29, 5, -1):
    pygame.draw.circle(screen, ((i%2)*255, 0, 0), (90,90), i, 1)

pygame.display.flip()
pygame.time.wait(5000)
pygame.image.save(screen, "test-draw-circles-"+pygame.version.ver+".png")
```
and post the result?

MightyJosip:
test-draw-circles-2 0 0 dev5

illume:
Thanks @MightyJosip It's looking good :)

Can we add an assertion in the tests? Something like...
```
assertEqual(bounding_rect.width, 2 * radius)
```

There is a test for the circle bounding rectangle in test/draw_test.py...
https://github.com/pygame/pygame/blob/a48f96d422d431f6b8e9976b466a1aed13448934/test/draw_test.py#L4699-L4732

illume:
for test/draw_test.py
```python
    def test_circle__diameter(self):
        """Ensures draw circle is twice size of radius high and wide."""
        surf = pygame.Surface((100, 100))
        color = (0, 0, 0, 50)
        center = surf.get_height() // 2, surf.get_height() // 2
        width = 1
        radius = 6
        bounding_rect = self.draw_circle(surf, color, center, radius, width)
        self.assertEqual(bounding_rect.width, radius * 2)
        self.assertEqual(bounding_rect.height, radius * 2)
```

MightyJosip:
Circle also had bad rect size (right and bottom were one pixel too much), it wouldn't affect drawing, but it would affect things connected to pygame.Rect

illume:
Here's the png from the test program:
test-draw-circles-2 0 0 dev3

illume:
I'm going to merge this now, and work in a separate branch to add more tests.

illume:
Related: https://github.com/pygame/pygame/issues/1122

PR #1388, opened 2019-10-08
Update Sprite.update and sprite Group.update docs

thiagojobson:
Hello,

I forgot to update the docs when I made #1375.

### DIFF
--- a/docs/reST/ref/sprite.rst
+++ b/docs/reST/ref/sprite.rst
@@ -86,7 +86,7 @@ Sprites are not thread safe. So lock them yourself if using threads.
    .. method:: update
 
       | :sl:`method to control sprite behavior`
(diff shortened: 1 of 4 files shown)


illume:
Oh. Thanks :)

Would you mind updating the _sprite.pyx as well? We need to keep them in sync now too.

thiagojobson:
Done. Please tell me if it is ok.

This will replace `sprites.py` in the future, right?

illume:
Cool. Thanks!

With the cython stuff it needs to be compiled to C, and the C file committed as well.
```
python3 setup.py cython build
```


I guess it might replace sprite.py in the future. It's a bit of an experiment at the moment.

thiagojobson:
Nice, I didn't know about cython!

PR #1393, opened 2019-10-10
Zero and negative sized circles do not draw anything or raise.

illume:
Added some commented out debugging printf for failing test.

For issue https://github.com/pygame/pygame/issues/1122 "Incorrect bounding rects returned from pygame.draw.circle"

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -710,11 +710,7 @@ circle(PyObject *self, PyObject *args, PyObject *kwargs)
 
     CHECK_LOAD_COLOR(colorobj)
 
-    if (radius < 0) {
-        return RAISE(PyExc_ValueError, "negative radius");
-    }
-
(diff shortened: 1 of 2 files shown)


src_c/draw.c
```diff
@@ -710,8 +710,8 @@ circle(PyObject *self, PyObject *args, PyObject *kwargs)
 
     CHECK_LOAD_COLOR(colorobj)
 
-    if (radius < 0) {
-        return RAISE(PyExc_ValueError, "negative radius");
+    if (radius < 1) {
+        return pgRect_New4(posx, posy, 0, 0);
```
charlesej: You could combine this check with the width check on line 717 since they are now returning the same thing.

illume: Good idea. Thanks.

PR #1402, opened 2019-10-11
Thick line bug fix.

### LINKED ISSUE #448: Thick Line Bug
Drawing thick lines is a bit buggy.

- pygame.draw.line with width>1 http://www.pygame.org/docs/ref/draw.html#pygame.draw.line
- clip_and_draw_line_width in https://github.com/pygame/pygame/blob/20f3b376ffb087fb4995f2d58570203355357a04/src/draw.c#L685
- [x] example showing error.
- [x] unit test (good first issue)
- [x] src/draw.c fix

---

pygame_bug

Which can result in this when making a bunch of short lines at different angles:
pygame_bug

Code for reproducing: https://pastebin.com/wNQj46E6
```python

#!/usr/bin/python3.4
import pygame, sys
mainClock = pygame.time.Clock()
from pygame.locals import *
pygame.init()
screen = pygame.display.set_mode((200,200),0,32)
while True:
    screen.fill((0,0,0))
    mx,my = pygame.mouse.get_pos()
        if event.type == QUIT:
            pygame.quit()
            sys.exit()
    pygame.draw.line(screen,(255,255,255),(50,50),(mx,my),30)
    pygame.display.update()
    mainClock.tick(40)

```

MightyJosip:
This should resolve #448. There was a bug when you draw lines that go over the edge of the surface it modified ending points before it draws it. This means it was possible that the slope of the lines drawn next to each other are not the same and that was the reason for the "holes" in the line with the width bigger than 1. In this pull moved rect calculation after you draw the line (to ensure that you draw the line with the starting points to have the same slope). With this pull you should no longer have this bug.


MightyJosip:
I really have no idea why is appveyor build failing. It works just fine for python 3.5, 3.6 and 3.7, but for some reason it doesn't for 2.7 (even though I didn't change a single line of python code). And the errors I am getting are even weirder (missing ';', '(', undeclared identifiers...). Shouldn't things like that also fail for other builds.

src_c/draw.c
```diff
@@ -946,12 +956,15 @@ static int
...
-    int loop;
     int xinc = 0, yinc = 0;
-    int newpts[4];
-    int range[4]; /* {xmin, ymin, xmax, ymax} */
+    int bounding_rect[4];
+    bounding_rect[0] = INT_MAX;
+    bounding_rect[1] = INT_MAX;
+    bounding_rect[2] = 0;
+    bounding_rect[3] = 0;
+    int original_values[4];
```
illume: You have to move type declarations above assignments for the old MSVC compiler that python2.7 uses.

src_c/draw.c
```diff
@@ -946,12 +956,15 @@ static int
...
+    memcpy(original_values, pts, sizeof(int) * 4);
```
illume: This memcpy needs to be moved below the type declarations.

src_c/draw.c
```diff
@@ -74,20 +74,20 @@ clip_and_draw_aaline(SDL_Surface *surf, SDL_Rect *rect, Uint32 color,
 static int
 clip_and_draw_line_width(SDL_Surface *surf, SDL_Rect *rect, Uint32 color,
                          int width, int *pts);
-static int
-clipline(int *pts, int left, int top, int right, int bottom);
+//static int
```
illume: C usually uses this type of comment.
```
/*
*/
```

I can't remember if `//` is ok in old-C.

illume:
Somehow the builders are stalling on the draw test (maybe an infinite loop?)
Both on travis linux : https://travis-ci.org/pygame/pygame/jobs/596903528

and on appveyor windows: https://ci.appveyor.com/project/pygame/pygame/builds/28060667/job/jwkw87f4p8nt6ib0

src_c/draw.c
```diff
@@ -958,13 +956,13 @@ clip_and_draw_line_width(SDL_Surface *surf, SDL_Rect *rect, Uint32 color,
...
     int bounding_rect[4];
+    int original_values[4];
+    int anydrawn = 0;
+    int loop;
     bounding_rect[0] = INT_MAX;
     bounding_rect[1] = INT_MAX;
     bounding_rect[2] = 0;
     bounding_rect[3] = 0;
-    int original_values[4];
-    memcpy(original_values, pts, sizeof(int) * 4);
```
illume: Did you accidentally remove this memcpy?

MightyJosip: I just moved it a bit down just before it is used first time, but I will try moving it up now

illume: Also, the memcpy src and dst should be swapped?

```
     memcpy(void *restrict dst, const void *restrict src, size_t n);
```


To this:
```
memcpy(pts, original_values, sizeof(int) * 4);
```

illume: (ignore my last comment about swapping it)

MightyJosip: Actually it was probably this line, I moved it too low (after I change pts) and that could maybe cause bug.

MightyJosip:
This was my last commit to this pull, if you decide to merge it that is OK with me because I won't change anything more here (unless you want something).

Issue #1423, opened 2019-10-15
The great examples/ cleanup of October 2019.


illume:
The great examples/ cleanup of October 2019.

- [x] reformat examples using black code formatter.
- [x] top description for every example.
- [x] check comments, add extra ones.
- [x] stop using `from pygame import *` because it's not very good practice. Use `import pygame as pg` instead.
- [ ] add a sprinkle of creative writing, and make them fun.

Stretch goals:
- [ ] make an examples/ zip file download.
- [ ] Include an example at bottom of each module.

related to: https://github.com/pygame/pygame/issues/494

mcpalmer1980:
is from pygame.locals import * okay?

illume:
Hi @mcpalmer1980 if you want to start mixer.fadein() or rec.collideline that'd be awesome. Drop into the #contributing channel on our Discord chat if you want to ask questions there on how to proceed (the first step is usually to write a unit test or two).

illume:
I'll close this now, because it's not October 2019 anymore.

PR #1427, opened 2019-10-16
Cleaning up example - eventlist.py

Starbuck5:
Removed the from pygame import * in favor of import pygame as pg and from pygame.locals import *. Also made the program quit when the user presses the close button in the top right.

In response to #1423

### DIFF
--- a/examples/eventlist.py
+++ b/examples/eventlist.py
@@ -4,13 +4,14 @@
 tool for learning about pygame events and input. At the
 top of the screen are the state of several device values,
 and a scrolling list of events are displayed on the bottom.
-
 This is not quality 'ui' code at all, but you can see how
 to implement very non-interactive status displays, or even
 a crude text output control.
 """
 
-from pygame import *
+import pygame as pg
+from pygame.locals import *
+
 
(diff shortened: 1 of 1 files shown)


illume:
Thanks!

Let me know if you want to continue on this example? If so here are some things...

- Looks like the constants like QUIT need to be pg.QUIT
- Once that is done from pygame.locals import * can be removed.
- I'm not sure why it's saying every line changed? What code editor are you using?

Otherwise we can merge your changes in, and someone else can continue it later.

Starbuck5:
I'll change the constants to be in terms of pygame.

Yeah the every line changed thing is weird. I just copy pasted the file from github into IDLE, messed with it, and then committed it on the github site to replace the existing one. Maybe it's because I copy pasted instead of downloading?

illume:
It might help to run the code through the [black](https://github.com/psf/black) code formatter.

```bash
pip install black
python -m black examples/eventlist.py
```

Starbuck5:
Alright, I'll fix the issues and submit another pull request that hopefully doesn't say every line is replaced.

Issue #1365, opened 2019-10-03
SCALED related tasks


illume:
Some SCALED related tasks left to do identified in the PR #1237 work.

- [ ] Docs (user-facing) for resizing API (example code is already there!)
- [ ] Document (user-facing) behaviour of 0,0 window size, in set_mode docs.
- [ ] document the SCALED/toggle_fullscreen example some more so it reads like a tutorial.
- [ ] API for SDL render quality hint, or just documenting how to use os.environ to set SDL hints?
- [ ] Format and delete commented out code. Make issues if there are todos, and comments about other stuff.
- [ ] Maybe in the future test if workarounds are still needed and add #ifdefs for old SDL versions
- [ ] Figure out which static state (using_GL) should be used where and document (C-side developer-facing) the behaviour of SDL_Renderer and GL context somewhere. Maybe also document that user-facing, it it interacts with modernGL or pyOpenGL
- [ ] Talk to upstream about the bugs we found. (make issues in bugzilla)

robertpfeiffer:
I can add a check, but I don't think division by zero is actually possible there. I suppose the static analysis errs on the side of "can't prove it won't happen"?

illume:
I can't link to it, but if you run the tool gives you the path in a browser it can reach to get there. So it tells you if X is A and Y is B, and you go down this path, and down this path... this error can happen.

robertpfeiffer:
I ran scan-build, and it told me that if `using_gl` is false and pg_renderer is not-NULL at the same time intoggle_fullscreen, bad things happen. I know these cannot happen at the same time. I added a check anyway, just to make the static analyser happy. Then I ran it away. Now it said in the check assuming this check fails (no exception raised) because using_gl is true, but pg_renderer is NULL. Then, later down the line it said assuming pg_renderer != NULL.

lordmauve:
> Get OPENGL|SCALED to work with mouse scaling

What's involved in this? Shall I try to create a patch?

robertpfeiffer:
Resizing SCALED windows should work without user intervention now, the API I added is preliminary for testing purposes, the previous preliminary resizing API is no longer needed.

> What's involved in this?
Access GL viewport and window size info in event handling code. Ensure this code path is not entered when GL is used via a renderer. For SCALED, we can just ask SDL for the default renderer dimensions.

> Shall I try to create a patch?
Sure, go ahead.

PR #1432, opened 2019-10-17
Allow pygame.SCALED | pygame.OPENGL

lordmauve:
Enable the use of `SCALED` with `OPENGL` in `pygame.display.set_mode()`.

This actually works fine; the OpenGL viewport is always the rectangle `(-1, -1), (1, 1)` regardless of the viewport size, so simply scaling the window (while preserving aspect ratio) means that OpenGL does the scaling. User code can continue assuming the original dimensions when constructing matrices.

The only place where the exact dimensions matter is when allocating framebuffer objects. Using the original dimensions will give a smaller texture; but texture access is usually linearly interpolated so the impact of this will usually just be slightly reduced quality. For more refinement there's already `pygame.display.get_window_size()`.

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -878,9 +878,6 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
             if (w == 0 || h == 0)
                 return RAISE(pgExc_SDLError,
                              "Cannot set 0 sized SCALED display mode");
-            if (flags & PGS_OPENGL)
-                return RAISE(pgExc_SDLError,
-                             "Cannot use OPENGL with SCALED mode");
             /*if (flags & PGS_RESIZABLE)
                 return RAISE(pgExc_SDLError,
                 "Cannot use RESIZABLE with SCALED mode");*/


illume:
Hey hey. Thanks :)

I tried with examples/glcube.py and changed the set_mode call to these two ones:
```
    pygame.display.set_mode((640, 480), OPENGL | DOUBLEBUF | FULLSCREEN | SCALED)
    pygame.display.set_mode((640, 480), OPENGL | DOUBLEBUF | SCALED)
```
and the example seems to run ok on a MacOS 10.13.6 with nvidia gfx.
- laptop no external monitor with resolution set to normal.
- laptop with external screen attached
- laptop with laptop screen set to scaled mode (high res), and the window takes up much of the screen.

However, I don't know this code as well as @robertpfeiffer and @dlon though, so I'll wait for one of them to come along to have a look before merging.

If there are some other tests I could do, please let me know?

robertpfeiffer:
I have three questions for @lordmauve 

1. Why do you want to do that? What is your use case? Is it just for mouse scaling?
2. Why don't you just use set_mode with OPENGL and render your own quad with the surface contents as its texture if you do openGL stuff anyway?
3. How should this work on Windows(DirectX) or macOS(Metal)? Did you test this on windows?

I don't think this should be merged in its current state, as it merely removes the three lines that raise an exception when somebody passes both `OPENGL` and `SCALED` to `set_mode()`.

At the very least, we would need to detect `OPENGL|SCALED`, and ensure that the normal OPENGL code path is not entered, and also ensure that SCALED uses the OpenGL backend of SDL_Renderer, even when others are available and faster. We would also have to raise an exception when OpenGL is not available.

And even then, trying to get a handle on the GL context used by SDL_Renderer means relying on implementation details. It would be another thing to get a handle on the SDL_Renderer used by SCALED with pygame.sdl2 functionality in order to superimpose e.g. high-res rendered text over blocky pixels. That would at least use the public API of SDL2 in a way that allows us to switch to newer versions of SDL2 without having to emulate old behaviour.

Right now, this works because of the order in which the GL context and renderer are created, but it will probably need the following changes to work properly: 

- when SCALED|OPENGL is passed don't enter OPENGL-only code path (make sure we don't create a dummy display surface twice and create a memory leak)
- when SCALED|OPENGL is passed, request openGL SDL_Renderer backend only
- raise an exception when that backend is not available
- set state->using_gl and gl_context variables

But depending on the use case and the intended behaviour, there might be more to do.

illume:
I tested with this on MacOS hinting at using metal and it seems to run still. The gl cube comes up.
```
        (640, 480), pygame.OPENGL | pygame.DOUBLEBUF | pygame.SCALED
SDL_HINT_RENDER_DRIVER=metal python3 examples/glcube.py
```

lordmauve:
The use case is to support retina displays.

I want to scale and to use the input scaling.

> Why don't you just use set_mode with OPENGL and render your own quad with the surface contents as its texture if you do openGL stuff anyway?

There is no surface, I am only doing OpenGL.

> At the very least, we would need to detect OPENGL|SCALED, and ensure that the normal OPENGL code path is not entered, and also ensure that SCALED uses the OpenGL backend of SDL_Renderer, even when others are available and faster.

I absolutely expect the normal OpenGL codepath. You

robertpfeiffer:
I thought about it some more, and I *think* you want to use GL with the automatic mouse scaling and viewport setting, in which case it should be easy to support by not using SDL_renderer at all

lordmauve:
If I could independently configure mouse scaling, and detect desktop resolution before initialising a mode, that would also be useful. Actually more useful - it would enable creating a "retro mode" where I can treat the viewport as, say, 320x200 and render it full-screen.

robertpfeiffer:
> If I could independently configure mouse scaling, and detect desktop resolution before initialising a mode, that would also be useful. Actually more useful - it would enable creating a "retro mode" where I can treat the viewport as, say, 320x200 and render it full-screen.

That was my motivation for SCALED with pixel graphics. What you want to do is already doable with gl libraries on top of pygame, but it would make the API more consistent to support it.

illume:
Ok. Thanks for all the thought on this.

We can merge this in?

... and handle the various other cases and combinations when using with a Renderer in the SCALED issue? https://github.com/pygame/pygame/issues/1365 
I still haven't tried on windows, but I guess I can add that task to #1365 too :)

robertpfeiffer:
I'm sorry for responding so late. This should not have been merged without a couple of test cases and additional exception handling, which I have now started working on.
Implementing this properly with toggle_fullscreen and handling resizing will duplicate a lot of code from SDL_Renderer (well, not copy-pasted, but the same functionality implemented again), which isn't pretty but doable.

To cover @lordmauve's use case (https://github.com/pygame/pygame/pull/1432#issuecomment-544119014), we should probably just expose a function that returns the size of the current desktop, or better yet a list with the sizes of each attached monitor. That would be much more future-proof than `list_modes()` or `mode_ok()`, which I think could even be deprecated (at least make a note in the docs) going forward. We could also think about scaling mouse coordinates to the range [-1,1] when the user requests that.

PR #1435, opened 2019-10-18
Examples cleanup

illume:
For https://github.com/pygame/pygame/issues/1423

- top description for every example.
- check comments, add extra ones.
- stop using `from pygame import *` because it's not very good practice. Use `import pygame as pg` instead.
- README.rst greatly improved with descriptions of each of the examples.


examples/glcube.py
```diff
@@ -1,16 +1,29 @@
...
+
+Amazing.
+
+Every frame we orbit the camera around a small amount
+creating the illusion of a spinning object.
+
+First we setup some points of a multicolored cube. Then we then go through
+a semi-unoptimized loop to draw the cube points onto the screen.
+
+Oopengl does all the hard work for us. :]
```
charlesej: Opengl has an extra o here.

illume: thanks

PR #1439, opened 2019-10-19
small deletions in time.rst

galexandreg:
I have no clue about how .rst files work, but i'm pretty sure this is not supposed to be there 😉


illume:
There some info about .rst and how to build docs here: https://www.pygame.org/wiki/Hacking?vv=1#Generating%20docs

PR #1441, opened 2019-10-20
set_mode and SCALED improvements

robertpfeiffer:
For SCALED issue: https://github.com/pygame/pygame/issues/1365
Related discussion was in: https://github.com/pygame/pygame/pull/1432 /cc @lordmauve 

---

This adds error handling, removes commented out code, and reverts the handling of unknown platforms in `toggle_fullscreen` to the previous, desired behaviour of toggling on supported platforms and failing silently on others.

Two new functions were added to the display module, exposing SDL2 functionality:
`pygame.display.get_renderer_info` and `pygame.display.get_desktop_sizes`

```python
>>> pygame.display.get_desktop_sizes()
[(1366, 768), (1920, 1080)]
>>> pygame.display.get_renderer_info()
("opengl", 0b1010)
```

`get_renderer_info()` is mainly here to replace the debug printfs I added to the C code and removed again to get this merged. I don't think it will need to be part of the API long-term.

`get_desktop_sizes()` is a replacement for `mode_ok(mode)` and `get_modes()` that is more useful in conjunction with modern SDL2 features such as borderless fullscreen. This could also return a list of rects, not sure about the API yet. I hope this covers @lordmauve 's use case better than `OPENGL|SCALED`.

While I was at it, I took another look at the API SDL2 gives us to query desktop geometry. There is a way to get a rect for the useable part of the screen, not covered by task bars, docks, and similar UI elements. I used this instead of the screen size when available (SDL2.0.5 and up).

I also enhanced the default behaviour of `set_mode()` when no display argument is given: Now it defaults to the display of an existing window, if there is none it uses the environment variable, and when there is no environment variable *or* existing window, it queries the OS for the mouse position and checks that against the desktop geometry. Without an explicit `display=N` argument, the screen that contains the mouse pointer will be used. If all else fails, it still defaults to the first display (display=0).

`OPENGL|SCALED` needed a fair bit of extra code to handle `toggle_fullscreen` and `resize_event` properly. Mouse scaling is still not implemented.


illume:
It looked like there was some testing of this already done in #contributing :)

Would you like me to test some things, or give feedback on something in particular? Or can I merge it in?

illume:
Fixed the conflict I made in the examples cleanup PR.

PR #1442, opened 2019-10-20
music fade in and example

mcpalmer1980:
This is my first time creating a pull request.

Added a fade_ms argument to the music.play() method and described it in music.rst.

Added a hopefully clean and well documented example of using it, along with set_endevent and other pygame/python features that might help new users.

### DIFF
--- a/examples/music_drop_fade.py
+++ b/examples/music_drop_fade.py
@@ -58,12 +58,14 @@ def play_file(filename):
         try:  # we must do this in case the file is not a valid audio file
             pg.mixer.music.load(file_list[-1])
         except pg.error as e:
-            print(e)  # print a description such as 'Not an Ogg Vorbis audio stream'
+            print(
+                e
+            )  # this will print a description such as 'Not an Ogg Vorbis audio stream'
             if filename in file_list:
                 file_list.remove(filename)
                 print("{} removed from file list".format(filename))
             return
-        pg.mixer.music.play()  # fade_ms=4000)
+        pg.mixer.music.play(fade_ms=4000)
         pg.mixer.music.set_volume(volume)
         pg.mixer.music.set_endevent(MUSIC_DONE)
 
@@ -88,7 +90,7 @@ def play_next():
 
         file_list.append(nxt)
         print("starting next song: ", nxt)
-    pg.mixer.music.play()  # fade_ms=4000)
+    pg.mixer.music.play(fade_ms=4000)
     pg.mixer.music.set_volume(volume)
     pg.mixer.music.set_endevent(MUSIC_DONE)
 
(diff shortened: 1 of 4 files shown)


illume:
Awesome :) Thanks. 🎈 🎉

mcpalmer1980:
As far as I can tell the python alerts are fixed and the travis-ci failure is not a problem with my code. I would guess it a fluke or a problem with pygame2/sdl2 on osx. Not sure what to do next.

examples/music_drop_fade.py
```diff
@@ -0,0 +1,206 @@
...
+    clock = pg.time.Clock()
+    pg.scrap.init()
+    pg.SCRAP_TEXT = pg.scrap.get_types()[0]  # TODO remove when  scrap module is fixed
+    clipped = pg.scrap.get(pg.SCRAP_TEXT)  # store the current text in the clipboard
+
+    # add the command line arguments to the file_list
+    for arg in sys.argv[1:]:
+        add_file(arg)
+    play_file("house_lo.ogg")  # play default music included with pygame
+    play_file("crap.ogg")  # TODO remove this test
```
illume: 💩 hehe

examples/music_drop_fade.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    screen = pg.display.get_surface()
+    surf = font.render(text, 1, (255, 255, 255))
+    y += surf.get_height()
+    x = (screen.get_width() - surf.get_width()) / 2
+    screen.blit(surf, (x, y))
+    return y
+
+
+MUSIC_DONE = pg.USEREVENT + 1  # create a user event called MUSIC_DOWN in the first slot
```
illume: We recently added: `pygame.event.custom_type()` https://www.pygame.org/docs/ref/event.html#pygame.event.custom_type

```python
MUSIC_DONE = pg.event.custom_type()
```

illume:
Looks great. I made two small comments. I guess just the 'crap.ogg' file should be removed (if it stops it from running?).

mcpalmer1980:
It was meant to be removed, but the 'crap.ogg' shouldn't stop it from running. It was actually alien1.gif renamed so to test a non audio file with an accepted extension, which prints "Not an Ogg Vorbis audio stream'. It should just say 'file not found.'

Maybe the line should stay? Accidental toilet humor never hurt anyone did it?

mcpalmer1980:
Still not sure about the Travis CI error. I don't remember it ocurring last time. Do you think it's because of the missing referenced file? The details don't seem to suggest this to me.

mcpalmer1980:
I think this is done now.

Issue #1210, opened 2019-07-30
SDL2: mouse vitual input mode not working


mcpalmer1980:
For some reason I can't get relative mouse input when the mouse is beyond the window's edge. I have grabbed input and hidden the mouse. The following code ran fine on pygame < 2

```
#! /usr/bin/env python2

import sys
import os
import pygame as pg
from pygame.constants import *

folder = '/home/mcpalmer/New Fonts/'

class fontviewer:
	KEY_SPEED = 10

	def __init__(self, directory = folder):
		pg.init()
		self.font_dir = directory
		info = pg.display.Info()
		w = info.current_w
		h = info.current_h
		pg.display.set_mode((int(w * 0.8), int(h * 0.8)))
		self.screen = pg.display.get_surface()
		self.clock = pg.time.Clock()
		self.y_offset = 0
		self.grabbed = True
		pg.mouse.set_visible(not self.grabbed)
		pg.event.set_grab(self.grabbed)


	def __del__(self):
		pg.quit()

		#sys.exit()
		#print('sys.exit')

	def Render(self, fonts, text = 'A display of font &N', **dparams):
		size = dparams.get('size', 32)
		color = dparams.get('color', (255,255,255))
		self.back = dparams.get('back', (0,0,0))

		fd = self.font_dir
		surfaces = []
		total_height = 0
		max_width = 0

		for name in fonts:
			print('loading {}'.format(fd+name))
			if name.endswith('.ttf'):
				try:
					font = pg.font.Font(fd + name, size)
				except:
					print('failed to load {}'.format(name))
			t = text.replace('&N', name.rpartition('.')[0])
			try:
				surf = font.render(t, 1, color, self.back)
			except:
				print('failed to render {}'.format(name))
			total_height += surf.get_height()
			max_width = max(max_width, surf.get_width())
			surfaces.append(surf)
		self.surfaces = surfaces
		rendered = pg.surface.Surface((max_width, total_height))
		rendered.fill(self.back)
		print('total height={}, max width={}'.format(total_height, max_width))
		print('surface={},{}'.format(rendered.get_width(), rendered.get_height()))

		y = 0
		center = int(max_width / 2)
		for surf in self.surfaces:
			w = surf.get_width()
			c = center - int(w/2)
			rendered.blit(surf, (c, y))
			y+= surf.get_height()
		self.surface = rendered.convert()
		self.max_y = rendered.get_height() - self.screen.get_height()

	def Save(self, name = 'fontviewer.png'):
[...]

PR #1451, opened 2019-10-23
Virtual Mouse Input

mcpalmer1980:
For https://github.com/pygame/pygame/issues/1210

It seems that virtual mouse input at least partially works on OSX, but on my Linux Mint 18.3 it definitely doesn't. These changes do fix it.

I'm not sure how to write an automated test for input code, but I can upload an example that works with this branch but not in any of the pygame 2.0 dev builds. It creates a huge surface showcasing all your fonts and lets you scroll vertically through them, entering and exiting virtual mode with a left click.


mcpalmer1980:
Tried a script that uses virtual input mode for dev4 on Windows and I got no input beyond window edges. I wanted to try my fork but instructions for building on Windows asks for msvc, which asked for 7gb install. Don't have that much space available for Windows right now.
I'm pretty sure my changes would fix the problem.

I think I'll upload the example I used, which is a bit messy, and a bare-bones one. If anyone thinks one is useful and worthy of inclusion, I'll clean it up. 

In the mean time, I'm open for any suggestions regarding a test for mouse.c and/or my changes. Other than manually setting position and then getting it, which isn't guaranteed to simulate real world mouse movement, I'm at a loss.

illume:
I'm seeing this error on Mac clang:
```
src_c/event.c:140:19: error: unused variable 'inputEvent' [-Werror,-Wunused-variable]
        SDL_Event inputEvent[2];
```

illume:
I like the examples :)
We don't have a font viewer type example already. I wonder if we can add the relative mouse stuff to the eventlist.py status display.

In the chat it sounded like you have already tested quite a lot of things. Let me know if there's anything in specific that you'd like me to test? I have mac/windows/linux machines set up.

For automated testing, just calling the code provides good value. It shows that the code paths do not segfault on 200+ different OS/python combinations. It also shows that the argument processing works. With runtime instrumentation in things like Valgrind, and ASan memory testers it allows those tools to test the code paths... and that the SDL code itself doesn't crash on different platforms. So just calling `pygame.mouse.set_visible()` in a test would be good. Then testing if you set it to visible, if the get_visible function returns that it's visible [and the opposite].

ps. We have 'interactive' tests. `python3 -m pygame.tests --interactive`, which allow you to ask the user a question. We don't have many of these, but it can be a good way to put into code things which are hard/impossible to test automatically.
- write the test a human should do as text. "Use examples.virtualmouse and see what the coordinates printed are XXX if you do ZZZ..."
- ask them a Y/N question if it worked.

illume:
There already is a basic test for get_visible set_visible.

src_c/event.c
```diff
@@ -137,10 +137,7 @@ pg_event_filter(void *_, SDL_Event *event)
 #pragma PG_WARN(Add event blocking here.)
 
     else if (type == SDL_KEYDOWN) {
-#ifdef WIN32
```
illume: here is the removed ifdef.

examples/font_viewer.py
```diff
@@ -0,0 +1,122 @@
+#! /usr/bin/env python3
+
+import pygame as pg
+from pygame.constants import *
+
+class fontviewer:
```
illume: `class fontviewer` should be `class FontViewer` to be pep8 friendly.

examples/font_viewer.py
```diff
@@ -0,0 +1,122 @@
...
+		self.clock = pg.time.Clock()
+		self.y_offset = 0
+		self.grabbed = False
+		self.Render('&N abcDEF789', size=48, color = (0,0,0), back_color = (255,255,255))
+		self.Display()
+		#viewer.Save()
+		#pg.mouse.set_visible(not self.grabbed)
+		#pg.event.set_grab(self.grabbed)
+
+	def Render(self, text = 'A display of font &N', **dparams):
```
illume: methods should be lower case for pep8 friendliness

Issue #1406, opened 2019-10-12
Add a method to check intersection between a rect and a line


thiagojobson:
The [SDL_IntersectRectAndLine](https://wiki.libsdl.org/SDL_IntersectRectAndLine) can be used to check intersection beetwen a rect and a line.

This could be something like:
`Rect.collideline((x1, y1), (x2, y2)) -> bool`

mcpalmer1980:
I would love that too. But collide_segment() probably shouldn't return a bool. Most of the time you would want to know which edge it collided with first. What we really need is some sort of raycasting functionality, which I have suggested before. https://github.com/pygame/pygame/issues/1152

illume:
@thiagojobson Good idea about a `Rect.collideline`
- It matches the other collide* functions (rect, point, ...)
- should be able to reuse the SDL_IntersectRectAndLine code.

Should it take a line like `((x0, y0, x1, y1))` as well/instead?

We could start with a Bool (because it's easy to use SDL_IntersectRectAndLine), and move to returning an edge(or intersection?) at a further date?

mcpalmer1980:
So I took a quick look at the docs for SDL_IntersectRectAndLine and it confirmed my assumption that it checks collisions for a segment, not a line. Also, although it returns a bool it also clips the line by modifying the line coordinate pointers. It then becomes almost trivial to determine which edge it collided with.

As for the python implementation I remain unsure. Other then with rects, I cannot think of any examples where pygame changes the values of a parameter. Perhaps collideline() should return the clipped line coordinates in the event of a collision, or false otherwise. Icing on the cake would order the returning endpoints so that the colliding edge is returned first so I don't need to perform a bunch of comparisons to determine where the collision occurred.

Then we need to consider how to describe our lines in a convenient way that fits with the other python methods. In pygame, 2-tuples seem the most common way to describe a point. I think that the collision method should be:
```
Rect.collideline((x1, y1), (x2, y2)) -> [(x1,y1), (x2,y2)] or False
```
We can then use it by calling something like:
``` bool
line = (enemy.get_pos(), target.get_pos())
clipped_line = my_sprite.rect.collideline(*line)
explosions.add(clipped_line[0])
```
As a side question, is pygame starting to adapt vectors to be used place of two-tuples? Should this be considered in the collidedline method? Because the pygame1.91 included in my distribution still lacked official vector support, I created my own class and have so much code that uses it that I never experimented with the vectors in pygame.

[...]

PR #1452, opened 2019-10-24
Export various SDL2 functions for resolution scaling

n0la:
This patch exports various needed functions for SDL2 resolution scaling, including `set_window_size()`, `get_logical_size()`, `set_logical_size()`, `set_scale()`, `get_scale()`, `set_integer_scale()`, `get_integer_scale()`, `set_hint()`, `get_hint()`, `get_viewport()` and
`set_viewport()`.

This patch also always make a renderer (which is needed for scaling), regardless of whether the `SCALED` flag is set. The `SDL_PumpEvents` is also gone, as it forced realignment of the window with the monitor, making the code fail in our multi monitor environment.

A bit of background perhaps up front: in our company we use C & SDL, and use various scaling techniques to bring a multimedia application that has been programmed in one particular resolution, to a wide variety of different embedded devices which have a wide variety of different monitors connected. We also have various multi monitor setups, where our SDL window stretches across various screens (often 2, sometimes three at a time). C gives us full control to implement various stretching/scaling methods, while pygame does not. And we really want to move to Python :)

Here are some stretching/scaling methods that are now available through these newly exported functions. First, let's assume that the physical monitor's best resolution is 1280x1024 (4:3), while the resolution of the media application is 1920x1080 (aka Full HD).

The first method is called "letter boxing", where SDL puts black bars top and bottom to fill the space that is remaining after scaling up as high as possible. This is implemented per default in SDL, if the physical aspect ratio and the logical aspect ratio are not the same.

```python
pygame.display.set_mode((1280, 1024), flags = 0)
pygame.display.set_logical_size((1920, 1080))
```

The next method is the inverse of letter boxing, called "overscan". Here the height (instead of the width) is scaled accordingly, and the difference in width from the actual and the logical aspect ratio is simply cut off left and right:

```python
pygame.display.set_mode((1280, 1024), flags = 0)
pygame.display.set_hint("SDL_RENDER_LOGICAL_SIZE_MODE", "overscan")
pygame.display.set_logical_size((1920, 1080))
```

The next method simply stretches the smallest possible fraction of the logical screen into the actual physical aspect ratio. Cubes now become rectangles, circles become ellipse. This might not be what you want, but it is useful in some scenarios. For example when the logical, and actual ratio are not that far apart (i.e. 16:10 vs. 16:9), and you don't mind a bit of stretching:

```python
pygame.display.set_mode((1280, 1024), flags = 0)
pygame.display.set_logical_size((1920, 1080))
# Calculate scaling factor between the two, and tell SDL2 the new factor
pygame.display.set_scale((1280 / 1920, 1024 / 1080))
# Set the view port to hard values, causing stretching
pygame.display.set_viewport(pygame.Rect(0, 0, 1920, 1080))
```

Last method is: fiddle with the letter boxing used in part one. The letter boxing algorithm of SDL2 does the following (roughly):

```python
pygame.display.set_mode((1280, 1024), flags = 0)
pygame.display.set_logical_size((1920, 1080))
# Reset scale so that any SDL/pygame internal shenanigans are removed
pygame.display.set_scale((1.0, 1.0))
# Figure out scaling factor for the width
scale = (1280 * 1.0) / 1920
x = 0
w = 1280
# And apply that to the height
h = int(math.ceil(1080 * scale))
# Magic happens here:
y = (1024 - h) / 2
pygame.display.set_viewport(pygame.Rect(x, y, w, h))
pygame.display.set_scale((scale, scale))
```

You can now fiddle with `y` to make letter boxing as you want it. Remove the division, and the black bar that is needed to fill the remaining space is on top, or set it to zero, and you have it at the bottom. Depending on your graphical particularities, you can actually draw something into this black space to make your media application look like it was designed for this resolution while preserving aspect ratio.

There are a lot of caveats to the changed code. Firstly, resizing is not handled. You'd have to do that manually (but its not hard). And secondly, many methods of SDL are rather peculiar **when** they want to be called. For example move the call to `set_hint()` before `set_mode()` and you have a segfault.

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -763,6 +763,234 @@ pg_gl_get_attribute(PyObject *self, PyObject *arg)
 
 #if IS_SDLv2
 static PyObject *
+pg_set_window_size(PyObject *self, PyObject *arg)
+{
+    int winw = 0, winh = 0;
+    SDL_Window *win = pg_GetDefaultWindow();
+
+    if (win == NULL) {
+        return RAISE(pgExc_SDLError, "No open window/renderer");
+    }
+
+    if (!PyArg_ParseTuple(arg, "(ii)", &winw, &winh)) {
+        return Py_False;
+    }
+
+    SDL_SetWindowSize(win, winw, winh);
+    return Py_True;
+}
+
+static PyObject *
+pg_get_logical_size(PyObject *self, PyObject *arg)
+{
+    int w = 0, h = 0;
+
+    if (pg_renderer == NULL) {
+        return RAISE(pgExc_SDLError, "No open window/renderer");
+    }
+
+    SDL_RenderGetLogicalSize(pg_renderer, &w, &h);
+    return Py_BuildValue("(ii)", w, h);
+}
+
+static PyObject *
+pg_set_logical_size(PyObject *self, PyObject *arg)
+{
+    int w = 0, h = 0;
+    SDL_Surface *surf = NULL;
+    PyObject *surface = NULL;
+
+    if (pg_renderer == NULL) {
+        return RAISE(pgExc_SDLError, "No open window/renderer");
+    }
+
+    if (!PyArg_ParseTuple(arg, "(ii)", &w, &h)) {
+        return Py_False;
+    }
+
+    if (SDL_RenderSetLogicalSize(pg_renderer, w, h) != 0) {
+        return Py_False;
+    }
+
+    if (pg_texture != NULL) {
+        SDL_DestroyTexture(pg_texture);
+        pg_texture = NULL;
+
+        pg_texture = SDL_CreateTexture(pg_renderer, SDL_PIXELFORMAT_ARGB8888,
+                                       SDL_TEXTUREACCESS_STREAMING, w, h);
+        if (pg_texture == NULL) {
+            return Py_False;
+        }
+
+        surf = SDL_CreateRGBSurface(SDL_SWSURFACE, w, h, 32,
+                                    0xff << 16, 0xff << 8, 0xff, 0);
+        if (surf == NULL) {
+            return Py_False;
+        }
+
+        surface = pgSurface_New2(surf, 1);
+        pg_SetDefaultWindowSurface(surface);
+        Py_XDECREF(surface);
+        surface = NULL;
+    }
+
+    return Py_True;
+}
+
+static PyObject *
+pg_set_scale(PyObject *self, PyObject *arg)
+{
+    float x = 1.0, y = 1.0;
+
+    if (pg_renderer == NULL) {
+        return RAISE(pgExc_SDLError, "No open window/renderer");
+    }
+
+    if (!PyArg_ParseTuple(arg, "(ff)", &x, &y)) {
+        return Py_False;
+    }
+
+    if (SDL_RenderSetScale(pg_renderer, x, y) == 0) {
+        return Py_True;
+    }
+
+    return Py_False;
+}
+
+static PyObject *
+pg_get_scale(PyObject *self, PyObject *arg)
+{
+    float x = 0, y = 0;
+
+    if (pg_renderer == NULL) {
+        return RAISE(pgExc_SDLError, "No open window/renderer");
+    }
+
+    SDL_RenderGetScale(pg_renderer, &x, &y);
+    return Py_BuildValue("(ff)", x, y);
+}
+
+static PyObject *
+pg_get_integer_scale(PyObject *self, PyObject *arg)
+{
+    if (pg_renderer == NULL) {
+        return RAISE(pgExc_SDLError, "No open window/renderer");
+    }
+
+    if (SDL_RenderGetIntegerScale(pg_renderer)) {
+        return Py_True;
+    }
+
+    return Py_False;
+}
+
+static PyObject *
+pg_set_integer_scale(PyObject *self, PyObject *arg)
+{
+    PyObject *enable = NULL;
+    int val = 0;
+
+    if (pg_renderer == NULL) {
+        return RAISE(pgExc_SDLError, "No open window/renderer");
+    }
+
+    if (!PyArg_ParseTuple(arg, "O", &enable) || enable == NULL) {
+        return Py_False;
+    }
+
+    val = PyObject_IsTrue(enable);
+    Py_DECREF(enable);
+
+    if (SDL_RenderSetIntegerScale(pg_renderer, val) == 0) {
+        return Py_False;
+    }
+
+    return Py_True;
+}
+
+static PyObject *
+pg_set_hint(PyObject *self, PyObject *arg)
+{
+    char *hint = NULL, *value = NULL;
+    SDL_bool ret = 0;
+
+    if (!PyArg_ParseTuple(arg, "ss", &hint, &value)) {
+        return Py_False;
+    }
+
+    ret = SDL_SetHint(hint, value);
+    PyMem_Free(hint);
+    PyMem_Free(value);
+
+    return (ret == SDL_TRUE ? Py_True : Py_False);
+}
+
+static PyObject *
+pg_get_hint(PyObject *self, PyObject *arg)
+{
+    char *hint = NULL;
+    char const *value = NULL;
+
+    if (!PyArg_ParseTuple(arg, "s", &hint)) {
+        return Py_False;
+    }
+
+    value = SDL_GetHint(hint);
+    PyMem_Free(hint);
+
+    if (value == NULL) {
+        Py_RETURN_NONE;
+    }
+
+    return Text_FromUTF8(value);
+}
+
+static PyObject *
+pg_get_viewport(PyObject *self, PyObject *arg)
+{
+    SDL_Rect rect = {0};
+
+    if (pg_renderer == NULL) {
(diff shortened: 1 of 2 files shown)


robertpfeiffer:
I would suggest adding this functionality to pygame._sdl2 instead. Some of this is likely to break existing games, and we are still working to make pygame 2.0 compatble with pygame 1.9, so games written for old versions of pygame can run on machines with only the new SDL2-based pygame installed. The whole point of SCALED was to not expose all this to existing games!

Some of this functionality doesn't belong into display.c but would be useful in genera (set_hint functionality should be exposed  via pygame._sdl2 probably). We could probably cherry-pick some of this, and move the rest into pygame._sdl2.

We could also provide a set_physical_size API for SCALED without breaking existing use cases.

robertpfeiffer:
Also, I don't think your code that changes the logical size of the renderer doesn't change the display surface?

n0la:
> EDIT: I see you're a first-time contributor. Welcome to the PyGame project!
> 

Thanks :)

> I would suggest adding this functionality to pygame._sdl2 instead. Some of this is likely to break existing games, and we are still working to make pygame 2.0 compatble with pygame 1.9, so games written for old versions of pygame can run on machines with only the new SDL2-based pygame installed. The whole point of SCALED was to not expose all this to existing games!
> 

That is still true, except for the `SDL_PumpEvents()` thing.

> Some of this functionality doesn't belong into display.c but would be useful in general (set_hint functionality should be exposed via pygame._sdl2 probably). We could probably cherry-pick some of this, and move the rest into pygame._sdl2.
> 

That is good to hear. Please apologise the fact that I stuffed everything into display, as I had no real sense on where things should be going in general. See this as a PoC, and not a final commit.

> I don't think your code that changes the logical size of the renderer returns the display surface. You would have to call pygame.display.get_surface to get it. Do you even use software rendering in your multimedia application? If you use the renderer directly, then I would definitely suggest you use pygame._sdl2, which exposes the SDL_Renderer API.
> 

That is true, we are getting the surface manually right now. That is simply an oversight on my part. Once the `_sdl2` exposes everything we need, we can fully utilise it.

> On the other hand, we can merge your code for changing the physical screen size `pg_set_window_size()` while keeping the logical size and the display surface as-is without impacting existing use cases.

Great to hear.

There is one more thing I'd like to say: SDL2 scaling API (i.e. `SDL_SetRenderLogicalSize()`) as it stands with letter boxing is not great, [so much so in fact that even the author said it needs reworking](https://forums.libsdl.org/viewtopic.php?p=38664). Much of the SDL2 code that I have seen (and written), only use this API partially, and then user the other methods (i.e. set scale, set viewpoint, overscan, self-written letter boxing etc.) to get the desired result that fits best for the given media application and target screen. What I am trying to say is: if your custom scaling API is not playing the rather specific whistle tones that a user might need, it is always good to have the appropriate SDL2 functions around to roll your own.

robertpfeiffer:
You know PyGame is kind of slow if you use it for full HD games and multimedia stuff, right? SCALED was meant as a simplified API for pixel art games so resolutions between gameboy and QVGA can be used on retina/HD/4K displays without worrying about the physical resolution or melting your CPU with `pygame.transform.scale`. Scaling the surface up also allows us to keep the physical screen size and use borderless windowed fullscreen.

We haven't paid much attention to scaling the game *down*, as that would be the worst of both worlds: First you push lots of pixels around in software, then you copy all these pixels to the GPU!

Another thing you might consider doing (something people did before SCALED was a thing) is using openGL to render a textured quad which you can position where-ever you want it. I posted a tutorial here: https://blubberquark.tumblr.com/post/185013752945/using-moderngl-for-post-processing-shaders-with because people asked on the Discord, but since you're a multimedia guy, you already know how to use openGL. We are not using GL for SCALED because Apple is discouraging it and we might get better mac os performance in the future with SDL_Renderer's metal backend, or with the DirectX backend on Windows, but if you want more control you can also draw your own quad or quads with GL.

In that case, you'd need to know desktop/window geometry and a way to position the window, and you'd keep track of scaling yourself.

> The SDL_PumpEvents is also gone, as it forced realignment of the window with the monitor, making the code fail in our multi monitor environment.

Can you elaborate on that part? I have a multi-monitor setup which I use for testing, as does @illume, and we haven't had any problems so far.

n0la:
After some careful consideration, perhaps it is best to put `{set,get}_logical_size` (among others) into `_sdl2.Renderer`. I will prepare another pull request for that in the near future.

> You know PyGame is kind of slow if you use it for full HD games and multimedia stuff, right? SCALED was meant as a simplified API for pixel art games so resolutions between gameboy and QVGA can be used on retina/HD/4K displays without worrying about the physical resolution or melting your CPU with `pygame.transform.scale`. Scaling the surface up also allows us to keep the physical screen size and use borderless windowed fullscreen.
> 
> We haven't paid much attention to scaling the game _down_, as that would be the worst of both worlds: First you push lots of pixels around in software, then you copy all these pixels to the GPU!
> 

The trick is: we don't move a lot of pixels around. Most of it is static textures, or changes rather slowly or not that often. Yet high resolution matters (it needs to be pretty), and it needs to stretch several monitors.

> Another thing you might consider doing (something people did before SCALED was a thing) is using openGL to render a textured quad which you can position where-ever you want it. I posted a tutorial here: https://blubberquark.tumblr.com/post/185013752945/using-moderngl-for-post-processing-shaders-with because people asked on the Discord, but since you're a multimedia guy, you already know how to use openGL. We are not using GL for SCALED because Apple is discouraging it and we might get better mac os performance in the future with SDL_Renderer's metal backend, or with the DirectX backend on Windows, but if you want more control you can also draw your own quad or quads with GL.
> 

Apple is discouraging OpenGL? That seems to be a strategic fault on Apple's part.

> In that case, you'd need to know desktop/window geometry and a way to position the window, and you'd keep track of scaling yourself.
> 

That sounds doable. Thanks for the input/suggestions, and your nicely written tutorial :)

> > The SDL_PumpEvents is also gone, as it forced realignment of the window with the monitor, making the code fail in our multi monitor environment.
> 
> Can you elaborate on that part? I have a multi-monitor setup which I use for testing, as does @illume, and we haven't had any problems so far.

I am still investigating. I suspect that some event is triggering an unwanted window resize. This change should definitely go from the push request though.

illume:
Very excited about exposing these changes in some form. Thanks @n0la !

I really appreciate your explanations of those features.


I'll explain a few things...

1) The header docs are generated from .rst files. See instructions here: https://www.pygame.org/wiki/Hacking?vv=1#Generating%20docs So rather than editing the .h files, you write the display.rst file and then run `python3 setup.py docs` to generate the .h doc files.

2) Looks like SDL_RenderGetIntegerScale is in a newer version of SDL. That would need to be #ifdef'd out for older versions.

3) be good if things could be made private (prefixed with "_"), and marked experimental in the docs until they are 'done' (more on this below).


As @robertpfeiffer mentioned, we're trying to concentrate on backwards compatible changes for SDL2 with pygame 2.0, and newer features in pygame 2.1. But of course we are getting side tracked a bit with the _sdl2 module, and the SCALED stuff (and other shiny new features). So it'd be preferable if you could put them in the experimental _sdl2 module if possible. In there we're trying to keep close to the SDL API, but making it 'pythonic/OO' where ever that makes sense.

I don't want to block you from doing what you're doing however. If they need to go into the display module, can you please prefix them with a "_", and also mark them as experimental in the docs?

To be non-experimental:
- the API would need to be discussed for usability with a few different people. See the discussion for the new Rect.clipline method as an example at https://github.com/pygame/pygame/issues/1406
- we need an example/ program(s) showing how features are used. Very much like your programs. Perhaps changing an existing example to use the new functionality where appropriate. This lets people see how the API is used, and also helps people test the features. The example(s) should be written to be read like a tutorial if possible, with perhaps too many comments than are normally sensible.
- unit tests at least calling the functions at a minimum. This helps us check if things are broken on various platforms automatically. It also tests the SDL functions themselves are not crashing things.
- then we'd need to seek feedback after a .dev release is published, and maybe a few changes of the API(s) would happen again at that point.

Obviously that's a lot of work, just explaining the process we'd like to take for new features. It's fine to work on things incrementally, but in that case prefixing things with "_" marking them private and experimental is the way to go IMHO. It's also fine if you don't want to do some/any of this work on making the APIs stable/polished. Just be aware we may move/change the experimental APIs in the future.

Sorry for the walls of text :) Thanks again.

robertpfeiffer:
Just a heads up for OP. This week I might add some functionality/try to use existing APIs on the android branch to get a reference to the window and renderer available via pygame._sdl2.

The reference to the renderer is needed to add HD on-screen controls to old pixel art games.

robertpfeiffer:
I got this to work ```python
import pygame as pg
pg.init()
from pygame._sdl2 import Window, Texture, Image, Renderer
screen = pg.display.set_mode((800, 600), pg.SCALED | pg.RESIZABLE)
win = Window.get_window_from_ID(pg.display.get_window_ID())
renderer = Renderer.get_window_renderer(win)
```

getting a direct handle on the renderer via the _sdl2 stuff is the way to go. I'll close this once the functionality is merged.

robertpfeiffer:
extremely WIP branch here https://github.com/pygame/pygame/pull/1472. I won't work on this for the forseeable future, but you get the gist.

PR #1456, opened 2019-10-25
Fix Virtual Mouse Input in SDL2

mcpalmer1980:
Calls SDL_MouseSetRelativeMouseMode from pygame.set_grab() and pygame.set_visible() as necessary.
Removed redundant and unrelated example.
Will add font_viewer.py separately

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -46,8 +46,8 @@ static int have_registered_events = 0;
  *objects through the event queue.
  */
 
-#define USEROBJECT_CHECK1 (Sint32)0xDEADBEEF
-#define USEROBJECT_CHECK2 (Sint32)0xFEEDF00D
+#define USEROBJECT_CHECK1 0xDEADBEEF
+#define USEROBJECT_CHECK2 0xFEEDF00D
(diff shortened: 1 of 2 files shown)


src_c/mouse.c
```diff
@@ -54,11 +54,11 @@ mouse_set_pos(PyObject *self, PyObject *args)
             SDL_RenderGetScale(sdlRenderer, &scalex, &scaley);
             SDL_RenderGetViewport(sdlRenderer, &vprect);
 
-            x += vprect.x;
-            y += vprect.y;
+            x+=vprect.x;
+            y+=vprect.y;
 
-            x = (int)(x * scalex);
```
illume: was changing this intentional?

src_c/event.c
```diff
@@ -46,8 +46,8 @@ static int have_registered_events = 0;
  *objects through the event queue.
  */
 
-#define USEROBJECT_CHECK1 (Sint32)0xDEADBEEF
```
illume: Did you mean to change this?

PR #1472, opened 2019-10-31
expose window, renderer created by display.c to _sdl2 video

robertpfeiffer:
---

for https://github.com/pygame/pygame/issues/605
related to: https://github.com/pygame/pygame/pull/1452


src_c/display.c
```diff
@@ -2715,6 +2728,8 @@ static PyMethodDef _pg_display_methods[] = {
      METH_NOARGS, "provisional API, subject to change"},
     {"get_desktop_sizes", (PyCFunction)pg_get_desktop_screen_sizes,
      METH_NOARGS, "provisional API, subject to change"},
+    {"get_window_ID", (PyCFunction)pg_get_window_ID,
```
dlon: Can we return a `Window` here?

robertpfeiffer: I might make this a private API.
I'll probably add the function to get the window to _sdl2.Window directly, and call it from_display() or something

illume: `Window.from_display()` sounds good to me. So `display` is not bound to `video`. display.c could export it in c_api so `video` can use it.

examples/exposerenderer.py
```diff
@@ -0,0 +1,58 @@
+import pygame as pg
```
illume: Needs a header like the other examples.

examples/exposerenderer.py
```diff
@@ -0,0 +1,58 @@
...
+
+screen = pg.display.set_mode(RES, pg.SCALED | pg.RESIZABLE)
+
+done = False
+
+i = 0
+j = 0
+
+win = Window.from_display_module()
+#win = Window.from_ID(pg.display._get_window_ID())
```
illume: _get_window_ID can be removed

src_c/cython/pygame/_sdl2/video.pyx
```diff
@@ -193,6 +198,26 @@ cdef class Window:
...
+        cdef Window self = cls.__new__(cls)
+        cdef SDL_Window* window = pg_GetDefaultWindow()
+        if not window:
+            raise error()
+        self._win=window
+        SDL_SetWindowData(window, "pg_window", <PyObject*>self)
+        return self
+
+    @classmethod
+    def from_ID(cls, windowid):
```
illume: Can from_ID be removed?

src_c/display.c
```diff
@@ -2197,6 +2197,19 @@ pg_get_desktop_screen_sizes(PyObject *self, PyObject *args)
     return result;
 }
 
+static PyObject *
+pg_get_window_ID(PyObject *self, PyObject *args)
```
illume: I think this can be removed too.

illume:
There's a couple of LGTM alerts.
The mac build on travis is crashing with:
```c
src_c/display.c:2203:15: error: unused variable 'result' [-Werror,-Wunused-variable]

    PyObject *result;

              ^
```

robertpfeiffer:
The one bug I can't fix easily without also cythonising display.c: How do we prevent the user from getting a renderer or a handle on a window and then invalidating it by creating a new video with set_mode? How do we prevent the user from changing the display size and invalidating the display surface? I'd say that for now, we can't. We can easily prevent the user from deallocating the borrowed window through refcounting though, which would reliably crash the game.

robertpfeiffer:
As far as I am concerned, we could merge this now. It could use some more documentation, but the whole api is in flux anyway.

PR #1473, opened 2019-10-31
Fix invalid character for surname in author comment

VilhelmPrytz:
The "ö" showed up as an invalid character. This has been replaced with a proper "ö". See attached image of how it showed up previously (I'm aware of the fact that GitHub interprets the invalid character correctly somehow).

Author in editor

### DIFF
--- a/src_c/_sdl2/touch.c
+++ b/src_c/_sdl2/touch.c
@@ -1,6 +1,6 @@
 /*
   pygame - Python Game Library
-  Copyright (C) 2019 David Lönnhager
+  Copyright (C) 2019 David LÃ¶nnhager
 
   This library is free software; you can redistribute it and/or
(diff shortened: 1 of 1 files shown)


illume:
Thanks :)

To understand this better, what did you change on that line? It looks the same to me.
What OS are you on, and which editor are you using?

With MacOS, and Sublime Text/vim it looks ok.

cheers,

VilhelmPrytz:
I am on Ubuntu 18.04. The editor I was using was Visual Studio Code but it looked the same when using vim. I literally just replaced it with "ö" :smile:

It's an odd one, looking at the diff generated by GitHub it looks like I've changed nothing at all. But on my local computer, `git diff` shows this.
pygame_git_diff

PR #1485, opened 2019-11-04
Aliens example: make bomb sprite background transparent

robertpfeiffer:
The bomb sprite had a blue background ~~on my machine~~ since 1.9.6. It's not transparent, just blue.

### DIFF
--- a/examples/aliens.py
+++ b/examples/aliens.py
@@ -198,6 +198,7 @@ class Bomb(pg.sprite.Sprite):
     def __init__(self, alien):
         pg.sprite.Sprite.__init__(self, self.containers)
         self.image = self.images[0]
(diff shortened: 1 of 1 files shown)


illume:
I guess it works with pygame 1.9.6 on your machine? And this is an issue with pygame and transparency?

illume:
This should be fixed in pygame, so closing this.

robertpfeiffer:
Actually, no. Even in 1.9.6, the background of the sprite is just a dark blue, like the sky.

robertpfeiffer:
It just bugged me because aliens is my goto test case

illume:
I'd prefer not to change this.

PR #1486, opened 2019-11-04
fix #738, zero-size rotozoom surfaces

### LINKED ISSUE #738: transform.rotozoom result in segmentation fault for surfaces with zero width or height
Here is a small example that results in a seg fault.
```
import pygame

s = pygame.Surface((10, 0))
pygame.transform.scale(s, (10, 2))
#pygame.transform.rotozoom(s, 30, 1) will fail as well
```
pygame should probably check before performing scale or rotozoom and throw an error if width or height is zero.

robertpfeiffer:


robertpfeiffer:
Hey @illume, this currently fails because the test tests for something that doesn't work. Should I mark the test as expected failure, or create another issue to merge the failing test and delete the test here?

illume:
Hi ya.

yeah, mark it with expectedFailure if it's broken. If it's broken in the 'causes memory corruption or segfaults' kind of way skip the test.

robertpfeiffer:
rotating an image 90° four times, or 180° two times, or 360° once don't quite produce the original image in the way you would expect, but that's not the original issue here.

PR #1488, opened 2019-11-05
Fix for emtpty string/None font render

### LINKED ISSUE #1478: Why do fonts that render nothing show up as a vertical black line??
reported by "The New St. Paul aka MintFan"

MightyJosip:
This should fix #1478.

I would like if we can use this pull for discussion about what should be the result of the pygame.Font.render() if you use empty string or None as the first parameter. So far it used to be Surface with the size of (1, height_of_font). This means it would be drawn with the surface.fill(). I changed this to be the size of (0, height_of_font). With this you won't see anything on the screen, so basically this bug won't exist anymore. Now other possibility are to return size (0, 0). This would maybe mean less memory usage, but could maybe disable other functionalities of the Font (not sure, just thought).

To conclude, should the result of pygame.font.Font().render("") be 
<Surface(0xf(height)x32 SW)>
or
<Surface(0x0x32 SW)>

### DIFF
--- a/docs/reST/ref/font.rst
+++ b/docs/reST/ref/font.rst
@@ -161,7 +161,7 @@ loaded instead.
 
       The Surface returned will be of the dimensions required to hold the text.
       (the same as those returned by Font.size()). If an empty string is passed
-      for the text, a blank surface will be returned that is one pixel wide and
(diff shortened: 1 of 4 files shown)


illume:
Thanks @MightyJosip :)

Issue #1491, opened 2019-11-07
set_viewport constructs broken SDL_Rect for SDL_RenderSetViewport


n0la:
Hi,

the recently added methods for setting the view port do not work for me. I use the current github master, and a simple test code looks like this:

```python
#!/usr/bin/python3

import pygame
from pygame._sdl2 import Texture, Renderer, Window

win = Window(title = "IN HOOTS WE TRUST", size = (800, 600))
r = Renderer(window = win)
r.logical_size = (1920, 1080)

print(r.get_viewport())
rect = pygame.Rect(0, 0, 1920, 1080)
print(rect)
r.set_viewport(rect)
print(r.get_viewport())
```

Output is as follows:

```
$ python3 pygame-error.py
pygame 2.0.0.dev7 (SDL 2.0.10, python 3.7.5)
Hello from the pygame community. https://www.pygame.org/contribute.html
<rect(0, 180, 1920, 1080)>
<rect(0, 0, 1920, 1080)>
<rect(0, 1080, -190810624, -223577536)>
```
I am not quite sure yet where this error happens within the bowels of pygame.

illume:
Thanks :) It's fixed with a unittest in #1496

PR #1496, opened 2019-11-10
SDL2: Add video_test.py and test Renderer.set_viewport.

illume:
Fixes https://github.com/pygame/pygame/issues/1491 issue.

- Renderer.set_viewport test added.
- There was an issue with using pgRect_FromObject, which always requires the returned rect to be used, not the one passed in. Because sometimes(as in when it was used here) it does not use the pointer passed in.

---
SDL_Render mega issue #605
related work done in: https://github.com/pygame/pygame/pull/1472

### DIFF
--- a/src_c/cython/pygame/_sdl2/video.pyx
+++ b/src_c/cython/pygame/_sdl2/video.pyx
@@ -1012,11 +1012,15 @@ cdef class Renderer:
             if SDL_RenderSetViewport(self._renderer, NULL) < 0:
                 raise error()
             return
-        cdef SDL_Rect rect
-        if pgRect_FromObject(area, &rect) == NULL:
-            raise TypeError("the argument is not a rectangle or None")
-        if SDL_RenderSetViewport(self._renderer, &rect) < 0:
+
+        cdef SDL_Rect tmprect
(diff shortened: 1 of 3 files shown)


src_c/cython/pygame/_sdl2/video.pyx
```diff
@@ -1012,12 +1012,16 @@ cdef class Renderer:
             if SDL_RenderSetViewport(self._renderer, NULL) < 0:
                 raise error()
             return
-        cdef SDL_Rect rect
-        if pgRect_FromObject(area, &rect) == NULL:
-            raise TypeError("the argument is not a rectangle or None")
-        if SDL_RenderSetViewport(self._renderer, &rect) < 0:
+
+        cdef SDL_Rect tmprect
+        cdef SDL_Rect *rectptr = pgRect_FromObject(area, &tmprect)
```
charlesej: There are a few other places in this file where the returned pointer from `pgRect_FromObject()` is not being used.

- Image methods
  - `__init__()`
  - `draw()`

- Renderer method
  - `to_surface()`

illume: I've updated these ones too. But didn't add tests for these.

PR #1500, opened 2019-11-11
Support getting and setting window position

einarf:
Getting and setting window position is fairly straight forward in SDL2.


einarf:
Someone suggested the argument to `set_window_position` should be a tuple, but I'm not entirely sure how important this. Will leave that up to the reviewer for now. I'm fine with both.

einarf:
Should be ready for review.

einarf:
What I got from the discussion in the `#contribute` channel on discord.

`2.0.dev7` will expose the `pygame._sdl2` module. Here you can get access to the 2.1 cython api. `pygame._sdl2.video.Window.from_display_module()` will create a window representation of the display that provides all the new SDL2 methods.

@robertpfeiffer also warned about the following:
> but you'd have to del the window reference when you run set_mode again and it creates a new window and when you use that window to do things that invalidate the display surface (liek changing window size) then it's your own fault when it segfaults

My intention with this commit was to add a simple and stable 2.0-specific option to set the window position for users in the `display` module. Then a lot of people can easily remove their `HWND` hacks etc (for single window user). I have no idea what plans you have for the `display` module in the future, so I'm not entirely sure yet why adding this feature is a bad idea considering `pygame.display.get_window_size()` and new attributes for `display` functions were added in 2.0. (Possibly you want to avoid adding new features in 2.0 and focus more on 1.x compatibility)

If this PR is considered in the future the following needs to be done:

* Function arguments should accept and `int` or a two-component `tuple` behaving just like `Window.position`.
* Missing `#if SDLv2 preprocessor` is pointed out by @robertpfeiffer

Please close this if the PR is indeed still a bad idea and we can move on and focus fully on the new cython `_sdl2` module instead in `2.0.dev7+` 😃

illume:
Hi,

sorry again for how this went. I should have been more clear with plans, and where things can go.
It's very useful functionality, and not a bad idea at all! 
I'm sorry that your first PR here went this way. It's entirely my fault.

Putting display.get_window_size there was a mistake, and should probably be removed. I really want to put the focus onto finishing the 'critical' labeled pygame 2 blocking bugs, so we can move onto new features. I also want to make sure we get the API correct for things, and to give each function and class the attention it deserves.

cheers,

einarf:
All good. I'll move my attention to `pygame._sdl2` and see if there are things to add or improve there.
I know for example that SDL2 are missing buffer swapping with dwmapi for Windows Vista and later to get properly smooth frame rates.

Issue #1471, opened 2019-10-31
docs/reST/tut needs to be classified


rumia0601:
I know there are some tutorials at docs/reST/tut with different writers.
However, every tutorials are putted in single directory. 
I think every files in docs/reST/tut should be divided to sub-directories by writter's ID.
Then it will looks like arranged.

illume:
Good idea.

illume:
One issue we need to think about is older URLs. How do we make sure the old web based URLs are redirected to the new ones? Can the Sphinx tool do this, or do we need to do this ourselves somehow. If Sphinx can't do this, we could add some redirects on the website. But we'd need a mapping of old_url -> new_url.

How do different languages folders fit in here? Is there a Sphinx recommended way to do this? What are other projects doing?

What about ordering by topic instead? From a users perspective it may be nicer.

illume:
Currently docs/reST/tut/ has these files in there.
Note all the image files.
```
CameraIntro.rst
ChimpLineByLine.rst
DisplayModes.rst
ImportInit.rst
MakeGames.rst
MoveIt.rst
PygameIntro.rst
SpriteIntro.rst
SurfarrayIntro-rest
SurfarrayIntro.rst
camera_average.jpg
camera_background.jpg
camera_green.jpg
camera_hsv.jpg
camera_mask.jpg
camera_rgb.jpg
camera_thresh.jpg
camera_thresholded.jpg
camera_yuv.jpg
chimp.py.rst
chimpshot.gif
common.txt
intro_ball.gif
intro_blade.jpg
intro_freedom.jpg
newbieguide.rst
surfarray.png
surfarray_allblack.png
surfarray_flipped.png
surfarray_redimg.png
surfarray_rgbarray.png
surfarray_scaledown.png
surfarray_scaleup.png
surfarray_soften.png
surfarray_striped.png
surfarray_xfade.png
tom_basic.png
tom_event-flowchart.png
tom_formulae.png
tom_games2.rst
tom_games3.rst
tom_games4.rst
tom_games5.rst
tom_games6.rst
tom_radians.png
```

illume:
MyreMylarLa mentioned this extension in the discord chat. https://github.com/sphinx-contrib/redirects

illume:
There is an internationalization guide for sphinx. http://www.sphinx-doc.org/en/master/usage/advanced/intl.html

However this works on translatable strings instead. I don't think it's suitable for whole tutorial documents.

illume:
[...]

PR #1502, opened 2019-11-14
classified tut directory

rumia0601:
When we looked at the pygame GitHub's tutorial directory, we saw a lot of tutorials. However, this directory is not classified because it is a single directory. It was a little bit uncomfortable to see.
 Therefore, we raised the issue that if we divide it into subdirectories by the author's ID, it would look more classified than before. We received your positive response to this issue and we divided subdirectories under the author's name. Then, the subdirectory is divided by game name.
 The only difference between current pygame directory and my requesting directory is just tut directory.


illume:
Cool!

Continuing the discussion in the issue: https://github.com/pygame/pygame/pull/1502

illume:
I think we will do this another way. Researching how in the issue #1471

PR #1512, opened 2019-11-18
My new tutorial for English and Korean

rumia0601:
As I Issued before,
I made tutorial for English and Korean.

Each tutorial is separated by (6+2) sub-tutorial.
Furthermore, two directories named en and ko are added to tut directory.


illume:
Well done :)

PR #1542, opened 2019-12-31
Declare UTF-8 encoding

stefanor:
File contains a ° symbol.

This breaks importing under non-UTF-8 locales on unix.


stefanor:
MacOS Travis failures look *very* unrelated.

PR #1543, opened 2019-12-31
Fix a number of spelling mistakes

stefanor:
Debian has tooling that looks for spelling mistakes in packages, and as a result we've been carrying this patch for a while.

But I can't see any record of us forwarding it.

### DIFF
--- a/docs/reST/ref/mouse.rst
+++ b/docs/reST/ref/mouse.rst
@@ -62,7 +62,7 @@ configured.
    | :sg:`get_pos() -> (x, y)`
 
    Returns the ``X`` and ``Y`` position of the mouse cursor. The position is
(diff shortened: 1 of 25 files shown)


docs/reST/ref/mouse.rst
```diff
@@ -62,7 +62,7 @@ configured.
    | :sg:`get_pos() -> (x, y)`
 
    Returns the ``X`` and ``Y`` position of the mouse cursor. The position is
-   relative the the top-left corner of the display. The cursor position can be
+   relative the top-left corner of the display. The cursor position can be
```
hpnok:
```suggestion
   relative to the top-left corner of the display. The cursor position can be
```

PR #1545, opened 2019-12-31
Skip portmidi on Debian GNU/kFreeBSD and GNU/Hurd too

stefanor:
Where on GNU/kFreeBSD:
```
>>> platform.system()
'GNU/kFreeBSD'
```
And on GNU/Hurd:
```
>>> platform.system()
'GNU'
```

Debian has been carrying a hacky patch to do this, but now that you're looking for FreeBSD, upstream, we can handle this better.


stefanor:
> I guess that's a freebsd + gnu debian patch to python.

FreeBSD kernel + GNU Userland

PR #1552, opened 2020-01-08
Code refactoring of draw.c

MightyJosip:
With the recent changes to the draw.c module a lot of old code is not necessary any more. With this update I deleted some old functions, removed old variables used to store bounding rects (now it is stored in list modified during drawing algorithm), changed how aaline is being drawn (end result is still the same, but the code is a lot smaller), fixed the naming of some variables and functions...

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -33,82 +33,41 @@
 
 #include <float.h>
 
-/*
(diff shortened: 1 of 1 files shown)


illume:
I'm testing it manually on Mac, because travis CI mac builds are currently broken.

Issue #728, opened 2019-01-09
pylint issues


illume:
pylint does not seem to find some constants inside pygame.

```python
stuntcat/gifmaker.py:140:29: E1101: Module 'pygame' has no 'KEYDOWN' member (no-member)
stuntcat/gifmaker.py:141:32: E1101: Module 'pygame' has no 'K_g' member (no-member)
stuntcat/gifmaker.py:145:34: E1101: Module 'pygame' has no 'K_g' member (no-member)
```
To fix this:

- perhaps the constants should be explicitly imported?
- perhaps __all__ should be used?

To consider: 
- different constants depending on some defines (at least SDL2 constants are more than SDL1 based pygame).

notpygame:
Related: https://github.com/PyCQA/pylint/issues/2144

MightyJosip:
Can someone check this with the most recent pygame version. Maybe adding type hints fixed this

MyreMylar:
I cleaned up the stuntcat pylint output to make it a little easier to see what was going on without disabling everything. That's in a PR here: 

https://github.com/pygame/stuntcat/pull/64

I also experimented with adding `__all__` to the pygame packages __init__.py, It does work, but it is a bit of a pain. You either have to re-import the public interface parts of pygame at the bottom because they all get deleted, or mess with the delicate mess of imports in the file.

It also seems like pylint doesn't actually care that much about `__all__` when c modules are involved anyway.

If you run with `--extension-pkg-whitelist=pygame` on then most of the remaining errors that still appear vanish. This is the output from `pylint --disable=fixme --extension-pkg-whitelist=pygame stuntcat` with the PR above.

    ************* Module stuntcat.cli
    stuntcat\cli.py:22:4: W0603: Using the global statement (global-statement)
    stuntcat\cli.py:26:8: C0415: Import outside toplevel (stuntcat) (import-outside-toplevel)
    stuntcat\cli.py:30:40: E1101: Instance of 'list' has no '_path' member (no-member)
    stuntcat\cli.py:30:40: W0212: Access to a protected member _path of a client class (protected-access)
    stuntcat\cli.py:57:12: C0415: Import outside toplevel (main) (import-outside-toplevel)
    stuntcat\cli.py:59:12: C0415: Import outside toplevel (stuntcat.main) (import-outside-toplevel)
    stuntcat\cli.py:60:8: C0415: Import outside toplevel (pygame) (import-outside-toplevel)
[...]

PR #1553, opened 2020-01-08
Remove unused imports

VilhelmPrytz:
See these LGTM alerts for reference. I used flake8 to verify that the imports weren't actually used.

https://lgtm.com/projects/g/pygame/pygame/snapshot/5ee0d3bf2ef4464a1031b89f9222e1c345779211/files/src_py/surfarray.py#xba4bec973938d53b:1
https://lgtm.com/projects/g/pygame/pygame/snapshot/5ee0d3bf2ef4464a1031b89f9222e1c345779211/files/src_py/freetype.py#xb1e899caa6e2e8ac:1
https://lgtm.com/projects/g/pygame/pygame/snapshot/5ee0d3bf2ef4464a1031b89f9222e1c345779211/files/src_py/ftfont.py#xa5121d4658b3d854:1

### DIFF
--- a/src_py/freetype.py
+++ b/src_py/freetype.py
@@ -1,15 +1,7 @@
 """Enhanced Pygame module for loading and rendering computer fonts"""
 
-from pygame._freetype import (
-   Font,
-   STYLE_NORMAL, STYLE_OBLIQUE, STYLE_STRONG, STYLE_UNDERLINE, STYLE_WIDE,
-   STYLE_DEFAULT,
-   init, quit, get_init,
-   was_init, get_cache_size, get_default_font, get_default_resolution,
-   get_error, get_version, set_default_resolution,
-   _PYGAME_C_API, __PYGAMEinit__,
-   )
-from pygame.sysfont import match_font, get_fonts, SysFont as _SysFont
+from pygame._freetype import Font
+from pygame.sysfont import SysFont as _SysFont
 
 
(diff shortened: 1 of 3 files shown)


illume:
Hi. Thanks for this.

Some of these are needed because they expose these functions to the user. The `src_py/freetype.py` ones are needed for this reason. So people can import `pygame.freetype` and then use `pygame.freetype.init`.

We need a list of things exposed in an `__all__` attribute.
> `__all__` is a list of public objects of that module, as interpreted by import *. It overrides the default of hiding everything that begins with an underscore.
See https://docs.python.org/3/tutorial/modules.html#importing-from-a-package for more details about `__all__`. I think this should keep LGTM happy.

Someone would need to look at each of the objects and see which ones need to be exposed and added to `__all__`.

If you would like to do this for these three files, please let me know?

VilhelmPrytz:
Thanks, I'll see if this is something I can do! 👍

VilhelmPrytz:
Actually, I don't think I have the time to do this now. I can remove the changes that are incorrect, which one should I remove? Or maybe you can edit my pull request, not too sure how this works.

illume:
Hi. I guess if someone has time to test this, they can figure out which of the changes work.
No one seems to have time for this though... so I'll close it.

I'll link this PR to the issue https://github.com/pygame/pygame/issues/728 in case someone wants to come back to this later.

PR #1566, opened 2020-01-21
Fix Python syntax warning

charlesej:
Overview of changes:
- Fixed the following warning (raised when using Python 3.8 to run setup install: `python setup.py install`)
  - SyntaxWarning: "is not" with a literal. Did you mean "!="?

System details:
- os: windows 10 (64bit)
- python: 3.8.1 (64bit) and 2.7.10 (64bit)
- pygame: 2.0.0.dev7 (SDL: 2.0.10) at 2ff089bfcf7fffc7f8a566a3cc3d0d1fb76fe9e1

### DIFF
--- a/test/test_utils/__init__.py
+++ b/test/test_utils/__init__.py
@@ -179,7 +179,7 @@ def rect_outer_bounds(rect):
    O      O
 
     """
-    return (rect.left != 0 and [(rect.left - 1, rect.top)] or []) + [
+    return ([(rect.left - 1, rect.top)] if rect.left else []) + [
         rect.topright,
         rect.bottomleft,
         rect.bottomright,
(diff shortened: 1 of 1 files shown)


McSinyx:
While `foo is not 0` is implementation dependent (I'm pretty sure there is only one 0 in CPython), `foo != 0` has different meaning which check for inequality.  Consider
```python
>>> False is not 0
True
>>> False != 0
False
```

For the later, simply `foo` is enough.  In this case, if you look carefully, the statement should be rephrased as 
```python
return ([(rect.left - 1, rect.top)] if rect.left else []) + [
    rect.topright,
    rect.bottomleft,
    rect.bottomright,
]
```

There are a lot of exiting code using `predicate and consequent or alternative`, which should be rewritten as `consequent if predicate else alternative` according to PEP 8.

charlesej:
Thanks for the code inspection. I have made the requested change.

McSinyx:
> I have made the requested change.

Wait that's just my opinion, I'm no maintainer of pygame or anything.  Happy Lunar New Year BTW!

Issue #1238, opened 2019-08-10
SDL2: pygame.event.post() does not post some event types


charlesej:
The following code shows that some event types are not appearing in the event queue after being posted using `pygame.event.post()`. Running the code and pressing 'q' creates and posts 2 different event types (`KEYDOWN` and `USEREVENT`).
- Using pygame 1.9.4, both events appear in the event queue after being posted.
- Using pygame 2.0.0.dev3, the `USEREVENT` event is posted and appears in the event queue, but the `KEYDOWN` event does not appear in the event queue when posted.

Note: Uncovered this issue while investigating #1236.

---

```python
import pygame


def add_event(event_type):
    new_event = pygame.event.Event(event_type, unicode='a', key=ord('a'))
    print('Adding event:', new_event)
    pygame.event.post(new_event)
    print('Peeking:', pygame.event.peek(event_type))


pygame.init()

screen = pygame.display.set_mode((400, 200))
pygame.display.flip()
clock = pygame.time.Clock()
running = True

while running:
    clock.tick(20)

    for e in pygame.event.get():
        if e.type == pygame.QUIT:
            running = False
        elif e.type == pygame.KEYDOWN:
            print('Handling event:', e)

            if e.key == pygame.K_ESCAPE:
                running = False
            elif e.key == pygame.K_q:
                # Pressing 'q' adds KEYDOWN and USEREVENT events.
                add_event(pygame.KEYDOWN)
                add_event(pygame.USEREVENT)
        elif e.type == pygame.USEREVENT:
            print('Handling event:', e)

pygame.quit()
```

---

Output using pygame 1.9.4. Both events appear in the event queue after being posted.
```
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
Handling event: <Event(2-KeyDown {'unicode': 'q', 'key': 113, 'mod': 0, 'scancode': 16})>
Adding event: <Event(2-KeyDown {'unicode': 'a', 'key': 97})>
Peeking: 1
Adding event: <Event(24-UserEvent {'unicode': 'a', 'key': 97})>
Peeking: 1
Handling event: <Event(2-KeyDown {'unicode': 'a', 'key': 97})>
Handling event: <Event(24-UserEvent {'unicode': 'a', 'key': 97})>
```

---

[...]

PR #1584, opened 2020-02-19
post keydown events, closes #1580

### LINKED ISSUE #1580: Regression: Cannot post keydown events (or any non-USEREVENT events)
This used to work with Pygame 1.9.6 (I tested it):

```python
import pygame
pygame.init()
screen=pygame.display.set_mode((50,50))
pygame.event.post(
            pygame.event.Event(pygame.KEYDOWN,
                           key=pygame.K_a))
pygame.event.get()
pygame.event.post(
        pygame.event.Event(pygame.KEYDOWN,
                           key=pygame.K_a,
                           unicode='',
                           mod=0,
                           scancode=4,
                           window=None))
pygame.event.get()
```

I.E. you could post arbitrary keydown events.

I relied on this behaviour in my tutorial here: https://blubberquark.tumblr.com/post/183424536830/pygame-and-asyncio-part-v-twitch-integration. TL;DR This tutorial explains how to interleave the `asyncio` event loop with a pygame game loop, so you can post `KEYDOWN` events from asyncio, so people can type commands in IRC and they show up in the game as key presses.

I tried to follow my own tutorial, and found out it doesn't work any more.

I looked at the code in event.c, and it seems to only create and populate USEREVENT structs, and fill their dicts with the supplied data. We probably need to check for known SDL event types and create the appropriate struct for SDL_KEYDOWN or whatever it's called instead. That is going to be a lot of code duplication if we do all known event types, but especially for KEY(DOWN|UP) events, it might be worth the effort to provide an easy way to retrofit touch controls onto existing games.

robertpfeiffer:
This obviously (if you look at the code) only covers *my own* use case, but it should serve as a rough sketch for #1238. The other event types should be analogous.

For #1580


src_c/event.c
```diff
@@ -1667,9 +1667,18 @@ pg_event_post(PyObject *self, PyObject *args)
...
 
-    if (pgEvent_FillUserEvent(e, &event))
-        return NULL;
+    if (e->type == SDL_KEYDOWN || e->type == SDL_KEYUP){
+        event.type =  e->type;
+        event.key.keysym.sym = PyLong_AsLong(PyDict_GetItemString(e->dict, "key"));
+        if (PyDict_GetItemString(e->dict, "scancode") != NULL)
+            event.key.keysym.scancode = PyLong_AsLong(PyDict_GetItemString(e->dict, "scancode"));
+        if (PyDict_GetItemString(e->dict, "mod") != NULL)
+            event.key.keysym.mod = PyLong_AsLong(PyDict_GetItemString(e->dict, "mod"));
```
charlesej: This line is generating the warning:
`warning C4244: '=': conversion from 'long' to 'Uint16', possible loss of data`

src_c/event.c
```diff
@@ -1667,9 +1667,18 @@ pg_event_post(PyObject *self, PyObject *args)
         Py_RETURN_NONE;
     }
 
-    if (pgEvent_FillUserEvent(e, &event))
-        return NULL;
+    if (e->type == SDL_KEYDOWN || e->type == SDL_KEYUP){
+        event.type =  e->type;
+        event.key.keysym.sym = PyLong_AsLong(PyDict_GetItemString(e->dict, "key"));
+        if (PyDict_GetItemString(e->dict, "scancode") != NULL)
+            event.key.keysym.scancode = PyLong_AsLong(PyDict_GetItemString(e->dict, "scancode"));
```
charlesej: You may want to also add the `window` field for `SDL_KEYDOWN `and `SDL_KEYUP `events
and the `unicode `field for `SDL_KEYDOWN `events.

robertpfeiffer: unicode is not a SDL2 thing, but a pygame thing!
https://wiki.libsdl.org/SDL_Keysym
https://wiki.libsdl.org/SDL_KeyboardEvent

src_c/event.c
robertpfeiffer: What should I do here? Cast explicitly or do overflow check? Should I throw an exception if the user puts in too big numbers here?

robertpfeiffer:
I'll look into it some more, but I disagree with at leat one of the tests.

`pygame.event.post(pygame.event.Event(pygame.KEYDOWN))` **should** fail

src_c/event.c
charlesej: A cast will deal with the warning. As for what to do if the value is too big, that might be something @illume can answer. I would suggest as a minimum that a note be made in the event post() documentation that key up/down events only support 'mod' values <= 0xFFFF.

illume:
Hi,

I guess raising an error would be nice?

Otherwise just leave a comment in there for later...?
TODO FIXME: this should raise an error
Key up/down events only support 'mod' values <= 0xFFFF and >= 0.

robertpfeiffer:
I added error handling. Now the tests are broken because they post invalid events.

illume:
What's the status of this one?

robertpfeiffer:
@illume The status is that I won't be fixing this for _all_ event types, because 1) that would be a lot of typing and 2) the bug is somewhere else. There is code in event.c already to mangle USEREVENTs into the right type with an event filter, but it's not loaded properly, except when you run the test suite. Forcing keydown events to have all the propertied of SDL2 keydown events ironically breaks the test cases about posting events based on empty python dicts.

If you think a stop-gap is better than nothing, I can clean this up a bit. After all, I use pygame.event.post in my twitch integration tutorial and would love it to work with pygame 2.0. Or maybe you see right away why putting
```python
pygame.display.init()
pygame.display.quit()
pygame.init()
``` at the beginning works as a workaround. I don't. Maybe it's a corner case in SDL. I just didn't finish this because I thought "there has to be a better way" but then there wasn't.

**Comment by illume** [comment]
> If you think a stop-gap is better than nothing, I can clean this up a bit.

Yeah, thanks. I think a stop-gap is better than nothing.

> Forcing keydown events to have all the propertied of SDL2 keydown events ironically breaks the test cases about posting events based on empty python dicts.

I guess if posting events worked before without filling in all the fields we should still allow that?

Here's a checklist for remaining stuff, feel free to edit it or ask me to do some bits.

- [ ] Moving this code into a function?
- [ ] a bit of doc
- [ ] event_test.test_get__event_sequence is failing on appveyor Windows.

```
FAIL: test_get__event_sequence (pygame.tests.event_test.EventModuleTest)
Ensure get() can handle a sequence of event types.
  File "C:\Python27\lib\site-packages\pygame\tests\event_test.py", line 366, in test_get__event_sequence
    self._assertCountEqual(retrieved_events, expected_events)
  File "C:\Python27\lib\site-packages\pygame\tests\event_test.py", line 208, in _assertCountEqual
    self.assertItemsEqual(*args, **kwargs)
AssertionError: Element counts were not equal:
First has 1, Second has 0:  <Event(768-KeyDown {'scancode': 1964920650, 'window': None, 'key': 32, 'unicode': '', 'mod': 8096})>
First has 0, Second has 1:  <Event(768-KeyDown {'key': 32})>
Ran 36 tests in 0.040s
FAILED (failures=1, skipped=1)
loading pygame.tests.fastevent_test
```

**Comment by robertpfeiffer** [comment]
Has anybody had any luck fixing the root cause in event.c? Otherwise we need to generate proper SDL events instead of Python-side event dicts.

**Comment by robertpfeiffer** [comment]
I think I made it work. What do the other two mean? documenting that event.post may or may not fail if you post SDL events without the right fields?

**Comment by robertpfeiffer** [comment]
The docs are a bit vague on purpose. I don't want to rigorously describe edge case behaviour, because I don't want to support this stopgap indefinitely.

**Comment by robertpfeiffer** [comment]
I don't see a way to trigger another build...

File: src_c/event.c [hunk 3]
```diff
@@ -1746,9 +1746,61 @@ pg_event_post(PyObject *self, PyObject *args)
...
+#endif /* IS_SDLv2 */
+    }
+    else if (e->type >= PGE_USEREVENT && e->type < PG_NUMEVENTS) {
+        if (pgEvent_FillUserEvent(e, &event))
+            return NULL;
+    }
+    else {
+         if (pgEvent_FillUserEvent(e, &event))
+            return NULL;
+    }
```
MyreMylar, line 1803: The `else if` here has no ultimate effect right now, is this intentional?

MyreMylar (changes requested):
Overall this seems fine, little bit of _seemingly_ unneeded code in there, but it doesn't seem to break anything and it fills in more key data when key events are manually posted to the queue with out all their data. At least so long as you do the `display.quit()` and `display.init()` dance beforehand.

Tested it with this:

```
import pygame

pygame.display.init()  # This is still weird   
pygame.display.quit()  # This is still weird   
pygame.init()

pygame.event.clear()

new_event = pygame.event.Event(pygame.KEYDOWN, key=pygame.K_a)
pygame.event.post(new_event)
print(new_event)
print(pygame.event.poll())
```

Output on dev10:
```
<Event(768-KeyDown {'key': 97})>
<Event(768-KeyDown {'key': 97})>
```

Output with this PR:

```
<Event(768-KeyDown {'key': 97})>
<Event(768-KeyDown {'unicode': '', 'key': 97, 'mod': 33824, 'scancode': 59098328, 'window': None})>
```

I think that was the eventual goal of this PR after all the back and forth? I'll approve once the above is resolved one way or the other.

robertpfeiffer:
@MyreMylar 

You shouldn't need any additional inits and quits to post KEYDOWN events now. Post just uses the SDL2 event queue with the proper event type for keydown events now.

illume:
LGTM
```
14:04:28-rene~/dev/pygame/pygame (robertpfeiffer-postkeydownevents)$ python3 postevent_test.py 
pygame 2.0.0.dev11 (SDL 2.0.12, python 3.8.2)
<Event(768-KeyDown {'key': 97})>
<Event(768-KeyDown {'unicode': '', 'key': 97, 'mod': 45200, 'scancode': 107738736, 'window': None})>

14:04:37-rene~/dev/pygame/pygame (robertpfeiffer-postkeydownevents)$ python postevent_test.py 
pygame 2.0.0.dev8 (SDL 2.0.12, python 2.7.16)
<Event(768-KeyDown {'key': 97})>
<Event(0-NoEvent {})>
```

docs/reST/ref/event.rst
```diff
@@ -346,6 +346,14 @@ Events reserved for :mod:`pygame.midi` use.
 
    If the event queue is full a :exc:`pygame.error` is raised.
 
+   Caution: In pygame 2.0, calling this function with event types defined by
+   pygame (such as ``pygame.KEYDOWN``) may put events into the SDL2 event queue.
+   In this case, an error may be raised if standard attributes of that event
+   are missing or have incompatible values, and unexpected properties may
+   be silently omitted. In order to avoid this behaviour, custom event
+   properties should be used with custom event types.
+   This behaviour is not guaranteed.
```
illume, line 355: This line is confusing to me. What behavior is not guaranteed?

illume (approved):
👍

Issue #1595, opened 2020-03-05
Sound is absent in examples.aliens because of bug introduced by refactoring


yzelensky:
`pg.mixer` unconditionaly assigned to `None` after initialisation that effectively disables sound in examples.aliens.
Looks like accident consequence of refactoring done by @illume in the middle of October 2019.

https://github.com/pygame/pygame/blob/d13a044744a05b775ec26bbf8bead771583a6309/examples/aliens.py#L248

This is a tiny issue, but it is quite annoying in particular for newcomers with OSX Catalina, because they have to install the newest version to make the video working, and after doing so they faced with not working sound (as they can think based on their attempt to run aliens example).

https://www.reddit.com/r/Python/comments/du8xuc/pygame_under_macos_catalina/

PR #1598, opened 2020-03-11
Updating sound issues for Aliens.py

Spl99:
On line 247, I've commented out the issue. It is preventing people from hearing sounds from this game. Comes from issue #1595

### DIFF
--- a/examples/aliens.py
+++ b/examples/aliens.py
@@ -244,7 +244,7 @@ def main(winstyle=0):
     pg.init()
     if pg.mixer and not pg.mixer.get_init():
         print("Warning, no sound")
-        pg.mixer = None
+        # pg.mixer = None
     pg.mixer = None
 
     fullscreen = False


examples/aliens.py
```diff
@@ -244,7 +244,7 @@ def main(winstyle=0):
     pg.init()
     if pg.mixer and not pg.mixer.get_init():
         print("Warning, no sound")
-        pg.mixer = None
+        # pg.mixer = None
     pg.mixer = None
```
hpnok: How can the outcome be different if `mixer` is still assigned to `None` when leaving the if clause?

illume, line 248: Thanks for finding this. 

However, I think you commented out the wrong line. Instead I think line 248 should be removed.

illume:
Closing this one. Because I need to get a fix for this before we do our 2.0.0.dev8 release.

Issue #1333, opened 2019-09-24
cppchecker


illume:
> Cppcheck is an analysis tool for C/C++ code. It provides unique code analysis to detect bugs and focuses on detecting undefined behaviour and dangerous coding constructs. The goal is to detect only real errors in the code (i.e. have very few false positives).

> You will find more bugs in your software by testing your software carefully, thanby using Cppcheck.

#### To Install cppcheck

http://cppcheck.sourceforge.net/ and https://github.com/danmar/cppcheck
The manual can be found here: http://cppcheck.net/manual.pdf

```bash
brew install cppcheck bear
sudo apt-get install cppcheck bear
```

### To run cppcheck on C code.

You can use [bear](https://github.com/rizsotto/Bear) (the build ear) tool to record a compilation database (compile_commands.json). cppcheck can then know what c files and header files you are using.

```bash
# call your build tool, like `bear make` to record. 
# See cppcheck manual for other C environments including Visual Studio.
bear python setup.py build
cppcheck --quiet --language=c --enable=all -D__x86_64__ -D__LP64__ --project=compile_commands.json
```

It does seem to find some errors, and style improvements that other tools do not suggest. Note that you can control the level of issues found to errors, to portability and style issues plus more. See `cppcheck --help` and the manual for more details about `--enable` options.

For example these ones from the pygame code base:
```
[src_c/math.c:1134]: (style) The function 'vector_getw' is never used.
[src_c/base.c:1309]: (error) Pointer addition with NULL pointer.
[src_c/scrap_qnx.c:109]: (portability) Assigning a pointer to an integer is not portable.
[src_c/surface.c:832] -> [src_c/surface.c:819]: (warning) Either the condition '!surf' is redundant or there is possible null pointer dereference: surf.
```

cppcheck reports 942 things in the pygame codebase. (633 without cython related things).

illume:
Here's the log from cppcheck.

[cppcheck.txt](https://github.com/pygame/pygame/files/3666052/cppcheck.txt)

ankith26:
Okay, I experimented a bit with cppchecker, and indeed it's quite a mature and useful piece of software, in my warnings on CI PR, I'm adding a CI check for running cppchecker. It also helped catch a memleak, and the best part is that it runs on every macro codepath!

PR #1603, opened 2020-03-22
mixer.c: Fix leak on out of memory with realloc

nthykier:
There is a common, but faulty code pattern, involving realloc of the
form:

    x = realloc(x, ...);

The bug occurs if realloc fails to allocate memory (usually in out of
memory situations).  In this case, realloc will return NULL *without*
free'ing the input leaking to a leak.

In the concrete situation, the code did not appear to handle realloc
failures at all and would immediately attempt to dereference the NULL
pointer.

Found with cppcheck.

### DIFF
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -1348,8 +1348,14 @@ set_num_channels(PyObject *self, PyObject *args)
 
     MIXER_INIT_CHECK();
     if (numchans > numchanneldata) {
+        struct ChannelData *cd_org = channeldata;
         channeldata = (struct ChannelData *)realloc(
             channeldata, sizeof(struct ChannelData) * numchans);
+        if (!channeldata) {
+            /* Restore the original to avoid leaking it */
+            channeldata = cd_org;
+            return PyErr_NoMemory();
+        }
         for (i = numchanneldata; i < numchans; ++i) {
             channeldata[i].sound = NULL;
             channeldata[i].queue = NULL;


nthykier:
Note to the reviewer: Please review the out of memory return extra carefully.  The Python <-> C interface is not my strong suit.

src_c/mixer.c
```diff
@@ -1348,8 +1348,15 @@ set_num_channels(PyObject *self, PyObject *args)
 
     MIXER_INIT_CHECK();
     if (numchans > numchanneldata) {
+        struct ChannelData *cd_org = channeldata;
         channeldata = (struct ChannelData *)realloc(
             channeldata, sizeof(struct ChannelData) * numchans);
+        if (!channeldata) {
+            /* Restore the original to avoid leaking it */
+            channeldata = cd_org;
+            PyErr_NoMemory();
```
illume, line 1357: This could be just `return PyErr_NoMemory();`

nthykier: Thanks for the suggestion; I have updated the branch to implement it.

nthykier:
@illume / @dlon: Thanks for the review.  :)

Could you help me understand the process from here?  As far as I can tell, this pull request and several other active pull requests look good to go at a glance but are not merged. The only "wart" on them are appear to be caused by #1606 (all Mac OSX builds are failing for unrelated reasons).  If #1606 is stalling the merging of all of these pull requests, would you be okay with a pull request to disable the Mac OSX builds? (alternatively, mark them as "ignore-failure")

nthykier:
(For reference, I have opened pre-emptively #1625 to "solve" #1606 to avoid a round trip).

illume:
Hi. Usually a couple of people review pull requests. If the changes are uncontroversial, there are tests, docs, and the PRs are small... I usually merge them in right away. But I'm catching up on the pull requests after not being able to attend to them for some months. I like to go through them all before merging in case there may be some commits that conflict with each other.

With MacOS CI being broken, testing takes a bit longer since I have to do it manually for MacOS.

illume:
Related issue: https://github.com/pygame/pygame/issues/1333

PR #1604, opened 2020-03-22
surface.c: Replace a redundant conditional with an assert

nthykier:
The conditional is confusing cppcheck into believing that
e.g. srcobject might be NULL assuming "itemlength < 2".  This is not
possible due to a conditional a bit earlier, but cppcheck does not
understand that.


illume:
Related https://github.com/pygame/pygame/issues/1333

Issue #1332, opened 2019-09-24
PyList_Append not checked for failure in some places


illume:
There's a few spots where the return code of PyList_Append is not checked in some places.

This isn't likely to affect much real world code, but is probably something that should be fixed at some point.

illume:
@nthykier  fixed a lot of these ones, except...

> About the two missing cases that I left:
>    - sdlmain_osx.m: Looks like Objective-C. As I am never good at objective-C nor have the ability to test it, I punted on that.
>    - surf_blits in surface.c: The error handling this function was too complex that I was sure I could do it right. Notably, I think there is a bug where ret is leaked on error (via goto bliterror;). I am leaving that to someone who knows/understands the code a bit better.

ankith26:
It appears that the surf_blits one is already fixed by nthykier 
That means that only the Objective-C one is left

PR #1605, opened 2020-03-22
Check PyList_Append for errors

nthykier:
This commit ensures that PyList_Append is consistently checked for
errors except two cases.


nthykier:
About the two missing cases that I left:

 * sdlmain_osx.m: Looks like Objective-C.  As I am never good at objective-C nor have the ability to test it, I punted on that.
 * surf_blits in surface.c: The error handling this function was too complex that I was sure I could do it right.  Notably, I think there is a bug where `ret` is leaked on error (via `goto bliterror;`).  I am leaving that to someone who knows/understands the code a bit better.

This was done in order to solve #1332

Issue #1606, opened 2020-03-22
Travis CI fails due to fatal: Authentication failed for 'https://github.com/Homebrew/homebrew-versions/'


nthykier:
```
$ if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then source buildconfig/ci/travis/.travis_osx_before_install.sh;  fi

remote: Invalid username or password.

fatal: Authentication failed for 'https://github.com/Homebrew/homebrew-versions/'

Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-versions failed!

/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby: warning: unknown argument for --enable: `frozen-string-literal'

/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby: warning: unknown argument for --disable: `did_you_mean'

-e:1:in `<main>': undefined method `canonical_segments' for #<Gem::Version "2.0.0"> (NoMethodError)

==> Downloading https://homebrew.bintray.com/bottles-portable-ruby/portable-ruby-2.6.3.mavericks.bottle.tar.gz
```

nthykier:
The OSX builds have been removed, so I guess this is closable now. :)

PR #1625, opened 2020-04-10
.travis.yml: Mark all OSX builds as "allowed_failures"

nthykier:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,41 +43,43 @@ jobs:
 
     # Should only need one OSX build because older builds can work on newer machines ok.
     #https://docs.travis-ci.com/user/osx-ci-environment/#OS-X-Version
-    - os: osx
-      osx_image: xcode7.3
-      language: generic
-      env:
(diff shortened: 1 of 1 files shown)


nthykier:
This "fixes" #1606 by flagging the Mac OSX as allowed to fail.  It is less optimal than fixing the Mac OSX builds. However, as I do not know how to do that, I settled with this alternative.  This will enable us to see if pull requests and commits regress on the remaining platforms.

(Closes #1606)

illume:
Good idea.

Would you mind commenting out all the `- os: osx` instead? Since the mac builds are completely broken, we should probably disable them until they are fixed -- to save wasted build time as well.

nthykier:
René Dudfield:
> Good idea.
> 
> Would you mind commenting out all the `- os: osx` instead? Since the mac builds are completely broken, we should probably disable them until they are fixed -- to save wasted build time as well.
> 

Sure. :)  I have replaced the commit to do that.

PR #1626, opened 2020-04-10
Issue templates

robertpfeiffer:
I propose using github issue templates. Maybe not just like this, maybe much more involved, split into bugs, features, and pull request, but I thought I'd start here. I don't want to have to ask people what version of pygame and python they are on, only to find out a bug can only be reproduced with PyGame 1.9.6 on a 32-bit Debian with Python 3.8 on a RaspberryPi.

### DIFF
--- a/docs/issue_template.md
+++ b/docs/issue_template.md
@@ -1,0 +1,30 @@
+<!-- This issue tracker is for bug reports and feature requests in PyGame. -->
+<!-- If you have a simple question about PyGame, please ask in the #help channel on Discord at https://discord.gg/EKfEbTd -->
+
+**Environment:**
+<!-- You can get most of this info from the interactive python prompt -->
+<!-- And from the output of "import pygame" -->
+
+- operating system:
+- Python version:
+- SDL version:
+- PyGame version:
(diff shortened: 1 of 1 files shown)


MyreMylar:
I tried restarting the travis build but it just produces the same nonsensical error about being unable to find a buildconfig .sh that is clearly there.

Perhaps making a small change to the issue template file will cause Travis to sort itself out again?

Maybe add a capital 'O' on Operating system or something.

MyreMylar:
It looks like this PR is trying to run old mac/OSX build configurations on Travis that no longer exist which is probably why it doesn't work. Happy to just grab this file out of this PR and try it in a new PR instead, I think this could be useful in prompting people reporting issues to give us a bit more detail.

MyreMylar:
Tried moving the bones of this over to a new PR here: https://github.com/pygame/pygame/pull/2073 to a) get it past Travis and b) move it to the latest style/folder structure GitHub has moved to.

MyreMylar:
May as well close this as there is more discussion of it in the new PR linked above.

PR #1628, opened 2020-04-11
Update SDL_ttf version

### LINKED ISSUE #1413: pygame.Font rendered, anti-aliased text with transparent background looks worse/faded in pygame 2
A subtle issue I've been noticing, particularly when working with small font sizes. Something in the way anti-aliased text with a transparent background is rendered has changed between pygame 1.9.6 and pygame 2, and not for the better in my opinion. This is visible on Windows, no idea if it is also present on other platforms at the moment.

I've personally seen the issue in dev3 and dev4 and can investigate different versions further if needed.

Here's how some text looks under pygame 1.9.6:
pygame_1_9

And here's the same text under pygame 2 dev4:
pygame_2

The font used here is Fira Code, but I don't think that particularly matters. 

If you render text with a background colour set then everything looks fine/consistent across versions, but that is obviously not ideal in all cases as text background colours have no alpha at all. Even if they did, in the example above when I want to stick text on to a semi-transparent background the text background would look odd blitted onto an already semi-transparent rectangle.

This issue is more noticeable with dark text on a light background.

MightyJosip:
This should fix #1413 for SDL2 builds. I tested this on Windows, as far as I found out Linux already use 2.0.15, and for the other OS I was unable to find what version do they use, so if can someone else check that I would be grateful.

### DIFF
--- a/buildconfig/download_win_prebuilt.py
+++ b/buildconfig/download_win_prebuilt.py
@@ -233,12 +233,12 @@ def place_downloaded_prebuilts(temp_dir, move_to_dir, x86=True, x64=True, sdl2=T
         copy(
             os.path.join(
                 temp_dir,
-                'SDL2_ttf-devel-2.0.14-VC/SDL2_ttf-2.0.14'
+                'SDL2_ttf-devel-2.0.15-VC/SDL2_ttf-2.0.15'
             ),
             os.path.join(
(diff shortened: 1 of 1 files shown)


MightyJosip:
Reopened this just in case we are unable to fix the code for the most recent SDL version

Issue #1612, opened 2020-03-30
Completion of error handling


elfring:
Would you like to [add more error handling](https://github.com/pygame/pygame/blob/77adb7ed2f851712d39a8dc63745195bda408a97/src_c/mixer.c#L447 "Update candidate: _init()") for return values from a function like [malloc()](https://pubs.opengroup.org/onlinepubs/9699919799/functions/malloc.html "Memory allocation")?

elfring:
How do you think about to improve [static source code analysis](https://en.wikipedia.org/wiki/List_of_tools_for_static_code_analysis#C.2C_C.2B.2B "Tools for code analysis") also for this software?

illume:
Hello @elfring. Yes, that would be nice. Pull requests would be appreciated :)

There are some tools in these issues that we have started to use. Some more than others.
https://github.com/pygame/pygame/issues?q=is%3Aissue+is%3Aopen+label%3Atooling+

MyreMylar:
I think this issue is done.

PR #1632, opened 2020-04-11
Fix some missing malloc error handling and memory-leak on error

nthykier:
Closes: #1612

This is an incremental patch to plug some memory leaks on error and missing error handling of malloc (etc.) returning error. It is not complete; I will be filing some issues some code cases that looks wrong but I cannot quite grok well enough to fix them.


shuttle1987:
Fixes like this are really nice to see, thanks!

elfring:
Thanks for your small source code improvement.

Issue #1322, opened 2019-09-21
Complexity of C code


illume:
GNU complexity can find complex C functions (that probably should be refactored).
http://www.gnu.org/software/complexity/

`brew install complexity`

```bash
complexity src_c/*.c
```

illume:
frama-c is another tool that can be used to find function complexity. https://frama-c.com/metrics.html

illume:
A pygame.image.tostring cleanup PR was started here: https://github.com/pygame/pygame/pull/1637

illume:
I'd like to start making a list of issues for each of the functions that are too complex. Before writing lots of issues, I'd like to explore doing one issue all the way through. I've started writing an issue about display.set_mode complexity #2316

These issues could be used for teaching, probably at an intermediate/advanced level.

We could learn from the image.tostring PR #1637

PR #1637, opened 2020-04-12
WIP: Refactor image_tostring to reduce its "GNU complexity" score

nthykier:
Related to #1322.

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -370,6 +370,14 @@ unpack_color_from_palette(char *data, Uint32 color, SDL_PixelFormat *format) {
     data[2] = (char)format->palette->colors[color].b;
 }
 
+static PG_INLINE Uint32 read_24bit_color(const Uint8 *ptr) {
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+    return ptr[0] | (ptr[1] << 8) | (ptr[2] << 16);
+#else
+    return ptr[2] | (ptr[1] << 8) | (ptr[0] << 16);
+#endif
+}
+
 PyObject *
 image_tostring(PyObject *self, PyObject *arg)
 {
@@ -469,11 +477,7 @@ image_tostring(PyObject *self, PyObject *arg)
                     Uint8 *ptr = (Uint8 *)DATAROW(surf->pixels, h, surf->pitch,
                                                   surf->h, flipped);
                     for (w = 0; w < surf->w; ++w) {
-#if SDL_BYTEORDER == SDL_LIL_ENDIAN
-                        color = ptr[0] + (ptr[1] << 8) + (ptr[2] << 16);
-#else
-                        color = ptr[2] + (ptr[1] << 8) + (ptr[0] << 16);
-#endif
+                        color = read_24bit_color(ptr);
                         ptr += 3;
                         unpack_RGB_color(data, color, surf->format);
                         data += 3;
@@ -548,11 +552,7 @@ image_tostring(PyObject *self, PyObject *arg)
                     Uint8 *ptr = (Uint8 *)DATAROW(surf->pixels, h, surf->pitch,
                                                   surf->h, flipped);
                     for (w = 0; w < surf->w; ++w) {
-#if SDL_BYTEORDER == SDL_LIL_ENDIAN
-                        color = ptr[0] + (ptr[1] << 8) + (ptr[2] << 16);
-#else
-                        color = ptr[2] + (ptr[1] << 8) + (ptr[0] << 16);
-#endif
+                        color = read_24bit_color(ptr);
                         ptr += 3;
                         unpack_RGB_color(data, color, surf->format);
                         data[3] =
@@ -625,11 +625,7 @@ image_tostring(PyObject *self, PyObject *arg)
                     Uint8 *ptr = (Uint8 *)DATAROW(surf->pixels, h, surf->pitch,
                                                   surf->h, flipped);
                     for (w = 0; w < surf->w; ++w) {
-#if SDL_BYTEORDER == SDL_LIL_ENDIAN
-                        color = ptr[0] + (ptr[1] << 8) + (ptr[2] << 16);
-#else
-                        color = ptr[2] + (ptr[1] << 8) + (ptr[0] << 16);
-#endif
+                        color = read_24bit_color(ptr);
                         ptr += 3;
                         unpack_ARGB_color(data, color, surf->format);
                         data += 4;
@@ -683,11 +679,7 @@ image_tostring(PyObject *self, PyObject *arg)
                     Uint8 *ptr = (Uint8 *)DATAROW(surf->pixels, h, surf->pitch,
                                                   surf->h, flipped);
                     for (w = 0; w < surf->w; ++w) {
-#if SDL_BYTEORDER == SDL_LIL_ENDIAN
-                        color = ptr[0] + (ptr[1] << 8) + (ptr[2] << 16);
-#else
-                        color = ptr[2] + (ptr[1] << 8) + (ptr[0] << 16);
-#endif
+                        color = read_24bit_color(ptr);
                         ptr += 3;
                         unpack_RGBA_color_alpha_mult(data, color, surf->format);
                         data += 4;
@@ -741,11 +733,7 @@ image_tostring(PyObject *self, PyObject *arg)
                     Uint8 *ptr = (Uint8 *)DATAROW(surf->pixels, h, surf->pitch,
                                                   surf->h, flipped);
                     for (w = 0; w < surf->w; ++w) {
-#if SDL_BYTEORDER == SDL_LIL_ENDIAN
-                        color = ptr[0] + (ptr[1] << 8) + (ptr[2] << 16);
-#else
-                        color = ptr[2] + (ptr[1] << 8) + (ptr[0] << 16);
-#endif
+                        color = read_24bit_color(ptr);
                         ptr += 3;
                         unpack_ARGB_color_alpha_mult(data, color, surf->format);
                         data += 4;
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -314,6 +314,62 @@ image_get_extended(PyObject *self, PyObject *arg)
     return PyInt_FromLong(GETSTATE(self)->is_extended);
 }
 
+static PG_INLINE void
+unpack_RGB_color(char *data, Uint32 color, SDL_PixelFormat *format) {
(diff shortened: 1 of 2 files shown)


illume:
Very nice.

Some initial thoughts...

- the code is **much** more readable now.
- The test coverage of image.tostring is not very high it seems, which is a bit worrying with a refactor.
- I haven't checked each change to see if it's equivalent yet.
- How about prefixing these with `_` or `_tostring_`? This would show that they are internal functions, and related to image_tostring. Maybe `_tostring_` makes them unreadable though... so marking the start and end of them with a comment could help the future reader?
- Passing function pointers to be used in an inner loop could result in a significant performance difference. I wonder if this is true though? It was true that function pointers are not inline candidates in all the major compilers (and also intels compiler) some years ago at least.

src_c/image.c
```diff
@@ -314,6 +314,129 @@ image_get_extended(PyObject *self, PyObject *arg)
...
+
+static PG_INLINE Uint32 read_24bit_color(const Uint8 *ptr) {
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+    return ptr[0] | (ptr[1] << 8) | (ptr[2] << 16);
+#else
+    return ptr[2] | (ptr[1] << 8) | (ptr[0] << 16);
+#endif
+}
+
+typedef int (*unpack_function)(char *, Uint32, SDL_PixelFormat *);
```
illume, line 384: `int` can have different sizes. It could make sense to set this to a sized type.

illume:
Instead of a function pointer, passing in a constant and switching on that to call the appropriate function should be faster. Branch predictors are pretty good in this case, and the compiler still has a possibility of in-lining the function. But we would need a basic benchmark first to see for sure.

Something like this?

```c
static PG_INLINE void
parse_16bit_colors(SDL_Surface *surf, char *data, int flipped, int unpack_func) {
    int w, h;
    for (h = 0; h < surf->h; ++h) {
        Uint16 *ptr = (Uint16 *)DATAROW(
            surf->pixels, h, surf->pitch, surf->h, flipped);
        for (w = 0; w < surf->w; ++w) {
            int color = *ptr++;
            if (unpack_func == UNPACK_RGB_COLOR) {
                data += unpack_RGB_color(data, color, surf->format);
            } else if (unpack_func == ...) {
                data += ...
            }
        }
    }
}

parse_16bit_colors(surf, data, flipped, UNPACK_RGB_COLOR);
```

src_c/image.c
```diff
@@ -314,6 +314,129 @@ image_get_extended(PyObject *self, PyObject *arg)
...
+
+static PG_INLINE void
+parse_8bit_colors(SDL_Surface *surf, char *data, int flipped, unpack_function unpack_func) {
```
illume, line 387: I'm not sure about `char *data`.

- data could be a different name? `dataout`?
- the `char *` could be `Uint8 *`?

nthykier:
> Very nice.
> 
> Some initial thoughts...
> 
>     * the code is **much** more readable now.
> 

I am glad you like it. :)

>     * The test coverage of image.tostring is not very high it seems, which is a bit worrying with a refactor.
> 

Ugh, that would be an issue indeed. I relied on the existing test to let me know if I broke something.

>  [...]
>     * How about prefixing these with `_` or `_tostring_`? This would show that they are internal functions, and related to image_tostring. Maybe `_tostring_` makes them unreadable though... so marking the start and end of them with a comment could help the future reader?
> 

Sure, happy to do that.

>     * Passing function pointers to be used in an inner loop could result in a significant performance difference. I wonder if this is true though? It was true that function pointers are not inline candidates in all the major compilers (and also intels compiler) some years ago at least.

Ok, I had expected that the compiler would see that there is a "hard-coded" pointer reference passed to an inlinable function (which would allow it to inline both).   Do we have any standard way of benchmarking code like this with a relevant test corpus? I would prefer using existing practise here to ensure that I am testing/benchmarking the right thing.

nthykier:
It appears gcc can inline those now with `-findirect-inlining` (default with `-O2`) according to https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html. Not sure about other reference compilers used by pygame.

illume:
You should be able to paste this into ipython.

```python
>>> import pygame                                                                                                                                                                                  
>>> surf = pygame.Surface((20,20))                                                                                                                                                                 
>>> bigsurf = pygame.Surface((500,500))                                                                                                                                                            
>>> %timeit rgba_buf = pygame.image.tostring(surf, "RGBA")                                                                                                                                         
>>> %timeit rgba_buf = pygame.image.tostring(bigsurf, "RGBA")                                                                                                                                      
```

Timings for my ye olde laptop...
```python
>>> import pygame                                                                                                                                                                                  
>>> surf = pygame.Surface((20,20))                                                                                                                                                                 
>>> bigsurf = pygame.Surface((500,500))                                                                                                                                                            
>>> %timeit rgba_buf = pygame.image.tostring(surf, "RGBA")                                                                                                                                         
2.3 µs ± 49.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
>>> %timeit rgba_buf = pygame.image.tostring(bigsurf, "RGBA")                                                                                                                                      
1.03 ms ± 10.2 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Second run timings of master branch...

```2.19 µs ± 21.9 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
1.03 ms ± 11.9 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

---

Timings with refactor branch...
```2.29 µs ± 18.1 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
1.1 ms ± 15.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Second run with refactor branch...

```2.28 µs ± 12.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
1.09 ms ± 4.77 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

This was the command line it used to compile image.c (note it uses clang):
```bash
clang -Wno-unused-result -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/usr/include -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/System/Library/Frameworks/Tk.framework/Versions/8.5/Headers -Ddarwin -D_THREAD_SAFE -DENABLE_NEWBUF=1 -I/usr/X11R6/include -I/usr/local/include/SDL2 -I/usr/local/include -I/usr/local/opt/openssl@1.1/include -I/usr/local/opt/sqlite/include -I/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/include/python3.7m -c src_c/image.c -o build/temp.macosx-10.13-x86_64-3.7/src_c/image.o

clang --version
Apple LLVM version 10.0.0 (clang-1000.11.45.5)
Target: x86_64-apple-darwin17.7.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

illume:
By adding `-S` to the command, and removing the `-o` part it could generate the assembly into the image.s file.

```bash
clang -S  -Wno-unused-result -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/usr/include -I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk/System/Library/Frameworks/Tk.framework/Versions/8.5/Headers -Ddarwin -D_THREAD_SAFE -DENABLE_NEWBUF=1 -I/usr/X11R6/include -I/usr/local/include/SDL2 -I/usr/local/include -I/usr/local/opt/openssl@1.1/include -I/usr/local/opt/sqlite/include -I/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/include/python3.7m -c src_c/image.c
```

In there it looked to me like `unpack_RGB_color` was not defined as a function, but was inlined.

nthykier:
Thanks for the timing suggestion. :) I have implemented an alternative proposal for solving this (#1643) that will replace this one assuming my alternative proposal is accepted.

nthykier:
> Instead of a function pointer, passing in a constant and switching on that to call the appropriate function should be faster. Branch predictors are pretty good in this case, and the compiler still has a possibility of in-lining the function. But we would need a basic benchmark first to see for sure.
> 
> Something like this?
> 
> ```c
> static PG_INLINE void
> parse_16bit_colors(SDL_Surface *surf, char *data, int flipped, int unpack_func) {
>     int w, h;
> [...]
> }
> 
> parse_16bit_colors(surf, data, flipped, UNPACK_RGB_COLOR);
> ```

I tried this with parse_32bit_colors (admittedly using an enum and a switch-case, but that should not matter performance-wise) and it had the same performance as the function pointers with the current optimization levels (i.e. a loss compared to master or the macro magic I did in #1643).

illume:
Thanks for trying it out.

nthykier:
From #1643:
> Some properties we can test:
> 
>     * output length is `w*h*depth_bytes` (covered in round trip test too)
> 
>     * round trip: `s1 -> tostring -> fromstring -> s2; assert s1 == s2`.
> 
> 
> I guess these formats are also not tested:
> 
>     * 'RGB'
> 
>     * 'RGBX'
> 
>     * 'P'
> 
> 
> Some test methods to add...
> 
> ```
>     def test_tostring_depth_8(self):
>     def test_tostring_depth_16(self):
>     def test_tostring_depth_24(self):
>     def test_tostring_fmt_RGB(self):
>     def test_tostring_fmt_RGBX(self):
>     def test_tostring_fmt_P(self):
> ```

illume:
Closing this to keep our PR list clean. If you want to continue it, please feel free to comment/reopen. I made a comment in the complexity issue about this PR in case someone wants to pick it up at another time.

PR #1639, opened 2020-04-13
SDL2: Ensure that SDL_GetWindowSurface is called on flip (Request for review)

nthykier:
Without this commit, resizing a pygame window under SDL2 immediately crashes with the error below.  I am not entirely sure why the `set_mode` call is not enough with this change either, so I am opening this as a draft pull request for now until can review whether this is the right approach.

```
  File "debugging/resize.py", line 28, in <module>
    pygame.display.flip()
pygame.error: Window surface is invalid, please call SDL_GetWindowSurface() to get a new surface
```

The `debugging/resize.py` has the following content:

```python
import pygame

pygame.init()
window = pygame.display.set_mode((800, 600), pygame.RESIZABLE)

running = True
resize = None
while running:
        if event.type == pygame.QUIT:
            running = False
        elif event.type == pygame.ACTIVEEVENT:
            print(event)
        elif event.type == pygame.VIDEORESIZE:
            resize = event.size
            print(event)
            print(f"width={window.get_width()}, height={window.get_height()}")

    if resize:
        pygame.event.set_blocked(pygame.VIDEORESIZE)
        pygame.event.clear(pygame.VIDEORESIZE)
        window = pygame.display.set_mode(resize, pygame.RESIZABLE)

        resize = None
            
    rect = window.fill((255, 0, 0))
    pygame.display.flip()
    pygame.event.set_allowed(pygame.VIDEORESIZE)
```

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -1924,6 +1924,14 @@ static PyObject *
 pg_GetDefaultWindowSurface(void)
 {
     /*return a borrowed reference*/
+    if (pg_default_window) {
+        /* With SDL2, resizing invalidates the existing surface.  Calling
+         * SDL_GetWindowSurface will recreate the surface when it is
+         * invalid (or return the current surface if it is still valid)
+         */
+        SDL_Surface *new_surface = SDL_GetWindowSurface(pg_default_window);
+        ((pgSurfaceObject *)pg_default_screen)->surf = new_surface;
+    }
     return pg_default_screen;
 }
 
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1396,6 +1396,13 @@ pg_flip_internal(_DisplayState *state)
             SDL_RenderPresent(pg_renderer);
         }
         else {
+#if IS_SDLv2
+            /* Force a re-initialization of the surface in case it
+             * has been resized to avoid "please call SDL_GetWindowSurface"
+             * errors that the programmer cannot fix
+             */
+            (void)pg_GetDefaultWindowSurface();
+#endif
             status = SDL_UpdateWindowSurface(win);
         }
     }


nthykier:
Related: https://forums.libsdl.org/viewtopic.php?t=11027

illume:
This one doesn't crash for me on macos: `pygame 2.0.0.dev7 (SDL 2.0.10, python 3.7.4)`. I tried resize.py and resized the window.

src_c/base.c
```diff
@@ -1924,6 +1924,14 @@ static PyObject *
 pg_GetDefaultWindowSurface(void)
 {
     /*return a borrowed reference*/
+    if (pg_default_window) {
+        /* With SDL2, resizing invalidates the existing surface.  Calling
+         * SDL_GetWindowSurface will recreate the surface when it is
+         * invalid (or return the current surface if it is still valid)
+         */
+        SDL_Surface *new_surface = SDL_GetWindowSurface(pg_default_window);
```
illume, line 1932: SDL_GetWindowSurface can return NULL on failure

nthykier: Ok.  My initial take on that would be to raise an `pgExc_SDLError` here if `SDL_GetWindowSurface` returns NULL.  However, I am not sure how to do that in base.c because the "standard" method for this does not work (`return RAISE(pgExc_SDLError, SDL_GetError());`).

Notably, neither `pgExc_SDLError` nor `PYGAMEAPI_GET_SLOT` is defined here (presumably to avoid breaking `base.c` that end up doing the true definition of `pgExc_SDLError` later)

illume: Hrmm.

Here's the docs for pg_GetDefaultWindowSurface.
```
.. c:function:: PyObject* pg_GetDefaultWindowSurface(void)

   Return a borrowed reference to the Pygame default window display surface,
   or *NULL* if no default window is open.

   Availability: SDL 2.
```

It doesn't talk about errors. I guess all the callers will have to check for the `surf` value of the return being NULL being an error - and then raise an exception? 

```c
PyObject *surface = pg_GetDefaultWindowSurface();
if(surface && !surface->surf) {
    RAISE(pgExc_SDLError, SDL_GetError());
}
```

It's kind of an error prone API though, but would work. What do you think?

illume: I don't think SDL_GetWindowSurface will fail that often. We can move this `check SDL_GetWindowSurface return code` issue to a separate issue report for another time... and merge this one in.

illume: I wrote this up here: https://github.com/pygame/pygame/issues/1671

robertpfeiffer:
I just checked with SDL2.0.8 again, and the behaviour of window resizing is actually *different* now in 2.0.12.

123username:
Could this PR be merged onto the dev version please ?
like pygame v2.0.0dev10 i guess

I had the problem when switching to fullscreen using v2.0.0dev10
thanks !

PR #1643, opened 2020-04-13
image.c: Extract pixel reading loops into macros

nthykier:
With a bit of macro magic, we can collapse the two for-loops
(surrounding the pixel unpacking in image_tostring) into a "simple"
FOREACH_XXBIT_PIXEL statement.  This changes code like:

```c
    for (h = 0; h < surf->h; ++h) {
        Uint32 *ptr = (Uint32 *)DATAROW(
             surf->pixels, h, surf->pitch, surf->h, flipped);
        for (w = 0; w < surf->w; ++w) {
            color = *ptr++;
            ... use color here ...;
        }
    }
```

into:

```c
    FOREACH_32BIT_PIXEL(surf, flipped, color, {
        ... use color here ...;
    });
```

There are 4 variants (one for each of the "bytes per pixel" values
supported).  As an added bonus, we can "hide" the byte-order issue
from 24bit variant in the macro definition.

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -314,13 +314,136 @@ image_get_extended(PyObject *self, PyObject *arg)
     return PyInt_FromLong(GETSTATE(self)->is_extended);
 }
 
+/**
+ * Used by image_tostring to implement the loop over pixels that are
+ * stored in 1 byte (8bit) per pixel.
+ *
+ * It takes the following parameters:
+ *  - surf: The SDL_Surface containing the pixels to loop over
+ *  - flipped: A boolean saying whether we should read the format from
+ *    top to bottom (false) or from bottom to top (true)
+ *  - color_var: The name of a caller defined variable.  This will be
+ *    in each loop iteration to the read color of the current pixel.
+ *    This must be able to handle at least as many bits as the bits
+ *    per pixel (i.e. the XX in FOREACH_XXBIT_PIXEL).  Note that
+ *    Uint32 works fine in all cases.
+ *  - code: A block of code to be executed for each pixel.
+ *
+ * Note that the macro defines some internal (scoped) variables with
+ * the following names: w_, h_, ptr_.  Please avoid using these names
+ * for variables when using the macro to avoid name shadowing or
+ * conflicts.
+ *
+ * The "do { ... } while(0)" is to make this a "statement" requiring
+ * semi-colon (as it looks like one).
+ */
+#define FOREACH_8BIT_PIXEL(surf, flipped, color_var, code) \
+    do { \
+        int w_, h_; \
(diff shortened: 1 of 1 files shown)


nthykier:
This is an alternative solution to #1637 (and would replace that PR if this is accepted).  This variant has two advantages over #1637:

 1. It makes the "same changes" to all the loops (of same pixel size), which reduces the number of chances to introduce bugs (that are not caught by testing).  #1637 had "unique changes" to most variants.
 1. It does not rely on the compiler to handle inlining nor optimizing function pointers due to its use of macros.

Thanks to the latter point, there is no performance regression in this PR compared to #1637.

illume:
Thanks for this. It makes it easier to read. 

This approach (macro approach):
- pro: it's a smaller change.
- pro: It's reduces the code line count (introduces more documentation).
- pro: equivalent speed
- con: macros are not as readable, and machine checkable as functions.

The other approach (function approach):
- pro: it uses functions rather than macros which are easier for people and tools to read.
- pro: further refactoring is easier.
- con: more extensive changes, and without enough tests, there's a higher chance of issues.
- con: slower at the moment. However I feel functions have the potential to beat a macro approach in future refactorings.

Where to go from here?

1. more tests are required to refactor this code safely (for either approach). I suggest we could start on these first.
2. I'm leaning towards your approach with functions, rather than macros. Mainly because I think future refactorings of this approach are easier without macros. I wonder did you try [my suggestion](https://github.com/pygame/pygame/pull/1637#issuecomment-612818946) to see if it's of a similar speed?

nthykier:
To answer your "where to go from here":

1. AFAICT, we are missing tests for `depth != 32`.  Unfortunately, it is not trivial to just loop over the depths and run the existing tests (among other because `tostring` expands the color format to 32bit and the tests assume that the original depth should be used when comparing).  I cannot exactly wrap my head around writing tests to cover the missing parts, so I am stuck there.  The best I can offer is a "round-trip" test a la `s1 -> tostring -> fromstring -> s2; assert s1 == s2`.
2. I have not tested the alternative yet.  I will wait until we have sufficient testing to ensure confidence in the refactoring.

illume:
Some properties we can test:
- output length is w*h*depth_bytes
- round trip: `s1 -> tostring -> fromstring -> s2; assert s1 == s2`.

I guess these formats are also not tested:
- 'RGB'
- 'RGBX'
- 'P'

Some test methods to add...
```
    def test_tostring_depth_8(self):
    def test_tostring_depth_16(self):
    def test_tostring_depth_24(self):
    def test_tostring_fmt_RGB(self):
    def test_tostring_fmt_RGBX(self):
    def test_tostring_fmt_P(self):
```

nthykier:
I will close this leaving #1637 instead.

Issue #1383, opened 2019-10-07
Calling surface's set_shifts() or set_masks() can impact other surfaces


charlesej:
Calling surface's `set_shifts()` or `set_masks()` methods can impact other surfaces.

The following code examples show that when using pygame 1.9.4 other surfaces are not impacted by a surface changing its shifts or masks, but when using pygame 2.0.0.dev3 they are. Note that in the given examples a segmentation fault occurs when the other surfaces are impacted, this doesn't always happen. Sometimes the surfaces continue to function with incorrect format data.

The `set_shifts()` and `set_masks()` functions are currently being used in these files:
- src_py/_camera_vidcapture.py
- test/surface_test.py
- test/surfarray_test.py

Note:
This may be related to #1381. When surfarray_test.py's `test_blit_array()` is changed to not use `set_shifts()` and `set_masks()` (comment out lines 358-380), no test fails are seen if surface_test.py's `test_src_alpha_issue_1289()` is included in the test suite (only seemed to be an issue with Python 2.7).

---

```python
# file: surface_set_shifts.py
import pygame


pygame.init()

surf_a = pygame.Surface((1, 1), 0, 32)
surf_b = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
surf_c = pygame.Surface((1, 1), pygame.SRCALPHA, 32)

surf_a.fill((1, 2, 3))
surf_b.fill((4, 5, 6))
surf_c.fill((7, 8, 9))

print('surf_a.get_at:', surf_a.get_at((0, 0)))
print('surf_b.get_at:', surf_b.get_at((0, 0)))
print('surf_c.get_at:', surf_c.get_at((0, 0)))

print()

print('surf_a.get_shifts:', surf_a.get_shifts())
print('surf_b.get_shifts:', surf_b.get_shifts())
print('surf_c.get_shifts:', surf_c.get_shifts())

print()

# Setting the shifts for surf_c.
c_shifts = [0, 16, 24, 8]
print('surf_c.set_shifts:', c_shifts)
surf_c.set_shifts(c_shifts)

print()

# The shifts for surf_b have also changed.
print('surf_a.get_shifts:', surf_a.get_shifts())
print('surf_b.get_shifts:', surf_b.get_shifts())
print('surf_c.get_shifts:', surf_c.get_shifts())

print()

# Segmentation fault trying to get_at() for surf_b.
print('surf_a.get_at:', surf_a.get_at((0, 0)))
[...]

PR #1644, opened 2020-04-14
Have set_shifts and set_masks raise AttributeError in SDL2

nthykier:
PixelFormat is read-only in SDL2 (because it is shared between
multiple surfaces).  Changing the PixelFormat can trigger seg. faults
when working with other surfaces.

This commit "solves" the problem by having the set_shifts and
set_masks methods raise an AttributeError to inform that programmer
that we cannot change these values any more.

Closes: #1383

### DIFF
--- a/docs/reST/ref/surface.rst
+++ b/docs/reST/ref/surface.rst
@@ -768,6 +768,9 @@
 
       This is not needed for normal pygame usage.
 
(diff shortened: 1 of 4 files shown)


nthykier:
Note that we can also (as an alternative) remove the methods completely.  Let me know if you prefer that - though in that case, should we just remove all references to them (including in the documentation) or should we leave some traces behind (e.g. the docs saying it is going)?

Issue #1181, opened 2019-07-20
SDL2 RLEACCEL segfault


robertpfeiffer:
This might be related to #721 and #742. I extracted this code from my old game Tower Of Babel as I tried to port it to pygame2, and I realised that I might have used RLEACCEL wrong back in 2014, so this might be low-priority.

Anyway, here is a minimal example of the code that triggers the same bug (you need to download the image file from https://blubbervision.neocities.org/tile-indexed.png to trigger the bug.

```python
import pygame
pygame.init()
screen = pygame.display.set_mode((320,240), pygame.SCALED, 24)
pygame.event.pump()
pygame.display.update()
bg=pygame.Surface((320,240), 24)
bg.set_colorkey((255,0,255), pygame.RLEACCEL)
bg.fill((255,0,255))
#https://blubbervision.neocities.org/tile-indexed.png
tilesheet=pygame.image.load("tile-indexed.png")
TILE=16
tilesheet.set_colorkey((255,0,255))

screen.blit(tilesheet,(0,0))
pygame.display.update()
pygame.time.wait(300)

bg.blit(tilesheet, (0,0))
for x in range(20):
  for y in range(15):
    bg.blit(tilesheet.subsurface(pygame.Rect(x*TILE,y*TILE,TILE,TILE)),(x*TILE,y*TILE))
screen.blit(bg,(0,0))

pygame.display.update()
pygame.time.wait(300)
bg.fill((255,0,255)) # comment this out for segfault 
bg.copy()
pygame.display.update()
```

This crashes on my machine with pygame 2.0-dev3. It doesn't of you remove the RLEACCEL flag.

A related issue might be the correct handling of colorkey on images with alpha transparency or of indexed/paletted images where one of the palette colours has alpha transparency.

robertpfeiffer:
I think we forgot to check for SDL_MUSTLOCK somewhere

nthykier:
This has the same crash signature (backtrace) as #1200.  I think we are better off considering the seg. fault from #1181 and #1200 as the same.

That said, the code snippet here with the `bg.fill((255, 0, 255))` line active fails with a big fat:

```
Traceback (most recent call last):
  File "debugging/gh-1181.py", line 29, in <module>
    bg.fill((255,0,255)) # comment this out for segfault 
pygame.error: SDL_FillRect(): You must lock the surface
```
Which also need fixing.

robertpfeiffer:
[...]

PR #1645, opened 2020-04-14
surf_fill: Add missing locking before calling SDL_FillRect

nthykier:
Spotted in #1181.  I am unsure if I should have used pgSurface_Lock/Unlock instead of SDL_LockSurface/UnlockSurface.  Let me know if I should change that.


nthykier:
I guess #1649 answered my question and appears to be a better solution than this. :)

PR #1649, opened 2020-04-16
RLE fixes

robertpfeiffer:


robertpfeiffer:
What maybe remains to be done here is looking at drawing and figuring out if we need to lock/prep there too.

robertpfeiffer:
Plus a test case that shows when exactly #1645 wouldn't work.

PR #1652, opened 2020-04-19
Add -Werror for a subset of compiler warnings

nthykier:
This commit introduces a "-pygame-ci" option that set certain compiler
warnings to errors.  It can be combined with the "-warnings" flag to
also include the warnings from that option.

The new option is also added to travis when run on Linux.


nthykier:
This will need a few CI runs to ensure it works as intended.  I will remove the draft/WIP flag when it is ready for review.

nthykier:
Related issue: #1316

This would create a baseline for which warnings we aim to have permanently fixed (though only measured on the Linux builds).

PR #1657, opened 2020-04-20
KMOD_MODE was not properly described. Is supposed to indicate that th…

glennmackintosh:
…e AltGr modifier key is in use

The description for the key modifier KMOD_MODE was described as 'mode' which does not indicate what modifer key is being depressed.
In SDL, KMOD_MODE is described as indicating that the AltGr modifier key is being pressed. This should be the same here in pygame.

### DIFF
--- a/docs/reST/ref/key.rst
+++ b/docs/reST/ref/key.rst
@@ -203,7 +203,7 @@ can be assembled by bitwise-ORing them together.
       KMOD_META     left meta or right meta or both
       KMOD_CAPS     caps lock
       KMOD_NUM      num lock
-      KMOD_MODE     mode
+      KMOD_MODE     AltGr
 
 
 The modifier information is contained in the ``mod`` attribute of the


glennmackintosh:
Glad I was able to contribute. I wanted to correct this since I spent a
bunch of time trying to figure out what this mod actually did. I'm hoping
to prevent the next guy from having to repeat that.

I am confused though. I can see that this is now corrected out on GitHub,
but when I look at the doc https://www.pygame.org/docs/ref/key.html it
still  says "KMOD_MODE mode". I thought that getting it fixed on GitHub
would get it corrected on pygame.org as well. DO I have to do something
else or will it just take a bit of time to filter to there?

Glenn Mackintosh

illume:
@glennmackintosh it's updated now. There was an problem with the webhook which updates the website.

glennmackintosh:
Great, thanks for the update.

Glenn Mackintosh

Issue #1535, opened 2019-12-19
pygame.errror exceptions are too generic


bendavis78:
- [x] unit test for FileNotFoundError exception for pygame.image.load (**good first issue**)
- [x] investigate where in the pygame code different exceptions could be raised (rwops.c? image_ext.c?)
- [x] file opening code in `src_c/` to raise FileNotFoundError when file isn't there.
- [x] look at mapping other SDL file related error codes to python exceptions. Not just FiltNotFoundError.

---

When loading an image with a file that doesn't exist, you get a `pygame.error` exception with a human-readable string description of the error. This is an anti-pattern. It would be better to allow the built-in exception `IOError` or `FileNotFoundError` to bubble up.

Let's say I want to handle an error where an image isn't found on disk. I currently have to check `os.path.isfile(path)`. If the file does exist, and I use `pygame.image.load(path)`, that's two file stats, which is not as efficient as simply catching the exception that is normally thrown when a file doesn't exist.

Either the original exception should be allowed to raise, or pygame should define its own version of `FileNotFoundError`.  There may be other instances of this anti-pattern throughout pygame, but so far this is all I've come across.

illume:
This is a good enhancement.

I don't have time to do this myself anytime soon, but PRs are welcome.

krkakade:
Can I work on it?

vjs22334:
I would like to work on this if it's still available

MightyJosip:
Of course, any help would be appreciated

jshultz:
I would be interested to work on this as well.

illume:
Here are a couple of error codes that [IMG_GetError](https://www.libsdl.org/projects/SDL_image/docs/SDL_image_45.html) returns.

```python
>>> import pygame
pygame 2.0.0.dev7 (SDL 2.0.12, python 3.7.4)
Hello from the pygame community. https://www.pygame.org/contribute.html
>>> pygame.image.load("asdf.png")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
pygame.error: Couldn't open asdf.png
>>> pygame.image.load("/etc/passwd")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
pygame.error: Unsupported image format
```

illume:
Closing thanks to the PR from @DotMars that was merged. Thanks!

PR #1660, opened 2020-04-22
image.load throws FileNotFoundError instead of the generic pygame.error

DotMars:
This PR is an answer to issue #1535. 
-Unit test for FileNotFoundError.
-pygame.image.load("nonexistent.png") throws a FileNotFoundError now.
-Fixed error in examples/camera.py

### DIFF
--- a/src_c/imageext.c
+++ b/src_c/imageext.c
@@ -215,9 +215,14 @@ image_load_ext(PyObject *self, PyObject *arg)
     }
 
     if (surf == NULL){
-        SDL_ClearError();
-        PyErr_SetString(PyExc_FileNotFoundError, "No such file or directory.");
-        return NULL;
+        if (!strcmp(IMG_GetError(), "Couldn't open fname.asdf")){
+            SDL_ClearError();
+            PyErr_SetString(PyExc_FileNotFoundError, "No such file or directory.");
+            return NULL;
+        }
+        else{
+            return RAISE(pgExc_SDLError, IMG_GetError());
+        }
     }
     final = pgSurface_New(surf);
     if (final == NULL) {
--- a/src_c/imageext.c
+++ b/src_c/imageext.c
@@ -216,7 +216,8 @@ image_load_ext(PyObject *self, PyObject *arg)
 
     if (surf == NULL){
         SDL_ClearError();
-        return RAISE(PyExc_FileNotFoundError, "No such file or directory.");
+        PyErr_SetString(PyExc_FileNotFoundError, "No such file or directory.");
+        return NULL;
     }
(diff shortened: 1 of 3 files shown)


src_c/imageext.c
```diff
@@ -214,8 +214,9 @@ image_load_ext(PyObject *self, PyObject *arg)
         PyMem_Free(ext);
     }
 
-    if (surf == NULL) {
-        return RAISE(pgExc_SDLError, IMG_GetError());
+    if (surf == NULL){
+        SDL_ClearError();
+        return RAISE(PyExc_FileNotFoundError, "No such file or directory.");
```
illume, line 219: Maybe something like this would compile (untested):
```c
SDL_ClearError();
PyErr_SetString(PyExc_FileNotFoundError, "No such file or directory.");
return NULL;
```

illume:
Great. Thanks for working on this.

IMG_GetError returns "Couldn't open fname.asdf"
You could use strcmp to see if it is this error string and raise the FileNotFoundError exception then? Otherwise continue raising the pygame.error.

MightyJosip:
Great job. It is nice to see the new contributors to pygame. Also you should change the `test_load_unknown_extension` function in imageext_test.py and make it check for raising the FileNotFoundError and than the travis and appveyor builds should work.

DotMars:
Guys thanks for the welcoming input. Right now I am stuck trying to get the tests to pass and here's my error code : 
`loading pygame.tests.draw_test
Segmentation fault (core dumped)`

Any idea how to get this pass ?

src_c/imageext.c
```diff
@@ -215,9 +215,14 @@ image_load_ext(PyObject *self, PyObject *arg)
     }
 
     if (surf == NULL){
-        SDL_ClearError();
-        PyErr_SetString(PyExc_FileNotFoundError, "No such file or directory.");
-        return NULL;
+        if (!strcmp(IMG_GetError(), "Couldn't open fname.asdf")){
```
MightyJosip, line 218: I think you should remove fname.asdf from your strcmp method. Basically you can only check
`if (!strcmp(IMG_GetError(), "Couldn't open")){`. In the comment illume left he meant that it returns filename.extension (fname.asdf was just the example) and then this should work.

DotMars: Thanks for the input. I ended up using strncmp() to compare only the beginning of IMG_GetError and I also add a check for PY3 because FileNotFoundError is not defined in Python2.7.
All the tests seem to be passing now. Please let me know if there was anything else to change.

MightyJosip:
Great work. Thanks 👍.

PR #1667, opened 2020-04-25
Some docs fixes

MightyJosip:
Fixed return type of some display functions and fixed a typo I made before in draw.rst

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -428,7 +428,7 @@ required).
 .. function:: iconify
 
    | :sl:`Iconify the display surface`
-   | :sg:`iconify() -> int`
+   | :sg:`iconify() -> bool`
 
    Request the window for the display surface be iconified or hidden. Not all
    systems and displays support an iconified display. The function will return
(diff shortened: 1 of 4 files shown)


nthykier:
Thanks for working on this. :) 

In regards to the return values, I personally think it would make sense to return a true bool instead.  There seems to be little risk of breaking things - even if people were using 0 and 1, python appears to be relatively forgiving:

```python
>>> True == 1
True
>>> False == 0
True
>>> 
```

MightyJosip:
This could also be another way of doing this. We can either change the docs or change return values from PyInt_FromLong to PyBool_FromLong.

nthykier:
I have open a PR #1673 to return the proper type from these functions **except** toogle_fullscreen.  The toggle_fullscreen can return -1, 0 and 1, so it is not a proper bool in the first place.  Here I think we should change the docs to match (to agree to stop returning -1).

illume:
Thanks.

@MightyJosip shall we remove the bool->int doc changes in favour of the #1673 PR (which converts them from ints to bools)?

MightyJosip:
Yes tomorrow I will remove it (you can merge #1673), and also I found some other errors while I was doing type hints so I will add them also

PR #1670, opened 2020-04-26
Optimised circle and rect drawing

robertpfeiffer:
Faster methods for drawing filled rects and circles of all kinds.

Edited to add: This pull request also changes the API of surflock to only accept pgSurfaceObject pointers for locking, instead of generic PyObjects. This type change doesn't break any existing code, is more correct on all fronts, and will ensure this bug I was tracking down for two weeks won't happen again.


illume:
Will the failure be fixed if you rebase against master? I think the 305a5e9dde3bba0679fc33d6cbe3b088737675e1 fix you committed last night fixes it?

robertpfeiffer:
You just requested changes without highlighting any code, right? I can't see anything when I click on "view changes"

MyreMylar:
Could Python 2 be struggling with this part?:

            SDL_Rect sdlrect;
            SDL_Rect cliprect;
            int result;
            SDL_Rect clipped;

Perhaps those need to be at the top of the function instead of in the middle.

robertpfeiffer:
> Could Python 2 be struggling with this part?:
> 
> ```
>         SDL_Rect sdlrect;
>         SDL_Rect cliprect;
>         int result;
>         SDL_Rect clipped;
> ```
> 
> Perhaps those need to be at the top of the function instead of in the middle.

I don't see how. They are defined at the beginning of the block. If it's a problem with SDL_Rect, it would be based on the SDL version, not the Python version.

The test segfaults when checking for key scankode constants

MyreMylar:
Well that didn't work.

Nothing is jumping out at me. Perhaps the easiest thing would be to split it into two additional PRs by the changed functions and then see which one is inexplicably killing python 2?

MyreMylar:
OK, running it locally on windows Python 2.7 it is this test that seg faults:

    def test_path_data_validation(self):
        """Test validation of multi-point drawing methods.

        See bug #521
        """
        surf = pygame.Surface((5, 5))
        rect = pygame.Rect(0, 0, 5, 5)
        bad_values = (
            "text",
            b"bytes",
            1 + 1j,  # string, bytes, complex,
            object(),
            (lambda x: x),
        )  # object, function
        bad_points = list(bad_values) + [(1,), (1, 2, 3)]  # wrong tuple length
        bad_points.extend((1, v) for v in bad_values)  # one wrong value
        good_path = [(1, 1), (1, 3), (3, 3), (3, 1)]
        # A) draw.lines
        check_pts = [(x, y) for x in range(5) for y in range(5)]

        for method, is_polgon in (
            (draw.lines, 0),
            (draw.aalines, 0),
            (draw.polygon, 1),
        ):
            for val in bad_values:
                # 1. at the beginning
                draw.rect(surf, RED, rect, 0)
                    if is_polgon:
                        method(surf, GREEN, [val] + good_path, 0)
                    else:
                        method(surf, GREEN, True, [val] + good_path)

                # make sure, nothing was drawn :
                self.assertTrue(all(surf.get_at(pt) == RED for pt in check_pts))

                # 2. not at the beginning (was not checked)
                draw.rect(surf, RED, rect, 0)
                    path = good_path[:2] + [val] + good_path[2:]
                    if is_polgon:
                        method(surf, GREEN, path, 0)
                    else:
                        method(surf, GREEN, True, path)

                # make sure, nothing was drawn :
                self.assertTrue(all(surf.get_at(pt) == RED for pt in check_pts))

using the command line:

    python -m pygame.tests.draw_test -v

Example output:

    test_lines__valid_color_formats (__main__.DrawLinesTest)
    Ensures draw lines accepts different color formats. ... ok
    test_lines__valid_points_format (__main__.DrawLinesTest)
    Ensures draw lines accepts different points formats. ... ok
    test_lines__valid_width_values (__main__.DrawLinesTest)
    Ensures draw lines accepts different width values. ... ok
    test_color_validation (__main__.DrawModuleTest) ... ok
    test_path_data_validation (__main__.DrawModuleTest)
    Test validation of multi-point drawing methods. ... Fatal Python error: (pygame parachute) Segmentation Fault

MyreMylar:
Specifically it hangs with this line:

    draw.rect(surf, RED, rect, 0)

where:

surf = <Surface(5x5x32 SW)>
RED = (255, 0, 0, 255)
rect = <rect(0, 0, 5, 5)>
width = 0

MyreMylar:
OK, from my testing removing these two lines gets rid of the seg fault:

            sdlrect.x = rect->x;
            sdlrect.y = rect->y;
            sdlrect.w = rect->w;
            sdlrect.h = rect->h;

            SDL_GetClipRect(surf, &cliprect);


            if (!SDL_IntersectRect(&sdlrect,
                                   &cliprect,
                                   &clipped)) {
                return pgRect_New4(rect->x, rect->y, 0, 0);
            }

            //pgSurface_Prep(self); // Causes seg fault on python 2
            result = SDL_FillRect(surf, &clipped, color);
            //pgSurface_Unprep(self); // Causes seg fault on python 2

            if (result == -1)
            return pgRect_New(&clipped);

Does that seem logical? The draw tests all seem to pass on Python 2 with them removed.

robertpfeiffer:
Well, shoot. I already know locking and prepping are broken, but I didn't expect this.

illume:
Does it make sense to use the same locking pattern as in surface.c?
```c
pgSurface_Prep(self);
pgSurface_Lock(self);
result = SDL_FillRect(surf, &clipped, color);
pgSurface_Unlock(self);
pgSurface_Unprep(self);
```

src_c/draw.c
```diff
@@ -897,19 +906,44 @@ rect(PyObject *self, PyObject *args, PyObject *kwargs)
...
+            sdlrect.w = rect->w;
+            sdlrect.h = rect->h;
 
-        ret = polygon(NULL, poly_args, NULL);
-        Py_DECREF(poly_args);
-        return ret;
+            SDL_GetClipRect(surf, &cliprect);
+
+
+            if (!SDL_IntersectRect(&sdlrect,
```
illume, line 933: Is the old draw.rect code returning the same rect as this one? We need a test case for this.

There is `test_rect__surface_clip` but it doesn't seem to assert against the returned rect.

MightyJosip: For that we have `test_rect__bounding_rect` function: https://github.com/pygame/pygame/blob/3ab109795a90489440856a269fd033545de6c694/test/draw_test.py#L4815

illume: Cool.

However, it doesn't use `Surface.set_clip`.

PR #1676, opened 2020-04-26
Gh 45 fuzzy color matching in all functions

nthykier:
Support more color definitions (e.g. "blue" or "0x123456") in most functions.

Note that as a part of this, I also made `font_render` more strict about the background color parameter (ad3c59dd59bccce7b671312971a8d99f70451006).  I am considering to cherry-pick that commit to master regardless of whether this branch is merged/accepted in its current form as I think this behaviour is a lot better. Originally, I did it to simplify the changes, but on reflection, I think it makes better sense to raise an exception rather than silent ignore the color argument.


illume:
Awesome. I've been wanting this feature for the longest time. 

Haven't had a chance to read through the code yet.

src_c/color.c
```diff
@@ -2094,6 +2119,12 @@ pg_RGBAFromColorObj(PyObject *color, Uint8 rgba[])
     return pg_RGBAFromObj(color, rgba);
 }
 
+static int
+pg_RGBAFromFuzzyColorObj(PyObject * color, Uint8 rgba[])
```
illume, line 2123: pg_RGBAFromFuzzyColorObj should probably be documented in docs/reST/c_api/color.rst

Issue #1515, opened 2019-11-19
Fatal Python error: PyEval_SaveThread: NULL tstate when using transform module.


MyreMylar:
**Platform**: Windows
**Pygame version**: 1.9.6+ (and probably some earlier versions)
**Python version**: 3.7 & 3.8 (at least) 

This is a frustratingly hard to pin down race condition threading bug I think. Basically every so often when the code below is run it will crash with a FatalError on the line where transform.smoothscale() is called.

Googling the problem leaves me more confused as most references are to this old python bug:
https://bugs.python.org/issue20891

But that was apparently resolved in 3.7. So is it possible the transform functions are trying to do some threading before they have have confirmed access to the GIL? Or another worker thread is snatching the GIL before one of the others is done with it? 

Or am I doing something crazy?

Apologies for the slightly long code, but with how infrequently the error pops (I'd guess at about 1/30 tries? it is hard to diagnose what is important and what is fluff.

**Error generated**
```
Fatal Python error: PyEval_SaveThread: NULL tstate
Python runtime state: initialized

Current thread 0x00000e7c (most recent call first):
  File "S:/Reddit_refactor/transform_bug.py", line 50 in apply_gradient_to_surface
  File "S:/Reddit_refactor/transform_bug.py", line 74 in <module>

```

**Code for reproduction:**

```
import pygame
from typing import Union


class ColourGradient:
    def __init__(self, angle_direction: int, colour_1: pygame.Color,
                 colour_2: pygame.Color, colour_3: Union[pygame.Color, None] = None):
        self.angle_direction = angle_direction
        self.colour_1 = colour_1
        self.colour_2 = colour_2
        self.colour_3 = colour_3

    def __eq__(self, other):
        if type(other) != ColourGradient:
            return False
        return (self.colour_1 == other.colour_1 and
                self.colour_2 == other.colour_2 and
                self.colour_3 == other.colour_3 and
                self.angle_direction == other.angle_direction)

    def apply_gradient_to_surface(self, input_surface: pygame.Surface, rect=None):
        """
        Applies this gradient to a specified input surface using blending multiplication.
        As a result this method works best when the input surface is a mostly white, stencil shape type surface.

        :param input_surface:
        :param rect: The rectangle on the surface to apply the gradient to.
        """
        input_surface_size = input_surface.get_size()
        inverse_rotated_input = pygame.transform.rotate(input_surface, -self.angle_direction)
        gradient_size = inverse_rotated_input.get_rect().size

[...]

PR #1678, opened 2020-04-27
smoothscale: Avoid calling GETSTATE without the gil

nthykier:
Under Python3, GETSTATE is PyModule_GetState and therefore a part of
the Python C-API (thus requiring the GIL when doing the call).
Compare with python2, where GETSTATE was just a trivial address-of
operation.

Closes: #1515

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -1457,20 +1457,23 @@ surf_scalesmooth(PyObject *self, PyObject *arg)
     if (width && height) {
         SDL_LockSurface(newsurf);
         pgSurface_Lock(surfobj);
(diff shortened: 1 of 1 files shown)


nthykier:
For reference, even with this commit, there is a potential bug in that the C object we are looking at could be modified from a different thread (e.g. via Python).  E.g. calling `set_smoothscale_backend` during a `smoothscale` call would make "interesting" results.  Though, that is a separate issue from calling a Python API function without the GIL in my opinion.

illume:
I tried on Mac, and made a test case to run on the CI that couldn't reproduce either: https://github.com/pygame/pygame/pull/1684

Issue #554, opened 2018-10-10
pygame.math vs math in pygame-Python-modules


e1000:
Very small technical internal issue: 

it seems that, within the new src_py/draw_py.py module,
the following code does not work in Python2 :

```python
from __future__ import relative_imports
from math import floor, ceil, trunc
```

Instead, **math** is interpreted as **pygame.math** and Python does not find the
functions ... But I would like to use those functions for an aaline algorithm.
Can it happen elsewhere? (eg usercode?)

For now it's not a big deal, and maybe we can wait until the problem disappears in 2020 :-)

I'd like to understand what is the fix of this problem though, it's a generic pb...

PR #1679, opened 2020-04-27
draw.py: Request absolute_import under Py2

nthykier:
Closes: #554


nthykier:
Another alternative is to close #554 given that Python2.7 support is official over now.

illume:
Nice one. absolute_import is documented here: https://www.python.org/dev/peps/pep-0328/

PR #1682, opened 2020-04-28
opinionated warnings

robertpfeiffer:


nthykier:
I think this is a good idea to raise warnings for most of these things to help programmers to avoid issues (I use "for most" as I don't recognize all the issues, so I do not feel I am an authority on that subject).

However, warnings may raise exception and the patches do not account for that. From the documentation:

"""
**Issuing warnings**

Use these functions to issue warnings from C code. They mirror similar functions exported by the Python warnings module. They normally print a warning message to sys.stderr; however, it is also possible that the user has specified that warnings are to be turned into errors, and in that case they will raise an exception. It is also possible that the functions raise an exception because of a problem with the warning machinery. *The return value is 0 if no exception is raised, or -1 if an exception is raised. (It is not possible to determine whether a warning message is actually printed, nor what the reason is for the exception; this is intentional.)* If an exception is raised, the caller should do its normal exception handling (for example, Py_DECREF() owned references and return an error value).
"""
Emphasis mine, source: https://docs.python.org/3/c-api/exceptions.html

illume:
There's no description in this PR.

illume:
Closing this to keep our PR list clean. Moving to an issue here https://github.com/pygame/pygame/pull/1682

nthykier:
ITYM: #2085

PR #1688, opened 2020-04-29
Added type check for pygame.sprite.LayeredDirty.set_timing_treshold()

MightyJosip:
set_timing_treshold() used to accept all values. the problem with it is for example in this code:
```
import pygame

class Dirty(pygame.sprite.DirtySprite):
    def __init__(self):
        pygame.sprite.DirtySprite.__init__(self)
        self.image = pygame.Surface((20, 20))
        self.rect = self.image.get_rect()

pygame.init()
screen = pygame.display.set_mode((100, 100))
layered_dirty = pygame.sprite.LayeredDirty(Dirty())
layered_dirty.set_timing_treshold("a")
layered_dirty.draw(screen)
```
used to raise `TypeError: '>' not supported between instances of 'int' and 'str'` on line `layered_dirty.draw(screen)`. With it you could think you did something wrong with your draw function, but actually you made the mistake where you set_timing_treshold to string
Now it raises `TypeError: Expected numeric value, got str instead` on the line `layered_dirty.set_timing_treshold("a")`.

### DIFF
--- a/docs/reST/ref/sprite.rst
+++ b/docs/reST/ref/sprite.rst
@@ -637,6 +637,8 @@ Sprites are not thread safe. So lock them yourself if using threads.
       Default is 1000./80 where 80 is the fps I want to switch to full screen
       mode.  This method's name is a typo and should be fixed.
 
+      :raises TypeError: if ``time_ms`` is not int or float
+
       .. ## LayeredDirty.set_timing_treshold ##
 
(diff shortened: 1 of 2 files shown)


nthykier:
I think it would definitely be an improvement to stop the bug in the `set_timing_treshold` (also, I think we have typo of `threshold` in that method name?).

PR #1690, opened 2020-04-29
Some lgtm fixes

MightyJosip:
For https://github.com/pygame/pygame/issues/1133

### DIFF
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -213,7 +213,7 @@ def warn_unwanted_files():
 
 
 try:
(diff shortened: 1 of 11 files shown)


src_py/__init__.py
```diff
@@ -213,7 +213,7 @@ def warn_unwanted_files():
 
 
 try:
-    from pygame.surface import *
+    from pygame.surface import Surface
```
illume, line 216: Should also import `SurfaceType`

PR #1694, opened 2020-05-01
Adds function to pygame.key module to get key code from name of key

MyreMylar:
Attempt to add a function to get a key code from a key name. There were some concerns in the original report #309 that this function wouldn't be SDL1 compatible but I'm not sure if that is a huge concern any more.

Also, do I need to add a:

    .. versionadded:: 2.0.0 

into the docs?

It's my first first attempt at adding a C extension Function so I may have done some obvious stuff wrong!


MightyJosip:
Great. Can you also add
`def key_code(name: str) -> int: ...`
somewhere in the buildconfig/pygame-stubs/key.pyi. And yes it would be a cool addition if you write version in the docs.

illume:
I think this needs to be more clear in the docs what this is returning.

Mention something about it matching K_* event types and an example like your unit test below?

```rst
   Example:

   ::

        >>> pygame.key.key_code("return") == pygame.K_RETURN
        True
        >>> pygame.key.key_code("0") == pygame.K_0
        True
        >>> pygame.key.key_code("space") == pygame.K_SPACE
        True

   .. ## pygame.key.key_code ##
```


We still want to be able to compile with SDL 1 for pygame 2.
```c
#if IS_SDLv1
/* maybe return 0? How would code that has to work on pygame 1 and 2 work? */
#else  /* IS_SDLv2 */
/*... your code goes here.*/
#endif /* IS_SDLv2 */
```

The SDL docs for SDL_GetKeyFromName have this note:
https://wiki.libsdl.org/SDL_Keycode
https://wiki.libsdl.org/SDL_GetKeyFromName

> Returns key code, or SDLK_UNKNOWN if the name wasn't recognized; call SDL_GetError() for more information. 

So your code needs to either raise an error if you get SDLK_UNKNOWN? Or should it just return 0? Maybe raising an error is better.

MyreMylar:
Added two exceptions:

1. Where the key name is unknown it returns 0 and raises a ValueError.
2. Where pygame is compiled with SDL 1 it should raise a NotImplementedError. Not certain how to test this one, I added a test for it but I don't know if the CI runs that setup.

And I added the example documentation suggested by @illume.

test/key_test.py
```diff
@@ -45,6 +47,17 @@ def test_name(self):
         self.assertEqual(pygame.key.name(pygame.K_0), "0")
         self.assertEqual(pygame.key.name(pygame.K_SPACE), "space")
 
+    def test_key_code(self):
```
MightyJosip, line 50: You should add first 4 asserts under `if not SDL1`, right now they are checked in every case (and as far as I have understood key_code should be undefined with SDL1

src_c/key.c
```diff
@@ -709,6 +709,37 @@ key_name(PyObject *self, PyObject *args)
...
+    };
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "s", kwids, &name))
+        return NULL;
+
+#if IS_SDLv1
+    PyErr_SetString(PyExc_NotImplementedError, "not supported with SDL 1");
+    return 0;
+#else
+    long code = SDL_GetKeyFromName(name);
```
MightyJosip: Can you try to add `long code;` under `const char * name;` and then replace this line with `code = SDL_GetKeyFromName(name);`. For some reasons sometimes on python 2 you are not allowed to declare variables after the start of function and that could be the reason why AppVeyor is failing with this weird syntax errors

test/key_test.py
MyreMylar: Good catch! I've fenced these off now... And now I see I could have just taken your version from teis conversation thread.

You can tell I'm not much of a GitHub master.

src_c/key.c
MyreMylar: I did this too.

MightyJosip:
Thanks 👍, I will approve this changes now, but since we are close to the release date I will let illume to decide if we include it in dev8.

Issue #202, opened 2014-06-07
Add timeout parameter while waiting for events


illume:
Originally reported by: **metagriffin (Bitbucket: [metagriffin](https://bitbucket.org/metagriffin), GitHub: [metagriffin](https://github.com/metagriffin))**

----------------------------------------

i think it would be great if the wait functions (`pygame.event.wait` and `pygame.fastevent.wait`) could accept a timeout parameter -- this would alleviate the need to duplicate a lot of code that uses the poll/sleep approach to reduce CPU consumption.

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/202

**To Do to resolve:**

- [x] Add timeout parameter to `pygame.event.wait()` after which time it will return True even if the event has not yet occurred.
- [ ] Add timeout parameter to `pygame.fastevent.wait()` after which time it will return True even if the event has not yet occurred.
- [x] Add unit tests for these new functions.
- [x] Update the docs.

**Related Docs**: https://www.pygame.org/docs/ref/event.html#pygame.event.wait

MyreMylar:
This was the original pull request since the `#31` link above actually points to an unrelated bug on github:

https://bitbucket.org/pygame/pygame/pull-requests/31/add-timeout-parameter-while-waiting-for/diff

I'm still a little lost on the use cases myself, even after reading the bitbucket thread.

MyreMylar:
**To Do to resolve:**

- [ ] Add timeout parameter to `pygame.event.wait()` after which time it will return True even if the event has not yet occurred.
- [ ] Add timeout parameter to `pygame.fastevent.wait()` after which time it will return True even if the event has not yet occurred.
- [ ] Add unit tests for these new functions.
- [ ] Update the docs.

This was started back on bitbucket with the use case suggested of a clock that updates the screen when a hand moves or, I guess, when the user prods it some way.

nthykier:
Wouldn't it be better (more reliable/simpler) to use a timer for this?

MyreMylar:
We now have a timeout parameter for `event.wait()` so just need one for the equivelent `fastevent.wait()`

ankith26:
I’ll probably try adding the timeout parameter to the fastevent.wait method

illume:
I'm not sure if fastevent is relevant anymore on SDL2. I'm pretty sure the new event loop is very fast with mulitple threads posting now... but haven't confirmed.

ankith26:
Hmm, so close this issue? 
[...]

PR #1695, opened 2020-05-02
Adds Fastevent module documentation 

MyreMylar:
This is to solve ##774

First pass is just transplanting the existing documentation from where it is buried in the code to using the existing system used by other modules so it appears in the auto generated docs.

I think this documentation also likely needs a review by someone at least passingly familiar with the fastevent module.

There are also a few issues and a related old pull request that mention the module here:

https://github.com/pygame/pygame/issues/946
https://github.com/pygame/pygame/issues/202
https://github.com/pygame/pygame/pull/1003

That may be relevent.

### DIFF
--- a/docs/reST/ref/fastevent.rst
+++ b/docs/reST/ref/fastevent.rst
@@ -1,0 +1,105 @@
+.. include:: common.txt
+
+:mod:`pygame.fastevent`
+=======================
+
+.. module:: pygame.fastevent
+   :synopsis: pygame module for interacting with events and queues from multiple
+              threads.
+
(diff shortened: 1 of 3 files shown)


illume:
Maybe add this line?
> The main benefit of pygame.fastevents is that you can `post()` events more efficiently from other threads. If you're not posting events from other threads, then just use the normal pygame.event module.

Is fastevents still faster with SDL2? I know the SDL2 event code underwent some serious improvements for threaded use cases, so perhaps fastevent isn't needed anymore. The test case is to use an app which post events into the main thread, and see if there is less or more latency with the two event modules.

PR #1704, opened 2020-05-03
SCALED related progress, for a cleaner merge this time

robertpfeiffer:
For #1365

- remove the TODO comment and extract screen detection into its own function
- add an explicit VSYNC parameter to set_mode, for both SCALED and OPENGL usage
- add Python runtime warnings for unstable API calls
- simplify nested conditionals
- add is_fullscreen method to be used instead of display surface flags. The display surface is not re-created with SCALED or OpenGL!
- force integer scaling for SCALED with nearest-neighbour scaling, but not when smooth scaling hints are detected, or in FULLSCREEN. This ensures the full screen is used in FULLSCREEN, even if some pixels have jagged edges as a result.


src_c/display.c
```diff
@@ -895,6 +901,14 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
     state->using_gl = (flags & PGS_OPENGL) != 0;
     state->scaled_gl = state->using_gl && (flags & PGS_SCALED) != 0;
 
+    if (state->scaled_gl){
+        PyErr_WarnEx(PyExc_FutureWarning, "SCALED|OPENGL is experimental and subject to change" ,1);
```
nthykier, line 905: Missing check of return code from `PyErr_WarnEx`

src_c/display.c
```diff
@@ -1079,6 +1130,25 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+
+            /* Even if this succeeds, we can never *really* know if vsync actually works.
+               There may be screen tearing, blocking double buffering, triple buffering,
+               render-offloading where the driver for the on-board graphics *doesn't* have
+               vsync enabled, or cases where the driver lies to us because the user has
+               configured vsync to be aways on or always off, or vsync is on by default
+               for the whole desktop because of wayland GL compositing. */
+            if (vsync) {
+                if (SDL_GL_SetSwapInterval(-1) != 0){
+                    PyErr_WarnEx(PyExc_Warning, "adaptive vsync for OpenGL not available, trying regular" ,1);
```
nthykier, line 1128: Missing check of return code from `PyErr_WarnEx`

src_c/display.c
```diff
@@ -1079,6 +1130,25 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+                    if (SDL_GL_SetSwapInterval(1) != 0){
+                        PyErr_WarnEx(PyExc_Warning, "regular vsync for OpenGL *also* not available", 1);
```
nthykier, line 1130: Missing check of return code from `PyErr_WarnEx`

src_c/display.c
```diff
@@ -1092,22 +1162,36 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+                              pg_renderer,
+                              !(flags & PGS_FULLSCREEN
+                                || SDL_GetHintBoolean("SDL_HINT_RENDER_SCALE_QUALITY",SDL_FALSE)));
                     SDL_RenderSetLogicalSize(pg_renderer, w, h);
+                    /* this must be called after creating the renderer!*/
                     SDL_SetWindowMinimumSize(win, w, h);
 
                     SDL_GetRendererInfo(pg_renderer, &info);
+                    if (vsync && !(info.flags & SDL_RENDERER_PRESENTVSYNC)) {
+                        PyErr_WarnEx(PyExc_Warning, "could not enable vsync" ,1);
```
nthykier, line 1179: Missing check of return code from `PyErr_WarnEx`

src_c/display.c
```diff
@@ -1092,22 +1162,36 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+                    }
+                    if (!(info.flags & SDL_RENDERER_ACCELERATED)) {
+                        PyErr_WarnEx(PyExc_Warning, "no fast renderer available" ,1);
```
nthykier, line 1182: Missing check of return code from `PyErr_WarnEx`

src_c/display.c
```diff
@@ -2281,6 +2389,7 @@ pg_toggle_fullscreen(PyObject *self, PyObject *args)
 #if SDL_VERSION_ATLEAST(2, 0, 4)
         case SDL_SYSWM_ANDROID:  // currently not supported by pygame
 #endif
+            PyErr_WarnEx(PyExc_Warning, "cannot leave FULLSCREEN on this platform" ,1);
```
nthykier, line 2381: Missing check of return code from `PyErr_WarnEx`

src_c/display.c
```diff
@@ -2401,6 +2515,7 @@ pg_toggle_fullscreen(PyObject *self, PyObject *args)
              * This is only relevant in the non-GL case. */
             int wx = SDL_WINDOWPOS_UNDEFINED_DISPLAY(window_display);
             int wy = SDL_WINDOWPOS_UNDEFINED_DISPLAY(window_display);
+            PyErr_WarnEx(PyExc_Warning, "re-creating window in toggle_fullscreen" ,1);
```
nthykier, line 2510: Missing check of return code from `PyErr_WarnEx`

src_c/display.c
```diff
@@ -2432,7 +2547,9 @@ pg_toggle_fullscreen(PyObject *self, PyObject *args)
             if (result != 0) {
                 return RAISE(pgExc_SDLError, SDL_GetError());
             }
-            if (r_info.flags & SDL_RENDERER_SOFTWARE) {
+            if (r_info.flags & SDL_RENDERER_SOFTWARE
+                && wm_info.subsystem == SDL_SYSWM_X11) {
+                PyErr_WarnEx(PyExc_Warning, "recreating software renderer in toggle_fullscreen" ,1);
```
nthykier, line 2544: Missing check of return code from `PyErr_WarnEx`

src_c/display.c
```diff
@@ -2481,6 +2601,7 @@ pg_toggle_fullscreen(PyObject *self, PyObject *args)
             display_surface->surf = SDL_GetWindowSurface(win);
         }
         else if (wm_info.subsystem == SDL_SYSWM_WAYLAND) {
+            PyErr_WarnEx(PyExc_Warning, "skipping toggle_fullscreen on wayland" ,1);
```
nthykier, line 2598: Missing check of return code from `PyErr_WarnEx`

src_c/display.c
```diff
@@ -2499,6 +2620,7 @@ pg_toggle_fullscreen(PyObject *self, PyObject *args)
                 }
                 display_surface->surf = SDL_GetWindowSurface(win);
                 pg_SetDefaultWindow(win);
+                PyErr_WarnEx(PyExc_Warning, "re-creating window in toggle_fullscreen" ,1);
```
nthykier, line 2617: Missing check of return code from `PyErr_WarnEx`

src_c/display.c
```diff
@@ -2218,6 +2307,25 @@ pg_get_desktop_screen_sizes(PyObject *self, PyObject *args)
     return result;
 }
 
+static PyObject *
+pg_is_fullscreen(PyObject *self, PyObject *args)
```
nthykier, line 2300: Looks great. :)

Should we extract this into a separate commit/branch? (It looks self-contained and a no-brainer to merge)

src_c/display.c
robertpfeiffer: What exactly does that return value mean? https://docs.python.org/3/c-api/exceptions.html#c.PyErr_WarnEx

src_c/display.c
robertpfeiffer: I'd really rather not. I would like to have this (and also #1705) in the same release with the warnings. I don't want people to depend on undocumented behaviour and then file bugs about it.

src_c/display.c
robertpfeiffer: Never mind I saw your comment in the other thread

src_c/display.c
robertpfeiffer: You were so right about the warnings. Obviously I didn't just *forget* to check the return code ten times or so, I had the completely wrong idea about how this works! In hindsight it *obviously* has to work this way, you can't just throw an exception in C code!

src_c/display.c
nthykier: Ok, fine with me. :)

src_c/display.c
```diff
@@ -1079,6 +1117,31 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+                if (SDL_GL_SetSwapInterval(-1) != 0){
+                    if (PyErr_WarnEx(PyExc_Warning, "adaptive vsync for OpenGL not available, trying regular", 1) != 0) {
+                        _display_state_cleanup(state);
+                        goto DESTROY_WINDOW;
+                    }
+                    if (SDL_GL_SetSwapInterval(1) != 0){
+                        if(PyErr_WarnEx(PyExc_Warning, "regular vsync for OpenGL *also* not available", 1) != 0) {
+                            _display_state_cleanup(state);
+                            goto DESTROY_WINDOW;
+                    }
```
nthykier, line 1137:
```suggestion
                            }
```

(Avoid confusion on about where the warning scrope ends)

robertpfeiffer:
cross-reference #1365

robertpfeiffer:
I did the thing, and the appveyor build broke while downloading dependencies probably because some third party host was offline.

illume:
Thanks. I reran the broken appveyor job.

illume:
Someone still needs to test this. I'll get around to it later if no one else does, but my macos build is currently not working.

MyreMylar:
So I think the conclusion on this after Discord discussion yesterday was that we should merge it even though there is still some odd behaviour using the SCALED flag on windows in a window (it works fine full screen) because the SCALED flag is only intended to be used 'for pixel art on HD screens and phones'. And because that behaviour is SDL's fault anyway.

As far as I can tell the new vsync parameter for set_mode() works fine as a replacement for the old environment flag for toggling vsync on and off (even if vsync support is incidental and not the intention of this PR).

I can try and phrase the vsync documentation (currently in a separate PR) in a way that doesn't imply that the current behaviour of the SCALED flag is the intended pygame path for making windowed desktop apps with vsync in the future.

The scaling up of low res screens (I tested with 640x480) worked fine full screen and in a window.

PR #1710, opened 2020-05-06
Adds version.SDL to version module, fixes #951

### LINKED ISSUE #951: Add SDL version info to pygame.version module
Maybe like this:

```python
>>> pygame.version.SDL
(1, 2, 15)
```

```python
>>> pygame.version.SDL
(2, 0, 9)
```

MyreMylar:
For @robertpfeiffer.

### DIFF
--- a/buildconfig/pygame-stubs/version.pyi
+++ b/buildconfig/pygame-stubs/version.pyi
@@ -1,6 +1,20 @@
-from typing import Tuple
+from typing import Tuple, Union
 
-SDL: Tuple[int, int, int]
+
+class SoftwareVersion(tuple):
+    def __new__(cls, major: int, minor: int, patch: int) -> PygameVersion: ...
+    def __repr__(self) -> str: ...
+    def __str__(self) -> str: ...
+
+    major: int
+    minor: int
+    patch: int
+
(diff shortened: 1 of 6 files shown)


nthykier:
Should be pre-emptively wrap that in a `PygameVersion` (or similar structure) as is done with Pygame's own version?

MyreMylar:
> Should be pre-emptively wrap that in a `PygameVersion` (or similar structure) as is done with Pygame's own version?

I was pondering that.. PygameVersion seemed like the wrong class to use because of the name but I guess if we renamed the class 'SoftwareVersion' or something it could work.

MyreMylar:
Tried out just reusing the PygameVersion class. The main problem is that when you print it for SDL it you get a string like this:

    >>> pygame.version.SDL
    PygameVersion(major=2, minor=0, patch=12)

Which seems like it could potentially be confusing.

Perhaps the best Idea is to move the functionality of PygameVersion to a base class and create two thin shims around it called PygameVersion and SDLVersion...

MyreMylar:
Thought I might as well add the better autocomplete data to the .pyi file in this PR rather than another one.

PR #1711, opened 2020-05-07
Adds MMX, SSE2 & Arm NEON optimised versions of blit_blend_premultiplied()

MyreMylar:
This should make using pre-multiplied alpha blits pretty efficient on most desktop OS's I believe (anything that supports the old MMX instruction set). It performs similar to (or slightly better) than standard SDL alpha blitting.

**Caveats:**

 - This is ported from SDL here: https://hg.libsdl.org/SDL/file/015943013626/src/video/SDL_blit_A.c#l330
with two of the instructions removed to change it from regular alpha blending to pre-multiplied alpha.

- MMX is a fairly outdated standard - supplanted by SSE, SSE2, SSE3 etc but it still seems to be what SDL is using for it's alpha blit operations on my computer - if I've followed the code paths correctly.

- This definitely won't do anything to speed up ARM architectures, they will still use the old slow implementation. There are similar functions to the one I ported for two different ARM SIMD standards but they are written in assembly e.g. here:
https://hg.libsdl.org/SDL/file/015943013626/src/video/arm/pixman-arm-neon-asm.S#l226
which I don't understand enough to port.

**Further related work**

- Building a performance test case and testing the performance of this on multiple desktop platforms e.g. mac, linux. I mean it's nice that it works fast on my PC, but if it's only on my PC then it's not much good.
- Consider exposing the various SDL functions for checking SIMD instruction set support from pygame so we can quickly query them, use them in code to branch, or ask users to report them. e.g. SDL_HasMMX(), SDL_Has3DNow(), SDL_HasSSE2() etc.
- Consider adding utility functions for working with pre-multiplied alpha to pygame.Color, pygame.image and pygame.font. Right now in pygame GUI I'm wrapping output from these modules to change surfaces over to a pre-multiplied alpha blending format. 
- Could other pygame implemented blends in alphablit.c be sped up in the same way - or _perhaps_ in a better way with a slightly more recent set of SIMD instructions (SSE2 is now as old as Python 2)?


MyreMylar:
This is related to: https://github.com/pygame/pygame/issues/1289

MyreMylar:
I ported the MMX implementation over to SSE2 to make it more portable to x64 platforms. 

It's as straight a port as I could make of the MMX code rather than taking advantage of anything special about SSE2. Though SSE2 was designed to allow this so it shouldn't suffer much from the straight conversion.

While poking around researching this I also came across [this article](http://0x80.pl/articles/sse4-alphaover.html) featuring SSE3/SSE4 versions of the same blending operation. However, its in assembly which I have no experience in. It does outline the algorithm though which might inspire an enterprising soul.

MyreMylar:
I'm still working on testing the ARM port on Raspberry Pi to see if it works. If it does then that'll potentially clear up all my initial caveats.

Note: I'm not certain on the correct place or method of incorporating the sse2neon header, it's under the MIT license - do we have to add a (another) copy of the MIT LICENSE.txt somewhere?

MyreMylar:
As far as I can tell from Pi the ARM sse2neon code works (it compiles and runs and seems to do alpha multiplied blending) - though my current test project has lots of visual errors on the pi under SDL2 anyway as it turns out whether using this code or not.

The only thing missing in this change list is a way to activate the ARM code only on ARM platforms. you need to add the flag `-mfpu=neon` to make it work. Without this flag the compiler will error.

I'm not sure how to add this in a platform safe way to the setup scripts.

MyreMylar:
I added a bit to setup.py that checks if the machine name being compiled on begins with 'arm' and then adds the `-mfpu=neon` flag. That should detect raspberry pis, I think the other main arm neon supporting platform is android so hopefully it will also detect that.

illume:
Very interesting about the ARM/Neon stuff.

Seems Neon isn't available on some raspberries (rpi 1, compute module 1, and the rpi zeros).
https://en.wikipedia.org/wiki/Raspberry_Pi#Specifications this is because NEON is optional on arm.

I'm not sure doing it like smoothscale work work for this, because it compiles slightly differently.
Perhaps compiling a separate python module with `-mfpu=neon` could work. This way neon support could be detected at runtime. Then, if neon, load the `surface_neon.so`, .

I'm not sure if you can say "just compile this function with -mfp=neon"? I don't think so.

Perhaps there is a distutils hack that could compile a separate surface_sse.c with `-mfp=neon`, then that one could just link the function at runtime too. This way it compiles with `-mfp=neon`, and then the function is available. Hrmm.

MyreMylar:
> Very interesting about the ARM/Neon stuff.
> 
> Seems Neon isn't available on some raspberries (rpi 1, compute module 1, and the rpi zeros).
> https://en.wikipedia.org/wiki/Raspberry_Pi#Specifications this is because NEON is optional on arm.
> 
> I'm not sure doing it like smoothscale work work for this, because it compiles slightly differently.
> Perhaps compiling a separate python module with `-mfpu=neon` could work. This way neon support could be detected at runtime. Then, if neon, load the `surface_neon.so`, .
> 
> I'm not sure if you can say "just compile this function with -mfp=neon"? I don't think so.
> 
> Perhaps there is a distutils hack that could compile a separate surface_sse.c with `-mfp=neon`, then that one could just link the function at runtime too. This way it compiles with `-mfp=neon`, and then the function is available. Hrmm.

There is a runtime check for ARM NEON support in the code as well as a compile time check - it should't try to compile it if ARM support isn't detected. The compile time check uses:

`#ifdef SDL_ARM_NEON_BLITTERS`

and the runtime check uses:

`if (SDL_HasNEON() == SDL_TRUE)`

But I don't think those SDL defines and functions are available to us in setup.py so I improvised with the 'arm' machine name detection. I may be wrong...

I assume that if neon is not supported, and we also don't pass it any neon code, then the compiler flag will do nothing.

MyreMylar:
(I'll fix that LGTM alert)

setup.py
```diff
@@ -101,6 +101,16 @@ def compilation_help():
...
     sys.argv.remove ('-pygame-ci')
 
+# This should detect Raspberry pis at least.
+from platform import machine
+if machine().startswith('arm'):
+    cflags = os.environ.get('CFLAGS', '')
+    if cflags:
+        cflags += ' '
+    cflags += '-mfpu=neon'
+    os.environ['CFLAGS'] = cflags
```
illume, line 111: I think this will apply to all files, making all fpu code use neon - even outside this blit routine.

For binary wheels, built on say a Debian or piwheels build server, they would compile with neon. But then a computer on say a RPI 1 loading that wheel would fail (I think).

MyreMylar: Yes, I think you are right. 

I also don't think that you can upload wheels that are that specific to PyPI, it looks like the 2014 'manylinux' standard sets armv7 as minimum suggesting that there is no support for armv6 differentiation.

I guess the option that SDL, openCV and tensorflow are using is to upload wheels/binaries  _without_ neon optimisations and then allowing a build flag to enable them when you build from source so users can choose to build with it or not. I think we should probably do the same.

SDL link:
https://www.phoronix.com/scan.php?page=news_item&px=SDL-ARM-NEON-Optimizations

OpenCV link for ENABLE_NEON flag:
https://www.pyimagesearch.com/2017/10/09/optimizing-opencv-on-the-raspberry-pi/

nthykier:
@MyreMylar Do we have any automated test cases that use `blit_blend_premultiplied()`?  It would be really helpful for writing a SSE4.2 port of that function.

MyreMylar:
> @MyreMylar Do we have any automated test cases that use `blit_blend_premultiplied()`? It would be really helpful for writing a SSE4.2 port of that function.

Not yet, I wasn't expecting SSE4.2 to happen :)

I can probably work something up for now and then modify it later when I add utility functions for working with pre-multiplied alpha colours.

setup.py
MyreMylar: Looking at SDL's CMake it appears they set the ARMNEON option _on_ by default here:
https://hg.libsdl.org/SDL/file/4298bf108b06/CMakeLists.txt#l332

Which then tries neon fpu in this block here I believe:
https://hg.libsdl.org/SDL/file/4298bf108b06/CMakeLists.txt#l704

Which does a little check to see if it can compile some assembly source code with ARM blitters before actually enabling them. So perhaps we can borrow this `check_c_source_compiles` check from SDL and use that to toggle the fpu flag on or off?

illume: Note, that raspberian is set up to load modules from https://www.piwheels.org/ first (see /etc/pip.conf). They try and keep everything compatible with `armv6`, because they still ship new hardware with that (the zero's), and this is the reason for the Raspberian distro - since Debian supported a higher version. https://www.raspbian.org/RaspbianFAQ#What_is_Raspbian.3F

It's disappointing that the manylinux effort messed up with raspberian support. At least now there is i386 and some ARM support in 2014. I haven't kept up with it the changes in the last half year.

---

> I guess the option that SDL, openCV and tensorflow are using is to upload wheels/binaries without neon optimisations and then allowing a build flag to enable them when you build from source so users can choose to build with it or not. I think we should probably do the same.

I guess compiling it with `CFLAGS=-mfpu=neon setup.py install` would work for this option? Platforms with these flags built in won't need to do that either.


---

Here's another way we could hack up distutils based on python modules and compilation flags. (obviously these ways would be out of scope for this PR)

Currently Setup.SDLX.in has stuff like this.
```
surface src_c/surface.c src_c/alphablit.c src_c/surface_fill.c $(SDL) $(DEBUG)
```

Adding CPU specific modules could be done like this:
```
surface_fpu_neon src_c/surface.c src_c/alphablit.c src_c/surface_fill.c $(SDL) $(DEBUG) -mfpu=neon
```
Then the init machinery (`__init__.py`) can load the correct `surface` (or `surface_fpu_neon`) module based on the CPU capabilities. I guess all the other C based routines would also be sped up because of better compiler optimizations too.

This is obviously complicated. I feel if we only rely on compiler flags, then most users will not be able to take advantage of the optimizations -- because most users will use the pre-built binaries.

illume: So, the way forward for this PR is to remove the setup.py changes, and to require people to add the CFLAGS themselves on build?

src_c/alphablit.c
```diff
@@ -25,6 +25,10 @@
 #define NO_PYGAME_C_API
 #include "_surface.h"
 
+#ifdef SDL_ARM_NEON_BLITTERS
+#include "include/sse2neon.h"
```
illume, line 29: A URL here to point out where sse2neon came from would be good.

nthykier:
> > @MyreMylar Do we have any automated test cases that use `blit_blend_premultiplied()`? It would be really helpful for writing a SSE4.2 port of that function.
> 
> Not yet, I wasn't expecting SSE4.2 to happen :)
> 
> I can probably work something up for now and then modify it later when I add utility functions for working with pre-multiplied alpha colours.

Not yet sure there will be one either. :)  I think there is a bug in `_mm_mulhi_epu16` / `pmulhuw` where it miscomputes "0xff00 * 0x0001 / 0xff" to be 0x00 rather than 0x01...

setup.py
MyreMylar: What I added in the latest commit was a flag for setup.py called `-ENABLE_NEON` mainly so that I could also add a matching define (PG_ENABLE_ARMNEON) to the c preprocessor so that the compiler instructions match whats in the code. 

In theory you should be able to enable the ARM intrinsics like this:

`python3 setup.py install -ENABLE_NEON`

This was me mimicking what is done for the CI build configuration which has this flag:

`python3 setup.py install -pygame-ci`

I'm not attached to the all-caps, probably they should be consistent.

Previously, I was using the SDL defines for including and removing the ARM code - but I realised there could be a mismatch between whatever SDL was doing with ARM and what pygame was enabling if we didn't have our own define.

I still need to run tests on my Pi4 with and without the flag to check it works but I think it's a reasonable stop gap for this pull request.

------

I think your idea of pulling the ARM code out into it's own .c file compiled with the flag for the wheels is a good one though. I suppose that since we have a runtime check for NEON capabilities as well we could just drag the actual function body over to the .c file compiled with `mfpu-neon` and import a header with just the function declarations. That the .c body wouldn't run on a Pi1 wouldn't matter because the Pi1 would never call the function in the first place.

MyreMylar:
Remaining tasks for this PR (I think):

- [ ] Add link to sse2neon github repository.
- [ ] Test build with and without setup.py flag on Pi4 to check it actually works.
- [ ] Add test for BLEND_PREMULTIPLIED to surface tests.

There is other, related work in the first comment in this PR and in the discussions with @illume and @nthykier  but I think those things would be better in their own pull requests to avoid this one getting any more unwieldy than it is.

nthykier:
Sounds good to me.   :)

@MyreMylar: How much speed up do you see with MMX/SSE2? I tried to optimize the RGBA_PREMULT version of image_tostring with SSE4 and I am stuck at a factor 2 improvement.  If that is in the same range as SSE2, then I am not sure SSE4 will buy anything considerable at the moment.

setup.py
nthykier: > So, the way forward for this PR is to remove the setup.py changes, and to require people to add the CFLAGS themselves on build?

I had a look at numpy and I _think_ they solve this problem by using function attributes.  But then, numpy seems to generate half their performance critical code from templates, so I am not sure it is trivial to apply directly to our situation.

MyreMylar:
> Sounds good to me. :)
> 
> @MyreMylar: How much speed up do you see with MMX/SSE2? I tried to optimize the RGBA_PREMULT version of image_tostring with SSE4 and I am stuck at a factor 2 improvement. If that is in the same range as SSE2, then I am not sure SSE4 will buy anything considerable at the moment.

Comparing 2.0.0dev8 vs this branch on my pygame_gui general test on my PC I see about a 5-6x speed up:

    dev8 frame time:           0.04 / about 25 FPS
    this branch frame time:    0.006 / about 150 FPS

Though I don't know how replicable that will be across a range of platforms and use cases. I'm not sure the Pi is seeing much boost at all - though it's very hard to tell as I'm getting performance impacting console error spamming there in Pi SDL 2 builds anyway (with this branch or not) and various visual errors (with this branch or not).

test/surface_test.py
```diff
@@ -2283,6 +2283,34 @@ def test_blit_blend_rgba(self):
                     % (dst.get_at((x, y)), tst.get_at((x, y)), x, y),
                 )
 
+    def test_blit_blend_premultiplied(self):
```
illume, line 2286: A couple of ideas to improve this test.

1. what is the formula it is supposed to hold to? Then the python formula can be easily compared to the other one.
2. edge cases for color could be used. 0, 2, 9, 127, 128, 200, 255.
3. edge case surfaces could be called. 0, 4, 9, 16, 17, 67, 69

MyreMylar:
Spent some time on this again today. 

- [ ] Added more, and stricter tests to cover more blending situations and in case anyone else writes a similar blending function in the future. (lets see if they pass)
- [ ] Fixed the setup.py flag to enable arm neon being broken. I suspect the NEWBUF flag I originally modelled it after may be broken as well.
- [ ] optimised the non-intrinsic version of pre-multiply blend a bit based on how the intrinsic version works. Should save time blending in some common situations.

MyreMylar:
If it all passes I'll run another round of testing on the raspberry pi with the neon optimisations on and off to see if I can get a better sense of any performance changes.

test/surface_test.py
MyreMylar: I have added tests to cover theses cases and a couple of others with surface formats (alpha on or off).

MyreMylar:
OK, managed to get some more testing in on the Raspberry Pi 4 today.

1. There are some other bugs with alpha blending on the Pi, but they are not related to this PR so I'll raise a separate issue about them shortly after gathering a little more info.
2. Using this PR I see a small, 10% or so speedup over 2.0.0.dev8, _without_ the `-enable-arm-neon` option set. Which is likely from the small changes I made to the base implementation.
3. Using this PR with `-enable-arm-neon` when building pygame I see an approximate 3x speed up using the same pygame_gui performance test setup as on the other platforms. Not as good as SSE2/MMX - but not bad for a freebie port.

I think I'm, personally, done messing around with this now, unless anyone else has anything that needs resolving.

setup.py
illume: ... continuing this discussion here: [CPU specific modules (neon, SSE4, etc)](https://github.com/pygame/pygame/issues/1723)

PR #1715, opened 2020-05-09
Add a SSE4.2 optimized version of image_tostring for some 32bit surfaces

nthykier:
Add a SSE4.2 optimized version of tostring_surf_32bpp that can speed
up operation on surfaces that satisfy certain restrictions such as:

 * have a width of at least 4 pixels
 * have masks that are at most 0xff (256)
 * do NOT use colorkeys
 * do NOT have any losses (or completely "loss out" a value)

When compiling, remember to set "-msse4.2" in CFLAGS a la:

```sh
CFLAGS="-msse4.2" python3 setup.py build
```

(Otherwise, the relevant code will not be included in the binary).

As for speed:

 * SSE4.2: 0.183ms - 0.195ms (0.189ms ± 0.006ms)
 * master: 1.606ms - 1.610ms (1.607ms ± 0.002ms)

Which is roughly a factor 8 improvement.

Benchmarked using the following code:

```python
import timeit
import pygame
import statistics

bigsurf = pygame.Surface((500,500))

def test_this_tostring():
    rgba_buf = pygame.image.tostring(bigsurf, "RGBA")

test_this = test_this_tostring

if __name__ == '__main__':
    number = 1000
    res = timeit.repeat('test_this()', number=number, setup="from __main__ import test_this")
    min_val = min(res) * 1000 / number
    max_val = max(res) * 1000 / number
    avg_val = (sum(res) * 1000 / len(res)) / number
    std_dev = statistics.stdev(res)
    print("%.3fms - %.3fms (%.3fms ± %.3fms)" % (min_val, max_val, avg_val, std_dev))

```

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -30,6 +30,10 @@
 
 #include "doc/image_doc.h"
 
+#if __SSE4_2__
+#include <emmintrin.h>
+#endif
+
 #if IS_SDLv1
 #include "pgopengl.h"
 #endif /* IS_SDLv1 */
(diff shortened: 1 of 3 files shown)


nthykier:
Updated to support the speed for any surface of `width >= 4`.

MyreMylar:
I'm sure you know this @nthykier but much as with the Arm neon code we will need to eventually pull these functions out into a separate .c file to include these optimisations by default in a pre-compiled python wheel. See issue:

https://github.com/pygame/pygame/issues/1723

Otherwise including the `-msse4.2` option generally could potentially build code in other imported headers with MSSE4 optimisations that won't run on platforms that don't support MSSE4.

illume:
I guess we don't know if we need to separate functions into separate files yet? Potentially, we just compile the whole module with different CPU instructions -- so no separate files needed.

illume:
Wow, I didn't realise how widely spread SSE4.2 is now. It's at 97% on the valve hardware survey, and SSE2 is at 100%.
https://store.steampowered.com/hwsurvey 
Click "Other Settings" accordion near the bottom.

MyreMylar:
> I guess we don't know if we need to separate functions into separate files yet? Potentially, we just compile the whole module with different CPU instructions -- so no separate files needed.

Yes, I suppose it would depend on whether everything we import is also guarding their platform specific code properly at runtime? It may be that we can just enable these flags all the time and they will just work on platforms that don't support the optimisations. 

I suppose what we really need is to build test wheels on a Pi4 and on a SSE4 supporting PC, copy them over to a Pi 1 and a old PC and see if they work.

nthykier:
I used the function attribute to automatically enable SSE4.2 compilation of the critical code piece to ensure it was compiled with optimization even with setting `CFLAGS=-msse4.2` (forgot to update the comment though).  It should ensure that optimization is enabled by default (on all x86 and x86_64 machines) and then activates assuming the runtime support it.

Issue #1718, opened 2020-05-11
SDL1 build broken


illume:
I think we should keep supporting SDL1 for pygame 2 at least, and decide after what to do. It's good for testing sometimes, and there is still a demand for SDL1 support on some older or exotic platforms (especially embedded ones where SDL2 lacks framebuffer support).

illume:
#1719 builds a SDL1 build on travis CI for us now.

PR #1719, opened 2020-05-11
SDL1 build fixes

illume:
For https://github.com/pygame/pygame/issues/1718

- fixes a few SDL1 build problems
- makes the travis CI linux python 3.5 worker use SDL1 (so we have one robot catching SDL1 issues). Note, that the manylinux build also tests python 3.5 with SDL2 on linux.

### DIFF
--- a/src_c/pgcompat.h
+++ b/src_c/pgcompat.h
@@ -195,7 +195,7 @@
 
 
 #if defined(SDL_VERSION_ATLEAST)
-#if !(SDL_VERSION_ATLEAST(2, 0, 5))
+#if (SDL_VERSION_ATLEAST(2, 0, 0)) && !(SDL_VERSION_ATLEAST(2, 0, 5))
 /* These functions require SDL 2.0.5 or greater.
 
   https://wiki.libsdl.org/SDL_SetWindowResizable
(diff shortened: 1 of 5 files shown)


src_c/pgcompat.h
```diff
@@ -195,6 +195,11 @@
 
 
 #if defined(SDL_VERSION_ATLEAST)
+
+#if !(SDL_VERSION_ATLEAST(2, 0, 0))
+/* SDL 1.x does not have SDL_Window */
+typedef void SDL_Window;
```
nthykier, line 201: If this solves a compile error, I suspect it replaces it with a lot of "unused variables" warnings (as I presume we cannot call anything that sets that variable nor pass it own to any function to consume that value).  If SDL 1 support is a long term goal for us, then I think we ought to keep that part of the code base for warning free as well and as such I would "vote" for a "proper warningfree fix" here.

nthykier:
Looks good to me except for the `typedef void SDL_Window;` part.  We should definitely have an SDL1 CI build for as long as we support SDL1. :)

src_c/pgcompat.h
illume: Good point. Thanks.

illume:
The typedef has been removed now.

PR #1721, opened 2020-05-12
Fix midi_to_ansi_note

GroovyUnicyclist:
Previously did not display the right note for some notes. Fixed so that notes array starts at C instead of A and adjusted offset accordingly.

### DIFF
--- a/src_py/midi.py
+++ b/src_py/midi.py
@@ -724,8 +724,8 @@ def midi_to_ansi_note(midi_note):
     >>> midi_to_ansi_note(108)
     'C8'
     """
-    notes = ['A', 'A#', 'B', 'C', 'C#', 'D', 'D#', 'E', 'F', 'F#', 'G', 'G#']
+    notes = ['C', 'C#', 'D', 'D#', 'E', 'F', 'F#', 'G', 'G#', 'A', 'A#', 'B']
     num_notes = 12
-    note_name = notes[int(((midi_note - 21) % num_notes))]
+    note_name = notes[int(((midi_note - 24) % num_notes))]
     note_number = (midi_note - 12) // num_notes
     return '%s%s' % (note_name, note_number)


GroovyUnicyclist:
> Hi!
> 
> Can you provide an example of a call to `midi_to_ansi_note()` that produces the wrong result, or if you are feeling adventurous, update the current test so it covers these cases:
> 
> Here is the current test in `test/midi_test.py`:
> 
> ```
> def test_conversions(self):
>     """ of frequencies to midi note numbers and ansi note names.
>     """
>     from pygame.midi import frequency_to_midi, midi_to_frequency, midi_to_ansi_note
> 
>     self.assertEqual(frequency_to_midi(27.5), 21)
>     self.assertEqual(frequency_to_midi(36.7), 26)
>     self.assertEqual(frequency_to_midi(4186.0), 108)
>     self.assertEqual(midi_to_frequency(21), 27.5)
>     self.assertEqual(midi_to_frequency(26), 36.7)
>     self.assertEqual(midi_to_frequency(108), 4186.0)
>     self.assertEqual(midi_to_ansi_note(21), "A0")
>     self.assertEqual(midi_to_ansi_note(71), "B4")
>     self.assertEqual(midi_to_ansi_note(82), "A#5")
>     self.assertEqual(midi_to_ansi_note(83), "B5")
>     self.assertEqual(midi_to_ansi_note(93), "A6")
>     self.assertEqual(midi_to_ansi_note(94), "A#6")
>     self.assertEqual(midi_to_ansi_note(95), "B6")
>     self.assertEqual(midi_to_ansi_note(96), "C7")
>     self.assertEqual(midi_to_ansi_note(102), "F#7")
>     self.assertEqual(midi_to_ansi_note(108), "C8") 
> ```

Ok it turns out I had an older version where this wasn't fixed yet. Although I could argue my way makes more intuitive sense since the octaves change on the C's, my fix doesn't change anything about the way this function works.

MyreMylar:
Ah, OK :) No worries.

I don't think it particularly matters how this function calculates the note for the end user (though my musical consultant agrees with you).

On the principle that we probably shouldn't change anything that's working, we should probably close this. But thank you for the pull request anyway! This was only my second attempt at reviewing a pull request - and I also learned a bit about musical notation :)

PR #1724, opened 2020-05-12
small speedup: Use Surface.blits in sprite.Group.draw.

illume:
Should be able to compare with this benchmark:
```
python3 examples/testsprite.py -noupdate_rects 10000
```


illume:
Seems there was a communication problem between travis-ci and github.  The tests all passed: https://travis-ci.org/github/pygame/pygame/builds/686228980

Issue #1429, opened 2019-10-16
Need Linux and macOS wheels for Python 3.8


illume:
- [x] Travis linux: https://travis-ci.community/t/add-python-3-8-support/5463 (available on xenial/bionic)
- [x] Travis MacOS: based on homebrew. Maybe need to save old Python3.7 build. (homebrew not upgraded to 3.8 yet) https://github.com/Homebrew/homebrew-core/pull/45337
- [x] AppVeyor are adding it in their next images update: appveyor/ci#3142 (not yet available, using a script which downloads python3.8 if not in the image.)
- [x] manylinux build
- [x] decide if doing a 1.9.6 build - ~~No. Concentrating on pygame 2 instead.~~ Yes.
- [x] decide if doing 2.0.0.dev6 release with 3.8 (depends on what is ready)
- [x] Windows python 3.8 pygame 1.9.6 builds uploaded.
- [x] ~~It seems the old travis Mac build worked for 3.6... so it should be possible to do a mac wheel.~~ the old mac ci based on homebrew isn't working now. Not going to invest effort into pygame 1.9.6 builds with SDL1, which doesn't work well on mac.
- [ ] manylinux 3.8 build for pygame 1.9.6 should be possible

illume:
travis ci python3.8 https://github.com/pygame/pygame/pull/1450

illume:
New docker images with manylinux1_base_x86_64 and manylinux1_base_i686 with python 3.8 are pushed.

mgedmin:
Appeyor has 3.8 support now!

illume:
Windows python 3.8 pygame 1.9.6 builds uploaded. The old appveyor config is broken it seems.

Old travis 3.6 build for pygame 1.9.6 is still working. https://travis-ci.org/pygame/pygame/jobs/610253803 So the new 3.8 python config could be merged into branch 1.9.6-python3.8 probably.

Similarly with the manylinux build... it should be not so hard to just build the image again with the branch 1.9.6-python3.8.

cclauss:
What is keeping this issue from being closed?

clbarnes:
> What is keeping this issue from being closed?

Linux and MacOS wheels.

cclauss:
@illume Could you please change the title of this issue to something like:
`Need Linux and macOS wheels for Python 3.8`

markuspi:
[...]

PR #1727, opened 2020-05-13
Travis CI: Test Py38 on Ubuntu Bionic

cclauss:
Related to #1429

1. `sudo apt-get install musescore-soundfont-gm` fails on Ubuntu 20.04 so replace with `timgm6mb-soundfont`

2. /usr/lib/gcc/x86_64-linux-gnu/9/include-fixed/bits/statx.h:38:25: error: missing binary operator before token "("  https://travis-ci.org/github/pygame/pygame/jobs/688841921#L1375


illume:
Thanks.

It would be good to have one builder using the new ubuntu. This way we can test with different versions of dependencies (like libsdl and such). Keeping other python versions in there lets us test with those older python versions too.

The builds for python 3.8 for pygame 1.9.6 could be made, but that would have to be done manually by someone - not with our travis CI config. I guess pygame 2 will still take a while to do, and more and more people are encountering issues building from source. So we should probably do that.

I'll squash-merge this into another branch, and do some more modifications on it in another PR.

nthykier:
@illume Any news on the branch you were talking about or should we go ahead and merge this one after it was squashed and rebased by @cclauss ?

illume:
I'll have a look at this again after I push a release out in the next days :)

nthykier:
> I'll have a look at this again after I push a release out in the next days :)

Ping?

illume:
Thanks for the ping.

illume (changes requested):
- Giving us good coverage of different systems and dependencies should be a goal for the CI. It seems Xenial reaches end of support next year in April 2021. https://wiki.ubuntu.com/Releases It would be good to cover these supported versions: (16.4, 18.4, 20.4, 20.10/Xenial, Bionic, Focal, Groovy). Keeping the older builds is required (because it helps us test with old versions of dependencies compared to dev machines which are often running on the latest). Running that ARM build on Xenial was required (maybe it still is). Because it would otherwise run on x86.

- I haven't tested the config change at the bottom of the travis.yml (near deploy). What is the reason for the change?

cclauss:
That change was made to fix a Travis Build Config Validation issue...
[Config - Build #3704 - pygamepygame - Travis CI.pdf](https://github.com/pygame/pygame/files/5178017/Config.-.Build.3704.-.pygamepygame.-.Travis.CI.pdf)

illume:
Makes sense, thanks :)

illume:
`skip_cleanup: true` can be removed now since [cleanup: false is the default](https://blog.travis-ci.com/2019-08-27-deployment-tooling-dpl-v2-preview-release).

illume (approved):
👍 thanks

cclauss:
> skip_cleanup: true can be removed now since cleanup: false is the default.

The merge to master undid that.
https://travis-ci.org/github/pygame/pygame/builds/725473713/config Click `Build config validation`

PR #1728, opened 2020-05-13
Constants added to SDL in 2.0.4 were not correctly ifdef'd out

MyreMylar:
Useful list of SDL constants by version number here to check against pygame's:

https://godoc.org/github.com/veandco/go-sdl2/sdl

This is related to the build on piwheels failing in python 3.4 here:
https://github.com/piwheels/packages/issues/114
http://paste.debian.net/1146529/


MyreMylar:
Could also use the:

`SDL2_AUDIODEVICE_SUPPORTED`

define from `pgcompat.h`.

bennuttall:
👍

PR #1729, opened 2020-05-14
Some type hint fixes

MightyJosip:
- Surface methods now accepts correct Rect values
- key.set_repeat doesn't require optional values any more
- Added support for Rect.\_\_getitem\_\_()
- Added support for len() function and iterating over sprite groups


MyreMylar:
I think these lines in sprite.pyi:

    class Sprite:
        image: Surface
        rect: Rect
        def __init__(self, *groups: Group) -> None: ...
        def update(self, *args, **kwargs) -> None: ...
        def add(self, *groups: Group) -> None: ...
        def remove(self, *groups: Group) -> None: ...
        def kill(self) -> None: ...
        def alive(self) -> bool: ...
        def groups(self) -> List[Group]: ...

definitely need Group switching to AbstractGroup. Otherwise they won't accept the other types of sprite group.

Also, this is more debatable, but should:

    image: Surface
    rect: Rect

be Optional[Surface] and Optional[Rect]? By default pygame doesn't create these attributes, it just treats them _as if_ they exist in the other functions so its sort of like they are initialised to None without that actually being the case.

Looks like font.pyi needs changing from this:

    class Font(object):
        def __init__(self, name: Union[str, None], size: int) -> None: ...

to something like:

    class Font(object):
        def __init__(self, filename: Union[str, io.IOBase, None], size: int) -> None: ...

And a similar change for image.load in image.pyi:

    def load(filename: Union[str, io.IOBase], namehint: Optional[str] = "") -> Surface: ...

MightyJosip:
@MyreMylar Thank you

MyreMylar:
I also noticed in Surface.pyi:

    class Surface(object):
        _pixels_address: int
        @overload
        def __init__(
            self,
            width_height: Tuple[float, float],
            flags: int = ...,
            depth: int = ...,
            masks: Optional[_ColorInput] = ...,
        ) -> None: ...
        @overload
        def __init__(
            self,
            width_height: Tuple[float, float],
            flags: int = ...,
            surface: Surface = ...,
        ) -> None: ...

width_height is in pixels so should probably be:

    class Surface(object):
        _pixels_address: int
        @overload
        def __init__(
            self,
            size: Tuple[int, int],
            flags: int = ...,
            depth: int = ...,
            masks: Optional[_ColorInput] = ...,
        ) -> None: ...
        @overload
        def __init__(
            self,
            size: Tuple[int, int],
            flags: int = ...,
            surface: Surface = ...,
        ) -> None: ...

Also, I was pondering if it is correct to provide two @overload variants but no regular signature in pygame's case where type checkers can't find the regular python signature? Wondering if this was what might have been baffling LGTM in stuntcat yesterday.

MightyJosip:
Surface was already made by someone long ago else so I didn't edit it too much and that is probably the reason why it has most bugs (together with rect (also done before), and sprite (done by me)). About int/float you can freely pass floats in function, it will just truncate decimal part, but it won't raise any error. Overload needs to be there for PyCharm (otherwise it wouldn't accept the second constructor, but PyCharm itself has some problems with overloaded type hints, for now I will leave it the way it is, until they do something (https://youtrack.jetbrains.com/issue/PY-42137 and https://youtrack.jetbrains.com/issue/PY-42229)

Issue #85, opened 2011-09-14
Add code to gl_cube example showing how to create a 3.2 or 4.1 Core Profile context


illume:
Originally reported by: **René Dudfield (Bitbucket: [illume](https://bitbucket.org/illume), GitHub: [illume](https://github.com/illume))**

----------------------------------------

Need to investigate how to create a 3.2 or 4.1 Core Profile context in pygame's set_display_mode.

Asked on the SDL mailing list if it is possible with SDL 1.2.

Need to report back findings on PyOpenGL mailing list.

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/85

MyreMylar:
Is this issue to support more modern OpenGL versions on pygame 1 or just to support setting the Ope GL version in pygame more generally?

From what I can tell SDL 2 supports this fairly straightforwardly, but no support for it has ever been added to SDL 1.2.

**Setting OpenGL version in SDL2:**

```
SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 2);
SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
```

Is the library going to let users set this when using OpenGL, in the normal way that you can set SDL variables, or is it going to do it for them as this issue implies. If it's doing it for them, don't we still need some way for them to select a version anyway (e.g. 2.1, 3.2 or 4.1).

illume:
There's a `pygame.display.gl_set_attribute` now (and a get).
See it documented here: docs/_sources/ref/display.rst.txt

It's not used in examples/ though.

robertpfeiffer:
So what's needed for this is an example? It looks like @einarf is already using the feature in production.

einarf:
Yup. Works like a charm cross platform (win/osx/linux) : https://github.com/moderngl/moderngl-window/blob/de6b5c0134ef078984418c0e87f8ff04e21eb2ce/moderngl_window/context/pygame2/window.py#L30-L35

MyreMylar:
> So what's needed for this is an example? It looks like @einarf is already using the feature in production.

Add something like this to the start of `examples/glcube.py` 

    def main():
        "run the demo"
        # initialize pygame and setup an opengl display
        pg.init()
    
[...]

PR #1730, opened 2020-05-14
pygame.examples.glcube updated to use 'modern' API.

MyreMylar:
My attempt to update the gl cube example to show how you can switch the context profile.

So it actually works I also included a modern gl version of the gl cube as close as I could make it to the original.

This finishes fixing issue #85

### DIFF
--- a/examples/glcube.py
+++ b/examples/glcube.py
@@ -30,8 +30,6 @@ import ctypes
 try:
     import OpenGL.GL as GL
     import OpenGL.GLU as GLU
-    import OpenGL.GL.shaders
-    import OpenGL.arrays.vbo as vbo
     from numpy import array, dot, eye, zeros, float32, uint32
 except ImportError:
     print("The GLCUBE example requires PyOpenGL & numpy")
--- a/examples/glcube.py
+++ b/examples/glcube.py
@@ -24,18 +24,22 @@ Keyboard Controls
 
 """
 import pygame as pg
+import math
+import ctypes
 
 try:
     import OpenGL.GL as GL
     import OpenGL.GLU as GLU
+    import OpenGL.GL.shaders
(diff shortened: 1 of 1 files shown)


robertpfeiffer:
Looks good. I ported glcube to modernGL, but your example seems to be the better way to go.

MyreMylar:
> OpenGL.arrays.vbo

I looked at the source code for this and it imports ctypes:
https://bazaar.launchpad.net/~mcfletch/pyopengl/trunk/view/head:/OpenGL/arrays/vbo.py#L38

So I'm not sure we'd be gaining anything by switching - it seems to be pretty much the same code in a wrapper. Not an openGL expert though.

> I wonder if we should just remove the old OpenGL style code in here so it's easier to read? Then the _modern parts could be removed.

The reason I kept the old parts is because the original point of updating the gl cube example was to show that you could switch the core context profile - in these lines:

    gl_version = (2, 0)  # GL Version number (Major, Minor)

    # By setting these attributes we can choose which Open GL Profile
    # to use, profiles greater than 3.2 use a different rendering path
    pg.display.gl_set_attribute(pg.GL_CONTEXT_MAJOR_VERSION, gl_version[0])
    pg.display.gl_set_attribute(pg.GL_CONTEXT_MINOR_VERSION, gl_version[1])
    pg.display.gl_set_attribute(pg.GL_CONTEXT_PROFILE_MASK,
                                pg.GL_CONTEXT_PROFILE_CORE)

..and I noticed that switching it higher than 3.1 didn't work because of the change between old gl and modern gl. If we remove the old style code, won't we just have the same problem in the other direction, where if you set the profile below 3 it won't work?

> Probably could use a few doc strings/comments in those transform functions.

Yes, I will add some.

PR #1733, opened 2020-05-14
Add an __all__ variable to constants.c

MyreMylar:
An experiment to see if this will fix the LGTM errors.


illume:
What was the result of your testing? Does this help any tool?

- LGTM
- pylint
- python help(pygame), help(pygame.locals)
- jedi
- pycharm

I expect that some analyze the python code statically, so that the `__all__` would need to also be static, not dynamic?

MyreMylar:
> What was the result of your testing? Does this help any tool?
> 
> * LGTM
> * pylint
> * python help(pygame), help(pygame.locals)
> * jedi
> * pycharm
> 
> I expect that some analyze the python code statically, so that the `__all__` would need to also be static, not dynamic?

From my testing it seems that pylint can pick up the `__all__` variable the way I have defined it here. Most of the alerts and errors we got previously have been covered by the .pyi files in pylint, jedi & PyCharm.

using 2.0.0.dev8 if you do:

`pylint --extension-pkg-whitelist=pygame --allow-wildcard-with-all y test.py`

on a file called `test.py` containing:

    """
    Doc
    """
    from pygame.constants import *
    print(K_END)

you will get:

    test.py:4:0: W0401: Wildcard import pygame.constants (wildcard-import)


But with this pull request compiled you won't. `allow-wildcard-with-all` is disabled by default though so I guess the more reliable way to fix pylint would be for us to never use wildcard imports in pygame and instead import each constant individually into pygame's `__init__.py` and `locals.py`

LGTM is a mystery to me.

MyreMylar:
On LGTM,

I don't know how it is checking for the `__all__` variable for its alerts, or if the incremental check we get here on pull requests works the same as a full scan (which seem to happen on a slower schedule in my experience). It's possible the version of all I generate in the c file here is not what  the LGTM checker is expecting, but without knowing how it checks it's hard to know. I could try emailing the LGTM team at GitHub?

I suspect, as with pylint, the full-proof way to eliminate these alerts is to directly import each constant individually.

MyreMylar:
I'll guess I'll try emailing semmle (creators of LGTM) about this just to see what they say, unless anyone has a better idea?

MyreMylar:
I have now tried emailing.

Issue #1707, opened 2020-05-04
Allow screen blanking while pygame is running


jfbauer432:
When a pygame program is running, it disables the systems screen blanking (likely by preventing DPMS power off).  I would like a way to disable that feature sometimes (e.g. when not in fullscreen)

pygame 1.9.6
x11 video driver
Ubuntu 19.10

PR #1734, opened 2020-05-14
display: Add support for controling screensaver state. part 1/2

nthykier:
... on platforms that support it with SDL2.

SDL2 does not provide any visible feedback on whether the controls
have an effect and may "silently" say it changed the state without
actually doing anything (except update the boolean variable, which
`SDL_IsScreenSaverEnabled` returns).  Accordingly, we document that
these functions may be unreliable on certain platforms without any
visible clue.

For SDLv1, we just raise a `TypeError` as these functions are not
supported there.

Closes: #1707

https://wiki.libsdl.org/SDL_HINT_VIDEO_ALLOW_SCREENSAVER


buildconfig/pygame-stubs/__init__.pyi
```diff
@@ -69,4 +69,6 @@ def encode_file_path(
     obj: Any, etype: Optional[Type[Exception]] = UnicodeEncodeError
 ) -> bytes: ...
 def register_quit(callable: Callable) -> None: ...
+def get_screensaver_enabled() -> bool: ...
```
MightyJosip, line 72: Nice changes, didn't know it is even possible to do something like this with SDL :+1:
just a note can you move this 2 functions from \_\_init\_\_.pyi to display.pyi (because with this it would mean that you call this functions with `pygame.get_screensaver_enabled()`, and as far as I have understood it should be called with `pygame.display.get_screensaver_enabled()`

nthykier: Thanks for catching that.  I intended to use `display.pyi` but apparently managed to write it in the wrong file. :)

docs/reST/ref/display.rst
```diff
@@ -571,4 +571,50 @@ required).
...
+   this function will raise an exception.
+
+   If the screen has been disabled due to this function (or by default
+   in SDL), it will automatically be re-enabled when
+   :func:``pygame.quit()` is called.
+
+   It is possible to influence the default value under SDL 2.0.2 or
+   later via the environment variable ``SDL_HINT_VIDEO_ALLOW_SCREENSAVER``,
+   which can be set to either ``0`` (disable) or ``1`` (enable).
+
```
MyreMylar, line 608: Could add a note here that `SDL_HINT_VIDEO_ALLOW_SCREENSAVER` defaults to False, so by default screen savers are disabled while pygame applications are running.

illume:
I made a few doc changes, and am making a few changes to display.c now.

docs/reST/ref/display.rst
```diff
@@ -573,47 +573,48 @@ required).
...
    .. note:: Disabling screensaver is subject to platform support.
              When platform support is absent, this function will
              silently appear to work even though the screensaver state
              is unchanged.  The lack of feedback is due to SDL not
              providing any supported method for determining whether
              it supports changing the screensaver state.
+             ``SDL_HINT_VIDEO_ALLOW_SCREENSAVER`` is available in SDL 2.0.2 or later.
+
+   .. versionadded:: 2.0 This function is only available with SDL2. Under SDL1,
+   this function will raise a NotImplementedError exception.
```
nthykier, line 616: The code currently uses `TypeError`.  As I understand it, it is the correct error in this case as the function will never be implemented for SDLv1 https://docs.python.org/3/library/exceptions.html#TypeError


> exception TypeError
> 
>     Raised when an operation or function is applied to an object of inappropriate type. The associated value is a string giving details about the type mismatch.
> 
>     This exception may be raised by user code to indicate that an attempted operation on an object is not supported, and is not meant to be. If an object is meant to support a given operation but has not yet provided an implementation, NotImplementedError is the proper exception to raise.
> 
>     Passing arguments of the wrong type (e.g. passing a list when an int is expected) should result in a TypeError, but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a ValueError.
> 


vs.



> exception NotImplementedError
> 
>     This exception is derived from RuntimeError. In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added.
> 
>     Note
> 
>     It should not be used to indicate that an operator or method is not meant to be supported at all – in that case either leave the operator / method undefined or, if a subclass, set it to None.
> 
>     Note
> 
>     NotImplementedError and NotImplemented are not interchangeable, even though they have similar names and purposes. See NotImplemented for details on when to use it.

illume: I guess since it's a hint, we can just have the function do nothing under SDL1? This would make it easier to use from user code that wanted to support pygame 2 and 1.9.x.

nthykier: That is certainly also an option. :)  For the "getter", should we then emulate SDL2 behaviour or just always return a fixed boolean value?

illume: Yeah, cool. I think emulating SDL2 behaviour is a good idea.

illume:
This may just be me, but I think the word 'enable' is misleading here?

`SDL_EnableScreenSaver` and `set_screensaver_enable` when read by themselves suggest that the screensaver will be turned on right then. Not that the screen saver will kick in after 10 minutes of inactivity (or whatever the setting is).

How about we follow the environment variable name a bit? 'Allow' from `SDL_HINT_VIDEO_ALLOW_SCREENSAVER` seems a better word choice than `enable` as used in the SDL function names.

- `display.set_allow_screensaver`
- `display.get_allow_screensaver`

https://wiki.libsdl.org/SDL_HINT_VIDEO_ALLOW_SCREENSAVER
https://wiki.libsdl.org/SDL_EnableScreenSaver
https://wiki.libsdl.org/SDL_DisableScreenSaver
https://wiki.libsdl.org/SDL_IsScreenSaverEnabled

nthykier:
Fine with me. :)

nthykier:
@MyreMylar I pinged you for a review because you had a `changes requested` that have not been resolved.

illume:
Ok, thanks :) I'll work on renaming now.

illume:
Merging into `sdl2-screensaver-support` branch so I can continue.

PR #1739, opened 2020-05-15
sysfont: Update windows, and ubuntu xenial font names and tests.

MyreMylar:
Trying to close off this issue here:

https://github.com/pygame/pygame/issues/179

Someone has already done the code on windows to automatically collect the installed system fonts, but the list in create aliases was out of date so I went through it and added standard fonts from more recent versions of windows.

I also noticed we weren't really testing this module, possibly for good reason, but I thought I'd try some basic multi-platform tests anyway.

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -245,13 +245,16 @@ def create_aliases():
     alias_groups = (
         ('monospace', 'misc-fixed', 'courier', 'couriernew', 'console',
          'fixed', 'mono', 'freemono', 'bitstreamverasansmono',
-         'verasansmono', 'monotype', 'lucidaconsole', 'consolas'),
+         'verasansmono', 'monotype', 'lucidaconsole', 'consolas',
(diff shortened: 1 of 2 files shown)


MyreMylar:
Hmm the default xenial font might be called 'ubuntumono' instead.

MyreMylar:
Looks like Xenial on Travis has these 7 fonts installed:

-  'dejavusans',
-  'dejavuserif',
-  'dejavusansmono',
-  'liberationsans',
-  'liberationserif',
-  'liberationmono'
-  'liberationsansnarrow',

I added the top 6 to our alias sets as well as the selection from windows and the default 'ubuntu' font.

All the new tests are passing now on CI, they are a little basic but it's hard to get much of a baseline when the number and which of default fonts are installed on each system varies so widely, particularly on the varied linux platforms. At least we have some coverage with this and should notice if we test on a platform that has no fonts at all in the alias groups.

Will switch from draft for any other ideas.

PR #1770, opened 2020-05-17
WIP: MacOS clang failed to build the alphablit SSE stuff.

illume:
Related: 
- https://github.com/pygame/pygame/pull/1711
- https://github.com/pygame/pygame/pull/1715

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -25,6 +25,12 @@
 #define NO_PYGAME_C_API
 #include "_surface.h"
 
+#if PG_COMPILE_SSE4_2
+#include <emmintrin.h>
+/* SSSE 3 */
+#include <tmmintrin.h>
+#endif
+
 #ifdef PG_ENABLE_ARM_NEON
 // sse2neon.h is from here: https://github.com/DLTcollab/sse2neon
 #include "include/sse2neon.h"


MyreMylar:
Curious to know what errors clang produced?

illume:
It looks like the error is happening with SDL1 and clang.
```c
src_c/alphablit.c:1101:5: error: use of undeclared identifier '__m128i'
    __m128i src1, dst1, mm_alpha, mm_zero, mm_alpha2, multmask2_128;
    ^
...
```

I guess SDL2 includes the intrinsics properly or so.

MyreMylar:
I feel like this define is slightly wrong for `alphablit.c` as it is not currently using any SSE4.2 intrinsic functions. Looking at this answer:

https://stackoverflow.com/questions/11228855/header-files-for-x86-simd-intrinsics

Perhaps it should it just be:

    #ifdef PG_ENABLE_ARM_NEON
        // sse2neon.h is from here: https://github.com/DLTcollab/sse2neon
        #include "include/sse2neon.h"
    #else 
        #include <immintrin.h>  // should pull in all supported intrinsic functions
    #endif /* PG_ENABLE_ARM_NEON */

I guess SDL1 just doesn't include the right intrinsic headers.

illume:
The intrinsics header here is only required for SDL1+clang it seems (maybe only on macos... haven't tried linux+clang). We could try including this in src_c/include/pgplatform.h when using SDL1?

I should probably try and get a macosx robot going on travis ci again.

illume:
There's a fix for this in https://github.com/pygame/pygame/pull/1832

PR #1771, opened 2020-05-17
Fix typo in mask stub

charlesej:


MightyJosip:
I fiil good because of this change :) . Thanks

Issue #251, opened 2015-07-14
pygame.display.update(Rect) updates the whole screen on Windows due to SDL2 UpdateWindowFramebuffer implementation


illume:
Originally reported by: **Richard Lange (Bitbucket: [wrongu](https://bitbucket.org/wrongu), GitHub: [wrongu](https://github.com/wrongu))**

----------------------------------------

I am using Mac OS 10.10, python 3.3, and pygame 1.9.2,. The issue is that update(rect) behaves just like update()

Some others have had the same problem:

[on stack exchange](http://gamedev.stackexchange.com/questions/83705/pygame-display-updatelist-updates-entire-screen-not-just-rectangles-in-list)

[on dreamincode](http://www.dreamincode.net/forums/topic/184546-problem-using-pygamedisplayupdate/)

(the user on dreamincode solved it by switching back to python 2.7, which doesn't seem like a sustainable solution)

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/251

MyreMylar:
***Reproduction  code:***

```
import pygame


pygame.init()

surface = pygame.display.set_mode((200, 200))

# Draw a square
pygame.draw.rect(surface, (0, 255, 0), (0, 0, 200, 200))

# update only half of it, so we see a gree rectangle
pygame.display.update(pygame.Rect(0, 100, 200, 100))

running = True
while running:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            running = False
```

This works fine on windows, needs the code testing again on mac in pygame 2 and perhaps pygame 1.9.6.

illume:
Works on 2.0.0.dev7 on Mac.

nthykier:
This works for me as well on Linux - both in 1.9.6 and 2.0.0.dev7.  

I propose we close this as solved (even if we never figured out if it was a pygame or a SDL bug)

MightyJosip:
Before we close this just to test one thing: what should be the output of this
```
import pygame


pygame.init()

surface = pygame.display.set_mode((200, 200))

# Draw a square
pygame.draw.rect(surface, (0, 255, 0), (0, 0, 200, 200))

# update only half of it, so we see a green rectangle
pygame.display.update(pygame.Rect(0, 100, 200, 0))

running = True
while running:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            running = False
```
[...]

PR #1773, opened 2020-05-17
Updating display.update() docs to reflect actual behaviour of SDL.

MyreMylar:
This is because the 'rect' parameter is treated by SDL as a suggestion rather than a rule and platform implementations are ignoring it.

Related to:
https://github.com/pygame/pygame/issues/251

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -212,10 +212,13 @@ required).
    It is more efficient to pass many rectangles at once than to call update
    multiple times with single or a partial list of rectangles. If passing a
    sequence of rectangles it is safe to include None values in the list, which
-   will be skipped.
+   will be skipped. These rectangles are a suggested optimisation to the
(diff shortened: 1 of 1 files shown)


illume:
Not sure about this one. Because the documented functionality of SDL_UpdateWindowSurfaceRects does not say it's a hint. (I saw Sam's comment).

MyreMylar:
Ah, I see you submitted a patch to SDL for this not working on windows. Does it work fine on mac & linux then? If so then we should likely just leave the docs as they are and close the original bug once the SDL patch is merged.

nthykier:
In my view, the alternative is to ask the SDL devs to update the documentation to reflect their view on that function.  If we keep the status quo, then we will end up here eventually when pygame is used on a new platform that does not support that feature again.

MyreMylar:
I'll switch this to draft for now so we don't accidentally merge it while there is uncertainty on the SDL position.

Issue #735, opened 2019-01-14
SDL2: vsync default to on


illume:
At the moment pygame+sdl2 vsync defaults to off, but we should change it to default to on to be backwards compatible with pygame+sdl1.

https://wiki.libsdl.org/SDL_HINT_RENDER_VSYNC

Also, we should perhaps expose a display.set_mode(vsync=True) option.

husano896:
How will vsync flag work in pygame+sdl1?

illume:
Good question. Maybe it will just do the current behavior, and it will be ignored. I'm not sure if SDL supports vsync in a way where it can be enabled or not.

husano896:
Previously in SDL1, OpenGL contexts are vsync'd by SDL_GL_SWAP_CONTROL

and surfaces are vsync'd by HWSURFACE | DOUBLEBUF + clock.tick().

But windowed surfaces can only use clock.tick(), which may cause screen jitters.

Probably related? #304

dlon:
The hint has no effect for me. Perhaps it only works for SDL_Renderer.

illume:
Talking about this some more, I agree it should probably be left "as is" for pygame 2. Since this is a new feature, it should be moved to pygame 2.1+ (if we do it at all).

MyreMylar:
My question; is there a way to enable vsynch at all in Pygame 2? 

The pygame 1.9 method is no longer available.

illume:
Sorry, which method was that?

MyreMylar:
In Pygame 1 you can enable v-sync (on windows) by:

1. Setting:  `os.environ['SDL_VIDEODRIVER'] = 'directx'` before you call pygame.init()
2. Setting these flags: `pygame.FULLSCREEN | pygame.DOUBLEBUF` when you call pygame.display.set_mode()
3. Don't restrict the frame rate with clock.tick(FPS).

On my computer this locks the frame rate to 60 FPS and if you scroll sprites or surfaces around they will move smoothly with no hitching. As far as I know this is the only way to enable v-sync/smooth graphic movement in pygame 1 so it's not ideal (doesn't work in windowed mode, no idea what the situation is on mac or linux) but it's better than nothing.

Step 1 is the one we can no longer do in pygame 2 as 'directx' is no longer an option for the SDL_VIDEODRIVER environment variable.

[...]

PR #1774, opened 2020-05-17
Updating display docs to describe new vsync argument to set_mode

MyreMylar:
This is related to:

https://github.com/pygame/pygame/issues/735


robertpfeiffer:
So should I add the environment variable back in then? I thought an extra keyword argument and the SDL hint are enough?

MyreMylar:
> So should I add the environment variable back in then? I thought an extra keyword argument and the SDL hint are enough?

In 2.0.0.dev8 on windows you need to use the pygame environment variable + the pygame.SCALED flag to turn vsync on and off, SCALED on it's own doesn't do it. I got the impression it also worked roughly like this on mac. 

I didn't know you had tried to remove it.

robertpfeiffer:
> I didn't know you had tried to remove it.

I thought an explicit parameter is a better idea than an environment variable. I only really put that in for quick testing purposes, it wasn't supposed to be part of the API. The SCALED branch was merged with a couple of TODO things left to do.

The explicit parameter will set vsync for both SCALED and OPENGL, but will raise an exception on software graphics. This could also be a flag in `flags`, but the extra parameter has the added option of being unspecified.

SDL_HINT_RENDER_VSYNC is already a hint for renderer that does the same thing as the pygame environment variable, when used in conjunction with SCALED.

MyreMylar:
> The explicit parameter will set vsync for both SCALED and OPENGL, but will raise an exception on software graphics.

I assume you mean 'software' graphics that don't use the SCALED flag here, not just non-openGL. Or when you use an environment variable to engage software scaling. The terminology gets confusing.

> SDL_HINT_RENDER_VSYNC is already a hint for renderer that does the same thing as the pygame environment variable, when used in conjunction with SCALED.

I tried swapping out `os.environ['PYGAME_VSYNC'] = "1"` for `os.environ['SDL_HINT_RENDER_VSYNC'] = "1"` in 2.0.0.devv8 but it doesn't have the same effect. Only setting PYGAME_VSYNC with SCALED actually seems to engage vsync. Which makes sense as it engages this bit of code here:

    if (vsync_env != NULL) {
        pg_renderer = SDL_CreateRenderer(
            win, -1, SDL_RENDERER_PRESENTVSYNC);
    }
    else {
        pg_renderer = SDL_CreateRenderer(win, -1, 0);
    }

in `display.c`. 

No objection to this behaviour being switched to a less confusing function, so long as it works. I'll switch this to draft and perhaps revise it or scrap it when a final API has been settled on.

MyreMylar:
I've updated this docs to match the SCALED Pull request here:

https://github.com/pygame/pygame/pull/1704

They should be merged after that one has been merged as they document the vsync parameter added in that PR.

MyreMylar:
I also tidied up `display.rst` a bit while I was in there as several of the chunks of pre-formatted text seemed to be overlapping their containers.

illume:
I'm not sure if we want `vsync=` there permanently as a keyword arg?

Everything else goes through the `flags`, which has a late 90s retro API feel about it. So probably keyword arguments are more `pythonic` (is that still a thing?). With the video API we experimented with using flags, and I think it works ok. They are more discoverable, especially with with tab completion editors. However the main drawback is that it takes up more than one variable - so passing around all those variables or saving them is more of a chore.

MyreMylar:
> I'm not sure if we want `vsync=` there permanently as a keyword arg?

After reading your whole comment I ended up unsure if you were in favour of keyword arguments in this instance or against :)

Anyway, this is just the documentation - the change is already in the code. I guess we could not document it but that seems odd.

illume:
TLDR; a keyword with an int seems the best, most loved option.

---

renpytom made an interesting comment: 
> I'd suggest making vsync a number. That would let make it possible to have vsync=2 (sync every 2 frames), or vsync=-1 (sync every frame, but tear if late).

[SDL_GL_SetSwapInterval](https://wiki.libsdl.org/SDL_GL_SetSwapInterval
) uses a number too: "0 for immediate updates, 1 for updates synchronized with the vertical retrace, -1 for adaptive vsync; see Remarks".

There were quite a lot of people who preferred flags, but more that liked the keyword arguments. See below.

Screenshot 2020-05-30 at 08 37 48
Screenshot 2020-05-30 at 08 37 02

illume:
I'm going to merge this into a pygame branch, and change it to int.

Issue #1741, opened 2020-05-15
_camera_videocapture.py uses .set_masks() & .set_shifts()


MyreMylar:
These methods no longer work in Pygame 2/ SDL2 so this portion of code will need to be replaced with some other method:

    def get_surface(self, dest_surf = None):
        """Returns a pygame Surface.
        """
        abuffer, width, height = self.get_buffer()
        if abuffer:
            surf = pygame.image.frombuffer(abuffer, (width, height), "RGB")

            # swap it from a BGR surface to an RGB surface.
            r,g,b,a = surf.get_masks()
            surf.set_masks((b,g,r,a))

            r,g,b,a = surf.get_shifts()
            surf.set_shifts((b,g,r,a))

            surf = pygame.transform.flip(surf, 0,1)

            # if there is a destination surface given, we blit onto that.
            if dest_surf:
                dest_surf.blit(surf, (0,0))
            else:
                dest_surf = surf
            return dest_surf

illume:
I started some new work on camera stuff in [parrotjoy](https://github.com/pygame/parrotjoy). Might be useful to anyone looking into this.

PR #1775, opened 2020-05-18
Fix camera video capture

MyreMylar:
A fix for this issue:

https://github.com/pygame/pygame/issues/1741

**Problem**
Basically the `surface.set_masks()` and `surface.set_shifts(`) functions are no longer available in SDL2 so to make the `_camera_videocapture.py` file work and display camera output we need to replace these functions.

**What I did**
The root of the problem is that camera output comes in in a different byte ordering to what we need in pygame (BGR versus RGB). So to fix it I added an extra block to the `image.frombuffer()` function that handles a 'BGR' format parameter.

**Docs**:
https://www.pygame.org/docs/ref/surface.html#pygame.Surface.set_masks
https://www.pygame.org/docs/ref/surface.html#pygame.Surface.set_shifts

This is the test code I used to verify that this works on windows:

    import pygame.camera
    
    pygame.init()
    pygame.camera.init()
    cam_list = pygame.camera.list_cameras()
    cam = None
    if cam_list:
        cam = pygame.camera.Camera(cam_list[0], (640, 480))
        cam.start()
    
    window_surface = pygame.display.set_mode((640, 480))
    
    is_running = True
    while is_running:
            if event.type == pygame.QUIT:
                is_running = False
    
        if cam is not None:
            image = cam.get_image()
            window_surface.blit(image, (0, 0))
    
        pygame.display.update()

**Bonus:** I rolled in some fixes to the camera.pyi file while I was at it.

**Remaining potential work:**

 - [x] The three functions:
    https://www.pygame.org/docs/ref/image.html#pygame.image.tostring
    https://www.pygame.org/docs/ref/image.html#pygame.image.fromstring
    https://www.pygame.org/docs/ref/image.html#pygame.image.frombuffer

    sort of share documentation of the format parameter. So, we could either add code to make 
    'BGR' a format option for all three, document that `frombuffer()` has an additional parameter or, 
    just leave it as a undocumented thing if we think it won't be that widely used?

- [x] check for any other uses of `set_masks()` and `set_shifts()` in the code base.
- [x] unit test `from_buffer()`


MyreMylar:
I think I'll just change the documentation for frombuffer() as it already has undocumented format differences from the other two functions.

buildconfig/pygame-stubs/image.pyi
```diff
@@ -13,5 +13,5 @@ def fromstring(
     flipped: Optional[bool] = False,
 ) -> Surface: ...
 def frombuffer(
-    string: str, size: Union[List[int], Tuple[int, int]], format: str
+    bytes: bytes, size: Union[List[int], Tuple[int, int]], format: str
```
illume, line 16: I don't think there is a buffer protocol type yet. https://github.com/python/typing/issues/593
I asked on Typing sig about it, maybe someone will have some good idea.

Maybe our own buffer type could start with something like this:
```
from pygame.bufferproxy import BufferProxy
Union[BufferProxy, bytes, bytearray, memoryview]
```
This doesn't work with array.array, or numpy.ndarray, but would be a good start.

buildconfig/pygame-stubs/camera.pyi
```diff
@@ -8,7 +8,7 @@ def list_cameras() -> List[str]: ...
 class Camera:
     def __init__(
         self,
-        device: Optional[int] = 0,
+        device: str,
```
illume, line 11: I guess other implementations use int? Can this be `Optional[Union[int, str]]` or does this require string here?

test/image_test.py
```diff
@@ -508,22 +508,173 @@ def test_tostring_depth_24(self):
...
+        self.assertEqual(eight_bit_surf.get_at((0, 0)),
+                         pygame.Color(255, 10, 20))
+        self.assertEqual(eight_bit_surf.get_at((1, 1)),
+                         pygame.Color(255, 255, 255))
+        self.assertEqual(eight_bit_surf.get_at((2, 2)),
+                         pygame.Color(0, 0, 0))
+        self.assertEqual(eight_bit_surf.get_at((3, 3)),
+                         pygame.Color(50, 200, 20))
+
+        rgb_buffer = bytearray([255, 10, 20,
```
illume, line 534: I guess these tests could be broken up into separate methods. The benefits being that you can run one at a time if debugging, and if one fails it's easier to see where the problem is.

src_c/image.c
```diff
@@ -1252,6 +1252,15 @@ image_frombuffer(PyObject *self, PyObject *arg)
 
         */
     }
+    else if (!strcmp(format, "BGR")) {
+        if (len != (Py_ssize_t)w * h * 3)
+            return RAISE(
+                PyExc_ValueError,
+                "Buffer length does not equal format and resolution size");
+        surf = SDL_CreateRGBSurfaceFrom(data, w, h, 24, w * 3,
+                                        0xFF << 16, 0xFF << 8,
```
illume, line 1261: I guess this might need a `SDL_BYTEORDER == SDL_LIL_ENDIAN` block?

buildconfig/pygame-stubs/camera.pyi
MyreMylar: Think this was just an error. Two code paths are here:

https://github.com/pygame/pygame/blob/16a8d3ea36ff09fc1c1d91c6492219a4bef944db/src_c/_camera.c#L1787-L1800

and here:

https://github.com/pygame/pygame/blob/16a8d3ea36ff09fc1c1d91c6492219a4bef944db/src_c/_camera.c#L1841-L1851

and it looks like both turn it into `char *dev_name = NULL;` so I think str type is correct. Also not optional because it comes before the pipe `|` character.

src_c/image.c
MyreMylar: I was going to add one but noticed it had been commented out for the RGB block above here:

https://github.com/pygame/pygame/blob/16a8d3ea36ff09fc1c1d91c6492219a4bef944db/src_c/image.c#L1240-L1254

And thought that maybe this had been removed on purpose and I should follow suit.

test/image_test.py
MyreMylar: I have done this now.

buildconfig/pygame-stubs/image.pyi
MyreMylar: I added this to the `image.pyi` file as a `_BufferStyle` type similar to the `RectStyle` type we already have.

buildconfig/pygame-stubs/camera.pyi
illume: 👍

src_c/image.c
illume: This can be tested later once the big endian test bots are setup.

buildconfig/pygame-stubs/image.pyi
```diff
@@ -1,6 +1,9 @@
 from typing import Optional, Tuple, List, Union, IO
 
 from pygame.surface import Surface
+from pygame.bufferproxy import BufferProxy
+
+_BufferStyle = Union[BufferProxy, bytes, bytearray, memoryview]
```
illume, line 6: Cool :)

docs/reST/ref/image.rst
```diff
@@ -167,12 +167,26 @@ following formats.
...
 
    Create a new Surface that shares pixel data directly from a bytes buffer.
-   This method takes the same arguments as ``pygame.image.fromstring()``, but
+   This method takes similar arguments to ``pygame.image.fromstring()``, but
    is unable to vertically flip the source data.
 
    This will run much faster than :func:`pygame.image.fromstring`, since no
    pixel data must be allocated and copied.
 
+   It accepts the following 'format' arguments:
```
MightyJosip, line 176: If this is the case can you add `_format = Literal['p', 'RGB', 'BRG', 'RGBX', RGBA', 'ARGB']` to the image.pyi and then replace `format: str` with `format: _format`. You can see the example of it in pixelcopy.pyi

MyreMylar: I have now done this for all of the to/from functions to match their code.

PR #1814, opened 2020-05-19
debugging distutils.sysconfig.expand_makefile_vars

illume:


illume:
The issue was found, we need to add this to the Setup.SDL2.in
```
PNG = -lpng
JPEG = -ljpeg
```

Issue #1403, opened 2019-10-12
MSVC compiler optimization hints


illume:
This comment https://github.com/ulfjack/ryu/pull/70#issuecomment-412168459 contains some interesting MSVC hints: 
- [x] use /O2 instead of /Ox optimization compiler flag.
- [x] use __forceinline instead of inline

MyreMylar:
Had a bash at adding the extra compiler flags that `/O2` includes over `/Ox`. According to MSDN:

https://docs.microsoft.com/en-us/cpp/build/reference/ox-full-optimization?redirectedfrom=MSDN&view=vs-2019

`/Ox` and `/O2` are basically the same except that `/O2` also includes `/GF /Gy` . 

- `GF` [reduces the size of the compiled code by eliminating duplicate strings](https://docs.microsoft.com/en-us/cpp/build/reference/gf-eliminate-duplicate-strings?view=vs-2019) - In my testing this takes the size of `pygame-2.0.0.dev9-py3.8-win32.egg` down from 6,662 KB to 6,655 KB a saving of approx 7KB.
- `Gy` [reduces the size of code by merging identical functions (such as those generated by templates) into one linker object](https://docs.microsoft.com/en-us/cpp/build/reference/gy-enable-function-level-linking?view=vs-2019). ([explained fairly well here](https://stackoverflow.com/questions/629894/function-level-linking-gy-switch-in-vc-what-is-it-good-for)) - In my testing this reduces the size of the egg file from 6,662 KB to 6,619 KB a saving of approx 43KB.

If we enable both together we get a .egg file size of 6,611 a total code size saving of approx 51KB. I can't see any downside to enabling these.

In my testing I added this to the `setup.py` file:

    # try adding extra compile args
    if sys.platform == 'win32':
        for e in extensions:
            e.extra_compile_args += ["/GF", "/Gy"]

Do we only build a 32 bit windows wheel for compatibility? I'm not sure, I'm also not sure how the mingw compiler factors in for Pygame 1/SDL1 which presumably has different compiler options to MSVC.

These stack overflow answers suggest a few methods for detecting the `clang` compiler that might be of use:

https://stackoverflow.com/questions/28641408/how-to-tell-which-compiler-will-be-invoked-for-a-python-c-extension-in-setuptool

MyreMylar:
I'm less certain about the benefits of us switching `__inline` to `__forceinline` always on MSVC. Microsoft don't seem to recommend it here:

[...]

PR #1816, opened 2020-05-19
MSVC compiler optimisation hinting

MyreMylar:
Related to this issue:
https://github.com/pygame/pygame/issues/1403

So far the code I've added to setup.py adds two flags to the extension modules compile process when the system is 'win32' and the compiler supports the two flags. These flags both serve to reduce the size of the generated c code reducing our overall .egg file size by just under 1%.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -563,7 +563,6 @@ if sys.platform == 'win32':
         """
         from distutils.errors import CompileError
         import tempfile
-        import os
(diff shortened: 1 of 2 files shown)


MyreMylar:
Also added a PG_FORCEINLINE macro that is the same as PG_INLINE on all compilers except `MSC_VER` where it means `__forceinline`.  I haven't used it anywhere yet though as Microsoft indicate that using `__forceinline` may sometimes make performance _worse_ so using it will have to be done on a case by case.

PR #1817, opened 2020-05-19
Some type hint and examples fixes

MightyJosip:
- Added type hints for operators for `pygame.Color` and `pygame.math.Vector`
- Fixed pygame.event type hints
- Fixed types in `pygame.font` (bool -> int)
- Fixed pygame.draw accepting string objects as color argument (it was added in #1676, around the same time I did type hints so it wasn't in sync)
- Fixed Sound constructor keywords
- Fixed syntax error in pygame.version
- Added flag if the class is hashable
- Fixed scrap_clipboard.py example (You can start it at least now, it doesn't work because pygame2 doesn't support scrap module)
- Fixed testsprite.py example

NOTE: If you use PyCharm you won't be able to see the changes with hashable/unhashable types, haven't checked if mypy recognize it, but I used example from https://github.com/python/typing/issues/74 (ok just checked mypy and it also doesn't accept __hash__ = None as non hashable object)


MyreMylar:
Oops, guess I missed out some `...`'s in that version.py.

What's the deal with the `bool` to `int` font changes? They would seem to be inconsistent with the docs e.g. here:

https://www.pygame.org/docs/ref/font.html#pygame.font.Font.set_bold

Is this a case of the underlying c code using ints and not casting back to python's bool type?

MightyJosip:
Docs is wrong (at least when I tested those functions

MyreMylar:
Looked over the changes and then tried it out in pygame_gui and it all seems to work well with the PyCharm inspection.

Looks like type checkers don't care actually about passing a bool into an int - though it is slightly confusing for the auto complete hints.

I think this should be merged and we can figure out any ints that should be bools later.

MightyJosip:
> Looks like type checkers don't care actually about passing a bool into an int - though it is slightly confusing for the auto complete hints.

This is because of numeric object "hierarchy". Basically it goes like float -> int -> bool -> None. So for example type hint for float will accept float, int, bool and None, int will also accept bool and None...

Issue #1328, opened 2019-09-23
MSVC /analyze static analyzer


illume:
https://docs.microsoft.com/en-us/cpp/build/reference/analyze-code-analysis

MyreMylar:
The PR above adds analyze to appveyor. There are some slight annoyances with the analysis data itself - disabling analysis warnings from system headers works, but the disabling itself generates warnings only on appveyor & some warnings are probably just failures of static analysis but overall I think the output is useful and you can search the generated warning error codes to usually get a pretty good Microsoft help page on what the warning is about, and often how to resolve it.

 `/analyze` also, unsurprisingly slows down build times. Build times on appveyor are quite good though at about 1-2 minutes each build now and only 2-3 minutes with analyze so it's probably not that big of a deal. I suspect we could set it up so analyze only runs on a few of the builds if it gets too slow.

If the initial PR gets checked in I can revist this periodically to try and squash more of the warnings.

illume:
This is merged in now. 

We can leave the issue open to address the remaining warnings it reports.

illume:
I think we should perhaps change /analyze to only run on one job.
- 14 mins (with no /analyze)
- 24 mins (with /analyze)

Recently there's been lots of PRs, commits, and such. With that when several PRs are merged it can take an hour or more to get through them. eg. 5*24 == 120 mins.

If we have to pick one, then probably we should use the most recent compiler /analyze? Since it's probably greatly improved over older versions of the compiler. Perhaps a 32bit one, and a 64bit one?

MyreMylar:
Yeah, I think maybe just leave it on one for now and maybe rotate it later if we run out of warnings. I think we can configure that on appveyor pretty easily IIRC. I'll make a PR sticking it on the 3.8 64 bit build first.

MyreMylar:
PR here:

https://github.com/pygame/pygame/pull/1867

ankith26:
Seems like this issue can be closed now, the static analyzer works well on Windows appveyor builds RN

PR #1820, opened 2020-05-19
Add /analyze flag to appveyor build and fix some reported warnings

MyreMylar:
Related to this:
https://docs.microsoft.com/en-us/cpp/build/reference/analyze-code-analysis?view=vs-2019

and this issue:
https://github.com/pygame/pygame/issues/1328

Lets see if it runs on appveyor... (It does though you need to click through to see the interesting output)

Here is a dump of the output from running it locally:
[msvc_analyze_output.txt](https://github.com/pygame/pygame/files/4651633/msvc_analyze_output.txt)


MyreMylar:
Though I might as well try and fix a few of the outputted warnings while I was doing this.

First up, a change to a macro that was spewing:

`warning C6246: Local declaration of '_module' hides declaration of the same name in outer scope.` over many places in the code base.

MyreMylar:
Noticed that we had `INT_CHECK(x)` defined in multiple places in the code so moved it to  pgcompat.h and made sure we were using it in all the places that were previously doing a `PyLong_Check(x) || PyInt_Check(x)`

This was generating an Appveyor warning about both sides of an or being equal to the same thing (which they previously were on python 3).

MyreMylar:
Don't want to make too many changes across too many files in one PR so I'll stop fiddling here I think and revisit with batches of fixes over time.

MyreMylar:
Looks like the /external flags were only added midway through the life of MSVC 2017 and python only updated to MSVC 2019 after Python 3.8.3, so I added an extra appveyor build using 3.8.3 to see if that would take the /external flags (as my build does locally)

Related:
https://devblogs.microsoft.com/cppblog/broken-warnings-theory/
https://bugs.python.org/issue39789

MyreMylar:
I've asked a query about why the 'experimental:external' flags generate warnings on appveyor here:

https://help.appveyor.com/discussions/problems/27377-command-line-warning-d9002-ignoring-unknown-option-experimentalexternal

illume:
So the tests can run, I'm going to merge master in (which runs SDL1 tests too).

PR #1822, opened 2020-05-19
Add get_size test

PeanutbutterWarrior:
Creates the test for #1810

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1127,16 +1127,10 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
 
         self.fail()
 
-    def todo_test_get_size(self):
-
-        # __doc__ (as of 2008-08-02) for pygame.surface.Surface.get_size:
-
-        # Surface.get_size(): return (width, height)
-        # get the dimensions of the Surface
-        #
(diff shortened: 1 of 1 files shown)


test/surface_test.py
```diff
@@ -1127,16 +1127,10 @@ def todo_test_get_shifts(self):
...
-        # Surface.get_size(): return (width, height)
-        # get the dimensions of the Surface
-        #
-        # Return the width and height of the Surface in pixels.
-
-        self.fail()
+    def test_get_size(self):
+        surface = pygame.Surface((15, 5))
+        size = surface.get_size()
+        self.assertEqual(size, (15, 5))
```
MyreMylar, line 1133: Hey,

Thanks for contributing! My main suggestion to make this test a bit more comprehensive is to test a bit of a wider range of input values in a loop, like so:

```suggestion
    def test_get_size(self):
        sizes = ((1,1), (119, 10), (1000,1000))  # Could add a few more ideas here 
        for width, height in sizes:
            surf = pygame.Surface((width, height))
            found_size = surf.get_size()
            self.assertEqual((width, height), found_size)
```

PeanutbutterWarrior:
Thanks, I've added them and some more

PR #1823, opened 2020-05-20
Adds test for Surface.get_height

kduggan15:
#1822

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -991,9 +991,11 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
         self.assertEqual(ck, colorkey)
 
     def test_get_height(self):
-        surf = pygame.Surface((2,4))
-        height = surf.get_height()
-        self.assertEqual(height, 4)
+        sizes = ((1,1), (119, 10), (1000,1000))  # Could add a few more ideas here
+        for width, height in sizes:
+            surf = pygame.Surface((width, height))
+            found_height = surf.get_height()
+            self.assertEqual(height, found_height)
 
     def todo_test_get_locked(self):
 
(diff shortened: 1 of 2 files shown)


test/surface_test.py
```diff
@@ -990,16 +990,10 @@ def todo_test_get_colorkey(self):
...
-        # Surface.get_height(): return height
-        # get the height of the Surface
-        #
-        # Return the height of the Surface in pixels.
-
-        self.fail()
+    def test_get_height(self):
+        surf = pygame.Surface((2,4))
+        height = surf.get_height()
+        self.assertEqual(height, 4)
```
MyreMylar, line 996: Hi! 
Thanks for having a go at this, my main suggestion to make this test a bit more comprehensive is to test more values in a loop, just in case the function does anything weird in edge case situations e.g. :

```suggestion
    def test_get_height(self):
        sizes = ((1,1), (119, 10), (1000,1000))  # Could add a few more ideas here 
        for width, height in sizes:
            surf = pygame.Surface((width, height))
            found_height = surf.get_height()
            self.assertEqual(height, found_height)
```

kduggan15:
Thank you for the feedback!

MyreMylar:
This ended up as quite a confusing merge, (if you peek at Files Changed). I could be wrong but I think this is because when you merged the remote changes you did it directly into this PR branch? 

Rather than:

1. pull changes direct from pygame/pygame 'upstream' into a local copy of of pygame/pygame (you should be able to pull from here but not push)
2. switch to your master branch and merge upstream pygame/pygame into your local master branch (by which I mean kduggan15/pygame, i.e. your fork)
3. finally switch to this branch and merge your fork's master branch into this branch.

Which is basically redoing the process that created the fork in the first place. 

Though likely for a change of this size you wouldn't need to be that up to date with the remote master as Git can probably work it out by itself. But I guess it is useful to know anyway, assuming I have properly understood what happened.

If you want to make it a cleaner, easier-to-understand merge you could just do the first two steps above, then create a new branch from your local master, paste your actual changes to the surface_test file in to that and create a fresh pull request.

kduggan15:
I think creating a fresh PR is best. I'll close up this one

Issue #1764, opened 2020-05-16
Add unit test: mouse.set_pos()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/mouse_test.py#L59-L61
**Docs:** https://www.pygame.org/docs/ref/mouse.html#pygame.mouse.set_pos
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/mouse.c#L33-L68

MyreMylar:
This test is now merged! Thanks @Graphcalibur !

PR #1824, opened 2020-05-20
Added tests for mouse.set_pos()

Graphcalibur:
Addresses #1764.

### DIFF
--- a/test/mouse_test.py
+++ b/test/mouse_test.py
@@ -59,10 +59,10 @@ class MouseModuleTest(unittest.TestCase):
     def test_set_pos(self):
         """Ensures set_pos works correctly."""
         expected_pos = ((0, 0), (1, 400), (912, 566), (9999, 9999), (1000, 0), (0, 9999))
-        for pos in expected_pos:
-            pygame.mouse.set_pos(pos)
+        for x, y in expected_pos:
+            pygame.mouse.set_pos(x, y)
             found_pos = pygame.mouse.get_pos()
-            self.assertEqual(found_pos, pos)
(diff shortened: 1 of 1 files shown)


Graphcalibur:
Thank you for the help! I think this commit should get it to work properly.

test/mouse_test.py
```diff
@@ -56,9 +56,27 @@ def test_set_pos__invalid_pos(self):
             with self.assertRaises(TypeError):
                 pygame.mouse.set_pos(invalid_pos)
 
-    def todo_test_set_pos(self):
-        """Ensures set_pos works correctly."""
-        self.fail()
+    def test_set_pos(self):
```
MyreMylar, line 59: Looks like the dummy video driver/window manager used on CI doesn't generate mouse events at all. That's OK though, we can skip this test if the video driver is a "dummy" like so:

```suggestion
    @unittest.skipIf(
            os.environ.get("SDL_VIDEODRIVER") == "dummy",
            'mouse events require a non-"dummy" SDL_VIDEODRIVER',
    )
    def test_set_pos(self):
```

MyreMylar: I suppose a clock timeout might also be good, so we don't ever get into an endless loop like:

    clock = pygame.time.Clock()
    for x, y in expected_pos:
        pygame.event.get()

        pygame.mouse.set_pos(x, y)
        ready_to_test = False
        clock.tick()
        time_passed = 0.0
        while not ready_to_test and time_passed < 1000.0:
            time_passed += clock.tick()

MyreMylar:
All looks good 👍 🎉

PR #1826, opened 2020-05-20
Add test_get_locked

PeanutbutterWarrior:
Test for #1803

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1002,41 +1002,50 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
         self.fail()
 
     def test_get_locked(self):
+        def blit_locked_test(surface):
+            newSurf = pygame.Surface((10, 10))
+            try:
+                newSurf.blit(surface, (0,0))
+            except pygame.error:
+                return True
+            else:
+                return False
+
         surf = pygame.Surface((100, 100))
-        # Surface should start unlocked
-        self.assertFalse(surf.get_locked())
+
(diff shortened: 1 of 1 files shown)


MyreMylar:
Looks pretty solid, I suppose we could also try 

- blitting a locked surface onto another one.
- blitting onto a locked surface and confirming it produces an error.

(you can check for assertions being raised with `self.assertRaises` and passing the function name and it's parameters as separate arguments, see examples here:

https://github.com/pygame/pygame/blob/master/test/blit_test.py#L135)

PeanutbutterWarrior:
Sorry, but how does that test .get_locked()? As another check that a surface is locked?

MyreMylar:
> Sorry, but how does that test .get_locked()? As another check that a surface is locked?

Yeah, my thinking was just a double check that the flag was accurately reporting what was happening deeper in the code somewhere. I mean _probably_ the code right now is just checking for the flag in blit before raising an assert so we would be just in a circle, but perhaps at some point in the codes life it might check for locked-ness of a surface in a different way internally in SDL or pygame and the flag we check with `get_locked()` could get out of sync with what is going on. 

The blit code is a big old mess of code so I could definitely see it :)

Honestly, I think this kind of testing could easily fit in the unit test for lock() (when that exists) - checking that surfaces locked there both can't be blitted on _and_ report that they are locked with get_locked() so we could just punt it over to there if you think it makes this test clearer?

PeanutbutterWarrior:
Yes, I think this would work as a test for lock(), then for get_locked() it will check against the blit check, rather than the expected value.

MyreMylar:
New changes also look good! :)

Issue #1812, opened 2020-05-18
Add unit test: surface.must_lock()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)
[Draft guide on how to write a unit test for pygame](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/surface_test.py#L1187-L1205
**Docs:** https://www.pygame.org/docs/ref/surface.html#pygame.Surface.mustlock
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/surface.c#L1144-L1150

MyreMylar:
This is merged now.

PR #1827, opened 2020-05-20
Adds test for surface.mustlock()

kduggan15:
#1812 
I'm writing tests for Surface.mustlock()
The only case I could find was subsurfaces. 
Another case might be when the HWSURFACE flag is set, but in this case on my machine the surface still returns False on mustlock(). 
I'm not sure what causes this, any feedback would be appreciated.

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1181,10 +1181,17 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
         self.assertEqual(surf.get_at((0, 0)), color)
 
     def test_mustlock(self):
+        #Test that subsurfaces mustlock
         surf = pygame.Surface((4,4))
         subsurf=surf.subsurface(pygame.Rect(1,1,2,2))
         self.assertTrue(subsurf.mustlock())
         self.assertFalse(surf.mustlock())
+        #Test RLEACCEL flag in set_colorkey
+        surf = pygame.Surface((100,100))
+        blit_surf = pygame.Surface((100,100))
+        blit_surf.set_colorkey((0,0,255), RLEACCEL)
+        surf.blit(blit_surf,(0,0))
+        self.assertTrue(blit_surf.mustlock())
 
     def test_set_alpha_none(self):
         """surf.set_alpha(None) disables blending"""
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1180,25 +1180,11 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
         surf.set_at((0, 0), c)
         self.assertEqual(surf.get_at((0, 0)), color)
 
-    def todo_test_mustlock(self):
-
-        # __doc__ (as of 2008-08-02) for pygame.surface.Surface.mustlock:
-
-        # Surface.mustlock(): return bool
-        # test if the Surface requires locking
-        #
-        # Returns True if the Surface is required to be locked to access pixel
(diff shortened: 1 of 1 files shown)


MyreMylar:
It looks like surface.mustlock wraps this SDL macro:

https://github.com/spurious/SDL-mirror/blob/6b6170caf69b4189c9a9d14fca96e97f09bbcc41/include/SDL_surface.h#L62 

...as well as checking for a subsurface.

So probably we should also try two surfaces with the RLEACCEL flag that you can set with these two functions:

https://www.pygame.org/docs/ref/surface.html#pygame.Surface.set_alpha
https://www.pygame.org/docs/ref/surface.html#pygame.Surface.set_colorkey

kduggan15:
I tried creating a surface and setting those flags as you described, but the surface still doesn't return True for must_lock. Here's what it looks like.
```python
        #Test RLEACCEL flag in set_alpha
        surf = pygame.Surface((10,10))
        surf.set_alpha(255, RLEACCEL)
        self.assertTrue(surf.mustlock())
        #Test RLEACCEL flag in set_alpha
        surf = pygame.Surface((10,10))
        surf.set_alpha(255, RLEACCEL)
        self.assertTrue(surf.mustlock())
```

Both of these assertions fail

MyreMylar:
> ```python
> self.assertTrue(surf.mustlock())
> ```

Hmm, I can't get the flag to appear with set_alpha() so far but a quick poke around the other tests reveals the trick to get it active with set_colorkey:

    import pygame
    
    base_surface = pygame.Surface((100, 100), 32)
    test_surface = pygame.Surface((100, 100), 32)
    test_surface.set_colorkey((255, 0, 255), pygame.RLEACCEL)
    base_surface.blit(test_surface, (0, 0))
    if test_surface.get_flags() & pygame.RLEACCEL:
        print('have pygame.RLEACCEL')
        if test_surface.mustlock():
            print("I must be locked")
 
Which seems like it would serve for our purposes here since we aren't actually testing set_alpha().

kduggan15:
I think these tests are more complete and ready for review

illume:
I'd like to merge this soon, because I'd like to get another release out this weekend. Other things can come in further PRs if they're needed :)

A couple of small notes (that also apply to many of the existing pygame tests).

1) The python formatter [black](https://github.com/psf/black) could be run on python changes, so that the python code formatting is [pep8](https://www.python.org/dev/peps/pep-0008/) friendly.
2) probably the surface size could be reduce to speed things up a tiny little bit.

3) It's good to separate tests into smaller tests of what they are testing.

- It makes it easier read smaller tests, since they are simpler and only testing one thing.
- Is a good way to document what is being tested in the method name, with the docstring comment being used to elaborate so_your_method_does_not_get_super_long.

eg.
```
def test_mustlock_subsurface(self):
    """ should use the locking of the parent surface.
         Locking the subsurface, should in turn lock the parent surface.
    """
```

- People debugging issues in the code that is tested can run smaller pieces of code. Imagine adding print statements, or being in a debugger... it's easier to pin point the problem if there is less code running each time. If a test calls the tested-code 5-10 times, then it's slightly harder for the person fixing the bug.
- When a test fails (in the future) it's quicker and easier to pinpoint the problem with smaller self contained tests than with a larger method that tests many things. If a larger test fails, then the person fixing it first has to narrow down where the real problem is.
- Smaller tests also reduce the chance of dependencies between parts of the test making the test incorrect by accident. Smaller tests can be run in random order so as to shake out these issues automatically - whereas a larger test can not be run in a random order automatically.

Issue #1199, opened 2019-07-26
SDL2: examples/overlay.py not working


charlesej:
The `examples/overlay.py` code is not working. The main issue appears to be that `SDL_Overlay` support has been removed in SDL2.0.

Some other, unrelated SDL2 issues, with the overlay example code:
- python 3 execution has a str/bytes issue
- the opened file not being closed

---

 The following code runs the overlay example code.  Different versions of python and pygame were tested with the results shown below. The overlay example code works using python 2.7.10 and pygame 1.9.4 and fails with the other combinations.

```python
# Filename: overlay_examples_issue.py
import os

import pygame
import pygame.examples
import pygame.examples.overlay as overlay

# Run the overlay example code.
DATA_DIR = os.path.join(os.path.split(pygame.examples.__file__)[0], 'data')
overlay.main(os.path.join(DATA_DIR, 'yuv_1.pgm'))
```

Output with python 3.7.2 and pygame 2.0.0.dev3.
```
pygame 2.0.0.dev3 (SDL 2.0.9, python 3.7.2)
Hello from the pygame community. https://www.pygame.org/contribute.html
Unknown format( len 2 ). Exiting...
```

Output with python 2.7.10 and pygame 2.0.0.dev3.
```
pygame 2.0.0.dev3 (SDL 2.0.9, python 2.7.10)
Hello from the pygame community. https://www.pygame.org/contribute.html
Traceback (most recent call last):
  File "overlay_examples_issue.py", line 10, in <module>
    overlay.main(os.path.join(DATA_DIR, 'yuv_1.pgm'))
  File "C:\Program Files\Python27\lib\site-packages\pygame-2.0.0.dev3-py2.7-win-amd64.egg\pygame\examples\overlay.py", line 52, in main
    vPlayer(fname)
  File "C:\Program Files\Python27\lib\site-packages\pygame-2.0.0.dev3-py2.7-win-amd64.egg\pygame\examples\overlay.py", line 36, in vPlayer
    ovl= pygame.Overlay(pygame.YV12_OVERLAY, (w,h))
AttributeError: 'module' object has no attribute 'YV12_OVERLAY'
```

Output with python 3.7.2 and pygame 1.9.4.
```
pygame 1.9.4
Hello from the pygame community. https://www.pygame.org/contribute.html
Unknown format( len 2 ). Exiting...
```

Output with python 2.7.10 and pygame 1.9.4. The overlay code works in this instance.
```
pygame 1.9.4
[...]

PR #1830, opened 2020-05-21
Try out some SDL1 tests for the overlay module.

MyreMylar:
The overlay module is pretty neglected, probably because it doesn't exist in SDL2 but we have one issue open about it here:

https://github.com/pygame/pygame/issues/1199

And I ran across a potential porting guide here:

https://stackoverflow.com/questions/17579286/sdl2-0-alternative-for-sdl-overlay

So, I thought I'd see if I could get some tests going for SDL1 and see if it will at least work under SDL1.

- [ ]  first issue: it occasionally crashes with my test data running on Python 3.8 and Pygame 1.9.6. I get this DeprecationWarning:
`DeprecationWarning: PY_SSIZE_T_CLEAN will be required for '#' formats
  overlay.display(raw)`. I followed the guidance in this issue: https://github.com/python-pillow/Pillow/issues/3750 and made a few modifications to overlay.c to see if that will fix it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ install:
       $PIP_CMD install --upgrade pip
       $PIP_CMD install --upgrade numpy
       $PYTHON_EXE setup.py -config -auto $WHICH_SDL_BUILD
-      $PYTHON_EXE setup.py install -pygame-ci -noheaders -auto
+      $PYTHON_EXE setup.py install -pygame-ci -noheaders
     fi
   - |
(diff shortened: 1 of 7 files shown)


MyreMylar:
This looks useful for understsanding the weird pixel formats used by overlay:

http://rawpixels.net/

MyreMylar:
The end result of this testing is that we can get the SDL1 build up and running on CI again without too much work, so I'm going to do that first in a separate PR - then revisit this one (and probably start it over again).

PR #1832, opened 2020-05-21
SDL 1 CI build fixes

MyreMylar:
This is a grab bag of small fixes to try and get the SDL 1 build building and passing tests again on Travis CI.

There is more detail in this draft PR where I discovered most of these errors in the process of trying to write some tests for the overlay module:

https://github.com/pygame/pygame/pull/1830

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -26,8 +26,14 @@
 #include "_surface.h"
 
 #ifdef PG_ENABLE_ARM_NEON
-// sse2neon.h is from here: https://github.com/DLTcollab/sse2neon
-#include "include/sse2neon.h"
+    // sse2neon.h is from here: https://github.com/DLTcollab/sse2neon
+    #include "include/sse2neon.h"
+#else
+    #if IS_SDLv1
+        // SDL 1 doesn't import the latest intrinsics, this should should pull
+        // them all in for us
+        #include <immintrin.h>
+    #endif /* IS_SDLv1 */
 #endif /* PG_ENABLE_ARM_NEON */
 
 /* The structure passed to the low level blit functions */
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -2403,14 +2403,13 @@ class SurfaceBlendTest(unittest.TestCase):
                                                dst_size=(69, 69),
                                                src_has_alpha=False,
                                                ))
-        self.assertEqual(*test_premul_surf(pygame.Color(30, 20, 0, 51),
-                                           pygame.Color(40, 20, 0, 255),
-                                           src_size=(17, 67),
-                                           dst_size=(69, 69),
-                                           dst_has_alpha=False,
-                                           ))
-        # These tests trigger some of the weird SDL1 alpha blending behaviour
-        if not SDL1:
+            self.assertEqual(*test_premul_surf(pygame.Color(30, 20, 0, 51),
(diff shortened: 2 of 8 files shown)


src_c/alphablit.c
```diff
@@ -26,8 +26,14 @@
 #include "_surface.h"
 
 #ifdef PG_ENABLE_ARM_NEON
-// sse2neon.h is from here: https://github.com/DLTcollab/sse2neon
-#include "include/sse2neon.h"
+    // sse2neon.h is from here: https://github.com/DLTcollab/sse2neon
+    #include "include/sse2neon.h"
+#else
+    #if IS_SDLv1
```
illume, line 32: 👍

MyreMylar:
Looks like two of the new mask tests are also failing on SDL 1, is that because of this bit of code perhaps:

https://github.com/pygame/pygame/blob/f439234df00f0ce80044a6aeedbeca12f29979d7/src_c/mask.c#L2072-L2076

Maybe @charlesej will know.

charlesej:
@MyreMylar I think there needs to be a check added to the `check_surface_pixel_format` function for SDL 1.
https://github.com/pygame/pygame/blob/f439234df00f0ce80044a6aeedbeca12f29979d7/src_c/mask.c#L2023-L2036


 Something like:
```c
static int
check_surface_pixel_format(SDL_Surface *surf, SDL_Surface *check_surf)
{
    if ((surf->format->BytesPerPixel != check_surf->format->BytesPerPixel) ||
        (surf->format->BitsPerPixel != check_surf->format->BitsPerPixel)
#if IS_SDLv2
        || (surf->format->format != check_surf->format->format)
#else
        || ((surf->flags & SDL_SRCALPHA) != (check_surf->flags & SDL_SRCALPHA))
#endif
    ) {
        return 0;
    }

    return 1;
}
```

MyreMylar:
Hooray, with that last fix from Charles it all runs. Probably should merge it in quick before we all break it again in every other PR 🍻 🤦

test/surface_test.py
```diff
@@ -2394,27 +2395,28 @@ def test_premul_surf(src_col,
...
-                                           ))
-        self.assertEqual(*test_premul_surf(pygame.Color(30, 20, 0, 255),
-                                           pygame.Color(40, 20, 0, 255),
-                                           src_size=(17, 67),
-                                           dst_size=(69, 69),
-                                           src_has_alpha=False,
-                                           dst_has_alpha=False,
-                                           ))
-
+        # These tests trigger some of the weird SDL1 alpha blending behaviour
```
illume, line 2398: It makes me a little nervous that these aren't correct on SDL1. Are the tests correct, or is SDL1 incorrect? Has the behaviour changed in these blit routines now? Do the SDL2 ones not match the SDL1 ones?

illume: This doesn't need a response necessarily :) I was just thinking out loud.

MyreMylar: No, you are right to ask I'm definitely going to poke at it and see if I can figure out why this works differently.

I mean in theory you don't _need_ to do pre-multiplied alpha blending in any of those cases because both surfaces don't have alpha but I believe I would expect it to just work boringly without going through the fancy intrinsic code....

Anyway, it just didn't seem the moment to go off on a tangent figuring that out in this PR.

PR #1833, opened 2020-05-22
Font type hints - use bool instead of int

MightyJosip:
Small fix, related to https://github.com/pygame/pygame/pull/1819


MyreMylar:
I also changed the set functions to accept a bool in python 3 as well - or at least as best as can be done in CPython using this argument parsing format:

> p (bool) [int]
> Tests the value passed in for truth (a boolean predicate) and converts the result to its equivalent C true/false integer value. Sets the int to 1 if the expression was true and 0 if it was false. This accepts any valid Python value. See Truth Value Testing for more information about how Python tests values for truth.
> 
> New in version 3.3.
(From here: https://docs.python.org/3/c-api/arg.html)

Dunno if that is enough to update the `set_bold(), set_italic(), set_underline()` & `render(antialias)` type hints. It seems there is no real difference between bool and int on the python side (see: https://blog.rmotr.com/those-tricky-python-booleans-2100d5df92c) so I guess there is no harm to us setting the type hint to bool since that is the expectation.

MightyJosip:
Actually I will place bools there

PR #1834, opened 2020-05-22
Some tweaks to the pre-multiply alpha blending

MyreMylar:
Goals:

 - [x] Trying to get all the test to pass on SDL1 on CI.
 - [x] Improve accuracy of blend formula based on this SDL issue: https://bugzilla.libsdl.org/show_bug.cgi?id=4761
 - [x] Handle 24 bit surfaces and 8 bit surfaces properly.
 - [x] Make SSE2 acceleration work in SDL1 on windows.
 - [x] Make Arm/Raspberry Pi compile `alphablit.c` with SDL1 changes.


MyreMylar:
Was also looking into this test case:

    import pygame
    
    
    pygame.init()
    
    screen=pygame.display.set_mode((200,200))
    screen.fill((255,255,255,255))
    
    # First make surf2 have a transparent background with SRCALPHA
    # This is a white image with transparency
    white_transparent_image = pygame.Surface((100, 100), pygame.SRCALPHA)
    white_transparent_image.fill((107, 107, 107, 107))  # colour is pre-multipled by the alpha
    
    # Just check a value of a pixel in the image
    print(white_transparent_image.get_at([10, 10]))
    # This returns (107, 107, 107, 107) for me... ok so far
    
    # First compose our transparent image onto a black transparent one
    surf2 = pygame.Surface((100, 100), pygame.SRCALPHA).convert_alpha()
    print(surf2.get_at([10, 10]))
    # This returns (0, 0, 0, 0) for me
    surf2.blit(white_transparent_image, [0, 0], special_flags=pygame.BLEND_PREMULTIPLIED)
    print(surf2.get_at([10, 10]))
    # After composing this returns (107, 107, 107, 107) for me
    
    # finally blit on the screen
    screen.blit(surf2, (0, 0), special_flags=pygame.BLEND_PREMULTIPLIED)
    
    # What is the value now showing on the screen where the image is blitted?
    print(screen.get_at([10, 10]))
    # This shows (255, 255, 255, 255) for me.
    screen.blit(surf2, (100, 100), special_flags=pygame.BLEND_PREMULTIPLIED)
    
    pygame.display.flip()
    input()

 ..,modified for pre-multipled alpha from the original issue here:
https://github.com/pygame/pygame/issues/1289#issuecomment-530970898

With this commit and the test case above I get the following output:

    (107, 107, 107, 107)
    (0, 0, 0, 0)
    (107, 107, 107, 107)
    (255, 255, 255, 255)

With SDL 1 & 2 which I believe preserves the spirit of the original problem (not getting a white screen & wanting compatability between SDL1 & SDL2)

MyreMylar:
I also tried the test case from this comment:
https://github.com/pygame/pygame/issues/1289#issuecomment-531307240

    import pygame
    pygame.init()
    
    screen=pygame.display.set_mode((200,200))
    screen.fill((255,255,255,255))
    pygame.display.flip()
    
    white_transparent_image = pygame.Surface((256, 256), pygame.SRCALPHA)
    white_transparent_image.fill((100, 100, 100, 100))
    
    #This returns (255, 255, 255, 100):
    print(white_transparent_image.get_at([10, 10]))
    
    surf2=pygame.Surface((256,256), pygame.SRCALPHA)
    surf2.blit(white_transparent_image, [0, 0], special_flags=pygame.BLEND_PREMULTIPLIED)
    
    # this shows a grey screen:
    screen.blit(surf2, (0,0), special_flags=pygame.BLEND_PREMULTIPLIED)
    # this gives the correct result (white):
    #screen.blit(white_transparent_image, (0,0))
    
    print(screen.get_at([10, 10]))
    
    pygame.display.flip()
    input()

again premultiplying the fill colours by the alpha and using BLEND_PREMULTIPLIED when blitting and get this output:

    (100, 100, 100, 100)
    (255, 255, 255, 255)

Which I believe is correct, though this test case is basically the same as the previous one. I think converting code like this would be generally easier with a function on a pygame.Color you could call to pre-multiply the colour by the alpha for you accurately.

MyreMylar:
Trying the test case from this comment with the same pattern of changes:

https://github.com/pygame/pygame/issues/1289#issuecomment-531314066

    import pygame
    
    pygame.display.init()
    
    pygame.display.set_mode((120, 120))
    
    white_transparent_image = pygame.Surface((256, 256), pygame.SRCALPHA)
    white_transparent_image.fill((100, 100, 100, 100))
    
    # This returns (255, 255, 255, 100):
    print(white_transparent_image.get_at([10, 10]))
    
    surf2 = pygame.Surface((256, 256), pygame.SRCALPHA).convert_alpha()
    surf2.fill((0, 0, 0, 0))
    # This returns (0, 0, 0, 0):
    print(surf2.get_at([10, 10]))
    
    # This returns (99, 99, 99, 99):
    surf2.blit(white_transparent_image, [0, 0], special_flags=pygame.BLEND_PREMULTIPLIED)
    print(surf2.get_at([10, 10]))

Output now:

    (100, 100, 100, 100)
    (0, 0, 0, 0)
    (100, 100, 100, 100)

Which is what you would expect.

MyreMylar:
Again, same changes on @illume's minimal test case from here:

https://github.com/pygame/pygame/issues/1289#issuecomment-538167187

    import pygame
    
    surf1 = pygame.Surface((1, 1), pygame.SRCALPHA)
    surf1.fill((100, 100, 100, 100))
    
    surf2 = pygame.Surface((1, 1), pygame.SRCALPHA)
    surf2.blit(surf1, (0, 0), special_flags=pygame.BLEND_PREMULTIPLIED)
    
    print("surf1.get_at((0, 0))", surf1.get_at((0, 0)))
    print("surf2.get_at((0, 0))", surf2.get_at((0, 0)))

Result:

    surf1.get_at((0, 0)) (100, 100, 100, 100)
    surf2.get_at((0, 0)) (100, 100, 100, 100)

Which is what you would expect.

MyreMylar:
Checking the tests from this comment:

https://github.com/pygame/pygame/issues/1289#issuecomment-557810444

with the same change over to premultiplied alpha:

import unittest
    import pygame
    
    
    class BlitIssueTest(unittest.TestCase):
    
        def test_src_alpha_issue_1289_0(self):
            """
            """
            surf1 = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
            surf1.fill((100, 100, 100, 100))
    
            surf2 = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
            surf2.fill((0, 0, 0, 0))
            self.assertEqual(surf2.get_at((0, 0)), (0, 0, 0, 0))
            surf2.blit(surf1, (0, 0), special_flags=pygame.BLEND_PREMULTIPLIED)
    
            self.assertEqual(surf1.get_at((0, 0)), (100, 100, 100, 100))
            self.assertEqual(surf2.get_at((0, 0)), (100, 100, 100, 100))
    
            # (100, 100, 100, 100)   (0, 0, 0, 0)
            # SDL1:  -> (100, 100, 100, 100)
            # SDL2:  -> (100, 100, 100, 100)
    
        def test_src_alpha_issue_1289_1(self):
            """
            """
            surf1 = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
            surf1.fill((100, 100, 100, 100))
    
            surf2 = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
            surf2.fill((0, 0, 0, 1))
            self.assertEqual(surf2.get_at((0, 0)), (0, 0, 0, 1))
            surf2.blit(surf1, (0, 0), special_flags=pygame.BLEND_PREMULTIPLIED)
    
            self.assertEqual(surf1.get_at((0, 0)), (100, 100, 100, 100))
            self.assertEqual(surf2.get_at((0, 0)), (100, 100, 100, 101))
    
            # (100, 100, 100, 100)   (0, 0, 0, 1)
            # SDL1:  -> (100, 100, 100, 101)
            # SDL2:  -> (100, 100, 100, 101)
    
        def test_src_alpha_issue_1289_2(self):
            """
            """
            surf1 = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
            surf1.fill((100, 100, 100, 100))
    
            surf2 = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
            surf2.fill((0, 0, 0, 255))
            self.assertEqual(surf2.get_at((0, 0)), (0, 0, 0, 255))
            surf2.blit(surf1, (0, 0), special_flags=pygame.BLEND_PREMULTIPLIED)
    
            self.assertEqual(surf1.get_at((0, 0)), (100, 100, 100, 100))
            self.assertEqual(surf2.get_at((0, 0)), (100, 100, 100, 255))
    
            # (100, 100, 100, 100) (0, 0, 0, 255)
            # SDL1: -> (100, 100, 100, 255)
            # SDL2: -> (100, 100, 100, 255)
    
        def test_src_alpha_issue_1289_3(self):
            """
            """
            surf1 = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
            surf1.fill((100, 100, 100, 100))
    
            surf2 = pygame.Surface((1, 1), pygame.SRCALPHA, 32)
            surf2.fill((0, 0, 0, 10))
            self.assertEqual(surf2.get_at((0, 0)), (0, 0, 0, 10))
            surf2.blit(surf1, (0, 0), special_flags=pygame.BLEND_PREMULTIPLIED)
    
            self.assertEqual(surf1.get_at((0, 0)), (100, 100, 100, 100))
            self.assertEqual(surf2.get_at((0, 0)), (100, 100, 100, 106))
    
            # (100, 100, 100, 100) (0, 0, 0, 10)
            # SDL1: -> (100, 100, 100, 106)
            # SDL2: -> (100, 100, 100, 106)
    
    
    if __name__ == "__main__":
        unittest.main()

They all pass with the same results on SDL1 and SDL2 and I think the results seem reasonable.

MyreMylar:
So scanning through this long issue (partly my fault):

https://github.com/pygame/pygame/issues/1289

I believe it covers two issues:

 1. There was no obvious way to do pre-multiplied alpha blending before in pygame. This is what causes the OP's issue as they are trying to do alpha compositing (blitting two layers with alpha ontop of one another) with the default SDL_BLENDMODE_BLEND ([formula here](https://wiki.libsdl.org/SDL_BlendMode)) and it works exactly like the formula, i.e. it looks bad because that is the basic problem with that type of alpha blending ([see here](https://microsoft.github.io/Win2D/html/PremultipliedAlpha.htm) for a decent overview). This was true in pygame 1 & pygame 2 at the tiem because even though we had basic BLEND_PREMULTIPLIED mode, it had no documentation and ran much slower than regular alpha blending because it had no intrinsic acceleration.

2. The second issue is the differences between alpha blending in SDL1 and SLD2. This can be broken down into two sub issues:
    - SDL1 uses a very slightly different blending formula, likely because it is slightly faster - this causes alpha blending results to differ by about 1 per colour channel between SDL1 and SDL 2. It's not visually noticeable and doesn't really cause that much bother.
    - SDL 1 had different behaviour when a pixel had exactly 0 alpha. In that specific case it behaved more like pre-multiplied alpha, not for any mathematical reason but probably because whoever wrote the formula noticed that the normal 'straight alpha' formula looks at it's worst/weirdest when you have a black coloured surface with 0 alpha and you try to blit onto it with another alpha'd surface.

**Conclusion**

On the SDL side, my feeling is that SDL probably will not reintroduce the special case 0 behaviour without a good reason because it is mathematically incorrect and inconsistent with their documentation. There is an issue about the general accuracy of the alpha formula [here](https://bugzilla.libsdl.org/show_bug.cgi?id=4761), which might change the output of SDL's normal blend mode slightly if the formula is change. I don't think it matters that much whatever the outcome as the differences are pretty minor, though I did attempt to use the modified formula from this issue in this changelist as it seemed more accurate.

I think that anyone having alpha compositing problems of this nature should probably be directed into trying out pre-multiplied blending in pygame 2 as that appears to be the generally accepted solution across the digital graphics world. It's possible in the future that SDL will add it's own pre-multiplied blend mode that is superior to pygame's and we could look at using that then.

On our own side, I think that as well as this blendmode we need two additional features to support it: 
    1. A `.premultiply_alpha()` method on `pygame.Color` that will take the current colour channel values and multiply them by the current alpha channel value. I think it should probably also set a flag to only let you do this process once. 
   2. A `.convert_premul_alpha()` method on `pygame.Surface` that modifies a surface in the same way.

Those would help adapt existing programs, though of course users could also load images with the alpha already pre-multiplied as some (though not all) image programs support this.

MyreMylar:
I think I should probably let this get merged in on it's own as it fixes some skipped tests and add the support functions in a different pull request.

src_c/pgcompat.h
```diff
@@ -223,3 +223,5 @@ SDL_Surface * SDL_CreateRGBSurfaceWithFormat(Uint32 flags, int width, int height
 #endif
 
 #endif /* ~PGCOMPAT_INTERNAL_H */
+
+#define WINDOWS_IGNORE_PACKING_MISMATCH
```
illume, line 227: What is this needed for? Do we need it for all files?

illume:
> I think I should probably let this get merged in on it's own as it fixes some previously skipped tests, and then add the support functions in a different pull request later on.

Which skipped tests does this fix? (I don't see any tests with the skip removed)

illume:
Very good. I read through all your posts to try and understand it, but I think I'll come back to it again later.

It sounds like you've got the blitter to work like it did in SDL1 with SDL2?

src_c/pgcompat.h
MyreMylar: It is only needed to compile with recent windows SDKs, and only with SDL1. And I believe it just turns off the warning inside the windows headers. You can read someone complainign about it here:

https://developercommunity.visualstudio.com/content/problem/739697/c2118-error-with-windows-kits-100183620-in-winnth.html

It's a bug/incompatibility between one of the windows headers and code compiled with non 8 byte aligned structs as far as I can tell:

https://stackoverflow.com/questions/56570681/static-assertion-failed-with-windows-headers-require-the-default-packing-option

There are several complaints of it generating warnings in old code across the internet and adding this define appears to be the easy fix out there. 

see:

https://github.com/microsoft/vcpkg/issues/8748



I believe it comes to us via SDL 1 as we don't seem to be doing anything fancy with structs and SDL 2 seems to work fine, so any modules that use SDL would need it. It's possible someone will come up with a better fix in the future but it may be that SDL 1 would need to be rebuilt without /Zp? It's not 100% clear. Anyway, far as I can tell SDL 1 was rubbing along OK with this mismatch before now so maybe ignoring the warning with this define is the best way.

MyreMylar:
> > I think I should probably let this get merged in on it's own as it fixes some previously skipped tests, and then add the support functions in a different pull request later on.
> 
> Which skipped tests does this fix? (I don't see any tests with the skip removed)

I removed an if from the code that was skipping a chunk of the pre-multiply blend tests on SDL1.

src_c/pgcompat.h
nthykier: Should't that define be _inside_ the "include-guard" if-statement rather than outside it to avoid issues if the file is included twice for some reason?

illume: Thanks for the explanation :)

illume:
> I removed an if from the code that was skipping a chunk of the pre-multiply blend tests on SDL1.

Ah yes. I see now.

src_c/pgcompat.h
MyreMylar: > Should't that define be _inside_ the "include-guard" if-statement rather than outside it to avoid issues if the file is included twice for some reason?

Probably! Though I don't think it'll make much difference for a simple define like this. I'll move it.

MyreMylar: I moved it and added more details of the files that are currently affected by the problem. 

While it definitely only happens on our SDL1 build, I'm now not sure it is coming in from SDL1 so it may be possible to fix it on our side somehow. The three .c files currently affected are scrap.c, display.c and event.c.

PR #1841, opened 2020-05-23
Reject attempts to delete attributes in that cannot be deleted

nthykier:


nthykier:
Try a "before / after" on this if you want to see it in action. :)
```
import pygame
c = pygame.color.Color('black')
del c.a
```

PR #1843, opened 2020-05-23
display: Add support for controling screensaver state. part 2/2

illume:
Continuing on https://github.com/pygame/pygame/pull/1734


src_c/display.c
```diff
@@ -2868,27 +2869,35 @@ pg_get_screensaver_enabled(PyObject *self) {
...
 
 static PyObject *
-pg_set_screensaver_enabled(PyObject *self, PyObject* args) {
-    int val;
-    if (!PyArg_ParseTuple(args, "i", &val))
+pg_set_screensaver_enabled(PyObject *self, PyObject *arg, PyObject *kwds) {
+    int val = 0;
+    static char *keywords[] = {"value", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(arg, kwds, "i", keywords, val)) {
```
nthykier, line 2881: Does this mean that `set_screensaver_enabled()` works (i.e. with no arguments)?  If so, I think "False" is a weird default for this function.

nthykier:
Didn't you want to rename these functions?

src_c/display.c
illume: I'm not sure I considered what to do with no value passed in really.


What should it do when called like `set_allow_screensaver()`? Reading the call like that, it reads like the intention is 'allow screensaver'. I guess the default should be True?

nthykier: If it accepts no arguments, I would assume the default was true.  I am a bit conflicted on it accepting no parameters (among other because I think `allow_screensaver([bool])` is a better variant for an argumentless variant - i.e. without set - but it does not follow our existing conventions and is therefore confusing for a different reason).

illume:
I pushed a change with the rename, and changed the default of `set_allow_screensaver(value=True)`.

PR #1638, opened 2020-04-12
pg_IntFromObj: Explicitly handle and convert floats

nthykier:
Prior to this commit, running test/draw_test.py would trigger
deprecation warnings like below.  The floats are passed "by design" to
ensure various PyGame API calls silentl accept floats (truncated to
ints).

```
python3 test/draw_test.py
pygame 2.0.0.dev7 (SDL 2.0.10, python 3.8.2)
Hello from the pygame community. https://www.pygame.org/contribute.html
......................ssss..................................................test/draw_test.py:5475: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  draw.circle(
test/draw_test.py:5487: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  draw.circle(
test/draw_test.py:5499: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  draw.circle(pygame.Surface((2, 2)), (0, 0, 0, 50), (1.3, 1.3), 1.2)
........test/draw_test.py:5406: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  bounds_rect = self.draw_circle(**kwargs)
....................................................test/draw_test.py:1393: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  bounds_rect = self.draw_line(**kwargs)
.test/draw_test.py:1367: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  bounds_rect = self.draw_line(**kwargs)
.......................test/draw_test.py:2062: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  bounds_rect = self.draw_lines(**kwargs)
..........................test/draw_test.py:3964: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  bounds_rect = self.draw_polygon(**kwargs)
[...]

PR #1853, opened 2020-05-24
Font bgcolor SDL1 test fix

### LINKED ISSUE #1852: SDL1: test_freetype_Font_bgcolor failing
```
FAIL: test_freetype_Font_bgcolor (__main__.FreeTypeFontTest)
  File "C:\Program Files\Python38\Lib\site-packages\pygame-2.0.0.dev9-py3.8-win-amd64.egg\pygame\tests\freetype_test.py", line 1247, in test_freetype_Font_bgcolor
    self.assertEqual(s.get_at((0, 0)), transparent_black)
AssertionError: (255, 255, 255, 2) != (0, 0, 0, 0)

```

System details:
- os: windows 10 (64bit)
- python: 3.8.1 (64bit)
- pygame: 2.0.0.dev9 (SDL: 1.2.15) at 0e761fd0541c53c5a20c6756b6c26d9626ef5c06

MyreMylar:
Fixes #1852 

Looks like the test render of a '0' wasn't in a large enough font for the SDL 1 font rendering to not make some mess in the top left pixel.

I also changed an implicit int cast to an explict int one as implict int casts are deprecated in python 3.8 and it was popping an annoying warning.

### DIFF
--- a/test/freetype_test.py
+++ b/test/freetype_test.py
@@ -1230,7 +1230,7 @@ class FreeTypeFontTest(unittest.TestCase):
         self.assertRaises(AttributeError, setattr, f, "fgcolor", None)
 
     def test_freetype_Font_bgcolor(self):
-        f = ft.Font(None, 12)
+        f = ft.Font(None, 32)
         zero = "0"  # the default font 0 glyph does not have a pixel at (0, 0)
         f.origin = False
         f.pad = False
--- a/test/freetype_test.py
+++ b/test/freetype_test.py
@@ -679,7 +679,7 @@ class FreeTypeFontTest(unittest.TestCase):
         rrect = font.render_to(surf, rect, "FoobarBax", color, None, size=24)
         self.assertEqual(rrect.top, rrect.height)
         self.assertNotEqual(rrect.size, rect.size)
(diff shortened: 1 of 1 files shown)


test/freetype_test.py
```diff
@@ -679,7 +679,7 @@ def test_freetype_Font_render_to(self):
         rrect = font.render_to(surf, rect, "FoobarBax", color, None, size=24)
         self.assertEqual(rrect.top, rrect.height)
         self.assertNotEqual(rrect.size, rect.size)
-        rrect = font.render_to(surf, (20.1, 18.9), "FoobarBax", color, None, size=24)
+        rrect = font.render_to(surf, (int(20.1), int(18.9)), "FoobarBax", color, None, size=24)
```
charlesej, line 682: I think the intention for this part of the test is to make sure floats are accepted (but will eventually be truncated to ints). The assert to test this (next line) is commented out so this might need some more investigation. There were similar deprecation warnings in `draw_test.py` and they were addressed by #1638. I would suggest undoing this change and maybe creating a new issue (or issues) to track if something similar to #1638 can be used to fix the deprecation warning and to figure out why the assert is commented out.

MyreMylar: Ah good call, 

I'll revert that part.

PR #1855, opened 2020-05-24
Freetype font int casting fix

MyreMylar:
**What this does**

@charlesej pointed out that there was a better way to solve implict int cast warnings when I applied a quick fix to the freetype tests. The change to `freetype.c` applies this fix by getting ints from the passed in dest param using the function `pg_IntFromObj` to get ints rather than the standard `PyIntAsLong()`.

### DIFF
--- a/docs/reST/ref/freetype.rst
+++ b/docs/reST/ref/freetype.rst
@@ -420,9 +420,8 @@ loaded. This module must be imported explicitly to be used. ::
       That is, the alpha values of the foreground, background, and destination
       target surface all affect the blit.
 
-      The return value is a :class:`Rect <pygame.Rect>` instance containing the
-      width and height of the text's bounding box and the position of the
-      text's origin. This is the same as ``get_rect()``.
+      The return value is a rectangle giving the size and position of the
+      rendered text within the surface.
 
       If an empty string is passed for text then the returned
       :class:`Rect <pygame.Rect>` is zero width and the height of the font.
--- a/docs/reST/ref/freetype.rst
+++ b/docs/reST/ref/freetype.rst
@@ -420,8 +420,9 @@ loaded. This module must be imported explicitly to be used. ::
       That is, the alpha values of the foreground, background, and destination
       target surface all affect the blit.
 
-      The return value is a rectangle giving the size and position of the
-      rendered text within the surface.
+      The return value is a :class:`Rect <pygame.Rect>` instance containing the
+      width and height of the text's bounding box and the position of the
+      text's origin. This is the same as ``get_rect()``.
 
       If an empty string is passed for text then the returned
       :class:`Rect <pygame.Rect>` is zero width and the height of the font.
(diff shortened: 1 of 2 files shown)


docs/reST/ref/freetype.rst
```diff
@@ -420,8 +420,9 @@ loaded. This module must be imported explicitly to be used. ::
       That is, the alpha values of the foreground, background, and destination
       target surface all affect the blit.
 
-      The return value is a rectangle giving the size and position of the
-      rendered text within the surface.
+      The return value is a :class:`Rect <pygame.Rect>` instance containing the
+      width and height of the text's bounding box and the position of the
+      text's origin. This is the same as ``get_rect()``.
```
illume, line 425: This part is confusing to me because I thought it talked about `Surface.get_rect()`.
> This is the same as ``get_rect()``.

illume: What does this part mean?
> position of the text's origin

Does render_to return a rect of where the text is drawn on the surface? So the text's origin if drawn at `(40, 40)` will be `(40, 40)`?

MyreMylar: Maybe my new documentation isn't much better. Basically, _wherever_ you render the text on the surface, you will get the same output rectangle returned from `freetype.Font.render_to()`. This returned rectangle is the same rectangle you get if you pass the same, non-positional text parameters (i.e. without a position on a surface) to `freetype.Font.get_rect()`.

E.g. with

    returned_render_to_rect = font.render_to(surf, (32, 64), "Some Text", color, size=24)

and:

    returned_get_rect = font.get_rect( "Some Text", size=24)

then:

    returned_render_to_rect == returned_get_rect 

The position part of this rectangle is described as the 'offset' in the code and the current `get_rect()` docs, but it does not contain the pixel position the text is rendered at on the surface for render_to as this part of the current render_to() docs seem to suggest:

> The return value is a rectangle giving the size and position of the rendered text within the surface.

illume: Ah, ok! I understand now.


> At this point it is likely too late to change what this method returns as it is already out in the world and altering it would break compatibility.

I did some research by searching github: `pygame render_to -filename:freetype_misc.py`

https://github.com/search?q=pygame+render_to+-filename%3Afreetype_misc.py&type=Code

There I didn't find a usage of anyone using the return value of render_to. There was one result on page 4 that assigned the value, but never used it afterwards. If no one used the return value because the documentation mislead everyone, or if it's not generally useful I'm not sure.

Probably it would be ok to change the behaviour to match the documentation? Since no one seems to be using the return value?

MyreMylar: Hmm, OK. 

That's probably a more complicated job then. I'll strip the doc change from this one and log an issue.

illume: Ok. Thanks.

PR #1858, opened 2020-05-25
Fix _pg_rw_write

### LINKED ISSUE #1714: Saving image to BytesIO/StringIO doesnt work
Hello ,  I need some help , thanks

I have been trying to write some text to an image then save it to a stream to convert it to base64.

I want to avoid acess to disk , so saving to adisk then reloading is not an option, I wrote this code :

`
from io import BytesIO    
f = BytesIO()
f.name="image.png"
pygame.image.save(text, f)
`

Thats the snippet people use online , but they used it all with python 2 , now in python 3 i guess pygame assumes that the second argument of save() is always a path to the target file , so for it that should be a string . thats why i get this error :

`UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte`

That line appears around 100 time , or more , for each byte i guess.

doing my reseach on this , i found people from around the world having this issue , like in here : https://oldj.net/blog/2012/02/19/text-to-image/  ( search for UnicodeDecodeError in that page ) 

people how to let pygame.image.save() accept a BytesIO as second argument , i dont want to switch to another library nor to save to disk :)

dangillet:
Fix #1714 
Calling the helper write function should be done with a Bytes type
instead of a String type.

### DIFF
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -2,6 +2,7 @@
 
 import array
 import binascii
+import io
 import os
 import tempfile
 import unittest
@@ -254,6 +255,30 @@ class ImageModuleTest(unittest.TestCase):
             finally:
                 # clean up the temp file, comment out to leave tmp file after run.
                 os.remove(temp_filename)
+
+    def test_save_to_fileobject(self):
+        s = pygame.Surface((1, 1))
+        s.fill((23, 23, 23))
+        bytes_stream = io.BytesIO()
+
+        pygame.image.save(s, bytes_stream)
+        bytes_stream.seek(0)
+        s2 = pygame.image.load(bytes_stream, "tga")
+        self.assertEqual(s.get_at((0, 0)), s2.get_at((0, 0)))
+
+    def test_save_tga(self):
+        s = pygame.Surface((1, 1))
+        s.fill((23, 23, 23))
+        with tempfile.NamedTemporaryFile(suffix=".tga", delete=False) as f:
+            temp_filename = f.name
+
+        try:
+            pygame.image.save(s, temp_filename)
+            s2 = pygame.image.load(temp_filename)
+            self.assertEqual(s2.get_at((0, 0)), s.get_at((0, 0)))
+        finally:
+            # clean up the temp file, even if test fails
+            os.remove(temp_filename)
 
     def test_save_colorkey(self):
         """ make sure the color key is not changed when saving.
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -27,6 +27,11 @@
 #define _PYGAME_INTERNAL_H
 
 #include "pgplatform.h"
+/*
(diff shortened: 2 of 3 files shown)


test/image_test.py
```diff
@@ -255,6 +256,28 @@ def test_save(self):
...
+        pygame.image.save(s, bytes_stream)
+        bytes_stream.seek(0)
+        s2 = pygame.image.load(bytes_stream, "tga")
+        self.assertEqual(s.get_at((0, 0)), s2.get_at((0, 0)))
+
+    def test_save_tga(self):
+        s = pygame.Surface((1, 1))
+        s.fill((23, 23, 23))
+        try:
+            temp_filename = "%s.tga" % "tmpimg"
```
illume, line 273: Perhaps use the tempfile.NamedTemporaryFile here?
This way the file will be written in a writable area.

```python
with tempfile.NamedTemporaryFile(delete=False) as afile:
    temp_filename = afile.name
```

test/image_test.py
```diff
@@ -255,6 +256,28 @@ def test_save(self):
...
 
+    def test_save_to_fileobject(self):
+        s = pygame.Surface((1, 1))
+        s.fill((23, 23, 23))
+        bytes_stream = io.BytesIO()
+
+        pygame.image.save(s, bytes_stream)
+        bytes_stream.seek(0)
+        s2 = pygame.image.load(bytes_stream, "tga")
+        self.assertEqual(s.get_at((0, 0)), s2.get_at((0, 0)))
```
illume, line 267: 👍

test/image_test.py
dangillet: Excellent suggestion! I was not sure tempfile was a thing in Python 2.7. But when adding it, I noticed it was already use elsewhere. :)

I had just copied over what the previous test was doing as a starting point.

I've made the changes and squashed my commit.

test/image_test.py
```diff
@@ -255,6 +256,26 @@ def test_save(self):
...
+
+    def test_save_tga(self):
+        s = pygame.Surface((1, 1))
+        s.fill((23, 23, 23))
+        with tempfile.NamedTemporaryFile(suffix=".tga") as f:
```
illume, line 272: It looks like this is failing on windows. This is because on windows, you can't open the file if it's open again.

```python
with tempfile.NamedTemporaryFile(delete=False) as afile:
    temp_filename = afile.name

pygame.image.save(s, temp_filename)
s2 = pygame.image.load(temp_filename)
self.assertEqual(s2.get_at((0, 0)), s.get_at((0, 0)))
```

dangillet: You can tell I'm not coding on a Windows machine. Thanks to CI for this problem!

I've amended the code trying to make sure the temp file gets deleted even if the test fails.

illume: Nice one! Looks like it's passing on windows now.

PR #1859, opened 2020-05-25
Adds test for surface.get_offset()

NachiketUN:
#1807 
I have added a test for surface.get_offset() only. But since I used 'black' to reformat the code, it shows multiple additions and deletions.


illume:
Hi,

thanks for your contribution.

Oh, it looks like some other people haven't been running the code formatter.
Sorry about that.


Here's the test amongst the formatting changes:
```python
    def test_get_offset(self):
        """get_offset returns the (0,0) if surface is not a child
                      returns the position of child subsurface inside of parent
        """
        surf = pygame.Surface((100, 100))
        self.assertEqual(surf.get_offset(), (0, 0))
        # subsurface offset test
        subsurf = surf.subsurface(1, 1, 10, 10)
        self.assertEqual(subsurf.get_offset(), (1, 1))
```

I had to fix some conflicts with a recently merged commit.

Issue #1800, opened 2020-05-18
Add unit tests: surface.get_clip()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)
[Draft guide on how to write a unit test for pygame](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/surface_test.py#L968-L982
**Docs:** https://www.pygame.org/docs/ref/surface.html#pygame.Surface.get_clip
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/surface.c#L2052-L2060

cruxicheiros:
Since a test was added in #1860, is there more that needs to be added to the test or should this be closed? 😁

MightyJosip:
This is now done thanks to the #1860 by @pierrehaou. Thanks 👍

PR #1860, opened 2020-05-26
Added unit test for surface.get_clip()

pierrehaou:
I added a unit test for this issue: https://github.com/pygame/pygame/issues/1800

As this is my first time contributing please let me know of any best practices when adding code. Thanks!

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -967,8 +967,12 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
         self.fail()
 
     def test_get_clip(self):
-        surf = pygame.Surface(200,200)
-        pygame.surf.get_clip()
+        s = pygame.Surface((800,600))
+        rectangle = s.get_clip()
+        self.assertEqual(rectangle, (0, 0, 800, 600))
+
+
+
 
 
     def todo_test_get_colorkey(self):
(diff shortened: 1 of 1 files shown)


illume:
Hi @pierrehaou 

thanks for your contribution!

If you look on the details link for the appveyor link, you will see that some tests are failing.

```
====================================================================
ERROR: test_get_clip (pygame.tests.surface_test.SurfaceTypeTest)
  File "C:\Python27\lib\site-packages\pygame\tests\surface_test.py", line 970, in test_get_clip
    surf = pygame.Surface(200,200)
ValueError: size needs to be (int width, int height)
```

Just calling the function is a useful test by itself. If you want to improve it, you can add an `assertEquals()` call in there with what you expect the return value of `Surface.get_clip()` to be.

pierrehaou:
Thanks for responding @illume ! I believe I have corrected the errors even though only 1/4 checks have passed.

Another thing I noticed when running the file, it returns:
```
Ran 91 tests in 2.696s

FAILED (failures=1, skipped=3, expected failures=2)
```
unsure why failures:1 and expected failueres: 2

illume:
> unsure why failures:1 and expected failueres: 2

It looks like we have a failure in there that slipped through. Currently we don't have any MacOS builders, and this is related to that.

PR #1862, opened 2020-05-26
font: Add attribute accessors for bold, italic and underline

nthykier:
Because `f.bold = True` is nicer than `f.set_bold(True)`. :)

### DIFF
--- a/docs/reST/ref/font.rst
+++ b/docs/reST/ref/font.rst
@@ -139,6 +139,56 @@ loaded instead.
    font with actual italic or bold glyphs. The rendered text can be regular
    strings or unicode.
 
+   .. attribute:: bold
+
+      | :sl:`Gets or sets whether the font should be rendered in (faked) bold.`
+      | :sg:`bold -> bool`
+
+      Whether the font should be rendered in bold.
+
+      When set to True, this enables the bold rendering of text. This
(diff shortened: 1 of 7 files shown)


docs/reST/ref/font.rst
```diff
@@ -139,6 +139,56 @@ loaded instead.
...
+      .. versionadded:: 2.0.0
+
+      .. ## Font.bold ##
+
+   .. attribute:: italic
+
+      | :sl:`Gets or sets whether the font should be rendered in (faked) italics.`
+      | :sg:`bold -> bool`
+
+      Whether the font should be rendered in bold.
```
illume, line 164: > Whether the font should be rendered in bold.
should be italic

docs/reST/ref/font.rst
```diff
@@ -139,6 +139,56 @@ loaded instead.
...
+      .. versionadded:: 2.0.0
+
+      .. ## Font.italic ##
+
+   .. attribute:: underline
+
+      | :sl:`Gets or sets whether the font should be rendered with an underline.`
+      | :sg:`bold -> bool`
+
+      Whether the font should be rendered in bold.
```
illume, line 182: >       Whether the font should be rendered in bold.
underline

Issue #1868, opened 2020-05-28
surface.blit area rect can't be a big number (>16bit)


galexandreg:
### Problem
I can't use big rect coordinates (more than 16bit I guess) in the area argument of surface.blit(s) to select a small area from a really big surface to blit on another one.

```
import sys
import pygame

window = pygame.display.set_mode((800, 600))

testsurface = pygame.Surface((100, 128000))
testsurface.fill((255, 0, 0))

testrect1 = pygame.Rect(0, 16000, 100, 100)
testrect2 = pygame.Rect(0, 32000, 100, 100)
testrect3 = pygame.Rect(0, 64000, 100, 100)

while True:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            sys.exit()

    window.fill((0, 255, 0))
    window.blit(testsurface, (100, 100), area=testrect1)
    window.blit(testsurface, (200, 200), area=testrect2)
    window.blit(testsurface, (300, 300), area=testrect3)
    pygame.display.flip()
```
I would expect to see three red squares on a green background, but I see only two.

### Possible cause
It seems like SDL1 rects were always 16 bits integers, but with SDL2, they are simple ints.

**SDL1 (SDL_video.h):**
```
typedef struct SDL_Rect {
    Sint16 x, y;
    Uint16 w, h;
} SDL_Rect;
```
**SDL2 (SDL_rect.h):**
```
typedef struct SDL_Rect
{
    int x, y;
    int w, h;
} SDL_Rect;
```

**Pygame always cast rect sizes and coordinates to shorts in Surface.blit and Surface.blits.**
```
dest_rect.x = (short)dx;
dest_rect.y = (short)dy;
dest_rect.w = (unsigned short)src_rect->w;
dest_rect.h = (unsigned short)src_rect->h;
sdlsrc_rect.x = (short)src_rect->x;
sdlsrc_rect.y = (short)src_rect->y;
sdlsrc_rect.w = (unsigned short)src_rect->w;
sdlsrc_rect.h = (unsigned short)src_rect->h;
```

**Pygame rects always use ints, even in Pygame1**
```
#if !defined(SDL_VERSION_ATLEAST) || PG_API_VERSION == 1
typedef struct {
    int x, y;
    int w, h;
} GAME_Rect;
#else /* SDL 2+ */
typedef SDL_Rect GAME_Rect;
#endif /* SDL 2+ */
```

[...]

PR #1869, opened 2020-05-28
Blits - Stop casting rects to 16bit ints when using SDL2

galexandreg:
This is my attempt to fix the #1868 issue.

**Internal Pygame rects are ints**
```
#if !defined(SDL_VERSION_ATLEAST) || PG_API_VERSION == 1
typedef struct {
    int x, y;
    int w, h;
} GAME_Rect;
#else /* SDL 2+ */
typedef SDL_Rect GAME_Rect;
#endif /* SDL 2+ */
```

But SDL_Rects are 16bit in SDL1 and normal ints in SDL2:
**SDL1 (SDL_video.h):**
```
typedef struct SDL_Rect {
    Sint16 x, y;
    Uint16 w, h;
} SDL_Rect;
```
**SDL2 (SDL_rect.h):**
```
typedef struct SDL_Rect
{
    int x, y;
    int w, h;
} SDL_Rect;
```

It's not necessary to cast rect data to shorts in blits when using SDL2.

This may be completely wrong... I don't really know. 😕  But heh... How am I going to learn if I don't try. 😄 

Thanks in advance for your feedback.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -2173,7 +2173,10 @@ surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
     PyObject *argpos, *argrect = NULL;
     pgSurfaceObject *srcobject;
     int dx, dy, result;
-    SDL_Rect dest_rect, sdlsrc_rect;
+    SDL_Rect dest_rect;
+#if IS_SDLv1
+    SDL_Rect sdlsrc_rect;
+#endif /* IS_SDLv1  */
     int sx, sy;
     int the_args = 0;
 
@@ -2216,6 +2219,7 @@ surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
         src_rect = &temp;
     }
 
+#if IS_SDLv1
     dest_rect.x = (short)dx;
     dest_rect.y = (short)dy;
     dest_rect.w = (unsigned short)src_rect->w;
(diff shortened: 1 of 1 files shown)


illume:
This reproducer is from the #1868 issue...

Works for me :)

```python
import sys
import pygame

window = pygame.display.set_mode((800, 600))

testsurface = pygame.Surface((100, 128000))
testsurface.fill((255, 0, 0))

testrect1 = pygame.Rect(0, 16000, 100, 100)
testrect2 = pygame.Rect(0, 32000, 100, 100)
testrect3 = pygame.Rect(0, 64000, 100, 100)

while True:
        if event.type == pygame.QUIT:
            sys.exit()

    window.fill((0, 255, 0))
    window.blit(testsurface, (100, 100), area=testrect1)
    window.blit(testsurface, (200, 200), area=testrect2)
    window.blit(testsurface, (300, 300), area=testrect3)
    pygame.display.flip()

```

MightyJosip:
Also works for me on Windows (dev8 has 2 rects, this version 3). Great job @galexandreg for detailed issue report, and especially for fixing it right away. This can be proved to be useful in open world games that could have huge sizes of surfaces.

galexandreg:
> Looks good to me.
> 
> We probably need a few tests though to see if it's working ok.

Thanks!
I see that you saw my issue with the test a few hours later.

do... you... want me to write a unittest for it? 😕 
😄 I'm really new to all this stuff. But anyway, I guess I can make one if that's what you mean.

illume:
Sorry, I missed your example in the linked issue before I wrote that.

If you added something like this to test/surface_test.py and committed it, it should run on all the different build robots.

```python

    @unittest.skipIf(
        pygame.get_sdl_version()[0] == 1,
        'only works in SDL2',
    )
    def test_blit_big_rects(self):
        """ SDL2 can have more than 16 bits for x, y, width, height.
        """
        surf = pygame.Surface((500, 500))
        testsurface = pygame.Surface((100, 128000))
        testsurface.fill((255, 0, 0))

        surf.fill((0, 255, 0))
        surf.blit(testsurface, (100, 100), area=(0, 16000, 100, 100))
        surf.blit(testsurface, (200, 200), area=(0, 32000, 100, 100))
        surf.blit(testsurface, (300, 300), area=(0, 64000, 100, 100))

        self.assertEqual(surf.get_at((101, 101)), (255, 0, 0, 255))
        self.assertEqual(surf.get_at((201, 201)), (255, 0, 0, 255))
        self.assertEqual(surf.get_at((301, 301)), (255, 0, 0, 255))

        self.assertEqual(surf.get_at((400, 301)), (0, 255, 0, 255))
        self.assertEqual(surf.get_at((400, 201)), (0, 255, 0, 255))
        self.assertEqual(surf.get_at((100, 201)), (0, 255, 0, 255))
        self.assertEqual(surf.get_at((99, 99)), (0, 255, 0, 255))
        self.assertEqual(surf.get_at((450, 450)), (0, 255, 0, 255))

```

PR #1877, opened 2020-05-30
vsync sdl2 keyword takes an int

illume:
continuing on from https://github.com/pygame/pygame/pull/1774 for the vsync issue at: https://github.com/pygame/pygame/issues/735

- [x] vsync docs 
- [x] use vsync in examples.testsprite (-vsync)


MyreMylar:
Looks good to me, I assume this is so we can add the ability to set it to 2 later on?

illume:
Thanks @MyreMylar. Yeah... I guess so. There are a few different sync modes that we could support. Including every second frame sync, 'adaptive sync', and gsync/freesync. This all needs more research though... so that can be continued in #735

Yohan27x:
Hey guys, 

I have the same problem than all of you with jittering movement.
I read all of your messages and i tried many things ( a little bit randomly because i'm new to programming and pygame ) 

I end up with a solution that seems to works to have smooth movement, there is my code : 

import pygame
import sys,os
os.environ['PYGAME_VSYNC'] = "1"

pygame.init()
screen = pygame.display.set_mode((800, 600),pygame.SCALED| pygame.DOUBLEBUF , vsync=1 )
pygame.display.set_caption('Movement test')

FPSCLOCK = pygame.time.Clock()



def main():
    done = False
    x = 0

    while not done:
            if event.type == pygame.QUIT:
                done = True

        x += 4
        screen.fill((0, 0, 0))
        pygame.draw.rect(screen, (0, 0, 255), (x, 300, 70, 80))
        pygame.display.flip()
        FPSCLOCK.tick(60)

        #pygame.time.delay(1)

    pygame.quit()
    sys.exit()

if __name__ == "__main__":
    main()

Keep me up please if it works or not for you !

PR #1878, opened 2020-05-30
Refactor setup, add -stripped build option

robertpfeiffer:
setup.py is currently full of dead code, code that *should* be dead but is still reachable, and side effects.

This pull request removes handling of cases that should not happen any more, like no newbuffer support. Current versions of PyPy and CPython (3.5 and up) implement the buffer API, and older versions that don't are not supported. It's the same with setuptools.

Instead of handling the absence of these features, we should just return early on unsupported Python versions.

This pull request also adds a "-stripped" parameter, to be invoked like this: `python3 setup.py -stripped bdist_wheel`. Stripped builds do not include any docs, tests, or code examples, so you can save space when embedding them in a binary distribution created with p4a or PyInstaller.

These features are already part of the android branch, and by porting them over to `master`, I hope we get closer to pointing the p4a build recipe to an official release tarball instead of a specialised version.


illume:
We support python 2.7. At least until pygame 2 is released, at which point we can assess the situation again.

setup.py
```diff
@@ -68,6 +73,17 @@ def compilation_help():
 if not hasattr(sys, 'version_info') or sys.version_info < (2,7):
     compilation_help()
     raise SystemExit("Pygame requires Python version 2.7 or above.")
+if sys.version_info >= (3, 0) and sys.version_info < (3, 5):
+    compilation_help()
+    raise SystemExit("Pygame requires Python3 version 3.5 or above.")
+if IS_PYPY and sys.version_info < (3, 5):
+    raise SystemExit("Pygame requires PyPy compatible with Python 3.5 or above.")
+
+def consume_arg(name):
```
illume, line 82: 👍

setup.py
```diff
@@ -100,19 +115,29 @@ def compilation_help():
...
     os.environ['CFLAGS'] = cflags
-    sys.argv.remove ('-pygame-ci')
+
+STRIPPED=False
+
+# STRIPPED builds don't have developer resources like docs or tests
+# TODO: mangle CFLAGS and strip debug symbols
+
+if "PYGAME_ANDROID" in os.environ:
+    # test cases and docs are useless inside an APK
```
illume, line 125: They're useful to some.
Testing on user devices is much easier with tests included.
People developing with pygame to learn might like the docs included.

setup.py
```diff
@@ -201,21 +223,6 @@ def compilation_help():
 
 revision = ''
 
-# Python 3.0 patch
```
illume, line 204: 👍

setup.py
```diff
@@ -335,34 +342,22 @@ def run_install_headers(self):
...
-enable_newbuf = False
-if sys.version_info >= (2, 6, 0):
-    try:
-        sys.pypy_version_info
-    except AttributeError:
-        enable_newbuf = True
+# decide whether or not to enable new buffer protocol support (PEP 3118)
+# old CPython versions without newbuf are no longer supported!
+# new PYPY also supports PEP 3118
+enable_newbuf = True
```
illume, line 348: If pypy supports it now, and python 2.7 supports it... then I think we're good.

setup.py
```diff
@@ -509,6 +506,13 @@ def write_version_module(pygame_version, revision):
 #required. This will be filled if doing a Windows build.
 cmdclass = {}
 
+def add_command(name):
```
illume, line 509: 👍

setup.py
```diff
@@ -68,6 +73,17 @@ def compilation_help():
 if not hasattr(sys, 'version_info') or sys.version_info < (2,7):
     compilation_help()
     raise SystemExit("Pygame requires Python version 2.7 or above.")
+if sys.version_info >= (3, 0) and sys.version_info < (3, 5):
+    compilation_help()
+    raise SystemExit("Pygame requires Python3 version 3.5 or above.")
+if IS_PYPY and sys.version_info < (3, 5):
+    raise SystemExit("Pygame requires PyPy compatible with Python 3.5 or above.")
```
illume, line 80: Last I looked, pypy (the 2.7 equivalent) worked. Am I misunderstanding this message?
I guess we should find the exact pypy version that pygame supports 'well enough'?

setup.py
```diff
@@ -28,6 +28,12 @@
 import sys
 import os
 
+# just import these always and fail early if not present
+import distutils
+from setuptools import setup
+
+IS_PYPY = '__pypy__' in sys.builtin_module_names
```
illume, line 35: 👍

An aside... I still wonder if some distros would still prefer not to have to depend on setuptools. I guess it's so widespread now they don't care.

setup.py
```diff
@@ -822,6 +805,19 @@ def run(self):
...
+    PACKAGEDATA = {
+    "cmdclass":    cmdclass,
+    "packages":    ['pygame',
+                    'pygame.threads',
+                    'pygame._sdl2'],
+    "package_dir": {'pygame': 'src_py',
+                    'pygame._sdl2': 'src_py/_sdl2',
+                    'pygame.threads': 'src_py/threads'},
+    "ext_modules": extensions,
+    "zip_safe":  False,
```
illume, line 818: An aside... I wonder if we can get zip_safe egg(zip file) based pygame working again. It makes load time quite a bit faster with slow IO like on many android devices. Might be something to research in the future.

setup.py
robertpfeiffer: On desktop, this is an *option*, on android, it's default. We can always add an option to disable it and include tests in the apk. The idea is to skip docs when packaging a binary with PyInstaller, not when installing via pip.

setup.py
robertpfeiffer: I haven't yet tested on PyPy 2.7, this might be a bit too conservative.

robertpfeiffer:
So what is left to be done here?

setup.py
```diff
@@ -335,34 +342,22 @@ def run_install_headers(self):
...
+# decide whether or not to enable new buffer protocol support (PEP 3118)
+# old CPython versions without newbuf are no longer supported!
+# new PYPY also supports PEP 3118
+enable_newbuf = True
 
-if enable_newbuf:
-    enable_newbuf_value = '1'
-else:
-    enable_newbuf_value = '0'
+# TODO: remove all PG_ENABLE_NEWBUF conditionals from C code
```
illume, line 350: I'd like to move towards leaving TODO notes as github issues instead of leaving them in the code as part of the (py)lint cleanups.

setup.py
```diff
@@ -100,19 +115,29 @@ def compilation_help():
               '-Werror=cast-align -Werror=int-conversion ' + \
               '-Werror=incompatible-pointer-types'
     os.environ['CFLAGS'] = cflags
-    sys.argv.remove ('-pygame-ci')
+
+STRIPPED=False
+
+# STRIPPED builds don't have developer resources like docs or tests
+# TODO: mangle CFLAGS and strip debug symbols
```
illume, line 122: Can you please remove this TODO (perhaps put it inside an issue).

Issue #1793, opened 2020-05-18
Add unit test: time.wait()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)
[Draft guide on how to write a unit test for pygame](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/time_test.py#L184-L199
**Docs:** https://www.pygame.org/docs/ref/time.html#pygame.time.wait
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/time.c#L163-L192

Gogotchuri:
@MyreMylar I made a [pull request](https://github.com/pygame/pygame/pull/1879) for time.delay() and time.wait(). Please, let me know if improvements are neccessary.

MyreMylar:
This is merged now.

PR #1879, opened 2020-05-30
Pygame time.wait() time.delay() tests impelmented

Gogotchuri:
Implemented tests for wait() and delay(), would like to get a review and try again if need be.

### DIFF
--- a/test/time_test.py
+++ b/test/time_test.py
@@ -138,25 +138,13 @@ class ClockTypeTest(unittest.TestCase):
 class TimeModuleTest(unittest.TestCase):
     def test_delay(self):
         """Tests time.delay() function."""
-        # call time.delay(millis) "iterations" times and check each time if
-        # wait "waited" for given millisecond (+- delta). At the end, take
-        # average time for each call (whole_duration/iterations), which should
-        # be equal to millis (+- delta - acceptable margin of error)
-        millis = 50
-        iterations = 20
-        # TODO: review margin of error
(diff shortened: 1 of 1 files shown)


Gogotchuri:
> The test for the expected behaviour looks good to me. The main remaining thing I would test is unexpected inputs. I see the function has some handling for passing in non-ints so it'd probably be a good idea to test that and make sure it works now, and more importantly, stays working in the future if someone changes the functions.
> 
> Basically, I would add 4 extra assert checks per function that pass a negative int, a string, a float and a Tuple to both functions and then see what output you get (I think it will be mostly TypeErrors & ValueErrors) then you can make a test for the correct assertions with unittest's assertRaises() method.
> 
> That's about all else I can think of for these functions they are fairly straightforward.
 Thanks, for your reply. I added 3 assertion cases mentioned above and created new methods to avoid code duplication. Regarding the negative value integer, the function doesn't seem to throw an exception in that case. So, I think it's ok.

Gogotchuri:
Thanks, for your reply. I added 3 assertion cases mentioned above and created new methods to avoid code duplication. Regarding the negative value integer, the function doesn't seem to throw an exception in that case. So, I think it's ok.

illume:
I'm going to merge this in now, so it can be included in the release this morning. MyreMylar may possibly have further comments, but those can go in the issue https://github.com/pygame/pygame/issues/1793 and any other commits can happen in other PRs if you want to.

cheers!

PR #1881, opened 2020-05-31
Added unit test for surface.get_colorkey(). Closes #1801

### LINKED ISSUE #1801: Add unit test: surface.get_colorkey()
This function needs a unit test finishing.

**Test stub:** https://github.com/pygame/pygame/blob/master/test/surface_test.py#L984-L991
**Docs:** https://www.pygame.org/docs/ref/surface.html#pygame.Surface.get_colorkey
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/surface.c#L1490-L1526

khuang0312:
Some of the original code was actually wrong. For instance, checking the type of the returned value from surface.get_colorkey() to see if it is a Color object or an instance of a child class of Color. 

I also added some more cases to make the test more comprehensive. For example, checking what happens when you set the color key using an RGBA value, the alpha not 255, and what surface.get_colorkey() should return.

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1029,16 +1029,20 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
         self.assertEqual(s.get_colorkey(), (r, g, b, 255))
         
         # test for using method after display.quit() is called...
-        with self.assertRaises(pygame.error):
+        def f():
             s = pygame.display.set_mode()
             pygame.display.quit()
-            s.get_colorkey() 
+            s.get_colorkey()
+
+        self.assertRaises(pygame.error, f)
 
         # test for using method when display is created with OpenGL and the SDL version is 1
-        if pygame.get_sdl_version()[0] > 1:
-            with self.assertRaises(pygame.error):
+        if SDL1: # SLD1 is a bool defined at the top...
+            def f():
                 s = pygame.display.set_mode(flags=pygame.OPENGL)
                 s.get_colorkey()
+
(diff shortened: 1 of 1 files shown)


khuang0312:
Okay, I just added the two new unit tests.

khuang0312:
I seem to be failing the integration tests by not raising an error, even though on my machine, it raises the error.

MyreMylar:
Hmm, it may be that the 'dummy' video driver, which is used on CI, doesn't invalidate surfaces in the same way? I'm not certain. I suppose you could test it locally by adding 

 `os.environ['SDL_VIDEODRIVER'] = 'dummy'` 

Somewhere at the top of the test file temporarily and see if it breaks the assert firing locally.

If so, perhaps the best thing to do is pull out that test and put it into a separate test_function (something like 'test_get_colorkey__quit_assertion' and then use the skip decorator if the video driver is == to dummy . e.g.


https://github.com/pygame/pygame/blob/ccfe35c3c16e960748443f461d92f02f0b8d3661/test/event_test.py#L555-L559

MyreMylar:
Should probably also note that color keys are an area with lots of bugs hovering around them right now so it may be that there are genuine code problems.

MyreMylar:
@illume I think we could probably just restart the Appveyor CI on our end couldn't we? At least that worked to fix Travis. I just don't have the power to restart Appveyor.

illume:
Oh yeah. We can just merge this in now... because tests pass. No need to do it in this pr.

Issue #1392, opened 2019-10-09
Documentation for event queue limitations, and call frequency requirements.


mcpalmer1980:
[Edit.] The SDL event queue (which pygame uses) needs to fetch events off the queue regularly otherwise they get lost.

- [x] Can you think of documentation change(s) so that this limitation of the event queue is more obvious?

---

I'm not sure if this is a bug in pygame or in my code. Unfortunately there is too much code to include it all here and I don't believe I can reproduce the problem with a smaller snippet. It could even be a problem with my Sony Dualshock 4, but I highly doubt it. I've spent hours trying to track down or work around the issue to no avail.

The problem manifests itself by returning a 'stuck' value from joystick.get_hat(). The method will return the stuck value for all successive calls until I press the hat along the same axis. It only occurs after I call a function that I use to show dialogue in my game, and only if I let go of the hat as soon as the animated dialogue begins to play.

This would of course suggest that the problem is in my own code, but I don't think so. Somehow the JOYHATMOTION event never returns a (0,0) value in these situations. If I hold the button a little longer the event is returned.

I tried to work around this problem by posting an event like this:

		ev = pygame.event.Event(JOYHATMOTION, joy=0, hat=0, value=(0,0))
		pygame.event.post(ev)

When I do this the event shows up in my debug log, but it doesn't effect the return value of joystick.get_hat(). If I don't post the event then the same debug text never prints. I don't see how my code could be blocking the event but it is.

The code for message.CheckInput() follows. It is separate from the game.GetInput() method that I used for normal gameplay input. It is designed to allow players to skip a page of dialogue by pressing any button on any device, or to skip all remaining dialogue by pressing ESCAPE or holding down any joystick button beyond a certain threshold.

	def CheckInput(self, lock=False):
		events = pygame.event.get()
		if lock or self.locked:
			lock = True
			self.can_cancel = False

		pressed = False
		for event in events:
			if event.type == QUIT:
				self.game.Terminate()
			elif event.type == MUSIC_DONE:
				pass
				#self.game.NextMusicTrack()
			elif event.type == KEYDOWN and event.key == K_ESCAPE and not lock:
				self.skip_rest = True
				return True
			elif event.type == JOYHATMOTION:
				print('in message.checkinput', event)

		keys = pygame.key.get_pressed()
		mouse = pygame.mouse.get_pressed()
		if True in keys or True in mouse:
			pressed = True
		if self.game.joystick:
			for i in range(self.game.joystick.get_numbuttons()):
[...]

PR #1883, opened 2020-05-31
Strengthen warning in event docs about handling the event queue

MyreMylar:
For this issue:

https://github.com/pygame/pygame/issues/1392

If you don't call `pygame.event.pump()` somehow (it's called by `pygame.event.get()` by default) every frame your app's event handling is going to behave weirdly. I think the docs for `pump()` make this clear already but the warning in the main event module docs was a bit ambiguous.

### DIFF
--- a/docs/reST/ref/event.rst
+++ b/docs/reST/ref/event.rst
@@ -19,9 +19,10 @@ events into the queue from other threads, please use the
 The event queue has an upper limit on the number of events it can hold
 (128 for standard SDL 1.2). When the queue becomes full new events are quietly
 dropped. To prevent lost events, especially input events which signal a quit
-command, your program must regularly check for events and process them.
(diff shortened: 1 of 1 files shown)


docs/reST/ref/event.rst
```diff
@@ -19,9 +19,10 @@ events into the queue from other threads, please use the
 The event queue has an upper limit on the number of events it can hold
 (128 for standard SDL 1.2). When the queue becomes full new events are quietly
 dropped. To prevent lost events, especially input events which signal a quit
-command, your program must regularly check for events and process them.
-To speed up queue processing use :func:`pygame.event.set_blocked()` to
-limit which events get queued.
+command, your program must handle events every frame (with
+``pygame.event.get()``, ``pygame.event.pump()`` or ``pygame.event.clear()``)
```
illume, line 23: I get pygame.event.wait can also be used as well.

Issue #456, opened 2018-05-27
Surface.set_alpha() sets the SRCALPHA flag


gummbum:
- [ ] a test case as a unit test showing the 'old' behaviour.
- [ ] document old behaviour, and which versions of pygame the behaviour changed.
- [ ] confirm if the change is with SDL version change, directx VS windb video driver with SDL1, or with a pygame change.
- [ ] change it back in SDL1
- [ ] change it back in SDL2

---

This issue is possibly related to #455.

# Windows 7, Python 2.7, pygame.1.9.2-cp27 and 1.9.3
import pygame
pygame.init()
pygame.display.set_mode((200, 200))
surf1 = pygame.Surface((20, 20))
print('surf1 before', surf1.get_alpha(), surf1.get_flags() & pygame.SRCALPHA)

surf1.set_alpha(128)
print('surf1 before', surf1.get_alpha(), surf1.get_flags() & pygame.SRCALPHA)

gummbum:
Here's a workaround for anyone in the same bind, i.e. where the routine handles generic surfaces and therefore needs to inspect them in order to apply the appropriate alpha effect.

1. save the surface alpha
2. set the surface alpha to None
3. test the surface for SRCALPHA
4. if SRCALPHA, do blending for per-pixel alpha
5. restore the surface alpha

gummbum:
Update. And this is surprising. I verified these results on Windows 7 and Raspbian. An image with an alpha layer renders in about half the time as the same image converted and with a colorkey.

The following renders 20 images at 770 FPS:

image = pygame.image.load('data/tomato.png').convert_alpha()

The following renders 20 images at 380 FPS:

image = pygame.image.load('data/tomato.png').convert_alpha()
image = image.convert()
image.set_alpha(None)
image.set_colorkey(image.get_at((0, 0)))

I seem to recall this did not used to be the case. The use of colorkey used to render much faster than per-pixel alpha.

Also I notice that in the first case, when I set alpha to None it changes how the image is rendered.

image = pygame.image.load('data/tomato.png').convert_alpha()
image.set_alpha(None)  # this produces an opaque image

This also did not used to be the case. It used to be that an image with an alpha layer was unaffected by set_alpha() and set_colorkey(). The alpha layer had to be stripped in order for them to have an effect. And even if this new way is intentional, the performance of doing this is still poorer than convert_alpha().

[...]

PR #1889, opened 2020-06-01
Check for presence of colorkey to determine setting BLENDMODE_BLEND

MyreMylar:
I'm not sure if this will work at all because I don't have working png loading when I build from source with sdl2 on windows. However, if someone wants to grab this branch build it for sdl2 and give it a go with solarwolf, or just the test case below:

    import pygame
    
    
    pygame.display.init()
    dsurf = pygame.display.set_mode((320, 240), 0)
    
    
    def update(surface):
        dsurf.fill((0, 0, 0))
        dsurf.blit(surface, (0, 0))
        pygame.display.flip()
    
    
    img_surf = pygame.image.load('data/menu_creds_off.png')
    
    color_key = img_surf.get_colorkey()
    print('initial color_key', color_key)
    img_surf.set_colorkey(color_key, pygame.RLEACCEL)
    img_surf = img_surf.convert()
    print('initial alpha', img_surf.get_alpha())
    print('initial bitsize', img_surf.get_bitsize())
    print('initial flags', img_surf.get_flags())
    
    update(img_surf)
    
    img_surf.set_alpha(90)
    update(img_surf)
    img_surf.set_alpha(255)
    update(img_surf)
    
    running = True
    while running:
            if event.type == pygame.QUIT:
                running = False

If it doesn't look red, we might be onto something.

Related to bugs we are having here:

https://github.com/pygame/pygame/issues/721

and here:

https://github.com/pygame/pygame/issues/456

and maybe a few other places.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1534,6 +1534,7 @@ surf_set_alpha(pgSurfaceObject *self, PyObject *args)
     PyObject *alpha_obj = NULL, *intobj = NULL;
     Uint8 alpha;
     int result, alphaval = 255;
+    Uint32 colorkey;
 #if IS_SDLv1
     int hasalpha = 0;
 #endif /* IS_SDLv1 */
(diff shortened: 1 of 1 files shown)


MyreMylar:
This should be a draft but I've lost where the button is.

illume:
It looks red to me :(

illume:
I'll try and get the conan package manager stuff working on windows, so hopefully you'll be able to compile with dependencies.

PR #1894, opened 2020-06-01
Issue 1809: Add unit test: surface.get_shifts

nstapler:


MyreMylar:
As I expect you've noticed, `set_shifts()` doesn't work on pygame 2/SDL 2. This is documented here:

https://www.pygame.org/docs/ref/surface.html#pygame.Surface.set_shifts

It's fine to use it in part of your testing for `get_shifts()` but you will have to restrict that part of the test to SDL 1 only (we have a build that uses SDL 1 on Travis so it will still get run on the Continuous Integration at least once).

Probably the easiest way to accomplish that is to split it into two test functions - one that is  skipped on SDL 2/pygame 2 - that uses `set_shifts()` and one that just uses get_shifts on various types of created surfaces. I'd recommend creating them with bit depths of 8, 24 and 32 and maybe setting alpha on and off to see if that changes what you get back from `get_shifts()`.

Thanks for contributing!

MyreMylar:
@nstapler should I close this PR, or are you likely to pick it up again?

nstapler:
@MyreMylar Sorry, for the late response. I think you should close it.

Issue #1720, opened 2020-05-11
PyGame window stops updating after moving the window to another screen (mac)


levsa:
If I move the PyGame window from the screen it was started on to another screen, it stops updating itself. It doesn't start updating again even if moved back to the original screen. Nothing in the terminal and Esc quits the application as normally. Tried own application and `python3 -m pygame.examples.aliens`

On macOS Mojave 10.14.6 (18G4032). Graphics Radeon Pro 450 2 GB, Intel HD Graphics 530 1536 MB

illume:
Hi, thanks for the report.

I think this is fixed in pygame 2. Are you using 1.9.6?
You can use the pre-release like so: `python3 -m pip install pygame==2.0.0.dev8`

levsa:
No, I was using `2.0.0.dev6`.

Tested `2.0.0.dev8` now, the issue is still present

illume:
I can't reproduce this on my Mac 10.13.6 with python 3.7.

Can you please try with these two examples to see if they work?

```bash
python3 -m pygame.examples.aliens
```

This one prints out a few diagnostic issues, and renders in a different way to the first.

```bash
python3 -m pygame.examples.setmodescale

renderer: metal flags: 0b1010
accelerated
render to texture
```

levsa:
`pygame.examples.setmodescale` works.
`pygame.examples.aliens` doesn't work :)

Output from `pygame.examples.setmodescale`
```
 lava@3.7.2 ~/src/lava/python/lavalamp/[master] > python3 -m pygame.examples.setmodescale
pygame 2.0.0.dev8 (SDL 2.0.12, python 3.7.2)
Hello from the pygame community. https://www.pygame.org/contribute.html
desktops [(2560, 1440), (1920, 1200)]
renderer: opengl flags: 0b1010
accelerated
render to texture
```

MyreMylar:
I assume this is related to this:

https://github.com/pygame/pygame/issues/1725

galexandreg:
> I assume this is related to this:
> 
> #1725

I don't think so. The aliens example dosen't use SCALED to get a SDL renderer, so I don't see how it could be related.



> I can't reproduce this on my Mac 10.13.6 Intel Iris Pro 1536 MB, NVIDIA GeForce GT 750M with python 3.7. Maybe it's the MacOS version or ATI driver triggering the issue.
> 

[...]

PR #1897, opened 2020-06-02
Keep all window events in SDL2

robertpfeiffer:
generate *additional* compatibility events instead of mangling.
This should help with #1720 on Linux, make pygame's usage of the SDL2 API "more correct", more future-proof, and make behaviour more consistent across platforms.

Especially when minimising and maximising, additional videoexpose events are generated, which are needed for SCALED to work correctly.

@xandark, does this work for you?

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -117,7 +117,7 @@ RemovePending_PGS_VIDEORESIZE_Events(void * userdata, SDL_Event *event)
 }
 
 /*SDL 2 to SDL 1.2 event mapping and SDL 1.2 key repeat emulation*/
-static int
+static int SDLCALL
 pg_event_filter(void *_, SDL_Event *event)
 {
     /* This event filter alters events inplace.
@@ -130,9 +130,16 @@ pg_event_filter(void *_, SDL_Event *event)
                 return 1;
             case SDL_WINDOWEVENT_RESIZED:
                 if(_pg_event_generate_videoresize) {
-                    /* keep resized event around for SDL_RendererEventWatch */
+                    /* keep resized event around for SDL_RendererEventWatch
+                       (SDL2-internal) and pygame-internal event watch in
+                       display.c */
                     SDL_Event newevent = *event;
                     newevent.type = SDL_VIDEORESIZE;
+
+                    /* all previous resize events are superseded.
+                       SDL2 already does this for SDL_WINDOWEVENT_RESIZED,
+                       so we only need to filter our own custom event before
+                       we push the new one*/
                     SDL_FilterEvents(RemovePending_PGS_VIDEORESIZE_Events, &newevent);
                     SDL_PushEvent(&newevent);
                     return 1;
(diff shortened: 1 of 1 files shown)


illume:
To reproduce:
- linux (X11 or wayland?), pygame with SDL2
- python3 -m pygame.examples.aliens
- bring another window in front (quick before aliens get you)
- see garbage in aliens window

https://github.com/pygame/pygame/issues/1720#issuecomment-637094351

xandark:
@robertpfeiffer looking at your description, that sounds good to me

@illume sorry, forgot to say it was X11 that I reported the original issue upon, but yes, those steps to reproduce are correct

MyreMylar:
> @robertpfeiffer looking at your description, that sounds good to me
> 

I think robert was hoping you would be able to test his fix to see if it worked. To do this you would need to download the zip of his branch from here:

https://github.com/robertpfeiffer/pygame/tree/keep-window-expose-events

unzip it and name the directory pygame, and then build and install it by navigating to the directory of the unzipped zip running something like:

    python -m buildconfig -sdl2
    python setup.py install
    pip install . -U

Unless you already did that and I misunderstood?

xandark:
@MyreMylar ah, thank you, I didn't know the protocol, I'll follow your instructions tomorrow

illume:
There's a way to install via a commit:
```bash
python -m pip install git+git://github.com/robertpfeiffer/pygame.git@3a3c8a8b1399890e429730f5040d3755ecc31cde --user
```

Or install via a branch:
```bash
python -m pip install https://github.com/robertpfeiffer/pygame/archive/keep-window-expose-events.zip --user
```

xandark:
Okay, I managed to build the special dev11 branch via the `python -m pip install https://github.com/robertpfeiffer/pygame/archive/keep-window-expose-events.zip --user` command, thank you, that worked great for me.

But the issue remains when testing on X11; it didn't fix it. In other words, bringing another window in front of the Aliens example and then bringing Aliens to the front again does not cause Aliens' window to repaint.

MyreMylar:
This seems more correct to me, even if it doesn't resolve the multi-monitor mac issue so I'm going to approve it on general principles.

illume:
> even if it doesn't resolve the multi-monitor mac issue so I'm going to approve it on general principles.

Was this tested by someone as well? I saw xandark above tested on X11.

MyreMylar:
> > even if it doesn't resolve the multi-monitor mac issue so I'm going to approve it on general principles.
> 
> Was this tested by someone as well? I saw xandark above tested on X11.

Nope, I'm just getting my bugs confused because Xandark posted about the issue in that issue thread here:
https://github.com/pygame/pygame/issues/1720

Sadly, I have only one monitor and 0 macs so I can neither confirm or deny that this PR does anything for them.

robertpfeiffer:
I think @xandark might be having a different problem altogether - neither this one, *nor* #1720 on mac. As long as pygame.examples.setmodescale works on KDE neon (with the white circle moving), we can merge this. It works on Linux on my machines.

PR #1909, opened 2020-06-03
Unit tests for time.Clock.get_time()

### LINKED ISSUE #1787: Add unit test: time.Clock.get_time()

**Test stub:** https://github.com/pygame/pygame/blob/master/test/time_test.py#L50-L72
**Docs:** https://www.pygame.org/docs/ref/time.html#pygame.time.Clock.get_time
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/time.c#L367-L372

pedrodelapena:
Closes #1787

### DIFF
--- a/test/time_test.py
+++ b/test/time_test.py
@@ -48,29 +48,33 @@ class ClockTypeTest(unittest.TestCase):
 
         self.fail()
 
-    def todo_test_get_time(self):
+    def test_get_time(self):
+        #Testing parameters
+        delay = 0.1 #seconds
+        delay_miliseconds = delay*(10**3)
+        iterations = 10
+        delta = 50 #milliseconds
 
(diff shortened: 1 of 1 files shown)


MyreMylar:
I'm slightly nervous about reducing the error delta as we just had two time module tests fail on CI today because their timing error delta was too small - but perhaps it'll be fine.

pedrodelapena:
> I'm slightly nervous about reducing the error delta as we just had two time module tests fail on CI today because their timing error delta was too small - but perhaps it'll be fine.

I'll revert the changes on the delta parameter and I'll squash the commits. The error delta I originally used was too small while the delay I used was too big, which lead me to increase the first later on. I hope this change gets it all right!

MyreMylar:
LGTM 👍

PR #1911, opened 2020-06-04
Add unit test: display.get_driver()

### LINKED ISSUE #1744: Add unit test: display.get_driver()

**Test stub**: https://github.com/pygame/pygame/blob/master/test/display_test.py#L106-L121
**Docs**: https://www.pygame.org/docs/ref/display.html#pygame.display.get_driver
**Code to be tested**: https://github.com/pygame/pygame/blob/master/src_c/display.c#L694-L723

gabsmoreira:
Add unit test: display.get_driver() and closes #1744

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -699,6 +699,7 @@ static PyObject *
 pg_get_driver(PyObject *self, PyObject *args)
 {
     const char *name = NULL;
+    VIDEO_INIT_CHECK();
     name = SDL_GetCurrentVideoDriver();
     if (!name)
         Py_RETURN_NONE;
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -105,21 +105,42 @@ class DisplayModuleTest(unittest.TestCase):
 
     def test_get_driver(self):
         drivers = [
-            'windib',
-            'directx',
-            'x11',
-            'dga',
-            'fbcon',
-            'directfb',
-            'ggi',
-            'vgl',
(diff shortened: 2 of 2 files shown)


gabsmoreira:
Thanks for your quick reply @MyreMylar, I added more drivers and tested `get_driver()` after `display.quit()` as requested.

MyreMylar:
The CI test is failing on that one specific build because that is our SDL 1 build and this function has a slightly different definition for SDL 1 and SDL 2.

Right now on SDL 1 the function calls the VIDEO_INIT_CHECK() macro before doing anything:

https://github.com/pygame/pygame/blob/c10a68e34c1357c5ab7cacf0bb815fc39faa6428/src_c/_pygame.h#L170-L172

which as you can see raises an assert if the display is not initialised. 

I think the best fix for this would likely be to add the `VIDEO_INIT_CHECK()` to the sdl2 version of this function here:

https://github.com/pygame/pygame/blob/9e8fedc9c2b002d85f04de5b7d31f9891262005e/src_c/display.c#L697-L706

In basically the same place as it is in the sdl1 version seen here:

https://github.com/pygame/pygame/blob/9e8fedc9c2b002d85f04de5b7d31f9891262005e/src_c/display.c#L741-L749

And then change the test to check this assert is being being raised using something like:

    display.quit()
    with self.assertRaises(errorTypeHere):
         driver = display.get_driver()

PR #1912, opened 2020-06-04
Add unit test: joystick.quit()

### LINKED ISSUE #1784: Add unit test: joystick.quit()

**Test stub:** https://github.com/pygame/pygame/blob/master/test/joystick_test.py#L76-L88
**Docs:** https://www.pygame.org/docs/ref/joystick.html#pygame.joystick.quit
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/joystick.c#L67-L72

raphacosta27:
Add unit test: joystick.quit() and closes #1784

### DIFF
--- a/test/joystick_test.py
+++ b/test/joystick_test.py
@@ -101,19 +101,17 @@ class JoytickModuleTest(unittest.TestCase):
 
         self.fail()
 
-    def todo_test_quit(self):
+    def test_quit(self):
+        """Test if joystick.quit works."""
+        
+        pygame.joystick.init()
 
-        # __doc__ (as of 2008-08-02) for pygame.joystick.quit:
+        self.assertIsNotNone(pygame.joystick.get_count()) #Is not None before quit
 
-        # pygame.joystick.quit(): return None
-        # uninitialize the joystick module
-        #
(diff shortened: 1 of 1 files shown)


MyreMylar:
I think this is about all that we can test for on Continuous Integration because the machines won't have a gamepad plugged in.

I do think the module is also going to need some interactive tests that don't run on CI and ask you to plug in a controller before beginning and there we could also test that any plugged in controllers are also quit when quit() is called.

Not worth worrying about that for now. I'll raise another issue and approve this.

raphacosta27:
> I think this is about all that we can test for on Continuous Integration because the machines won't have a gamepad plugged in.
> 
> I do think the module is also going to need some interactive tests that don't run on CI and ask you to plug in a controller before beginning and there we could also test that any plugged in controllers are also quit when quit() is called.
> 
> Not worth worrying about that for now. I'll raise another issue and approve this.

Yeah you're right! I also assigned myself to the test for joystick.get_count() and realized that this tests are actually just confirming that the function itself is working. To really test if the joystick library is not broken, the best way would be with the interactive tests you mentioned. Or, in some way, we could simulate joystick plugs.

PR #1914, opened 2020-06-04
Increase time.delay() error delta to 50

MyreMylar:
Based on this CI fail here:
https://ci.appveyor.com/project/pygame/pygame/builds/33301393/job/jw5s7t6yrf9w01l9
 

    ==============================================================
    FAIL: test_delay (pygame.tests.time_test.TimeModuleTest)
    Tests time.delay() function.
      File "C:\Python35-x64\lib\site-packages\pygame\tests\time_test.py", line 144, in test_delay
        self._wait_delay_check(pygame.time.delay, millis, iterations, delta)
      File "C:\Python35-x64\lib\site-packages\pygame\tests\time_test.py", line 210, in _wait_delay_check
        self.assertAlmostEqual(wait_time, millis, delta=delta)
    AssertionError: 88 != 50 within 5 delta

delta here represents the worst case error between requested delay and actual delay measured by time.time().

### DIFF
--- a/test/time_test.py
+++ b/test/time_test.py
@@ -139,7 +139,7 @@ class TimeModuleTest(unittest.TestCase):
         """Tests time.delay() function."""
         millis = 50  # millisecond to wait on each iteration
         iterations = 20  # number of iterations
-        delta = 5  # Represents acceptable margin of error for wait in ms
+        delta = 50  # Represents acceptable margin of error for wait in ms
         # Call checking function
         self._wait_delay_check(pygame.time.delay, millis, iterations, delta)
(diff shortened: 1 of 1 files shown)


MyreMylar:
Saw a second fail for this test with an error of 11 so it wasn't just a one off. It looks like `time.delay()` still has a fairly inaccurate worst case error even if it's average case appears to be better than `time.wait()`

illume:
Yeah. A shared overloaded CI based on containers running in VMs is probably the worst case for timing functions. It could potentially take a long time for the process to get a time slice.

Issue #1910, opened 2020-06-03
surface.set_at() inverting color positions on a png surface


galexandreg:
### Problem
_Maybe this is "normal". I don't know much about SDL surface internals... But it seems odd that get_at() and set_at() don't behave the same._

**Setting individual pixel colors on a surface created from a png file will invert color positions. (BGR, not RGB)**
get_at() will return colors from the png as RGB, but when setting a color with set_at() and reading it back with get_at(), they will be inverted.
It works fine with a "normal" surface.

### Test
```
import pygame

pygame.init()

image = pygame.image.load("test_get_set.png")
surface = pygame.Surface((500, 500))
surface.fill((0, 0, 0))


print("With the png (red, green, blue):")
print(image.get_at((0, 0)))
print(image.get_at((0, 200)))
print(image.get_at((0, 400)))
print()
print("Set pixels to red, green, blue with set_at:")
image.set_at((0, 0), (255, 0, 0))
image.set_at((0, 200), (0, 255, 0))
image.set_at((0, 400), (0, 0, 255))
print(image.get_at((0, 0)))
print(image.get_at((0, 200)))
print(image.get_at((0, 400)))
print()
print("Normal surface :")
print("Set pixels to red, green, blue with set_at:")
surface.set_at((0, 0), (255, 0, 0))
surface.set_at((0, 200), (0, 255, 0))
surface.set_at((0, 400), (0, 0, 255))
print(surface.get_at((0, 0)))
print(surface.get_at((0, 200)))
print(surface.get_at((0, 400)))
```
The test png file:
![test_get_set](https://user-images.githubusercontent.com/45190261/83690799-6a5c6c80-a5bf-11ea-9118-d9aff9b4927f.png)

illume:
Interesting. I wonder if this is related to the solarwolf png issue.

MyreMylar:
Modified the test slightly so you can see the error visually and printed out some more info:

    import pygame
    
    pygame.init()
    
    window_surf = pygame.display.set_mode((500, 500))
    
    image = pygame.image.load("test_get_set.png")
    print("bitsize:", image.get_bitsize())
    print('shifts:', image.get_shifts())
    
    
    print("With the png (red, green, blue):")
    print(image.get_at((0, 0)))
    print(image.get_at((0, 200)))
    print(image.get_at((0, 400)))
    print()
    print("Set pixels to red, green, blue with set_at:")
    image.set_at((10, 10), pygame.Color(255, 0, 0))
[...]

PR #1919, opened 2020-06-05
Remove hard coded RGB bitshifts that don't work with SDL_MapRGB()/SDL_MapRGBA() 

MyreMylar:
I think there are some places in the code where these functions are being used interchangeably, or perhaps code has been updated from GetRGB() to MapRGB(). Anyway, it looks like GetRGB would return data that was always in RGB order but MapRGB() returns colour data that is in the byte order of whatever format was passed in when it was called.

This means we have to be careful when using hardcoded bit shifts as we are in several places. Where GetRGB() is being used it's fine as the output is always the same but with MapRGB we could have different output from different input surfaces. At some point SDL started treating PNGs as Big Endian (which is correct) so they are now loaded in BGR order. When we map colours to them they also end up in BGR order, but then in some places we are assuming the colours are in RGB order when trying to stick them back into the surface after modifying them. 

Related to (& fixes): 
https://github.com/pygame/pygame/issues/1910


MyreMylar:
There is one more similar pattern in transform.c but it's called with an SDL_GetRGB() generated colour so should be fine.

MyreMylar:
> Very good! We probably should have tests in there for this.

Yes, I was thinking that this morning. If I shrink down the original .png from the issue it should make a good test case.

MyreMylar:
The other function this affects is 'laplacian'. Not super certain how to test it.

MyreMylar:
Figured out the last test. 

All three of the new tests fail on current master, but pass with this PR.

PR #1921, opened 2020-06-05
Pylint fixes for locals.py

MyreMylar:
I put all the imported constants into a static all block that pylint can understand rather than a dynamically generated one. 

If we (soon-ish) have pylint-ing enabled on check-in it should then be impossible to miss adding a constant here because pylint will automatically report it as an error.

Zero pylint errors after this PR using:

`pylint src_py/locals.py --extension-pkg-whitelist=pygame`


MyreMylar:
One upside(?) of declaring the constants like this is that it jogged loose which ones were undefined when building with SDL 1. 

I added the missing ones, declared as `-1` which seems to be the approach taken with the other missing constants to date.

PR #1922, opened 2020-06-05
Force LGTM to grab the latest compiled version of the pygame for analysis

MyreMylar:
Experiment to see if a config file will make LGTM grab the c extensions from 2.0.0.dev10.

**Edit:** it works

### DIFF
--- a/buildconfig/pygame-stubs/overlay.pyi
+++ b/buildconfig/pygame-stubs/overlay.pyi
@@ -1,0 +1,13 @@
+from typing import Optional, Tuple
+
+from pygame.rect import Rect
+
+
+class Overlay(object):
+    def __init__(self, format: int, size: Tuple[int, int]) -> None: ...
+
+    def display(self, pixel_data: Optional[Tuple[bytes, bytes, bytes]]) -> None: ...
+
+    def set_location(self, rect: Rect) -> None: ...
+
+    def get_hardware(self, rect: Rect) -> int: ...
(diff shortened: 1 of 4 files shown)


MyreMylar:
That gives a slightly more accurate picture of the current state of pygame LGTM alerts. I think before it was using pygame 1.9.6 for the c extensions.

MyreMylar:
Guess we'll just have to update it to the latest release each time, or at least a close enough one to catch c extension/dynamically built python file fixes/alerts.

As an aside it looks like this fixes the `__all__` alert for version.py, but still doesn't catch the `__all__` in constants.c which should be present in dev10 so I guess there is still a mystery to puzzle out there.

MyreMylar:
Just thought I'd try another experiment with a .pyi file to see if I could determine whether they get picked up when committing to LGTM, or whether they are only grabbed from the wheel.

MyreMylar:
Looks like the .pyi files for the c extensions are likely also only grabbed from the wheel.

MyreMylar:
Looks like that works on those 'wrong number of arguments' errors - interesting.

MyreMylar:
I think perhaps the errors LGTM is picking up might be caused by this block here:

    
    try:
        from pygame.surface import Surface, SurfaceType
        Surface = lambda: Missing_Function
    
    
    try:
        import pygame.mask
        from pygame.mask import Mask
        Mask = lambda: Missing_Function
    
    try:
        from pygame.pixelarray import PixelArray
        PixelArray = lambda: Missing_Function
    
    try:
        from pygame.overlay import Overlay
        Overlay = lambda: Missing_Function

So I'm going to try and replace it with something similar to MissingModule and see if that helps.

MyreMylar:
Think this idea should wait until the next release after we've cleaned up the .py files some more, including `__init__.py` to see if that helps.

It's interesting to know though.

Issue #1785, opened 2020-05-18
Add unit test: time.Clock.get_fps()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)
[Draft guide on how to write a unit test for pygame](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/time_test.py#L14-L35
**Docs:** https://www.pygame.org/docs/ref/time.html#pygame.time.Clock.get_fps
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/time.c#L360-L365

MyreMylar:
I think we have enough unit tests in the works for this one now.

illume:
Seems like the test for this is failing sometimes in the appveyor build robot.  
https://ci.appveyor.com/project/pygame/pygame/builds/33439459/job/3finofao4jei308n

Probably the delta needs to be increased?

```
======================================================================
FAIL: test_get_fps (pygame.tests.time_test.ClockTypeTest)
test_get_fps tests pygame.time.get_fps()
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python37\lib\site-packages\pygame\tests\time_test.py", line 26, in test_get_fps
    self._fps_test(c, 60, delta)
  File "C:\Python37\lib\site-packages\pygame\tests\time_test.py", line 36, in _fps_test
    self.assertAlmostEqual(clock.get_fps(), fps, delta=fps*delta)
AssertionError: 38.61003875732422 != 60 within 12.0 delta (21.38996124267578 difference)

```

PR #1923, opened 2020-06-05
Added unit test for Issue #1785

SabrinaSimao:
Thanks for the quick reply!


MyreMylar:
Somehow this test got three submissions in a short time period!

In the end I went with the most recent one because it has the most checks:
https://github.com/pygame/pygame/pull/1926

Thank you for your contribution though, it helped in reviewing the tests to compare three takes on the same idea. Some people have been asking to be assigned an issue before working on it, and I guess that would be a way to avoid this unfortunate & unlikely occurrence happening again?

PR #1925, opened 2020-06-05
A start on linting the sprite module

MyreMylar:
**Warnings disabled = 5:**

1. `layer=None  # noqa pylint: disable=unused-argument; supporting legacy derived classes that override in non-pythonic way`
2. `return self.__class__(self.sprites()) # noqa pylint: disable=too-many-function-args; needed because copy() won't work on AbstractGroup`
3. `class collide_circle_ratio(object):  # noqa pylint: disable=invalid-name; this is a function-like class`
4. `class collide_rect_ratio:  # noqa pylint: disable=invalid-name; this is a function-like class`
5. `def draw(self, surface, bgd=None):  # noqa pylint: disable=arguments-differ; unable to change public interface`

--------------------------------------------------------

**Remaining issues after this PR**:

    src_py\sprite.py:1:0: C0302: Too many lines in module (1735/1000) (too-many-lines)
    src_py\sprite.py:69:2: W0511: TODO:  a group that holds only the 'n' most recent elements. (fixme)
    src_py\sprite.py:1097:4: R0914: Too many local variables (16/15) (too-many-locals)
    ------------------------------------------------------------------
    Your code has been rated at 9.95/10 (previous run: 9.97/10, +0.00)

Apart from the TODO which I'm ignoring, the remaining issues are:

 1. The module is too large. This is a historic problem. If I was coding it from the beginning i'd probably put all the classes into separate files (making them technically mini-modules) and then pull them all together into a sprite module with an `__init__.py` or something. This isn't pygame's general coding style for modules though so I'm not sure whether we should just leave it as it is and disable the warning or divide up the module in some other way...
2. too many locals in LayeredDirty's draw function. I've left this error in for now in case someone wants to revisit this method again in the future. The problem is that having lots of local variables actually speed this function up a bit so while it is relatively easy to get rid of them it costs in performance. However, there may be some broader issues with the whole technique switching between two drawing modes that it's hard to assess that properly right now on my main development platform (windows) because display.update() currently doesn't function in the same way it did originally in SDL 2.0.12 on windows. I also felt that re modelling this method if it was going to be done was probably an issue for a more focused PR.


MyreMylar:
If anyone has any ideas for the last two warnings I'd like them.

illume:
Reasons can be added after `;` if you want.

```python
class Sprite(object): # pylint: disable=useless-object-inheritance; required for python 2
```

src_py/sprite.py
```diff
@@ -1597,14 +1659,13 @@ def spritecollideany(sprite, group, collided=None):
...
-        for s in group:
-            if collided(sprite, s):
-                return s
+    if collided is not None:
+        for group_sprite in group:
+            if collided(sprite, group_sprite):
+                return group_sprite
     else:
         # Special case old behaviour for speed.
-        spritecollide = sprite.rect.colliderect
```
illume, line 1606: This small optimization was removed here.

src_py/sprite.py
```diff
@@ -1523,25 +1582,26 @@ def spritecollide(sprite, group, dokill, collided=None):
...
+                if sprite.rect.colliderect(group_sprite.rect):
+                    group_sprite.kill()
+                    append(group_sprite)
 
         return crashed
 
-    elif collided:
-        return [s for s in group if collided(sprite, s)]
-    else:
-        spritecollide = sprite.rect.colliderect
```
illume, line 1543: optimization removed here

src_py/sprite.py
```diff
@@ -1432,6 +1482,11 @@ def __init__(self, ratio):
         """
         self.ratio = ratio
 
+    def __repr__(self):
+        """
+        Turn the class into a string.
+        """
+        return 'ratio: ' + str(self.ratio)
```
illume, line 1489: This doesn't follow a standard repr format. Is repr useful here?

src_py/sprite.py
```diff
@@ -1040,131 +1066,137 @@ def draw(self, surface, bgd=None):
...
-        _update[:] = []
+        self.lostsprites[:] = []
 
         # -------
         # restore original clip
-        _surf.set_clip(_orig_clip)
+        surface.set_clip(_orig_clip)
         return _ret
 
+    @staticmethod
```
illume, line 1128: Nice improvement moving these to separate functions.

I haven't read through these two to see if they were refactored correctly however.

src_py/sprite.py
```diff
@@ -66,41 +66,37 @@
...
+# function, instead of just a background surface. the function
+# would take a surface and a rectangle on that surface to erase.
+#
+# perhaps more types of collision functions? the current two
+# should handle just about every need, but perhaps more optimized
+# specific ones that aren't quite so general but fit into common
+# specialized cases.
 
-import pygame
-from pygame import Rect
```
illume, line 90: I was wondering if it's worth binding Rect here for speed, but I guess anywhere it's used in a loop it should just bind it to a local first.

src_py/sprite.py
```diff
@@ -656,13 +685,13 @@ def add_internal(self, sprite, layer=None):
 
         if layer is None:
             try:
-                layer = sprite._layer
+                layer = sprite._layer  # noqa pylint: disable=protected-access, explicitly trying to break protected access
```
illume, line 688: `_layer` really probably should have been a public variable. Not really sure how to do that in a backwards compatible way. Maybe by making `_layer` a property alias to `layer`.

src_py/sprite.py
```diff
@@ -1040,131 +1066,137 @@ def draw(self, surface, bgd=None):
...
-        _surf.set_clip(_orig_clip)
+        surface.set_clip(_orig_clip)
         return _ret
 
+    @staticmethod
+    def _draw_dirty_internal(_old_rect, _rect, _sprites, _surf_blit,
+                             _update):
+        for spr in _sprites:
+            if spr.dirty < 1:
+                if spr.visible:
```
illume, line 1133: These two if blocks could be collapsed?
```
if spr.dirty < 1 and spr.visible:
elif spr.visible:
```

src_py/sprite.py
```diff
@@ -1040,131 +1066,137 @@ def draw(self, surface, bgd=None):
...
+                    # sprite not dirty; blit only the intersecting part
+                    if spr.source_rect is not None:
+                        # For possible future speed up, source_rect's data
+                        # can be pre-fetched outside of this loop.
+                        _spr_rect = _rect(spr.rect.topleft,
```
illume, line 1138: Why are these variables with `_` in front?

src_py/sprite.py
MyreMylar: I added those quickly because pylint won't let you can't have a function with only one public method. 

It being `__repr__` isn't particularly important, just needs to be something and that seemed like it might be marginally useful. I'll try to make them have the proper format like the other examples in the sprite module.

src_py/sprite.py
MyreMylar: Huh, I'd never heard of assigning functions to temporary variables being a python optimisation before. Looking it up though it appears to be reported several times on the internet.

I made this change because pylint starts to complain about temporary variables in a function if you have more than some limit (7?). I'll change it back and see if we are still under that limit, if not I guess we might just have to disable that warning.

src_py/sprite.py
MyreMylar: That seems to have been the coding style of whoever originally wrote the LayeredDirty class, it's just been preserved in my refactor. I don't think the leading underscore has any meaning for temporary variables in a function.

src_py/sprite.py
MyreMylar: looks like it 👍

src_py/sprite.py
MyreMylar: or perhaps to a double underscore `__layer` ?

I'd vaguely like to turn all the public variables of the Sprite and DirtySprite class into properties, but perhaps that'd have some impacts on speed.

src_py/sprite.py
MyreMylar: :thumbs:

I guess performance tests for the sprite module would be useful for refactorings.

src_py/sprite.py
nthykier: Sadly, it gets even more complicated with Cython where such an assign _defeats_ optimization (it can cause a C-callable function to degenerate into a python call).  As I understand it, Python 3.9 will hopefully make this optimization much less needed (with its optimization of the `LOAD_GLOBAL` instruction, which _tends_ to be the slowest part)

src_py/sprite.py
nthykier: If speed is relevant, then I think we should use Cython for the sprite classes instead of pure python - especially if we make `pygame.Rect` C-callable for Cython (I do not remember if it is already)

MyreMylar:
- I added back the functions I previously removed that were pulled in as local variables, a few didn't need to to return as they are now passed in to functions as parameters which will also make them local variables inside the function. 

- Changed the two new `__repr__` methods to be like @illume suggestion.

- Fixed the pylint disable reason comments so they actually work.

- Disabled the draw function `arguments-differ` warning as well since I think that one is out of our hands.

- Updated the first post.

**To Do**:

- consider turning `_layer` into a property and making the variable `__layer`

src_py/sprite.py
MyreMylar: I think Cythoniz-ing concerns are probably beyond the scope of this PR.

src_py/sprite.py
MyreMylar: I checked and it looks like Rect is bound to a local variable/function argument prior to any time it is used in a loop.

MyreMylar:
Added `layer` as a read-only property on Sprite and DirtySprite. This is actually consistent with the DirtySprite docs, weirdly, which seem to written as if layer existed as a read-only property already. 

The property uses `gettattr` for dynamic property access for the spite class which doesn't define the `._layer` variable normally, but it uses regular access for DirtySprite, which does.

I also edited the tests slightly because they seemed to be directly assigning values fruitlessly to `layer` in two places which up to now did nothing of any use and now raises an AttributeError as the property is genuinely read only rather than non-existent.

It's probable that there is other code out there right now directly assigning values to  `.layer` which will now raise attribute errors where it didn't before; but I think that is a good thing because that code is currently not doing anything useful as the intended behaviour is that only groups should be setting the layer value of sprites.

Where the groups set the layer of a sprite they now use `setattr` to set the protected variable directly, I think this is the only way to keep the intended interface where only groups should set the layer property of sprites, the `_layer` variable is still protected and also also avoid pylint warnings.

Let me know if I'm wrong or if this is too controversial.

illume:
Awesome. I haven't had a chance to read it yet, but I will soon.

illume:
Thinking about a couple of things...

So that the `layer` attribute is the fast one, should we reverse which one is the property? 
- `_layer` as a property for backwards compatibility
- `layer` as a normal attribute

Should layer be read only?

src_py/sprite.py
```diff
@@ -66,41 +66,37 @@
...
 from operator import truth
 
-# Python 3 does not have the callable function, but an equivalent can be made
-# with the hasattr function.
-if 'callable' not in dir(__builtins__):
-    callable = lambda obj: hasattr(obj, '__call__')
+import pygame
+
 
 # Don't depend on pygame.mask if it's not there...
```
illume, line 93: At this point we can probably just assume pygame.mask is there.

src_py/sprite.py
```diff
@@ -215,7 +220,20 @@ def alive(self):
         return truth(self.__g)
 
     def __repr__(self):
-        return "<%s sprite(in %d groups)>" % (self.__class__.__name__, len(self.__g))
+        return "<%s sprite(in %d groups)>" % (self.__class__.__name__,
```
illume, line 223: Lowercase sprite here was an old mistake. 
- Changing it now might break some doctest based tests, but that's about it.
- Fixing it would make it slightly more readable.

So, probably worth fixing to `<Sprite...>`.

illume:
Here's some scripts to test with.

```bash
python3 -m pygame.examples.testsprite
python3 -m pygame.examples.testsprite -layered_dirty
python3 -m pygame.examples.testsprite -layered_dirty 5000
python3 -m pygame.examples.testsprite -static
python3 -m pygame.examples.testsprite -noupdate_rects
python3 -m pygame.examples.aliens
python3 -m pygame.examples.chimp
python3 -m pygame.examples.mask
```

illume:
Ran a few tests:

`timeout 20s python3 -m pygame.examples.testsprite -layered_dirty 5000`
- main: 44.853084, 45.242989, 46.128330, 48.962511, 47.106263
- this pr: 44.677747, 46.851475, 45.097078, 44.756806, 42.854572

`timeout 20s python3 -m pygame.examples.testsprite`
- main: 594.032699, 705.467832, 552.540765, 563.473420, 566.394236
- this pr: 451.268912, 559.743952, 567.836148, 557.078712, 566.515386

MyreMylar:
> Thinking about a couple of things...
> 
> So that the `layer` attribute is the fast one, should we reverse which one is the property?
> 
> * `_layer` as a property for backwards compatibility
> * `layer` as a normal attribute

No, I think that would probably be more confusing because the leading underscore generally indicates a protected variable only to be used inside the class or for inherited classes. Generally users won't be using '.layer' or '._layer' in performance critical code anyway because it is designed to just be a read only indication of the sprite's layer, changing the layer variable doesn't change a sprite's actual layer in a group so there is no point setting the layer variable in a tight update loop or anything like that. Actual layer management is done through the groups which have layers and the layer property of a sprite is a convenience thing for reading the current layer of a sprite.

The whole concept of a sprite having a layer variable is a little bit odd as sprites can be in multiple groups. If a sprite is present in multiple groups with layers, on a different layer in each group - what value should the sprite's layer (.layer or ._layer) variable have? Usually it doesn't come up because people don't use multiple groups that way, but it is still a little odd.

> Should layer be read only?

Yes, given the current way the code works - because otherwise it creates confusion when you set the layer of a sprite and it doesn't change layers in (one of) its groups. The restriction makes sense in the docs and given how layers actually work in the code, the change over to making it a property just enforces things. Making sure that people incorrectly assigning a value to a sprite's layer will now get a warning that what they are doing won't actually do anything. I'm still sure it'll cause someone's code somewhere to raise an error because a lot of people make a mess when they are coding.

A more dramatic change would be to change the `.layer` property to have a setter pick a group the sprite is assigned to (the first one? probably all of them?) and then try calling `.set_layer()` on it with the new layer (I guess it will silently fail if the group doesn't have a layer, because otherwise you'd get lots of errors using multiple groups in fairly normal ways). Probably pylint will moan about that though because the base group type doesn't have a `.set_layer()` method so it'd be another change to add that to all groups even if they don't have layers or something like that. May also break code but in different ways to the current commits. I think if I was designing it again I would have just enforced having a `LayeredSprite` in a `LayeredSpriteGroup` rather than the current situation where any sprite can go in any group, even if they don't really work quite right in that type of group properly without a bunch of hacks.

I think the way I've done it makes the most sense that can be made of it under the current code logic but these are definitely questions to ponder when (if?) there is a sprite redesign after 2.0.

MyreMylar:
> Ran a few tests with python 3.8... TLDR; benchmarking is hard.

Looks like there isn't any dramatic change in performance with this benchmark which I guess is a win?

src_py/sprite.py
MyreMylar: 👍

src_py/sprite.py
MyreMylar: 👍

MyreMylar:
Testing all these:

> python3 -m pygame.examples.testsprite
> python3 -m pygame.examples.testsprite -layered_dirty
> python3 -m pygame.examples.testsprite -layered_dirty 5000
> python3 -m pygame.examples.testsprite -static
> python3 -m pygame.examples.testsprite -noupdate_rects
> python3 -m pygame.examples.aliens
> python3 -m pygame.examples.chimp
> python3 -m pygame.examples.mask

After lots of test runs they all seem basically the same except the LayeredDirty ones which seem a teeny bit slower in the new version versus the old mega function. 

This one:
`python3 -m pygame.examples.testsprite -layered_dirty 5000` 

Seems to run at about 21 FPS on average with the current mega draw() function, but at about 20 FPS on average with the new split into two arrangement. It's tricky to be 100% with a small change like this but it seems fairly consistent so far.

I'm not sure exactly what is causing the change so far, other than that it is in the LayeredDirty draw() method. I tried a few small things but didn't have much luck (the latest commit reintroduces a too-many-locals pylint error trying to chase that missing frame). It may be I have to abandon refactoring this one and just revert to the old slightly faster mess for now to get this checked in. I'll poke at it again tomorrow for a bit with a fresh brain and some biscuits.

Everything else seemed fine though.

src_py/sprite.py
```diff
@@ -853,7 +895,7 @@ def change_layer(self, sprite, new_layer):
             mid += 1
         sprites.insert(mid, sprite)
         if hasattr(sprite, 'layer'):
-            sprite.layer = new_layer
+            setattr(sprite, '_layer', new_layer)
```
illume, line 898: Is this correct? Before it was assigning to `layer` and now to `_layer`.

illume:
Probably it's worth committing your cleaner version, over the messy one IMHO. If we can't detect a difference at the macro level, I think it's fine :)

On the other hand, if you want to continue... It could be worth doing a benchmark at the method level with just the one you found a bit slower. Perhaps even do some profiling to see if you notice anythin.

src_py/sprite.py
MyreMylar: It's a little confusing because this is the only place in the original *code* that actually uses `.layer` as opposed to `._layer` so I think it was actually a bug. Made more confusing by the docs which interchangeably refer to both `.layer` and `._layer`. 

Poking through the docs again I believe it was intended that users could set the `._layer` value of a sprite **before** it is added to a group and that value would then be respected when the sprite was added, but then after entering a group the `._layer` should be read only and not settable for users. 

I suppose I could try to encode that idea into a setter for the new `.layer` property? So that it checks to see if the sprite is in any group first, and if so refuses to set the layer (as it works now) but otherwise allows it

I suspect the person/persons writing the code and docs was very confused at some point about `._layer` and `.layer` or they originally had it one way and then changed it to the other, as right now the current state is very confusing. See this from the `LayeredUpdates` class docs:

```
This group is fully compatible with pygame.sprite.Sprite class for visible game objects..

You can set the default layer through kwargs using 'default_layer' and an integer for the layer. The default layer is 0.

If the sprite you add has an attribute `_layer` then that layer will be used. If the **kwarg contains 'layer' then the sprites passed will be added to that layer (overriding the `sprite.layer` attribute). If neither sprite has attribute `layer` nor **kwarg then the default layer is used to add the sprites.
```

I mean they switch between talking about `._layer` and `.layer` from one sentence to the next . The code was actually using `._layer` not `.layer` everywhere sprites are added so setting `.layer` would do basically nothing in the current code - except that line you highlight would be activated meaning when you changed a sprite's layer (but not when you originally add it) the sprite's new `.layer` attribute would be kept up to date.

My feeling is that this code was originally written using `.layer` and then at some point in its development it was decided to change to `._layer` and several places in the docs were missed and this one place in the code. Though I could be mistaken and there is some other more confusing purpose for this. 

TLDR

- Original code is slightly confused between `._layer` (used the most times in the code and thus the most functional attribute) and `.layer` (used once in the code in a way that is mainly cosmetic).
- Original docs are confused as well using `._layer` and `.layer` seemingly interchangeably to describe behaviour that is governed solely by `._layer`.
- Making things even more confusing there are two sprite classes. One called `DirtySprite` defines `._layer` by default and the other, called `Sprite` does not, but both can be added to groups with layers in and in some cases adding a regular Sprite to a layered group will set the `._layer` property on the Sprite but not always.
- New code somewhat resolves this ambiguity by leaving `._layer` in place but adding a python property called `.layer` that returns the value of `.layer`. Right now it is read only which is consistent with the way the code works once a sprite is in a group.

**To Do**

- I think I'm going to change both of the `if hasattr(sprite, 'layer'):` lines to read `if hasattr(sprite, '_layer'):` as I think that will more accurately preserve the spirit of the original code where you could potentially have a regular Sprite object added to a layered group that didn't itself store any layer data, so long as you specified a layer when adding it to the group. This will also resolving the original bug where the code was confused here between `._layer` and `.layer`.
- I will probably also add a setter to the new `.layer` properties that sets `._layer` so long as the Sprite or DirtySprite is not already in a group. Once they are in a group they will become read only an produce an error if you try to set them.

src_py/sprite.py
```diff
@@ -226,10 +226,24 @@ def layer(self):
...
         :return: layer as an int, or raise AttributeError.
         """
         return getattr(self, '_layer')
 
+    @layer.setter
+    def layer(self, value):
+        if not self.alive():
+            setattr(self, '_layer', value)
+        else:
+            raise AttributeError("Can't set layer directly after "
```
illume, line 242: Did you consider changing the layer in each group here?

Something like...
```python
[group.change_layer(self, value) for group in self.groups]
```

src_py/sprite.py
illume: Yes, I see. Thanks for the description.

src_py/sprite.py
MyreMylar: Yeah, I discussed it a bit [further above](https://github.com/pygame/pygame/pull/1925#issuecomment-647468318). One implementation difficulty is that not all types of groups have layers so the change_layer() method doesn't exist on those groups and will throw an error with an implementation like the above. A fairly common use case might be where people have some subset of sprites in a layered group, but also have all their sprites in one single basic mega-group for convenience. Probably we could work around that with a more complicated setter.

It's also a slightly more dramatic functionality change which I'm mostly steering clear of in these linting passes, and I guess we'd also need to update the docs to reflect that you can now change the layer of a sprite by setting it's layer. I'm sort of ambivalent on it for now, unless you think it's super useful?

MyreMylar:
> On the other hand, if you want to continue... It could be worth doing a benchmark at the method level with just the one you found a bit slower. Perhaps even do some profiling to see if you notice anything.

I poked at this a little and noticed that right now with both the current layered_dirty tests the draw method goes down the slower rendering path which seemed wrong based on the intentions. A little investigation later showed that this was partly because `get_ticks()` was **always** returning 0. 

That turned out to be because the testsprite example was not calling `pygame.init()`. Once that was re-enabled, it was easy to see that the layered_dirty 5000 test was now fighting constantly between the two drawing paths, as the simpler one was always faster than the threshold it would then trip the code that tells it to go down the slower path which would then trip the code that tells it to go back to the faster path. I assume at some point the default threshold values for the dirty flag technique were faster at an 80FPS threshold so this arrangement made sense but right now it doesn't, at least not on my PC.

Anyway since that was obviously bad, the easiest quick fix without tearing down the whole house is just to raise the default threshold a chunk for when it switches to the other (currently slower) path. With that change this test:
```
python3 -m pygame.examples.testsprite -layered_dirty 5000
```

Now runs twice as fast after the refactor. 

I expect we should add to one of those issues discussing a new sprite class whether the Dirty flag technique is still useful at all, and if it is we need a test that demonstrates its worth because this one with current blitting speeds seems to show that so far the time spent setting it up costs more than the time saved in all circumstances, at least in SDL2. I'll have a check with SDL1 as well to see if it makes a difference.

MyreMylar:
Looks like it is always faster to use the non-dirty rect version of draw always on SDL1 as well. Interesting, think I need to go and make some changes to my gui module.

MyreMylar:
it'd be useful to get some more testing on different platforms with the default threshold change using:

```
python3 -m pygame.examples.testsprite -layered_dirty 5000
python3 -m pygame.examples.testsprite -layered_dirty
```

those sets of settings (though I personally set it up via Pycharm's Run->Edit configuration to make it work). Basically comparing 2.0.0.dev10 or master versus the latest on this branch just in case it only runs faster on windows. 

To run the test you just have to let it run for a few seconds (I was counting to ten) then close the window and it should report the average FPS. For me the difference was very noticeable with current master, sdl2 config showing about 20ish FPS on the 5000 test and showing more like 40ish FPS with the latest on this branch just from the threshold change.

If you can't be bothered to sync to this branch you should get the same results by changing line 214 of `testsprite.py` to look like this:
```
        sprites = pg.sprite.LayeredDirty(_time_threshold=0)
```

illume:
Did you see there is a static option? LayeredDirty works best when there are static(non-moving) sprites. `python3 -m pygame.examples.testsprite -static`

I'm thinking it might be best to factor out the threshold algorithm into a function, so it can be tested easily? This would be easier than testing different apps on different platforms.

illume:
Note, that on windows with SDL < 2.0.13 partial updates are not implemented by SDL.

MyreMylar:
> Did you see there is a static option? LayeredDirty works best when there are static(non-moving) sprites. `python3 -m pygame.examples.testsprite -static`

Tried running the testing again with the static flag as well. It went a bit better for the default threshold (i.e. using DirtyFlag) but was still slightly worse than not using it. But eh, I think it is probably not the time to puzzle this out on windows with SDL 2.0.12. 

I'll just leave it at what it was before.

MyreMylar:
My conclusion is that I'll stop messing with LayeredDirty's draw function. I don't think I've made it worse in any significant way performance wise (though it's hard to tell on windows).

Is there anything else important to address?

illume:
Some future work came our of this PR, it would be good to capture it and make issues.

- LayeredDirty choosing dirty/non-dirty function doesn't always work. Probably the threshold stuff needs to be factored out and tested properly.
- get_ticks always returns 0, when not pygame.init. Find a way to get this to work, without init all pygame?
- assigning to Sprite.layer, see `@layer.setter` and comment https://github.com/pygame/pygame/pull/1925#discussion_r444273396

Issue #1755, opened 2020-05-16
Add unit test: display.toggle_fullscreen()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/display_test.py#L392-L405
**Docs:** https://www.pygame.org/docs/ref/display.html#pygame.display.toggle_fullscreen
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/display.c#L2676-L2688 & https://github.com/pygame/pygame/blob/master/src_c/display.c#L2221-L2507

tsadama:
Hi, I would like to work on this.

MyreMylar:
This test has been merged.

Issue #1743, opened 2020-05-16
Add unit test: display.get_active()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub**: https://github.com/pygame/pygame/blob/master/test/display_test.py#L68-L81
**Docs**: https://www.pygame.org/docs/ref/display.html#pygame.display.get_active
**Code to be tested**: https://github.com/pygame/pygame/blob/master/src_c/display.c#L260-L273

tsadama:
Hi, I would like to work on this.

illume:
Closing this, because the test was done by @tsadama 🎉

Issue #161, opened 2013-03-08
get_active doesn't work


illume:
Originally reported by: **Sidnei Pereira (Bitbucket: [sidnei](https://bitbucket.org/sidnei), GitHub: [sidnei](https://github.com/sidnei))**

----------------------------------------

This function doesn't work. I tested the state of display when the game is out of focus and when minimized and just keep returning 1

By the way, the docs says it should return a bool instead of a int.

OS: Ubuntu 12.04 x64
Python: 2.7.2
Pygame: 1.9.2pre

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/161

MyreMylar:
On windows right now this works in pygame 1.9.5 & 1.96, but it doesn't work in pygame 2.0-dev6.

**Code:**

```
import pygame


pygame.init()
window_surface = pygame.display.set_mode((640, 480))

is_running = True
while is_running:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            is_running = False

    if not pygame.display.get_active():
        print(str(pygame.display.get_active()))

    pygame.display.update()

```

**Reproduction steps:**

1. Run the code above.
2. Minimise the pygame window.
3. console should flood with '0' if it works.


I guess it could use re-testing on other platforms again, and perhaps a unit test somehow? Though it's not easy to force minimise a window cross platform with the standard libraries.

illume:
This works for me on MacOS, python3.7 pygame 2.0.0.dev7 SDL2.0.10

nthykier:
Works for me in pygame 1.9.6 and in pygame 2.0.0.dev7 on minimize but neither reacts to the window being "hidden" by another window nor losing focus.

I propose we fix this by updating the documentation to say "minimized" and then fix the return value to be Bool as we promised from the docs.

PR #1927, opened 2020-06-07
Added unit tests for display.toggle_fullscreen and display.get_active

tsadama:
Hello, this pull request addresses #1743 and #1755. 

The unit tests created for display.toggle_fullscreen() and display.get_active() were created with stipulations from the documentation and from my understanding of the code. I used pylint to check my code.

Two things I noticed when testing locally:

- The documentation for display.get_active() states that it will return false when an image is iconified. However, iconify returned true, but get_active returned true still. 

- When running display.set_mode() with width/height of (0,0) and then display.toggle_fullscreen(), the resulting width/height becomes (800, 600) with a display.toggle_fullscreen() return value that is not 1. However for all other width/heights I tried, the resulting width/height is the same as the original with a display.toggle_fullscreen() return value that is not 1.

 Please let me know if I should make any changes!


tsadama:
Follow up: it looks like one of the reasons for the unsuccessful build is because of the bullet point #1. Guidance on how I should address this would be appreciated!

MyreMylar:
> Hello, this pull request addresses #1743 and #1755.
> 
> The unit tests created for display.toggle_fullscreen() and display.get_active() were created with stipulations from the documentation and from my understanding of the code. I used pylint to check my code.
> 
> Two things I noticed when testing locally:
> 
> * The documentation for display.get_active() states that it will return false when an image is iconified. However, iconify returned true, but get_active returned true still.

Since this is happening locally and nor just on CI it sounds like a bug with get_active(). One of the issues we are trying to solve with getting unit tests for everything is finding actual bugs lurking in the code. I will have a poke at this tomorrow.

Sometimes the Continuous Integration builds behave differently to local builds.

> * When running display.set_mode() with width/height of (0,0) and then display.toggle_fullscreen(), the resulting width/height becomes (800, 600) with a display.toggle_fullscreen() return value that is not 1. However for all other width/heights I tried, the resulting width/height is the same as the original with a display.toggle_fullscreen() return value that is not 1.

Interesting, I guess it can't really make a full screen display with 0 pixels to play with so picking a reasonable default seems OK to me?

tsadama:
Thank you for the response. In the meantime, I will fix some of the other errors that popped up with the CI.

MyreMylar:
I had a quick look at `get_active()` and remembered this old bug:

https://github.com/pygame/pygame/issues/161

Which, reading between the lines, seems to indicate that `get_active()` maybe doesn't work only on windows?

 Are you on windows @tsadama ?

MyreMylar:
I'm using this MOR from the old bug:

```
import pygame


pygame.init()
window_surface = pygame.display.set_mode((640, 480))

is_running = True
while is_running:
        if event.type == pygame.QUIT:
            is_running = False

    if not pygame.display.get_active():
        print(str(pygame.display.get_active()))

    pygame.display.update()

```

Last time @illume reported that it worked on mac and @nthykier reported that it worked on his machine, which I'm guessing is probably linux of some flavour?

MyreMylar:
Tried that same code on Raspberry Pi OS (which is basically Debian) and it works fine. So it's looking like it is probably a windows issue.

MyreMylar:
Poked at the C code some, it looks like on windows `SDL_WINDOW_SHOWN` is always true whatever the minimised state of the window. However, I notice that there is also another flag called `SDL_WINDOW_MINIMIZED` which does work on windows.

Perhaps you could change `display.c` to read something like:

    #if IS_SDLv2
    static PyObject *
    pg_get_active(PyObject *self, PyObject *args)
    {
        Uint32 flags = SDL_GetWindowFlags(pg_GetDefaultWindow());
        return PyBool_FromLong((flags & SDL_WINDOW_SHOWN) &&
                               !(flags & SDL_WINDOW_MINIMIZED));
    }
    #else  /* IS_SDLv1 */
    static PyObject *
    pg_get_active(PyObject *self, PyObject *args)
    {
        return PyBool_FromLong((SDL_GetAppState() & SDL_APPACTIVE) != 0);
    }
    #endif /* IS_SDLv1 */

And see if it makes your tests work locally at least?

tsadama:
Hi @MyreMylar , I am actually using Ubuntu 16.04. I tried the MOR a few posts above, and my command line is being flooded with False so it seems like it's working. 

In terms of the change to the C code, I applied it and then did the following:
python3 setup.py build 
sudo python3 setup.py install 
ran the unit test on get_active

I still got the output where get_active would return true after a call to iconify, although I'm not entirely sure that I am rebuilding/recompiling it properly after changing the C code.

MyreMylar:
Not looked over what you've done yet, but I wonder if a couple of calls to pygame.event.get() or pygame.event.pump() after calling iconify() will sort out the issue for the unit tests. I suspect that it might be the sort of SDL instruction where it sends a request to the OS and then waits for an event to come back to know if it worked.

tsadama:
Follow up: 
I've tried modifying the C code as mentioned above and I've also tried updating the display, and using calls to pygame.event.get() and pygame.event.pump(). Unfortunately, I am unable to determine why iconify does not cause get_active() to be false. Could it be possible that iconify is buggy here? If so, it might be worth saving the iconify/get_active() check for an iconify unit test.

MyreMylar:
> Could it be possible that iconify is buggy here? If so, it might be worth saving the >iconify/get_active() check for an iconify unit test.

I guess an iconify() test case would be useful to check that. Modifying the code above:

```
import pygame


pygame.init()
window_surface = pygame.display.set_mode((640, 480))

is_running = True
while is_running:
        if event.type == pygame.QUIT:
            is_running = False
            pygame.display.iconify()

    if not pygame.display.get_active():
        print(str(pygame.display.get_active()))

    pygame.display.update()
```

For me, on windows, that makes pressing any key minimize the window just as if I'd pressed the minimize button on the menu bar. It doesn't print False when minimized on dev10, but it does if I add the C code change mentioned above.

If .iconify() doesn't minimize the window on linux then I guess we have our answer.

MyreMylar:
Just tried the above code using `pygame.display.iconify()` and it works fine on Raspberry Pi OS (which is Debian based) and using dev10. It prints False continuously after pressing a key to minimize the window.

tsadama:
I tried the above code and it looks like it’s working! To test the code without manually pressing a key, I used pygame.event to post a keydown event. I will make a commit once I factor this in to the unit test and make sure everything’s working locally

tsadama:
Update: the previous commits were my attempts to fix get_active. However, the CI builds are all terminating bc of a lack of output for 10 + minutes. I thought this might have been because of an infinite loop, but when I commented out the loop, the CI builds are still terminating.

robertpfeiffer:
I would expect his test to fail if the user is using a tiling window manager.

MyreMylar:
OK, I investigated this a bit. 

1. From what I can tell you can trim the failing part of the test right down to something like this:

```
pygame.init()

pygame.display.set_mode((640, 480))

pygame.event.clear()
pygame.display.iconify()

self.assertEqual(pygame.display.get_active(), False)
```

On windows and so long as you have also added the C code change above it will pass locally.

2. This part of the test at least will not work on the CI (Travis & Appveyor) as `iconify()` doesn't work with the 'dummy' video driver. This means you need to split the test into two parts and add a skip decorator to the top. that looks something like this:

```
    @unittest.skipIf(
        'requires the SDL_VIDEODRIVER to be a non "dummy" value',
    )
    def test_get_active_iconify(self):
```

3. I'm not sure what the issue is on Ubuntu but if the original test code was working for you and the test above doesn't then it sounds like it might be related to calling `display.update()` and `event.pump()` a few times so I would be tempted to just a a for loop that calls those functions a few times after calling iconify and seeing if that resolves the difference. I'll have a look on raspberry pi later myself.

tsadama:
Hi I made the changes requested above, but I am still getting the issue where my CI builds are terminating because of a lack of output for 10 minutes. The previous 3 commits were to make sure everything is updated with my fork/branch, but I don't know where the issue stems from.

MyreMylar:
From what I can tell Travis is failing on the SDL1 build on line 74. It looks to me it like there is an arguable bug in SDL 1 there. It appears that:

` return PyBool_FromLong((SDL_GetAppState() & SDL_APPACTIVE) != 0);`

is returning True before a window has even been created there.  

I think we could make the case for changing SDL 1's get_active to something like:

```
#else  /* IS_SDLv1 */
static PyObject *
pg_get_active(PyObject *self, PyObject *args)
{
    if (!pgDisplaySurfaceObject)
         return PyBool_FromLong(0);
    return PyBool_FromLong((SDL_GetAppState() & SDL_APPACTIVE) != 0);
}
#endif /* IS_SDLv1 */
```

to fit better with how the docs describe the function working only after set_mode has been called.

Both of these new tests passed on travis on SDL2 so if there is something else up with the Travis builds it is almost certainly not your fault. It looks like a whole bunch of new pull requests were merged so perhaps one of them has sewed chaos. 

Appveyor hasn't caught up with this PR yet either, but if it passes on there it is probably just Travis having a bad day which happens occasionally. I would try to fix the SDL1 problem either by changing the display.c code (better) or skipping that part of the test on SDL1 (OK). You shousl be able to test on SDL1 locally by changing the build config andthen doing a clean and rebuild, something like:

```
python -m buildconfig -sdl1
python setup.py clean --all
python setup.py install
pip install . -U 
```

Or however you do that process on your platform/build environment. You can switch back to sdl2 by doing the same thing but with sdl2 in place of sdl1 on the first line.

test/display_test.py
```diff
@@ -65,20 +65,53 @@ def todo_test_flip(self):
 
         self.fail()
 
-    def todo_test_get_active(self):
+    def test_get_active(self):
+        """Test the get_active function"""
+
+        #Initially, the display is not active
+        pygame.quit()
```
MyreMylar, line 72: don't need this pygame.quit() here. The display tests only init the display module before each test with the setUp() function:

https://github.com/pygame/pygame/blob/4940b9c664f8b063e8f634857c9f79ec5eb736d7/test/display_test.py#L18-L19

test/display_test.py
```diff
@@ -65,20 +65,53 @@ def todo_test_flip(self):
...
+        pygame.display.quit()
+        self.assertEqual(pygame.display.get_active(), False)
+
+        #get_active defaults to true after a set_mode
+        pygame.init()
```
MyreMylar, line 77: likewise should change this to `pygame.display.init()` since we are just testing the display module.

test/display_test.py
```diff
@@ -65,20 +65,53 @@ def todo_test_flip(self):
...
+
+
+    @unittest.skipIf(
+        os.environ.get("SDL_VIDEODRIVER") == "dummy",
+        'requires the SDL_VIDEODRIVER to be a non dummy value'
+    )
+    def test_get_active_iconify(self):
+        """Test the get_active function after an iconify"""
+
+        pygame.init()
```
MyreMylar, line 95: should delete this as `setUp()` will already have called `pygame.display.init()` for us.

test/display_test.py
```diff
@@ -416,20 +449,42 @@ def todo_test_set_palette(self):
 
         self.fail()
 
-    def todo_test_toggle_fullscreen(self):
+    def test_toggle_fullscreen(self):
```
MyreMylar, line 452: toggle_fullscreen() doesn't work with the dummy video driver. We should skip this whole test on CI.

According to the docs it also only works on X11 (linux) - this is clearly untrue looking at the current code which indicates support varies widely on a per platform basis. Probably we should also raise an issue to try and update the docs with more accurate info on what platforms toggle_fullscreen() now works on.

Two approaches here, you could either parse through the C code to try and work out which platforms currently support toggle_fullscreen(). This is the relevant part of the code:

https://github.com/pygame/pygame/blob/7d3900d5b04dc8102f218cc60f3fbf5e06dd8fa1/src_c/display.c#L2515-L2559

And then add checking for those platforms as part of the test (looks like it should work on windows, mac (cocoa) and linux (x11 & wayland drivers) at least but fail on android which is always full screen). Or, probably better, you could just make the test fail if the function raises an unexpected exception (as I mention below) and then build up a skip list of list of platforms that the test would fail on.

To figure out platforms for this you can probably mostly rely on the SDL_VIDEODRIVER environment variable again which also has 'android' in addition to the 'dummy' on used on CI. See this list in the get_driver test:

https://github.com/pygame/pygame/blob/48e94ef6106f3c1b79f95e0012f8f7ac2a964066/test/display_test.py#L106-L137

MyreMylar: TLDR: skip the test on dummy SDL_VIDEODRIVER and android at least.

MyreMylar:
I think probably the toggle fullscreen problems are why this is causing the CI to lock up.

MyreMylar:
I expect there may still be some edge cases where the `toggle_fullscreen()` test fails but given it only runs locally and most of us don't personally have the many potential configurations & platforms that pygame could run under I think it might be easier to see if we get any reports of failures for this test and then figure out how to skip the test in that particular situation.

MyreMylar:
Checked the current commit on windows under SDL1 & SDL2 and all the tests pass.

I'll submit the change to SDL2's version of get_active() as a separate PR, unless you want to roll it into here?

tsadama:
I don't mind if you submit the SDL2 as a separate PR! Thank you for all the help @MyreMylar !

MyreMylar:
Added issue about updating the `toggle_fullscreen()` docs here: https://github.com/pygame/pygame/issues/1990

MyreMylar:
Just holding off merging this until the windows fix can get merged at the same time.

PR #1932, opened 2020-06-08
Add a tox.ini to ease testing

mgorny:
Since the test suite requires the package being installed, it is
non-trivial to run from a fresh git checkout.  Add a config file for tox
to take care of creating a virtualenv, installing the package
and testing it.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,17 @@
+[tox]
+envlist = py{27,36,37,38,39}
+skipsdist = True
+skip_missing_interpreters = True
+
+[testenv]
+deps =
+    numpy
+setenv =
+    SDL_VIDEODRIVER=dummy
+    SDL_AUDIODRIVER=disk
+passenv =
+    PORTMIDI_INC_PORTTIME
+commands =
+    python -m buildconfig -auto
+    pip install .
+    python -m pygame.tests


illume:
Good idea.

Can you move it into the `buildconfig/` folder somehow? We moved various config into there to avoid cluttering up the root.

Maybe an instruction on how to use it as a comment at the top?

tox.ini
```diff
@@ -0,0 +1,17 @@
+[tox]
+envlist = py{27,36,37,38,39}
```
illume, line 2: We support python 3.5.

tox.ini
```diff
@@ -0,0 +1,17 @@
...
+[testenv]
+deps =
+    numpy
+setenv =
+    SDL_VIDEODRIVER=dummy
+    SDL_AUDIODRIVER=disk
+passenv =
+    PORTMIDI_INC_PORTTIME
+commands =
+    python -m buildconfig -auto
```
illume, line 15: I wonder if this `python -m buildconfig -auto` line is necessary?

tox.ini
mgorny: Ah, sorry, I've included only the versions I was able to test locally.

tox.ini
mgorny: It's needed if system is missing some dependencies. I presumed we want to let people also test how things work when some of the libs are missing.

mgorny:
> Can you move it into the `buildconfig/` folder somehow? We moved various config into there to avoid cluttering up the root.

That might be possible but would require non-trivial hacks. I can't find a proper way to define another working directory and using `cd` doesn't work (it expects real programs as `commands`). That said, moving it out of the root directory might mean people won't notice it.

> Maybe an instruction on how to use it as a comment at the top?

The usual way is to just call `tox`. Not sure if that's worth mentioning since it's the same way as any other tox project.

MyreMylar:
One alternative to consider over having lots of different .ini files is to use `setup.cfg`. Looks like there is still a lot of debate on a 'standard' single file for python tools - see this blog post for a useful table:

https://blog.pilosus.org/posts/2019/12/26/python-third-party-tools-configuration/

illume:
Cool.

We could perhaps remove the appveyor.yaml file, to make room for a setup.cfg. Since appveyor can be configured inside appveyor to use a different path (travis can't last I looked). I can't remember if there was a problem with that (forks not finding the appveyor config).  https://www.appveyor.com/docs/build-configuration/#yaml-file-alternative-naming I'll make another PR for this, since it would be good to do anyway.

We might need mypy and pylint config in the setup.cfg file too? (but do we really need any config for these tools?). Also, we could perhaps configure pytest in there to be able to run our test suite (not moving to pytest, but being able to use the runner would be useful sometimes).

Even if we only need config for pylint (we do), and not other tools it'd be worth it IMHO.

Let's use setup.cfg for the tox config?
It's documented how to do so here: https://tox.readthedocs.io/en/latest/config.html


---

ps. There's another post on my blog about using setup.cfg to avoid config file soup ;) It may be possible for us to engage in successful advocacy for black using setup.cfg... But maybe we don't need black config anyway, or maybe there's a friendlier alternative to black?  http://renesd.blogspot.com/2017/02/setupcfg-solution-to-python-config-file.html

pps.
### setup.cfg only (and remove setup.py)

Ok, this part is getting a bit off topic for this PR... but kind of related, so here we go :)

Removing the setup.py to be replaced by setup.cfg might also be a possibility. Probably it'd be a challenge.
- I think this is only supported in the last couple of years by setuptools. I never looked into it in detail, and can't remember the version setuptools started supporting this. Going setup.cfg only would mean requiring a recent setuptools.
- setuptools versions are **very** fragmented across different distros, but it might be worth looking at if we can now move to using setuptools via setup.cfg. 
- There's 20 years of `python setup.py install` docs out there
- Before we had the option of being setuptools free. I don't think this is necessary anymore, but it does actually make it a bit harder for people to install from source on Debian for example.

MyreMylar:
> We might need mypy and pylint config in the setup.cfg file too

I think the only thing that would definitely be nice for `pylint` in setup.cfg is the thing that makes it trust pygame c extensions (the default is not to trust any c extensions). This:

`extension-pkg-whitelist=pygame`

In general I prefer disabling things on a line-by-line, but I guess there might also be a case to be made for disabling any python 2 related compatibility warnings (e.g. inheriting from object) across the whole project up until we decide to remove python 2 support as it would be less work to change one line in `setup.cfg` than however many lines in all the `.py` files

If you have: 

```
[pylint.MESSAGES CONTROL]
extension-pkg-whitelist=pygame
```

in `setup.cfg` then you can just do:

    pylint src_py 

to check for errors in the whole directory rather than having to append the whole thing each time you call it.

illume:
@mgorny aha! thanks for letting us know. Good luck with your Gentoo python!
I'll continue this in another PR.

illume:
Since setuptools 30.3.0 (8 Dec 2016) declarative configuration in setup.cfg is available: https://github.com/pypa/setuptools/pull/862

Which is most modern linux distros: https://repology.org/project/python:setuptools/versions
(tldr; Centos 7, Debian old stable (stretch), Ubuntu 18.04).

illume:
Continued on here: https://github.com/pygame/pygame/pull/1944

PR #1936, opened 2020-06-09
Pylint fixes for py_draw.py

MyreMylar:
As far as I can tell this module is used exclusively for testing the real draw functions in `draw.c`. As such I was a little more ruthless in linting and tried to maintain only the function interfaces that were equivalent to the `draw.c` functions. 

The other problem with it is that there are lots of TODOs all over it that I didn't really want to get into. I feel like linting errors and fixing TODOs are two entirely different things that shouldn't really be jumbled together, as such I ran pylint with `fixme` disabled for this file.

**Disabled alerts**:

- Only one this time: 

 ```
def _multi_lines(surf, color, closed,  # pylint: disable=too-many-arguments
                  points, width=1, blend=False,
                  aaline=False):
 ```

This function is used to implement the interfaces of both `draw_line()` and `draw_aaline()` two functions that are already right at pylint's limit for function arguments, thus it wasn't really possible to cut the fat without splitting it into two functions - which would remove the point of having it all in one shared function in the first place.

**Remaining warnings after this PR (with fixme disabled)**

0

Tested using:

`pylint src_py/draw_py.py --extension-pkg-whitelist=pygame --disable=fixme`

### DIFF
--- a/src_py/draw_py.py
+++ b/src_py/draw_py.py
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
-'''Pygame Drawing algorithms written in Python. (Work in Progress)
+"""Pygame Drawing algorithms written in Python. (Work in Progress)
 
 Implement Pygame's Drawing Algorithms in a Python version for testing
 and debugging.
-'''
+"""
 from __future__ import absolute_import, division
+from collections import namedtuple
 import sys
 
 if sys.version_info >= (3, 0, 0):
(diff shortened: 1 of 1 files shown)


illume:
Yeah, this module was supposed to be a python implementation of the drawing algorithms.

- comparison testing with different implementations
- debugging, and implementing new algos is easier with python
- useful for teaching.
- perhaps very fast when JIT compiled on things like pypy

Issue #1788, opened 2020-05-18
Add unit test: time.Clock.tick()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)
[Draft guide on how to write a unit test for pygame](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/time_test.py#L74-L110
**Docs:** https://www.pygame.org/docs/ref/time.html#pygame.time.Clock.tick
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/time.c#L288-L352

AaronLi:
I'll give this a try

AaronLi:
Should this be closed?

MightyJosip:
This is now done thanks to the #1938 by @AaronLi. Thanks 👍

PR #1938, opened 2020-06-10
Added test for time.Clock.tick()

AaronLi:
This solves issue #1788.  I uncommented the old test that was there which tested `tick()`'s delay reporting ability and polished it up a bit, and also added a few new tests to verify that `tick()` will control a stable and proper framerate

### DIFF
--- a/test/time_test.py
+++ b/test/time_test.py
@@ -102,15 +102,15 @@ class ClockTypeTest(unittest.TestCase):
             collection.append(c.tick())
 
         # removes the first highest and lowest value
-        # for outlier in [min(collection), max(collection)]:
-        #     if outlier != milliseconds:
-        #         collection.remove(outlier)
+        for outlier in [min(collection), max(collection)]:
+            if outlier != milliseconds:
+                collection.remove(outlier)
 
         average_time = sum(collection) / len(collection)
 
         # assert the deviation from the intended framerate is within the
         # acceptable amount (the delay is not taking a dramatically long time)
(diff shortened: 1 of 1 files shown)


AaronLi:
It seems like there's 2 reasons why it's failing: Calling tick with a desired fps of n repeatedly n times takes less than 1 second for some reason, and there may be outliers disrupting the test for tick reporting accuracy. I have a fix for the latter

MyreMylar:
Good work so far!

Looks like your test is getting foxed by integer division on python 2 (on python 2 there is no implicit cast to float when you divide, so you'll need to cast to float before dividing to make it work on python 2 & 3).

It looks like this test also fails:

```

FAIL: test_tick (pygame.tests.time_test.ClockTypeTest)
Tests time.Clock.tick()
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/pygame-2.0.0.dev11-py3.5-linux-x86_64.egg/pygame/tests/time_test.py", line 135, in test_tick
    self.assertGreater(end - start, 1)
AssertionError: 0.9663300514221191 not greater than 1

```

Possibly you may be overestimating the accuracy of the frame rate limiting? I'd guess that it limits the framerate to pretty close to 100 on average but not exactly based on observation of it set to 60.

I suspect that this line in the docs:

> the program will never run at more than 40 frames per second.

may be over selling it a bit, especially given the next line is:

> Note that this function uses SDL_Delay function which is not accurate on every platform

PR #1939, opened 2020-06-10
Add unit test to Clock.get_rawtime

### LINKED ISSUE #1786: Add unit test: time.Clock.get_rawtime()

**Test stub:** https://github.com/pygame/pygame/blob/master/test/time_test.py#L37-L48
**Docs:** https://www.pygame.org/docs/ref/time.html#pygame.time.Clock.get_rawtime
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/time.c#L374-L379

jiquach:
Closes #1786 

Tests the logic:
- Clock initiation.
- `get_rawtime()`gets actual elapse time with a framerate argument in tick.
- `get_rawtime()` and `get_time()` return the same thing when there is no framerate argument.

### DIFF
--- a/test/time_test.py
+++ b/test/time_test.py
@@ -35,18 +35,33 @@ class ClockTypeTest(unittest.TestCase):
         # We should get around fps (+- fps*delta -- delta % of fps)
         self.assertAlmostEqual(clock.get_fps(), fps, delta=fps*delta)
 
-    def todo_test_get_rawtime(self):
+    def test_get_rawtime(self):
+        
+        iterations = 10
+        delay = 0.1
+        delay_miliseconds = delay*(10**3) #actual time difference between ticks
+        framerate_limit = 5
+        delta = 50 #allowable error in milliseconds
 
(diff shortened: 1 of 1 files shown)


jiquach:
It seems that one of the AppVeyor jobs failed on a different test (test_get_fps). I checked out a couple of other AppVeyor builds and it seems that the specific test passes for those.

- Is there a way to kick off another AppVeyor build to confirm whether it is a valid issue to look into?
- The `test_get_fps` uses a 20% delta from the true FPS vs the calculated FPS from `get_fps()` which is large already so increasing the delta might make the test less concrete.

illume:
Hi @jiquach thanks for mentioning the get_fps fail. I made a note about it at the issue here: https://github.com/pygame/pygame/issues/1785

The problem with the VMs and containers on highly loaded systems is that they can get very small time slices sometimes, and take quite a while to come back from sleeps. Making it pretty difficult to test timing related things.

Issue #1750, opened 2020-05-16
Add unit test: display.set_gamma()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/display_test.py#L295-L311
**Docs:** https://www.pygame.org/docs/ref/display.html#pygame.display.set_gamma
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/display.c#L1896-L1932

MightyJosip:
This is now done thanks to the #1942 by @gtsopus. Thanks 👍

PR #1942, opened 2020-06-10
Add display.set_gamma unit test

gtsopus:
Adds a unit test function for display.set_gamma for   Issue #1750

I do that by first creating a list containing different gamma values ranging from 0.0 to 1.0 and I use them for the red=blue=green gamma values by only inputting the red value. 
Similarly, I re-create the list, filling it with tuples in (r,g,b) style and input them as different r,g,b values in the function.
In order for the test to pass, I expect the function to return True for every case.

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -315,7 +315,7 @@ class DisplayModuleTest(unittest.TestCase):
         self.assertFalse(display.get_init())
 
     @unittest.skipIf(
-    os.environ.get("SDL_VIDEODRIVER") == "dummy",
+    os.environ.get("SDL_VIDEODRIVER") == "dummy" and not SDL2,
     "Can't set gamma on SDL 1 with the dummy video driver",
     )
     def test_set_gamma(self):
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -315,14 +315,14 @@ class DisplayModuleTest(unittest.TestCase):
         self.assertFalse(display.get_init())
 
     def test_set_gamma(self):
-        pygame.display.set_mode((1, 1))
+        if(not SDL2):
+            pygame.display.set_mode((1, 1))
         gammas = [0.0,0.25,0.5,0.88,1.0]
         for gamma in gammas:
             self.assertEqual(pygame.display.set_gamma(gamma),True)
         gammas = [(0.5,0.5,0.5),(1.0,1.0,1.0),(0.22,0.33,0.44),(0.0,0.0,0.0)]
         for gammaTuple in gammas:
             self.assertEqual(pygame.display.set_gamma(gammaTuple[0],gammaTuple[1],gammaTuple[2]),True)
-
 
     def todo_test_set_gamma_ramp(self):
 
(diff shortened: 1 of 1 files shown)


MyreMylar:
The build that is failing specifically is the SDL 1 one, which uses this SDL function:

http://sdl.beuc.net/sdl.wiki/SDL_SetGamma

You can build pygame with SDL 1 using:

```
python setup.py clean --all
python -m buildconfig -sdl1
python setup.py install
pip install . -U
```

Or something similar, depending on your platform requirements. The thing to test is whether it is always returning False under SDL1, returning False only with certain values, or returning False only on the CI. 

It could be that this function is hardware dependent only on SDL 1.

Once we know what the circumstances of failure are it's easier to know how to proceed, it could be a bug in the code we can fix, or just a test we need to skip on the CI.

gtsopus:
I tried everything as advised,, with the existing code and different inputs but it doesn't seem to fail when using SDL1, at least locally on my PC. Maybe just like you said this function is hardware dependent or just something that I didn't notice yet.

MyreMylar:
It could be the 'dummy' video driver that is used on CI. You can try it lcoally by setting the SDL_VIDEODRIVER enviornment variable to 'dummy' near the top of your program (I thin definitely before pygame.init() or display.init() are called:

`os.environ['SDL_VIDEODRIVER'] = 'dummy'`

(that's from memory so I may have it slightly wrong)

gtsopus:
Looks like a dummy video card really fails the test, thank you for your help. I'll try and skip it if it detects a dummy video card and see how it does.

MyreMylar:
The appveyor test failures are not your fault, that is just ongoing issues we are having with testing the `time` module figuring out error deltas that work on appveyor.

I do think we could probably do the skipping a little more carefully though as it looked like the test was running fine on the SDL 2 build. Hang on I'll do one of those request changes things.

test/display_test.py
```diff
@@ -314,23 +314,19 @@ def test_quit__multiple(self):
...
-        # if the function succeeds it will return True.
-        #
-        # A gamma value of 1.0 creates a linear color table. Lower values will
-        # darken the display and higher values will brighten.
-        #
-
-        self.fail()
+    @unittest.skipIf(
+    os.environ.get("SDL_VIDEODRIVER") == "dummy",
+    'OpenGL requires a non-"dummy" SDL_VIDEODRIVER',
```
MyreMylar, line 319:
```suggestion
    @unittest.skipIf(
    os.environ.get("SDL_VIDEODRIVER") == "dummy" and not SDL2,
    'Can't set gamma on SDL 1 with the dummy video driver',
```

Think this is a skip that will still let us run the test on most of the Continuous Integration builds.

MyreMylar:
Just closing and re-opening to kick off appveyor again.

MyreMylar:
I also added this:

`and not SDL2`

here:

`os.environ.get("SDL_VIDEODRIVER") == "dummy" and not SDL2`

in my suggested change so that the SDL 2 tests would still run on CI - because I think they worked before and more testing is better than none

MyreMylar:
Looks good to me and all the tests pass :)

👍 🎉 

Thank you for contributing!

PR #1944, opened 2020-06-11
setup.cfg for tooling config, starting with tox

illume:
This is the PR from @mgorny continued https://github.com/pygame/pygame/pull/1932

Moves tox config into the setup.cfg

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
-[tox]
-envlist = py{27,36,37,38,39}
+[tox:tox]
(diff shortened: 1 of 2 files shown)


MyreMylar:
I think we may end up putting `fixme` into the pylint warning disable list as well, just because fixing TODO comments is generally a much larger job than fixing most (though not all) regular pylint warnings.

Either that or we could purge TODOs from the codebase and turn them into github issues instead?

MyreMylar:
Anyway, this seems like a great start.

PR #1945, opened 2020-06-12
Increase error deltas for some tests that failed on appveyor

MyreMylar:
These two tests failed:

```
FAIL: test_set_timer (pygame.tests.time_test.TimeModuleTest)
Tests time.set_timer()
  File "C:\Python36-x64\lib\site-packages\pygame\tests\time_test.py", line 248, in test_set_timer
    self.assertEqual(events, test_number)
AssertionError: 7 != 8
Ran 9 tests in 14.959s


FAIL: test_tick (pygame.tests.time_test.ClockTypeTest)
Tests time.Clock.tick()
  File "C:\Python35-x64\lib\site-packages\pygame\tests\time_test.py", line 130, in test_tick
    self.assertAlmostEqual(average_time, milliseconds,delta = epsilon)
AssertionError: 6.112244897959184 != 5.0 within 0.75 delta
Ran 9 tests in 14.409

```

I've noticed that it seems to be always Appveyor that fails on these error delta tests while Travis does fine. Perhaps what we need for a more general solution is some kind of generic error delta multiplier for the time module on appveyor.

### DIFF
--- a/test/time_test.py
+++ b/test/time_test.py
@@ -95,17 +95,17 @@ class ClockTypeTest(unittest.TestCase):
         """Tests time.Clock.tick()"""
         """
         Loops with a set delay a few times then checks what tick reports to
-        verify its accuracy. Then calls tick with a desired framerate and
-        verifies it is not faster than the desired framerate nor is it taking
+        verify its accuracy. Then calls tick with a desired frame-rate and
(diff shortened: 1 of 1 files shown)


illume:
Cool. Let's see if this fixes it :)
Another option would be to disable the tests on appveyor.

PR #1946, opened 2020-06-13
Pylint pass for midi module

MyreMylar:
**Tested using:** `pylint src_py/midi.py`

This PR reduces the pylint warnings down to this list:

```
src_py\midi.py:18:2: W0511: TODO: finish writing tests. (fixme)
src_py\midi.py:21:2: W0511: TODO: create a background thread version for input threads. (fixme)
src_py\midi.py:103:2: W0511: TODO: find all Input and Output classes and close them first? (fixme)
src_py\midi.py:377:45: W0613: Unused argument 'buffer_size' (unused-argument)
```
Three TODOs, which are beyond the scope of a pylint pass, and one genuine error in the public interface. It appears that the `buffer_size` parameter passed to the midi.Output class is never used at all.

I'll raise a separate issue about the `buffer_size` parameter.

**Disabled**:

`def quit():  # pylint: disable=redefined-builtin`

This is a sort of generic problem across all pygame modules -  that their quit functions shadow python's built-in `quit()` function. For example, if you do:

```
from pygame.midi import *

quit()
```

You will call the `pygame.midi` modules `quit()` function rather than the built-in `quit()` function. I don't think we can change this at this point, all those module `quit()` functions are out there now.

### DIFF
--- a/src_py/midi.py
+++ b/src_py/midi.py
@@ -15,11 +15,10 @@ point in the future.  The pyportmidi bindings are included with pygame.
 New in pygame 1.9.0.
 """
 
-
-#TODO:
-#    - finish writing tests.
-#        - likely as interactive tests... so you'd need to plug in a midi device.
-#    - create a background thread version for input threads.
+# TODO: finish writing tests.
+#        - likely as interactive tests... so you'd need to plug in
+#          a midi device.
+# TODO: create a background thread version for input threads.
 #        - that can automatically inject input into the event queue
 #          once the input object is running.  Like joysticks.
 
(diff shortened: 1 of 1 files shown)


MyreMylar:
I did some weird code to avoid having to use the global keyword for module level state here:

```
def _module_init(state=None):
    # this is a sneaky dodge to store module level state in a non-public
    # function. Helps us dodge using globals.
    if state is not None:
        _module_init.value = state
        return state

    try:
        _module_init.value
    except AttributeError:
        return False
    else:
        return _module_init.value
```

I'm not particularly happy about it, though it does work. 

The standard advice for getting rid of mutable globals is to transform the code using them into a class, but that isn't really an option here because the midi module is already in widespread use. Thus we come back to this. Does anyone have a cleaner way to purge globals like this?

illume (approved):
Nice cleanups. Thanks 👍

Issue #1846, opened 2020-05-24
Add minimum window size support


nthykier:
I would like to see `SDL_SetWindowMinimumSize` exposed somehow, so I can set a minimum size for the window and avoid having to fight with the SDL resizing (with resizable windows).

In Pygame 1/SDL 1 the "wisdom" was to call `set_mode` after a resize (not that it worked reliably). When #1705 is merged, I think the recommendation will be to stop with the `set_mode` calls but many games cannot sanely render their graphics under a certain window size.  This is where I hope that `SDL_SetWindowMinimumSize` will be useful.

**Related docs**: https://www.pygame.org/docs/ref/display.html

illume:
I'd prefer we don't add new SDL2 functionality before pygame 2.1. One reason especially for window related functionality is that there is a Window class, and these are probably better exposed through that. Sorry I didn't make this clearer earlier.

Closing this issue for now so someone else doesn't waste more effort. Sorry @Starbuck5

PR #1950, opened 2020-06-15
Minimum and maximum window size

Starbuck5:
Makes an interface for SDL_SetWindowMinimumSize, SDL_GetWindowMinimumSize, SDL_SetWindowMaximumSize, and SDL_GetWindowMaximumSize.

In response to issue #1846 

I added everything to the docs as well.

Things to note:
- I didn't know whether these functions were SDL2 only or not, so I didn't put them in one of the "if SDL1" or "if SDL2" blocks
- Setting a maximum size is significantly less useful, so maybe it shouldn't be exposed to try to keep pygame simpler
- I don't actually know C, hopefully I was able to pick up things in the other functions well enough
- I couldn't find a way to get rid of a minimum or maximum size constraint once it was established. You can change it to something else, but you can't go back to being unconstrained.

Other idea:
If setting a maximum or minimum size is a one time operation (can't be undone), maybe it should be in pygame.display.set_mode.

Like:
`pygame.display.set_mode((480,480), pygame.RESIZABLE, minsize = (100,50), maxsize=(1920,1080))`

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -593,7 +593,7 @@ required).
 
 .. function:: get_allow_screensaver
 
-   | :sl:`Return whether the screensaver is allowed to run.`
+   | :sl:`Return whether the screensaver is allowed to run`
    | :sg:`get_allow_screensaver() -> bool`
 
    Return whether screensaver is allowed to run whilst the app is running.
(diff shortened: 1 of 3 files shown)


MyreMylar:
Looks like you need to change:

`Py_BuildValue("(ii)", min_w, min_h);`

to:

`return Py_BuildValue("(ii)", min_w, min_h);`

And similar in the other get function.

illume:
I'd prefer we don't add new SDL2 functionality before pygame 2.1. One reason especially for window related functionality is that there is a Window class, and these are probably better exposed through that.

Starbuck5:
Seems legit, I'll close this then.

illume:
Sorry I didn't make that clearer earlier.

Issue #1762, opened 2020-05-16
Add unit test: mouse.get_cursor()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/mouse_test.py#L16-L18
**Docs:** https://www.pygame.org/docs/ref/mouse.html#pygame.mouse.get_cursor
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/mouse.c#L281-L323

Reminisque:
Hello. I would like to contribute. May I get assigned this issue?

MyreMylar:
Absolutely! 👍

Reminisque:
I forked and clone the repo, but I could not get the testing to run as the `pygame` module could not be found when importing. Am I supposed to install `pygame` through `pip` before working on it or am I supposed to follow the instructions here https://www.pygame.org/wiki/CompileWindows ?

Also are there any other dependencies I need to download or other instructions I need to follow?

MyreMylar:
Ah windows.

In my experience it is easier to get the sdl1 build working now. So from the pygame directory try:

```
python setup.py clean --all
python -m buildconfig -sdl1
python setup.py install
pip install . -U
```

That's what I tend to do. I think this unit test should have the same behaviour on SDL 1 and SDL 2 so you can probably rely on the CI.

Reminisque:
I tried the instructions, but I'm getting the following error:
```
distutils.errors.DistutilsPlatformError: Microsoft Visual C++ 14.2 is required. Get it with "Microsoft Visual C++ Build Tools": https://visualstudio.microsoft.com/downloads/
```
I downloaded and installed the build tools required and updated the `setuptools` package. However it's still giving me the error when I try to run the first command again.

MyreMylar:
Ah yes, you may also need to setup environment variables or run the developer command prompt to unstick it:

For example, here are my environment variable for the many versions of Python I have installed:

![image](https://user-images.githubusercontent.com/13382426/84119625-897a5480-aa2c-11ea-9df0-409720181b53.png)

the MSVC compiler confusingly has lots of different version names. It looks like you need this one to build C extensions with your version of Python:

`MSVC++ 14.2  _MSC_VER == 1920 (Visual Studio 2019 Version 16.0)`

Possibly you can set the command line environment variables like this:

[...]

PR #1955, opened 2020-06-16
Added unit test for mouse.get_cursor()

Reminisque:
Hello this is my first contribution to pygame. 

This pull request is a unit test for `mouse.get_cursor` and addresses issue #1762. The test handles display initialization, SDL version, NULL mouse cursor and expected output based on the code documentation and what I understood the method does.

In regards to the display, I saw that there was a class at the top of `mouse_test.py` that was handling the initialization and quitting of `pygame.display`. There was supposed to be a `finally` block at the end where `pygame.display.quit()` would be executed, but it ended up failing the rest of the tests since `test_get_cursor` ran first and the other tests don't have `pygame.display.init()`.

If there's anything I need to change please let me know. Thank you.

### DIFF
--- a/test/mouse_test.py
+++ b/test/mouse_test.py
@@ -57,9 +57,43 @@ class MouseModuleInteractiveTest(MouseTests):
 
 class MouseModuleTest(MouseTests):
 
-    def todo_test_get_cursor(self):
+    def test_get_cursor(self):
         """Ensures get_cursor works correctly."""
-        self.fail()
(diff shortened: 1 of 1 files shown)


MyreMylar:
Looks good to me! 👍 🎉 

Thank you very much for the contribution and good work getting it compiling on windows.

Reminisque:
Thank you for all the help with compiling on windows.

Issue #1947, opened 2020-06-14
Use of deprecated APIs.


methane:
Py_UNICODE and relating APIs are deprecated since Python 3.3.
We are planning to remove them in Python 3.11.  Please remove followings:

```
./pygame-1.9.6/src_c/math.c:1599:    attr = PyUnicode_AsUnicode(attr_unicode);
./pygame-1.9.6/src_c/math.c:1679:    attr = PyUnicode_AsUnicode(attr_unicode);
./pygame-1.9.6/src_c/pgcompat.h:50:#define Text_GetSize PyUnicode_GetSize
```

* PyUnicode_AsUnicode may be replaced with PyUnicode_AsUTF8
* Text_GetSize can be removed.

robertpfeiffer:
To anybody willing to take this on: Remember that Python 2.7 legacy support will not be dropped until after the stable pygame 2.0 release.

MyreMylar:
Another PR reminded me to update this issue. We still need to remove these functions from `ft_unicode.c`

PR #1956, opened 2020-06-16
Avoid PEP393 deprecated APIs.

methane:
There is remaining use of deprecated APIs in ft_unicode.c.
But it is not simple, straightforward.  So I exclude it from this pull request.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -3825,7 +3825,7 @@ _view_kind(PyObject *obj, void *view_kind_vptr)
 #else
         if (PyUnicode_GET_LENGTH(obj) != 1) {
             PyErr_SetString(PyExc_TypeError,
-                            "expected a length 1 string for argument 3");
+                            "expected a length 1 string for argument 1");
             return 0;
         }
         ch = PyUnicode_READ_CHAR(obj, 0);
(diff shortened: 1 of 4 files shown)


src_c/surface.c
```diff
@@ -3815,12 +3815,21 @@ _view_kind(PyObject *obj, void *view_kind_vptr)
...
         }
         ch = *PyUnicode_AS_UNICODE(obj);
+#else
+        if (PyUnicode_GET_LENGTH(obj) != 1) {
+            PyErr_SetString(PyExc_TypeError,
+                            "expected a length 1 string for argument 3");
+            return 0;
+        }
+        ch = PyUnicode_READ_CHAR(obj, 0);
+#endif
```
MyreMylar, line 3832: Looks like a small cut and paste issue here as the error string for python 2 says 'argument 1', but the error string for python 3 says 'argument 3'

MyreMylar:
This is related to this issue:
https://github.com/pygame/pygame/issues/1947

nthykier:
Looks good to me.

@MyreMylar: Could you check if you are ok with it after the recent changes? :)  (I cannot merge it while you have a "pending" Changes requested note).

Issue #1766, opened 2020-05-16
Add unit test: threads.WorkerQueue.threadloop()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/threads_test.py#L56-L62
**Docs:** None.
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_py/threads/__init__.py#L187-L202

PR #1961, opened 2020-06-18
Adds test for threadloop

kduggan15:
From what I understand, threadloop isn't really meant to be used by the user. It's just the loop used by each thread in the pool. 

The idea behind this test is that if you make a single threaded worker queue, and make the worker do threadloop, it's behaviour shouldn't change since it's just busy running threadloop anyway.

stop() works as expected, but wait() doesn't work, since threadloop is infinite

### DIFF
--- a/test/threads_test.py
+++ b/test/threads_test.py
@@ -53,13 +53,33 @@ class WorkerQueueTypeTest(unittest.TestCase):
 
         self.assertIs(wq.queue.get(), STOP)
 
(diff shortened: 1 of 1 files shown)


kduggan15:
#1766

illume (approved):
Looks good to me. Thanks :)

I see it's still in Draft mode, do you still want to do something else to it? If so, I think it can be done in further a PR.

Issue #1866, opened 2020-05-27
freetype.Font.render_to() returns an offset rect rather than a position rect


MyreMylar:
The docs for this function imply that the rect returned will contain the position of the text in the surface being rendered to. In fact the function returns the same data as the function freetype.Font.get_rect() with the same shared inputs (render_to() also takes some input parameters get_rect() does not as you might expect).

I originally suggested amending the docs but it was pointed out that hardly anyone has used the return value of this function so far, so we might as well make it do what it says it does instead.

Current docs: https://www.pygame.org/docs/ref/freetype.html#pygame.freetype.Font.render_to

Code that will have to be changed: 
https://github.com/pygame/pygame/blob/master/src_c/freetype/ft_render.c#L262-L406

In particular:
https://github.com/pygame/pygame/blob/master/src_c/freetype/ft_render.c#L396-L397

MyreMylar:
Looks like this was noticed before in this issue: 
https://github.com/pygame/pygame/issues/1182

mmurrs:
Can I be assigned this issue?

mmurrs:
Is there a person who would be good to contact about some syntax that is used within the function? I cannot find any documentation for "FX6_TRUNC" within pygame nor online.

illume:
@mmurrs It's defined in src_c/freetype/ft_wrap.h
```c
#define FX6_TRUNC(x)  ((x) >> 6)
```

IgorSondors:
@MyreMylar @mmurrs @mmurrs This problem is still exists in pygame==2.0.1 installed via pip. Also I tried to build pygame==2.0.2.dev1 from source and use solution from [this](https://github.com/pygame/pygame/pull/2410) pull request that not solved the problem too. I really need to use freetype.render_to, please help

zoldalma999:
Can you share an example script? I tried this on windows and it worked:
```py
import pygame
import pygame.freetype
pygame.init()

surf = pygame.Surface((200, 200))
font = pygame.freetype.Font("freesansbold.ttf", 20)

print(font.render_to(surf, (23, 37), "Test"))
```

Output:
```
pygame 2.0.1 (SDL 2.0.14, Python 3.7.7)
Hello from the pygame community. https://www.pygame.org/contribute.html
<rect(23, 37, 41, 15)>
```
The font file ("freesansbold.ttf") copied from [here](https://github.com/pygame/pygame/blob/main/src_py/freesansbold.ttf)

[...]

PR #1971, opened 2020-06-23
Addresses Issue #1866 for render_to() to return a position rect

mmurrs:
The FX6_TRUNC was resulting in incorrect values caused by bit-shifting

### DIFF
--- a/test/freetype_test.py
+++ b/test/freetype_test.py
@@ -669,7 +669,8 @@ class FreeTypeFontTest(unittest.TestCase):
         rrect = font.render_to(surf, (32, 32), "FoobarBaz", color, None, size=24)
         self.assertIsInstance(rrect, pygame.Rect)
         self.assertEqual(rrect.top, rrect.height)
-        ## self.assertEqual(rrect.left, something or other)
+        ##test that render_to is position
+        self.assertEqual(rrect.left, something or other)
 
         rcopy = rrect.copy()
         rcopy.topleft = (32, 32)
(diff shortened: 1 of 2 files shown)


MyreMylar:
Amazing!

The main thing we need here to finish this off is a unit test of the functionality here:

https://github.com/pygame/pygame/blob/48571a6b060b5c528ba584684e9120cec96afa99/test/freetype_test.py#L661-L749

Currently there are some commented out lines in there that refer to the returned rectangle but obviously (since the test passes) there are no actual checks of the supposed functionality. Ideally we could make an understandable change that would make this test fail on the current master but then work with your new fix.

mmurrs:
Ok thanks for the feedback, I will go through those unit tests to make sure render_to() is working properly.

test/freetype_test.py
```diff
@@ -669,7 +669,9 @@ def test_freetype_Font_render_to(self):
         rrect = font.render_to(surf, (32, 32), "FoobarBaz", color, None, size=24)
         self.assertIsInstance(rrect, pygame.Rect)
         self.assertEqual(rrect.top, rrect.height)
-        ## self.assertEqual(rrect.left, something or other)
+        ##test that render_to is position
+        self.assertEqual(rrect.left, 32)
+        self.assertEqual(rrect.right, 32)
```
MyreMylar, line 674: This seems wrong, the returned rect (according to the docs) is supposed to have the size as well as the position of the rendered text - so the right hand side of the rectangle should definitely be further to the right right than the left hand side with the text 'FoobarBaz' at font size 24.

**Example picture:**

image

It won't come out to exactly 156px on the right hand side of the rect as that's m,e just eyeballing it in a paint program, but it should definitely have at least _some_ width.

test/freetype_test.py
```diff
@@ -680,7 +682,8 @@ def test_freetype_Font_render_to(self):
         self.assertEqual(rrect.top, rrect.height)
         self.assertNotEqual(rrect.size, rect.size)
         rrect = font.render_to(surf, (20.1, 18.9), "FoobarBax", color, None, size=24)
-        ## self.assertEqual(tuple(rend[1].topleft), (20, 18))
+        # Added the test back
+        self.assertEqual(tuple(rend[1].topleft), (20, 18))
```
MyreMylar, line 686: Could also check the bottomright here once the width & height part is working.

MyreMylar:
Looks like no progress on this one. Gonna just close it up, no hard feelings, just useful to keep a clean-ish list of open PRs. You can always re-open it later if you have new inspiration.

Issue #1754, opened 2020-05-16
Add unit test: display.set_palette()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/display_test.py#L376-L390
**Docs:** https://www.pygame.org/docs/ref/display.html#pygame.display.set_palette
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/display.c#L1806-L1894 & https://github.com/pygame/pygame/blob/master/src_c/display.c#L1934-L1999

hanzohasashi33:
Hello, may I be assigned this issue?

MightyJosip:
@hanzohasashi33 Sure

MightyJosip:
This is now done thanks to the #1972 by @hanzohasashi33. Thanks 👍

PR #1972, opened 2020-06-23
Update math docs and add test for display.set_palette()

### LINKED ISSUE #1957: scale_to_length raises ValueError, docs say ZeroDivisionError
https://www.pygame.org/docs/ref/math.html#pygame.math.Vector2.scale_to_length says:
>  If the vector is the zero vector (i.e. has length 0 thus no direction) an ZeroDivisionError is raised.

However, the code says ValueError instead:
https://github.com/pygame/pygame/blob/4562396964cca81bceb2d12d3f09aa96746cb3fa/src_c/math.c#L1294-L1296

One of them (the docs, in my opinion) should change to match the other.

**Steps to resolve:**

- [x] Change the docs for pygame.math.Vector2 to report that a ValueError is returned when attempting to scale a zero length vector. Source docs [here](https://github.com/pygame/pygame/blob/4562396964cca81bceb2d12d3f09aa96746cb3fa/docs/reST/ref/math.rst). Online Docs [here](https://www.pygame.org/docs/ref/math.html#pygame.math.Vector2.scale_to_length)
- [x] Change the docs for pygame.math.Vector3 to report that a ValueError is returned when attempting to scale a zero length vector. Source docs [here](https://github.com/pygame/pygame/blob/4562396964cca81bceb2d12d3f09aa96746cb3fa/docs/reST/ref/math.rst). Online docs [here](https://www.pygame.org/docs/ref/math.html#pygame.math.Vector3.scale_to_length).

hanzohasashi33:
Changed the docs for pygame math.Vector.2 and math.vector.3 to report that a ValueError is raised instead of ZeroDivisionError when attempting to scale a zero length vector.
fixes #1957

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -148,8 +148,8 @@ Multiple coordinates can be set using slices or swizzling::
 
       Scales the vector so that it has the given length. The direction of the
       vector is not changed. You can also scale to length 0. If the vector is
-      the zero vector (i.e. has length 0 thus no direction) an
-      ZeroDivisionError is raised.
+      the zero vector (i.e. has length 0 thus no direction) a
+      ValueError is raised.
 
       .. ## Vector2.scale_to_length ##
 
@@ -421,8 +421,8 @@ Multiple coordinates can be set using slices or swizzling::
 
       Scales the vector so that it has the given length. The direction of the
       vector is not changed. You can also scale to length 0. If the vector is
-      the zero vector (i.e. has length 0 thus no direction) an
-      ZeroDivisionError is raised.
+      the zero vector (i.e. has length 0 thus no direction) a
+      ValueError is raised.
 
       .. ## Vector3.scale_to_length ##
 
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -401,6 +401,9 @@ class DisplayModuleTest(unittest.TestCase):
 
     @unittest.skipIf(SDL2, "set_palette() not supported in SDL2") 
     def test_set_palette(self):
+        with self.assertRaises(UnboundLocalError) :
+            palette = [1,2,3]
+            screen.set_palette(palette)
         screen = pygame.display.set_mode((1024,768),pygame.DOUBLEBUF,8)
         palette = []
         self.assertIsNone(screen.set_palette(palette))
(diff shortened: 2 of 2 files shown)


MightyJosip:
Most recent change also address: #1754. You should fix the error you are getting right now
```
ERROR: test_set_palette (pygame.tests.display_test.DisplayModuleTest)
  File "C:\Python36-x64\lib\site-packages\pygame\tests\display_test.py", line 406, in test_set_palette
    self.assertIsNone(pygame.display.set_palette(palette))
pygame.error: Display mode is not colormapped
```
and also probably add some more palettes (right now you only check for the empty list)

robertpfeiffer:
bit depth is ignored by set_mode on SDL2, and modern hardware doesn't do 8-bit color-mapped display modes more. This test should just be marked to be skipped on SDL2.

hanzohasashi33:
@robertpfeiffer , So shall I just close the pr and mark it to be skipped on SDL2?

robertpfeiffer:
> @robertpfeiffer , So shall I just close the pr and mark it to be skipped on SDL2?

Don't close the PR, but mark to skip the palette test on SDL2.

hanzohasashi33:
> 
> 
> > @robertpfeiffer , So shall I just close the pr and mark it to be skipped on SDL2?
> 
> Don't close the PR, but mark to skip the palette test on SDL2.

@robertpfeiffer May I know how to do the above ?

MyreMylar:
Here is an example of a test that uses a decorator to skip the test when running on SDL2:

https://github.com/pygame/pygame/blob/48571a6b060b5c528ba584684e9120cec96afa99/test/display_test.py#L152-L166

You can change the string in the decorator to describe why this test is being skipped e.g. "set_palette() not supported in SDL2"

hanzohasashi33:
> 
> 
> Here is an example of a test that uses a decorator to skip the test when running on SDL2:
> 
> https://github.com/pygame/pygame/blob/48571a6b060b5c528ba584684e9120cec96afa99/test/display_test.py#L152-L166
> 
> You can change the string in the decorator to describe why this test is being skipped e.g. "set_palette() not supported in SDL2"

Thanks for the help!

MyreMylar:
Test it on SDL1 on windows and the test passes locally too, looks decent on a look over.

Issue #1789, opened 2020-05-18
Add unit test: time.Clock.tick_busy_loop()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)
[Draft guide on how to write a unit test for pygame](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/time_test.py#L112-L134
**Docs:** https://www.pygame.org/docs/ref/time.html#pygame.time.Clock.tick_busy_loop
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/time.c#L354-L358

MightyJosip:
This is now done thanks to the #1976 by @cruxicheiros. Thanks 👍

PR #1976, opened 2020-06-23
Add basic tests for tick_busy_loop

cruxicheiros:
I decided to have a go at Issue #1789.

I wrote some basic tests that just check whether the return value of the function is what's expected.

I think that based on docs, the main test for `tick_busy_loop` should be the same as for `tick` but with a higher standard for accuracy - the thing is that I don't know _how_ much more accurate so I don't personally feel confident in finishing the test.

Thanks!

### DIFF
--- a/test/time_test.py
+++ b/test/time_test.py
@@ -163,26 +163,33 @@ class ClockTypeTest(unittest.TestCase):
         # Test whether the return value of tick_busy_loop is the length of the
         # tick in milliseconds
         second_length = 1000
+        shortfall_tolerance = 1  # (ms) The amount of time a tick is allowed to run short of, to account for underlying rounding errors
         sample_fps = 40
 
-        self.assertLessEqual(c.tick_busy_loop(sample_fps), second_length/sample_fps)
+        self.assertGreaterEqual(c.tick_busy_loop(sample_fps), (second_length/sample_fps) - shortfall_tolerance)
+        pygame.time.wait(10)  # incur delay between ticks that's faster than sample_fps
+        self.assertGreaterEqual(c.tick_busy_loop(sample_fps), (second_length/sample_fps) - shortfall_tolerance)
+        pygame.time.wait(200)  # incur delay between ticks that's slower than sample_fps
+        self.assertGreaterEqual(c.tick_busy_loop(sample_fps), (second_length/sample_fps) - shortfall_tolerance)
 
         high_fps = 500
-        self.assertLessEqual(c.tick_busy_loop(high_fps), second_length/high_fps)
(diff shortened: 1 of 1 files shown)


cruxicheiros:
Why does this fail on one Travis build but not the others?

cruxicheiros:
Fixed the issue: the issues with Build 7 were being caused by that build running slower than the given framerate specifies (which is explicitly allowed in the spec). The problem was solved by switching equality checks to less-than-or-equal checks, which more closely fits the specification and allows the tests to pass. :raised_hands:

cruxicheiros:
Currently doing revision on this PR :construction:

cruxicheiros:
Made a new PR to account for the fact that:

- Ticks (as in the time spent between the last call to `clock.tick_busy_loop` and the next one) can take longer than the time set by the FPS
- Ticks sometimes take less than the time set by the FPS, in current observation only by 1 millisecond, suggesting they are caused by an occasional rounding error
- There is (currently undocumented) behaviour that changes the duration of each call to  `clock.tick_busy_loop` based on the time elapsed between the end of the last call and the start of the second to keep FPS consistent.

MyreMylar:
> There is (currently undocumented) behaviour that changes the duration of each call to clock.tick_busy_loop based on the time elapsed between the end of the last call and the start of the second to keep FPS consistent. I would like this to become documented behaviour.

Is this undocumented? I guess the wording is perhaps a little vague, but the purpose of this function and `tick()` with an FPS argument is to try and keep a consistent frame rate in a game loop so the delay would have to be variable based on how much work was done on any given game loop.

These are the current relevant lines in the docs (with a similar one for tick():

> If you pass the optional framerate argument the function will delay to keep the game running slower than the given ticks per second. This can be used to help limit the runtime speed of a game. By calling Clock.tick_busy_loop(40) once per frame, the program will never run at more than 40 frames per second.

I guess you could add an extra line on the end that makes it clear that the length of the delay with a frame rate argument passed will vary based on how much time has passed since the previous call to this function, in order to try and keep the frame rate consistent with what is requested. Do it in a separate pull request though to keep things clean - docs for the time module are here:

https://github.com/pygame/pygame/blob/master/docs/reST/ref/time.rst

cruxicheiros:
@MyreMylar Gave it a go. There's no point writing tests for the string one because [the interface limits the type](https://github.com/pygame/pygame/blob/master/buildconfig/pygame-stubs/time.pyi) and if it isn't castable to int then you'll just get a TypeError. But here are some others.

cruxicheiros:
Re requested review since the merge appeared to be blocked by the unresolved change request.

PR #1977, opened 2020-06-24
Added unit test for the surface.lock() function

41aaronb:
I added in a unit test for the surface.lock() function, I think it is good to go but review would be appreciated.

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1386,7 +1386,35 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
         # only be unlocked after the final lock is released.
         #
 
-        self.fail()
+        # Basic
+        surf = pygame.Surface((100, 100))
+        surf.lock()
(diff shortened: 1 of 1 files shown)


robertpfeiffer:
Based on the current behaviour of SDL2, this test case does nothing except increasing metrics of coverage.

robertpfeiffer:
On further reflection, we can go ahead and merge this, I'll open another issue for surface locking.

41aaronb:
Hey guys, wondering how we can actually get this merged, let me know if there's anything more I need to do!

MyreMylar:
> Hey guys, wondering how we can actually get this merged, let me know if there's anything more I need to do!

Nothing more for you to do, we just generally leave issues open for a little while so everyone with an interest can have a chance to look at it and comment.

It's approved now so it'll probably get merged in a batch with a bunch of other issues in a few days.

Thanks again for the contribution!

PR #1980, opened 2020-06-24
unit test for Surface.get_locks function

jtoloff:
I believe I added all the tests needed to ensure the expected functionality, please let me know if there is anything I can improve upon.


robertpfeiffer:
This needs to check for pixelarray locks as well.

jtoloff:
How does the new commit look? Does the added test check pixelarray locks sufficiently?

robertpfeiffer:
I just realised that there are two PRs for surface locking. I guess this is good enough for get_locks.

Issue #1765, opened 2020-05-16
Add unit test: threads.WorkerQueue.do()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/threads_test.py#L28-L35
**Docs:** None
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_py/threads/__init__.py#L173-L176

codescientist703:
Hi, I would like to contribute to this issue.

PR #1983, opened 2020-06-25
Adds test for thread do

kduggan15:
#1765
Adds test for threads.WorkerQueue.do()


MyreMylar (commented):
This test seems to be basically the same as the one bundled up in: https://github.com/pygame/pygame/pull/2064

I think I'll end up preferring that PR due to the overall similarity and the fact it has another test bundled up in it - if you think there is anything to add from this test to that one that I've missed then please let me know.

Apologies for the length of time in getting to review this, the summer holidays are always chaotic round here.

PR #1986, opened 2020-06-26
updated draw_test.py with test_ellipse

Sowmyasree101:


MightyJosip:
This code doesn't test anything

Sowmyasree101:
> _No description provided._



> This code doesn't test anything

Is this code not accepted?

MyreMylar:
@Sowmyasree101 you need to have at least one self.assert type line in a test for it to do anything. 

The name of this test also clashes with one of the 10+ existing ellipse tests in this file, see:

https://github.com/pygame/pygame/blob/264c74a288fa36d8a90c9e2115bd8610cd6b3fd4/test/draw_test.py#L498-L537

If you thing something is going untested about the ellipse drawing, then it would be good to specify what is being tested in the name of the test, similar to the other double underscore `def test_ellipse__invalid_color_formats(self)` style names.

PR #1989, opened 2020-06-27
Allow setting several paths in PYGAME_EXTRA_BASE

Gosha:
The variable is split on "`:`" and is stored in an array called `extrabases`, which implies that several paths are valid.

However, the loop just overwrites `incdirs` and `libdirs`, so only the last PATH in that variable is used.


MyreMylar:
This looks like a straight logic error to me - not much point iterating through a loop if you are just overwriting everything each cycle. @illume is likely the best person to confirm.

Issue #1759, opened 2020-05-16
Add unit test: touch.get_finger()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/touch_test.py#L37-L39
**Docs:** https://www.pygame.org/docs/ref/touch.html#pygame._sdl2.touch.get_finger
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/_sdl2/touch.c#L86-L123

joshi-b:
Hi, I would like to contribute to this!

joshi-b:
PR [here](https://github.com/pygame/pygame/pull/1999)

PR #1999, opened 2020-07-03
Added test for touch.get_finger()

joshi-b:
This should fix issue #1759


MyreMylar:
Gonna close this for now, feel free to re-open in new inspiration strikes. This needs some sort of interactive test like we have now started for the joystick module.

Issue #1997, opened 2020-07-03
In setup.py add_command decorator is broken


jtiai:
When trying to invoke setup.py:
```
$ python setup.py --help
Traceback (most recent call last):
  File "setup.py", line 616, in <module>
    class WinBuildExt(build_ext):
  File "setup.py", line 510, in decorator
    assert issubclass(command, Command)
AssertionError
```

In setup.py there is the code:
```python
def add_command(name):
    def decorator(command):
        assert issubclass(command, Command)
        cmdclass[name]=command
        return command
    return decorator
```

Later in the setup.py decorator is used:
```python
    @add_command('build_ext')
    class WinBuildExt(build_ext):
```

`build_ext` class inherits from `distutils.cmd.Command` not `distutils.core.Command` which is tested in code causing setup.py to fail with assertion error when `WinBuildExt` class definition is found,

`command` = `WinBuildExt` and let's check it's MRO list.
```python
(Pdb++) command
<class '__main__.WinBuildExt'>
(Pdb++) command.__mro__
(<class '__main__.WinBuildExt'>, <class 'distutils.build_ext'>, <class 'distutils.command.build_ext.build_ext'>, <class 'distutils.cmd.Command'>, <class 'object'>)
```

Let's compare MRO list of two different `Command` classes.
```python
(Pdb++) distutils.cmd.Command.__mro__
(<class 'distutils.cmd.Command'>, <class 'object'>)
(Pdb++) distutils.core.Command.__mro__
(<class 'setuptools.Command'>, <class 'distutils.cmd.Command'>, <class 'object'>)
```

Let's make check with `issubclass`:
```python
(Pdb++) issubclass(command, distutils.cmd.Command)
True
(Pdb++) issubclass(command, distutils.core.Command)
False
```

Using windows 10, Python 3.8.1 32bit.
setuptools 47.3.1

jtiai:
After more digging this happens only in virtualenv environment created by pipenv. Doesn't happen with standard virtualenv tools.

Wonder if pipenv picks some old distuitls/setuptools...

MyreMylar:
Ahh, that explains it. Well, no reason not to fix it as far as I can see since it works outside of pipenv with the change as well.

robertpfeiffer:
That's what I get for using clever features like decorators!

PR #2001, opened 2020-07-04
Pipenv: Fixed incorrect base class comparison (#1997)

jtiai:
Fixes issue #1997 where comparison is made against wrong `Command` class.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -507,7 +507,7 @@ cmdclass = {}
 
 def add_command(name):
     def decorator(command):
-        assert issubclass(command, Command)
+        assert issubclass(command, distutils.cmd.Command)
         cmdclass[name]=command
         return command
     return decorator


MyreMylar:
Tried this locally on windows and didn't notice it breaking anything, everything still builds as it did before.

Under what circumstances was this causing an error - the issue doesn't make it clear?

jtiai:
Sorry forgot to add original error:
```
When trying to invoke setup.py:
```
$ python setup.py --help
  File "setup.py", line 616, in <module>
    class WinBuildExt(build_ext):
  File "setup.py", line 510, in decorator
    assert issubclass(command, Command)
AssertionError
```

Same error appeared when trying to install pygame from sources.

MyreMylar:
Hmm, I don't get this error on windows as far as I can tell. Perhaps it depends on a specific python version?

MyreMylar:
Added Pipenv to the title so we can find this again if we get any more reports/related issues.

Issue #1749, opened 2020-05-16
Add unit test: display.iconify()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/display_test.py#L205-L221
**Docs:** https://www.pygame.org/docs/ref/display.html#pygame.display.iconify
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/display.c#L2153-L2162

## remaining work

- [ ] Add issue to add WINDOWEVENT_MINIMIZED as a constant to pygame in the same way we have WINDOWEVENT_CLOSE.
- [ ]  Add issue to update the documentation for pygame 2 to indicate the event now generated by iconify()/minimizing in general under SDL2.
- [ ] double check this display.iconify test works locally on windows & raspberry pi.

zoldalma999:
Hey!
I would like to try and do this unit test

PR #2006, opened 2020-07-06
Adding unit test to iconify

zoldalma999:
#1749

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -256,17 +256,22 @@ class DisplayModuleTest(unittest.TestCase):
         self.fail()
 
     def test_iconify(self):
-        pygame.display.set_mode((640, 480))
+        screen = pygame.display.set_mode((640, 480))
         Iconified_event = False
         success = pygame.display.iconify()
 
+        #'ActiveEvent' (gain 0 state 2) didnt seem to work but is the same as 'WindowEvent' 7
         for e in pygame.event.get():
-            if e.type == pygame.ACTIVEEVENT:
-                if e.gain == 0 and e.state == 2 and success:
-                    Iconified_event = True
+            # if e.type == pygame.ACTIVEEVENT:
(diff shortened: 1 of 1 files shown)


zoldalma999:
Im sorry it took me so long. I made the changes.
Is there a way to test sdl 1 versions? I tried looking for it but couldnt find one.
Also should i open an issue about the event/docs?

MyreMylar:
> Im sorry it took me so long. I made the changes.

No worries at all, we probably wouldn't notice unless you vanished for longer than a month :)

> Is there a way to test sdl 1 versions? I tried looking for it but couldn't find one.

Yes. There is an SDL 1 build that runs on Travis which should now be labelled as such, but if I recall this test only runs locally so you would need to build pygame with the SDL 1 build configuration. For me the process of switching SDL versions goes something like this:

```
python setup.py clean --all
python -m buildconfig -sdl1
python setup.py install
pip install . -U
```

But the exact commands you use may depend on how you build python modules/pygame normally and your platform (if you've not built pygame from source yet you may also want to have a look at the [wiki documentation for different platforms](https://www.pygame.org/wiki/Compilation) e.g. here are the windows instructions: https://www.pygame.org/wiki/CompileWindows - which I know from experience can be tricky )

> Also should i open an issue about the event/docs?

You can if you like, or I'll do it myself at some point when this PR gets wrapped up.

MyreMylar:
Remaining work for me:

- [ ] Add issue to add WINDOWEVENT_MINIMIZED as a constant to pygame in the same way we have WINDOWEVENT_CLOSE.
- [ ]  Add issue to update the documentation for pygame 2 to indicate the event now generated by iconify()/minimizing in general under SDL2.
- [ ] double check this test works locally on windows & raspberry pi.

illume (approved):
Thanks. 🎉  

I think we can merge this in now. I moved the TODO list to the issue.

Issue #1796, opened 2020-05-18
Add unit test: surface.convert_alpha()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)
[Draft guide on how to write a unit test for pygame](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/surface_test.py#L854-L872
**Docs:** https://www.pygame.org/docs/ref/surface.html#pygame.Surface.convert_alpha
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/surface.c#L1963-L2012

williams-c:
I haven't contributed to an open source project before but read over the guides and everything, mind if I take a shot at this one?

MyreMylar:
Good luck!

CobyCoding:
@williams-c Does the pull above fix the issue. If so can we close this issue.

PR #2008, opened 2020-07-07
test for surface.convert_alpha()

williams-c:
Completed test for #1796, let me know if this work or you have any feedback, thanks!


williams-c:
Added the other tests. Wasn't exactly sure how the test the bit depths that are changed by convert alpha, they all seemed to by changed to 32 which is what I looked for in the test cases.

MyreMylar:
Looks good to me, and all the test pass 👍 

Thanks for the contribution!

PR #2009, opened 2020-07-07
Added unit test: JoystickModuleTest.test_get_count

AdditionalPylons:
First PR! :)

### DIFF
--- a/test/joystick_test.py
+++ b/test/joystick_test.py
@@ -53,6 +53,8 @@ class JoystickModuleTest(unittest.TestCase):
         with self.assertRaises(pygame.error):
             pygame.joystick.Joystick(count)
 
+        pygame.joystick.quit()
+
 
     def test_get_init(self):
         # Check that get_init() matches what is actually happening
--- a/test/joystick_test.py
+++ b/test/joystick_test.py
@@ -3,10 +3,6 @@ from pygame.tests.test_utils import question, prompt
 
 import pygame
 
-pygame.joystick.init()
-# The number of joysticks available for testing.
-JOYSTICK_COUNT = pygame.joystick.get_count()
-pygame.joystick.quit()
 
 class JoystickTypeTest(unittest.TestCase):
     def todo_test_Joystick(self):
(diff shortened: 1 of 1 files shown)


MyreMylar:
As I said in the issue, this sort of 'returns 0' test is fine for the CI where there will always be zero joysticks connected - but to truly check the functionality we need an interactive sort of test where you are prompted to confirm that you have plugged in a controller or two first, then after the prompt the test would call get_count() and then ask the interactive tester to confirm that the number of controllers detected is equal to the number the tester things is connected.

If you run the cdrom tests module under SDL1 it does this sort of testing, prompting you to pop a CD drive in and out.

AdditionalPylons:
@MyreMylar Thanks for the feedback! I just ran cdrom_test.py and see what you're talking about...I'm a total beginner to open source and hadn't seen an interactive test like this before--pretty awesome! As it is, I checked that the test I wrote works for 0, 1, and 2 controllers, but doesn't give any visual feedback...just pass or fail. Would you like me to rewrite it as an interactive test confirming the number of controllers and resubmit?

test/joystick_test.py
```diff
@@ -31,21 +37,38 @@ def todo_test_Joystick(self):
 
 
 class JoystickModuleTest(unittest.TestCase):
-    def todo_test_get_count(self):
-
-        # __doc__ (as of 2008-08-02) for pygame.joystick.get_count:
+    @unittest.skipIf(0 == JOYSTICK_COUNT, "No joysticks detected")
```
MyreMylar, line 40: I think it is best to skip this test in a different way.

There is a 'tag' you can apply to a test case class that means all the tests in the class are skipped unless you run the tests with the interactive flag. This is the correct approach to take here as well.

Check out this example in the font_test module:

https://github.com/pygame/pygame/blob/599a76c0e8edf6059001b22fd9e7160d8e050ce8/test/font_test.py#L476-L478

I think we need to basically add a new class to the joystick_test module called `JoystickInteractiveTest` that has the 'interactive' tag as above. That way you can include a basic functionality unit test like your first draft that runs on the CI platforms (Travis, Appveyor) in the `JoystickModuleTest` class, and then a second test in the JoystickInteractiveTest class that does the whole prompt, question scenario in this version of the test. Make sense? Two tests, two different classes. I suspect when other people write more joystick tests they will also fit into this pattern. 

This will also have the advantage of not skipping the interactive test if you start it with no controllers connected.

test/joystick_test.py
```diff
@@ -31,21 +37,38 @@ def todo_test_Joystick(self):
...
+    def test_get_count(self):
+        # Test get_count correctly identifies number of connected joysticks
+        prompt("Please add or remove joysticks as necessary before the test.")
```
MyreMylar, line 43: For this line it would probably be clearer for a modern audience if we went with something like:

> "Please connect any joysticks/controllers now before starting the joystick.get_count() test."

That tells people which test is happening and also gives some indication they can connect console controllers and not just an old Thrustmaster 7000 from the loft or something.

test/joystick_test.py
```diff
@@ -31,21 +37,38 @@ def todo_test_Joystick(self):
...
+
+        self.assertGreaterEqual(JOYSTICK_COUNT, 0, ("joystick.get_count() must "
+                                            "return a value >= 0"))
+
+        response = question(
+            ("NOTE: Having Steam open may add an extra virtual joystick for "
+             "each joystick physically plugged in.\n"
+             "Is the correct number of joysticks connected to this system [{}]?"
+             .format(JOYSTICK_COUNT))
+        )
```
MyreMylar, line 58: Again for this message it would probably be useful to indicate that by 'joystick' pygame also refers to console controllers. Perhaps what the confirmation message is after could be a little clearer too. Something like:

> joystick.get_count() thinks there is {} joystick(s) or controller(s) connected to this system. Is this correct?

AdditionalPylons:
Knew skipping the whole test if count == 0 was not ideal but didn't know there was a better way and was afraid prompt would break CI. Moved the auto test to the bottom of the class to keep it close to the new interactive test and to make sure funcs called in my code have already been tested above.

Really thorough and helpful feedback, definitely appreciate it! Love pygame and trying to get started in open source, really hoping to help out and learn as much as I can in the process. :)

Here's hoping this one does the trick!

AdditionalPylons:
Awesome, thanks so much for taking the time to look over it and give feedback! Since this is my first PR, is there anything else I need to do to close out the issue? Looking forward to the merge and to starting in on the next one :)

MyreMylar:
Usually I leave up PR's for about a week-ish after I approve review them in case any other reviewers want to weigh in. Then there is a merging day when I, or someone else scoops up a bunch of PRs and merges them all.

Feel free to create another branch after checking out your master branch again if you want to carry on.

PR #2014, opened 2020-07-12
added unit test for image.load_extended()

bubo-py:
it's my very first contribution as well as unit test, so please tell me if something is wrong and how should i fix that

### DIFF
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -732,13 +732,9 @@ class ImageModuleTest(unittest.TestCase):
 
         self.fail()
 
-    def todo_test_load_extended(self):
+    def test_load_extended(self, image):
 
-        # __doc__ (as of 2008-08-02) for pygame.image.load_extended:
-
-        # pygame module for image transfer
-
-        self.fail()
+        self.assertRaises(pygame.error, image.load_extended, [])
 
     def todo_test_save_extended(self):
 


MyreMylar:
Thank you for having a go at writing a test, but I suspect this function needs a little more testing of its purpose which is to load image formats other than .bmp. I would expect a test to have a go at loading at least some file types from this list in the docs page:

- JPG
- PNG
- GIF (non-animated)
- BMP
- PCX
- TGA (uncompressed)
- TIF
- LBM (and PBM)
- PBM (and PGM, PPM)
- XPM

I believe some of these file types are already included in the pygame package in the example/demo folder. See `example_path` in this test file. A few of the file types are already loaded in other tests but not all of them and this test is a decent place to just try out the list in the docs in one place. There are other lots of tests of file loading in general in this test file and in `imageext_test.py` and different several sets of bad data permutations - so this particular unit test should probably just focus on the core purpose of the `load_extended()` function (can it load all the file types in the docs list?) 

For example, the test you almost succeeded in writing is actually already covered in another test here:

https://github.com/pygame/pygame/blob/master/test/imageext_test.py#L21-L22

If you need to add any new data files that are not already in:

https://github.com/pygame/pygame/tree/master/examples/data

Then they should be small files (e.g. 32x32 pixels) and added to that directory.

If that sounds too difficult then feel free to let us know and move on and I'll add this extra information to the original issue.

bubo-py:
Hey, I am writing to you with a request of some help... Even though I tried
I think I don't really understand how to write a unit test. There are a
couple of examples that I thoroughly understood, but when it comes to real
world examples like that in pygame, my brain starts to struggle. Could you
please show me some tests pretty similar to that one I was assigned to do
and explain it to me? I would be glad to help with improving pygame, but
first of all I need to learn how to do it, lol. Thanks in advance!!!

MyreMylar:
> Hey, I am writing to you with a request of some help... Even though I tried I think I don't really understand how to write a unit test. There are a couple of examples that I thoroughly understood, but when it comes to real world examples like that in pygame, my brain starts to struggle. Could you please show me some tests pretty similar to that one I was assigned to do and explain it to me? I would be glad to help with improving pygame, but first of all I need to learn how to do it, lol. Thanks in advance!!!

Sure no problem, there is a guide written by @illume on how to write unit tests [here](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html) and there are lots of example tests in the [tests folder](https://github.com/pygame/pygame/tree/master/test) of the GitHub repo. For example here is one for `save_extended()`: 

https://github.com/pygame/pygame/blob/6dbff9193cbee07ec6c05091a866ea71f868566d/test/image_test.py#L809-L841

Don't worry if you don't figure out writing tests the first time you try, it's all good practice! 

The general approach when writing a function unit test is to check that everything the function is supposed to do works (the documentation is a good starting place, but reading the code helps too) and everything it's supposed to fail at fails correctly (here you generally need to look at the code for assertions/errors). You can also try to push some limits with parameter inputs, unexpected but still accepted ones for example and see if the behaviour is what you would expect.

I'll close this PR for now as we are gearing up for a release, but feel free to come back at a later date if you have a fresh take.

Issue #1805, opened 2020-05-18
Add unit test: surface.get_losses()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)
[Draft guide on how to write a unit test for pygame](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/surface_test.py#L1028-L1040
**Docs:** https://www.pygame.org/docs/ref/surface.html#pygame.Surface.get_losses
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/surface.c#L2747-L2894

Reminisque:
May I work on this issue?

MyreMylar:
Absolutely.

Reminisque:
I've looked through the documentation for the function and the associated SDL objects such as `SDL_Surface` and `SDL_PixelFormat`, but I'm not entirely clear how everything works. 

From what I could understand of the docs on masks, shifts, and losses the values in mask are for getting the bits for a specific color component (R or G or B or A) from a pixel's color value with length equal to the depth passed to the `pygame.Surface` object it belongs to, the values in shifts are the number of bits to the right of the corresponding component, and losses is the number of left shifts needed to make the component 8-bit after using the mask to isolate the bits and shifts to bring those bits all the way to the right. Also each of these are already set when creating a new instance of `pygame.Surface` depending on the depth set. 

I noticed that after playing around with the `set_masks` and `set_shifts` the losses do not change. What happens if someone changes the mask and shift values for one of the color components so that it ends up having a length larger than 8 bits after being isolated, shifted to the right, and shifted to the left? I also found that for an 8-bit depth, the masks show that the number of bits assigned is in a 2-3-2-0 configuration which only adds up to 7 bits and only 2 bits are used for red. I only just started learning more about color depth since starting this issue, so I am not sure if this is intentional since I read on wikipedia that its normally 3-3-2 for RGB.

Are there any other functions for `pygame.Surface` that I would have to consider beisdes those metnioned above? I'm thinking I might need some cases regarding alpha, but would it only be through me setting the alpha masks when creating the surface and not through `set_alpha`? It didn't change masks, shifts, or losses after setting it with the set function.

[...]

PR #2020, opened 2020-07-21
Add unit test for surface.get_losses

Reminisque:
This pull request addresses issue #1805.

I tested standard color depths from 8-bit to 32-bit. For 8-bit, the default mask was 2 bits for red, 3 bits for green, 2 bits for blue, and no bits for alpha. I only looked up color depths after starting this issue so please feel free to correct me if I am wrong, but is it supposed to be missing a bit for red? On wikipedia it had said the usual configuration is 3 bits for red and green and 2 bits for blue. As for 32-bit the default mask does not include bits for alpha, so I added another 32-bit depth surface with an included alpha mask.

Are there any test cases that I may be missing?


Reminisque:
I am getting this error in CI:
```
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/pygame-2.0.0.dev11-py2.7-linux-x86_64.egg/pygame/tests/surface_test.py", line 1272, in test_get_losses
    self.assertEqual(surf_8bit.get_losses(), (6, 5, 6, 8))
AssertionError: Tuples differ: (8, 8, 8, 8) != (6, 5, 6, 8)
First differing element 0:
8
6
- (8, 8, 8, 8)
?  ^  ^  ^
+ (6, 5, 6, 8)
?  ^  ^  ^
```
It was my mistake to assume that the losses will be the same for for a given color depth on all systems. However, I am unsure how to find out what losses I'll be given except when passing my own masks when creating a new Surface.

MyreMylar:
> I tested standard color depths from 8-bit to 32-bit. For 8-bit, the default mask was 2 bits for red, 3 bits for green, 2 bits for blue, and no bits for alpha.

7 bits smacks of leaving space for 1 bit alpha to me (opaque or transparent) but since you said no alpha I may be wrong. Perhaps pygame uses 7 bits for colour to make it easy to transition from 8 bit without alpha to 8-bit with alpha with no colour alteration.

> It was my mistake to assume that the losses will be the same for for a given color depth on all systems. However, I am unsure how to find out what losses I'll be given except when passing my own masks when creating a new Surface. Should I pass a mask for all surfaces, except the one with no depth parameter as that one seems to work fine?

656 is another one of those compressed colour formats a bit like the old DXT that used to be popular. 

If the losses are different on CI compared to local I suspect you can figure out what the losses are likely to be from the colour format of the default display surface (returned from `set_mode()`)? I recall that mapped colours have to do with the display surface and that this is different between CI and most local configurations.

Perhaps also try setting the display driver to 'dummy' locally and see if you can reproduce the behaviour. I'll have a poke at that myself.

Reminisque:
```
surf_set_mode = pygame.display.set_mode((100, 100))
surf_default = pygame.Surface((100, 100))
surf_8bit = pygame.Surface((100, 100), depth=8)
surf_8bit_332 = pygame.Surface((100, 100), depth=8, masks=rgba8)
surf_15bit = pygame.Surface((100, 100), depth=15)
surf_16bit = pygame.Surface((100, 100), depth=16)
surf_24bit = pygame.Surface((100, 100), depth=24)
surf_32bit_rgb = pygame.Surface((100, 100), depth=32)
surf_32bit_rgba = pygame.Surface((100, 100), depth=32, masks=rgba32)

print(surf.get_losses())
print(surf_default.get_losses())
print(surf_8bit.get_losses())
print(surf_8bit_332.get_losses())
print(surf_15bit.get_losses())
            print(surf_16bit.get_losses())
            print(surf_24bit.get_losses())
            print(surf_32bit_rgb.get_losses())
            print(surf_32bit_rgba.get_losses())
```

Reminisque:
Is there a way to print out values in CI? I tried to see what losses I am getting in the CI, but am unable to get any output from the print statements I added to the test. The losses seem to be different from what I get locally with the video driver set to dummy, but I can only tell when a test fails so I would only get test output per commit.

Also it seems that alphas aren't valid for 8-bit and 15-bit. I tried to put 1 bit for alpha by removing 1 bit from blue to keep the depth the same, but it only worked locally. I haven't tried 32-bit yet on the CI, since it's one of the last tests and the failures come before it.

Reminisque:
I got the losses to work. Only the 8-bit depth surface with no masks given is returning losses of 8 bits for all color components. Every other surface is returning losses that match what I get locally without setting the driver to dummy.

test/surface_test.py
```diff
@@ -1240,19 +1239,55 @@ def todo_test_get_locks(self):
...
+            elif (display_surf.get_losses() == (8, 8, 8, 8)):
+                self.assertEqual(losses, (8, 8, 8, 8))
+
+            self.assertEqual(surf_8bit.get_losses(), (5, 5, 6, 8))
+            self.assertEqual(surf_15bit.get_losses(), (3, 3, 3, 8))
+            self.assertEqual(surf_16bit.get_losses(), (3, 2, 3, 8))
+            self.assertEqual(surf_24bit.get_losses(), (0, 0, 0, 8))
+            self.assertEqual(surf_32bit.get_losses(), (0, 0, 0, 0))
+            
+            # Method should fail when display is not intialized
```
MyreMylar, line 1284: A small typo here in the word initialized/initialised.

MyreMylar (changes requested):
Overall it looks good, test passes locally for me and it all makes sense I'll approve it with the small typo fixed.

Reminisque:
I've added fixes for the typo and the extra parentheses. Looks like a bit of javascript syntax got mixed in while I was working on it.

Is it okay for the appveyor check to fail? I looked at what the error message, but there wasn't any indication that it was an error being raised during testing. I think it was a 403 HTTP error code.

MyreMylar (approved):

Issue #1760, opened 2020-05-16
Add unit test: key.get_focused()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/key_test.py#L27-L39
**Docs:** https://www.pygame.org/docs/ref/key.html#pygame.key.get_focused
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/key.c#L768-L778

leopoldwe:
Can I be assigned this issue?

MyreMylar:
Assigned.

leopoldwe:
@MyreMylar  I submitted pull request #2026 with my code changes. Can you take a look?

MyreMylar:
This test is merged now.

PR #2026, opened 2020-07-31
Add key.get focused unit test

leopoldwe:
This is a pull request to fix issue #1760.

### DIFF
--- a/test/key_test.py
+++ b/test/key_test.py
@@ -24,19 +24,10 @@ class KeyModuleTest(unittest.TestCase):
         "does it import"
         import pygame.key
 
-    def todo_test_get_focused(self):
-
-        # __doc__ (as of 2008-08-02) for pygame.key.get_focused:
-
-        # pygame.key.get_focused(): return bool
-        # true if the display is receiving keyboard input from the system
-        #
-        # This is true when the display window has keyboard focus from the
-        # system. If the display needs to ensure it does not lose keyboard
-        # focus, it can use pygame.event.set_grab() to grab all input.
-        #
-
-        self.fail()
+    def test_get_focused(self):
+        """Ensure get_focused returns the correct type"""
+        focused = pygame.key.get_focused()
+        self.assertIsInstance(focused, int)
 
     def test_get_pressed(self):
         states = pygame.key.get_pressed()
--- a/test/key_test.py
+++ b/test/key_test.py
@@ -25,9 +25,12 @@ class KeyModuleTest(unittest.TestCase):
         import pygame.key
 
     def test_get_focused(self):
-        """Ensure get_focused returns the correct type"""
+        self.assertFalse(pygame.key.get_focused()) #No window to focus
+        pygame.display.set_mode((500, 500))
+        pygame.event.get() #Pump event queue to get window focus
         focused = pygame.key.get_focused()
         self.assertIsInstance(focused, int)
+        self.assertTrue(focused)
 
     def test_get_pressed(self):
         states = pygame.key.get_pressed()
--- a/test/key_test.py
+++ b/test/key_test.py
@@ -1,3 +1,4 @@
+import os
 import unittest
 import pygame
 import pygame.key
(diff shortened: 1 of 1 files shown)


test/key_test.py
```diff
@@ -24,19 +24,24 @@ def test_import(self):
...
-        # This is true when the display window has keyboard focus from the
-        # system. If the display needs to ensure it does not lose keyboard
-        # focus, it can use pygame.event.set_grab() to grab all input.
-        #
-
-        self.fail()
+    def test_get_focused(self):
+        focused = pygame.key.get_focused()
+        if not SDL1:
+            self.assertFalse(focused) #No window to focus
```
MyreMylar, line 30: Does this mean that under SDL1 `key.get_focused()` returns True, even before you call `display.set_mode()` to create a window? That seems a bit screwy and maybe something we could fix in the C code.

test/key_test.py
```diff
@@ -24,19 +24,24 @@ def test_import(self):
...
+        self.assertIsInstance(focused, int)
+        display_sizes = pygame.display.list_modes()
+        if display_sizes == -1:
+            display_sizes = [(500, 500)]
+        if not display_sizes:
+            return
```
MyreMylar, line 36: Does returning here cause this test to fail? Genuinely not sure.

test/key_test.py
```diff
@@ -24,19 +24,24 @@ def test_import(self):
...
+        pygame.display.set_mode(size = display_sizes[-1], flags = pygame.FULLSCREEN)
+        pygame.event.set_grab(True)
+        pygame.event.pump() #Pump event queue to get window focus on macos
+        if not pygame.event.get_grab():
+            return
+        focused = pygame.key.get_focused()
+        self.assertIsInstance(focused, int)
+        self.assertTrue(focused)
```
MyreMylar, line 44: Just getting back to pygame after a summer holiday, apologies for the long wait.

My overall feedback is that this is a fine test for getting and testing a keyboard focus positive, but to fully test it is all working we'd want to try something we would expect would make the window _lose_ keyboard focus (i.e. something like minimising the window with `display.iconify()` - though you will want to check the test for that function to see what shenanigans are required for the iconify() to be registered by the OS and report back to pygame).

Complicating all of this (as with many unit tests) some of the window and display functionality does not work correctly on the 'dummy' display driver used on the continuous integration machines, I would not be at all suprised if you need to split this test into two functions - one, like your test above, that works on the CI - and one that has more testing in it but will only work on 'local' machines that use a real display driver.

You can mimic the 'dummy' test environment locally by setting the display driver used with something like:

```
import os

os.environ['SDL_VIDEODRIVER'] = 'dummy'
import pygame
````

(more info  here: https://www.pygame.org/wiki/DummyVideoDriver)

Just going to whiz through my inbox now but I may pop back to this later depending on time.

File: test/key_test.py (hunk 1 above)
Comment (leopoldwe) [reply test/key_test.py:30]: Yes, it appears that key.get_focused() returns true with SDL1 without a window to focus, at least on the build test machine. I do think it is odd, but I assumed the function was working as currently written.

File: test/key_test.py (hunk 2 above)
Comment (leopoldwe) [reply test/key_test.py:36]: No, returning causes the test to succeed. The idea I had was that if I was unable to get a fullscreen window to test, I would just stop the test.

File: test/key_test.py (hunk 3 above)
Comment (leopoldwe) [reply test/key_test.py]: I have changed the test to include iconifying the display. I also tested using SDL 1 and using the dummy display. I see get_focused() always returns true when using SDL 1. When using SDL 2 and a dummy display, get_focused always returns false. I have rewritten the test to sidestep most of the assertions when using a dummy display driver, but I have allowed the test to continue when using SDL 1, as always returning 1 seems like it is behaving incorrectly to me. Let me know if I should make more changes.

**Comment by MyreMylar** [comment]
I believe the display.quit() in this test is causing subsequent tests to fail on CI, tests sometimes run in an unusual order on CI. Perhaps it could be fixed with a quick update to:

```
    def setUp(cls):
        # This makes sure pygame is always initialized before each test (in
        # case a test calls pygame.quit()).
        if not pygame.get_init():
            pygame.init()

        if not pygame.display.get_init():
            pygame.display.init()
```

The iconify part also doesn't work on windows but I believe that is because of a thing I need to fix.

illume:
I tried this locally on a MacOS machine with python 2.7 and python 3.8, and the test passes.

illume:
I think we can move this iconify issue on windows into a separate bug report. For now I'll leave a note in the #1760 issue, and leave that open.
> The iconify part also doesn't work on windows but I believe that is because of a thing I need to fix.

illume (approved):
Looks good to me. I'll leave it open for MyreMylar to have another look. But it looks like you addressed all their concerns :)

MyreMylar (approved):

MyreMylar:
Agreed. Approved. 👍

PR #2027, opened 2020-07-31
Add SDL2-specific methods to Joystick; rename event fields

### LINKED ISSUE #2036: IDs in joystick events don't correspond to joystick "IDs"
I have been developing with gamepad support on Windows and Linux using Pygame 2.0.0dev10. On Linux this works. But on Windows I found that the `joy` attribute of a joystick event does not correct to the correct gamepad.

When I dug into this, I found SDL1 used device indices and doesn't support hotplugging. Device indexes are essentially the numbers given by `range(pygame.joystick.get_count())`.

SDL2 is very different: device indexes are a momentary snapshot, and each index may refer to different devices at different moments in time. Instead it introduces "instance IDs" which are persistent, and it is these that are given in joystick events (jstick->which is now an instance ID). But you can only get an instance ID from an open joystick object (until 2.0.12 when a function was added to get an instance for a device index).

As documented at the top of SDL_joystick.h:

> The term "device_index" identifies currently plugged in joystick devices between 0 and SDL_NumJoysticks(), with the exact joystick
> behind a device_index changing as joysticks are plugged and unplugged.
>
> The term "instance_id" is the current instantiation of a joystick device in the system, if the joystick is removed and then re-inserted
>  then it will get a new instance_id, instance_id's are monotonically increasing identifiers of a joystick plugged in.

So on Windows, with my gamepads at least, instance ids don't match device indexes even before hotplugging. Essentially, hotplugging is ingrained in the SDL2 joystick model, and joystick support cannot be correct without it.

To provide backwards compatibility with Pygame on SDL1, we would have to maintain a mapping from the instance ID provided by SDL events back to the device index a joystick was created with. This will work as long as no joystick devices are disconnected.

But joysticks do get disconnected, so that situation is also broken. It is not possible to make SDL2 follow the SDL1 model. I propose to also fully expose the SDL2 joystick model as a backwards-incompatible API, and deprecate using device indexes for anything.

lordmauve:
Fixes #2036: incorrect joystick IDs provided in joystick events.

Fully support hotplugging and the corresponding instance ids, while maintaining backwards compatibility (but only as long as no devices are plugged in/unplugged - it's impossible to do better than this).

SDL1 used device indices and doesn't support hotplugging. SDL2 is very different: device indexes can refer to different devices at different moments in time. Instead it introduces "instance IDs" which are persistent, and it is these that are given in joystick events (`jstick->which` is now an instance ID).  I found that instance IDs *do not match up* to device indexes, at least on Windows, even in the absence of hotplugging. So the status quo prior to this patch is:

* Stick state (Joystick objects) works
* The `joy` field of joystick events carries an instance ID, which you cannot match up to anything (Instance IDs are not exposed anywhere)
* Hotplugging doesn't work; you can't open the new joysticks

A note on device indexes: in SDL2, they're pretty much only valid at the moment you scan joysticks. However, existing code assumes they're always valid. Where do we get them from? There's Joystick.get_id(). pygame_sdl2 addressed this by [making Joystick.get_id() return the instance ID](https://github.com/renpy/pygame_sdl2/blob/f60bc1da402c4592311821f674da83c5378ec2fd/src/pygame_sdl2/joystick.pyx#L75).

However, get_id() is documented to return what you passed in in the constructor; I bet few people call it. They probably write:

```python
{id: Joystick(id) for id in range(get_count())}
```

So we don't have a way to start giving people instance IDs and never giving them device indexes; device indices are implied by the idea you can loop over `range(get_count())`. Maybe in future we should avoid letting users get hold of device indexes: deprecate `get_count()` and add a `available_devices()` which returns "Device objects" that can be `open()`ed.

Device indices will never work with hotplugging, of course. There's no way we can ever let people use device IDs and get working hotplugging.

The best we can do, if we want to offer compatibility, is map everything so that device IDs work as before as long as you don't unplug or plug in a joystick while a game is running. If you do, then all bets are off (your code won't match events to the correct device). 

What I have done is this:

- [x] Expose instance IDs on Joystick objects (and GUIDs, and other facts), as `get_instance_id()`
- [x] Drop the existing idea of storing SDL_Joystick objects in a table keyed by device_index - this is fundamentally what breaks hotplugging. I've replaced this with a doubly-linked list. Opening a new device searches this list and returns the existing instance, so now `Joystick(1) is Joystick(1)`. Before, the objects shared all state but were different objects.
- [x] Open joystick objects immediately that they are constructed, so that there's no possibility that joystick objects are constructed without being initialised and opened later after device indexes have been reallocated. This is more ergonomic anyway. However, it does prevent getting a device name without opening the Joystick object. I believe we should add a new discovery API that unifies sticks and game controllers, and started writing one, but that will be a different PR.
- [x] Enter instance_id -> device_index in a dict `pygame.event._joy_instance_map` as soon as we have it. If `SDL_JoystickGetDeviceInstanceID` is available this happens entirely in event.c, otherwise it happens when the Joystick object is initialised.
- [x] Use `pygame.event._joy_instance_map` to map instance_id to device_index for the existing joystick event `joy` attribute
- [x] Expose instance_id under a new event attribute, to enable correct future-proof behaviour.
- [x] Document it all.

I've tested this with instance_ids and hotplugging works perfectly. I also adapted my code to use the existing `joy` attribute  and it does work as long as you don't open/close joysticks after devices are removed.


lordmauve:
cc @illume - as discussed on Discord, longer write-up above

robertpfeiffer:
This is looking good. I think an example like a smash bros style character select screen with handling of unplugging and reconnecting of controllers should be in the examples directory, but we can add that later.

PR #2030, opened 2020-08-01
Fix sprite.collide_circle_ratio

### LINKED ISSUE #2029: sprite.collide_circle_ratio stores incorrect value in the radius attribute of the sprites.
When sprite.collide_circle_ratio is used to check for collision of sprites that do not have the "radius" attribute set, it calculates the radius and stores it in said attribute of the sprites so it can be used again next time.

The problem is that the radius is multiplied by the ratio **before** being stored.

Then all further collision checks on those sprites using sprite.collide_circle_ratio will get the radius multiplied by the ratio from that attribute and then multiply it **again** by the ratio. Any collision checks on those sprites using sprite.collide_circle will have a similar problem, the radius **multiplied by the ratio** will be fetched from that attribute and used instead of just the radius itself.

siggisv:
Fixes #2029 

- Add tests that reveals the bug in issue #2029 (incorrect value stored in the "radius" attribute of sprites)
- Fix sprite.collide_circle_ratio so it now stores the correct value in the "radius" attribute of those sprites that it is collision checking, that did not have that attribute previously set.

### DIFF
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -1483,27 +1483,27 @@ def collide_circle(left, right):
     ydistance = left.rect.centery - right.rect.centery
     distancesquared = xdistance ** 2 + ydistance ** 2
 
-    if hasattr(left, 'radius'):
+    try:
         leftradius = left.radius
-    else:
+    except AttributeError:
         leftrect = left.rect
         # approximating the radius of a square by using half of the diagonal,
         # might give false positives (especially if its a long small rect)
         leftradius = (0.5 * ((leftrect.width ** 2 +
                               leftrect.height ** 2) ** 0.5))
         # store the radius on the sprite for next time
-        setattr(left, 'radius', leftradius)
+        left.radius = leftradius
 
-    if hasattr(right, 'radius'):
+    try:
         rightradius = right.radius
-    else:
+    except AttributeError:
         rightrect = right.rect
         # approximating the radius of a square by using half of the diagonal
         # might give false positives (especially if its a long small rect)
         rightradius = (0.5 * ((rightrect.width ** 2 +
                                rightrect.height ** 2) ** 0.5))
         # store the radius on the sprite for next time
-        setattr(right, 'radius', rightradius)
+        right.radius = rightradius
     return distancesquared <= (leftradius + rightradius) ** 2
 
 
@@ -1562,24 +1562,24 @@ class collide_circle_ratio(object):  # noqa pylint: disable=invalid-name; this i
         ydistance = left.rect.centery - right.rect.centery
         distancesquared = xdistance ** 2 + ydistance ** 2
 
-        if hasattr(left, "radius"):
+        try:
             leftradius = left.radius
-        else:
+        except AttributeError:
             leftrect = left.rect
             leftradius = (0.5 * ((leftrect.width ** 2 +
                                   leftrect.height ** 2) ** 0.5))
             # store the radius on the sprite for next time
-            setattr(left, 'radius', leftradius)
+            left.radius = leftradius
         leftradius *= ratio
 
-        if hasattr(right, "radius"):
+        try:
             rightradius = right.radius
-        else:
+        except AttributeError:
             rightrect = right.rect
             rightradius = (0.5 * ((rightrect.width ** 2 +
                                    rightrect.height ** 2) ** 0.5))
             # store the radius on the sprite for next time
-            setattr(right, 'radius', rightradius)
+            right.radius = rightradius
         rightradius *= ratio
 
         return distancesquared <= (leftradius + rightradius) ** 2
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -1563,22 +1563,24 @@ class collide_circle_ratio(object):  # noqa pylint: disable=invalid-name; this i
         distancesquared = xdistance ** 2 + ydistance ** 2
 
         if hasattr(left, "radius"):
-            leftradius = left.radius * ratio
+            leftradius = left.radius
         else:
             leftrect = left.rect
-            leftradius = (ratio * 0.5 * ((leftrect.width ** 2 +
-                                          leftrect.height ** 2) ** 0.5))
+            leftradius = (0.5 * ((leftrect.width ** 2 +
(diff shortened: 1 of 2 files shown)


siggisv:
While fixing this bug, I couldn't help noticing that the same code is repeated four times, twice in sprite.collide_circle() and twice in sprite.collide_circle_ratio() with the only difference being that it is finding the radius of different sprites:

        if hasattr(left, "radius"):
            leftradius = left.radius
        else:
            leftrect = left.rect
            leftradius = (0.5 * ((leftrect.width ** 2 +
                                  leftrect.height ** 2) ** 0.5))
            # store the radius on the sprite for next time
            setattr(left, 'radius', leftradius)

This left me wondering if there are some speed optimization reasons for this or if it could be refactored into a single function.

siggisv:
After refactoring the radius calculation to be in only one place (inside the getter-method of the radius attribute of the sprite object) I ran some profiling tests where I compared the time needed to run `pygame.sprite.spritecollide(s, group, dokill=false, collided=pygame.sprite.collide_circle)` (where `s` is a single `pygame.sprite.Sprite` and `group` is an `pygame.sprite.AbstractGroup` containing 500 000 sprites).

The conclusion was that even after trying any optimization hint I could find (e.g. some of those in [https://wiki.python.org/moin/PythonSpeed/PerformanceTips](https://wiki.python.org/moin/PythonSpeed/PerformanceTips)), the time that the collision check would take would be about 4% longer for the refactored version compared to the original version (ca. 4% longer when the radius attribute had not been set and ca. 3.3% longer after it had been set).

If the slowing down of the `sprite.collide_circle` and `sprite.collide_circle_ratio` by some 4%, just to get some "cleaner code", is deemed acceptable then I could add the refactored version to this PR.

(Note that I did neither compare the speed of the creation of the sprites nor their memory usage in the refactored version vs the original version).

siggisv:
I did some more speed test on my refactored code, this time also measuring the time taken to create those 500 000 sprites.

Finding out that creating new sprites was about 9-10% slower and also considering that the speed of the collision checks after the attribute has been set is so much important than the check where the attribute has not been set (since it is at most once that each sprite goes through a `collide_circle` or `collide_circle_ratio` check without that attribute being set), I made changes to that code, ending with having three versions:

1. This current pull request that just fixes the bug and adds units tests.
2. A version where the radius calculations has been refactored into a getter-method of the sprite:
siggisv/pygame@1e38f06b07776b7c5421ac1109bc5bf069c30ac8
The creation of sprites and the circle-collision checks (with the radius attribute being set) are about as fast as in version 1 at the cost of the circle-collision checks (with that attribute not being set) being about 24% slower than in version 1.
3. A version where the radius calculations are still inline inside the `collide_circle()` and `collide_circle_ratio()` but have been optimized by using `try ... except` instead of `if hasattr() ... else`, making the collision checks about 12-13% faster when the attribute has been set, at the cost of them being about 16-17% slower the first time when the attribute has not yet been set:
siggisv/pygame@af2733494dd978eb4092a985c60a52a3519df312

Whoever will be reviewing this pull request, please let me know if you would like me to merge one of those two optimization attempts into this pull request.

MyreMylar:
I suspect your initial refactoring was floundering on the surprising fact that python loves extra, seemingly pointless, temporary variables for speed reasons - despite some linting tools having limits of the amount of 'unnecessary' temporary variables in a function and how messy it looks. I've run across the same struggle in the past trying to clean up old pygame code where the better looking code always runs slightly slower.

From your commentary, it sounds like version 3 would be preferable as anyone using a collide_radius type function should be setting up their sprite with a radius on setup anyway.

siggisv:
If I read correctly the output of `cProfile.run()` then the issue with my first attempt at refactoring the inline code into a getter-method is that it slowed down the collision checks a bit by adding an extra function call into the process. 

My attempts to optimize the refactored code were all centered around making the getter-method faster. When my final attempt (version 2) only managed to be as fast as the original code **when** the attribute is set, **at the cost** of being so much slower when the attribute is not set (by using `try ... except` and by assigning directly to `._radius` instead of calling the function `settattr()`), then I decided to check how much better the original code would be if I would use those optimizations in the original, inline code -- resulting in version 3. 

Regarding your worries of linter errors, I just tried running `pylint` on the original sprite code:

```bash
(venv-linting) siggisv@PyGameDev:~/work/projects/pygame-dev/pygame$ pylint src_py/sprite.py
src_py/sprite.py:1:0: C0302: Too many lines in module (1735/1000) (too-many-lines)
src_py/sprite.py:69:2: W0511: TODO:  a group that holds only the 'n' most recent elements. (fixme)                                                                                          
pygame 2.0.0.dev11 (SDL 2.0.9, python 3.7.3)
src_py/sprite.py:1097:4: R0914: Too many local variables (16/15) (too-many-locals)

------------------------------------------------------------------
Your code has been rated at 9.95/10 (previous run: 9.95/10, +0.00)
```

And then on my code (version 3). Looks like there are no new linter errors this time :) :

```bash
(venv-linting) siggisv@PyGameDev:~/work/projects/pygame-dev/pygame$ pylint src_py/sprite.py
src_py/sprite.py:1:0: C0302: Too many lines in module (1737/1000) (too-many-lines)
src_py/sprite.py:69:2: W0511: TODO:  a group that holds only the 'n' most recent elements. (fixme)                                                                                          
pygame 2.0.0.dev11 (SDL 2.0.9, python 3.7.3)
src_py/sprite.py:1097:4: R0914: Too many local variables (16/15) (too-many-locals)

------------------------------------------------------------------
Your code has been rated at 9.95/10 (previous run: 9.95/10, +0.00)
```

MyreMylar:
Well that's reassuring, hopefully we'll have some kind of CI linting pass one day that'll deal with this kind of thing automagically.

I say we go with version 3 then. I'm not sure if I can pull it from that commit into this PR somehow with some git magic I'm not clever enough to know, but hopefully you can just apply it from your end and we'll go from there.

siggisv:
I have added that commit to this PR.

MyreMylar:
I've now tested this locally reproducing the bug in dev10 and seeing it fixed with this changelist. Tests & pylint all pass locally as well.

Approved! 🎉 👍

MyreMylar (approved):
As above.

PR #2037, opened 2020-08-10
Added unit test: surface.get_pitch()

anguye13:
[#1808 ](https://github.com/pygame/pygame/issues/1808)

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1408,7 +1408,7 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
             rect2 = rect1.inflate(-width/2, -height/2)
             subsurf2 = surf.subsurface(rect2)
             sub_buff2 = subsurf2.get_buffer()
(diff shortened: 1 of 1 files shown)


test/surface_test.py
```diff
@@ -1388,7 +1388,29 @@ def todo_test_get_pitch(self):
         #
         # This value is not needed for normal Pygame usage.
 
-        self.fail()
+        # Test get_pitch() on several surfaces of varying size
+        sizes = ((2, 2), (7, 33), (33, 7), (2, 734), (734, 2), (734, 734))
+        for width, height in sizes:
```
MyreMylar, line 1393: My main thought here is - bit depths. It should be possible to add them into your tuples and then the creation of Surfaces and different bit depths should affect the values returned via `get_pitch()`, though if it works right it won't affect your tests as they start with `get_buffer()` which should also be affected in the same way.

anguye13:
I made the changes :)

MyreMylar (approved):
Excellent! Approved 👍 🎉

illume (approved):
🎉 thank you.

Issue #1751, opened 2020-05-16
Add interactive unit test: display.set_gamma_ramp()


MyreMylar:
This function has a unit test, but it now needs an interactive test (see comment below).

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/display_test.py#L313-L328
**Docs:** https://www.pygame.org/docs/ref/display.html#pygame.display.set_gamma_ramp
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/display.c#L2578-L2611 & https://github.com/pygame/pygame/blob/master/src_c/display.c#L2049-L2084

illume:
There is some further work to do on this test from the PR https://github.com/pygame/pygame/pull/2039

MyreMylar wrote:
>I feel like this test doesn't provide a great example of what setting the gamma ramp actually does. My intuition (based on no knowledge or testing) is that you set a look up table for each channel (e.g. red[50] now equals 60 instead of 50) and then that is applied to the display surface somehow. I assume it happens as some sort of final step when you .flip()/.update() the screen but having that show in a test somehow would be useful even if it is in a separate test that will only run locally. If it is a post .update()/flip() change to the display, that is tricky to test in the standard ways, like examining the pixels of the surface (as they will still be the same).

> So, failing any other good ideas coming to mind, my thought is that you could keep the current test but add a local interactive test that has a couple of dramatic gamma shifts displayed (e.g. start with a black screen shift blue[0] to 255) and ask users to confirm via the console that the window contents are blue/red/green or whatever ramp is set. Does that make sense?

> You can see examples of interactive tests in the cdrom or joystick test modules. They are basically a special type of test that only runs when the tests are run locally on a users machine and prompt for feedback.

PR #2039, opened 2020-08-10
Added unit test for display.set_gamma_ramp()

anguye13:
[#1751 ](https://github.com/pygame/pygame/issues/1751)

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -377,7 +377,7 @@ class DisplayModuleTest(unittest.TestCase):
         # hardware support gamma ramps, if the function succeeds it will
         # return True.
         #
-        pygame.display.set_mode(5, 5)
+        pygame.display.set_mode((5, 5))
         r = list(range(256))
         g = [number + 256 for number in r]
         b = [number + 256 for number in g]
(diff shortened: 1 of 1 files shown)


MyreMylar:
I feel like this test doesn't provide a great example of what setting the gamma ramp actually does. My intuition is that you set a look up table for each channel (e.g. red[50] now equals 60 instead of 50) and then that is applied to the display surface somehow. I assume it happens as some sort of final step when you `.flip()/.update()` the screen but having that show in a test somehow would be useful even if it is in a separate test that will only run locally. If it is a post .update()/flip() change to the display surface change that is tricky to test in the standard way by examining the pixels of the surface (as they will still be the same).

So, failing any other good ideas coming to mind, my thought is that you could keep the current test but add a local interactive test that has a couple of dramatic gamma shifts displayed (e.g. start with a black screen shift blue[0] to 255) and ask users to confirm via the console that the window contents are blue/red/green or whatever ramp is set. Does that make sense?

You can see examples of interactive tests in the cdrom or joystick test modules. They are basically a special type of test that only runs when the tests are run locally on a users machine and prompt for feedback.

anguye13:
Yea that makes sense, I’ll try and make the changes when I get a chance 👍

illume:
I think this test provides value, and that we can merge it in. I've added the extra test requirements into the #1751 issue (which we can leave open).

illume (approved):
Thanks :)

PR #2040, opened 2020-08-11
Unit test for image.save_extended()

anguye13:
Checking to see if image.save_extended() works as intended

### DIFF
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -740,13 +740,39 @@ class ImageModuleTest(unittest.TestCase):
 
         self.fail()
 
-    def todo_test_save_extended(self):
+    def test_save_extended(self):
+        surf = pygame.Surface((5, 5))
+        surf.fill((23, 23, 23))
 
-        # __doc__ (as of 2008-08-02) for pygame.image.save_extended:
+        passing_formats = ['jpg', 'png']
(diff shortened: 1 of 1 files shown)


anguye13:
save_extend seems to raise "pygame.error: Unrecognized image type" for all image types that aren't .jpg or .png

MyreMylar:
Looking at the [docs ](https://www.pygame.org/docs/ref/image.html#pygame.image.save)for `.save()` which I assume calls `save_extended()` pygame supports saving in these formats:

- BMP
- TGA
- PNG
- JPEG

It's possible that bmp saving is taken care of inside `save()` and not `save_extended()` as I think this is how it happens for load. I'm not sure about TGA files. Pygame definitely doesn't support saving in all of the types listed in your test so we should probably scope the testing correctly first.

anguye13:
Updated the error file types to only include .bmp and .tga and added tests to show save_extended() successfully work on .jpg and .png

MyreMylar:
I double checked the C code and indeed the regular 'save' function of the image module loads both .bmp and .tga file types by itself only passing .jpg (and .jpeg) or .png file types to save_extended().

MyreMylar (approved):
Approving this test 🎉 👍

Issue #1752, opened 2020-05-16
Add unit test: display.set_icon()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/display_test.py#L330-L351
**Docs:** https://www.pygame.org/docs/ref/display.html#pygame.display.set_icon
**Code to be tested**: https://github.com/pygame/pygame/blob/master/src_c/display.c#L2134-L2150 & https://github.com/pygame/pygame/blob/master/src_c/display.c#L2641-L2659

PR #2042, opened 2020-08-13
Adds a unit test of the display.set_icon() function. (Issue #1752)

DGMcKenney:
I've created a unit test to verify that using the set_icon() function returns None. I hope that this is what you wanted for this test!

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -439,7 +439,7 @@ class DisplayModuleTest(unittest.TestCase):
         with self.assertRaises(ValueError):
             palette = [[123,123,123]*10000]
             screen.set_palette(palette)
-        with self.assertRaises(ValueError):
+        with self.assertRaises(ValueError): 
             palette = [1,2,3]
             screen.set_palette(palette)
 
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -5,6 +5,7 @@ import os
 
 import pygame, pygame.transform
 from pygame.compat import unicode_
+from pygame.locals import *
 
 from pygame import display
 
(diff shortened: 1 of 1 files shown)


MyreMylar:
I think an interactive test might be more useful here if it can't be easily verify automatically that the icon has changed (I suspect it looks different on different OSes making automated detection hard).

An interactive test is when we prompt local users for feedback based on what is happening or being displayed on the screen, and is useful in cases where we don't have the hardware on the continuous integration platforms (e.g. cdroms or joysticks) or where it is hard to automatically verify whether a call has worked because it varies widely based on platform (this case). There are lots of helper functionality for these test and you can see examples in the joystick test module or the cdrom test module.

In this case I would set the icon of the window to a recognisable image or colour with the function and then ask the tester if the image is a banana (for example) instead of the normal pygame python. 

I would also bear this in mind from the documentation:

> Some systems do not allow the window icon to change after it has been shown. This function can be called before pygame.display.set_mode() to create the icon before the display mode is set.

DGMcKenney:
Hey, thank you for the feedback- I really appreciate it!

I've added an interactive test for this- displaying a new icon (a red square), and getting the user to press 'f' if the new icon is not a red square, and pressing any other key if it is a red square. I am not sure if that's what you were thinking or not, so let me know if you'd like something different.

I did have an issue where the window was being created off-screen, so I set the window position manually.

I also just noticed as I was typing this that the AppVeyor build failed. I'm afraid that is beyond my skill level to understand, but if there's something that I can do to fix it, please let me know.

Thanks!

DGMcKenney:
I created a new class for the interactive test, and included __tags__ = ["interactive"], but the AppVeyor failed again after my changes (the Travis CI is still in progress as of this writing).

I've compared what I wrote with joystick_test.py, and I'm afraid I can't find any differences that would explain why the CI platforms aren't ignoring it, based on what you've told me.

MyreMylar:
Travis has completed successfully, so that means your changes are good. 

Appveyor appears to have another generic issue downloading something from the internet this week -  possibly the pre-built windows .lib/.dll library files. Probably that means someone's web host is down, I'll investigate.

illume (approved):
👍 looks good to me. Thanks :)

illume:
ps. @DGMcKenney we use the black python code formatter https://github.com/psf/black

DGMcKenney:
Awesome, thanks! I really appreciate all your help and patience on my first open-source contribution. I had a good time and learned a lot!

PR #2050, opened 2020-08-21
type stub updates

judej:
I hope you find this stubs update useful: The following is what I have done:

1. Added stubs for a couple more classes (audio device)
2. Removed as many of the Any(s) as possible and made it specific from reading the code. This needs review since I am not a Pygame expert (although I have learned a lot from this exercise)
3. Removed the default value in the stubs
4. Made it acceptable by type shed by getting it to pass mypy's stub test (there is one error I could not resolve)
5. Fixed formatting and import sorting per type shed specs
6. Removed some stubs for members that should be private
7. Added types to untyped params and returns

Like I said I hope this is useful. While this makes the experience on editors that use stubs really nice - that is, it helps folks write correct code quickly, personally I would like to take this to the next step and add the typing to the source itself. But that is another conversation. :) thanks for letting me contribute.

Related to https://github.com/pygame/pygame/issues/839


illume:
Thanks @judej :)

It looks like your email address in the commit is not linked to your github profile? Can you please confirm it's correct or update the commit?

illume:
We don't have any tests for the stubs in

buildconfig/pygame-stubs/__init__.pyi
```diff
@@ -1,71 +1,55 @@
...
-import pygame.pixelarray
-import pygame.pixelcopy
-import pygame.sndarray
-import pygame.surfarray
-import pygame.math
-import pygame.fastevent
-
-# Other
-import pygame.scrap
+from . import bufferproxy as bufferproxy
```
illume, line 3: I'm not sure if this syntax is ok with py2/mypy/jedi/etc?

buildconfig/pygame-stubs/__init__.pyi
```diff
@@ -1,71 +1,55 @@
...
-@overload
-def encode_file_path(
-    obj: Any, etype: Optional[Type[Exception]] = UnicodeEncodeError
-) -> bytes: ...
-def register_quit(callable: Callable) -> None: ...
-def __getattr__(name) -> Any: ...  # don't error on missing stubs
+def encode_file_path(obj: Union[str, bytes, object], etype: Optional[Type[Exception]] = ...) -> bytes: ...
+def register_quit(callable: Callable[[], None]) -> None: ...
+
+# def __getattr__(name) -> Any: ...  # don't error on missing stubs
```
illume, line 55: Why was this commented out? Is it still needed?

buildconfig/pygame-stubs/camera.pyi
```diff
@@ -1,27 +1,23 @@
-from typing import List, Optional, Union, Tuple
+from typing import Optional, Sequence, Tuple, Union
+
 from pygame.surface import Surface
 
+
 def init() -> None: ...
 def quit() -> None: ...
-def list_cameras() -> List[str]: ...
```
illume, line 6: If it was an input type, I can see the benefit of a Sequence, but not on a return type. Seems to me List is better here?

buildconfig/pygame-stubs/__init__.pyi
```diff
@@ -1,71 +1,55 @@
...
 
 def get_error() -> str: ...
 def set_error(error_msg: str) -> None: ...
 def get_sdl_version() -> Tuple[int, int, int]: ...
 def get_sdl_byteorder() -> int: ...
 def encode_string(
-    obj: Union[str, bytes],
-    encoding: Optional[str] = "unicode_escape",
-    errors: Optional[str] = "backslashreplace",
-    etype: Optional[Type[Exception]] = UnicodeEncodeError,
```
illume, line 60: I wonder should defaults really not be used in the stubs?

@MightyJosip is there some benefit to having default values in the stubs?

buildconfig/pygame-stubs/freetype.pyi
```diff
@@ -60,62 +56,56 @@ class Font:
...
-        invert: Optional[bool] = False,
+        style: Optional[int] = ...,
+        rotation: Optional[int] = ...,
+        size: Optional[float] = ...,
+        invert: Optional[bool] = ...,
     ) -> Tuple[bytes, Tuple[int, int]]: ...
     def render_raw_to(
         self,
-        array: Any,
+        array: BufferProxy,
```
illume, line 104: I wonder if this accepts memoryview or things like arrays? I think it does, but haven't checked.

The python type system doesn't have support for buffer objects. I opened a thread on it in the types mailing list some months ago, but I don't think there's a good solution.

buildconfig/pygame-stubs/image.pyi
```diff
@@ -1,24 +1,18 @@
...
 
-from pygame.surface import Surface
 from pygame.bufferproxy import BufferProxy
+from pygame.surface import Surface
 
 _BufferStyle = Union[BufferProxy, bytes, bytearray, memoryview]
-_to_string_format = Literal['p', 'RGB', 'RGBX', 'RGBA', 'ARGB', 'RGBA_PREMULT', 'ARGB_PREMULT']
-_from_buffer_format = Literal['p', 'RGB', 'BRG', 'RGBX', 'RGBA', 'ARGB']
-_from_string_format = Literal['p', 'RGB', 'RGBX', 'RGBA', 'ARGB']
+_to_string_format = Literal["p", "RGB", "RGBX", "RGBA", "ARGB", "RGBA_PREMULT", "ARGB_PREMULT"]
```
illume, line 7: nitpick: We are moving to use the black python formatter code style which decided on single quotes.

buildconfig/pygame-stubs/image.pyi
```diff
@@ -1,24 +1,18 @@
...
+_from_buffer_format = Literal["p", "RGB", "BRG", "RGBX", "RGBA", "ARGB"]
+_from_string_format = Literal["p", "RGB", "RGBX", "RGBA", "ARGB"]
 
-def load(filename: Union[str, IO], namehint: Optional[str] = "") -> Surface: ...
-def save(surface: Surface, filename: Union[str, IO]) -> None: ...
+def load(filename: Union[str, IO[bytes]], namehint: Optional[str] = ...) -> Surface: ...
```
illume, line 11: What is this change for with filename? From`IO` -> `IO[bytes]`

buildconfig/pygame-stubs/mixer.pyi
```diff
@@ -34,17 +35,12 @@ def get_sdl_mixer_version(linked: bool) -> Tuple[int, int, int]: ...
 
 class Sound:
     @overload
-    def __init__(self, file: IO) -> None: ...
+    def __init__(self, file: IO[bytes]) -> None: ...
```
illume, line 38: Same question about `IO[bytes]` here as above.

buildconfig/pygame-stubs/sprite.pyi
```diff
@@ -27,41 +14,37 @@ class Sprite:
     image: Optional[Surface] = None
     rect: Optional[Rect] = None
     def __init__(self, *groups: AbstractGroup) -> None: ...
-    def update(self, *args, **kwargs) -> None: ...
+    def update(self, *args: str, **kwargs: int) -> None: ...  # TODO: verify
```
illume, line 17: There's a "TODO: verify" here. 

Think it was ok as it was before. I don't think args is of type str, and kwargs is of type int?

buildconfig/pygame-stubs/sprite.pyi
```diff
@@ -27,41 +14,37 @@ class Sprite:
...
     def __iter__(self) -> Iterator[Sprite]: ...
     def copy(self) -> AbstractGroup: ...
-    def sprites(self) -> List[Sprite]: ...
+    def sprites(self) -> Sequence[Sprite]: ...
     def add(self, *sprites: Sprite) -> None: ...
     def remove(self, *sprites: Sprite) -> None: ...
     def has(self, *sprites: Sprite) -> bool: ...
-    def update(self, *args, **kwargs) -> None: ...
-    def draw(self, surface: Surface) -> None: ...
+    def update(self, *args: str, **kwargs: int) -> None: ...  # TODO verify
```
illume, line 41: I think this is wrong too (as above).

buildconfig/pygame-stubs/sprite.pyi
```diff
@@ -27,41 +14,37 @@ class Sprite:
...
+    def draw(self, surface: Surface) -> Sequence[Rect]: ...
```
illume, line 42: I think this was correct before? The default Group implementation returns None.

buildconfig/pygame-stubs/sprite.pyi
```diff
@@ -72,58 +55,48 @@ class RenderClear(Group):
 
 class RenderUpdates(Group):
     def copy(self) -> RenderUpdates: ...
-    def draw(self, surface: Surface) -> List[Rect]: ...
+    def draw(self, surface: Surface) -> Sequence[Rect]: ...
```
illume, line 58: Maybe List is better here because it's a return type of List?

buildconfig/pygame-stubs/sprite.pyi
```diff
@@ -72,58 +55,48 @@ class RenderClear(Group):
...
 
 class OrderedUpdates(RenderUpdates):
     def copy(self) -> OrderedUpdates: ...
 
 class LayeredUpdates(AbstractGroup):
-    def __init__(self, *sprites: Sprite, **kwargs: Dict[str, Any]) -> None:
-        AbstractGroup.__init__(self)
+    def __init__(self, *sprites: Sprite, **kwargs: int) -> None: ...
```
illume, line 64: `**kwargs: int` here is wrong I think?

buildconfig/pygame-stubs/joystick.pyi
```diff
@@ -5,7 +5,7 @@ def quit() -> None: ...
 def get_init() -> bool: ...
 def get_count() -> int: ...
 
-class Joystick(object):
```
illume, line 8: Is (object) needed for py2 support?

buildconfig/pygame-stubs/surface.pyi
```diff
@@ -95,21 +73,19 @@ class Surface(object):
...
+    def subsurface(self, left: float, top: float, width: float, height: float) -> Surface: ...
     def get_parent(self) -> Surface: ...
     def get_abs_parent(self) -> Surface: ...
     def get_offset(self) -> Tuple[int, int]: ...
     def get_abs_offset(self) -> Tuple[int, int]: ...
     def get_size(self) -> Tuple[int, int]: ...
     def get_width(self) -> int: ...
     def get_height(self) -> int: ...
-    def get_rect(self, **kwargs) -> Rect: ...
+    def get_rect(self, **kwargs: Any) -> Rect: ...  # this needs to be Any. Applied to any attribute. eg. center - Tuple
```
illume, line 88: I don't understand this :) Could you please elaborate?

buildconfig/pygame-stubs/surface.pyi
```diff
@@ -1,49 +1,32 @@
...
+_ColorInput = Union[Color, str, Sequence[int], Tuple[int, int, int], Tuple[int, int, int, int]]
 _RgbaOutput = Tuple[int, int, int, int]
 _RectStyle = Union[
     Rect,
     Tuple[float, float, float, float],
     Tuple[Tuple[float, float], Tuple[float, float]],
-    List[float],
-    List[Vector2],
+    Sequence[float],
+    Sequence[Vector2],
```
illume, line 15: I'm not sure this is the case at the moment.

```python
>>> pygame.Rect({1,2,3,4})
TypeError: Argument must be rect style object
>>> pygame.Rect([1,2,3,4])
<rect(1, 2, 3, 4)>
```

buildconfig/pygame-stubs/surface.pyi
```diff
@@ -1,49 +1,32 @@
...
 from pygame.color import Color
-from pygame.rect import Rect
-
 from pygame.math import Vector2
+from pygame.rect import Rect
 
-_ColorInput = Union[
-    Color, str, List[int], Tuple[int, int, int], Tuple[int, int, int, int]
-]
+_ColorInput = Union[Color, str, Sequence[int], Tuple[int, int, int], Tuple[int, int, int, int]]
```
illume, line 8: I'm not sure this takes any sequence at the moment.

buildconfig/pygame-stubs/sprite.pyi
```diff
@@ -72,58 +55,48 @@ class RenderClear(Group):
...
-    def __init__(self, *sprites: DirtySprite, **kwargs: Dict[str, Any]):
-        LayeredUpdates.__init__(self, *sprites, **kwargs)
+    def __init__(self, *sprites: DirtySprite, **kwargs: int) -> None: ...
     def copy(self) -> LayeredDirty: ...
-    def draw(self, surface: Surface, bgd: Optional[Surface] = None) -> List[Rect]: ...
+    def draw(self, surface: Surface, bgd: Optional[Surface] = ...) -> Sequence[Rect]: ...
     def clear(self, surface: Surface, bgd: Surface) -> None: ...
     def repaint_rect(self, screen_rect: _RectStyle) -> None: ...
-    def set_clip(self, screen_rect: Optional[_RectStyle] = None): ...
+    def set_clip(self, screen_rect: Optional[_RectStyle] = ...) -> None: ...  # TODO: Verify
```
illume, line 88: There's a TODO comment here.

set_clip returns None, or raises an error.

buildconfig/pygame-stubs/sprite.pyi
```diff
@@ -72,58 +55,48 @@ class RenderClear(Group):
...
     def get_clip(self) -> Rect: ...
-    def set_timing_treshold(
-        self, time_ms: SupportsFloat
-    ) -> None: ...  # This actually accept any value
+    def set_timing_treshold(self, time_ms: SupportsFloat) -> None: ...  # This actually accept any value
```
illume, line 90: This accepts int or float (now... I think this changed only in the last months). Remove the comment?

buildconfig/pygame-stubs/sprite.pyi
```diff
@@ -72,58 +55,48 @@ class RenderClear(Group):
...
     def copy(self) -> GroupSingle: ...
 
 def spritecollide(
-    sprite: Sprite,
-    group: AbstractGroup,
-    dokill: bool,
-    collided: Optional[Callable[[Sprite, Sprite], bool]] = None,
-) -> List[Sprite]: ...
+    sprite: Sprite, group: AbstractGroup, dokill: bool, collided: Optional[Callable[[Sprite, Sprite], bool]] = ...,
+) -> Sequence[Sprite]: ...
```
illume, line 99: Not sure this should be Sequence when it returns List.

buildconfig/pygame-stubs/sprite.pyi
```diff
@@ -135,7 +108,7 @@ def collide_circle(left: Sprite, right: Sprite) -> bool: ...
 
 class collide_circle_ratio:
     ratio: float
-    def __init__(self, ratio: float): ...
+    def __init__(self, ratio: float) -> None: ...
```
illume, line 111: Is this correct for an init returning None?

buildconfig/pygame-stubs/sprite.pyi
```diff
@@ -144,10 +117,9 @@ def groupcollide(
     group2: AbstractGroup,
     dokill: bool,
     dokill2: bool,
-    collided: Optional[Callable[[Sprite, Sprite], bool]] = None,
-) -> Dict[Sprite, Sprite]: ...
+    collided: Optional[Callable[[Sprite, Sprite], bool]] = ...,
+) -> Mapping[Sprite, Sprite]: ...
```
illume, line 121: I think it should return Dict type here, not Mapping? Since it returns a dict.

buildconfig/pygame-stubs/transform.pyi
```diff
@@ -1,50 +1,39 @@
-from typing import Tuple, List, Union, Optional, Sequence
-from pygame.surface import Surface
-from pygame.math import Vector2
+from typing import Optional, Sequence, Tuple, Union
+
 from pygame.color import Color
```
illume, line 3: I guess these multiple import lines could be changed to:
`from pygame import Color, Rect, Surface, Vector2`

illume (changes requested):
👍 I've left some questions on various lines.

buildconfig/pygame-stubs/__init__.pyi
judej: Re-read the specs and recommendations on this. Will update

buildconfig/pygame-stubs/__init__.pyi
judej: I commented it out so that only the methods stubbed here are recognized in strict mode. Have uncommented it.

buildconfig/pygame-stubs/freetype.pyi
judej: Will do

buildconfig/pygame-stubs/__init__.pyi
judej: The two issues with having defaults in stubs is a) it can become out of sync with the source and lead to unexpected behavior b) mypy explicitly requires that there be no default values.

buildconfig/pygame-stubs/camera.pyi
judej: Fixed

buildconfig/pygame-stubs/joystick.pyi
judej: Yes, will put all those back. 

Are there many developers using Python 2 for Pygame? I noticed there were some tutorials that used Python 2. But was not clear about the prevlaence.

buildconfig/pygame-stubs/sprite.pyi
judej: found issue 5876 in mypy where BDFL ok'ed it. Will revert

buildconfig/pygame-stubs/sprite.pyi
judej: Fixed

buildconfig/pygame-stubs/sprite.pyi
judej: fixed

buildconfig/pygame-stubs/sprite.pyi
judej: Changed that to Optional[_RectStyle] which is effectively `Union[_RectStyle, None`

buildconfig/pygame-stubs/sprite.pyi
judej: Done

buildconfig/pygame-stubs/sprite.pyi
judej: removed `int`

buildconfig/pygame-stubs/sprite.pyi
judej: fixed

buildconfig/pygame-stubs/sprite.pyi
judej: made it a list

buildconfig/pygame-stubs/sprite.pyi
judej: Yes, PEP 484

buildconfig/pygame-stubs/sprite.pyi
judej: Fixed

buildconfig/pygame-stubs/surface.pyi
judej: Will revert to List

buildconfig/pygame-stubs/surface.pyi
judej: Will revert

buildconfig/pygame-stubs/surface.pyi
judej: Removing the Any

buildconfig/pygame-stubs/transform.pyi
judej: Changing all the imports to be import the module and not the symbol. So, it will be:
from . import color

and when using it, it will be color.Color - per guidance

buildconfig/pygame-stubs/image.pyi
judej: I am using the Black formatter as required by mypy, and black converts to double quotes.

buildconfig/pygame-stubs/image.pyi
judej: The IO can be bytes or str, in this case I figured it had to be bytes

buildconfig/pygame-stubs/__init__.pyi
judej: fixed
Changed all the "from . import foo as foo" to " from . import foo"

buildconfig/pygame-stubs/__init__.pyi
judej: Fixed

buildconfig/pygame-stubs/camera.pyi
judej: fixed

buildconfig/pygame-stubs/freetype.pyi
judej: done

buildconfig/pygame-stubs/mixer.pyi
judej: IO[AnyStr] can be str or bytes; since this is a sound file, I presuming it has to be bytes

judej:
@illume, @MightyJosip, please lmk if there are other changes you would like.
thanks
Jude

MightyJosip (commented):
Hi, sorry for late response. Some notes:
1. You need to include all constants in pygame namespace (they can be called by writing for example: pygame.QUIT, and in your case you need to write pygame.constants.QUIT.
2. There are still some Union[Sequence being returned, and key.get_pressed should return Sequence (it actually returns <class 'pygame.key.ScancodeWrapper'>, and the Sequence is the most similar "class" to it
3. We will probably keep default values because they provide extra information to users in some editors. For example in PyCharm with default values
a 
and without default values:
b
If we have default values written than the user can see that the default frequency is 44100 and similar information.
4. if you replace all List[int], Tuple[int, int, int] with Sequence[int], than the functions would accept the tuples with any amount of elements, even though it must have 3 elements and other similar examples

judej:
@illume, @MightyJosip, sorry for the delay in adding back the default values. Working on it as time slices allow :)

Question: what is your opinion in just adding the types inline - that way:
- we can keep all the default values and have type info
- have type info right there in the code and don't have to go update the type info in the .pyi files
- can do it gradually, file by file - start the files with no dependencies and making progress

I am volunteering to do this :)

MyreMylar:
I believe that is quite awkward to do for python 2.7, and also likely for our modules written in C. I think you have to add something like:

`thing_function.__annotations__ = {'param_type_hint_1': str, 'param_type_hint_2': int, 'return': None}` 

...to make it work, rather than the neat syntax in Python 3. Unless you know otherwise?

MightyJosip:
If you know how to do it you can

judej:
For compiled code, I agree, we will need type stubs. no way around that.

I was interested in the Python code in PyGame. For that we can use type comment annotations in the code (to maintain compatibility with Python 2). If there wasn't a Python 2 constraint, we could do regular type annotations in the code. We can also incorporate the default values (2 and 3 compatible). 

thoughts?

Also, is there a significant number for Python2 apps that PyGame is being used in? Is that diminishing? Do you have a plan to discontinue support at some time?

MyreMylar:
> Also, is there a significant number for Python2 apps that PyGame is being used in? Is that diminishing? Do you have a plan to discontinue support at some time?

Python 2 is still the default python on macs and is installed on current Debian and probably other linux distros too. Which means that a lot of users on those platforms doing `pip install pygame` end up installing it on python 2. Luckily, Apple are removing Python 2 in the next edition of OSX and the same is true of Debian which should mean installs of python 2 will start drying up much faster from next year or so when those are both out.

I do think we should start tracking stats from PyPI at least on how many python 2 wheel downloads we get to track this change, with a view towards eventually cutting people off when the numbers get very low - but I haven't personally looked into it yet. Backwards compatibility has been the big focus for the initial release of pygame 2.0 rather than cutting users off, so removing Python 2 support will be a thing for future versions like Pygame 2.1/2.2+

PR #2053, opened 2020-08-23
Add optional 'namehint' argument to image.save

### LINKED ISSUE #48: Add optional 'format' argument to pygame.image.save
Originally reported by: **René Dudfield (Bitbucket: [illume](https://bitbucket.org/illume), GitHub: [illume](https://github.com/illume))**

== Arnau Sanchez, 2010-05-19 11:22:25 -0700

    
    Sometimes we need to store a surface image into a string instead of a filename. I tested with StringIO and it worked:
    
    import StringIO
    fd = StringIO.StringIO()
    pygame.image.save(surface, fd)
    print len(fd.getvalue()) 
    # 1334
    
    However, since we have no real filename, we cannot set the file format. That's why I propose to add a third optional argument that forces the function to use a given format:
    
    pygame.image.save(Surface, filename, format="TGA"): return None
    

----

- Bitbucket: https://bitbucket.org/pygame/pygame/issue/48

**Related Docs:** https://www.pygame.org/docs/ref/image.html#pygame.image.save

siggisv:
Resolves #48

- [x] Add unittests.
- [x] Make changes to `src_c/image.c` and `src_c/imageext.c` so that when saving a surface to a fileobject (instead of passing in a string contained the name of the file to be created), an optional `namehint` argument can be given to chose the filetype.
Works for:
    - [x] TGA
    - [x] BMP
    - [x] PNG
    - [x] JPEG ~~(this should work but doesn't)~~
 - [x] Update documentation to reflect the changes.
 - [x] Update type hints.

For some reason that I haven't figured out yet, the `write_jpeg()` function either throws an 'Segmentation Fault' or 'Output file write error --- out of disk space?' when writing to a `SDL_RWops*` object that was created from a fileobject (using `pgRWops_FromFileObject()`) instead of one created from a filename string (using `SDL_RWFromFile()`). At the same time, `write_png()` has no problem with writing to either one of them.

Note: Have tried using `IMG_SaveJPG()` and `IMG_SaveJPG_RW()` (they have been part of SDL2_image since version 2.0.2) instead of the internal `SaveJPG()` and that worked perfectly.


siggisv:
Saving a jpeg to a fileobject by using the namehint argument now works as long as using Pygame that is built with SDL_Image 2.0.2 or newer (as long as that was built with jpeglib).
Trying that with SDL_Image older than 2.0.2 will raise an error explaining that SDL_Image 2.0.2 or newer is needed.

Now I just need to... huh? ... wait, ... what? ... why?
Why is travis throwing an error on my use of the RAISE macro? :O
I am using it in exactly the same way as the original code was using it in multiple places. wtf?
I need some sleep.

One night's sleep and a few hours later I realize that the original code was using the macro "incorrectly" (it's purpose is to make `return RAISE(error_type, error_string);` call `PyErr_SetString()` while returning a `(PyObject *)NULL` at the same time) but it seems the original code passed all test because it was within preprocessor directives that turned it off unless building without jpeglib or png.h

A short grep confirmed that those were the only instances of incorrect use of that macro in the whole codebase.

I can fix that :D

MyreMylar:
It's looking like the version of SDL_image on Travis (It's on SDL 2.0.4 - so from 2014) does not have SDL_image 2.0.2 which I believe came out in 2018. I expect we can update the SDL library versions on Travis somehow.

siggisv:
I will start with changing the unittest to expect an error to be raised when using SDL_Image older than 2.0.2.

siggisv:
It looks like I might have to add a `get_sdl_image_version()` function. I see that we have a `get_sdl_version()` defined in `src_c/base.c` and a `get_sdl_mixer_version()` defined in `src_c/mixer.c` but haven't found anything similar for SDL_Image.
That should be doable.

illume:
Nice work so far :) Looking forward to this.

siggisv:
Yay! The Microsoft compiler used to compile for Python 2.7 on Windows seems to be stuck in the early 90s:

> Compiler error C2143: "In a C program, variables must be declared at the beginning of the function, and they cannot be declared after the function executes non-declaration instructions." [https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2143?view=vs-2019](https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2143?view=vs-2019)

C99 took that restriction away more than 20 years ago.

Oh well. I can fix that.

siggisv:
Found a 90s playlist to get me in the mood to code like it's still the last century. Hope Bill is happy now.

Looks like I now only have to update the documentations and the type hints.

siggisv:
This should be ready now but I hope I didn't mess anything up with my update to the docs.
I tried running `python setup.py docs` but even when switching back to upstream/master, it results in Sphinx (v3.2.1) throwing a bunch of warnings on some of the `pygame/docs/reST/c_api`-files before coming to a halt with the following error:
```
Extension error:
Handler <function collect_document_info at 0x7f568b4127b8> for event 'doctree-read'
threw an exception (exception: No fullname: desc's child has empty names list)   
```

MyreMylar:
Looks like the docs are only compiling with sphinx 2 at the moment. If you install that version it all builds.

docs/reST/ref/image.rst
```diff
@@ -93,18 +96,40 @@ following formats.
...
+
+   .. note:: To be able to save the ``JPEG`` file format to a file-like object,
+   SDL2_Image version 2.0.2 or newer is needed.
+
+   .. note:: When saving to a file-like object, it seems that for most
+   formats, it needs to be flushed after saving to be able to load from it.
 
    .. versionadded:: 1.8 Saving PNG and JPEG files.
+   .. versionchanged:: 2.0.0.dev11 The *namehint* parameter was added to make
+   it possible to save other formats than ``TGA`` to a file-like object.
```
MyreMylar, line 115: to make these work properly with the docs word wrapping you need to indent the second line of each `note `and `versionchanged` block so they match the line above, something like so:

```
   .. note:: To be able to save the ``JPEG`` file format to a file-like object,
             SDL2_Image version 2.0.2 or newer is needed.

   .. note:: When saving to a file-like object, it seems that for most
             formats, it needs to be flushed after saving to be able to load from it.

   .. versionadded:: 1.8 Saving PNG and JPEG files.
   .. versionchanged:: 2.0.0.dev11
                       The *namehint* parameter was added to make it possible
                       to save other formats than ``TGA`` to a file-like object.

```

docs/reST/ref/image.rst
```diff
@@ -93,18 +96,40 @@ following formats.
...
 
    .. ## pygame.image.save ##
 
+.. function:: get_sdl_image_version
+
+   | :sl: `get version number of the SDL_Image library used`
+   | :sg: `get_sdl_image_version() -> None`
+   | :sg: `get_sdl_image_version() -> (major, minor, patch)`
```
MyreMylar, line 123: You need to delete the spaces between `:sg:` and the start of the description so it looks like so:

```
   | :sl:`get version number of the SDL_Image library used`
   | :sg:`get_sdl_image_version() -> None`
   | :sg:`get_sdl_image_version() -> (major, minor, patch)`
```

MyreMylar (changes requested):
I wrote a long thing up about supporting keyword arguments - but I think it might end up being more confusing because of the type overloading that `save()` has going on for the second parameter (what single keyword name do you give a parameter that is either a file object or a file name?). I like keyword arguments, but ultimately pygame wasn't designed with them in mind.

Overall, it works! I tested it by loading a gif into a file object and saving it back out to a jpg using the namehint. I got the docs compiling and noticed a couple of teeny formatting issues.

MyreMylar (approved):

Issue #1742, opened 2020-05-16
Add unit test: display.flip()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/display_test.py#L51-L66
**Docs:** https://www.pygame.org/docs/ref/display.html#pygame.display.flip
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/display.c#L1609-L1675

amipy:
I tried to make a unit test for display.flip(). This is only my second time, so if I do something incorrectly, please tell me.
#2054

PR #2054, opened 2020-08-23
Added a unit test for display.flip()

amipy:
This is a unit test for display.flip().
I was unable to come up with a method to test the last error condition in the code, because that condition involves SDL giving an error not caught by pygame, and I do no know how to do that.


amipy:
Here is the issue I wrote this for.
#1742

MightyJosip:
This unit test should be part of [this file](https://github.com/pygame/pygame/blob/master/test/display_test.py#L51-L66)

amipy:
Alright, I moved it to the display_test file.

amipy:
There was an error in my test, so I fixed it.

illume (approved):
Thanks 🎉

PR #2055, opened 2020-08-24
set_system_cursor function added

Starbuck5:
Provides a way to set the cursor to a system specific variant using SDL_CreateSystemCursor.

Added documentation. I rephrased a bit of the mouse cursor documentation, because it referred to the mouse cursor as the "system cursor", which would be confusing with set_system_cursor.

I've never written a unit test before, but I did my best to use the other mouse tests as examples. I also used the black code formatter on mouse_test.py


src_c/constants.c
```diff
@@ -1056,6 +1056,22 @@ MODINIT_DEFINE(constants)
     DEC_CONST(APPACTIVE);
 #endif /* IS_SDLv2 */
 
+/* cursor constants */
+#if IS_SDLv2
+    DEC_CONST(SYSTEM_CURSOR_ARROW);
```
MightyJosip, line 1061: Hi. Since you modified constants can you also modify [type hints](https://github.com/pygame/pygame/blob/master/buildconfig/pygame-stubs/constants.pyi) for it. You can just run the commented script at the top and you will get new file

MightyJosip (changes requested):

src_c/constants.c
Starbuck5: Got it. Thanks.

illume:
Thanks! 🎉

illume (approved):
👍

PR #2061, opened 2020-08-30
Fix mouse.set_system_cursor when using SDL1 to do nothing.

illume:
SDL1 broke with https://github.com/pygame/pygame/pull/2055/

This makes the function do nothing on SDL1.


Starbuck5:
Thanks for the fix illume.

PR #2062, opened 2020-08-31
Fix grammar in sprite docs

douglas-cpp:


MyreMylar (approved):
Every little change helps make the docs more readable.

PR #2063, opened 2020-08-31
Added tests for pygame.transform.smoothscale(); fixes #1758

### LINKED ISSUE #1758: Add unit test: transform.smoothscale()

**Test stub:** https://github.com/pygame/pygame/blob/master/test/transform_test.py#L1043-L1062
**Docs:** https://www.pygame.org/docs/ref/transform.html#pygame.transform.smoothscale
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/transform.c#L1416-L1488

adamandrews1:


adamandrews1:
Fixes #1758 . We may want to add a more sophisticated pixel-color-blending test portion. @MyreMylar

test/transform_test.py
```diff
@@ -1176,8 +1176,33 @@ def todo_test_smoothscale(self):
...
+            smoothscaled_surface = pygame.transform.smoothscale(starting_surface, (10, 10))
+        self.assertRaises(ValueError, smoothscale_low_bpp)
+
+        def smoothscale_high_bpp():
+            starting_surface = pygame.Surface((20, 20), depth=48)
+            smoothscaled_surface = pygame.transform.smoothscale(starting_surface, (10, 10))
+        self.assertRaises(ValueError, smoothscale_high_bpp)
+
+        def smoothscale_invalid_scale():
+            starting_surface = pygame.Surface((20, 20), depth=36)
```
MyreMylar, line 1191: should the depth here be 32 ?

MyreMylar (changes requested):
I suspect you could write a reasonable test of the blending filter algorithm by doing something like:

```
two_pixel_surface = pygame.Surface((2, 1), depth=32)
two_pixel_surface.fill(pygame.Color(0, 0, 0), pygame.Rect(0, 0, 1, 1))
two_pixel_surface.fill(pygame.Color(255, 255, 255), pygame.Rect(1, 0, 1, 1))

two_five_six_surf = pygame.transform.smoothscale(two_pixel_surface, (256, 1))
self.assertEqual(two_five_six_surf.get_at((128, 0)), pygame.Color(127, 127, 127))
```

And perhaps something similar starting larger and going the other way to see if you get roughly a midway blend at the halfway point.

test/transform_test.py
adamandrews1: It definitely should be. This is not the value error we are looking for in this test! Thanks

MyreMylar (approved):
Looks good, there is another (already existing) test that covers some other functionality of smoothscale and some tests of setting it's backend as well so I'm happy that we have pretty good coverage of it with this test as well.

Thanks for the contribution! 👍 🎉

adamandrews1:
Thanks for the thoughtful guidance!

PR #2065, opened 2020-09-01
use rwops for font loading

robertpfeiffer:
This is needed for console/android export. Fonts, sounds and images already use RWops. This code path should work without any adjustments on PC/desktop platforms.

(Custom text data is outside of the scope of this pull request)


MyreMylar (approved):
Looks good to me, gave it a good eyeballing - ran it through the font tests locally and through my GUI module (since that uses a lot of fonts). Everything seems to work just as it did before on windows.

🎉 👍

PR #2068, opened 2020-09-03
Alphabetised colourdict.py colour names & minor linting

### LINKED ISSUE #2067: Color Dictionary (colordict for Color class constructor)
It would be handy to change the definition of THECOLORS in colordict.py so that the source listing is in alphabetical order by color name.  Unless the order is currently optimized (most used first?), it would be useful to see all grays together, all reds together, etc., in order to make decisions while coding.  Obviously, I can do this for myself, but I use the online docs quite a bit.  Thanks.

MyreMylar:
closes #2067 . 

As the title, I don't see any reason for the dictionary keys here to be in a random order and we needed to do some linting on this file anyway.


nthykier (approved):
LGTM

illume (approved):
👍

PR #1954, opened 2020-06-16
Update event.rst

red436:
Comment for proposed documentation link.

PR #2072, opened 2020-09-04
Added link in event docs to key constants list.

MyreMylar:
There was a PR ([here](https://github.com/pygame/pygame/pull/1954)) that pointed out that it wasn't easy to find the key constants from the event module docs, they are however already included in the 'key' module docs so I added a link from the event docs to the list in the key docs.

I think it'll be useful to new users as the K_ type key constants are frequently used in combination with the KEYUP & KEYDOWN event type constants that are explained on this page.


illume (approved):
👍

PR #2073, opened 2020-09-04
GitHub issue templates: New style

MyreMylar:
This is an update to [this old PR](https://github.com/pygame/pygame/pull/1626) that travis hated, with the templates slightly-redone in the new GitHub style. If it works it should present users with a choice when they click 'New Issue' on GitHub that looks something like [this (from pygame_gui)](https://github.com/MyreMylar/pygame_gui/issues/new/choose
):

Then from there it auto fills in an outline for a new issue into the text box (which you can also just delete) and adds some labels we can specify, we could also assign issues of a certain type to someone if we ever wanted to do that. 

The main upsides: 
 - some scaffolding will hopefully encourage/remind issue reporters to fill in details like their platform and pygame version so that we don't have to always ask that as the first question.
- Using the system provides some tools to automate issue labelling/assignment that we could go further with one day.

The slight downsides I can think of:
 - I believe it means we'll have to add the `.github` folder to the repository, I think this is the data folder GitHub is using for all it's new repository features like the built in CI type tool called 'workflows' (though in my experience that is actually pretty speedy compared to Travis so we might want to use it one day).
- The bug report template might discourage some people from reporting issues when they see the amount of details asked for. Though I'm not too bothered by that because a half finished bug report without enough details is often useless anyway.

We could take this issue templating further and add more types of issues if people like it.

**Caveat:** I'm not 100% this will work perfectly, I've based it off existing templates but you can't test these GitHub features in branches apparently.


illume:
TLDR; I'm kind of against templates and the .github folder without some discussion and consensus. I think it's worth discussing various aspects of the problems more before we decide on .github issue templates as a solution.


### issue label automation
Do you have a link to the issue labeling automation tools you're interested in?

Is labeling actually useful, and worth the effort of doing it manually? It's kind of useful to be able to click a 'docs' link and see all docs related issues. On the other hand I can just search for 'docs' and find much the same list. It's kind of useful to see in PRs and issues to get a quick visual idea of what areas they are related to.


## platform, and version details 

An alternative to increase the amount of people reporting the platform could be to include it in the pygame welcome message next to the pygame version and SDL version. Since the welcome message has been in pygame, people paste it more often when reporting issues.

> pygame 2.0.0.dev10 (SDL 2.0.12, python 2.7.16, Darwin-19.6.0-x86_64-i386-64bit)
> pygame 2.0.0.dev8 (SDL 2.0.12, python 2.7.16, Windows-10-10.0.10240-SP0)

Sometimes platform and version information is relevant to issues. But how often? Looking at previous issues shows that mostly the platform and version info is not actually interesting for the majority of issues reported, and that many people do report version/system information.


### .github (and config file soup)
.github folder hurts, as does the .travis.yml and all other config files. Travis is still very useful because it provides multiple platform CI. There are alternatives for Mac and other OS now. Also there are separate solutions for ARM, and power/ppc available... but not with 1 line of config for each like with Travis. I don't see us replacing Travis any time soon... but I'm very interested if someone could replace it and clean up our repo. Another option is to replace src_py and src_c with src/c and src/py.

The real question is... does the .github folder provide us with significant useful things? We had this discussion about the setup.cfg, and decided that various tools can be configured in there... and that outweighed the negative part of it cluttering up the root of the repo.


### issue templates

- stops people reporting by presenting people with a wall of text.
- feels less human, and less friendly to have bug templates.
- it makes the issues harder to read. Because the signal is diluted with the wall-of-text template, which often aren't important to the issue. If the version number of something isn't important, it shouldn't be written there. Repetitively reading the same format issue reports written in the same way over and again is more soul destroying imho.

For me, these outweigh the pros for having them. Mostly it's because the issue is worse to act on.

There's a difference between github templates that pop up automatically, and workflows and templates people can use. I think it's useful to be able to document our processes for different issues and guide people forward.

### issue triage, and next steps.

Mostly issue triage needs to happen -- Is it a problem, can it be reproduced, will we fix it, are we open to other people fixing it, what are the next steps. Sometimes people have time to write down just the subject of the issue. Sometimes they will go much further.

How do they make proposals, how do they fix documentation, how do they have a conversation about a topic or experiment?

MyreMylar:
TLDR; I'm not that invested in having, or not having, issue templates. I was mostly just lookign to clear out old PRs that seemed to be stuck in limbo and realised that there was anew format for issue templates since the originasl PR was made.
> 
> ### issue label automation
> Do you have a link to the issue labeling automation tools you're interested in?

The tool here is just the template, it has a little field at the top of the template that lets you auto assign some labels to a new issue depending on the template the reporter picks. So a bug report would get the 'bug' tag and an 'enhancement' would get the 'enhancement' tag.

> 
> Is labeling actually useful, and worth the effort of doing it manually? It's kind of useful to be able to click a 'docs' link and see all docs related issues. On the other hand I can just search for 'docs' and find much the same list. It's kind of useful to see in PRs and issues to get a quick visual idea of what areas they are related to.
> 

I find the labels fairly useful when I want to give people a list of themed/related issues (i.e. 'Here are all the good first issues on windows' or 'here are all the problems we've had on ArchLinux with the mixer module') and I think the tags probably help the search tool return better results? You could probably also do a reasonable job of this by updating the issue titles extensively and using search too. Having the titles there does serve as prompt to add a bit of extra detail to the issue with a click sometimes too. That's how I'm using them mostly right now anyway.

> ## platform, and version details
> An alternative to increase the amount of people reporting the platform could be to include it in the pygame welcome message next to the pygame version and SDL version. Since the welcome message has been in pygame, people paste it more often when reporting issues.
> 
> > pygame 2.0.0.dev10 (SDL 2.0.12, python 2.7.16, Darwin-19.6.0-x86_64-i386-64bit)
> > pygame 2.0.0.dev8 (SDL 2.0.12, python 2.7.16, Windows-10-10.0.10240-SP0)
> 
> Sometimes platform and version information is relevant to issues. But how often? Looking at previous issues shows that mostly the platform and version info is not actually interesting for the majority of issues reported, and that many people do report version/system information.

I agree it depends on the issue, I'm thinking a lot of how quite often when we get a 'pygame isn't working' type issue posted it turns out to be because the user was using pygame 1.9.6 on Mac OS for example, but if they don't say they were on mac up front there can be a bit of a 'are you on a recent macOS?' 'Yes', 'OK you need pygame 2' type conversation which we might be able to cut down a step. It's obviously also helpful to know details when you get a hard to pin down bug that we struggle to reproduce.

> 
> ### .github (and config file soup)
> .github folder hurts, as does the .travis.yml and all other config files. Travis is still very useful because it provides multiple platform CI. There are alternatives for Mac and other OS now. Also there are separate solutions for ARM, and power/ppc available... but not with 1 line of config for each like with Travis. I don't see us replacing Travis any time soon... but I'm very interested if someone could replace it and clean up our repo. Another option is to replace src_py and src_c with src/c and src/py.
> 
> The real question is... does the .github folder provide us with significant useful things? We had this discussion about the setup.cfg, and decided that various tools can be configured in there... and that outweighed the negative part of it cluttering up the root of the repo.

I guee how useful the folder will end up being will be determined by if & how much we use GitHub workflows in the future rather than issue templates which are much more of a trivial concern.

> ### issue templates
> * stops people reporting by presenting people with a wall of text.
> * feels less human, and less friendly to have bug templates.
> * it makes the issues harder to read. Because the signal is diluted with the wall-of-text template, which often aren't important to the issue. If the version number of something isn't important, it shouldn't be written there. Repetitively reading the same format issue reports written in the same way over and again is more soul destroying imho.
> 
> For me, these outweigh the pros for having them. Mostly it's because the issue is worse to act on.
> 
> There's a difference between github templates that pop up automatically, and workflows and templates people can use. I think it's useful to be able to document our processes for different issues and guide people forward.

Yeah an issue tutorial might be a friendly and happier compromise, but it sounds annoying to write compared to a lazy template that @robertpfeiffer already wrote :)

> How do they make proposals, how do they fix documentation, how do they have a conversation about a topic or experiment?

Well option 1 is - more templates! and option 2 is that there is always the option to have a blank issue (though if you have templates active it is a little less obvious you can create one) we could make the blank issues option explicit by including a freeform blank issue template that works just like the current ones.

s0lst1ce:
I agree with the idea keeping the root directory as clean as possible. However I also find GitHub (GH) templates very useful, be it for issues or PRs.
Indeed I've found that providing them allows for a more consistent format over issues/PRs which speeds up the review process.
This is especially true with projects with many users being unfamiliar with the tools.
To me the key points are the following:
- providing information on what to include, the scope, and so on (both for issues & PRs). Do note that this can be hidden in the resulting issue.
- having a consistent format
- you can provide default labels for each template 
I don't know if this outweighs the cons though.

CobyCoding (approved):
I think that is good. It will definitely help speed up fixing issue/bug reports.

illume:
- we can experiment with it, and if we don't like it remove it.
- there is a blank/empty template available for people
- we can add the files in the docs/ folder rather than in a .github folder

MyreMylar:
- Moved the template folder to docs. 
 - Added a blank issue template file to make that option more prominent.
 - Edited the description of the bug report template to make clear that it is supposed to be a guided experience to help people who aren't that comfortable/familiar with making issue reports.

illume (approved):
👍 great :)

PR #2076, opened 2020-09-04
Fix off by one error in src_c/event.c

siggisv:
The custom user event types returned by pygame.event.custom_type()
should be up to pygame.NUMEVENTS, but not include it.

 - [x] Add unittest to catch off-by-one bug in event.c
 - [x] Fix the bug.
 - [x] Find a way to prevent the unittest to mess up other tests by not leaving any custom user event types left for them.

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -1988,10 +1988,11 @@ int _custom_event = PGE_USEREVENT + 1;
 static PyObject *
 pg_event_custom_type(PyObject *self, PyObject *args)
 {
-    if (_custom_event > PG_NUMEVENTS) {
-        return RAISE(pgExc_SDLError, "pygame.event.custom_type made too many event types.");
+    if (_custom_event < PG_NUMEVENTS) {
+        return PyInt_FromLong(_custom_event++);
     }
-    return PyInt_FromLong(_custom_event++);
+    else
+        return RAISE(pgExc_SDLError, "pygame.event.custom_type made too many event types.");
 }
 
 static PyMethodDef _event_methods[] = {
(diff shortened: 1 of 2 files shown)


siggisv:
Looks like my unittest is messing other unitests by using up all the custom user event types :(

I need to find a way to fix that.

MyreMylar:
_Copying from Discord in case you miss it there:_

hmm I guess I would add it to the event modules's quit function so you can `pygame.quit()` to de-register all custom events. It should be this one in event.c: 
```
static void
_pg_repeat_cleanup(void)
{
    if (_pg_repeat_timer) {
        SDL_RemoveTimer(_pg_repeat_timer);
        _pg_repeat_timer = 0;
    }
    _pg_event_is_init = 0;
}
```
just add a `_custom_event = PGE_USEREVENT + 1;` to the end. Probably also have to move up the global definition of `_custom_event` too.
You could also rename the function to `_pg_event_cleanup` for thoroughness and add a comment to explain - it's 6:30 in the morning though so I might be speaking nonsense

siggisv:
I see that the SDL 1.2 build test on Travis is failing because I had not noticed that `_pg_repeat_cleanup()` was within a `#if IS_SDLv2` preprocessor directive when I moved the definition of `_custom_event`.

Need sleep. Will fix that tomorrow

siggisv:
@MyreMylar Your 6:30 in the morning comment was right, apart from the minor detail that I needed to make some changes to make it also work for SDL 1.2 builds.

MyreMylar (approved):
All looks good to me now, and it works locally. Approved! 👍 🎉 

Thanks for the contribution!

illume (approved):
👍 Thanks

PR #2077, opened 2020-09-05
Closes #1794. Added unittest for Surface.blit() for sizing and color.

### LINKED ISSUE #1794: Add unit tests: surface.blit()
This function needs several unit tests.

**Test stubs:** 

- [ ] https://github.com/pygame/pygame/blob/master/test/surface_test.py#L622-L652
- [ ] https://github.com/pygame/pygame/blob/master/test/surface_test.py#L667-L674
- [ ] https://github.com/pygame/pygame/blob/master/test/surface_test.py#L676-L693

**Docs:** https://www.pygame.org/docs/ref/surface.html#pygame.Surface.blit
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/surface.c#L2167-L2235

adamandrews1:
Looking forward to feedback. Let me know if I missed any cases.
I didn't include special flags testing here because I see others are treating those in different unit tests.

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -815,7 +815,7 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
             self.assertEqual(dst.get_at(pt)[1], src.get_at(pt)[1])
 
     def test_blit__blit_to_self(self):
-        """Test that blit operation works on self, and that no distortion occurs. """
+        """Test that blit operation works on self, alpha value is correct, and that no RGB distortion occurs."""
         test_surface = pygame.Surface((128, 128), SRCALPHA, 32)
         area = test_surface.get_rect()
 
@@ -826,56 +826,58 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
 
         test_surface.blit(test_surface, (0, 0))
 
-        # The below does not work, need to resolve. #TODO
-        # for x in range(area.width):
-        #     for y in range(area.height):
-        #         self.assertEqual(reference_surface.get_at((x,y)), test_surface.get_at((x,y)))
+        for x in range(area.width):
+            for y in range(area.height):
+                (r, g, b, a) = reference_color = reference_surface.get_at((x,y))
+                expected_color = (r, g, b, (a+(a*((256-a)//256))))
+                self.assertEqual(reference_color, expected_color)
 
         self.assertEqual(reference_surface.get_rect(), test_surface.get_rect())
 
 
     def test_blit__SRCALPHA_to_SRCALPHA_non_zero(self):
         """Tests blitting a nonzero alpha surface to another nonzero alpha surface
-         both premultiplied and straight alpha compositing methods."""
+         both straight alpha compositing method. Test is fuzzy (+/- 1/256) to account for
+         different implementations in SDL1 and SDL2.
+        """
 
-        (w, h) = size = (32, 32)
+        size = (32, 32)
 
         def check_color_diff(color1, color2):
-            """Returns True if two colors are within (1, 1, 1, 1)"""
+            """Returns True if two colors are within (1, 1, 1, 1) of each other."""
             for val in (color1 - color2):
                 if abs(val) > 1:
                     return False
             return True
 
         def high_a_onto_low(high, low):
+            """Tests straight alpha case. Source is low alpha, destination is high alpha"""
             high_alpha_surface = pygame.Surface(size, pygame.SRCALPHA, 32)
             low_alpha_surface = high_alpha_surface.copy()
-            high_alpha_color = Color((32, 32, 32, high))
-            low_alpha_color = Color((32, 32, 32, low))
+            high_alpha_color = Color((high, high, low, high)) # Injecting some RGB variance. Any numbers can be chosen.
+            low_alpha_color = Color((high, low, low, low))
             high_alpha_surface.fill(high_alpha_color)
             low_alpha_surface.fill(low_alpha_color)
-    
-            # Premultiplied case, src = low alpha, dst = high alpha
-            high_alpha_surface.blit(low_alpha_surface, (0, 0), special_flags=BLEND_PREMULTIPLIED)
+            
+            high_alpha_surface.blit(low_alpha_surface, (0, 0))
             
             expected_color = (low_alpha_color) + Color(tuple(((x*(255-low_alpha_color.a))//255) for x in high_alpha_color))
             self.assertTrue(check_color_diff(high_alpha_surface.get_at((0, 0)), expected_color))
         
         def low_a_onto_high(high, low):
+            """Tests straight alpha case. Source is high alpha, destination is low alpha"""
             high_alpha_surface = pygame.Surface(size, pygame.SRCALPHA, 32)
             low_alpha_surface = high_alpha_surface.copy()
-            high_alpha_color = Color((32, 32, 32, high))
-            low_alpha_color = Color((32, 32, 32, low))
+            high_alpha_color = Color((high, high, low, high)) # Injecting some RGB variance. Any numbers can be chosen.
+            low_alpha_color = Color((high, low, low, low))
             high_alpha_surface.fill(high_alpha_color)
             low_alpha_surface.fill(low_alpha_color)
     
-            # Premultiplied case, src = high alpha, dst = low alpha
-            low_alpha_surface.blit(high_alpha_surface, (0, 0), special_flags=BLEND_PREMULTIPLIED)
+            low_alpha_surface.blit(high_alpha_surface, (0, 0))
             
             expected_color = (high_alpha_color) + Color(tuple(((x*(255-high_alpha_color.a))//255) for x in low_alpha_color))
             self.assertTrue(check_color_diff(low_alpha_surface.get_at((0, 0)), expected_color))
         
-        # Test premultiplication on array of alpha values #TODO Straight Alpha Compositing
         for low_a in range(0,128):
             for high_a in range(128, 256):
                 high_a_onto_low(high_a, low_a)
@@ -1501,7 +1503,7 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
 
 
     def test_get_shifts(self):
-
+        """Tests whether Surface.get_shifts returns proper RGBA shifts under various conditions."""
         # __doc__ (as of 2008-08-02) for pygame.surface.Surface.get_shifts:
 
         # Surface.get_shifts(): return (R, G, B, A)
@@ -1514,22 +1516,22 @@ class SurfaceTypeTest(AssertRaisesRegexMixin, unittest.TestCase):
 
         # Test for SDL1 -> set_shifts() raises AttributeError in SDL2
         if SDL1:
-          surface = pygame.Surface((32, 32))
-          surface.set_shifts((2, 4, 8, 16))
-          r, g, b, a = surface.get_shifts()
-          self.assertEqual((r, g, b, a), (2, 4, 8, 16))
+            surface = pygame.Surface((32, 32))
+            surface.set_shifts((2, 4, 8, 16))
+            r, g, b, a = surface.get_shifts()
+            self.assertEqual((r, g, b, a), (2, 4, 8, 16))
         # Test for SDL2 on surfaces with various depths and alpha on/off
         else:
-          depths = [8, 24, 32]
-          alpha = 128
-          off = None
-          for bit_depth in depths:
-            surface = pygame.Surface((32, 32), depth=bit_depth)
-            surface.set_alpha(alpha)
-            r1, g1, b1, a1 = surface.get_shifts()
-            surface.set_alpha(off)
-            r2, g2, b2, a2 = surface.get_shifts()
-            self.assertEqual((r1, g1, b1, a1), (r2, g2, b2, a2))
+            depths = [8, 24, 32]
+            alpha = 128
+            off = None
+            for bit_depth in depths:
+                surface = pygame.Surface((32, 32), depth=bit_depth)
+                surface.set_alpha(alpha)
+                r1, g1, b1, a1 = surface.get_shifts()
+                surface.set_alpha(off)
+                r2, g2, b2, a2 = surface.get_shifts()
+                self.assertEqual((r1, g1, b1, a1), (r2, g2, b2, a2))
 
 
     def test_get_size(self):
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1,5 +1,4 @@
 import os
-import math
 import unittest
 from pygame.tests import test_utils
 from pygame.tests.test_utils import (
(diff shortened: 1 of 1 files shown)


adamandrews1:
Working on adding test to address alpha issue. Looks like the documentation for blit is pretty vague on what is supposed to happen here. For what is is worth, the following code:
```python3
def test_blit__SRCALPHA_to_SRCALPHA_non_zero(self):  # TODO
        # " There is no unit test for blitting a SRCALPHA source with non-zero
        #   alpha to a SRCALPHA destination with non-zero alpha " LL

        (w, h) = size = (32, 32)

        s = pygame.Surface(size, pygame.SRCALPHA, 32)
        s2 = s.copy()

        s.fill((32, 32, 32, 111))
        s2.fill((32, 32, 32, 31))

        s.blit(s2, (0, 0))

        self.assertEqual(s.get_at((0, 0)), (32, 32, 32, 31))```

Produces:
```
FAIL: test_blit__SRCALPHA_to_SRCALPHA_non_zero (__main__.SurfaceTypeTest)
  File "surface_test.py", line 786, in test_blit__SRCALPHA_to_SRCALPHA_non_zero
    self.assertEqual(s.get_at((0, 0)), (32, 32, 32, 31))
AssertionError: (31, 31, 31, 127) != (32, 32, 32, 31)
```
So the alpha here, by default, changed alpha values 31, 111 -> 127.

**Comment by MyreMylar** [comment]
Not certain this is relevant to what you are talking about, but....

There is [a long issue](https://github.com/pygame/pygame/issues/1289) about alpha compositing problems/difficulties with blit. The short summary is that the behaviour of alpha composite style blits (where _both_ surfaces have alpha) has changed between pygame 1 and pygame 2 in two ways (well really between SDL1 & SDL2):

- One makes blitting more mathematically correct in pygame 2 - but makes it look a little weirder in some cases that 'worked' in pygame 1 thanks to an SDL 1-era hack. I recently fixed up the new-ish PREMULTIPIED alpha blend mode especially for these situations to give people a usable choice and a proper fix. There is a quick primer/summary of 'straight' alpha and 'premultiplied' alpha [here](https://microsoft.github.io/Win2D/html/PremultipliedAlpha.htm).
- The second is just a minor formula tweak so that that all the numbers are often very slightly different between 1 and 2. See a briefer issue about it [here](https://github.com/pygame/pygame/issues/864).

**Comment by adamandrews1** [comment]
Unit testing for PREMULTIPLIED flag implemented, passing on my machine in ~0.8s.

Planning on implementing straightline alpha compositing unittest next. Expecting to have a tougher time given all of the issues you linked to. Let me know if you think this worth implementing the unittest at this point.

Did some investigation on blit-to-self unittest. Was thinking I was done, but I ran into an issue where apparently the blitted-to-self surface had color pixels that differed from the reference copy I created pre-blit...
Not sure why this would occur. See surface_test.py L829-832. @MyreMylar

File: test/surface_test.py [hunk 1]
```diff
@@ -1,5 +1,4 @@
 import os
-import math
```
adamandrews1, line 2: unused, caught by linter

test/surface_test.py
```diff
@@ -815,41 +814,81 @@ def test_blit__SRCALPHA_opaque_source(self):
...
-            src.set_at(pt, color)
+        reference_surface = test_surface.copy()
 
-        src.blit(src, (0, 0))
+        test_surface.blit(test_surface, (0, 0))
 
-    def todo_test_blit__SRCALPHA_to_SRCALPHA_non_zero(self):  # TODO
-        # " There is no unit test for blitting a SRCALPHA source with non-zero
-        #   alpha to a SRCALPHA destination with non-zero alpha " LL
+        # The below does not work, need to resolve. #TODO
```
adamandrews1, line 829: @MyreMylar This fails here. Not sure why.

test/surface_test.py
```diff
@@ -815,41 +814,81 @@ def test_blit__SRCALPHA_opaque_source(self):
...
+            for high_a in range(128, 256):
+                high_a_onto_low(high_a, low_a)
+                low_a_onto_high(high_a, low_a)
+                
 
     def test_blit__SRCALPHA32_to_8(self):
         # Bug: fatal
         # SDL_DisplayConvert segfaults when video is uninitialized.
         target = pygame.Surface((11, 8), 0, 8)
-        color = target.get_palette_at(2)
```
adamandrews1, line 792: This was overwriting something in another namespace -> renamed.

test/surface_test.py
```diff
@@ -815,41 +814,81 @@ def test_blit__SRCALPHA_opaque_source(self):
...
+    def test_blit__SRCALPHA_to_SRCALPHA_non_zero(self):
+        """Tests blitting a nonzero alpha surface to another nonzero alpha surface
+         both premultiplied and straight alpha compositing methods."""
 
-        s.blit(s2, (0, 0))
+        (w, h) = size = (32, 32)
 
-        # TODO:
-        # what is the correct behaviour ?? should it blend? what algorithm?
+        def check_color_diff(color1, color2):
```
adamandrews1, line 843: This is a slightly fuzzy test, if you check strict equality it is difficult to match the exact rounding behaviour of the algorithm. Does this mean it should fail?

adamandrews1 (commented):
Added context comments.

test/surface_test.py
MyreMylar: Just tested it manually and it's failing because of alpha compositing (blending a surface with non 255 alpha to another surface with non 255 alpha). 

The test should fail because you wouldn't expect a surface with alpha pixels blitted on top of another with the same to produce the same surface with the default alpha belnd mode (or most alpha blend modes). Perhaps this test should use an opaque gradient instead?

This is the utility function in test_utils:

```
def gradient(width, height):
    """

    Yields a pt and corresponding RGBA tuple, for every (width, height) combo.
    Useful for generating gradients.
    
    Actual gradient may be changed, no tests rely on specific values.
    
    Used in transform.rotate lossless tests to generate a fixture.

    """

    for l in xrange_(width):
        for t in xrange_(height):
            yield (l, t), tuple(map(rgba_between, (l, t, l, l + t)))
```

the bit that says `l + t` would have to be changed to 255 in a new `opaque_gradient()` utility function.

test/surface_test.py
MyreMylar: For 'straight alpha' I think it's fine for it to be a little bit fuzzy because the difficulty for us is that the algorithm is different for the default 'straight' alpha blend mode between SDL1 & SDL2. 

The pre-multiplied algorithm should be testable, it's just pygame code so it should be consistent between SDL1 & SDL2. If you look later in the file for the `test_blit_blend_premultiplied()` test you should see the formula (it has bitshifts!).

I wouldn't worry too much about the blend modes as most of them have their own test afaik.

test/surface_test.py
MyreMylar: 👍 thumbs

MyreMylar:
Added a few comments above.

test/surface_test.py
adamandrews1: I changed the alpha value expression to more general form. Thanks for the heads up on the non-opaque gradient.

test/surface_test.py
adamandrews1: Great - Fuzzy test for both straight alpha SDL1 & SDL2 cases is working and I will leave the premultiplied to its dedicated test. Thanks!

test/surface_test.py
```diff
@@ -1417,22 +1516,22 @@ def test_get_shifts(self):
...
+            depths = [8, 24, 32]
+            alpha = 128
+            off = None
+            for bit_depth in depths:
+                surface = pygame.Surface((32, 32), depth=bit_depth)
+                surface.set_alpha(alpha)
+                r1, g1, b1, a1 = surface.get_shifts()
+                surface.set_alpha(off)
+                r2, g2, b2, a2 = surface.get_shifts()
+                self.assertEqual((r1, g1, b1, a1), (r2, g2, b2, a2))
```
adamandrews1, line 1534: Inconsistent indentation, caught by linter.

test/surface_test.py
```diff
@@ -1417,22 +1516,22 @@ def test_get_shifts(self):
 
         # Test for SDL1 -> set_shifts() raises AttributeError in SDL2
         if SDL1:
-          surface = pygame.Surface((32, 32))
-          surface.set_shifts((2, 4, 8, 16))
-          r, g, b, a = surface.get_shifts()
-          self.assertEqual((r, g, b, a), (2, 4, 8, 16))
+            surface = pygame.Surface((32, 32))
```
adamandrews1, line 1519: inconsistent indentation

test/surface_test.py
```diff
@@ -826,56 +826,58 @@ def test_blit__blit_to_self(self):
...
         test_surface.blit(test_surface, (0, 0))
 
-        # The below does not work, need to resolve. #TODO
-        # for x in range(area.width):
-        #     for y in range(area.height):
-        #         self.assertEqual(reference_surface.get_at((x,y)), test_surface.get_at((x,y)))
+        for x in range(area.width):
+            for y in range(area.height):
+                (r, g, b, a) = reference_color = reference_surface.get_at((x,y))
+                expected_color = (r, g, b, (a+(a*((256-a)//256))))
```
adamandrews1, line 832: Corrected alpha expression

test/surface_test.py
```diff
@@ -826,56 +826,58 @@ def test_blit__blit_to_self(self):
...
                     return False
             return True
 
         def high_a_onto_low(high, low):
+            """Tests straight alpha case. Source is low alpha, destination is high alpha"""
             high_alpha_surface = pygame.Surface(size, pygame.SRCALPHA, 32)
             low_alpha_surface = high_alpha_surface.copy()
-            high_alpha_color = Color((32, 32, 32, high))
-            low_alpha_color = Color((32, 32, 32, low))
+            high_alpha_color = Color((high, high, low, high)) # Injecting some RGB variance. Any numbers can be chosen.
```
adamandrews1, line 857: RGB values were previously hard-coded. I _think_ this improves the test coverage.

test/surface_test.py
```diff
@@ -826,56 +826,58 @@ def test_blit__blit_to_self(self):
...
+            low_alpha_color = Color((high, low, low, low))
             high_alpha_surface.fill(high_alpha_color)
             low_alpha_surface.fill(low_alpha_color)
-    
-            # Premultiplied case, src = low alpha, dst = high alpha
-            high_alpha_surface.blit(low_alpha_surface, (0, 0), special_flags=BLEND_PREMULTIPLIED)
+            
+            high_alpha_surface.blit(low_alpha_surface, (0, 0))
```
adamandrews1, line 862: Default, straight alpha.

test/surface_test.py
MyreMylar: :thumbs:

test/surface_test.py
MyreMylar: 👍

test/surface_test.py
MyreMylar: sounds good 👍

test/surface_test.py
MyreMylar: 👍

illume:
Thanks! 👍 🎉

test/surface_test.py
```diff
@@ -741,8 +741,65 @@ def todo_test_blit(self):
...
+            """Resets our blank canvas destination surface"""
+            dst_surface.fill(pygame.Color(0,0,0))
+        
+        #Init
+        src_surface = pygame.Surface((256, 256), 32)
+        src_surface.fill(pygame.Color(255,255,255))
+        dst_surface = pygame.Surface((64, 64), 32)
+        reset_dest(dst_surface)
+
+        # Full coverage w/ overflow, specified with Coordinate
```
illume, line 754: How about putting all of these tests into a class with separate methods?

- makes it easier to run an individual method in development
- it's possible to see if more than one test fails

test/surface_test.py
```diff
@@ -757,41 +814,83 @@ def test_blit__SRCALPHA_opaque_source(self):
...
-        s2.fill((32, 32, 32, 31))
+        self.assertEqual(reference_surface.get_rect(), test_surface.get_rect())
 
-        s.blit(s2, (0, 0))
 
-        # TODO:
-        # what is the correct behaviour ?? should it blend? what algorithm?
+    def test_blit__SRCALPHA_to_SRCALPHA_non_zero(self):
+        """Tests blitting a nonzero alpha surface to another nonzero alpha surface
+         both straight alpha compositing method. Test is fuzzy (+/- 1/256) to account for
```
illume, line 840: I think this is a good way to work around the SDL1 -> SDL2 differences (which is considered a bug, but that's covered by another test).

test/surface_test.py
```diff
@@ -757,41 +814,83 @@ def test_blit__SRCALPHA_opaque_source(self):
...
+            high_alpha_surface = pygame.Surface(size, pygame.SRCALPHA, 32)
+            low_alpha_surface = high_alpha_surface.copy()
+            high_alpha_color = Color((high, high, low, high)) # Injecting some RGB variance. Any numbers can be chosen.
+            low_alpha_color = Color((high, low, low, low))
+            high_alpha_surface.fill(high_alpha_color)
+            low_alpha_surface.fill(low_alpha_color)
+    
+            low_alpha_surface.blit(high_alpha_surface, (0, 0))
+            
+            expected_color = (high_alpha_color) + Color(tuple(((x*(255-high_alpha_color.a))//255) for x in low_alpha_color))
```
illume, line 878: It's in general a good idea to avoid using other modules when testing as much as possible. But here `high_alpha_color.a` is more readable than `high_alpha_color[3]`.

illume (approved):
Looks good to me. It's good to merge in I think.

I left some comments about things for your consideration, but they're not necessary to do.

Will leave it open a bit longer for MyreMylar or someone else to have a look.

test/surface_test.py
adamandrews1: Went ahead and created a proper class for the blit tests as suggested. I didn't want my setUp() polluting/slowing down the following unrelated tests, so I created GeneralSurfaceTests class to pick up the following tests. I think the other tests could also be segmented somewhat like TestSurfaceBlit, but it's out of scope for now. Let me know what you think.

MyreMylar (changes requested):
Looking good - just noticed a few PEP-8 type formatting things while reading through the code. 

I'm not eagle-eyed or anything, my IDE (PyCharm) puts squiggly lines under things that are going to cause problems for code linters.

test/surface_test.py
```diff
@@ -742,7 +742,65 @@ def todo_test_blit(self):
...
+        self.assertEqual(self.dst_surface.get_at((1, 1)), (255, 255, 255))
+        self.assertEqual(self.dst_surface.get_at((63, 63)), (255, 255, 255))
+
+    def test_blit_area_contraint(self):
+        """Testing area constraint"""
+        result = self.dst_surface.blit(self.src_surface, dest=pygame.Rect((1, 1, 1, 1)),\
+            area=pygame.Rect((2, 2, 2, 2)))
+        self.assertIsInstance(result, pygame.Rect)
+        self.assertEqual(result.size, (2, 2))
+        self.assertEqual(self.dst_surface.get_at((0, 0)), (0, 0, 0)) #Corners
```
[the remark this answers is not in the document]
adamandrews1: No worries- I will take care of it.

MyreMylar (approved):
All cleaned up. Approved! 👍 

Thanks for the contribution 🎉

adamandrews1:
and thank you for the feedback and maintenance!
🎉

illume (approved):
👍

illume:
@adamandrews1 is your email in these git commits correct? It doesn't seem like it's linked up to your github account. I just wanted to make sure you get proper credit for it.

If it matters to you, you could:
- amend the commit(s) with your correct email address `git commit --amend --author "Adam Andrews < adamandrews@berkeley.edu >"` https://stackoverflow.com/questions/750172/how-to-change-the-author-and-committer-name-and-e-mail-of-multiple-commits-in-gi
- or add your other email address into github: https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/adding-an-email-address-to-your-github-account

Please let me know either way?

adamandrews1:
@illume 
I went ahead and added my other email to Github, so each commit should be good to go as-is.
Thanks for the heads-up!

PR #2080, opened 2020-09-05
Fix sprite drawing on Renderer

### LINKED ISSUE #2056: texture_sprite example broken by #1724
I noticed that my texture sprite test fails in dev10. The same error occurs with examples/texture_sprite.py. The error code is as follows:
```
  File "sdl2test.py", line 174, in <module>
    group.draw(renderer)
  File "/home/michael/.local/lib/python3.7/site-packages/pygame/sprite.py", line 477, in draw
    surface.blits((spr.image, spr.rect) for spr in sprites)
AttributeError: 'pygame._sdl2.video.Renderer' object has no attribute 'blits'
```
JoKing recommended I submit this issue even though it uses to the experimental sdl2_render api. He suggested that the error relates to #1724 and will be an easy fix. I concur.

MightyJosip:
Fix #2056

### DIFF
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -539,12 +539,16 @@ class AbstractGroup(object):
 
         """
         sprites = self.sprites()
-        self.spritedict.update(
-            zip(
-                sprites,
-                surface.blits((spr.image, spr.rect) for spr in sprites)
+        if hasattr(surface, "blits"):
+            self.spritedict.update(
+                zip(
+                    sprites,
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
Verified that this works locally and has no effect on performance of regular sprites. Tested using method from original PR:

```
python examples/testsprite.py -noupdate_rects 10000
```

FPS was the same on my PC before and after this PR (64.7 FPS). I see from discord that it solves the original problem too. Approved 👍

PR #2083, opened 2020-09-08
First pass at removing PG_ENABLE_NEWBUF

### LINKED ISSUE #1974: remove all PG_ENABLE_NEWBUF conditionals from C code
We should remove all PG_ENABLE_NEWBUF conditionals from C code, and delete the now unreachable code that handles the non-NEWBUF case. All of the supported Python versions now have PEP 3118 support. Instead, as a sanity check/defensive coding measure, we could just abort when Py_TPFLAGS_HAVE_NEWBUFFER is not defined.

adamandrews1:
This built on my local machine just fine, so I wanted to try it on Travis.

Still needs the failsafe for if Py_TPFLAGS_HAVE_NEWBUFFER is not present.

Fixes #1974 , eventually.

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -2217,7 +2217,7 @@ MODINIT_DEFINE(base)
         MODINIT_ERROR;
     }
 
-    if (PyModule_AddIntConstant(module, "HAVE_NEWBUF", PG_ENABLE_NEWBUF)) {
+    if (PyModule_AddIntConstant(module, "HAVE_NEWBUF", 1)) {
         Py_XDECREF(atexit_register);
         Py_DECREF(pgExc_BufferError);
         DECREF_MOD(module);
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -1046,8 +1046,6 @@ pgObject_GetBuffer(PyObject *obj, pg_buffer *pg_view_p, int flags)
     flags |= PyBUF_PYGAME;
 #endif
 
-#if PG_ENABLE_NEWBUF
-
     if (PyObject_CheckBuffer(obj)) {
         char *fchar_p;
 
@@ -1124,10 +1122,7 @@ pgObject_GetBuffer(PyObject *obj, pg_buffer *pg_view_p, int flags)
             return -1;
         }
         success = 1;
-    }
-
-#endif
-    if (!success && pgGetArrayStruct(obj, &cobj, &inter_p) == 0) {
+    }    if (!success && pgGetArrayStruct(obj, &cobj, &inter_p) == 0) {
         if (pgArrayStruct_AsBuffer(pg_view_p, cobj, inter_p, flags)) {
             Py_DECREF(cobj);
             return -1;
@@ -2156,13 +2151,9 @@ MODINIT_DEFINE(base)
         MODINIT_ERROR;
     }
 
-#if PG_ENABLE_NEWBUF
     pgExc_BufferError =
         PyErr_NewException("pygame.BufferError", PyExc_BufferError, NULL);
-#else
-    pgExc_BufferError =
-        PyErr_NewException("pygame.BufferError", PyExc_RuntimeError, NULL);
-#endif
+
     if (pgExc_SDLError == NULL) {
         Py_XDECREF(atexit_register);
         DECREF_MOD(module);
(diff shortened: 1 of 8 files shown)


src_c/pgcompat.h
```diff
@@ -189,7 +189,7 @@
                          start, stop, step, slicelength)
 #endif
 
-/* Support new buffer protocol? */
```
adamandrews1, line 192: Note - this file probably needs a bit more work

robertpfeiffer:
This is a good first step. You probably need to add some OLDBUF code back in to help with Python 2.7 compatibility.

adamandrews1:
Looks like Py2.7 is happy again. Last step is to add the failsafe.

robertpfeiffer:
There's a slight formatting mishap, and I think we should add a
```C
#ifndef Py_TPFLAGS_HAVE_NEWBUFFER
  #error No support for PEP 3118/Py_TPFLAGS_HAVE_NEWBUFFER
#endif 
````
somewhere, perhaps in pgcompat.h?

File: src_c/base.c [hunk 2]
```diff
@@ -1124,10 +1122,7 @@ pgObject_GetBuffer(PyObject *obj, pg_buffer *pg_view_p, int flags)
             return -1;
         }
         success = 1;
-    }
-
-#endif
-    if (!success && pgGetArrayStruct(obj, &cobj, &inter_p) == 0) {
+    }    if (!success && pgGetArrayStruct(obj, &cobj, &inter_p) == 0) {
```
robertpfeiffer, line 1125: You deleted one linebreak too many, now it looks like an else if branch.

robertpfeiffer (changes requested):

src_c/base.c
adamandrews1: Good catch - fix is in.

adamandrews1:
It looks like the old method of determining PG_ENABLE_NEWBUF:
```cpp
#if !defined(PG_ENABLE_NEWBUF)  /* allow for command line override */
#if !defined(PYPY_VERSION)
#define PG_ENABLE_NEWBUF 1
#else
#define PG_ENABLE_NEWBUF 0
#endif
#endif /* !defined(PG_ENABLE_NEWBUF)
```
suggests
```cpp
#if defined(PYPY_VERSION)
#error No support for PEP 3118/Py_TPFLAGS_HAVE_NEWBUFFER. Please use a supported Python version.
#endif
```
would be effective.
Not sure if this implies a huge change to the codebase.

When I add
```cpp
#ifndef Py_TPFLAGS_HAVE_NEWBUFFER
  #error No support for PEP 3118/Py_TPFLAGS_HAVE_NEWBUFFER
#endif 
```
to pgcompat.h, I hit the error unless I place it below the first ifndef (which would make the error line unreachable code).

Thoughts on how to proceed?

robertpfeiffer:
No, PYPY (a Python implementation that is not CPython) supports newbuffer now. That was the reason why I thought we can take this conditional out in the first place. PYPY version is not a reason to error. We can probably merge this without the check, but now I think about it, of course this produces an error, it needs to be in a file that incudes <Python.h>. My mistake. That was the wrong header file.

It should be in src_c/_pygame.h, under the commented-out check for SDL2.
 
```C
#if !defined(Py_TPFLAGS_HAVE_NEWBUFFER)
#error No support for PEP 3118/Py_TPFLAGS_HAVE_NEWBUFFER. Please use a supported Python version.
#endif
```

adamandrews1:
Makes sense. Here are the changes.
 _pygame.h seems a good a place as any for the check and is behaving properly on my machine. I wasn't sure where Py_TPFLAGS_HAVE_NEWBUFFER was to be included from, so thanks for the tip. @robertpfeiffer 

I think this is it for this issue, so I cleaned up the comments in setup.py and will promote this PR from Draft.

robertpfeiffer (approved):

MyreMylar:
This all seems to work for me and all references to ENABLE_NEWBUF have gone from the codebase.

I noticed while looking at this that there is also a PG_ENABLE_OLDBUF. If python 2.7 supports the new buffer implementation now is there any reason to maintain the OLDBUF code in bufferproxy @robertpfeiffer ? Probably we should add that as another issue as I've no doubt this was annoying and boring to do 👍 

Thankyou very much for the contribution! 🎉

MyreMylar (approved):

illume:
I added an extra task to https://github.com/pygame/pygame/issues/1974 for the PG_ENABLE_OLDBUF removal.

illume (approved):
👍 thanks :)

PR #1681, opened 2020-04-28
make set_alpha behave like in 1.9.6

robertpfeiffer:
set_alpha(255) should disable blending. get_alpha() should return 255 instead of None.

PR #2087, opened 2020-09-10
Surface.set_alpha(255) should turn off blending (1.9.6 compat) part 2

illume:
Continuing on from: https://github.com/pygame/pygame/pull/1681 in this pygame repo branch so we can collaborate more easily.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1584,6 +1584,11 @@ surf_set_alpha(pgSurfaceObject *self, PyObject *args)
     else
         alpha = (Uint8)alphaval;
 
+    if (alpha==255 && !SDL_ISPIXELFORMAT_ALPHA(surf->format->format)) {
+        if (SDL_SetSurfaceBlendMode(surf, SDL_BLENDMODE_NONE) != 0)
+            return RAISE(pgExc_SDLError, SDL_GetError());
+    }
+
     pgSurface_Prep(self);
 #if IS_SDLv1
     result = SDL_SetAlpha(surf, flags, alpha);
@@ -1627,8 +1632,8 @@ surf_get_alpha(pgSurfaceObject *self, PyObject *args)
     if (SDL_GetSurfaceBlendMode(surf, &mode) != 0)
         return RAISE(pgExc_SDLError, SDL_GetError());
 
-    if (mode != SDL_BLENDMODE_BLEND)
-        Py_RETURN_NONE;
+    if (mode == SDL_BLENDMODE_NONE)
+        return PyInt_FromLong(255);
 
     if (SDL_GetSurfaceAlphaMod(surf, &alpha) != 0)
         return RAISE(pgExc_SDLError, SDL_GetError());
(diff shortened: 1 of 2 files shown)


illume:
### 1) The Ubuntu Xenial 16.4 build with SDL 2.0.4 fails. 

I suggest we make an expected fail/skip with this old SDL <= 2.0.4 on this failing test.

### 2) SDL1 build issue
```python
  File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/pygame-2.0.0.dev11-py3.5-linux-x86_64.egg/pygame/tests/surface_test.py", line 1583, in test_set_alpha_none
    self.assertEqual(255, s.get_alpha())
AssertionError: 255 != None
```
Need to check what is different between 1.9.6 with SDL1 and this.

- Here is the 1.9.6 surface.c `set_alpha` https://github.com/pygame/pygame/blob/1.9.6/src_c/surface.c#L1479
- Does this same test fail in 1.9.6 with SDL1? It would. Here is the test showing that set_alpha(None) returns None. https://github.com/pygame/pygame/blob/1.9.6/test/surface_test.py#L1119

test/surface_test.py
```diff
@@ -1580,7 +1580,7 @@ def test_set_alpha_none(self):
         s = pygame.Surface((1, 1), SRCALPHA, 32)
         s.fill((0, 255, 0, 128))
         s.set_alpha(None)
-        self.assertEqual(None, s.get_alpha())
+        self.assertEqual(255, s.get_alpha())
```
illume, line 1583: set_alpha(None) used to make get_alpha() to return None.

Should we keep this for compatibility? I'm not sure.

If we keep the new behaviour, we should have a separate test for SDL1 or fix SDL1 to do the same?

src_c/surface.c
```diff
@@ -1627,8 +1634,8 @@ surf_get_alpha(pgSurfaceObject *self, PyObject *args)
     if (SDL_GetSurfaceBlendMode(surf, &mode) != 0)
         return RAISE(pgExc_SDLError, SDL_GetError());
 
-    if (mode != SDL_BLENDMODE_BLEND)
-        Py_RETURN_NONE;
+    if (mode == SDL_BLENDMODE_NONE)
```
illume, line 1637: Because this wasn't in 1.9.6, I think we should keep compatibility of returning None when there is no alpha.

test/surface_test.py
```diff
@@ -3133,6 +3133,46 @@ def test_overlap_check(self):
                 surf.blit(surf, (d_x, d_y), (s_x, s_y, 50, 50))
                 self.assertEqual(surf.get_at(test_posn), rectc_right)
 
+    def test_overlap_check_unrolled(self):
```
illume, line 3136: We can remove this unrolled test now?

It'd be great if we had paramatised tests. So for situations like this we can see where it fails. Unfortunately unit test doesn't have this feature until newer versions. Perhaps there is a good polyfill for old versions around however.
See subTest in python 3.4 https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests I started an issue for paramatized tests here: https://github.com/pygame/pygame/issues/2088

PR #2089, opened 2020-09-10
Commenting

CobyCoding:
I am a new programmer and i have been using pygame for a while, I added some commenting to help out new coders like me.

### DIFF
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -31,6 +31,7 @@ import os
 if os.name == 'nt':
     #pypy does not find the dlls, so we add package folder to PATH.
     pygame_dir = os.path.split(__file__)[0]
+    # Adds pygame to the enviroment variables
     os.environ['PATH'] = os.environ['PATH'] + ';' + pygame_dir
 
 # when running under X11, always set the SDL window WM_CLASS to make the
@@ -48,7 +49,9 @@ class MissingModule:
         self.info = str(exc_msg)
         self.reason = "%s: %s" % (exc_type.__name__, self.info)
         self.urgent = urgent
+        # If urgent is above 0
         if urgent:
+            # Warn the user of the missing module
             self.warn()
 
     def __getattr__(self, var):
@@ -62,11 +65,14 @@ class MissingModule:
         return 0
 
     def warn(self):
+        # If there is a missing module this function is called
         msg_type = 'import' if self.urgent else 'use'
+        # This defines the message that will be shown to the user
         message = '%s %s: %s\n(%s)' % (msg_type, self.name, self.info, self.reason)
         try:
             import warnings
             level = 4 if self.urgent else 3
+            # This warns the user of the missing module
             warnings.warn(message, RuntimeWarning, level)
         except ImportError:
             print(message)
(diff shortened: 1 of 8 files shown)


robertpfeiffer:
These comments are either obvious (and don't add anything to the code) or subtly wrong. You said you're a new coder, and that might be the problem. You don't need to explain how Python works, what an object is, and that the string "linux" means Linux. That's a given. The comments should explain the things that aren't clear from context.
Feel free to drop by the Discord if you have questions.

src_py/__init__.py
```diff
@@ -62,11 +65,14 @@ def __nonzero__(self):
         return 0
 
     def warn(self):
+        # If there is a missing module this function is called
```
illume, line 68: A doctstring for the class would be useful to explain the purpose of it.

```python
class MissingModule:
    """ Summary line goes here.

    Further elaboration on what the class does, how to use it, and things to watch out for goes here.
    """
```

Here is a good example and tutorial on how to use docstrings: https://thomas-cokelaer.info/tutorials/sphinx/docstring_python.html#template-py-source-file

src_py/__init__.py
```diff
@@ -100,61 +106,97 @@ def warn(self):
     try:
         import pygame.cdrom
     except (ImportError, IOError):
+        # This creates a missing module object
```
illume, line 109: Generally we only need to comment one time to make something clear.

src_py/__init__.py
```diff
@@ -184,6 +226,7 @@ def warn_unwanted_files():
 
     files = py_files + extension_files
 
+    # Creates a list of unwanted files
```
illume, line 229: A comment which just describes what a simple piece of code does is not very helpful for the reader.
Here is a longer explanation of why: https://realpython.com/python-comments-guide/#python-commenting-worst-practices

src_py/__init__.py
illume: Adding a docstring to the MissingModule class would make this more clear.

Also this comment could probably be improved:
```
# next, the "standard" modules
# we still allow them to be missing for stripped down pygame distributions
```

- Sentences with capital letters and full stops.
- Perhaps it could talk about how missing modules are assigned the MissingModule class instance. Because that's a bit of a weird thing to do compared to other python modules.

src_py/_camera_opencv_highgui.py
```diff
@@ -22,12 +22,12 @@ def quit():
 
 
 class Camera:
-
+    # This object will store all the infomation and functions about a certain camera
```
illume, line 25: Pretty good. You could add something like this as a docstring.

illume:
Hi @CobyCoding 

Thanks for the PR. I think you've found a lot of areas of code which could be more clear. I added a few links+comments about commenting and docstrings.

CobyCoding:
@illume Thanks so much for reviewing my PR. I will make the changes later on today.😃

CobyCoding:
@illume Thanks for all your helpful comments. As you can probably tell I'm fairly new to coding but I am very keen to learn and help out where ever I can. I really appreciate you taking the time to give me some great feedback on how I can improve my contribution to this project.

lordmauve:
Someone just linked me to this article on the different roles that code comments can play, which you might find useful: http://antirez.com/news/124

Issue #2088, opened 2020-09-10
Paramatized tests (subTest)


illume:
subTest introduced in python 3.4 https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests

> When there are very small differences among your tests, for instance some parameters, unittest allows you to distinguish them inside the body of a test method using the subTest() context manager.

This way when something fails you know which of the parameters failed with error messages. Additionally all the paramaters are tested if one param fails.

We would need a polyfil for our tests running in python 2.7.

ps. This was investigated around this test cleanup time: https://github.com/pygame/pygame/issues/496

nthykier:
According to https://stackoverflow.com/questions/54890586/subtest-like-feature-in-python-2-7, then the unittest2 module might suffice as polyfil for our needs.

illume:
My dreams of taking subTest from unittest2 were thwarted. Seems the unittest2 code is pretty hairy and 2015 old. The implementation of subTest in there is very tied to other parts of the implementation and uses lots of private members. I'd be worried about using unittest2 as a drop in replacement for unittest in case it breaks other things. I think unittest2 sort of evolved into what is used in the current python unittest.

- unittest2 "def subTest" https://hg.python.org/unittest2/file/tip/unittest2/case.py#l528
- current "def subTest" https://github.com/python/cpython/blob/master/Lib/unittest/case.py#L480
- 2.7 TestCase, https://github.com/python/cpython/blob/2.7/Lib/unittest/case.py#L131

https://stackoverflow.com/questions/32899/how-do-you-generate-dynamic-parameterized-unit-tests-in-python\

One disadvantage (or is it an advantage?) of subTest is that it doesn't do set up or tear down. You have to call them manually. Whereas paramatized tests in pytest and such do that.

It looks like there is a plugin for pytest so it can support subTest in it's runners. https://github.com/pytest-dev/pytest-subtests 

If we use a separate custom paramatize module, then that probably won't be supported well by pytest, since that includes this functionality natively.

We could do a simple implementation that just executes the context 
```python
def subTest(self, msg=None, **params):
    yield
    return
```

PR #2092, opened 2020-09-12
unittest subTest empty implementation for python2.7

illume:
python 2.7 doesn't have subTest, so we monkeypatch/polyfill an empty implementation.

It doesn't have all the features, but will just stop with a fail.
This will let us use subTest with better results on python 3. 

Issue: https://github.com/pygame/pygame/issues/2088

The first commit is black formatting. This commit 48857d91ccfadb5ea2cbaea1f1030939f43fd3b1 has the polyfill.


MyreMylar (approved):
Looks good.

Failures in a parameterised sub test now look something like this in the console output:

```
ERROR: test_scroll (__main__.GeneralSurfaceTests) (x=11, y=7)
  File "C:/Users/USERNAME/Programming/pygame/test/surface_test.py", line 1908, in test_scroll
    surf.get_at((x, y)),
IndexError: pixel index out of range
```

Which should make it a lot easier to figure out where things are going wrong when testing a large range of inputs.

👍

PR #2093, opened 2020-09-12
Joystick.c failed to compile on mac python2.7.

illume:


MyreMylar (approved):
Joystick module seems to still function fine on windows with this change (not that it should have had any effect) and it seems like an overall improvement to me. If it fixes 2.7 problems on mac too then it's definitely worth doing.

👍  Approved.

PR #2094, opened 2020-09-12
small surface.c cleanup, set_alpha, clang-formated.

illume:
- To make Surface.set_alpha more readable.
- clang-format, stop sorting include files. It's a bit dangerous.


MyreMylar (approved):
Looks neat in my IDE and set_alpha() looks a lot cleaner divided in two. All compiles and works on windows, tested with the surface tests, aliens example and some pygame_gui example programs nothing seems broken.

Approved! 👍

PR #2103, opened 2020-09-20
Fix mouse.pyi@set_system_cursor type hints

McSinyx:
This fixes the regression introduced by 3251c8305c2b74457d42e82f68e6961cb798a25c, which causes mypy to complain:

    .../site-packages/pygame/mouse.pyi:22: error: invalid syntax


MightyJosip (approved):
Thanks :+1:. This is great notice/change. I would have merged it, but since AppVeyor has some problems with downloading dependencies I will leave it for now. But it should definitely be merged.

McSinyx:
Thanks for the prompt review!  It seems that there was some 3rd party server outage at the time.  Let me try to trigger a rerun.

MightyJosip:
Thanks :+1:

PR #2104, opened 2020-09-20
Add mypy support, fix pygame._sdl2 importing touch module type hints

McSinyx:
As discussed in https://github.com/pygame/pygame/issues/839#issuecomment-687723556, mypy relies on py.typed if stubs are bundled together with modules.  Since it's an empty file, I think it'd be OK to include it in the stub source folder instead of using a tempfile.NamedTemporaryFile which will requires cleanup.

### DIFF
--- a/buildconfig/pygame-stubs/_sdl2/__init__.pyi
+++ b/buildconfig/pygame-stubs/_sdl2/__init__.pyi
@@ -1,1 +1,1 @@
-from .touch import *
+
--- a/setup.py
+++ b/setup.py
@@ -380,6 +380,7 @@ data_files = [('pygame', pygame_data_files)]
 add_stubs = True
 # add *.pyi files into distribution directory
 if add_stubs:
+    pygame_data_files.append(os.path.join('buildconfig', 'pygame-stubs', 'py.typed'))
     type_files = glob.glob(os.path.join('buildconfig', 'pygame-stubs', '*.pyi'))
     for type_file in type_files:
         pygame_data_files.append(type_file)


MightyJosip (commented):
Hi. Thanks a lot for your first contribution to pygame 🎉. So far I have tested pygame code by adding stubs to typeshed, but having this would be great QoL change. Even though AppVayor fails you can ignore it, every now and then something crashes, but that is not connected to your changes. We will soon rerun that code again on AppVeyor and than it should work again. I will write just a few more notes:
1. There is also _sdl2 folder with a few documented functions. I don't know is py.typed check for .pyi files in sub directories but can you check the following code with your changes/mypy:
```
import pygame._sdl2.touch
pygame.display.init()
pygame._sdl2.touch.get_finger(1, 2)
```
Mypy shouldn't say there is any error, if it does than we will find a way to modify it a bit.
2. When you are already changing typehints can you also remove everything from [this file](https://github.com/pygame/pygame/blob/master/buildconfig/pygame-stubs/_sdl2/__init__.pyi). It should be empty (because right now it returnes false positives in some cases), but don't remove that file since there are still some undocumented experimental things so it will probably be changed in the future.

Thanks once again.

McSinyx:
Thank you for the prompt response.  Per [PEP 561](https://www.python.org/dev/peps/pep-0561/#packaging-type-information),

> Package maintainers who wish to support type checking of their code MUST add a marker file named `py.typed` to their package supporting typing. This marker applies recursively: if a top-level package includes it, all its sub-packages MUST support type checking as well.

and mypy is happy with your snippet added.  I'll clean `_sdl/__init__.py` and retrigger the CIs.

MightyJosip:
Since SDL 2.0 they introduced Android/iOS support so there are some new functions like those in touch module. With some code changes right now you should be able to port (most) games to Android with p4a.

Great change. Thanks 👍

MightyJosip (approved):

McSinyx:
Thanks for the quick merge! I was thinking about porting to games to GNU/Linux on mobile like postmarketOS or Mobian though :smile:

PR #2105, opened 2020-09-20
Clean up README and source distribution

McSinyx:
ab6bd827ec32a53065ee2ef55482e1be3babba53 is mostly whitespace changes, so if one is reviewing it on GitHub, perse may toggle the *Hide whitespace changes* for a cleaner diff.


setup.py
```diff
@@ -669,13 +654,6 @@ def run(self):
         return install_data.run(self)
 
 
-@add_command('sdist')
```
illume, line 672: I'd prefer MANIFEST.in to stay in the buildconfig/ folder.

McSinyx: I'm OK with reverting this and fixes the actual bug in the function, but my reasons against this are (as stated in c3c6abce0c07a99b6e2a3231952ccac3573f14e1):

1. The others files in the project's root directory are only README and setuptools' configuration (slightly unrelated: setup.cfg happens to store linters' config, which IMHO should be moved to e.g. tox.ini), which makes sense for MANIFEST.in to be in the same dir.
2. Somehow (I can't find this documented) with this specified, setuptools will write MANIFEST to the root directory.

illume: We want to try and make it as easy as possible for people to contribute. Things like MANIFEST.in aren't useful for 95% of changes people make to pygame, for that reason it's best to hide it away in the buildconfig/ folder. If there was someway the MANIFEST.in could be made to not be written, that would be good. But I guess it should be moved into a separate PR?

The reason for the setup.cfg file is that all python tools can be configured in there. Including pylint, mypy, and others we want to make more use of.

README.rst
```diff
@@ -1,207 +1,192 @@
...
-
-
-     Marcus Von Appen - many changes, and fixes, 1.7.1+ freebsd maintainer.
-
-     Lenard Lindstrom - the 1.8+ windows maintainer, many changes, and fixes.
-
-     Brian Fisher - for svn auto builder, bug tracker and many contributions.
+------------
 
-     Rene Dudfield - many changes, and fixes, 1.7+ release manager/maintainer.
```
illume, line 67: How about we move the credits into a docs/CREDITS.rst file, and link from here?

README.rst
```diff
@@ -1,207 +1,192 @@
+pygame
```
illume, line 1: What do you think about removing this header? It takes up quite a lot of space, and most people will know this is about pygame.

README.rst
```diff
@@ -1,207 +1,192 @@
...
+======
 
+|TravisBuild| |AppVeyorBuild| |LaunchpadBuild|
+|PyPiVersion| |PyPiLicense| |Python2| |Python3| |GithubCommits|
+|LGTMAlerts| |LGTMGradePython| |LGTMGradeC| |Coverity|
 
-About pygame
-============
+This library is cross-platform and designed to make it easy to
```
illume, line 8: How about "The pygame library is a cross-platform"

In the last years we try to use "pygame" everywhere in the last couple of years to try and reinforce the point that lower case matters in python. We also have a better introduction that maybe can be used here... https://github.com/pygame/pygameweb/blob/master/pygameweb/templates/news/welcome.html It includes various links.

README.rst
```diff
@@ -1,207 +1,192 @@
...
+* SDL_tff >= 2.0.11
+* SDL_gfx (optional, vendored in)
+* NumPy >= 1.6.2 (optional)
 
 
 Contribute
-==========
+----------
 
-     https://www.pygame.org/contribute.html
```
illume, line 154: Maybe numpy could be (optional, but recommended).

We also have some other dependencies that could be added. Like portmidi, and freetype. 

The dependencies have dependencies (like SDL_image uses libpng, libjpeg and others)... I wonder if they are worth listing here(as an indented sub list?).

Another improvement here would be to make these links.

examples/README.rst
```diff
@@ -1,116 +1,145 @@
+These examples should help get you started with pygame.  Here is a
```
illume, line 1: Nice cleanups.

setup.py
```diff
@@ -409,19 +406,7 @@ def run_install_headers(self):
 
 #examples
 add_datafiles(data_files, 'pygame/examples',
-              ['examples',
-                  ['readme.rst',
-                   ['data',
-                       ['*']],
-                   ['macosx',
```
illume, line 416: The macosx ones are gone, cool.

setup.py
```diff
@@ -6,10 +6,8 @@
 # To configure, compile, install, just run this script.
 #     python setup.py install
 
-DESCRIPTION = """Pygame is a Python wrapper module for the
-SDL multimedia library. It contains python functions and classes
-that will allow you to use SDL's support for playing cdroms,
-audio and video output, and keyboard, mouse and joystick input."""
+with open('README.rst') as readme:
```
illume, line 9: I think this will fix it for unicode?
```
 with open('README.rst', encoding='utf-8') as readme:
```

Does our README.rst validate with the pypi rst I wonder? I noticed twine now has a checker, we could add to our CI( `twine check dist/*`). https://packaging.python.org/guides/making-a-pypi-friendly-readme/

setup.py
```diff
@@ -21,7 +19,7 @@
     "author":           "A community project.",
     "author_email":     "pygame@pygame.org",
     "description":      "Python Game Development",
-    "long_description": DESCRIPTION,
+    "long_description": LONG_DESCRIPTION,
```
illume, line 22: Great, this will make our pypi project page much better :)

illume (changes requested):
Thanks. I left some notes.

The only change required before we merge (IMHO) would be to revert the MANIFEST.in change.

setup.py
McSinyx: Thanks, I'll address the generation of `MANIFEST` later.

README.rst
McSinyx: I suppose that we can list them in the linked wiki page.  I don't really have an opinion on this, all I did was reformatting so that it looks better rendered :smile:

setup.py
McSinyx: Thanks for the unicode reminder, I keep forget that UTF-8 is not the default on all platforms.  I'll check for PyPI compatibility as I address this.

README.rst
McSinyx: That sounds like a nice idea, but I'd prefer to do this in another PR once I have better understanding on how pygame docs are built.

README.rst
McSinyx: Personally I like the way it looks, especially to compliment the badges to be displayed right below.  Not a strong opinion though, just that I'm used to see many READMEs looking that way.

README.rst
McSinyx: I rephrased this as

> pygame is a Free and Open Source cross-platform library for the development of multimedia applications like video games on Python. It uses the Simple DirectMedia Layer library and several other popular libraries to abstract most common functions and makes writing these program a more intuitive task.

Does that sound better?

McSinyx:
Yay it's green now!  BTW would a PR cleaning up `setup.py` be welcomed?  I'd like to organize the imports and possibly enforces 79-column limit and newline usage.

illume:
Thanks @McSinyx :) Nice changes.

Yeah, setup.py cleanups are very welcome :)

We're trying to move towards having python files formatted with [black](https://github.com/psf/black), and be [pylint](https://www.pylint.org/) clean.

illume (approved):
👍

McSinyx:
Thanks for the merge!

> We're trying to move towards having python files formatted with black, and be pylint clean.

May I suggest [flake8](https://flake8.pycqa.org) instead?  The reasoning is that it's less invasive than black and more modular than pylint.  Also I think many places would span too long after black is applied, e.g.

https://github.com/pygame/pygame/blob/c70d2baf4829ce12f50ef1cef24ef8d3b1f96660/setup.py#L391-L398

I'll do this with setup.py first for a preview of how it'll look with PEP 8-compatible flags in flake8 enabled and you (pyagme maintainers) will decide if it's esthetically pleasing enough.

If autoformat is ever needed though, I'd suggest [autopep8](https://pypi.org/project/autopep8).

PR #2111, opened 2020-09-21
Added unit test for surface.get_masks()

### LINKED ISSUE #1806: Add unit test: surface.get_masks()

**Test stub:** 
https://github.com/pygame/pygame/blob/4562396964cca81bceb2d12d3f09aa96746cb3fa/test/surface_test.py#L1257-L1267
**Docs:** https://www.pygame.org/docs/ref/surface.html#pygame.Surface.get_masks
**Code to be tested:** 
https://github.com/pygame/pygame/blob/4562396964cca81bceb2d12d3f09aa96746cb3fa/src_c/surface.c#L2720-L2729

seenemikk:
Closes #1806 
Please let me know if I've done something wrong or there is anything I can improve.

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1441,17 +1441,46 @@ class GeneralSurfaceTests(AssertRaisesRegexMixin, unittest.TestCase):
         finally:
             pygame.display.quit()
 
-    def todo_test_get_masks(self):
+    def test_get_masks__rgba(self):
+        """
+        Ensure that get_mask can return RGBA mask.
+        """
+        masks = [
+            (0x0f00, 0x00f0, 0x000f, 0xf000),
+            (0x00ff0000, 0x0000ff00, 0x000000ff, 0xff000000)
+        ]
+        depths = [16, 32]
+        for expected, depth in list(zip(masks, depths)):
+            surface = pygame.Surface((10, 10), pygame.SRCALPHA, depth)
+            self.assertEqual(expected, surface.get_masks())
 
-        # __doc__ (as of 2008-08-02) for pygame.surface.Surface.get_masks:
+    def test_get_masks__rgb(self):
+        """
+        Ensure that get_mask can return RGB mask.
+        """
+        masks = [
+            (0x60, 0x1c, 0x03, 0x00),
+            (0xf00, 0x0f0, 0x00f, 0x000),
+            (0x7c00, 0x03e0, 0x001f, 0x0000),
+            (0xf800, 0x07e0, 0x001f, 0x0000),
+            (0xff0000, 0x00ff00, 0x0000ff, 0x000000),
+            (0xff0000, 0x00ff00, 0x0000ff, 0x000000)
+        ]
+        depths = [8, 12, 15, 16, 24, 32]
+        for expected, depth in list(zip(masks, depths)):
+            surface = pygame.Surface((10, 10), 0, depth)
+            if depth == 8 and pygame.get_sdl_version()[0] == 2:
+                expected = (0x00, 0x00, 0x00, 0x00)
+            self.assertEqual(expected, surface.get_masks())
 
-        # Surface.get_masks(): return (R, G, B, A)
-        # the bitmasks needed to convert between a color and a mapped integer
-        #
-        # Returns the bitmasks used to isolate each color in a mapped integer.
-        # This value is not needed for normal Pygame usage.
-
-        self.fail()
+    def test_get_masks__no_surface(self):
+        """
+        Ensure that after display.quit, calling get_masks raises pygame.error.
+        """
+        with self.assertRaises(pygame.error):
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
Seems very solid, good coverage, runs on windows and passes my linter too.

Approved! 🎉 👍 Thank you for the contribution.

test/surface_test.py
```diff
@@ -1441,17 +1441,46 @@ def test_get_losses(self):
...
+            (0x7c00, 0x03e0, 0x001f, 0x0000),
+            (0xf800, 0x07e0, 0x001f, 0x0000),
+            (0xff0000, 0x00ff00, 0x0000ff, 0x000000),
+            (0xff0000, 0x00ff00, 0x0000ff, 0x000000)
+        ]
+        depths = [8, 12, 15, 16, 24, 32]
+        for expected, depth in list(zip(masks, depths)):
+            surface = pygame.Surface((10, 10), 0, depth)
+            if depth == 8 and pygame.get_sdl_version()[0] == 2:
+                expected = (0x00, 0x00, 0x00, 0x00)
```
illume, line 1473: The masks for 8 bit on SDL2 are interesting.

seenemikk: I was more wondering about why SDL1 doesn't have all zero masks. Because when bit depth is 8, surface uses a palette. And [SDL_PixelFormat](https://wiki.libsdl.org/SDL_PixelFormat) says that `A pixel format has either a palette or masks. If a palette is used Rmask, Gmask, Bmask, and Amask will be 0`.

illume: Seems like we are specifying these masks for SDL2 here:
https://github.com/pygame/pygame/blob/4d044213fa9a1185cd1af4758b0007cc17c11af2/src_c/surface.c#L662-L670

Using git blame (and ignoring a clang formatting) I find those lines were set in an initial commit from llindstrom 
```bash
git blame --ignore-rev 15051b4f1 src_c/surface.c
git log -p d7305c610
```
Unfortunately there's not any commit history for that one change.


I looked into the sdl 2 source code SDL_MasksToPixelFormatEnum where it has two cases for 8 bit.
```C
    case 8:
        if (Rmask == 0) {
            return SDL_PIXELFORMAT_INDEX8;
        }
        if (Rmask == 0xE0 &&
            Gmask == 0x1C &&
            Bmask == 0x03 &&
            Amask == 0x00) {
            return SDL_PIXELFORMAT_RGB332;
        }
        break;
```
It seems we have two cases for 8 bit surfaces and masks.
- SDL_PIXELFORMAT_INDEX8, 8bit surfaces with a pallet (the colors are indexed from there).
- SDL_PIXELFORMAT_RGB332, and when the colors are packed in there (r 3bits, g 3bits, b 2 bits)

In these cases the masks will be different for a surface.

I think we have to:
- [ ] consider what bugs in pygame there are because of this.
- [ ] update the test to support these two cases for 8 bit surfaces.

illume: It seems pygame is mostly assuming 8 bit surfaces have a palette. So the current behaviour is ok.

The test could add a special case in SDL2 for the special SDL_PIXELFORMAT_RGB332 masks, or a note about it.
```c
if (Rmask == 0xE0 &&
            Gmask == 0x1C &&
            Bmask == 0x03 &&
            Amask == 0x00) {
```

However, that could be done later. I think this PR is ok to merge :)

illume (approved):
Thanks :)

Let me know if you're ok to merge this?

test/surface_test.py
seenemikk: So am I understanding it correctly, that SDL1 uses SDL_PIXELFORMAT_RGB332 and SDL2 uses SDL_PIXELFORMAT_INDEX8? If I am, shouldn't the SDL1 Rmask in surface.c be `0xFF >> 5 << 5`?

seenemikk:
Yes I am. Thanks!

test/surface_test.py
illume: In SDL1 I think they are always supposed to be palette surfaces. It's ancient history that I'm not aware of the reason for those mask values. I guess setting the masks is allowed if people wanted to use the surfaces as some sort of packed format anyway in some situations.

Why do you think the SDL1 Rmask in surface.c be 0xFF >> 5 << 5?

seenemikk: Because in SDL1 the Rmask is set to 0x60, but in SDL_MasksToPixelFormatEnum it check if Rmask is equal to 0xe0.

illume: Yes, it seems the mask values used by pygame+SDL1 here are different to SDL_PIXELFORMAT_RGB332.

Future work could be to support SDL_PIXELFORMAT_RGB332 surfaces better in pygame, rather than just SDL_PIXELFORMAT_INDEX8 ones.

PR #2112, opened 2020-09-22
Python3 needs __bool__ instead of __nonzero__

Starbuck5:
I noticed that bool(MissingModule) == True, and it should equal False.

I looked through src_py for more __nonzero__, which I found in the sprite module.

I set up __bool__ to link to __nonzero__ so the code with it will work on Python 2 and 3.

### DIFF
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -59,7 +59,9 @@ class MissingModule:
         raise NotImplementedError(missing_msg)
 
     def __nonzero__(self):
-        return 0
+        return False
+
+    __bool__ = __nonzero__
 
     def warn(self):
         msg_type = 'import' if self.urgent else 'use'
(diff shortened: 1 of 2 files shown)


McSinyx:
Hello there, I think it would be nicer if you add the tests for the cases where previously incorrect as well!  (I'm not a maintainer, just someone happens to be watching the repo :smile:)

Starbuck5:
Thank you for the suggestion, it helped me figure more stuff out.

I tried writing a test case for the Sprite module bool() functionality and it worked without my changes, because when bool() can't find __bool__ in py3 or __nonzero__ in py2, it looks for __len__, which the sprite groups have defined.

I stand by my adding __bool__ though, because explicit is better than implicit.

MissingModule actually needs it, the Sprite groups worked fine without it, but the clarity is good for it.

I don't think I need to fully write out the test case now that I've explored it, because I'm not ending up fixing a bug with the Sprites, and MissingModule (the class) isn't something you can currently use outside of pygame (`pygame.MissingModule` raises an AttributeError)

MyreMylar (approved):
This seems like a simple improvement to me :thumbs:

PR #2114, opened 2020-09-24
Small addition to cursors docs

ankith26:
A small update to the cursor docs. 

I noticed that the source code for cursors module contain the textmarker_strings variable, but there is no reference to this variable in the docs. I don’t know wether this is intentional or not, if it is then please ignore this PR. 

I felt it needed to be in the docs, so I have put this PR. Please let me know if I did something wrong while submitting this PR, as it’s my first time contributing.

### DIFF
--- a/docs/reST/ref/cursors.rst
+++ b/docs/reST/ref/cursors.rst
@@ -55,6 +55,8 @@ The following strings can be converted into cursor bitmaps with
    * ``pygame.cursors.sizer_y_strings``
 
    * ``pygame.cursors.sizer_xy_strings``
+   
+   * ``pygame.cursor.textmarker_strings``
 
 .. function:: compile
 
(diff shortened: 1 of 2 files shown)


Starbuck5 (approved):
Thanks for contributing to pygame. This is a good simple addition. 

The cursor docs could be further improved though. To set the mouse to one of those cursors, you also need to have the size of the cursor, but that isn't documented anywhere.

Also, `pygame.cursor.compile` and `pygame.mouse.set_cursor` both reference each other, but they don't link to each other. This could be achieved using the `:func:` tag in sphinx, [like here in the display module](https://www.pygame.org/docs/ref/display.html#pygame.display.get_num_displays)

If you want to add fixes for those problems to this PR, that would be welcome.

ankith26:
My second commit updates on the cursors docs.

ankith26:
The above two commit are for improving the mouse docs. I have put in some useful information in the docs. Please do verify that I have done things correctly, as I am kind of new to all this.

illume (approved):
More good improvements :) Thanks again.

PR #2121, opened 2020-09-27
Change to support message describing how to disable it.

xcodz-dot:
required because sometimes user do not know about that and many cli applications use pygame libraries but for sound and other things and due to support message there cli does not look good.


MyreMylar (changes requested):
My feeling is that this would be better in the pygame docs somewhere, probably on this page:

https://www.pygame.org/docs/ref/pygame.html

Perhaps as a new paragraph in the module introduction where it talks about importing. Increasing the size of the support message for people who don't like it in the first place seems slightly counter-intuitive.

Also this change misspells 'variable'

xcodz-dot:
well then

PR #2123, opened 2020-09-27
Update pygame.rst

ankith26:
Another small update to the docs.

### DIFF
--- a/docs/reST/ref/pygame.rst
+++ b/docs/reST/ref/pygame.rst
@@ -251,6 +251,14 @@ check which version of pygame has been imported.
    package was built. If the identifier ends with a plus sign '+' then the
    package contains uncommitted changes. Please include this revision number
    in bug reports, especially for non-release pygame builds.
+   
+   Important note: pygame development has moved to github, this variable is 
+   obsolete now. As soon as development shifted to github, this variable started
+   returning an empty string ``""``. 
+   It has always been returning an empty string since ``v1.9.5``.
+   
+   .. versionchanged:: 1.9.5
+      Always returns an empty string ``""``.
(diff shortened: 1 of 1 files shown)


docs/reST/ref/pygame.rst
```diff
@@ -251,6 +251,14 @@ check which version of pygame has been imported.
...
    package contains uncommitted changes. Please include this revision number
    in bug reports, especially for non-release pygame builds.
+   
+   Important note: pygame development has moved to github, this variable is 
+   obsolete now. As soon as development shifted to github, this variable started
+   returning an empty string ``""``. 
+   It has always been returning an empty string since ``v1.9.5``.
+   
+   .. versionchanged:: 1.9.5
+      Always returns an empty string ``""``.
```
McSinyx, line 261: I suggest rephrasing the paragraph since the three sentences deliver very little information, removing the it in favor of the following:

```suggestion
      Always returns an empty string ``""`` since pygame switched to git for version control.
```

Not within the scope of this PR, but I'm wondering if it's possible to have similar behavior but for git commit.

ankith26: I guess we don’t need to specify twice that pygame is using github for Dev, it’s already mentioned in the line above right. However, I guess you could add a similar function for use with git, but that’s something for the devs to decide, not me.

MyreMylar (approved):
This seems like a good change to me for now - though medium term we should probably just scrap the rev field in the version module since it does nothing. 

I also think the versioning system we have now, combined with reasonably regular releases is fine grained enough that we usually have a good idea what's going on with reported issues, version wise.

Issue #1329, opened 2019-09-23
SDL2: Needs testing on mouse with two wheels


illume:
- [x] test a 2 scroller mouse what events are produced.
- [x] decide on backwards compatibiliity.
- [x] decide on forwards compatibility.
- [x] it would be nice if get_pressed returns a list of all buttons available on the mouse.
- [x] confirm test results on mouse with two mouse wheels.

> This very small issue only affected the debug features of my current project. I have the two 'extra' buttons on my Logitech m570 trackball/mouse set to teleport the player(or an enemy) to the cursor location for testing line-of-sight and ai features, and to help me move around easier while testing.

> A recent issue report mentioned keyboard codes being different in pygame2. The same applies to mouse buttons beyond 3 (or 5 considering the wheel). The extra buttons on my mouse were reported 8 and 9 on sdl1.2 but are reported 6 and 7 on sdl2. FYI there are only two 'extra' buttons, so the 8 and 9 codes returned from sdl1.2 were erroneous.

Originally reported by @mcpalmer1980 in here: https://github.com/pygame/pygame/issues/1128

**Related Docs**: https://www.pygame.org/docs/ref/mouse.html#pygame.mouse.get_pressed

mcpalmer1980:
Sorry I hid this report. I thought it was related to the other issue and would be better posted there. Yet it took over two months for someone to notice it.

Just wondering. Any estimate for a 2.0 release? Is there somewhere on GitHub to find such info?

Thanks to everyone for their hard work
Michael

mcpalmer1980:
Sorry I hid this report. I thought it was related to the other issue and would be better posted there. Yet it took over two months for someone to notice it.

Just wondering. Any estimate for a 2.0 release? Is there somewhere on GitHub to find such info?

Thanks to everyone for their hard work
Michael

illume:
@mcpalmer1980 No worries, I've been trying to go through the issues to prioritise ones for 'finishing' pygame 2 with the 'critical' label, and clean them up so it's easier for people to work on them.  There's a '[2](https://github.com/pygame/pygame/milestone/4)' milestone with issues linked. Already some people are shipping games with pygame 2, but there's obviously at least all those problems remaining in the issue lists. If your apps aren't affected by those issues, then using the pygame==2.0.0.dev3 is probably better (and dev4, and future ones). Bugs are hard to estimate, and estimating volunteer contributions is also hard to do... so I don't have an good estimate for you.

[...]

PR #2127, opened 2020-09-30
Adds a num_buttons parameter to mouse.get_pressed()

MyreMylar:
This is for smooth backwards compatibility with pygame 1 the parameter defaults to 3, but can also be set to 5 to support five button mice. Previously we were breaking compatibility by always returning five buttons.

Related to:
https://github.com/pygame/pygame/issues/1329


ankith26:
This change is nice, it will make the function backwards compatible.

illume (approved):
👍

illume:
Thinking about this from a game writers perspective... how do I use this function so it will work with different versions of pygame (at least pygame 1.9.6 and pygame 2)? Using `mouse.get_pressed()` with an argument will not raise an error on `pygame < 2.0.0.dev13`, so we are safe to pass in 5.

--

To use pygame.mouse.get_pressed(5) so it is compatible with old versions of pygame you can use some code like this:
```python
buttons = pygame.mouse.get_pressed(5)
buttons5 = buttons[:3] + (buttons[-2:] if len(buttons) == 5 else (False, False))
```
For mice without 5 buttons they will always be False.

illume:
If SDL suddenly supports 6 button mice (strangely they do exist)... then having the type error will give users who want to support that a bit more trouble in the future. Can we remove the type error with an incorrect count?

Probably a `dict[button_num]-> pressed` would have been a better return type... but we have what we have.

MyreMylar:
I guess the ValueError might give a few people problems but can they not just use `try: except:` blocks? For example in a hypothetical 6 button future, where they need to support 1+ versions of pygame something like:


```
try:
    buttons = pygame.get_pressed(num_buttons=6)
    if len(buttons != 6):  # Need this if you want to support pygame 1
         raise ValueError
except ValueError:
    try:
        buttons = pygame.get_pressed(num_buttons=5)
        if len(buttons != 5):  # Need this if you want to support pygame 1
            raise ValueError
    except:
        buttons = pygame.get_pressed()
```

And if you only want to support pygame 2+ with 6 or 5 button mice it would just be:

```
try:
    buttons = pygame.get_pressed(num_buttons=6)
except ValueError:
    buttons = pygame.get_pressed(num_buttons=5)
      
```
In which you could pad the tuple or do whatever else you need as necessary. The pygame 1+ version is a bit messy, but once you drop pygame 1 support it cleans up fairly nicely.

I feel like we should probably give clear feedback on what amount of buttons are _currently_ supported for most users and let the small minority of pygame library makers (people like me basically) who are going to manage different mouse button configurations over different versions of pygame figure that out.

I'm mostly thinking of people right now trying:

```
buttons = pygame.get_pressed(num_buttons=6)
```

With no ValueError it will just silently carry on and they might miss that their sixth button isn't getting recorded until some later point in the code.

PR #2133, opened 2020-10-02
Fix issues with pygame.font.Font

### LINKED ISSUE #206: pygame.font.SysFont(None, 20) crashes pyinstaller generated .exe
Creating a SysFont object with name = `None` causes pyinstaller generated .exes to crash with no traceback for explanation instead reporting that "xxxxx.exe has stopped working".

Question: `pygame.font.SysFont(None, 20)` should select the default pygame font (or `'Sans'`)?

Docs: https://www.pygame.org/docs/ref/font.html#pygame.font.SysFont
Code: https://github.com/pygame/pygame/blob/master/lib/sysfont.py
Test: https://github.com/pygame/pygame/blob/master/test/sysfont_test.py

```python
import pygame
import sys

pygame.init()
windowSurface = pygame.display.set_mode( (450, 25), 0, 32 )
pygame.display.set_caption('Breaking pygame')

x = None  # broken_pygame.exe
#x = 'Sans'  # non-broken_pygame.exe

print "This is the line that breaks the .exe's"
basicFont = pygame.font.SysFont(x, 20)
print 'See, I never make it here when that is None'

text = basicFont.render("I can crash pyinstaller .exe's that use pygame!" , True, (255, 0, 0))
textRect = text.get_rect()
windowSurface.blit (text, textRect)
pygame.display.update()

while True:
		if event.type == pygame.QUIT:
			pygame.quit()
			sys.exit()
```

Using a valid name (ie. 'Sans') does not crash pyinstaller generated .exe files.

see: http://www.pyinstaller.org/export/develop/project/doc/Manual.html#adapting-to-being-frozen

Originally reported by: **Ben Schroeder (Bitbucket: [ochawkeye](https://bitbucket.org/ochawkeye), GitHub: [ochawkeye](https://github.com/ochawkeye))**

- Bitbucket: https://bitbucket.org/pygame/pygame/issue/206

ankith26:
Fixes #206

### DIFF
--- a/src_py/__pyinstaller/hook-pygame.py
+++ b/src_py/__pyinstaller/hook-pygame.py
@@ -1,6 +1,9 @@
 """
 binaries hook for pygame seems to be required for pygame 2.0 Windows.
 Otherwise some essential DLLs will not be transfered to the exe.
+
+And also put hooks for datas, resources that pygame uses, to work 
+correctly with pyinstaller
 """
 
 import os
@@ -8,23 +11,27 @@ import platform
 
 from pygame import __file__ as pygame_main_file
 
-# Any resources that pygame depends upon must be put into this list,
-# so that pyinstaller includes it while running
-pygame_res = ["freesansbold.ttf", "pygame.ico", "pygame_icon.bmp",
-              "pygame_icon.icns", "pygame_icon.svg", "pygame_icon.tiff"]
-
 # Get pygame's folder
 pygame_folder = os.path.dirname(os.path.abspath(pygame_main_file))
 
 # datas is the variable that pyinstaller looks for while processing hooks
 datas = []
-for res in pygame_res:
-    res_path = os.path.join(pygame_folder, res)
+
+# A helper to append the relative path of a resource to hook variable - datas
+def _append_to_datas(file_path):
+    global datas
+    res_path = os.path.join(pygame_folder, file_path)
     if os.path.exists(res_path):
         datas.append((res_path, "pygame"))
 
-if platform.system() == "Windows":
-    
+# First append the font file, then based on the OS, append pygame icon file
+_append_to_datas("freesansbold.ttf")
+if platform.system() == "Darwin":
+    _append_to_datas("pygame_icon.tiff")
+else:
+    _append_to_datas("pygame_icon.bmp")
+
+if platform.system() == "Windows": 
     from PyInstaller.utils.hooks import collect_dynamic_libs
 
     pre_binaries = collect_dynamic_libs('pygame')
--- a/src_py/__pyinstaller/hook-pygame.py
+++ b/src_py/__pyinstaller/hook-pygame.py
@@ -3,7 +3,25 @@ binaries hook for pygame seems to be required for pygame 2.0 Windows.
 Otherwise some essential DLLs will not be transfered to the exe.
 """
 
+import os
 import platform
+
+from pygame import __file__ as pygame_main_file
+
+# Any resources that pygame depends upon must be put into this list,
+# so that pyinstaller includes it while running
+pygame_res = ["freesansbold.ttf", "pygame.ico", "pygame_icon.bmp",
+              "pygame_icon.icns", "pygame_icon.svg", "pygame_icon.tiff"]
+
+# Get pygame's folder
+pygame_folder = os.path.dirname(os.path.abspath(pygame_main_file))
+
+# datas is the variable that pyinstaller looks for while processing hooks
+datas = []
+for res in pygame_res:
+    res_path = os.path.join(pygame_folder, res)
+    if os.path.exists(res_path):
+        datas.append((res_path, "pygame"))
 
 if platform.system() == "Windows":
     
--- a/src_py/pkgdata.py
+++ b/src_py/pkgdata.py
@@ -59,8 +59,13 @@ def getResource(identifier, pkgname=__name__):
     rather than use it as a file-like object.  For example, you may
     be handing data off to a C API.
     """
-    if resource_exists(pkgname, identifier):
-        return resource_stream(pkgname, identifier)
+    
+    # When pyinstaller (or similar tools) are used, resource_exists may raise NotImplemented error
+    try:
+        if resource_exists(pkgname, identifier):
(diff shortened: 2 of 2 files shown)


ankith26:
Essentially, I have included pyinstaller hooks to load resource files used by pygame into pyinstaller generated exe. This not only fixes the default font not found issue (#206) but also ensures that the default pygame icon is also available on pyinstaller exe.

MyreMylar:
One potential issue I noticed in a quick scan through is that you are using `__file__` which is somewhat falling out of favour recently because it might not always be defined by the import system. See:

https://docs.python.org/3/reference/import.html#__file__
https://www.reddit.com/r/Python/comments/gjyrre/is_using_file_bad_practice/
https://pyoxidizer.readthedocs.io/en/v0.5.0/packaging_pitfalls.html

Though I'm not sure if those risks apply that strongly in this context as we are dealing with a hook for a packaging tool in the first place which is fairly unlikely to be run in a packaged context itself. But I thought it was worth being aware of.

ankith26:
Hmmm. Even I was not sure of this method. Is there a better way to get the absolute path of the pygame module, instead of the way I have done it. 

I did test on my raspberry pi with 3 code samples and found no issues so far. Soon I’m going to test on my windows machine.

ankith26:
> Though I'm not sure if those risks apply that strongly in this context as we are dealing with a hook for a packaging tool in the first place which is fairly unlikely to be run in a packaged context itself. But I thought it was worth being aware of.

Yes, a hook file is executed only one, when pyinstaller is freezing a project. After the package is frozen, there is no use for the hook-pygame file (it is not going to get executed again), so we are in clear for using `__file__` attribute.

I did search for alternatives, as using `__file__` didnt look like an elegant solution to me, but all the alternatives I found, nothing could beat `__file__`

MyreMylar (approved):
OK, now got a chance to test this PR.

I used the latest version of pyinstaller from their repo which has Python 3.8 support using this link:

`pip install https://github.com/pyinstaller/pyinstaller/archive/develop.tar.gz`

Tried using current master of pygame, running pyinstaller in the most basic configuration:

`pyinstaller pyinstaller_test.py`

...with this test code in a script called `pyinstaller_test.py` from the linked issue:

```
import pygame
import sys

pygame.init()
windowSurface = pygame.display.set_mode((450, 25), 0, 32)
pygame.display.set_caption('Breaking pygame')

x = None  # broken_pygame.exe
# x = 'Sans'  # non-broken_pygame.exe

print("This is the line that breaks the .exe's")
basicFont = pygame.font.SysFont(x, 20)
print('See, I never make it here when that is None')

text = basicFont.render("I can crash pyinstaller .exe's that use pygame!", True, (255, 0, 0))
textRect = text.get_rect()
windowSurface.blit(text, textRect)
pygame.display.update()

while True:
        if event.type == pygame.QUIT:
            pygame.quit()
            sys.exit()

```

And (unsurprisingly) got the pyinstaller crash as described in that issue.

- 

Next, I tried the same process but instead compiling with this PR and.... it all worked perfectly!

Congratulations, this all seems to be fixed and I'm sure will provide relief to a lot of people trying to compile their pygame scripts with pyinstaller.

Thankyou for the contribution. 🎉 👍

ankith26:
Thanks for the approval.

ankith26:
I have a question, relevant to this PR. 

In the src_py directory, There are many images for the pygame icon. My question is that are all images really used by pygame anywhere? I know that the bmp image is used to set the default icon. What about the other images? I have written hooks to include those with pyinstaller generated exe’s, but the size of those image together exceed 200kb, an extra overhead for every pyinstaller generated pygame exe. If those images are not used by pygame, maybe I cannot remove them from the hooks

MyreMylar:
from what I can see these are used:

 - `pygame_icon.tiff` - Used on mac/OSX for the icon
 - `pygame_icon.bmp` - Used on windows/linux for the icon
 - `pygame.ico` - Used in the docs for the html favicon.

And these ones are not:

 - `pygame_icon.icns` - Seems to be a mac icon format, perhaps it was replaced by tiff at some point or this is included in src_py for another reason?
 - `pygame_icon.svg` - Doesn't appear to be used anywhere in the code.

Though I may have missed something.

You could likely drop the .ico file as it appears to be the largest and only used in the docs and the svg file. I'm not sure if you can use Pyinstaller to package for mac when not running on a mac and visa versa for windows/linux. If that is how it works then you might be able to drop the platform specific icons when not running on their platform.

ankith26:
Latest commit is to ensure that only required files are loaded. This is an untested commit, that I will be testing soon.

illume:
I wonder if it's worth adding a tests/pyinstaller_test.py ? To test on mac/win/linux.

As a unit test, it could create an executable and run it. Maybe using something from examples/.
It would skip the test if pyinstaller can not import. So that local test runs without pyinstaller will not fail.
Then we'd need to add a `python -m pip install pyinstaller` to each of the CI jobs we want to test it on.

ankith26:
Yeah, because if there is any bug in this hooks file, then CI/CD is not going to catch it. Btw, I needed confirmation on wether including only these 3 data files in the hooks is fine or I am missing some file.

illume:
I can't think of any others required.

ankith26:
Ok, so I guess after a bit of testing even this PR will be ready for merging.

ankith26:
> As a unit test, it could create an executable and run it. Maybe using something from examples/.

Yeah about that, I don’t think any of the examples are going to work when we use them like
```python
import pygame.examples.aliens
pygame.examples.aliens.main()
```
and bundle them with pyinstaller. Simply because the resources those programs need will not be loaded by pyinstaller. 

It is not that difficult to write hooks to include those. But the problem is you will have to include all the data files, and every pyinstaller generated pygame application will have megabytes of more data that most programs are not even going to use.

illume:
Sorry about not being very clear. We could/should run a test for pyinstaller on our different CI platforms. So we can keep pyinstaller working with pygame in the future. I'll write another issue for that.

mustfa1980:
can you help me with passing none as font thing?

mustfa1980:
can you help me with passing none as font thing?

Issue #2110, opened 2020-09-20
manylinux builds not uploading to github on release


illume:
For the 2.0.0.dev12 release. They did upload to pypi however.

https://travis-ci.org/github/pygame/pygame/jobs/728620021#L28646
```
InvalidDistribution: Cannot find file (or expand pattern): 'dist/*.tar.gz'
```

Issue #2108, opened 2020-09-20
"setup.py sdist" broken


illume:
We should test as part of CI.

McSinyx:
I think I accidentally fixed this as part of GH-2105.

PR #2137, opened 2020-10-04
Test sdist

illume:
Adding a basic test for `python setup.py sdist`, since it was broken for python2.7 in the last release.

https://github.com/pygame/pygame/issues/2108

May also fix the issue with manylinux builds not uploading to github: https://github.com/pygame/pygame/pull/2110


illume:
(note I cancelled the appveyor jobs because the tests are run on travis CI, they didn't fail)

Issue #2109, opened 2020-09-20
timedelta failing on VMs like used on mac Travis release build


illume:
time delta based tests in time_test.py are failing on VMs like Travis CI Mac.

- [x] skip tests on Travis Mac builds
- [x] look at skipping on all CI builds, and reducing the time delta to acceptable ranges for user interactive systems.

Here are the failures: https://travis-ci.org/github/pygame/pygame/builds/728629710#L18075
```
======================================================================
FAIL: test_get_fps (pygame.tests.time_test.ClockTypeTest)
test_get_fps tests pygame.time.get_fps()
----------------------------------------------------------------------
Traceback (most recent call last):

  File "/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp2ayzmg3d/lib/python2.7/site-packages/pygame/tests/time_test.py", line 25, in test_get_fps
    self._fps_test(c, 100, delta)
  File "/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp2ayzmg3d/lib/python2.7/site-packages/pygame/tests/time_test.py", line 36, in _fps_test
    self.assertAlmostEqual(clock.get_fps(), fps, delta=fps*delta)

AssertionError: 12.195121765136719 != 100 within 30.0 delta

======================================================================
FAIL: test_get_rawtime (pygame.tests.time_test.ClockTypeTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp2ayzmg3d/lib/python2.7/site-packages/pygame/tests/time_test.py", line 55, in test_get_rawtime
    self.assertAlmostEqual(delay_miliseconds, c1, delta=delta)

AssertionError: 100.0 != 163 within 50 delta

======================================================================
FAIL: test_get_time (pygame.tests.time_test.ClockTypeTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/tmp2ayzmg3d/lib/python2.7/site-packages/pygame/tests/time_test.py", line 82, in test_get_time
    self.assertAlmostEqual(delay_miliseconds, c1, delta=delta)
AssertionError: 100.0 != 250 within 50 delta

======================================================================
FAIL: test_tick (pygame.tests.time_test.ClockTypeTest)
[...]

PR #2139, opened 2020-10-04
Disable timing based tests on mac CI, because issues.

illume:
For https://github.com/pygame/pygame/issues/2109

Because delta time is never high enough on Mac CI vms it seems.


MyreMylar (approved):
Seems good to me, and likely could be re-used if we decide to disable timing tests on CI in general.

PR #2140, opened 2020-10-05
adds a timeout option to event.wait()

### LINKED ISSUE #2131: SDL2: pygame.event.wait() with timeout
pygame.event.wait() blocks indefinitely, matching the semantics of the [SDL_WaitEvent() function in SDL 1.2](https://www.libsdl.org/release/SDL-1.2.15/docs/html/sdlwaitevent.html).

However, SDL2 brings a function [SDL_WaitEventTimeout()](https://wiki.libsdl.org/SDL_WaitEventTimeout?highlight=%28%5CbCategoryEvents%5Cb%29%7C%28CategoryEnum%29%7C%28CategoryStruct%29) which allows specifying a timeout in milliseconds.

This is ideal in applications that are driven by a scheduler and rather than trying to hit a constant frame-rate, eg. a UI that only updates in response to user input but wants to do other scheduled things like blinking the cursor.

The key paradigm is

* Wake up immediately if there is an input event
* Otherwise wake up at the time of the next scheduled event

and `pygame.event.wait(timeout=t)` would make this much easier and more performant than it is today.

Starbuck5:
Closes #2131, addresses most of #202.

Returns a pygame.NOEVENT event if it doesn't get an event after x milliseconds.

Works as a keyword or as an optional argument.

I'm not sure that I did the event.pyi typing right.


src_c/event.c
```diff
@@ -2083,7 +2103,7 @@ static PyMethodDef _event_methods[] = {
     {"get_grab", get_grab, METH_NOARGS, DOC_PYGAMEEVENTGETGRAB},
 
     {"pump", pg_event_pump, METH_NOARGS, DOC_PYGAMEEVENTPUMP},
-    {"wait", pg_event_wait, METH_NOARGS, DOC_PYGAMEEVENTWAIT},
+    {"wait", pg_event_wait, METH_VARARGS | METH_KEYWORDS, DOC_PYGAMEEVENTWAIT},
```
ankith26, line 2106: I am no expert with C, but I guess this line should have a `(PyCFunction)` like
`{"wait", (PyCFunction)pg_event_wait, METH_VARARGS | METH_KEYWORDS, DOC_PYGAMEEVENTWAIT},`

Starbuck5: Thanks for pointing that out, I missed that. It built locally for me, but hopefully this fixes it for the CI.

MightyJosip (commented):
Your code (probably) fails on python 2 because you initialized kwids after you run another function. Try to move it before `VIDEO_INIT_CHECK()`, so you will get something like this:
```
    SDL_Event event;
    int status;
    int timeout = 0;
    static char *kwids[] = {
        "timeout",
        NULL
    };

    VIDEO_INIT_CHECK();
```

Starbuck5:
YES it finally passes! Thank you for the help MightyJosip and ankith26.

ankith26:
Finally... the test passed. The last fix was a weird one, but worked nonetheless

MightyJosip:
Np, I had the same problem a lot of times so I am used to seeing something like that

docs/reST/ref/event.rst
```diff
@@ -229,9 +229,12 @@ and the ``joy`` attribute was deprecated.
...
+   | :sg:`wait(timeout) -> EventType instance`
 
    Returns a single event from the queue. If the queue is empty this function
-   will wait until one is created. The event is removed from the queue once it
+   will wait until one is created. If a ``timeout`` argument is given, the
+   function will return an event of type ``pygame.NOEVENT`` if no events enter the
+   queue in ``timeout`` milliseconds. The event is removed from the queue once it 
    has been returned. While the program is waiting it will sleep in an idle
    state. This is important for programs that want to share the system with
    other applications.
```
MyreMylar, line 240: These docs are great, but they need to mention somewhere that the timeout parameter only exists in pygame 2, just so our users still using pygame 1 don't get confused.

Usually we are doing this with one of these:

```
.. versionadded:: 2.0.0 ``timeout`` parameter
```

Added under the main block of documentation for the function, but I'd also add something like:

```
From pygame 2.0.0, if a ``timeout`` argument is given, the 
function will return an event of type ``pygame.NOEVENT`` 
if no events enter the queue in ``timeout`` milliseconds.
```

MyreMylar (changes requested):
All seems great. I tested it on locally on windows and the event tests all pass and it functions as I'd expect with this test case:

```
import pygame

pygame.display.init()

screen = pygame.display.set_mode([500, 500])
screen.fill((255, 0, 255))

clock = pygame.time.Clock()

running = True

while running:
    start_time = clock.tick()
    pygame.display.flip()

    wait_event = pygame.event.wait(timeout=10000)
    if wait_event.type == pygame.NOEVENT:
        end_time = clock.tick()
        print('received no event for', int((end_time - start_time)/1000.0), 'seconds.')
    else:
        print(wait_event)

```
Output:

```
pygame 2.0.0.dev13 (SDL 2.0.12, python 3.8.5)
<Event(512-WindowEvent {'event': 1, 'window': None})>
<Event(32768-ActiveEvent {'gain': 1, 'state': 1, 'window': None})>
<Event(512-WindowEvent {'event': 12, 'window': None})>
<Event(770-TextEditing {'text': '', 'start': 0, 'length': 0, 'window': None})>
<Event(32770-VideoExpose {})>
<Event(512-WindowEvent {'event': 3, 'window': None})>
received no event for 10 seconds.
```

I quickly hacked up.

Just one small niggle with the documentation to fix and I think we are good to merge.

ankith26:
Also, a small detail I noticed.
If an sdl error occurs, the pygame user who called event.wait with timeout is going to get pygame.noevent

docs/reST/ref/event.rst
Starbuck5: Thanks, thats a good catch.
I looked around and in other places, the docs were using ``.. versionchanged::`` for new arguments, like in pygame.image.save and with all the new keyword arguments in draw functions. So I changed the events doc ``.. versionadded::`` to ``.. versionchanged::`` to be consistent with those other ones.

Starbuck5:
Yes, that is intentional.
[SDL_WaitEventTimeout](https://wiki.libsdl.org/SDL_WaitEventTimeout) returns 0 if there is an error OR if it times out. So there isn't a reliable way of telling if it errored. I thought about timing it with SDL_GetTicks, but that might break down at lower precisions (like if someone puts in ``timeout=1``).

illume (approved):
👍 Thanks :)

illume:
Merging, because the requested changes have all been done.

Issue #2113, opened 2020-09-22
pre-multiplied alpha continued


notpygame:
BLEND_PREMULTIPLIED was implemented in this issue: https://github.com/pygame/pygame/issues/1289

To make it more usable for people, the following could be done:

- [ ] pygame.color method for working with pre-multiplied alpha (details needed)
- [ ] Surface.convert() method for working with pre-multiplied alpha (details needed).
- [ ] tutorial on using pre-multiplied alpha to docs or website wiki.

MyreMylar:
I think my initial idea for adapting `.convert_alpha()` was too unfocused and messing with the clarity of an existing function.

Instead I think I should mirror the PR for the Color class and just add a `.premul_alpha()` method to Surface that begins by erroring out if the surface is not 32bit with an alpha channel.

PR #2142, opened 2020-10-05
Adding a premul_alpha helper method to Color class

MyreMylar:
Related to the first part of #2113. 

The idea is that this could be used with methods like `surface.fill()` to create colours that are ready to be used with the already existing `BLEND_PREMULTIPLIED` blend mode.

Here is a quick example to show the usage and the difference between the alpha blend modes for those that don't know what I'm talking about:

```python
import pygame
import sys

pygame.init()
window_surface = pygame.display.set_mode((300, 100), 0, 32)
window_surface.fill((255, 0, 0))
pygame.display.set_caption('Pre-multiplied Alpha')

basicFont = pygame.font.SysFont(None, 20)

# pre-multiplied (visible on left hand side)
alpha_surface_1 = pygame.Surface((50, 50), pygame.SRCALPHA, 32)
alpha_surface_1.fill(pygame.Color(255, 255, 255, 200).premul_alpha())

alpha_surface_2 = pygame.Surface((100, 100), pygame.SRCALPHA, 32)
alpha_surface_2.fill(pygame.Color(0, 0, 200, 15).premul_alpha())

alpha_surface_2.blit(alpha_surface_1, (0, 0), special_flags=pygame.BLEND_PREMULTIPLIED)

window_surface.blit(alpha_surface_2, (0, 0), special_flags=pygame.BLEND_PREMULTIPLIED)
pygame.display.update()

# not pre-multiplied (visible on right hand side)
alpha_surface_1 = pygame.Surface((50, 50), pygame.SRCALPHA, 32)
alpha_surface_1.fill(pygame.Color(255, 255, 255, 200))

alpha_surface_2 = pygame.Surface((100, 100), pygame.SRCALPHA, 32)
alpha_surface_2.fill(pygame.Color(0, 0, 200, 15))

alpha_surface_2.blit(alpha_surface_1, (0, 0))

window_surface.blit(alpha_surface_2, (200, 0))
pygame.display.update()

while True:
        if event.type == pygame.QUIT:
            pygame.quit()
            sys.exit()
```

Output:

image


nthykier:
Open question: Should we prevent programmers from calling `premul_alpha()` on an already `premul_alphed()`'ed color?  It would require an additional field to track whether a given color instance was premul'ed or not.

MyreMylar:
I thought about it, but thought it might become quite confusing/involved/unreliable because there are lots of ways you can get hold of Colors in pygame that may already be _visually_ pre-multiplied but that we couldn't reliably tag. e.g.

- Load an image saved in a paint program with pre-multiplied alpha with `pygame.image.load()` then use `surface.get_at((x,y))` to get a Color back. This colour can't be tagged as pre-multiplied even though it is.
- Convert a filled alpha blended surface with the new `convert_alpha(premul=True)` method. It's now also got pre-multiplied alpha colour on it, but we can't easily track that color around a program unless we modify every function that interacts with a surface and every function that takes a Color from a surface.

So you end up with some alpha pre-multiplied Colors bearing a tag, and perhaps some others in an application without the tag.

If we did want to do it, I guess the best way would be to use something like setattr/hasattr (that's PyObject_SetAttr in CPython) to add a variable that doesn't exist on a normal Color and then test if that variable exists in the function. Then we wouldn't have an extra variable on regular Colors for people who don't care about this blend mode. I guess the variable would become invalid if any of the other attributes on the Color were modified so would need to be removed then with delattr. 

Slightly relatedly, its also not an easily reversible operation, e.g. if you pre-multiply a bright Color with an alpha of 5 you are going to lose a lot of colour data in any attempt to reverse from the pre-multiplied Color. So I think if someone wants the original Colors then they should save them separately and name the pre-multiplied version differently. This function returns a Color rather than modifying in place so that should be straightforward.

docs/reST/ref/color.rst
```diff
@@ -224,4 +224,21 @@
...
+      channels have been multiplied by the alpha channel of the original
+      color. The alpha channel remains unchanged.
+
+      This is useful when working with the ``BLEND_PREMULTIPLIED`` blending mode
+      flag for :meth:`pygame.Surface.blit()`, which assumes that all surfaces using
+      it are using pre-multiplied alpha colors.
+
+      .. versionadded:: 2.0.0
+
+      .. ## Color.lerp ##
```
nthykier, line 242:
```suggestion
      .. ## Color.premul_alpha ##
```

nthykier (approved):
LGTM other than one comment (see my suggestion)

illume (approved):
👍 great :) Can I merge now?

MyreMylar:
Yeah, I don't think there is anything else to do here.

PR #2143, opened 2020-10-05
Adding an optional 'premul' parameter to surface.convert_alpha()

MyreMylar:
This is related to the second part of #2113. 

A lot more work to do here but I got something basically running on SDL2.

Things to address:

 - [x] Trim down code to only parts that are actually being used. Currently built from blitting code which has many more options.
 - [x] Need to decide on any handling for non-32 bit surface formats with per pixel alpha (Are there any?) - _There are only 16 bit surfaces that can also have per pixel alpha - I'm just passing these through to the non-SSE2 version._
 - [x] Needs an SSE2 version for speed.
 - [x] Needs tweaks for SDL 1.
 - [x] Needs testing.
 - [x] Needs documentation.
 - [ ] Figure out how to convert 8 bit surfaces from font renders to premultiplied alpha.


MyreMylar:
I guess the main question remaining is whether to stick this in convert_alpha() as I have it now, or put it in a new separate method. 

**Arguments for:**

- It is an alpha conversion operation on a surface.
- In general, fewer methods are preferable to more.

**Arguments against:**

- It doesn't work on exactly the same range of inputs as a `convert_alpha()` without the `premul=True` parameter because you have to start with an alpha channel to multiply the RGB channels by it, whereas regular convert_alpha() will take any type of surface and return one with an alpha channel.
- Requires extra `if` check in `convert_alpha()` slowing it down a miniscule amount.

illume:
Will one of these work? Or do you need to do convert_alpha first, then a premul?
```python
pygame.load.image().convert_alpha(premul=True)

pygame.load.image().convert_premul()
```

MyreMylar:
This:

`pygame.load.image(f).convert_alpha(premul=True)`

 will work on it's own to take a 16 or 32 bit image with per pixel alpha to a 'display-surface-bit' image with alpha, which is normally 32 bit.

The only difference in what surfaces this works on vs the standard `surface.convert_alpha()` with no parameters, is that the surface needs to start with an alpha channel for it to do anything, which make sense since we are multiplying the R, G & B channels by the alpha channel - which is hard to do if there is no alpha channel. As far as I could see the only surfaces with per pixel alpha supported by pygame are ones with 16 or 32 bit depth. 

You could call:

```
# where f is a 24bit surface
pygame.load.image(f).convert_premul()
pygame.load.image(f).convert_alpha(premul=True)
```
but it would be a waste of time to do the premul because every pixel would have 255 alpha and thus no change would take place from the multiply.

PR #2148, opened 2020-10-06
Adds draw_doc.h and font_doc.h changes.

MyreMylar:
These files are only generated when the docs are compiled so I guess they were missed off a commit somewhen.


Starbuck5 (approved):
Looks good. I saw these changed files locally and I didn't know what the deal was.

illume (approved):
👍

PR #2151, opened 2020-10-06
Improved the speed of pygame.draw.line()

### LINKED ISSUE #2033: pygame.draw.line is terrible slow in pygame. 2.0.0.x
Compared with pygame 1.9.6 the pygame.draw.line method is terrible slow in pygame 2.0.0.dev10 

It is so slow that turns my application to a slideshow where as with 1.9.3 it runs fluid (sorry, no fps comparison).

Using Windows and python 3.7.3

MightyJosip:
For people who want TL;DR: gotta line fast.

For others, this should fix #2033.

If you draw the line that is completely inside of the screen and has `width = 1` it is the same.

If you draw the line that has `width != 1` it should be faster by a factor of `width * c`, where c is really small number.

But if you draw for example this line: `pg.draw.line(screen, "WHITE", (1000000000, 1000000000), (-1000000000, -1000000000))` than you will see the noticeable difference (in dev6-dev12 it would almost completely crash your game, however now it should be as fast as drawing any other line that goes from one edge to another).

PS: probably the most important improvement in this PR is that I updated pygame draw example image with the recent ellipse changes.


MyreMylar (approved):
Tested with this:

```
import pygame

pygame.init()

window_surf = pygame.display.set_mode((800, 600))

background_surf = pygame.Surface((800, 600))
background_surf.fill(pygame.Color('#000000'))

pygame.draw.line(background_surf, "WHITE", (1000000000, 1000000000), (-1000000000, -1000000000))

clock = pygame.time.Clock()
running = True

while running:
    frame_time = clock.tick(60)/1000.0

        if event.type == pygame.QUIT:
            running = False

    window_surf.blit(background_surf, (0, 0))

    pygame.display.update()
```

Which hangs pygame for several second for me in dev12 (on windows) and now runs perfectly fine with this PR. 

The other draw tests all pass locally as well.

Approved! 👍 🎉

illume:
Was this file supposed to be removed? docs/reST/ref/code_examples/draw_module_example.png

illume (approved):
Very, very good :)

Issue #2152, opened 2020-10-07
Results of unit tests on raspi, python 3.7 vs python 3.9


ankith26:
I have the latest raspberry pi OS (buster), on my raspberry pi 3b.
Python version 3.7 is pre-installed, I have installed pygame 2.0.0dev12 using pip

I have also installed python v3.9 using pyenv and pygame 2.0.0dev12

SDL version is 2.0.9 in both cases.

Below I have pasted the last part of the output text of running pygame standard unittests.

For Python 3.9

```
======================================================================
FAIL: test_SRCALPHA (pygame.tests.blit_test.BlitTest)
SRCALPHA tests.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/pi/.pyenv/versions/3.9.0/lib/python3.9/site-packages/pygame/tests/blit_test.py", line 19, in test_SRCALPHA
    self.assertEqual(s.get_at((0, 0)), d.get_at((0, 0)))
AssertionError: (0, 0, 255, 0) != (0, 0, 255, 255)

======================================================================
FAIL: test_get_active_iconify (pygame.tests.display_test.DisplayModuleTest)
Test the get_active function after an iconify
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/pi/.pyenv/versions/3.9.0/lib/python3.9/site-packages/pygame/tests/display_test.py", line 112, in test_get_active_iconify
    self.assertEqual(pygame.display.get_active(), False)
AssertionError: True != False

======================================================================
FAIL: test_iconify (pygame.tests.display_test.DisplayModuleTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/pi/.pyenv/versions/3.9.0/lib/python3.9/site-packages/pygame/tests/display_test.py", line 286, in test_iconify
    self.assertEqual(minimized_event, True)
AssertionError: False != True

======================================================================
FAIL: test_blanket_alpha (pygame.tests.surface_test.SurfaceSelfBlitTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/pi/.pyenv/versions/3.9.0/lib/python3.9/site-packages/pygame/tests/surface_test.py", line 3280, in test_blanket_alpha
    self._assert_same(surf, comp)
  File "/home/pi/.pyenv/versions/3.9.0/lib/python3.9/site-packages/pygame/tests/surface_test.py", line 3195, in _assert_same
    self.assertEqual(
[...]

PR #2153, opened 2020-10-07
Mostly this fixes failing unit tests

ankith26:
Mostly deals with the `iconify()` function unittest, #2152

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -2,6 +2,7 @@
 
 import unittest
 import os
+import time
 
 import pygame, pygame.transform
 from pygame.compat import unicode_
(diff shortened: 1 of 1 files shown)


ankith26:
Previously those unit tests would cycle for a Magical 5000 number.. That was not so consistent while testing across different platforms, particularly the low power ones. This PR makes sure that it waits for 1 second for the event so to happen

MyreMylar:
That seems a lot more sensible.

MyreMylar (approved):
Thanks for the contribution. Hopefully this will make these tests less likely to fail across more hardware. 

👍 🎉

illume (approved):
👍

PR #2149, opened 2020-10-06
Misc doc updates

ankith26:
Some more minor misc doc updates. I find myself at pygame docs most of the time, so I am polishing them a bit to help people like me :)

PR #2157, opened 2020-10-07
Adding windowevent constants

ankith26:
This PR is adding the new SDL WINDOWEVENT constants


ankith26:
I will add docs for these changes as part of #2149

MyreMylar (approved):
Thanks for the contribution. 

I suspect later we may need to document and test these events, assuming they are all generated & working on the pygame side, but this is a good change that'll make it easier to do those things.

👍 🎉

ankith26:
Yes, I have seen that WINOWEVENT events are consistently being generated, and the event attribute is always set to the respective integer, which I see why I have opened the PR to add these constants.

I’m aldready working on docs for these, and maybe we can add a few unittests for these too

illume (approved):
👍 thanks. Nice improvement.

I agree, docs and tests can come later in further PRs.

PR #2158, opened 2020-10-07
A few more type hint fixes

MightyJosip:
- Add new joystick functions that were added in #2027.
- Add support for functions that accept argument from pgRect_FromObject(). Right now code like this:
```
class MyRect():
    def __init__(self):
        self.rect = pg.Rect(100, 100, 100, 100)
pg.display.update(MyRect())
```
would give you the error, however with this changes it should work.


MyreMylar:
I'm wondering if it is time to have a `pygame_stub_helper_types.py` file with all the Union types like: `_CanBeRect` & `_ColorValue` in, just so we don't need to have them defined across multiple files?

buildconfig/pygame-stubs/gfxdraw.pyi
```diff
@@ -8,8 +9,17 @@ from pygame.rect import Rect
...
     Color, Tuple[int, int, int], List[int], int, Tuple[int, int, int, int]
 ]
-_RectValue = Union[Rect, Tuple[int, int, int, int]]
 _Coordinate = Union[Tuple[float, float], List[float], Vector2]
+_CanBeRect = Union[
+    Rect,
+    Tuple[int, int, int, int], List[int],
+    Tuple[_Coordinate, _Coordinate], List[_Coordinate]
+]
+class _HasRectAttribute(Protocol):
```
illume, line 18: 👍

buildconfig/pygame-stubs/display.pyi
```diff
@@ -1,4 +1,5 @@
 from typing import Union, Tuple, List, Optional, Dict, Sequence
+from typing_extensions import Protocol
```
illume, line 2: 🦆

illume (approved):
👍

PR #2159, opened 2020-10-07
Remove ancient file kept for python v2.4 compatibility

ankith26:


ankith26:
The file that was deleted was kept for python v2.4 compat (a version released almost 16 years ago). So I guess we can safely delete it.

src_py/threads/__init__.py
```diff
@@ -19,9 +19,7 @@
 if sys.version_info[0] == 3:
     from queue import Queue
     from queue import Empty
-elif (sys.version_info[0] == 2 and sys.version_info[1] < 5):
-    from Py25Queue import Queue
-    from Py25Queue import Empty
+
```
McSinyx, line 22: Nitpick: I think you accidentally left a trailing newline here.  Also since you're touching this part of the import, would it be a good idea to join the import statement above and below as `from queue import Empty, Queue`?

```suggestion
```

ankith26: Made the changes

MyreMylar (approved):
No reason for anyone to be using Python 2.5 now that I can see. I believe it is policy that we are supporting 2.7 only for python 2.

illume (approved):
Nice cleanup.

Issue #2099, opened 2020-09-14
SDL2 display.set_palette


illume:
set_palette is not supported in SDL2

See display.set_palette PR and issue:
- https://github.com/pygame/pygame/issues/1754
- https://github.com/pygame/pygame/pull/1972

robertpfeiffer:
It's easy to emulate this with OpenGL, or to have a layer of indirection to perform the palette swap every time on pygame.display.update. I'm not sure if it's worth implementing the ugly but reliable solution here.

illume:
From what I can tell, there is no easy solution for this. There are some [users of this function](https://github.com/search?q=%22pygame.display.set_palette%22&type=code), so it's important. However, there's no way to address it before the release of pygame 2 on the 28th of Oct. For these reasons I'll remove the critical label, and this can be added to a list of things that are not compatible (along with cdrom, and scrap) in the release.

PR #2160, opened 2020-10-07
Changed display.set_palette() test to test correct function.

MyreMylar:
Was investigating #2099 - which looks like a pain to resolve with the current rendering setup - and noticed that the test that kicked it off was actually testing setting the palette on the display _surface_, rather than the display as it was supposed to. Two identically named functions in different modules - easy to miss.

Anyway, this changes just switches the test in `display_test.py` to call the display module version instead and removed the parts of the test that don't apply (display.set_palette() explicitly _doesn't_ change the palette of the display surface). There is already a test for `surface.set_palette()` in that module's tests.

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -206,12 +206,12 @@ class DisplayModuleTest(unittest.TestCase):
         surface = pygame.display.get_surface()
 
         self.assertIsNone(surface)
-        
+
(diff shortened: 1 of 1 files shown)


illume (approved):
👍 good catch. Thanks.

PR #2161, opened 2020-10-10
Fix repeated VIDEOEXPOSE events

ankith26:
A PR to fix an issue where too many VIDEOEXPOSE events are placed in the event queue.

@MyreMylar suggested the code for solving this issue.

@robertpfeiffer This is the fix

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -144,6 +144,19 @@ RemovePending_PGS_VIDEORESIZE_Events(void * userdata, SDL_Event *event)
     return 1;
 }
 
+static int SDLCALL
+RemovePending_PGS_VIDEOEXPOSE_Events(void * userdata, SDL_Event *event)
+{
+    SDL_Event *new_event = (SDL_Event *)userdata;
+
+    if (event->type == SDL_VIDEOEXPOSE &&
+        event->window.windowID == new_event->window.windowID) {
+        /* We're about to post a new size event, drop the old one */
+        return 0;
+    }
+    return 1;
+}
+
 /*SDL 2 to SDL 1.2 event mapping and SDL 1.2 key repeat emulation*/
 static int SDLCALL
 pg_event_filter(void *_, SDL_Event *event)
@@ -191,6 +204,8 @@ pg_event_filter(void *_, SDL_Event *event)
                 {
                     SDL_Event newevent = *event;
                     newevent.type = SDL_ACTIVEEVENT;
+                    
+                    SDL_FilterEvents(RemovePending_PGS_VIDEOEXPOSE_Events, &newevent);
                     SDL_PushEvent(&newevent);
                     return 1;
                 }
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -192,6 +192,8 @@ pg_event_filter(void *_, SDL_Event *event)
                 {
                     SDL_Event newevent = *event;
                     newevent.type = SDL_VIDEOEXPOSE;
+                    
+                    SDL_FilterEvents(RemovePending_PGS_VIDEOEXPOSE_Events, &newevent);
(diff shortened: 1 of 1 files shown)


ankith26:
Oops I guess I did a mistake here. Ill fix it in the next commit

ankith26:
Anyone’s has any clue why the appveyor python 3.7 x64 build I still randomly failing. I’ve seen that happen with other PRs too

ankith26:
For those of you who are wondering why we are opening and closing PRs, we are just testing and debugging issues with the CI

ankith26:
:( Appveyor builds failing

ankith26:
After a bit of local testing I can tell that this fix is working. So this PR is done from my end. 
Now for the CI to pass and then this PR can be reviewed.

ankith26:
Started the CI

ankith26:
The builds are finally successful

MyreMylar (approved):
Looks like it works. Tested it on windows with a simple test program than prints the events to the console and moved the window on and off the screen. 

Before this PR many `<Event(32770-VideoExpose {})>` are generated for each `<Event(512-WindowEvent {'event': 4, 'window': None})>`. Afterwards there is a one-one correspondence which seems a lot clearer to me.

ankith26:
Actually, the thanks must go to you @MyreMylar, you suggested fix :)

src_c/event.c
```diff
@@ -144,6 +144,19 @@ RemovePending_PGS_VIDEORESIZE_Events(void * userdata, SDL_Event *event)
...
 }
 
+static int SDLCALL
+RemovePending_PGS_VIDEOEXPOSE_Events(void * userdata, SDL_Event *event)
+{
+    SDL_Event *new_event = (SDL_Event *)userdata;
+
+    if (event->type == SDL_VIDEOEXPOSE &&
+        event->window.windowID == new_event->window.windowID) {
+        /* We're about to post a new size event, drop the old one */
```
robertpfeiffer, line 154: This copy-pasted comment is not correct here.

ankith26: There you go, fixed it

src_c/event.c
```diff
@@ -144,6 +144,19 @@ RemovePending_PGS_VIDEORESIZE_Events(void * userdata, SDL_Event *event)
     return 1;
 }
 
+static int SDLCALL
+RemovePending_PGS_VIDEOEXPOSE_Events(void * userdata, SDL_Event *event)
```
illume, line 148: This should be probably be called: `_pg_remove_pending_PGS_VIDEOEXPOSE`.

Because of the naming conventions:
- `_` prefix because it's for private internal use.
- `pg` prefix because it's a pygame function.

ankith26: Yeah, but I had just copy pasted fro my the functions name above, let’s me make that modification to that function too

ankith26: Did it :)

illume:
I can't replicate the problem on Macos.

I tried with the eventlist example, and dragging the window around to the edge of the screen, and into a second screen. But did not see any VIDEOEXPOSE events at all.
```bash
python3 -m pygame.examples.eventlist
```

illume (changes requested):
Before merging, it would be good to confirm it's ok:
- on another windows machine
- on an X11 machine

illume:
There is a difference:

- On windows 10 1.9.6: If I move the window so it is visible, I do **NOT** get an VIDEOEXPOSE.
- On windows 10 this PR: If I move the window so it is visible, I do get a VIDEOEXPOSE and a WINDOWEVENT_MOVED.

But I don't think it's due to this PR.
I didn't try different video backends.

illume (approved):
Great :) No more VideoExposed spam. Thanks.

Issue #592, opened 2018-10-30
Deprecate draw.aaline : blend=False


e1000:
I agree with @artifactz (cf #575) that it does not make much sense not to blend an aaline (it is just in fact blended with black .... )

If we (deprecate and then) remove it, we should also remove it in `aalines` ...

**To Do to close:**

- [ ] Add python deprecation warning (use warnings module?) for calls to draw.aaline with blend=False for 2.0 release saying function will be deprecated in pygame 2.2.
- [ ] Add unit tests for deprecation warning.
- [ ] Wait until 2.2 then remove this option.

**Related Docs**: https://www.pygame.org/docs/ref/draw.html#pygame.draw.aaline

e1000:
I don't mind so much, it's just weird.
But I thought it's also a question of performance to remove this "if" for each pixel we draw.

And deprecating a strange feature and remove it later on, seems to me quite natural...

illume:
Yop, I'm not really clear on the use case... but I guess it was added for a reason, and people do use it.🤷‍♂️ 

Yeah, the `if` could be nice to remove for speed. Something like the alpha blit functions could be written. Where there is a separate loop for each case, but that's a bit unwieldy.

e1000:
Another thing is that we will have to add some more **if**s for resolving the issues of drawing on the border of the surface (#512), so that will slow down *aaline* again. I think I will manage to have other performance improvements in new version of the code, but still.

I would propose, we deprecate it, and if somebody complains, he should explain why he needs it and why it makes sense; and we would remove the deprecation warning at next release. On the other hand, if nobody complains, we're fine to remove it...

dlon:
The compiler should be able to [unswitch](https://en.wikipedia.org/wiki/Loop_unswitching) that loop, so I probably wouldn't worry about the if statement. Well, in the C implementation anyway. You could manually unswitch it in Python, if we can't remove it altogether.

MyreMylar:
**To Do to close:**

- [ ] Add python deprecation warning (use warnings module?) for calls to draw.aaline with blend=False for 2.0 release saying function will be deprecated in pygame 2.2.
- [ ] Add unit tests for deprecation warning.
- [ ] Wait until 2.2 then remove this option.

ankith26:
Thanks to zoldalma999, the deprecation warning is now, implemented. The only thing left to do is to implement it in pygame 2.2, and that is being tracked in #2405 
So closing this issue, 
Cheers!

PR #2162, opened 2020-10-10
Add deprecation warning when draw.aaline is called with blend=False

zoldalma999:
#592

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -115,6 +115,13 @@ aaline(PyObject *self, PyObject *arg, PyObject *kwargs)
         return NULL; /* Exception already set. */
     }
 
+    if (!blend) {
+        PyErr_WarnEx(PyExc_DeprecationWarning,
+                     "blend=False will be deprecated in pygame 2.2 and will "
+                     "default to True",
+                     1);
+    }
+
     surf = pgSurface_AsSurface(surfobj);
 
     if (surf->format->BytesPerPixel <= 0 || surf->format->BytesPerPixel > 4) {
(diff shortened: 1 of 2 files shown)


MightyJosip:
Closing this just to restart appveyor

src_c/draw.c
```diff
@@ -115,6 +115,13 @@ aaline(PyObject *self, PyObject *arg, PyObject *kwargs)
         return NULL; /* Exception already set. */
     }
 
+    if (!blend) {
+        PyErr_WarnEx(PyExc_DeprecationWarning,
```
nthykier, line 119: Missing check of the return value from `PyErr_WarnEx`  If it returns `-1` then an exception was thrown ( see https://docs.python.org/3/c-api/exceptions.html#issuing-warnings), which in turn requires an early return from this function.

src_c/draw.c
```diff
@@ -264,6 +271,13 @@ aalines(PyObject *self, PyObject *arg, PyObject *kwargs)
                             surf->format->BytesPerPixel);
     }
 
+    if (!blend) {
+        PyErr_WarnEx(PyExc_DeprecationWarning,
```
nthykier, line 275: Missing check of the return value from `PyErr_WarnEx`  If it returns `-1` then an exception was thrown ( see https://docs.python.org/3/c-api/exceptions.html#issuing-warnings), which in turn requires an early return from this function.

nthykier (changes requested):
Thanks for looking at this.  I have added some remarks about the code.

Issue #2145, opened 2020-10-06
Python 3.9 support


mgedmin:
Python 3.9 was released yesterday.

`pip install pygame` fails on 3.9 with a _lot_ of warnings hiding a couple of real errors:

```
    building 'pygame.pypm' extension
    gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -D_REENTRANT -DENABLE_NEWBUF=1 -I/usr/X11R6/include -I/usr/include/SDL -I/usr/include -I/usr/include -I/home/mg/src/pyspacewar/.tox/py39/include -I/home/mg/opt/python39/include/python3.9 -c src_c/pypm.c -o build/temp.linux-x86_64-3.9/src_c/pypm.o
    src_c/pypm.c: In function ‘__pyx_pf_4pypm_6Output___init__’:
    src_c/pypm.c:2247:20: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
     2247 |     __pyx_v_errmsg = Pm_GetErrorText(__pyx_v_err);
          |                    ^
    src_c/pypm.c: In function ‘PyInit_pypm’:
    src_c/pypm.c:6419:26: error: ‘PyTypeObject’ {aka ‘struct _typeobject’} has no member named ‘tp_print’
     6419 |   __pyx_type_4pypm_Output.tp_print = 0;
          |                          ^
    src_c/pypm.c:6434:25: error: ‘PyTypeObject’ {aka ‘struct _typeobject’} has no member named ‘tp_print’
     6434 |   __pyx_type_4pypm_Input.tp_print = 0;
          |                         ^
```

`tp_print` was [removed in 3.9](https://docs.python.org/3.9/whatsnew/3.9.html#id3).

This looks like Cython-generated code, so perhaps all that's needed is to re-run a newer version of Cython?

MightyJosip:
Can you try to install pygame with `pip install pygame==2.0.0.dev12`

MyreMylar:
Yeah, looking at the current code in the repo this looks like it has been resolved:

https://github.com/pygame/pygame/blob/0c6043debb5b1f7c4401c717c7069f05834bea06/src_c/pypm.c#L6050-L6051

MyreMylar:
Example commit showing this change on somebody else's repo:

https://github.com/cython/cython/commit/aa56a6b108de18138dfcf8722610327024b11dd9

[...]

PR #2163, opened 2020-10-10
python3.9 wheels

illume:
For https://github.com/pygame/pygame/issues/2145

- cibuildwheel on travis mac should do 3.9 without any changes. https://github.com/joerick/cibuildwheel/pull/440
- Appveyor doesn't have python 3.9 on windows yet. https://github.com/appveyor/ci/issues/3541 added a work around.
- added 3.9 to manylinux  (unfortunately it dropped python 3.4 support... so we have to as well). Unfortunately some test is crashing on travis... but is ok locally.
- travis 3.9 job doesn't exist yet. Use 3.9-dev still. (the manylinux part does the wheels with 3.9).

### DIFF
--- a/buildconfig/appveyor.yml
+++ b/buildconfig/appveyor.yml
@@ -63,6 +63,13 @@ environment:
       USE_SDL2: "-sdl2"
       USE_ANALYZE: ""
 
+    - PYTHON: "C:\\Python39\\python"
+      PYTHONPATH: "C:\\Python39"
+      PYTHON_VERSION: "3.9.0"
+      PYTHON_ARCH: "32"
+      USE_SDL2: "-sdl2"
+      USE_ANALYZE: ""
+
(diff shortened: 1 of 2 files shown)


illume:
Hrmm...
The appveyor python script is here: https://github.com/appveyor/build-images/blob/master/scripts/Windows/install_python.ps1#L88-L108

Something like this might work...
```ps1
        InstallPythonEXE "3.9.0" "x64" "$env:SystemDrive\Python39-x64"
        InstallPythonEXE "3.9.0" "x86" "$env:SystemDrive\Python39"
```

Maybe I should do a PR there.

illume:
The 3.9 manylinux build crashes on travis: https://travis-ci.org/github/pygame/pygame/jobs/734596927#L11892
It passes locally however.

PR #2166, opened 2020-10-11
Revert PR #2127 num_buttons parameter to mouse.get_pressed.

illume:
This https://github.com/pygame/pygame/pull/2127 caused intermittent crashes in python 3.7 x64 windows appveyor.


ankith26:
If #2167 works, then this revert won’t be necessary

PR #2167, opened 2020-10-11
Hopefully fix random crash

ankith26:


ankith26:
The braces were missing. Added those and fixed indentation.

PR #2180, opened 2020-10-12
bufferproxy: initialize PyTypeObject/PyModuleDef using dot notation

cbxbiker61:
First file to use dot initializers.

### DIFF
--- a/src_c/bufferproxy.c
+++ b/src_c/bufferproxy.c
@@ -653,8 +653,15 @@ static PyBufferProcs proxy_bufferprocs = {
     (Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE | Py_TPFLAGS_HAVE_GC)
 #endif
 
+/* Temporary Testing Macro */
+#if defined(MS_WINDOWS)
+#define PYVAROBJECT_HEAD_INIT(a,b) PyVarObject_HEAD_INIT(0,b)
+#else
+#define PYVAROBJECT_HEAD_INIT(a,b) PyVarObject_HEAD_INIT(a,b)
+#endif
+
 static PyTypeObject pgBufproxy_Type = {
-	PyVarObject_HEAD_INIT(&PyType_Type, 0)
+	PYVAROBJECT_HEAD_INIT(&PyType_Type, 0)
     .tp_name = PROXY_TYPE_FULLNAME,
     .tp_basicsize = sizeof(pgBufproxyObject),
     .tp_dealloc = (destructor)proxy_dealloc,
--- a/src_c/bufferproxy.c
+++ b/src_c/bufferproxy.c
@@ -654,7 +654,7 @@ static PyBufferProcs proxy_bufferprocs = {
 #endif
 
 static PyTypeObject pgBufproxy_Type = {
-    TYPE_HEAD(NULL, 0)
+	PyVarObject_HEAD_INIT(&PyType_Type, 0)
     .tp_name = PROXY_TYPE_FULLNAME,
     .tp_basicsize = sizeof(pgBufproxyObject),
     .tp_dealloc = (destructor)proxy_dealloc,
--- a/src_c/bufferproxy.c
+++ b/src_c/bufferproxy.c
@@ -654,44 +654,22 @@ static PyBufferProcs proxy_bufferprocs = {
 #endif
 
 static PyTypeObject pgBufproxy_Type = {
-    TYPE_HEAD(NULL, 0) PROXY_TYPE_FULLNAME, /* tp_name */
-    sizeof(pgBufproxyObject),               /* tp_basicsize */
-    0,                                      /* tp_itemsize */
-    (destructor)proxy_dealloc,              /* tp_dealloc */
-    0,                                      /* tp_print */
-    0,                                      /* tp_getattr */
-    0,                                      /* tp_setattr */
-    0,                                      /* tp_compare */
-    (reprfunc)proxy_repr,                   /* tp_repr */
-    0,                                      /* tp_as_number */
-    0,                                      /* tp_as_sequence */
-    0,                                      /* tp_as_mapping */
-    0,                                      /* tp_hash */
-    0,                                      /* tp_call */
-    0,                                      /* tp_str */
-    0,                                      /* tp_getattro */
-    0,                                      /* tp_setattro */
-    PROXY_BUFFERPROCS,                      /* tp_as_buffer */
-    PROXY_TPFLAGS,                          /* tp_flags */
-    DOC_PYGAMEBUFFERPROXY,
-    (traverseproc)proxy_traverse,         /* tp_traverse */
-    0,                                    /* tp_clear */
-    0,                                    /* tp_richcompare */
-    offsetof(pgBufproxyObject, weakrefs), /* tp_weaklistoffset */
-    0,                                    /* tp_iter */
-    0,                                    /* tp_iternext */
-    proxy_methods,                        /* tp_methods */
-    0,                                    /* tp_members */
-    proxy_getsets,                        /* tp_getset */
-    0,                                    /* tp_base */
-    0,                                    /* tp_dict */
-    0,                                    /* tp_descr_get */
-    0,                                    /* tp_descr_set */
-    offsetof(pgBufproxyObject, dict),     /* tp_dictoffset */
-    0,                                    /* tp_init */
-    PyType_GenericAlloc,                  /* tp_alloc */
-    proxy_new,                            /* tp_new */
-    PyObject_GC_Del,                      /* tp_free */
+    TYPE_HEAD(NULL, 0)
+    .tp_name = PROXY_TYPE_FULLNAME,
+    .tp_basicsize = sizeof(pgBufproxyObject),
+    .tp_dealloc = (destructor)proxy_dealloc,
+    .tp_repr = (reprfunc)proxy_repr,
+    .tp_as_buffer = PROXY_BUFFERPROCS,
+    .tp_flags = PROXY_TPFLAGS,
+    .tp_doc = DOC_PYGAMEBUFFERPROXY,
+    .tp_traverse = (traverseproc)proxy_traverse,
+    .tp_weaklistoffset = offsetof(pgBufproxyObject, weakrefs),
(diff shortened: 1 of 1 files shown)


MyreMylar:
Looks like the Python 2.7 build on windows is failing here:

```
src_c/bufferproxy.c(658) : error C2059: syntax error : '.'
```

Does it need an extra comma on the line above? The Python 2.7 build always tends to be the most fussy.

cbxbiker61:
> Looks like the Python 2.7 build on windows is failing here:
> 
> ```
> src_c/bufferproxy.c(658) : error C2059: syntax error : '.'
> ```
> 
> Does it need an extra comma on the line above? The Python 2.7 build always tends to be the most fussy.

In pgcompat.h...

Python3:
#define TYPE_HEAD(x,y) PyVarObject_HEAD_INIT(x,y)

Python2:
#define TYPE_HEAD(x,y)                          \
    PyObject_HEAD_INIT(x)                       \
    0,

Just looking at that, I'd say that we need to define something like TYPE_HEAD_DOT or something of the sort for the short term. Maybe you could test that.  On my gcc 10.x/python3.9 install adding a comma after TYPE_HEAD(NULL,0) breaks the compile since you end up with two sequential commas when the macro expands.

cbxbiker61:
> > Looks like the Python 2.7 build on windows is failing here:
> > ```
> > src_c/bufferproxy.c(658) : error C2059: syntax error : '.'
> > ```
> > 
> > 
> > Does it need an extra comma on the line above? The Python 2.7 build always tends to be the most fussy.
> 
> In pgcompat.h...
> 
> Python3:
> #define TYPE_HEAD(x,y) PyVarObject_HEAD_INIT(x,y)
> 
> Python2:
> #define TYPE_HEAD(x,y)
> PyObject_HEAD_INIT(x)
> 0,
> 
> Just looking at that, I'd say that we need to define something like TYPE_HEAD_DOT or something of the sort for the short term. Maybe you could test that. On my gcc 10.x/python3.9 install adding a comma after TYPE_HEAD(NULL,0) breaks the compile since you end up with two sequential commas when the macro expands.

Just looked at the source code of Python 2.7.18.  In the news file...

- PEP 3123: Provide forward compatibility with Python 3.0, while keeping
  backwards compatibility. Add Py_Refcnt, Py_Type, Py_Size, and
  PyVarObject_HEAD_INIT.

In effect we should be able to do away with the TYPE_HEAD macro and just use the standard PyVarObject_HEAD_INIT.  That'd clean up the code just a bit more anyways.

cbxbiker61:
> > > Looks like the Python 2.7 build on windows is failing here:
> > > ```
> > > src_c/bufferproxy.c(658) : error C2059: syntax error : '.'
> > > ```
> > > 
> > > 
> > > Does it need an extra comma on the line above? The Python 2.7 build always tends to be the most fussy.
> > 
> > 
> > In pgcompat.h...
> > Python3:
> > #define TYPE_HEAD(x,y) PyVarObject_HEAD_INIT(x,y)
> > Python2:
> > #define TYPE_HEAD(x,y)
> > PyObject_HEAD_INIT(x)
> > 0,
> > Just looking at that, I'd say that we need to define something like TYPE_HEAD_DOT or something of the sort for the short term. Maybe you could test that. On my gcc 10.x/python3.9 install adding a comma after TYPE_HEAD(NULL,0) breaks the compile since you end up with two sequential commas when the macro expands.
> 
> Just looked at the source code of Python 2.7.18. In the news file...
> 
>     * PEP 3123: Provide forward compatibility with Python 3.0, while keeping
>       backwards compatibility. Add Py_Refcnt, Py_Type, Py_Size, and
>       PyVarObject_HEAD_INIT.
> 
> 
> In effect we should be able to do away with the TYPE_HEAD macro and just use the standard PyVarObject_HEAD_INIT. That'd clean up the code just a bit more anyways.

Also from 2.7.18 source tree in newtypes.rst:

This is so important that we're going to pick the top of it apart still
further::

   PyVarObject_HEAD_INIT(NULL, 0)

This line is a bit of a wart; what we'd like to write is::

   PyVarObject_HEAD_INIT(&PyType_Type, 0)



End result is we should replace all TYPE_HEAD(NULL, 0) with  PyVarObject_HEAD_INIT(&PyType_Type, 0).  I'll change that in the pull request.

siggisv:
Using dot notation for initializing a struct was introduced in C99, right?
The MS compiler needed for the Windows build for python 2.7 isn't supporting much of C99 so that might be a problem until we drop support for 2.7.

cbxbiker61:
Everything but the VS9/Python2.7.13 builds fine.

I'd venture a guess that VS 9.0 used in the 2.7.x test build is not handling the dot notation properly. Can we bump up the VS version for the 2.7.x builds?  If not, maybe there's a flag for C99 mode (C99, how old is that? ;), I don't know what VS 9 does by default.  Probably also should update that env to build against Python 2.7.18 rather than 2.7.13.

cbxbiker61:
> Everything but the VS9/Python2.7.13 builds fine.
> 
> I'd venture a guess that VS 9.0 used in the 2.7.x test build is not handling the dot notation properly. Can we bump up the VS version for the 2.7.x builds? If not, maybe there's a flag for C99 mode (C99, how old is that? ;), I don't know what VS 9 does by default. Probably also should update that env to build against Python 2.7.18 rather than 2.7.13.

A discussion on python.org shows which VS versions are required for certain Python versions.  It doesn't say you can't use later versions for 2.7.  Why not use a later VS for the Python 2.7 builds?

MyreMylar:
> A discussion on python.org shows which VS versions are required for certain Python versions. It doesn't say you can't use later versions for 2.7. Why not use a later VS for the Python 2.7 builds?

So far I think we've been mostly relying on this old page for which Windows Compiler to use:

https://wiki.python.org/moin/WindowsCompilers

But I dug a bit deeper this morning. Looking at the readme from the 'PCBuild' folder in the 2.7.18 release of CPython gives us this nugget:

```
***WARNING***
Building Python 2.7 for Windows using any toolchain that doesn't link
against MSVCRT90.dll is *unsupported* as the resulting python.exe will
not be able to use precompiled extension modules that do link against
MSVCRT90.dll.
```

Which sounds bad, 2.7 Numpy will link against that .dll for example. It is less clear from there that python 2.7 compiled against 9.0 can't do the reverse, and use extension modules compiled with a later MSVC dll, but a discussion here:

https://stackoverflow.com/a/56210724

Indicates that it doesn't work and that you get problems with malloc and free at least.

There is a discussion here on Microsoft's slow adoption of C99 which sort of explains why it isn't in Visual Studio 2008/MSVC 9.0:

https://stackoverflow.com/questions/9610747/which-c99-features-are-available-in-the-ms-visual-studio-compiler/28118893

It seems C99 support was only complete with Visual Studio 2015/MSVC 14.0.


---

Overall conclusion seems to be that we can't use C99 features with Python 2.7 on windows, but they will work fine once 2.7 support is dropped and pygame is on the Python 3.5+ train. Sadly we are still waiting for various version of linux and macs to stop supporting python 2.

The next OSX release 'Big Sur' is finally dropping python 2 which should massively cut down the 2.7 user base in a couple of years. Debian is 'aiming' to remove support for python 2 in the next Bullseye version, which will then take a decent while to filter downstream and into the hands of users. Probably we should start some kind of monthly stats issue tracking downloads of pygame that use the 2.7 build where we can do that.

siggisv:
I also did some digging and found [this blog post](https://stevedower.id.au/blog/building-for-python-3-5) of [a CPython core developer](https://stevedower.id.au/about) that explains **why** we **MUST** use MSVC 9.0 to compile for Python 2.7 (although building for Python 3.5 was the main point of that post, it did start with an introduction explaining problems with compiling for previous versions).

The most important parts are:
> The CRT comes into play because it is shared between the host application and every plugin. Or rather, it assumes that it is shared. Because of the way Windows loads DLLs, if the host application and all its plugins are built with the same MSVC version and hence use the same CRT version, the state kept within that CRT would be shared.
> [...]
> However, when a plugin is built with a different CRT, this state is no longer shared. File descriptors opened by the plugin do not exist (or worse, refer to a different file) in the host, file and console buffering gets confused, error handling is no longer synchronised, memory allocated in one cannot be freed in the other and so on.

MyreMylar:
So it sounds like we can't do the dot notation unless we have a `#if PY2` section (or a `#if PY2` plus an `#ifdef WIN32`) in every .c file with the old style version as well, which I guess would be OK - and definitely useful later when we can drop Python 2 support, but probably not what was originally in mind for this PR. 

Up to you @cbxbiker61, do you want to go ahead with adding the new version with a PY2 split or just leave it and we'll try and collectively remember to revisit in the magical future time when we no longer have to care about Python 2.

siggisv:
To me it sounded like the main benefit of switching to using the dot notation was that it would be less error prone and easier to maintain than the old style.

Having to maintain both the old style and the dot notation would be the exact opposite.

cbxbiker61:
> To me it sounded like the main benefit of switching to using the dot notation was that it would be less error prone and easier to maintain than the old style.
> 
> Having to maintain both the old style and the dot notation would be the exact opposite.

Yep, totally agree.  I think monitoring usage of the python 2.7 downloads is the way to go.  We'll revisit this when it makes more sense.

cbxbiker61:
> > To me it sounded like the main benefit of switching to using the dot notation was that it would be less error prone and easier to maintain than the old style.
> > Having to maintain both the old style and the dot notation would be the exact opposite.
> 
> Yep, totally agree. I think monitoring usage of the python 2.7 downloads is the way to go. We'll revisit this when it makes more sense.

One thing that came out of this though is changing the TYPE_HEAD macro would make sense. PyVarObject_HEAD_INIT is available in Python 2.7 and 3.x.

`#if defined(MS_WINDOWS)
#define PYVAROBJECT_HEAD_INIT(a,b) PyVarObject_HEAD_INIT(0,b)
#else
#define PYVAROBJECT_HEAD_INIT(a,b) PyVarObject_HEAD_INIT(a,b)
#endif

`

MyreMylar:
I've added an issue for changing over the TYPE_HEAD macros here: #2208 - and we still have the original issue for tracking the dot notation switchover, waiting on the end of Python 2.

Issue #2098, opened 2020-09-13
mouse.get_cursor on SDL2


illume:
This PR https://github.com/pygame/pygame/pull/1955 added a unit test for mouse.get_cursor. However, it skips running it in SDL2. There's more discussion in the issue https://github.com/pygame/pygame/issues/1762

- [x] investigate if there a problem with pygame.mouse.get_cursor on SDL2?
- [ ] fix pygame.mouse.get_cursor on SDL2.
- [ ] fix unit test to run on SDL1 and SDL2.

MightyJosip:
Connected: https://github.com/pygame/pygame/issues/941

MyreMylar:
The issue (thanks to @ankith26 ) appears to be that `GetCursor()` now returns platform dependent data and is no longer interoperable with `SetCursor()` as it was in SDL1.

This is the line and comment in the SDL source code:

`typedef struct SDL_Cursor SDL_Cursor;   /**< Implementation dependent */`

I think all we could do here is store, pygame side, cursor data that _we_ set via `set_cursor()` and return that if we have it - but return None or something if `set_cursor()` has never been called.

ankith26:
Yeah I had opened a PR to try and get this working. After an hour of going through sdl source code, figured out that the data is stored in a platform dependent way (and I got no clue on the parameters of how the data is stored)

PR #2184, opened 2020-10-13
Just playing around with mouse.get_cursor

ankith26:
Related to #2098 

Just testing why get cursor fails on sdl2. According to docs, it should be available.
https://wiki.libsdl.org/SDL_GetCursor

### DIFF
--- a/src_c/mouse.c
+++ b/src_c/mouse.c
@@ -319,14 +319,6 @@ mouse_set_system_cursor(PyObject *self, PyObject *args)
     Py_RETURN_NONE;
 }
 
-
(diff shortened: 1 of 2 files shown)


ankith26:
Hmmm

ankith26:
SDL2 has a completely new way of doing SDL_Cursor AFAICT

ankith26:
After being Sherlock Holmes for about an hour, figured out that the strict SDL_Cursor is no way not having the standard height, width, andmarks, XORmarks attributes.

SDL Cursor is platform dependent now

PR #2188, opened 2020-10-15
[Draft] Make time module more precise

ankith26:
EDIT: 
NOTE: This is a draft PR, thats going to remain a draft for a while, because this brings in a lot of new changes, while pygame is preparing for a new major release. So this PR must wait.

Original comment:

This makes a lot of changes to the way pygame.time works. Mainly the changes I made add to the accuracy ot the module


ankith26:
If all works as per plan, here are the changes this PR makes:

Internal changes
1) It uses time.h over the sdl time module, using the extended accuracy provided to give float-millisecond accuracy to SOME of the functions in the time module
2) It has changed the clock class internally (but from users end it remains the same)

Usage changes
1) The function `pygame.time.delay()` now accepts the millisecond argument as an int or float (previously it was only int), this change is fully backwards compatible.
2) The `pygame.time.Clock' object now is much more accurate in the method for `tick_busy_loop` which now uses the time module to do its busy loop to give sub-millisecond accuracy. The `tick` method still depends on the SDL_Delay, so its accuracy has not changed
3) The methods `get_time` and `get_rawtime` now return floats, for extended accuracy

ankith26:
That was expected, all builds went red (except the Mac build, surprisingly)

src_c/time.c
```diff
@@ -278,10 +274,11 @@ time_set_timer(PyObject *self, PyObject *arg)
...
-    int fps_count, fps_tick;
-    float fps;
-    int timepassed, rawpassed;
+    PyObject_HEAD
+    int juststarted;
+    int fps_count;
+    clock_t last_tick, fps_tick;
+    float fps, timepassed, rawpassed;
     PyObject *rendered;
 } PyClockObject;
```
illume, line 283: I know it's not part of this PR... but if you feel like cleaning it up:

Our own pygame code should not start with Py, since that's supposed to be for python things.

src_c/time.c
```diff
@@ -309,40 +307,44 @@ clock_tick_base(PyObject *self, PyObject *arg, int use_accurate_delay)
...
         else {
             // this uses sdls delay, which can be inaccurate.
             if (delay < 0)
-                delay = 0;
+                delay = 0.0f;
 
-            Py_BEGIN_ALLOW_THREADS;
-            SDL_Delay((Uint32)delay);
-            Py_END_ALLOW_THREADS;
+            if (delay) {
```
illume, line 316: Maybe this should be `if (delay > 0.01f)`?
It would be interesting to see how often the `if(delay)` block is avoided in normal use? Is it never, or often?

src_c/time.c
ankith26: Ok I will replace that with Pg

src_c/time.c
```diff
@@ -140,24 +132,28 @@ time_get_ticks(PyObject *self)
...
+    double ticks;
     PyObject *arg0;
 
     /*for some reason PyArg_ParseTuple is puking on -1's! BLARG!*/
     if (PyTuple_Size(arg) != 1)
-        return RAISE(PyExc_ValueError, "delay requires one integer argument");
-    arg0 = PyTuple_GET_ITEM(arg, 0);
-    if (!PyInt_Check(arg0))
-        return RAISE(PyExc_TypeError, "delay requires one integer argument");
+        return RAISE(PyExc_ValueError, "delay requires only one argument");
```
illume, line 140: `delay requires one number argument`

src_c/time.c
```diff
@@ -140,24 +132,28 @@ time_get_ticks(PyObject *self)
...
+        ticks = accurate_delay(ticks);
+        return PyInt_FromLong((long)ticks);
+    } 
+    else if (PyFloat_Check(arg0)) {
+        ticks = PyFloat_AsDouble(arg0);
+        ticks = accurate_delay(ticks);
+        return PyFloat_FromDouble(ticks);
+    } 
+    else {
+        return RAISE(PyExc_TypeError, "delay must be an integer or floating point value");
```
illume, line 154: Can it require a number instead? Probably using kwargs instead of this custom argument parsing.

src_c/time.c
ankith26: Maybe that can be replaced by delay >= 1, because it’s going to be rounded to 0 if it is between 0 and 1 anyways

ankith26: I fixed that

src_c/time.c
ankith26: Changed :)

src_c/time.c
ankith26: I will try using py argparse tuple

ankith26: Done 👍

ankith26:
Note to self: this PR is getting messy with too many commits. When the time comes to merge this, create a new PR and close this one

PR #2122, opened 2020-09-27
WIP: SRC_ALPHA blit in SDL2 not compatible with SDL1, RLE bug fixes with SDL2.

illume:
For https://github.com/pygame/pygame/issues/1289

- [x] added a test for pygame 1.9.6/SDL1 alpha blit (surface_test.py@test_src_alpha_compatible)
- [x] surface_test.py@test_src_alpha_compatible passes with SDL2 and SDL1
- [ ] found bugs with RLE and surf->flags in SDL2. I asked about an issue/bug with RLE here: https://discourse.libsdl.org/t/is-there-an-rle-bug-with-sdl-setsurfacerle/28188 Need to make a more detailed bug report in the bugzilla. Note however, to support old SDL versions we will need a hack anyway :( But hey... would be good to not have to use a hack when it's fixed.
- [x] fixed surf.set_color_key(SDL_RLEACCEL) (via a hack)
- [x] fixed surf.copy() with RLE (via a hack)
- [ ] SDL_ConvertSurface removes the INPUT surf->flags SDL_RLEACCEL (todo, add hack everywhere SDL_ConvertSurface is used?)
- [ ] need to clean up the code a lot.
- [ ] currently with the PR solarwolf looks ugly again, need to fix the SDL_ConvertSurface RLE issue, and I think it'll be much better. But maybe there'll be more issues, so check it.

PR #2189, opened 2020-10-15
Making RLE flags & RLE behaviour more consistent between pygame 1 and pygame 2.

MyreMylar:
This follows on from discussion in the PR #2122 and is the fruits of time spent looking over the SDL 1.2.15 code and SDL 2 code.

My conclusions are:

**In SDL 1:**

- Providing the `SDL_RLEACCEL` flag to `SDL_SetColorKey()` immediately sets the `SDL_RLEACCELOK` flag on the surface, not the `SDL_RLEACCEL` flag.
- The `SDL_RLEACCEL` is only set on the surface itself later, after it has been blitted.

**In SDL 2:**

- The `SDL_RLEACCELOK` flag has gone, instead `SDL_SetColorKey()` now calls `SDL_SetSurfaceRLE()` which sets a private flag called `SDL_COPY_RLE_DESIRED`.
- Once again `SDL_RLEACCEL` is only set on a surface after it has been blitted.

So the pattern is roughly the same between the two, the main difference is that in SDL1 we had direct access to the `SDL_RLEACCELOK` flag on a surface amongst the other surface flags. The SDL2 flags also have less confusing names, more in line with the true purpose of the flag.

The other main difference is that RLE acceleration doesn't work on surfaces with per pixel alpha in SDL1, but it does in SDL2. 

So what this PR does is:

1. Grab the RLE tests from the PR #2122 and rejiggs them to conform to the reality of how SDL 1 works.
2. Adds a pointer casting hack to surface.c that allows pygame to read the state of the private flag SDL_COPY_RLE_DESIRED.
3. Adds a call to this new function in `get_flags()` so that it returns the state of `SDL_COPY_RLE_DESIRED` whenever `RLEACCELOK` is checked.

This should make the behaviour between pygame 1/SDL1 & pygame 2/SDL2 the same, except in the case of surfaces with per pixel alpha where I have disabled the test for SDL1.


test/surface_test.py
```diff
@@ -284,8 +309,27 @@ def test_subsurface_rle2(self):
...
+            img.set_colorkey(clear, RLEACCEL)
+            return img.convert()
+
+        pygame.display.init()
+        try:
+            window_surf = pygame.display.set_mode((640, 480))
+
+            image = pygame.image.load(example_path(os.path.join("data",
+                                                        "ship-up-boost1.png")))
+            image = optimize(image)
```
illume, line 328: Is this line duplicated on purpose?

MyreMylar: Yes, calling it twice was what caused the crash in the other PR once I stripped away all the other unessential parts of the test.

src_c/surface.c
```diff
@@ -65,6 +65,43 @@ typedef struct pg_bufferinternal_s {
     Py_ssize_t mem[6];      /* Enough memory for dim 3 shape and strides  */
 } pg_bufferinternal;
 
+/* copy of SDL Blit mapping definitions to enable pointer casting hack
+   for checking state of SDL_COPY_RLE_DESIRED flag */
+#define SDL_COPY_RLE_DESIRED        0x00001000
```
illume, line 70: How about a PG prefix? `PG_SDL_COPY_RLE_DESIRED`

test/surface_test.py
```diff
@@ -236,25 +236,51 @@ def test_fill(self):
...
         color = (250, 25, 25, 255)
         color2 = (200, 200, 250, 255)
-        sub_rect = pygame.Rect(16, 16, 16, 16)
-        s0 = pygame.Surface((32, 32), 24)
-        s1 = pygame.Surface((32, 32), 24)
-        s1.set_colorkey((255, 0, 255), pygame.RLEACCEL)
-        s0.blit(s1, (0, 0))
-        s1.fill(color)
-        self.assertTrue(s1.get_flags() & pygame.RLEACCEL)
+        # s0 = pygame.Surface((32, 32), 24)
```
illume, line 239: These commented out ones can go I guess.

test/surface_test.py
illume: If it's not supposed to be duplicated, the optimize function could be removed.

illume:
I wonder why the alpha blit test was removed?

src_c/surface.c
MyreMylar: can do 👍

test/surface_test.py
```diff
@@ -284,8 +309,27 @@ def test_subsurface_rle2(self):
...
+            clear = img.get_colorkey()
+            img.set_colorkey(clear, RLEACCEL)
+            return img.convert()
+
+        pygame.display.init()
+        try:
+            window_surf = pygame.display.set_mode((640, 480))
+
+            image = pygame.image.load(example_path(os.path.join("data",
+                                                        "ship-up-boost1.png")))
```
illume, line 327: Instead of adding this file, we could use alien1.png?

```
alien1.png: PNG image data, 80 x 71, 8-bit colormap, non-interlaced
ship-up-boost1.png: PNG image data, 128 x 32, 8-bit colormap, non-interlaced
```

MyreMylar: Sure, I was just overzealously copying from solarwolf to begin with to reproduce the crash.

MyreMylar:
> Oh, that blit change is not in here.

Yeah, I just pulled out the RLE stuff to here because it was getting quite confusing in the mega PR.

src_c/surface.c
```diff
@@ -2647,7 +2709,9 @@ surf_get_flags(PyObject *self, PyObject *args)
         flags |= PGS_SRCCOLORKEY;
     if (sdl_flags & SDL_PREALLOC)
         flags |= PGS_PREALLOC;
-    if (sdl_flags & SDL_RLEACCEL)
+    if (pg_HasSurfaceRLE(surf))
```
illume, line 2712: Should it look at sdl_flags too?
```c
if ((sdl_flags & SDL_RLEACCEL) || SDL_HasSurfaceRLE(surf))
```

test/surface_test.py
```diff
@@ -284,8 +310,28 @@ def test_subsurface_rle2(self):
...
+            return img.convert()
+
+        pygame.display.init()
+        try:
+            pygame.display.set_mode((640, 480))
+            image = pygame.image.load(example_path(os.path.join("data",
+                                                                "alien1.png")))
+            image = optimize(image)
+            image = optimize(image)
+            self.assertTrue(isinstance(image, pygame.Surface))
```
illume, line 332: I think this should also assert about the color key, and the RLEACCEL state.

MyreMylar: done.

MyreMylar:
I added two new tests.

- The first, checks the solarwolf-fixing RLE hack in `set_alpha()` which passes fine on SDL1 and (with the hack) on SDL2.
- The second identifies other differing `set_alpha()` with RLE behaviour between SDL1 & SDL2. It seems that SDL 1 is better able to always restore the pixels of a surface that have been RLE'd to their original state. In my test pixels are only 1 value away in each colour channel on SDL1 after RLE has been applied and then removed from a surface. In SDL2 however adding RLE via `set_alpha()` and then removing RLE via `set_alpha()` from a surface with a colorkey & surface alpha (but without per pixel alpha) causes the surface's color key to get used in the alpha blending of the blit calculation. So the values end up looking much different. I wonder if SDL2 is also stripping an internal colorkey flag somewhere in there.

I'll tinker with it a bit more but I think this PR will at least get solarwolf working again even if it doesn't fix all the RLE issues.

MyreMylar:
Added another test which triggers the same, blending the colorkey into the r,g & b channels issue as the last test but via this sequence:

- `surf.set_alpha(90, with RLEACCEL flag)`
- `surf.blit()`
- `surf.set_colorkey(original_key, with RLEACCEL flag)` 
- `surf.blit()`

So it can also occur even if you keep RLE active the whole time.

illume:
The `examples/data/ship-up-boost1.png` file can be removed.

illume:
Seems these are failing on SDL1... 

https://travis-ci.org/github/pygame/pygame/jobs/736055955#L4694

```python
FAIL: test_solarwolf_rle_usage (pygame.tests.surface_test.SurfaceTypeTest)
Test for error/crash when calling set_colorkey() followed
  File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/pygame-2.0.0.dev15-py3.5-linux-x86_64.egg/pygame/tests/surface_test.py", line 338, in test_solarwolf_rle_usage
    self.assertEqual(image.get_colorkey(), orig_colorkey)
AssertionError: Tuples differ: (146, 0, 0, 255) != (128, 0, 0, 255)

First differing element 0:
146
128
- (146, 0, 0, 255)
?   ^^
+ (128, 0, 0, 255)
?   ^^
```

https://travis-ci.org/github/pygame/pygame/jobs/736055955#L4714

```python
FAIL: test_solarwolf_rle_usage_2 (pygame.tests.surface_test.SurfaceTypeTest)
Test for RLE status after setting alpha
  File "/home/travis/virtualenv/python3.5.7/lib/python3.5/site-packages/pygame-2.0.0.dev15-py3.5-linux-x86_64.egg/pygame/tests/surface_test.py", line 364, in test_solarwolf_rle_usage_2
    self.assertTrue(image.get_flags() & pygame.RLEACCEL)
AssertionError: 0 is not true
```

illume (approved):
🎉🕺💃

src_c/surface.c
```diff
@@ -1609,6 +1648,26 @@ surf_set_alpha(pgSurfaceObject *self, PyObject *args)
...
+    if ((surf->flags & SDL_RLEACCEL) && (!(flags & PGS_RLEACCEL)))
+    {
+        /* hack to strip SDL_RLEACCEL flag off surface immediately when
+           it is not requested */
+        sdlrect.x = 0;
+        sdlrect.y = 0;
+        sdlrect.h = 0;
+        sdlrect.w = 0;
+
+        surface = SDL_CreateRGBSurface(0, 1, 1, 32,
```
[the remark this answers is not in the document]
MyreMylar: I also didn't know whether this would be faster.

However it probably doesn't matter too much as the hack should right now only come into play in rare-ish circumstances and likely only in some kind of loading phase of a game - as you have to be first adding RLE acceleration to a surface and then removing it. This combination of circumstances likely only happens in solarwolf by accident, or perhaps because of behaviour that may just exist because of changes in SDL and/or pygame since solarwolf was created.

PR #2198, opened 2020-10-16
Hopefully fix failing arm unittests

ankith26:


ankith26:
This PR doesn’t look like it’s helping. The issue might be due to travis in this case

ankith26:
The issue looks like it’s fixed, there is no need for this PR

PR #2199, opened 2020-10-16
draw.line fix with thick lines. More of the line was clipped off on screen borders than expected.

### LINKED ISSUE #2193: Lines across the border of the clipping area not always drawn correctly by `draw.line()`
**Environment:**
- **pygame version**: 2.0.0.dev14 

**Current behavior:**
When drawing a line that goes across the border of the clipping area of a surface, more is clipped than expected.

This image shows the result of drawing a line from the center of the screen to the position of the mouse pointer on a surface whose clipping area is shown by the red rectangle:

draw_line_test

This bug only surfaces when the border being crossed is perpendicular to the direction in which the line thickness grows.

**Expected behavior:**

The missing area shown in this picture was expected to also be drawn:

draw_line_test_notes

That was the behavior in pygame 2.0.0.dev12 as can be seen here:
draw_line_test_2

**Steps to reproduce:**

The behavior can be manually tested with:

```python
import pygame as pg

pg.init()
screen = pg.display.set_mode((400, 300))
clock = pg.time.Clock()
first_point = (200, 150)
while True:
    for event in pg.event.get():
        if event.type == pg.QUIT:
            quit()
    screen.set_clip(None)
    screen.fill("BLACK")
    screen.set_clip(pg.Rect(50, 50, 300, 200))
    screen.fill("RED")
    mouse_pos = pg.mouse.get_pos()
    pg.draw.line(screen, "WHITE", first_point, mouse_pos, 50)
    pg.display.flip()
    clock.tick(60)
```

MightyJosip:
This should close #2193

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -987,6 +987,12 @@ compare_int(const void *a, const void *b)
     return (*(const int *)a) - (*(const int *)b);
 }
 
+static int
+sign(int x, int y)
+{
+    return (x > 0) ? 1 : ((x < 0) ? -1 : y);
+}
+
 static Uint32
 get_antialiased_color(SDL_Surface *surf, int x, int y, Uint32 original_color,
                       float brightness, int blend)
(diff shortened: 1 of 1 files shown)


illume:
It tested fine on macos with the test case in #2193

MyreMylar (approved):
It works as I would expect now on Windows locally using the test case, and all the draw tests also still pass locally.

I'll approve this but, as requested, we should probably wait for the @siggisv verdict before merging.

siggisv (approved):
Looks good to me (tested on Debian Buster).  :+1: 

(It would have been nice to have some new unit tests that check that particular edge case, but I can take care of that in a separate PR.)

PR #2201, opened 2020-10-17
Add small placeholder function for future update

ankith26:
Related to #2188 

As we know, #2188 is a WIP and a draft for a future release (post v2.0.0)

The time PR is going to be adding support for using more precise clocks in many of the functions. To make it better, I’m going to add functionality to fall back to the sdl_time module if the new clock could not be used (rather than raising errors). So it makes sense that there must be some way to check whether pygame is using a clock of added accuracy.

So I have added the function (as a placeholder), and I hope that the placeholder can make it with v2.0.0 and have actual functionality implemented later. 

I will put docs, unittests etc in future versions (post v2.0.0)

My apologies, if I am just adding a bit to your work, (but it’s useful and a super small change and will not take any of your time at all)

### DIFF
--- a/src_c/time.c
+++ b/src_c/time.c
@@ -130,6 +130,12 @@ accurate_delay(int ticks)
 }
 
 static PyObject *
+time_is_preciseclock(PyObject *self)
+{
+    Py_RETURN_FALSE;
+}
+
+static PyObject *
 time_get_ticks(PyObject *self)
 {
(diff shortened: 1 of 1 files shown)


ankith26:
The idea of this PR is that code written in pygame v2.0.1 (or whichever version the time module is upgraded) should not break in v2.0.0

ankith26:
This PR is not required for now, closing it.

PR #2203, opened 2020-10-17
add android build, default data

robertpfeiffer:
Include some data files like freesansbold.ttf in the stripped dist


illume (approved):
👍 thanks :)

PR #2204, opened 2020-10-17
Skip a lot of failing tests on pypy, build pypy wheels on mac+travis.

illume:
- pypy2 on mac+travis is not working. (I'll report an issue to them)
- mac pypy builds are much longer because numpy needs to be built from source (seems also true on windows). (I'll report an issue later.)
- linux builds for pypy will need to wait until a manylinux2014 upgrade happens. Since the older manylinux does not support pypy.


MyreMylar:
Seems like a great start on getting pypy functional. 

Those rect test failures seem the weirdest. It also looks like we might need an `IS_PYPY_2` and a `IS_PYPY_3` define from your comments on those test skips.

MyreMylar (approved):

PR #2209, opened 2020-10-18
One of the math.c functions still used deprecated UTF functions

MyreMylar:
Related to #1947

Think this might also be the cause of one of the pypy test failures.


illume (approved):
👍 Nice. Good find.

PR #2211, opened 2020-10-19
Doc updates (SDL2 -> pygame 2)

ankith26:
Some doc updates. SDL2 changed to pygame 2 all over docs, along with misc fixes. I’m going to be commiting file by file.


ankith26:
This PR is done from my end. Of course, I have made only a few changes to the docs, and there might be some more changes that could be done to make the docs better. But the scope of this PR was only to replace SDL2 with pygame 2 (alongside with a bit of misc polish)

illume (approved):
👍 thank you.

PR #2212, opened 2020-10-19
pypy and manylinux2010

illume:
Continuing on from the closed PR https://github.com/pygame/pygame/pull/978 by @takluyver 

- This uses manylinux2010 to build wheels for x86 and x64.
- pypy27, pypy36, and pypy37 (which requires at least manylinux2010).
- manylinux1 still works, (use `make base-images-manylinux wheels-manylinux` instead)
- also fixed a sysfont test, and disabled a threads timing based test which intermittently fails.
- it's structured so changing to another image (like manylinux2014 or a steam runtime based one) is possible.

One downside is that Ubuntu Xenial 16.04 with the old pip on there won't work with manylinux2010 builds.


MyreMylar:
> One downside is that Ubuntu Xenial 16.04 with the old pip on there won't work with manylinux2010 builds

I assume the fix for this is just a:

`python -m pip install –upgrade pip`

?

Everything else looks fine, but hard to tell the effects until we build some wheels I suppose.

Issue #742, opened 2019-01-16
SDL2: Surfaces from Font with background crashes in SDL2


husano896:
Some tasks to do to fix this issue.

- [ ] add unit test for problem (see code below to start) *good first issue*
- [ ] determine why no error after `image.set_colorkey((0,0,0)),`
- [ ] fix issue with 255 not properly setting alpha in SDL2
- [ ] fix alpha(127) in SDL1 (should be 50%); alpha(127) in SDL2 is working properly

---
```python
    import pygame as pg

    pg.init()

    surf = pg.display.set_mode((320,240))
    font = pg.font.SysFont("Arial", 24)
    image = font.render("Test", 0, (255,255,255), (0,0,0))
    print(image.get_colorkey()) #returns None in both SDL1/SDL2.
    image.set_alpha(255)
    surf.blit(image,(0,0))
    pg.quit()
```

When the alpha of image is set to 255 (0~254 is fine) , it raised an error:

> pygame.error: Blit combination not supported

works fine in SDL1.

But will not raise error after image.set_colorkey((0,0,0)),

Also, 
```python
    image = font.render("Test", 0, (255,255,255))
    image.set_alpha(127)
```
take no effect in SDL1 (no transparent changes) but SDL2 (shows half transparent image)

**Related Docs:** https://www.pygame.org/docs/ref/surface.html

markph0204:
- [ ] fix issue with 255 not properly setting alpha in SDL2
- [ ] determine why no error after `image.set_colorkey((0,0,0)),`
- [ ] fix alpha(127) in SDL1 (should be 50%); alpha(127) in SDL2 is working properly

@illume can you validate this checklist?

illume:
Thanks @markph0204 I added a check list of tasks to the top.

nthykier:
Having dumpster dived into the SDL2 code, it appears that there are the following 3 cases where bilt can throw that error (AFAICT):

1. SDL thinks that there are less than 8 bits per pixel in the destination (`dst->format->BitsPerPixel < 8`)  
1. SDL thinks that source OR destination has more than 8 bits per pixel (`...->format->Rloss > 8`)
1. SDL fails to locate an interal blit function (not sure it can happen as it appears to always fallback to a "Slow blit" except in test cases)

[...]

PR #2213, opened 2020-10-19
First pass at backwards compatible pygame alpha blit

MyreMylar:
This adds an SDL1 style blitter for alpha blending where both source and destination surfaces have an alpha channel. It also adds a flag to disable this blitter in favour of reverting to the SDL2 one.

To Do
-------

- [x] ~~Currently we are only handling ARGB surfaces in the SSE2 version, can we tweak this for other 32 bit surface formats e.g. RGBA. Do we need to?~~ Looks like SDL only handles 'ARGB/ABGR' formats in my tests, basically where the alpha is at shift 24, though whether to call this ARGB or RGBA with big endian and little endian seems to be a minefield in itself. The current code should handle any 32bit surfaces where the alpha is shifted by 24bits anyway.
- [x] 16 bit surfaces - ~~is it worth doing anything for these? They should be handled by the slow, non-sse2, blitter but we aren't testing that right now.~~ These are now handled by the non-sse2 blitter & tested. It is about half the speed of the SDL2 16 bit surface blitter, if you care about speed over accuracy in 16bit blits you'll want to use the SDL2 blitter.
- [x] ~~blitting surface with alpha to surface without and visa versa - does this need handling or do the SDL1 & SDL2 blitters behave the same?~~ Investigated this, it appears that SDL1 goes down a different path when the destination surface is opaque so the new blitter path doesn't match the SDL1 output. I put these blits down the new path anyway because it looks more accurate to SDL1 than the SDL2 blitter. 
- [x] Check 'surface alpha' (set with set_alpha()). Also, we don't want to get involved in RLE stuff accidentally.
- [x] Split SSE2 blitter into multiple blitters to see if we can improve speed in most common surface setup versus rarer cases.
- [x] Add test for blit surface with per pixel alpha and surface alpha - this one will be different from normal SDL2 and is not a possible combo in SDL1, so will have to use carefully selected values. Should be SDL1 blending formula with SDL2 alpha capabilities.  
- [x] Check blitter against original test cases in #1289 
- [x] Add environment variable to makes SDL2 blitter the default instead of SDL1-mimic blitter.
- [x] Investigate splitting SSE2 blitter a third time to handle common case where there is no alpha on destination surface.


MyreMylar:
Some speed tests:

new-default SDL1-style alpha-blit (using SSE2):

```
         256002 function calls in 1.854 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.166    0.166    1.854    1.854 pygame_blit_sdl1_speed_test.py:9(speed_test_blits)
   128000    1.055    0.000    1.055    0.000 {method 'fill' of 'pygame.Surface' objects}
    64000    0.610    0.000    0.610    0.000 {method 'blit' of 'pygame.Surface' objects}
    64000    0.023    0.000    0.023    0.000 {method 'get_at' of 'pygame.Surface' objects}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

old-default SDL2 style blit (uses some form of intrinsics in SDL2):

```
         256002 function calls in 1.832 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.185    0.185    1.832    1.832 pygame_blit_sdl2_speed_test.py:10(speed_test_blits)
   128000    1.054    0.000    1.054    0.000 {method 'fill' of 'pygame.Surface' objects}
    64000    0.568    0.000    0.568    0.000 {method 'blit' of 'pygame.Surface' objects}
    64000    0.025    0.000    0.025    0.000 {method 'get_at' of 'pygame.Surface' objects}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

new-default SDL1 style alpha-blit (without any SSE2/NEON intrinsics)
```
         256002 function calls in 6.070 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.172    0.172    6.070    6.070 pygame_blit_sdl1_speed_test.py:9(speed_test_blits)
    64000    4.836    0.000    4.836    0.000 {method 'blit' of 'pygame.Surface' objects}
   128000    1.038    0.000    1.038    0.000 {method 'fill' of 'pygame.Surface' objects}
    64000    0.024    0.000    0.024    0.000 {method 'get_at' of 'pygame.Surface' objects}
        1    0.000    0.000    0.000    0.000 {method 'disable' of '_lsprof.Profiler' objects}
```

So the new SDL1 blitter is fairly comparable, though slightly slower, than the default SDL2 one and both of them are a lot faster than not using intrinsics. 

It's possible we could improve the speed a bit without affecting the accuracy, I'm not an SSE2 intrinsics expert.

MyreMylar:
Was using this to do speed testing (changing the flag to switch to the SDL2 blitter):

```
from sys import stdout
from pstats import Stats
from cProfile import Profile

import pygame
pygame.init()


def speed_test_blits():

    nums = [0, 1, 65, 126, 127, 199, 254, 255]
    results = {}
    for iterations in range(0, 1000):
        for dst_r, dst_b, dst_a in zip(nums, reversed(nums), reversed(nums)):
            for src_r, src_b, src_a in zip(nums, reversed(nums), nums):
                src_surf = pygame.Surface((66, 66), pygame.SRCALPHA, 32)
                src_surf.fill((src_r, 255, src_b, src_a))
                dest_surf = pygame.Surface((66, 66), pygame.SRCALPHA, 32)
                dest_surf.fill((dst_r, 255, dst_b, dst_a))

                dest_surf.blit(src_surf, (0, 0))
                key = ((dst_r, dst_b, dst_a), (src_r, src_b, src_a))
                results[key] = dest_surf.get_at((65, 33))


if __name__ == '__main__':
    profiler = Profile()
    profiler.runcall(speed_test_blits)
    stats = Stats(profiler, stream=stdout)
    stats.strip_dirs()
    stats.sort_stats('cumulative')
    stats.print_stats()
```

MyreMylar:
Checked over the test cases in the original report (#1289) and they all look/print like they do in SDL1 now. So that's nice.

MyreMylar:
With this PR you should now be able to do:

```
import os
import pygame
os.environ['PYGAME_BLEND_ALPHA_SDL2'] = '1'

pygame.init()
```

To set the blitter back to how it worked before this PR.

MyreMylar:
Ok, that is probably enough fiddling. To summarise:

- This adds a new default blitter for surfaces with alpha. The default blitter prioritises accuracy to what SDL1 did over speed.
- There are SSE2 optimised versions of the blitter that cover three common cases:
    - Two surfaces with per-pixel alpha and no surface alpha.
    - A source surface with per pixel alpha _and_ surface alpha, blitted to a destination surface with regular pixel alpha.
    - A source surface with per-pixel alpha blitted to an opaque surface.
- There is a regular non-intrinsic alpha blitter (that has been in pygame for ages) that is used for any cases directed to the alpha blitter outside of the three above.
- The blitter doesn't handle colorkey alpha, that is passed over to SDL2 as before.
- The blitter doesn't handle blits with RLE encoded surface, they are is also passed over to SDL2.
- You can switch from this new default blitter to SDL2 by passing the flag `BLEND_ALPHA_SDL2` when blitting, or by setting the environment variable `PYGAME_BLEND_ALPHA_SDL2` before calling `pygame.init()`.

This should resolve all the cases raised in #1289 and maybe will help with some others in the tracker, I haven't checked through carefully yet.

MyreMylar:
Speed wise, in my tests (rough code posted above):

 - slightly slower than SD2 blitter when doing pixel alpha surface to pixel alpha surface.
 - equivalent or slightly faster when blitting with a surface with pixel alpha & surface alpha (from set_alpha()) or to an opaque surface.

I don't have any particular ideas for improving that right now.

illume:
Nice work :) This also fixes the issue in https://github.com/pygame/pygame/issues/742 

This example used to error with:
> pygame.error: Blit combination not supported

```python
import pygame as pg
pg.init()
surf = pg.display.set_mode((320,240))
font = pg.font.SysFont("Arial", 24)
image = font.render("Test", 0, (255,255,255), (0,0,0))
print(image.get_colorkey()) #returns None in both SDL1/SDL2.
image.set_alpha(255)
surf.blit(image,(0,0))
```

But now it doesn't.

MyreMylar:
Looks like the arm CI build doesn't use the SSE2 optimised path at all, so the test issue was likely coming from a lack of support for surface_alpha() in the non-sse2 version of the alpha blend which I've now changed.

illume (approved):
```
_/=====\_
| great! |
---------\
          `_  
             🦜 
```

MyreMylar:
Hopefully this doesn't cause chaos in everyone's apps, but I'm bracing for it just in case.

illume:
The unit tests pass locally on Mac.

I see there's a difference in solarwolf:

Screenshot 2020-10-22 at 18 44 26

In the screen where you can load a save game.

illume (changes requested):
There's something weird going on with solarwolf.

MyreMylar:
This is my current solarwolf related issue with this PR on windows:

```
import pygame

pygame.init()


def textshadowed(color, text, center=None, pos='center'):
    font = pygame.font.Font(None, 32)
    darkcolor = [int(c // 2) for c in color]
    if text is None: text = ' '
    try:
        img1 = img2 = font.render(text, 1, color)
        img2 = font.render(text, 1, darkcolor)
    except (pygame.error, TypeError):
        img1 = img2 = pygame.Surface((10, 10))

    newsize = img1.get_width() + 2, img1.get_height() + 2
    img = pygame.Surface(newsize)

    print("start:")
    print("img.get_flags():", img.get_flags())
    print("img.get_at((0, 0)):", img.get_at((0, 0)))
    print("img.get_colorkey():", img.get_colorkey())
    print("img.get_shifts():", img.get_shifts())
    print("img.get_masks():", img.get_masks())
    print("img.get_bitsize():", img.get_bitsize())
    print("img.get_bitsize():", img.get_bytesize())

    img.blit(img1, (0, 0))
    img = img.convert()

    print("\nafter blit:")
    print("img.get_flags():", img.get_flags())
    print("img.get_at((0, 0)):", img.get_at((0, 0)))
    print("img.get_colorkey():", img.get_colorkey())
    print("img.get_shifts():", img.get_shifts())
    print("img.get_masks():", img.get_masks())
    print("img.get_bitsize():", img.get_bitsize())
    print("img.get_bitsize():", img.get_bytesize())

    img.set_colorkey((0, 0, 0), pygame.RLEACCEL)

    print("\npost set_colorkey:")
    print("img.get_flags():", img.get_flags())
    print("img.get_at((0, 0)):", img.get_at((0, 0)))
    print("img.get_colorkey():", img.get_colorkey())
    print("img.get_shifts():", img.get_shifts())
    print("img.get_masks():", img.get_masks())
    print("img.get_bitsize():", img.get_bitsize())
    print("img.get_bitsize():", img.get_bytesize())

    return img


screen = pygame.display.set_mode((320, 240))
screen.fill((255,255,255))

text_render = textshadowed((160, 200, 250), 'A', (190, 170), "midright")
screen.blit(text_render, (50, 50))

running = True
while running:
        if event.type == pygame.QUIT:
            running = False

    pygame.display.update()

```

Output pre-PR:
image

Output post-PR:
image

So far I can't figure out why the colorkey isn't working after using the new blitter before it is set.

illume:
With the current commit it looks like your "Output pre-PR" one to me:
Screenshot 2020-10-23 at 14 41 37

PR #2216, opened 2020-10-20
Just playing around with (PyCFunction) cast

ankith26:
Now I don’t know much techincal details of this, since I’m a newbie, but I felt like we have to apply the (PyCFunction) cast to the functions that do not accept any arguments. According to python docs https://docs.python.org/3/c-api/structures.html?highlight=meth_noargs#METH_NOARGS

If applying the type casts do not hurt, then we might as well use it as highlighted in the docs.

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -2081,31 +2081,32 @@ pg_event_custom_type(PyObject *self, PyObject *args)
 }
 
 static PyMethodDef _event_methods[] = {
-    {"__PYGAMEinit__", pgEvent_AutoInit, METH_NOARGS,
+    {"__PYGAMEinit__", (PyCFunction)pgEvent_AutoInit, METH_NOARGS,
      "auto initialize for event module"},
 #if IS_SDLv2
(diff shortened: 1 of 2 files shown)


illume:
You're correct.

Looking at our code we have other casting mistakes. Including with METH_KEYWORDS where should be PyCFunctionWithKeywords and not PyCFunction. However, I'm not sure how casting the function pointer makes any difference. Are there compiler warnings? I don't think there are any runtime differences.

I'm not sure if there's a way to type check in C that the functions are all using the correct type signature. At least not with compilers... maybe some other tools can do it.

ankith26:
As I said, I’m a bit new to C, but from what I have observed recently, not using the (PyCFunction) cast on the functions that accept no arguments tends to give occasional weird errors. And using PyCFunction on keyword functions is ok (I guess)

ankith26:
Ok, from what I can tell, as long as a function written in C accepts two arguments, self and arg, the cast is not needed. But the cast becomes compulsory otherwise. 

This PR does not look like it’s doing something’s important, you cannot fix things that are not broken, so I’m closing this.

PR #2218, opened 2020-10-21
Probably fix an issue in travis yml file

ankith26:
I dont know much about .yml scripting, but I was going through the file and the last bit looked wrong, so I guess i'm fixin it


illume (approved):
👍

PR #2219, opened 2020-10-21
Update manifest to include readme file (PR needs priority)

ankith26:
The readme file is important and must be included in the sdist. Without this PR, all attempts to build pygame from source (source installed from PyPI) are going to fail immediately (as discussed on discord)


illume (approved):
Thanks 👍

PR #2225, opened 2020-10-21
Added test_load_basic

dtomci:
Hello, so here is the test for load_basic. Hopefully its somewhat correct and what you expected.

### DIFF
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -788,15 +788,52 @@ class ImageModuleTest(unittest.TestCase):
             for item in version:
                 self.assertIsInstance(item, expected_item_type)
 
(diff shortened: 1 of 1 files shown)


MyreMylar:
It should be OK to actually call `load_basic()` in this test for now, we are discouraging the use of `load_basic()` in the documentation but for the unit test it makes sense to call it directly. I know `load()` redirects to `load_basic()` but that's why we have tests for `load()` too :)

Other things that could be tried is loading an actual file from the examples data directory e.g.:

`load_basic(example_path("data/asprite.bmp"))`

and perhaps loading from a file-like object like `io.BytesIO`?

Issue #1732, opened 2020-05-14
pygame.image docs missing functions and load_basic ignore namehint argument


MightyJosip:
https://github.com/pygame/pygame/blob/ea4a028100776774b8b7431bbd8a7c4d1005e00f/src_c/image.c#L62 This can be seen in the function above. name variable gets parsed, but never used. Also there are missing functions in docs: ('load_basic', 'load_extended', 'save_extended')

**Related Docs:** https://www.pygame.org/docs/ref/image.html

**To Do to Close**

- [ ] Doc string for function 'load_basic'.
- [ ] Doc string for function 'load_extended'.
- [ ] Doc string for function 'save_extended'
- [x] Regular .rst docs for these functions.
- [x] Investigate unused namehint argument for load_basic.

ankith26:
I guess I have answer to the last two questions. In the case of pygame.image.load_basic(), the namehint argument serves no purpose because the function is for loading BMP images anyways, it does not need the help of namehint argument to determine what kind of image it needs to load.

But it is still required nonetheless because when SDL-image module is not used while building pygame, pygame.image.load() is going to fall back to using pygame.image.load_basic(), and pygame.image.load() takes the optional name-hint argument, so even load_basic() must take the optional argument.

ankith26:
And yes, I want to be documenting all the other functions (and maybe some minor fixes to pygame.image)

siggisv:
In my opinion, @ankith26 is right that there is no need to document `load_basic()`, `load_extended()` and `save_extended()`.

Furthermore, I feel that not only should they not be documented, the should be renamed to `_load_basic()`, `_load_extended()` and `_save_extended()` to make it clearer that they are helper functions.

Regarding the unused name-hint argument, it should be easy to change `load_basic()` to check it and return error if it is not ending with "BMP" (or "bmp").

MyreMylar:
I agree that `load_basic()`, `load_extended()` and` save_extended()` _should_ be renamed with the preceding underscore so they don't appear in the public interface...

However, a quick GitHub search reveals that there are already lots of people using them despite them being undocumented. See:

https://github.com/search?q=pygame.image.load_basic%28&type=code
https://github.com/search?q=pygame.image.load_extended%28&type=code

Which is the perpetual curse of working with pygame's age and large userbase.

siggisv:
Well, that is a problem.

Maybe we could:

[...]

PR #2226, opened 2020-10-22
Some updates to image module

ankith26:
Just some updates to image module. Working on #1732, and some other minor changes

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -371,7 +371,7 @@ image_get_sdl_image_version(PyObject *self)
         return PyObject_CallObject(extverobj, NULL);
 }
 
-#if PG_COMPILE_SSE4_2 && (SDL_VERSION_ATLEAST(2, 0
+#if PG_COMPILE_SSE4_2 && (SDL_VERSION_ATLEAST(2, 0, 0))
 #define SSE42_ALIGN_NEEDED 16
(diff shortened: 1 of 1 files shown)


src_c/image.c
```diff
@@ -64,6 +53,15 @@ opengltosdl(void);
     ((flipped) ? (((char *)data) + (height - row - 1) * width) \
                : (((char *)data) + row * width))
 
+static PyObject *extloadobj = NULL;
+static PyObject *extsaveobj = NULL;
+static PyObject *extverobj = NULL;
+
+/* define docs for undocumented (kinda deprecated) functions */
+#define DOC_PYGAMEIMAGELOADBASIC "internal function to load images, use pygame.image.load() instead of this"
```
MightyJosip, line 61: This docstrings should be in this file: https://github.com/pygame/pygame/blob/master/src_c/doc/image_doc.h

MightyJosip (changes requested):

src_c/image.c
ankith26: Yeah, but these are temporary-undocumented functions, they cannot go in that file right.

ankith26: Plus those files are auto generated from the rest files, I guess they will dissapear the next time, those files get generated.

PR #2242, opened 2020-10-24
Do MOUSEBUTTONUP event on MOUSEWHEEL

### LINKED ISSUE #2240: pygame2 does not generate MOUSEBUTTONUP events for mouse wheel
**Environment:**

- **Python version**: 3.7.4
- **SDL version**: 2.0.12
- **PyGame version**: 2.0.0.dev22
- **Relevant hardware**: logitech wired optical mouse (basic model)

**Current behavior:**

Rolling the mouse wheel generates MOUSEBUTTONDOWN events, but no MOUSEBUTTONUP events. It does not matter if the mouse is moving or still.

**Expected behavior:**

The same behavior as pygame 1.9.6 is expected: it generates MOUSEBUTTONUP events.

**Screenshots**

NA

**Steps to reproduce:**

1. Run the provided code in pygame 1.9.6.
2. Scroll the mouse wheel up and down.
3. See the MOUSEBUTTONUP events are printed.
4. Run the provided code in pygame 2.0.0dev22 (or previous pygame2 version).
5. Scroll the mouse wheel up and down.
6. See the MOUSEBUTTONUP events are not printed.

**Test code**

If possible add a simple test program that shows the problem described in this report.
```
import pygame
from pygame.locals import KEYDOWN, K_ESCAPE, QUIT, MOUSEBUTTONUP, MOUSEBUTTONDOWN
pygame.init()
clock = pygame.time.Clock()
screen = pygame.display.set_mode((400, 400))
while 1:
    clock.tick(30)
    for e in pygame.event.get():
        if e.type == KEYDOWN and e.key == K_ESCAPE:
                quit()
        elif e.type == QUIT:
            quit()
        elif e.type == MOUSEBUTTONDOWN:
            print('Mouse button down: {}'.format(e.button))
        elif e.type == MOUSEBUTTONUP:
            print('Mouse button up: {}'.format(e.button))
```

```
paste other relevent logs or stack traces here, if applicable
```

ankith26:
Probably fixes #2240

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -265,40 +265,66 @@ pg_event_filter(void *_, SDL_Event *event)
         }
     }
     else if (type == SDL_MOUSEWHEEL) {
-        SDL_Event newevent;
+        SDL_Event newdownevent, newupevent;
         int x, y;
 
         if (event->wheel.x == 0 && event->wheel.y == 0) {
             //#691 We are not moving wheel!
             return 1;
         }
+        SDL_GetMouseState(&x, &y);
+        
         // Generate a MouseButtonDown event for compatibility.
         // https://wiki.libsdl.org/SDL_MouseWheelEvent
-        newevent.type = SDL_MOUSEBUTTONDOWN;
+        newdownevent.type = SDL_MOUSEBUTTONDOWN;
+        newdownevent.button.x = x;
+        newdownevent.button.y = y;
 
-        SDL_GetMouseState(&x, &y);
-        newevent.button.x = x;
-        newevent.button.y = y;
-
-        newevent.button.state = SDL_PRESSED;
-        newevent.button.clicks = 1;
+        newdownevent.button.state = SDL_PRESSED;
+        newdownevent.button.clicks = 1;
 
         if (event->wheel.y != 0) {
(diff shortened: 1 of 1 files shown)


ankith26:
Tested this PR locally, and it works as expected

ankith26:
Overall, this PR makes many changes to ensure that output is much more similar to pygame 1.9.6

src_c/event.c
```diff
@@ -157,6 +157,14 @@ _pg_remove_pending_PGS_VIDEOEXPOSE(void * userdata, SDL_Event *event)
     return 1;
 }
 
+static int
```
siggisv, line 160: Small note: C already has a function, `int abs(int n)`, defined in `<stdlib.h>`, that does exactly what this function is doing (see: https://en.cppreference.com/w/c/numeric/math/abs).

ankith26: Hmmm ok, I didn’t know it was there. Anyways, we can save an include by defining the function ourselves right ;)

illume: There is also SDL_abs.

illume (approved):
👍 thanks. Seems to work fine on my mac.

Issue #1618, opened 2020-04-04
Slow blitting with combined pixel and surface alpha


mcpalmer1980:
I recently discovered that surfaces can use both pixel and surface alpha in pygame2dev6. Curious about performance, I wrote a quick test. My results were somewhat surprising. When combined, the two alphas blit four times slower than either one alone.

Blubberquark suggested I submit an issue, presumably to remind him to take a look at some point. Feel free to close this at any time. This combination wasn't supported in previous releases, and future users hoping for fancy effects with pygame2 will likely use the renderer instead of software surfaces.

```
Test 0: 1000 blits of image size 1333, 1600
tested Blit no alpha: 68.406ms
tested Blit surface alpha: 464.457ms
tested Blit pixel alpha: 381.999ms
tested Blit pixel and surface alpha: 1925.622ms

Test 1: 1000 blits of image size 1333, 1600
tested Blit no alpha: 63.964ms
tested Blit surface alpha: 451.907ms
tested Blit pixel alpha: 380.308ms
tested Blit pixel and surface alpha: 1929.704ms

Test 2: 1000 blits of image size 1333, 1600
tested Blit no alpha: 64.182ms
tested Blit surface alpha: 457.898ms
tested Blit pixel alpha: 375.786ms
tested Blit pixel and surface alpha: 1927.906ms
```

Here's the code I used

```
import time
import pygame

filename = 'flyboy.png'

pygame.display.init()
screen = pygame.display.set_mode((720, 480))
ino_alpha = pygame.image.load(filename).convert()
isurf_alpha = pygame.image.load(filename).convert()
isurf_alpha.set_alpha(100)
ipixel_alpha = pygame.image.load(filename).convert_alpha()
idual_alpha = pygame.image.load(filename).convert_alpha()
idual_alpha.set_alpha(100)

def test(methods, tries = 1000):
	for method in methods:
		start = time.time()
		name = method(tries)
		results = (time.time() - start) * 1000
		print('tested {}: {}ms'.format(name, round(results, 3)))

def no_alpha(tries):
	for t in range(tries):
		screen.blit(ino_alpha, (10,10))
	return 'Blit no alpha'

def surf_alpha(tries):
	for t in range(tries):
		screen.blit(isurf_alpha, (10,10))
	return  'Blit surface alpha'

def pixel_alpha(tries):
	for t in range(tries):
		screen.blit(ipixel_alpha, (10,10))
	return  'Blit pixel alpha'

def dual_alpha(tries):
[...]

PR #2243, opened 2020-10-24
New Alpha blitter - fixes and speed ups

MyreMylar:
Noticed that some combos of surfaces were wrongly being redirected down the slow path of the new blitter when there was no need to.

This should fix that with a better test on who gets to enter the new blitter thunderdome.

Noticed it when testing:
https://github.com/pygame/pygame/issues/1618

New blitter still slightly slower in alpha to alpha case after this PR (same on other two as it's no longer being used) but is actually better in alpha+surf_alpha case.

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -2694,23 +2694,29 @@ alphablit_alpha_sse2_argb_no_surf_alpha (SDL_BlitInfo * info)
         3. One pixel at a time.
     */
 
+    /* 64 bit variables */
+    Uint64          *srcp64 = (Uint64 *)info->s_pixels;
+    Uint64          *dstp64 = (Uint64 *)info->d_pixels;
+    Uint64          src_amask64 = ((Uint64)srcfmt->Amask << 32) | srcfmt->Amask;
+
+    Uint64          rgb_mask = 0x00FFFFFF00FFFFFF;
+    Uint64          offset_rgb_mask = 0xFF00FFFFFF00FFFF;
+
+    /* 32 bit variables */
+    Uint32          *srcp32 = (Uint32 *)info->s_pixels;
+    Uint32          *dstp32 = (Uint32 *)info->d_pixels;
+    Uint32          src_amask32 = srcfmt->Amask;
+    Uint32          dst_amask32 = dstfmt->Amask;
+
+    __m128i src1, dst1, temp, sub_dst;
+    __m128i temp2, mm_src_alpha, mm_dst_alpha, mm_sub_alpha;
+    __m128i mm_alpha_mask, mm_zero, rgb_mask_128, offset_rgb_mask_128, alpha_mask_128;
+
     if (((width % 2) == 0) && ((srcskip % 2) == 0) && ((dstskip % 2) == 0))
     {
         width = width/2;
         srcskip = srcskip/2;
         dstskip = dstskip/2;
-
-        Uint64          *srcp = (Uint64 *)info->s_pixels;
-        Uint64          *dstp = (Uint64 *)info->d_pixels;
-
-        Uint64          src_amask = ((Uint64)srcfmt->Amask << 32) | srcfmt->Amask;
-        Uint64          dst_amask = ((Uint64)dstfmt->Amask << 32) | dstfmt->Amask;
-
-        Uint64          rgb_mask = 0x00FFFFFF00FFFFFF;
-        Uint64          offset_rgb_mask = 0xFF00FFFFFF00FFFF;
-
-        __m128i src1, dst1, temp, temp2, mm_src_alpha, mm_dst_alpha;
-        __m128i mm_alpha_mask, mm_zero, rgb_mask_128, offset_rgb_mask_128, alpha_mask_128;
 
         mm_zero = _mm_setzero_si128();
         mm_alpha_mask = _mm_cvtsi32_si128(0x00FF00FF);
@@ -2721,7 +2727,7 @@ alphablit_alpha_sse2_argb_no_surf_alpha (SDL_BlitInfo * info)
            that are not available in 32 bit */
         rgb_mask_128 = _mm_cvtsi64_si128(rgb_mask);
         offset_rgb_mask_128 = _mm_cvtsi64_si128(offset_rgb_mask);
-        alpha_mask_128 = _mm_cvtsi64_si128(src_amask);
+        alpha_mask_128 = _mm_cvtsi64_si128(src_amask64);
 
         while (height--)
         {
@@ -2730,9 +2736,9 @@ alphablit_alpha_sse2_argb_no_surf_alpha (SDL_BlitInfo * info)
 
                 /* load the pixels into SSE registers */
                 /* src(ARGB) -> src1 (00000000ARGBARGB) */
-                src1 = _mm_cvtsi64_si128(*srcp);
+                src1 = _mm_cvtsi64_si128(*srcp64);
                 /* dst(ARGB) -> dst1 (00000000ARGBARGB) */
-                dst1 = _mm_cvtsi64_si128(*dstp);
+                dst1 = _mm_cvtsi64_si128(*dstp64);
                 /* src_alpha -> mm_src_alpha (00000000A000A000) */
                 mm_src_alpha = _mm_and_si128(src1, alpha_mask_128);
                 /* dst_alpha -> mm_dst_alpha (00000000A000A000) */
@@ -2806,13 +2812,13 @@ alphablit_alpha_sse2_argb_no_surf_alpha (SDL_BlitInfo * info)
                 temp = _mm_and_si128(temp, rgb_mask_128);
                 /* add alpha to RGB */
                 temp = _mm_add_epi16(mm_dst_alpha, temp);
-                *dstp = _mm_cvtsi128_si64(temp);
+                *dstp64 = _mm_cvtsi128_si64(temp);
 
-                ++srcp;
-                ++dstp;
+                ++srcp64;
+                ++dstp64;
             }, n, width);
-            srcp += srcskip;
-            dstp += dstskip;
+            srcp64 += srcskip;
+            dstp64 += dstskip;
         }
 #else /* 32 bit 2 pixel path */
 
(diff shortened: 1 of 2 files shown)


MyreMylar:
I have a fix for the failing test locally, but I also have some speed up ideas for the blitter that seem to be panning out OK. So I'll probably wrap them up together.

MyreMylar:
OK, Time for some speed/performance chat, I'll do a couple of comments. First up here is the latest test case I'm using:

```
import time
import os
import pygame

#os.environ['PYGAME_BLEND_ALPHA_SDL2'] = '1'

filename = 'flyboy.png'

pygame.init()
screen = pygame.display.set_mode((720, 480))
pixel_screen = pygame.Surface((720, 480), flags=pygame.SRCALPHA)
pixel_screen.fill((200, 200, 200, 200))
ino_alpha = pygame.image.load(filename).convert()
isurf_alpha = pygame.image.load(filename).convert()
isurf_alpha.set_alpha(100)
ipixel_alpha = pygame.image.load(filename).convert_alpha()
idual_alpha = pygame.image.load(filename).convert_alpha()
idual_alpha.set_alpha(100)

def test(methods, tries = 1000):
	for method in methods:
		start = time.time()
		name = method(tries)
		results = (time.time() - start) * 1000
		print('tested {}: \t\t\t{}ms'.format(name, round(results, 3)))

def no_alpha(tries):
	for t in range(tries):
		screen.blit(ino_alpha, (10,10))
	return 'Blit no alpha\t\t\t'

def surf_alpha(tries):
	for t in range(tries):
		screen.blit(isurf_alpha, (10,10))
	return  'Blit surface alpha\t\t'

def pixel_alpha(tries):
	for t in range(tries):
		screen.blit(ipixel_alpha, (10,10))
	return  'Blit pixel to opaque alpha'

def pixel_to_pixel_alpha(tries):
	for t in range(tries):
		pixel_screen.blit(ipixel_alpha, (10, 10))
	return  'Blit pixel to pixel alpha'

def dual_alpha(tries):
	for t in range(tries):
		screen.blit(idual_alpha, (10,10))
	return 'Blit pixel and surf alpha'

def odd_no_alpha(tries):
	for t in range(tries):
		screen.blit(ino_alpha, (11,10))
	return 'Blit no alpha\t\t\t'

def odd_surf_alpha(tries):
	for t in range(tries):
		screen.blit(isurf_alpha, (11,10))
	return  'Blit surface alpha\t\t'

def odd_pixel_alpha(tries):
	for t in range(tries):
		screen.blit(ipixel_alpha, (11,10))
	return  'Blit pixel to opaque alpha'

def odd_pixel_to_pixel_alpha(tries):
	for t in range(tries):
		pixel_screen.blit(ipixel_alpha, (11, 10))
	return  'Blit pixel to pixel alpha'

def odd_dual_alpha(tries):
	for t in range(tries):
		screen.blit(idual_alpha, (11,10))
	return 'Blit pixel and surf alpha'


for i in range(1):
	tries = 1000
	w, h = ino_alpha.get_size()
	print(f'\nTest {i}: {tries} blits of image size {w}, {h}')
	print('\nEven blit width:\n-----------')
	test((no_alpha, surf_alpha, pixel_alpha, pixel_to_pixel_alpha, dual_alpha), tries)
	print('\nOdd blit width:\n-----------')
	test((odd_no_alpha, odd_surf_alpha, odd_pixel_alpha, odd_pixel_to_pixel_alpha, odd_dual_alpha), tries)
```

Based off the one in #1618 .

It is split into even and odd sized blits because that became relevant with how I was able to improve the speed of the blitter. I suspect that the majority of blits are probably of an even width - over half anyway.

MyreMylar:
First up here's the output from the test above running 2.0.0.dev22:

```
Even blit width:
-----------
tested Blit no alpha: 			               4026.666ms
tested Blit surface alpha: 			       4016.958ms
tested Blit pixel to opaque alpha: 			465.423ms
tested Blit pixel to pixel alpha: 			712.182ms
tested Blit pixel and surf alpha: 			864.756ms

Odd blit width:
-----------
tested Blit no alpha: 			               4008.666ms
tested Blit surface alpha: 			       4009.51ms
tested Blit pixel to opaque alpha: 			463.421ms
tested Blit pixel to pixel alpha: 			713.648ms
tested Blit pixel and surf alpha: 			864.785ms
```

And here's how it looks before the new blitter was merged in 2.0.0.dev20, when SDL2 was running everything:

```
Even blit width:
-----------
tested Blit no alpha: 			                 43.038ms
tested Blit surface alpha: 			        213.203ms
tested Blit pixel to opaque alpha: 			314.789ms
tested Blit pixel to pixel alpha: 			314.285ms
tested Blit pixel and surf alpha: 			1548.582ms

Odd blit width:
-----------
tested Blit no alpha	: 		                 51.872ms
tested Blit surface alpha: 		                206.187ms
tested Blit pixel to opaque alpha: 			309.281ms
tested Blit pixel to pixel alpha: 			308.282ms
tested Blit pixel and surf alpha: 			1528.916ms
```

Oops, we lost quite a bit of performance there and the opaque-to-opaque blits are not supposed to be running through the new blitter anyway.

MyreMylar:
For comparison here's the same test running in 1.9.6:

```
Even blit width:
-----------
tested Blit no alpha			: 		79.072ms
tested Blit surface alpha		: 		255.232ms
tested Blit pixel to opaque alpha: 			431.408ms
tested Blit pixel to pixel alpha: 			3517.745ms
*tested Blit pixel and surf alpha: 			432.414ms

Odd blit width:
-----------
tested Blit no alpha			: 		82.088ms
tested Blit surface alpha		: 		249.358ms
tested Blit pixel to opaque alpha: 			406.369ms
tested Blit pixel to pixel alpha: 			3499.715ms
*tested Blit pixel and surf alpha: 			407.004ms
```

*SDL1 doesn't actually support using pixel alpha together with surface alpha so the last row should be discounted as no proper blend actually takes place.

As you can see the SDL2 blitter is faster on all of these blits than SDL1. However, the new blitter was only faster at pixel alpha-to-pixel alpha blits. 

My goal is to try and equal SDL1 at least because in theory the new blitter and SDL1 have the same underlying formula, meanwhile SDL2 can 'cheat' in this contest a bit because it uses a different, faster approximation for alpha blending and doesn't have any special cases to handle like SDL1 and the new blitter that mimics it.

src_c/alphablit.c
```diff
@@ -2707,209 +2685,421 @@ alphablit_alpha_sse2_argb_surf_alpha_big_end (SDL_BlitInfo * info)
...
-                   modulated alpha */
-                *dstp = (*srcp & 0xFFFFFF00) | src_alpha;
-            }
-            else
+        Uint64          *srcp = (Uint64 *)info->s_pixels;
+        Uint64          *dstp = (Uint64 *)info->d_pixels;
+
+        Uint64          src_amask = ((Uint64)srcfmt->Amask << 32) | srcfmt->Amask;
+        Uint64          dst_amask = ((Uint64)dstfmt->Amask << 32) | dstfmt->Amask;
+
```
siggisv, line 2708: You might have figured it out by the time you read this, but this is what the old MS VC 9.0 compiler freaks out about.

It mostly follows C90 which forced you to have all your variables declared at the top of the function (or at the top of a code block).

MyreMylar: Yeah I knew it was going almost certainly going to complain about some variable declarations :) Just thought it would be easier to have the CI tell me where to look while I wrote up some other info.

MyreMylar:
So post-optimisation, where did I get to? Here is the current test case output on 64 bit python:

```
Even blit width:
-----------
tested Blit no alpha			: 		43.039ms
tested Blit surface alpha		: 		213.526ms
tested Blit pixel to opaque alpha: 			233.02ms
tested Blit pixel to pixel alpha: 			416.979ms
tested Blit pixel and surf alpha: 			864.525ms

Odd blit width:
-----------
tested Blit no alpha			: 		53.048ms
tested Blit surface alpha		: 		210.191ms
tested Blit pixel to opaque alpha: 			370.36ms
tested Blit pixel to pixel alpha: 			691.628ms
tested Blit pixel and surf alpha: 			874.847ms
```
 
This represents the best case situation running 64 bit code. It looks like the majority of computers now are 64 bit, so hopefully being fastest on 64 bit is a good thing. 

In this configuration all blits are finally faster than SDL1, and even width per-pixel to opaque blits are now also faster than SDL2. I suspect per pixel alpha to opaque blits are one of the most common types so this is somewhat cheering.

That's the best case though, the worst case is a 32 bit blitting:

```
Even blit width:
-----------
tested Blit no alpha			: 		46.042ms
tested Blit surface alpha		: 		203.185ms
tested Blit pixel to opaque alpha: 			311.292ms
tested Blit pixel to pixel alpha: 			471.606ms
tested Blit pixel and surf alpha: 			919.799ms

Odd blit width:
-----------
tested Blit no alpha			: 		54.049ms
tested Blit surface alpha		: 		202.184ms
tested Blit pixel to opaque alpha: 			371.337ms
tested Blit pixel to pixel alpha: 			667.11ms
tested Blit pixel and surf alpha: 			906.192ms
```

The actual worst case of an odd width blit doesn't get any worse than the 64 bit odd width blit. Still faster than SDL1 but slower than SDL2 (except per pixel with surface alpha). However, the best case on 32 bit is not as rosy. The per pixel to opaque is basically the same speed as SDL2 (about 80ms slower than 64bit) and the per-pixel-to-per-pixel is about 50ms slower in this test (about 170ms slower than SDL2).

MyreMylar:
**Summary of the summary**

Basing solely off this test, the new blitter is:

- Faster on average than SDL2 at per-pixel-to-opaque blits and per-pixel+surface-alpha-to-per-pixel blits.
- Slower on average than SDL2 at per-pixel-to-per-pixel blits.
- The new blitter performs at it's best on 64bit with even width blits.

Of course, this is one test metric and perhaps others will show this to be a pile of nonsense.

src_c/alphablit.c
```diff
@@ -25,6 +25,20 @@
...
+/* See if we are compiled 64 bit on GCC or MSVC */
+#if _WIN32 || _WIN64
+   #if _WIN64
+     #define ENV64BIT
+   #endif
+#endif
+
+// Check GCC
+#if __GNUC__
+  #if __x86_64__ || __ppc64__
```
illume, line 37: I guess there are other compilers/systems with 64bitness.
The existing code detects 64bit for clang as well (I guess it defines `__GNUC__`).

So I don't know if something like below is necessary.

```c
/* stdint.h is missing from some versions of MSVC. */
#ifdef _MSC_VER
   #if _WIN64
     #define ENV64BIT
   #endif
#elif __GNUC__ && (__x86_64__ || __ppc64__)
     #define ENV64BIT
#else

#include <stdint.h>

#if INTPTR_MAX == INT64_MAX
#define ENV64BIT
#elif INTPTR_MAX == INT32_MAX
#define ENV32BIT
#else
#error Unknown pointer size or missing size macros!
#endif
#endif /* _MSC_VER */
```

illume (approved):
🎉

Issue #1516, opened 2019-11-21
pg.image.load(fileobj) doesn't work in pygame 2


DeluxeYang:
```
jpg_file = BytesIO(jpg_bytes)
unit_image = pg.image.load(jpg_file).convert()
```

this code works in pygame 1.9.6, but return `pygame.error: Unsupported image format` in pygame 2 dev5

robertpfeiffer:
I could not reproduce your problem. Could you post the JPEG file?

I tried this:
```
import pygame
import io
import sys
pygame.init()
screen=pygame.display.set_mode((640,480))

filename=sys.argv[1]

contents = open(filename, "rb").read()
f_bytes = io.BytesIO(contents)

f_opened = open(filename, "rb")

img0=pygame.image.load(filename).convert()
img1=pygame.image.load(f_opened).convert()
img2=pygame.image.load(f_bytes).convert()

screen.blit(img0, (0,0))
pygame.display.flip()
pygame.time.delay(1000)

screen.blit(img1, (0,0))
pygame.display.flip()
pygame.time.delay(1000)

screen.blit(img2, (0,0))
pygame.display.flip()
pygame.time.delay(1000)
```

Starbuck5:
I could reproduce on Win10.

It seems to be just jpgs, I tested gif, tif, and png without issue.

Output:
```
pygame 2.0.0.dev6 (SDL 2.0.10, python 3.7.5)
Hello from the pygame community. https://www.pygame.org/contribute.html
Traceback (most recent call last):
  File "C:\Users\Starbuck\Desktop\test.py", line 16, in <module>
    img2=pygame.image.load(f_bytes).convert()
pygame.error: Unsupported image format
```

slackerbob:
I can also confirm this issue in dev10

```
res = urlopen(imgurl, timeout=10)
image_str = res.read()
res.close()
image_file = io.BytesIO(image_str)
img_logo = pygame.image.load(image_file)

```
results in:

`pygame.error: Unsupported image format`

I was banging my head against the wall until I realized it worked fine in pygame 1.9.6

illume:
Is this with all image urls or one in particular?

slackerbob:
For me it's happening on all .jpg BytesIO objects and not on any .png BytesIO objects.

BUT I've been testing this and so far I can only reproduce this issue on **Raspbian Buster**.

I could **not** reproduce the issue on OSX Catalina, nor on Arch Linux (x64).

[...]

PR #2249, opened 2020-10-26
WIP: Tests for loading images with BytesIO.

illume:
For https://github.com/pygame/pygame/issues/1516 It's failing on windows python 32bit (works on 64bit).

PIL also has the same issue with BytesIO and jpeg. https://stackoverflow.com/questions/23587426/pil-open-method-not-working-with-bytesio

```
retval = size:1: retval:2: maxnum:2:
read_data = (0, 16, )
read retval:2:
seek offset:0: whence:1:
seek retval:6:
seek offset:14: whence:1:
calling pyseek offset:14: whence:1:
seek retval:20:
```

The return value should be 20, but on windows python 32 it is returning 14. Because on windows it passes in (14, 0) rather than (14,1).
Why it is getting a 0, even when passing in 1? No idea. Maybe some sort of function prototype size with dlls miss match kind of issue.

I could try and work around it perhaps... by creating a SDL_RWOPS_MEMORY from the memory in the BytesIO.

### DIFF
--- a/src_c/rwobject.c
+++ b/src_c/rwobject.c
@@ -676,6 +676,13 @@ _pg_rw_read(SDL_RWops *context, void *ptr, size_t size, size_t maxnum)
 
     retval = Bytes_GET_SIZE(result);
     printf("retval = size:%d: retval:%d: maxnum:%d:\n", size, retval, maxnum);
+    unsigned char * result_string = Bytes_AsString(result);
+    printf("read_data = (");
+    for(int ii=0; ii < retval; ii++) {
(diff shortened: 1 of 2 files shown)


illume:
I don't think I'll be able to figure out the image.load(BytesIO(jpegdata)) issue on windows 32bit python quickly. I could make a quick dirty hack, but won't. Probably the next step is to make a small C example to try and replicate the issue and report a bug to the SDL bug tracker. Maybe that won't work, because it could be some sort of python->SDL dll compilation ABI issue.

PR #2252, opened 2020-10-26
Updates the joystick docs to include controller mappings for X360 & PS4 controllers

### LINKED ISSUE #1167: Joystick Mapping, Triggers, and SDL2
**Joystick Mapping and SDL_GAMECONTROLLERCONFIG**

I use a Sony Wireless DualShock 4 controller which has a known issue with newer kernel drivers in my Linux Distribution. For some reason the left trigger is reported as the third axis and the right stick's y axis comes up as the fifth.

This unfortunate mapping was a big problem for my current project, an overhead adventure shooter that supports WASD Keyboard+Mouse or dual-stick gamepad controls.

Because of this issue I developed a joystick config dialog that allows users to ensure that the correct axis and direction is used for dual stick games. It asks users to Move Up, Move Right, Aim Up, and then Aim Right, then saves the axis number and direction for each one.

I thought this issue would go away with SDL2 because of the global joystick config file that should be used. Perhaps I haven't set that up properly, or perhaps pygame2 doesn't use it. I'm not sure about the details and I'm offline as I write this. 

Perhaps some documentation regarding SDL_GAMECONTROLLERCONFIG is called for.

**Triggers are not Axes** 
Another thing worth mentioning is pygame's handling of triggers as an axis. It was never an issue until I designed the joystick config dialog, which looped through the available axes and looked for a value +/- some threshold(.15 I think). Triggers, like other axes, report 0 until they are moved for the first time, but unlike other axes, they report +1 or -1 after they get moved.

This caused me quite a lot of trouble until I realized that triggers report 1 instead of 0 when they're unpressed. The first time I opened my joystick config dialog after starting the game everything would be fine because the triggers still reported 0. Once I pulled a trigger to configure it for, say, FireLeftWeapon, the trigger reported +1 instead. The next time I ran the joystick dialog it would get stuck in my AwaitRelease() method because the left trigger's absolute value never dropped below the 'centered' threshold. This effectively hung the game for some apparently inexplicible reason.

This long-winded and likely confusing story can be summarized like this: triggers shouldn't report 0 up until they get pressed, and then report +1 until they get pressed again. This can confuse the crap out of someone. The axis should report values between 0 and 1 instead of between -1 and +1 with '0' hiding somewhere in the middle. This might be an SDL issue and not a pygame issue, but I assume that SDL2 reports triggers differently than regular axes and that pygame can return more appropriate values.

If not, at least document trigger behavior to reduce future hardships.

**To Do To Close:**

- [x] Confirm that joystick triggers (both X360 & PS4) behave the same between 1.9.6 and 2.0.0.dev8.
- [ ] Add documentation on the common pattern/ordering of axes for these popular controllers and their ranges.

**Related Docs:** https://www.pygame.org/docs/ref/joystick.html

MyreMylar:
I added controller mappings for the X360 pad and the PS4 pad because A) they are the controllers I possess and B) I think they are two of the most popular controller models and thus most likely to be useful when setting up defaults.

Fixes #1167

### DIFF
--- a/docs/reST/ref/joystick.rst
+++ b/docs/reST/ref/joystick.rst
@@ -232,6 +232,9 @@ So call one of pygame.event.get, pygame.event.wait, or pygame.event.pump regular
 
       The axis number must be an integer from ``0`` to ``get_numaxes() - 1``.
 
+      When using gamepads both the control sticks and the analog triggers are
+      usually reported as axes.
+
       .. ## Joystick.get_axis ##
 
    .. method:: get_numballs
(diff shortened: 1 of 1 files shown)


illume (approved):
👍 thanks.

PR #2253, opened 2020-10-26
Documents that integers underlying key code constants differ between Pygame 1 & 2

### LINKED ISSUE #1128: SDL2: constant keyboard IDs different.
**Probably not a high priority issue, but one found by someone in a real game...**

They were storing the numeric ids, rather than using constants. This meant when they loaded their game the arrow keys were not working.

This can happen when saving keyboard preferences, but saving the numeric ids.

Doing this:
- UP = 273

Instead of this:
- UP = `pygame.K_UP`

Results in the error. Also, if `pygame.K_UP` is saved to a file as 273 with pygame SDL1, and then loaded from disk with SDL2 there will be a problem.

```python
>>> import pygame
pygame 1.9.6
>>> pygame.K_UP
273
```

```python
>>> import pygame
pygame 2.0.0.dev2 (SDL 2.0.9, python 2.7.16)
>>> pygame.K_UP
1073741906

```

A fix might be to use SDL1 constants where possible, but first do a mapping to SDL2 constant numbers. I'm not sure if this is possible however. Additionally, maybe the constants can be turned into some sort of object which is a bit smarter than an integer.

**To Do**
- [ ] Document that key identifier ints differ between SDL 1 & SDL 2. 

**Related Docs:** 
https://www.pygame.org/docs/ref/key.html

MyreMylar:
Fixes #1128

### DIFF
--- a/docs/reST/ref/key.rst
+++ b/docs/reST/ref/key.rst
@@ -36,7 +36,9 @@ The ``pygame.KEYDOWN`` event has the additional attributes ``unicode`` and
 .. _key-constants-label:
 
 The following is a list of all the constants (from :mod:`pygame.locals`) used to
-represent keyboard keys.
+represent keyboard keys. These integers underlying these key code constants differ
+between pygame 1 and 2 and should not be used directly if you need your code to be
+portable between the two major versions.
 
 ::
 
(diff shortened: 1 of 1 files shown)


illume:
@MyreMylar I tried some different wording here. Feel free to revert back if you don't like it.

MyreMylar:
Looks good to me, I was probably being a bit programmer speak :)

PR #2254, opened 2020-10-26
Sets SDL Hint for hidden mouse cursor at window edges when full screen

### LINKED ISSUE #1479: Mouse pointer with mouse.set_visible(False) appears at screen edges when using SCALED flag & scaled desktop
Reported by AtlasDiseaseToday on discord. 

> Found an issue with Pygame 2.0.0.dev6 on Windows 10 in which after calling pygame.mouse.set_visible(False) the mouse sometimes appears (seemingly at random) when moving the mouse (I noticed this happens more when the mouse's x position is at 0).

**Related Docs:** https://www.pygame.org/docs/ref/mouse.html#pygame.mouse.set_visible

MyreMylar:
It's SDL_HINT_WINDOW_FRAME_USABLE_WHILE_CURSOR_HIDDEN. Set it to 0 while we are fullscreen with a hidden mouse cursor, and back to '1' otherwise.

fixes #1479 

Test Case (must have windows desktop scaled to something other than 100%):
```python
import pygame

pygame.init()
window_surface = pygame.display.set_mode((1920, 1080), flags=pygame.FULLSCREEN | pygame.SCALED)

background = pygame.Surface((1920, 1080))
background.fill(pygame.Color('#fb8691'))

fake_cursor = pygame.Surface((32, 32))
fake_cursor.fill(pygame.Color('#88CA99'))

pygame.mouse.set_visible(False)
while True:
        if event.type == pygame.QUIT:
            pygame.quit()
            pygame.quit()

    window_surface.blit(background, (0, 0))

    window_surface.blit(fake_cursor, pygame.mouse.get_pos())
    pygame.display.update()
```

You may also want to set the screen resolution to your normal monitor resolution. Anyway with that in place when I move the cursor to the edges of the screen and push at the edges the mouse cursor will flicker in and out of visibility.

### DIFF
--- a/src_c/mouse.c
+++ b/src_c/mouse.c
@@ -160,7 +160,7 @@ mouse_get_pressed(PyObject *self, PyObject *args, PyObject *kwargs)
     PyTuple_SET_ITEM(tuple, 0, PyBool_FromLong((state & SDL_BUTTON_LMASK) != 0));
     PyTuple_SET_ITEM(tuple, 1, PyBool_FromLong((state & SDL_BUTTON_MMASK) != 0));
     PyTuple_SET_ITEM(tuple, 2, PyBool_FromLong((state & SDL_BUTTON_RMASK) != 0));
-    
+
     if (num_buttons == 5) {
         PyTuple_SET_ITEM(tuple, 3, PyBool_FromLong((state & SDL_BUTTON_X1MASK) != 0));
         PyTuple_SET_ITEM(tuple, 4, PyBool_FromLong((state & SDL_BUTTON_X2MASK) != 0));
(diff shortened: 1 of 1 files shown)


MyreMylar:
Looks like the mac travis build just fell asleep at the end there.

illume:
I added your test case from the issue: #1479 Is that right?

It's been tested on windows already? I can try to reproduce on Ubuntu/Mac if the test case is the one I should use.

illume:
I can't reproduce the problem on Mac.

MyreMylar:
Yeah its been tested it on windows and it fixed the issue. I don't really know what the desktop scaling situation is like on other platforms or whether this bug was happening there at all. The OSes all have different windowing and mouse behaviour too so who knows.

It at least shouldn't do any harm on linux or mac.

Issue #2241, opened 2020-10-23
pygame window not repainting itself from window damage events


xandark:
### Environment:

Operating system: Linux - KDE neon
Python version: 3.8.5
SDL version: 2.0.12
PyGame version: 2.0.0.dev22
Relevant hardware: n/a

### Current behavior:

With my personal game and some of the included pygame examples, the pygame window's background does not repaint when another window is brought in front temporarily. When I run either...

python3 -m pygame.examples.aliens
python3 -m pygame.examples.blend_fill

...and I bring an arbitrary window in front of the game window, and then bring the game window to the front again, the game window doesn't repaint its background; you see the objects drawing on the garbage of the previous window's contents.

### Expected behavior:

The same behavior as pygame 1.9.6 is expected: it repaints the background from window damage events.

### Screenshots

![pygame window not repainting itself](https://user-images.githubusercontent.com/5048913/97061806-0f66ea00-154d-11eb-824f-acbdffe3004c.png)

### Steps to reproduce:

python3 -m pygame.examples.blend_fill
Bring a window in front of the pygame window, then move it away. The pygame window doesn't repaint.

tags: screen, update, updating, repaint, corrupted

ankith26:
Woah, even im getting some weird (related) issues on my pi, when I run aliens or blend_fill. 
Not only is it not repainting the screen, its also giving some sudden crashes and things from other windows are being drawn into the pygame window.

This does not happen on my other pygame app (which does not use things like alpha blit), which leads me to believe that this is an effect of #2213 )

MyreMylar:
I suspect this may be related to using `pygame.display.update(rects)`

MyreMylar:
Just confirmed that this doesn't happen on windows. It's not solely `pygame.display.update(rects)` because the `blend_fill` example doesn't use that - however `blend_fill` also doesn't update the screen unless you press a key.

MyreMylar:
New idea someone who can build pygame and uses linux regularly might want to try, sticking:

```
SDL_SetHint(SDL_HINT_VIDEO_X11_NET_WM_BYPASS_COMPOSITOR, "0")
```

Somewhere in the pygame C code (test it in initialization?), to see if it has any effect on this. It looks like that X11 hint is enabled by default in SDL for all apps, but it appears to be intended only for full screen apps so if it is related we might want to turn it on and off in the correct places in `set_mode()` and `toggle_fullscreen()`.

MyreMylar:
Details on the hint:

[...]

PR #2255, opened 2020-10-27
Sets SDL Hint to bypass the x11 compositor based on fullscreen state

MyreMylar:
Previously this hint, new in pygame 2, always defaulted to 'on' and may be causing some issues with windowed apps using x11.

This is related to #2241 but whether it fixes it is untested as I am unable to reproduce the issue on my available platforms.


MyreMylar:
I don't think it is a good idea to merge this as it is as the issue seems to only affect KDE linux and some windowed apps may want to run with the compositor disabled. 

I think ultimately we have to push this compositor decision out to the person making the app and document it. At least we know what it is now though.

Issue #2197, opened 2020-10-16
Should we use deprecation warnings anyway?


illume:
There's some arguments for and against.  Can we gather some research and tests to then make an informed decision together on which way we go?

Some notes from the discord:

> you can filter them: https://stackoverflow.com/questions/14463277/how-to-disable-python-warnings

> Regarding deprecation warnings and people not liking being spammed by them, it looks like they are by default filtered out by Python. See: https://docs.python.org/2/library/warnings.html#default-warning-filters and https://docs.python.org/3.5/library/warnings.html#default-warning-filters (and so on for 3.6, 3.7, etc.)
... with the exception that since 3.7 DeprecationWarning is once again shown by default when triggered directly by code in __main__.

MyreMylar:
Reading the docs it looks like the situation with DeprecationWarnings now for most users going forward (most people are post 3.7 already I would guess? Plus that will only increase) will be that they are printed once for each occurrence in the code. The relevant line is this:

```
"default" - print the first occurrence of matching warnings for each location (module + line number) where the warning is issued
```

That means if you have a loop that features a line triggering a DeprecationWarning you should only see the warning once indicating the line where the warning was triggered I think. A test program proving this would be useful though. We could try it with the current PR that adds a DeprecationWarning.

If it does work like that then to me this seems like the most useful setting, if we add them then most users will encounter the warning when choosing to upgrade pygame and see a simple printed list of where the deprecated code is located and can opt to remove/adapt it. 

For those that don't want to adapt they can either stay on an older pygame pre-warnings, or filter the warnings themselves. e.g.

```
import warnings
warnings.filterwarnings("ignore::DeprecationWarning:pygame[.*]")
```

Should filter out all `DeprecationWarning` messages produced by pygame and subpackages thereof (I haven't tested this yet).

zoldalma999:
>That means if you have a loop that features a line triggering a DeprecationWarning you should only see the warning once indicating the line where the warning was triggered I think. A test program proving this would be useful though.

```py
import pygame
pygame.init()

screen = pygame.display.set_mode([640, 360])
running = True

while running:
    for e in pygame.event.get():
        if e.type == pygame.QUIT:
            running = False
        if e.type == pygame.KEYDOWN:
            if e.key == pygame.K_ESCAPE:
                running = False

    screen.fill([20, 20, 20])
    if pygame.key.get_pressed()[pygame.K_a]:
[...]

PR #2261, opened 2020-10-28
Add deprecation warning when draw.aaline is called with blend=False

zoldalma999:
#592 [old pr](https://github.com/pygame/pygame/pull/2162)

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -115,6 +115,13 @@ aaline(PyObject *self, PyObject *arg, PyObject *kwargs)
         return NULL; /* Exception already set. */
     }
 
+    if (!blend) {
+        PyErr_WarnEx(PyExc_DeprecationWarning,
+                     "blend=False will be deprecated in pygame 2.2 and will "
+                     "default to True",
+                     1);
+    }
+
     surf = pgSurface_AsSurface(surfobj);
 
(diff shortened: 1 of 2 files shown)


MyreMylar:
Should link the general discussion issue on these:

https://github.com/pygame/pygame/issues/2197

My final verdict on DeprecationWarnings is that we should start using them from 2.0.0 onwards, they don't spam, there are options to filter them if the user doesn't care at all and it is better for us to provide feedback before changing any APIs. Since the focus post 2.0.0 is slightly less on compatibility with the past, and more on useful features to catch pygame up with the modern world DeprecationWarnings seem like a useful tool in that box.

I won't approve this for now in case everyone else is against, but I'm for them in general and I think the way it's done in this PR is a good method.

Starbuck5 (approved):
Nobody seems to have come out against deprecation warnings, and it has been a while.
These changes look good to me.

Starbuck5:
@illume - I think this is ready to merge.

After merging this I was thinking it would good to open an issue that reminds us to actually deprecate this and `set_system_cursor()` by 2.2.

illume (approved):
Thanks :)

illume:
Thanks @Starbuck5 . I created an issue here: https://github.com/pygame/pygame/issues/2405 with the milestone 2.2.

PR #2265, opened 2020-10-28
Experimental markings in the docs, remove docscomments.json file.

illume:
Anything that isn't fully tested, where the API is unclear or might need improving can be marked experimental to warn users of potential change.

Also remove the docscomments.json file, since it's not really something we should distribute.


MyreMylar (approved):
Looks good 👍

PR #2266, opened 2020-10-28
Environment variables docs

### LINKED ISSUE #1364: environment variables documentation
We should document the SDL environment variables, and also the pygame ones.

MyreMylar:
A start on some environment variable documentation. I want to expand from here if it seems reasonable. This is how it looks built locally. 

**Current look:**

image

Fixes #1364


MyreMylar:
Did some reformatting to the list and added more details on different ways to set environment variables.

MyreMylar:
Added a few SDL environment variables that get used fairly frequently.

illume (approved):
Great :)

PR #2269, opened 2020-10-28
Change opaque surface test to use SDL_ISPIXELFORMAT_ALPHA() macro

MyreMylar:
This is because if a destination surface has no pixel alpha and surface alpha is set on it it's blend mode will no longer be SDL_BLENDMODE_NONE.

However, we still want to treat it as an opaque destination surface from the blitter's perspective because it will have 0 pixel alpha and we can't use the optimised 'no surface alpha' blitter which assumes two surfaces with pixel alpha to do it's thing.

This fixes the menu bug in FlyBoy.


illume (approved):
👍

PR #2270, opened 2020-10-29
Documented pygame.image load_basic, load_extended, save_extended. image.c code cleanups.

### LINKED ISSUE #1732: pygame.image docs missing functions and load_basic ignore namehint argument

ankith26:
Documented pygame.image load_basic, load_extended, save_extended 
- fixes #1732
- image.c code cleanups, and CPython API usage fixes.

Continued on from #2226 PR.

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -371,7 +371,7 @@ image_get_sdl_image_version(PyObject *self)
         return PyObject_CallObject(extverobj, NULL);
 }
 
-#if PG_COMPILE_SSE4_2 && (SDL_VERSION_ATLEAST(2, 0
+#if PG_COMPILE_SSE4_2 && (SDL_VERSION_ATLEAST(2, 0, 0))
 #define SSE42_ALIGN_NEEDED 16
 #define SSE42_ALIGN __attribute__((aligned(SSE42_ALIGN_NEEDED)))
 
@@ -1671,9 +1671,9 @@ static PyMethodDef _image_methods[] = {
     
     {"save_extended", image_save_extended, METH_VARARGS, DOC_PYGAMEIMAGESAVEEXTENDED},
     {"save", image_save, METH_VARARGS, DOC_PYGAMEIMAGESAVE},
-    {"get_extended", image_get_extended, METH_NOARGS,
+    {"get_extended", (PyCFunction)image_get_extended, METH_NOARGS,
      DOC_PYGAMEIMAGEGETEXTENDED},
-    {"get_sdl_image_version", image_get_sdl_image_version, METH_NOARGS,
+    {"get_sdl_image_version", (PyCFunction)image_get_sdl_image_version, METH_NOARGS,
      DOC_PYGAMEIMAGEGETSDLIMAGEVERSION},
 
     {"tostring", image_tostring, METH_VARARGS, DOC_PYGAMEIMAGETOSTRING},
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -59,7 +59,7 @@ static PyObject *extverobj = NULL;
 
 /* define docs for undocumented (kinda deprecated) functions */
 #define DOC_PYGAMEIMAGELOADBASIC "internal function to load images, use pygame.image.load() instead of this"
-#define DOC_PYGAMEIMAGELOADAEXTENDED DOC_IMAGELOADBASIC
+#define DOC_PYGAMEIMAGELOADEXTENDED DOC_PYGAMEIMAGELOADBASIC
 #define DOC_PYGAMEIMAGESAVEEXTENDED "internal function to save images, use pygame.image.save() instead of this"
 
 static const char *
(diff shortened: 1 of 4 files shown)


illume:
Hi,

can you please add a description of what this PR is for? Is it ready to review?
It looks like a code cleanup?

cheers,

ankith26:
Well, this PR is basically code cleanup, but is also closing  #1732 and making the function doc-strings more accurate. It’s been ready for code review since the last commit

src_c/image.c
```diff
@@ -64,6 +53,15 @@ opengltosdl(void);
     ((flipped) ? (((char *)data) + (height - row - 1) * width) \
                : (((char *)data) + row * width))
 
+static PyObject *extloadobj = NULL;
+static PyObject *extsaveobj = NULL;
+static PyObject *extverobj = NULL;
+
+/* define docs for undocumented (kinda deprecated) functions */
```
illume, line 60: I don't think these functions are deprecated, or will be. They're still useful, and used fairly widely.

Can you please add these docs into the .rst file, and have them come in via the doc headers (generated with `python setup.py docs`)?

illume: > pygame.image.load_basic is always be available, no matter how pygame is compiled. Unfortunately they only support loading .bmp files (not png, jpeg or others). Most people would use pygame.image.load() instead.

ankith26: Ok, so these functions are useful, then. And I had made a PR a while back discouraging their use in the docs.
Ok, it’s understandable that load_basic() might fall useful so I can make the changes you are asking.
What about load_extended and save_extended

src_c/image.c
```diff
@@ -1715,106 +1724,26 @@ MODINIT_DEFINE(image)
...
-            Py_DECREF(extmodule);
+        extsaveobj = PyObject_GetAttrString(extmodule, "save_extended");
+        if (!extsaveobj) {
             DECREF_MOD(module);
             MODINIT_ERROR;
         }
-        if (PyModule_AddObject(module, "get_sdl_image_version", sdlImageV)) {
-            Py_DECREF(sdlImageV);
-            Py_DECREF(extmodule);
+        extverobj = PyObject_GetAttrString(extmodule, "_get_sdl_image_version");
```
illume, line 1741: I think this block of changes introduces some references counting issues? When there's an error it doesn't decrement some of the objects it has made from using PyObject_GetAttrString.

ankith26: Well, I’m kind of a newbie, so never really checked about ref-count issues. I will look into it now

ankith26: My idea when I wrote that block is that, if an error occurred, then that variable is set to NULL, which is useful for later checks on wether the extended module is available or not, my question is should we decrement the refcount of a NULL pointer.

src_c/image.c
illume: There could be an error block at the end which can do all the cleanup in one place. There are decrement functions which do not fail if the variable is NULL. Which means on failure, all of them would be decremented (and then set back to NULL).

ankith26: It’s kind of under the wrong block of comments, but I get it :) 
We do not want to decrement these variables at the end of modinit stuff, because they get use different throughout the code too.

src_c/image.c
ankith26: I fixed the issue :)

src_c/image.c
ankith26: Fixed the refcount issue, and handling errors clearly. As to any discussion on the usage of the load_basic and docs stuff, we can continue that outside this review block (it’s  a bit confusing in here)

ankith26:
I documented the functions, so that this PR is ready to merge.

ankith26:
Oops, I forgot [skip ci] on this one. It was merely a docs change.
Sorry

illume:
> I documented the functions, so that this PR is ready to merge.
> PS: I would like if this PR gets squash-merged, just the because I don’t want my messy commits lingering around in pygame commit-history :)

Ok, cool.

> Oops, I forgot [skip ci] on this one. It was merely a docs change.

We're not currently testing docs, but that would be a nice optimization to skip jobs based on paths changed. Maybe some new new stuff allows that now.

illume:
Looking good, thanks :) The code is much easier to read now.

I updated the title, and PR description so it can be used in the commit message.

Even though there are no new tests in the PR... The functions changed are already quite well tested :) Testing save_extended with the missing SDL2_image is possible, but quite hard since we're not already testing that configuration in CI.

illume (approved):
👍

PR #2271, opened 2020-10-29
WIP: Improve precision of time module

ankith26:
Continued in #2411


ankith26:
**TODO**
- [ ] Update the names of the functions in time.c, to fall inline with conventions
- [ ] Revamp the unittests for time module, making them flexible yet strict at the same time
- [ ] Run unittests on Windows and Linux CI

ankith26:
Well, this PR is almost done with the code parts, now only docs is left to do.

Since its a bit of an API change, It needs some thinking before merging. 
Well, the timing functions being more accurate is a good thing, but now some functions return floats instead of ints, so what can we do about that?

Also notice that I have aliased two methods to other methods

ankith26:
Well hopefully now it works, in my local testing, low values of delta would work, apparantly it does not on CI (now windows is all fine, its just linux)

illume:
Just a note, Travis CI are starting to limit our build minutes https://github.com/pygame/pygame/issues/2300

ankith26:
Which version of pygame does this PR get merged. I’m just asking because I have to mention the version in which this change occurred, in the docs.
Is this PR something for pygame 2.0.1 or must it wait till 2.1 (a very long time, I prefer the former)

src_c/time.c
```diff
@@ -93,107 +90,125 @@ timer_callback(Uint32 interval, void *param)
...
+    else if (millis >= 15)
+        sleeptime = (int)millis - 4;
+    else if (millis >= 10)
+        sleeptime = (int)millis - 3;
+    else if (millis >= 5)
+        sleeptime = (int)millis - 2;
+    else if (millis >= 3)
+        sleeptime = (int)millis - 1;
+    else
+        sleeptime = 0;
```
MyreMylar, line 176: What is this block of code doing? There is no comment and at first glance it seems fairly puzzling to make a delay for SDL Delay that is whatever was requested minus 5, if the delay is bigger than 20 milliseconds.

I'm sure it was written for a reason but to look at the reason is very unclear so a comment or perhaps moving this and the SDL_Delay() part below to a separate function with a descriptive name might help.

My uniformed guess is that the function is trying to use  SDL_Delay() to do the bulk of the delaying work because it is faster than a python while loop? But why different numbers have to be subtracted from this delay amount is a little bit lost on me so far.

test/time_test.py
```diff
@@ -4,6 +4,15 @@
 
 Clock = pygame.time.Clock
 
+SDL1 = pygame.get_sdl_version()[0] < 2
+
+# Because timing based tests are twitchy, just test stuff again if they fail,
+# but only upto NUM_TRIES
+NUM_TRIES = 3
+
```
MyreMylar, line 12: Could use an extra line break here to make linters happy with PEP8 compliance.

test/time_test.py
```diff
@@ -4,6 +4,15 @@
...
+# here, delta represents allowable error ratio
+def _is_almost_equal(a, b, delta):
+    return b * (1 - delta) <= a <= b * (1 + delta)
```
MyreMylar, line 16: And a line break here before the class definition starts.

test/time_test.py
```diff
@@ -4,6 +4,15 @@
...
 
 class ClockTypeTest(unittest.TestCase):
     __tags__ = ['timing']
```
MyreMylar, line 18: also one line break after this line, before the first function definition.

test/time_test.py
```diff
@@ -19,30 +28,35 @@ def test_get_fps(self):
...
     def test_get_rawtime(self):
-
-        iterations = 10
-        delay = 0.1
-        delay_miliseconds = delay*(10**3) #actual time difference between ticks
-        framerate_limit = 5
-        delta = 50 #allowable error in milliseconds
+        iterations = 5
+        delay = 0.02
+        delay_miliseconds = delay * 1000
```
MyreMylar, line 56: typo here in word 'miliseconds' - looks like it is an old typo that has just been preserved. Should be 'milliseconds' with two L's.

test/time_test.py
```diff
@@ -53,23 +67,36 @@ def test_get_rawtime(self):
...
         delay = 0.1 #seconds
-        delay_miliseconds = delay*(10**3)
-        iterations = 10
-        delta = 50 #milliseconds
+        delay_milliseconds = delay * 1000
+        iterations = 5
+        delta = 0.1
+        errcnt = 0
 
         #Testing Clock Initialization
```
MyreMylar, line 101: should be a space after the # here and two before it. Looks like it is an old error in the test, but we may as well clean house in here with all the changes.

test/time_test.py
```diff
@@ -80,17 +107,27 @@ def test_get_time(self):
...
+        # Comparing get_time() results with the 'time' module
         for i in range(iterations):
             t0 = time.time()
             time.sleep(delay)
             c.tick()
             t1 = time.time()
             c1 = c.get_time() #elapsed time in milliseconds
-            d0 = (t1-t0)*(10**3) #'time' module elapsed time converted to milliseconds
-            self.assertAlmostEqual(d0, c1, delta=delta)
+            d0 = (t1-t0) * 1000 #'time' module elapsed time converted to millis
```
MyreMylar, line 125: Again a missing space after the # and should be two before the # as well.

test/time_test.py
```diff
@@ -53,23 +67,36 @@ def test_get_rawtime(self):
...
             c1 = c.get_rawtime()
             c2 = c.get_time()
-            self.assertAlmostEqual(c1, c2, delta=delta)
+            
+            self.assertEqual(c1, c2)
+            if not _is_almost_equal(delay_miliseconds, c1, delta):
+                errcnt += 1
+        
+        if errcnt >= NUM_TRIES:
+            self.fail("error while doing get_rawtime()")
```
MyreMylar, line 92: too many line breaks here. Should be one.

test/time_test.py
```diff
@@ -115,16 +152,18 @@ def test_tick(self):
 
         # verify time.Clock.tick() will measure the time correctly
         c.tick()
-        for i in range(100):
+        for i in range(50):
             time.sleep(milliseconds / 1000) # convert to seconds
```
MyreMylar, line 156: Should be two spaces before the # in this comment.

test/time_test.py
```diff
@@ -137,86 +176,67 @@ def test_tick(self):
...
-
-    def test_tick_busy_loop(self):
-        """Test tick_busy_loop"""
-
+        
+        # now call _tick_test(), which are essentially more rigourous tests
+        if not SDL1:
+            self._tick_test()
+
+    def _tick_test(self, reccnt = 0):
```
MyreMylar, line 203: linter doesn't like the spaces in this part `reccnt = 0` should be `reccnt=0`.

test/time_test.py
```diff
@@ -137,86 +176,67 @@ def test_tick(self):
...
-        frame_length_without_decimal_places_3 = int(second_length/fractional_fps)
-        self.assertGreaterEqual(c.tick_busy_loop(fractional_fps), frame_length_without_decimal_places_3 - shortfall_tolerance)
-
-        bool_fps = True
-        self.assertGreaterEqual(c.tick_busy_loop(bool_fps), (second_length/bool_fps) - shortfall_tolerance)
-
-
+        self.assertAlmostEqual(c.tick(negative_fps), 0, delta=0.5)
+            
 class TimeModuleTest(unittest.TestCase):
```
MyreMylar, line 238: Should be an extra line break before the start of the class definition.

MyreMylar (changes requested):
Overall it's looking in pretty good shape. I did a quick review across the internet and it seems lots of people who've tried it recommend using `SDL_PerformanceCounter()` for making clocks now, so we are not alone in thinking this is a decent idea.

I also tried my original test program from one of the issues where this first came up and got good results from a wide range of `clock.tick(FPS)` values. It seems that tick() now always comes in microscopically under the FPS target in my measurement which is good and a lot more accurate than before.

I noticed one odd part in the C code that I didn't understand, which likely means other people won't either so it's in need of some additional explanation  (unless I'm a total moron today).

I also did a linter scan on the python parts which picked up a few trivial things that should be easy to fix. Likely most of it is from the remnants of the old test, but thought it was worth fixing them now since the file was changing so much anyway.

This is the FPS accuracy test program I was using (I think it the original design was the issue reporters), including it here in case anyone else wants to test this on other platforms:

```
import sys, time, pygame, decimal

from decimal import Decimal

decimal.getcontext().prec = 64

pygame.display.init()

size = width, height = 320, 240
screen = pygame.display.set_mode(size)
clock = pygame.time.Clock()
fps = 120
measures = []
t = 0
last_t = 1 / fps

while True:
    t = time.perf_counter()
        if event.type == pygame.QUIT: sys.exit()

    measures.append(Decimal(1.0) / Decimal(t - last_t))
    if len(measures) > fps * 2:
        del measures[0]
    print(f"manual: {Decimal(sum(measures)) / Decimal(len(measures))}")

    pygame.display.flip()

    clock.tick_busy_loop(fps)
    last_t = t
```

All my testing was done on windows, python 3.9 64bit.

src_c/time.c
ankith26: Yeah, I probably should have explained more with comments, I will add that soon.
My basic idea here is to get a good balance between a sleep and an accurate busy-loop delay, I try to do both, first by sleeping the processor for a while, and then doing a busy loop to cover for the remaining time. This way, the delay is accurate and the processor does not burn off by doing repeatedly busy loops.
That bit of code is just doing crude checks on the time delay to determine the time it has to sleep
If the delay is more than 20 millis, busy loop for 5-ish millis
If the delay is between 15 and 20 millis, busy loop for 4-ish millis and so on

ankith26:
Thanks for the review MyreMylar :) 
Will do 👍

test/time_test.py
ankith26: Done :)

test/time_test.py
ankith26: Done :)

test/time_test.py
ankith26: Done :)

test/time_test.py
ankith26: Done :)

test/time_test.py
ankith26: Done :)

test/time_test.py
ankith26: Done :)

ankith26:
autopep8 time_test.py and explained weird code with comments

test/time_test.py
```diff
@@ -19,78 +31,106 @@ def test_get_fps(self):
...
+        # Testing get_rawtime() = get_time()
         for f in range(iterations):
             time.sleep(delay)
             c.tick()
             c1 = c.get_rawtime()
             c2 = c.get_time()
-            self.assertAlmostEqual(c1, c2, delta=delta)
 
+            self.assertEqual(c1, c2)
+            if not _is_almost_equal(delay_miliseconds, c1, delta):
```
MyreMylar, line 89: One remaining typo here:

`delay_miliseconds` here should be `delay_milliseconds` with two l's.

MyreMylar (changes requested):
Could only find one typo this time. Everything still else still seems to run fine in my tests. 

What else is left to do here? Testing it on more platforms and docs? I'd be happy to see this in the 2.0.1.dev2 or dev4 release to get some more public feedback. I guess my only worry/reservation is that this somehow causes performance problems on slower/obscure hardware - though that seems unlikely - and I don't personally have the wide range of hardware to try it with. 

Otherwise the accuracy (and thus 'smoothness' of frames) is better, the implementation is cross-platform via SDL and it will resolve a few issues on the tracker complaining about the timer.

Anyway, my overall inclination is to merge it into a dev release and see if it generates any negative feedback that we need to adapt to from there.

test/time_test.py
ankith26: Nice spot :)
fixed

ankith26:
This PR is done from my end. Added docs now.

illume (changes requested):
> Anyway, my overall inclination is to merge it into a dev release and see if it generates any negative feedback that we need to adapt to from there.

If it could be merged in as a separate hidden method, I'd be ok with that assuming concerns below are addressed and the API doesn't change. I'm against merging this in still without more latency testing on all the platforms. Having it in as a separate hidden method would allow people to test it more easily.

My major concerns:
- tick() is turning into a busy loop? They are kept separate on purpose, and no-busy-loop as default on purpose.
- return type changing into a float is an API break. Since this number is often passed into movement functions, I expect fairly wide breaking changes.
- it doesn't account for different performance counters on different cores, and that frequency can change rapidly on cores(and be different amongst cores). This is the main reason the performance counters are supposed to be for profiling, and aren't really meant for other types of timing. This doesn't mean it's not possible to use it, just that I'm skeptical without empirical evidence. Turning off frequency scaling, and pinning processes to a certain core is often what people do for profiling to avoid these issues.
- I'm not sure about removing handling of worst clock accuracy. Because bad clocks still widely exist, how is this handled?


I'd suggest:
- making this a separate hidden method, so it can be merged in with more safety. The existing functions should not change.
- make a test plan available so different people can more easily test it on their hardware. (hopefully of the kind... "run these scripts, and paste me the numbers").

MyreMylar:
@illume I can find a fair amount of evidence of people using `QueryPerformanceCounter()/ for calculating delta time. For example:

https://subscription.packtpub.com/book/game_development/9781849694803/1/ch01lvl1sec13/working-with-game-time

And:

https://walbourn.github.io/understanding-game-time-revisited/

Which both recommend using `QueryPerformanceCounter()/QueryPerformanceFrequency()`. I think it is also the recommended method by Microsoft these days. I suspect that the multiple cores thing is probably a bit of a red herring too dating back to issues in the windows XP era when multicore CPUs were first appearing.

ankith26:
Hmm. Let me reason why I have aliased tick() with tick_busy_loop() and delay() with wait(). tick() and wait() used to sleep the processor, while tick_busy_loop() and delay() would keep looping to get a more accurate delay. This would put CPU under heavy load.

My solution was to come up with an accurate_delay function that sleeps for most of the time, and does a busy loop towards the end. This way, it’s both accurate and also CPU friendly. I have done extensive testing on my pi, and I can confirm that this works pretty well.

Another thing is that this new clock, is much more accurate than the previous one anyways. So I don’t see why that should become an issue.

Yes, the API break because some functions are returning floats is quite a valid point. And I’ve been thinking what to do about it. My initial idea was to put an argument to all the functions called returnfloat which is set to false by default, and users can set them to be true. I feel that this way is not so elegant, though

mcpalmer1980:
I'd definitely like to try the new clock and like the idea of adding something like a pg._time so users can test it. A more accurate clock.tick() would be great, but I agree with Illume that it should be a separate function if it effects resource usage more than a negligible amount. Battery usage my be important in some use cases.

Adding a float=False parameter to time functions that can be made to return a float is a fine solution, but there are 2 other possibilities I can think of:

time.set_accurate(True)
import pygame.new_time as time

Just my three cents

ankith26:
I guess I will implement a new experimental module for this. Something like _time
For people to test out stuff. 
And make the shift sometimes in the future, maybe during 2.1

ankith26:
Latest commit moves everything to a separate _time module.
The new tests I wrote are in _time_test.py
Nothing from the current module has been changed.
So it’s safe to merge

ankith26:
This one is almost done, but needs a bit more work. I can finish it by tomorrow, though

ankith26:
#2411 Has superseded this PR, closing this

PR #2272, opened 2020-10-29
Makes pygame/pyportmidi pass the output buffersize down to portmidi

### LINKED ISSUE #1949: midi.Output() buffer_size parameter does nothing
As per the title the `buffer_size` parameter passed to a pygame.midi.Output() class constructor is not used.

**To Do**

- [x] Either use the parameter or document that it does nothing. 

I know that issuing deprecation warnings is not popular, but if buffer_size cannot be implemented then perhaps it should be considered as the parameter is out there and likely being used unknowingly in some projects.

**Docs:** https://www.pygame.org/docs/ref/midi.html#pygame.midi.Output

**Code:** 

https://github.com/pygame/pygame/blob/4562396964cca81bceb2d12d3f09aa96746cb3fa/src_py/midi.py#L379-L434

MyreMylar:
This is used for buffering midi events and the size of the buffer represents how many midi events to buffer. Up to now it has been defaulting to 256 as that is the minimum value set in portmidi, and the pyportmidi wrapper was always passing in 0. I set the default in pygame to 256 so nothing should change in the default behaviour, only now if someone chooses to pass in a buffersize it will actually be used.

fixes #1949

### DIFF
--- a/docs/reST/ref/midi.rst
+++ b/docs/reST/ref/midi.rst
@@ -131,7 +131,7 @@ midi events to pygame events.
    | :sl:`Output is used to send midi to an output device`
    | :sg:`Output(device_id) -> None`
    | :sg:`Output(device_id, latency=0) -> None`
-   | :sg:`Output(device_id, buffer_size=4096) -> None`
+   | :sg:`Output(device_id, buffer_size=256) -> None`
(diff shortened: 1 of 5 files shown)


illume (approved):
👍 looks good.

PR #2277, opened 2020-10-30
Add color.update and rect.update

### LINKED ISSUE #1049: Suggestion: Add Color.update() and Rect.update()?
Is this a good or bad idea? Vectors have a function like this.
* Rect.update(tuple or Rect or x,y,w,h)
* Color.update(tuple or Color or r,g,b,a)

Rect can already be updated using `rect[:] = 1,2,3,4`, but the same is true of vectors.

**To Do:**

- [ ] Add unit test for Color.update() function checking all the same things we check when setting an individual element of the Color.
- [ ] Add unit test for Rect.update() function checking all the same things we check when setting an individual element of the Rect.
- [ ] Create C implementation of Color.update() that passes unit tests.
- [ ] Create C implementation of Rect.update() that passes unit tests.

**Related Docs:** 
https://www.pygame.org/docs/ref/color.html
https://www.pygame.org/docs/ref/rect.html
https://www.pygame.org/docs/ref/math.html#pygame.math.Vector2.update
https://www.pygame.org/docs/ref/math.html#pygame.math.Vector3.update

zoldalma999:
closes #1049 

- [x] C code for rect.update()
- [x] Tests for rect.update()
- [x] Docs for rect.update()
- [x] C code for color.update()
- [x] Tests for color.update()
- [x] Docs for color.update()


src_c/color.c
```diff
@@ -920,6 +923,60 @@ _premul_alpha(pgColorObject *color, PyObject *args)
...
+        if (_parse_color_from_single_object(r_or_obj, rgba)) {
+            return NULL;
+        }
+    }
+    else {
+        Uint32 color = 0;
+
+        /* Color(R,G,B[,A]) */
+        if (!_get_color(r_or_obj, &color) || color > 255) {
+            PyErr_SetString(PyExc_ValueError, "invalid color argument");
```
ankith26, line 949: Nice work on the PR :) 👍 
A small nitpick, you can use a macro to set error and return NULL, just like.
`return RAISE(PyExc_ValueError, "invalid color argument");`

ankith26 (changes requested):

ankith26 (approved):

ankith26:
Nice work on the PR, the new methods will be really useful :) 👍

ankith26:
Close and open to run the CI again :)

MightyJosip (changes requested):
Can you also modify the files connected to the type hints: (https://github.com/pygame/pygame/blob/main/buildconfig/pygame-stubs/color.pyi and https://github.com/pygame/pygame/blob/main/buildconfig/pygame-stubs/rect.pyi)

MightyJosip (approved):
Thanks

nthykier:
For reference, I would prefer if Colors are immutable.  See https://github.com/pygame/pygame/issues/1049#issuecomment-721990272 for my rationale.

illume:
I'd like to leave this open until we reach some sort of consensus on immutability with Color.

illume (approved):
👍 very nice, thank you.

Just waiting for more discussion on the Color immutable thing before merging.

ankith26:
I totally support the idea of mutable colours and this PR looks very good to me, considering the fact that colours are already mutable (you can set the r, g, b and a attribute of the Color object)

illume:
Good point @ankith26.

PR #2278, opened 2020-10-30
Adds an interactive test for set_gamma_ramp()

### LINKED ISSUE #1751: Add interactive unit test: display.set_gamma_ramp()

MyreMylar:
This lets us verify that the function is actually having an effect  in a way the non-interactive test is unable to. It is not supported on all platforms and has anti-abuse measure in place on at least Windows to try and stop it being used to blank out the screen.

fixes #1751

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -516,28 +516,56 @@ class DisplayModuleTest(unittest.TestCase):
                 self.assertEqual((test_surf.get_width(), test_surf.get_height()), \
                 width_height)
 
+
 class DisplayInteractiveTest(unittest.TestCase):
 
     __tags__ = ["interactive"]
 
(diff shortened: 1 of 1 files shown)


illume (approved):
👍

PR #2283, opened 2020-10-31
Unittests for event module

ankith26:
Finish writing unittests for last two pending functions in the event module


MyreMylar (changes requested):
The `get_grab()` test doesn't work locally - I assume because you don't create a window and `set_grab()` specifically checks for the presence of a window.

See the test: `test_set_grab__and_get_symmetric`

ankith26:
Thanks MyreMylar for pointing that out :)
Hopefully fixed now

MyreMylar (approved):

MyreMylar:
Yep, it all works now.

Thanks for the contribution! 👍 🎉

Issue #1896, opened 2020-06-02
setup.py lint


illume:
How about we have a command to run all the lint tools?

- black
- pylint
- clang-format

MyreMylar:
I think this is a good idea - once we have the number of linter errors down to a mangeable number (or preferably zero). I suspect if we turned it on now it would be a terrifying flood of hundreds, at least I'm sure pylint would do that. I think black auto formats, and I have no idea what clang-format does.

 Do all the linters agree with each other or will they get into linter fights? :O

illume:
> once we have the number of linter errors down to a mangeable number (or preferably zero). I suspect if we turned it on now it would be a terrifying flood of hundreds, at least I'm sure pylint would do that

We could start off with a list of lint-clean files that `setup.py lint` can check. Adding more files as they are cleaned.

>  I think black auto formats, and I have no idea what clang-format does.

Yeah, clang-format auto formats C/C++ similar to black for python.

I was thinking of some sort of pre-commit command - similar to a `make check`. To make it easier to run all the different things. Probably should add `setup.py docs` to this to make sure header files are generated. I think you can chain distutils commands so `setup.py docs lint test` could perhaps work.

> Do all the linters agree with each other or will they get into linter fights? :O
black and pylint agree at least

Pierre-Sassoulas:
Hello, 

Can I be assigned to this ? I would add a `pre-commit` configuration, like the one for pylint here:
https://github.com/PyCQA/pylint/blob/master/.pre-commit-config.yaml

Then with
```
pip3 install pre-commit
pre-commit install
```
pre-commit hooks are added to your local git repository and are executed on `git commit`. I could also update the travis job.

Pierre-Sassoulas:
@MyreMylar there is some configuration to do so the linter do not enter a linter war. Main thing to consider is the number of character per line. (Look like yours is 79, right ?)

illume:
Hello @Pierre-Sassoulas,

thanks for starting on this.  I don't think this issue has had enough discussion for implementation. Is it possible you could detail your thoughts on some of the choices you've made?


A few notes, and some background...

This issue was about a `setup.py lint` command using the tools we had selected (black, pylint, clang-format). The main benefit is that we build with other commands through the setup.py.

[...]

PR #2284, opened 2020-10-31
Add parentheses to print so it works in py2 and py3

Pierre-Sassoulas:
`print("foo")` works in both python2 and python3. `except Exception as e:` too. The python2 only syntax prevented `mypy` from running on the codebase for #1896 .

### DIFF
--- a/test/util/gen_stubs.py
+++ b/test/util/gen_stubs.py
@@ -104,8 +104,8 @@ def get_instance(type_):
 
     try:
         return helper(*arg)
-    except Exception, e:
-        raw_input("FAILED TO CREATE INSTANCE OF %s\n%s\n" 
+    except Exception as e:
+        raw_input("FAILED TO CREATE INSTANCE OF %s\n%s\n"
                   "Press Enter to continue" % (type_, e))
         return type_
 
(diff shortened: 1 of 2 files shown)


ankith26:
Welcome :)
Noticed that your other PR #2285 makes the sameness changes that this PR makes, so this PR is kinda redundant ??

Pierre-Sassoulas:
Good catch, #2285 need this, but it is more of a proposition to discuss and could be rebased when this is merged. I wanted to separate the various issues and have manageable review for maintainers :) I also expect the compat between python 2  and 3 to be easier to merge because I think it's non controversial.

ankith26:
Yes you are right, this PR is easier to approve.
The code you have updated in this PR was mostly old, unused code. 
But still, thanks for the contribution :)

MyreMylar:
As a note we can delete `svn_log_to_whatsnew.py` since pygame is no longer using svn and will not return.

I'm not as sure if `gen_stubs.py` (and the other connected `.py` files in the `util` sub-directory) are still useful. Theoretically `gen_stubs.py` is there to make unit test stubs whenever somebody makes a new pygame module. However, clearly that has not happened for some time. As well as it being written in the python 2 style, several of the imports it was relying on now no longer function.

I think on balance it would be useful if it worked for 2.1+ so I think raising an issue to investigate seeing if `gen_stubs.py` can be made functional again is the right way forward. It has documentation in the README.TXT so it's not a completely impossible task.

MyreMylar (approved):
Thanks for the contribution! 👍

PR #2285, opened 2020-10-31
Create a setup.py lint function 

### LINKED ISSUE #1896: setup.py lint

Pierre-Sassoulas:
This add automated linting to the repository, some of it is automated during commit (isort/black), some of them are really floody and should probably be activated one by one with a plan (mypy/flake8/pylint). We could also add autoflake to remove unused import and variable automatically.

Resolve #1896


illume (approved):
Thanks @Pierre-Sassoulas 

Sorry this took so long to get in.

I added a couple of commits in as well: fixed the conflict, removed the non-cmd based docs argument processing, got clang-format working, and disabled isort (for now).

PR #2286, opened 2020-10-31
Move threads.py out of its directory (minor update)

ankith26:


ankith26:
The Import that LGTM is complaining about, is a required import

illume:
It doesn't look like `Empty` is used to me. I think LGTM is correct.

ankith26:
It is not used in the file, but is used when some other file does
from pygame.threads import Empty

Which is why I included the import again

ankith26:
Ahh, just realised. 
Moving the file to the base directory will not work because threads module want some to import pythons time module, and pygame has it’s own time module, so it’s not going to import stuff correctly.
I’m gonna have to drop this PR

PR #2291, opened 2020-10-31
Change range(len(x)) to enumerate(x) when applicable

Pierre-Sassoulas:
Fix all C0200 consider-using-enumerate signaled by pylint

### DIFF
--- a/src_py/cursors.py
+++ b/src_py/cursors.py
@@ -268,8 +268,8 @@ def compile(strings, black='X', white='.', xor='o'):
     tells which characters will represent black pixels, and which
     characters represent white pixels. All other characters are
     considered clear.
-    
-    Some systems allow you to set a special toggle color for the 
+
+    Some systems allow you to set a special toggle color for the
     system color, this is also called the xor color. If the system
     does not support xor cursors, that color will simply be black.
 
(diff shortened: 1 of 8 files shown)


src_py/cursors.py
```diff
@@ -365,22 +365,22 @@ def bitswap(num):
...
     cursdata = tuple(cursdata)
-
-    for line in range(len(mask)):
-        if mask[line].startswith('static char') or mask[line].startswith('static unsigned char'):
+    for line in mask:
+        if line.startswith(possible_starts):
             break
 
-    data = ' '.join(mask[line+1:]).replace('};', '').replace(',', ' ')
+    data = ' '.join(mask[i+1:]).replace('};', '').replace(',', ' ')
```
illume, line 383: I don't think this is correct to use i instead of line?

Pierre-Sassoulas: Good catch, there's a problem here, there is no enumerate(), I'm going to fix it.

Pierre-Sassoulas: I changed `for line in mask:` to ` for i, line in enumerate(mask):` so the `i` now exists.

src_py/cursors.py
```diff
@@ -365,22 +365,19 @@ def bitswap(num):
 
     info = width, height, hotx, hoty
 
-    for line in range(len(curs)):
-        if curs[line].startswith('static char') or curs[line].startswith('static unsigned char'):
+    possible_starts = ['static char', 'static unsigned char']
+    for i, line in enumerate(curs):
+        if line.startswith(possible_starts):
```
illume, line 370: TIL: Seems startswith can't be a list, but can be a tuple.

Pierre-Sassoulas: Yeap, TIL too :) I fixed it.

src_py/cursors.py
illume: cool :)

illume (approved):
👍 thanks.

Issue #2282, opened 2020-10-31
SDL_VIDEODRIVER windib should not fail


illume:
It should probably translate this to 'windows' automatically.
```
os.environ['SDL_VIDEODRIVER'] = 'windib'
pygame.error: windib not available
```

illume:
There's some other ones we should handle...

https://wiki.libsdl.org/FAQUsingSDL

### Video
...

### Audio 

dsound -> directsound
pulse -> pulseaudio
dma -> ... nothing equivalent, maybe use 'dsp' in this case.

PR #2292, opened 2020-10-31
backwards compatible env vars, SDL_VIDEODRIVER=windib, SDL_AUDIODRIVER=pulse or dsound 

illume:
For https://github.com/pygame/pygame/issues/2282

Otherwise apps that use SDL_VIDEODRIVER=windib or SDL_AUDIODRIVER=pulse or dsound will fail when using SDL2.

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -245,6 +245,17 @@ pg_quit(PyObject *self, PyObject *arg)
 static PyObject *
 pg_init(PyObject *self, PyObject *args)
 {
+
+#if IS_SDLv2
+    /* Compatibility:
+        windib video driver was renamed in SDL2, and we don't want it to fail.
+    */
+    const char *drivername = SDL_getenv("SDL_VIDEODRIVER");
+    if (drivername && SDL_strncasecmp("windb", drivername, SDL_strlen(drivername) == 0)) {
+        SDL_setenv("SDL_VIDEODRIVER", "windows", 1);
+    }
+#endif
+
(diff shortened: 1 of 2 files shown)


src_py/__init__.py
```diff
@@ -103,6 +103,11 @@ def warn(self):
         import pygame.cdrom
     except (ImportError, IOError):
         cdrom = MissingModule("cdrom", urgent=1)
+else:
+    if os.environ.get('SDL_VIDEODRIVER') == 'windib':
+        # windib isn't in SDL2, we don't want it to fail.
+        environ['SDL_VIDEODRIVER'] = 'windows'
+
```
MyreMylar, line 110: Should we do this here, or in `display.init()` ? I suspect there may be people who do:

```
import os
import pygame
os.environ['SDL_VIDEODRIVER'] = 'windb'

pygame.display.init()  # or regular pygame.init()
```

Which I think will work with SDL, but will not be caught by this.

illume: Good point.

illume: Updated it to do the business in `pygame.display.init()` instead.

illume:
(note, I cancelled the travis build, it didn't fail).

ankith26 (approved):
This change looks good for me ;) 
Cheers!

illume:
I fixed the SDL_strncasecmp, and added backwards compatible SDL_AUDIO_DRIVER names too.

PR #2295, opened 2020-11-01
Makes mixer.music.get_busy() return False when the music is paused

### LINKED ISSUE #1449: Should music.get_busy() detect pause or get_paused() be added?
There is no way currently to detect if mixer.music is actually playing. get_busy() returns true when music is paused.

Will using SDL_MixPausedMusic() to detect paused music in mixer.music.get_busy() adversely effect legacy code? I don't think it will.

Should get_paused() be added? I think the above is better.

Note that SDL_MixerPausedMusic will return false if the music is stopped, a combination of the to calls is necessary to determine if music is actually playing.

I could add this functionality if it seems useful. It's not very important because users can code their own states, but it would be convenient. Also, I think the way that get_busy currently works is silly, quirky behaviour from SDL that could have been improved for pygame

**To Do**

 - [ ] Make `get_busy()` track whether music is paused on pygame side and return False if music is paused.
 - [ ] Document this change from pygame 1.

**Related Docs**: https://www.pygame.org/docs/ref/music.html#pygame.mixer.music.get_busy

MyreMylar:
Fixes #1449

### DIFF
--- a/docs/reST/ref/music.rst
+++ b/docs/reST/ref/music.rst
@@ -154,7 +154,11 @@ can crash the program, ``e.g``. Debian Linux. Consider using ``OGG`` instead.
    | :sg:`get_busy() -> bool`
 
    Returns True when the music stream is actively playing. When the music is
-   idle this returns False. It returns True even if the music is paused.
+   idle this returns False. In pygame 2.0.1 and above this function returns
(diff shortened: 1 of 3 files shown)


nthykier (approved):

illume (approved):
Oh, cool! 👍

PR #2303, opened 2020-11-05
Delete obsolete test/util files

### LINKED ISSUE #2289: Remove everything test/util/ as these scripts are no longer used.
Since it's been a long time since we last added a new python module this script has stopped working, functions it imported no longer exist in the code base and until recently it didn't even run in python 3.

There is documentation for it in README.TXT so it shouldn't be too impossible to reverse engineer how it worked from the existing code and the expected output.

ankith26:
closes #2289


illume (approved):
Thanks 👍

PR #2310, opened 2020-11-08
expose GL version/debug info

robertpfeiffer:
This would be useful for diagnosing portability/driver compatibility problems on platforms where PyOpenGL and modernGL are not available, like Raspberry Pi and android.


illume:
I'm against this because:

- glGetString can be easily used from other OpenGL bindings.
- no tests, docs, or any discussion about the API. 
- includes an extra OpenGL header

PR #2250, opened 2020-10-26
rearrange event posting, zero-init event struct

robertpfeiffer:
This PR improves the code in event.c and documents problems that became apparent during the work on #2244
- Zero-initialise the posted event struct to avoid posting garbage in the padding that could become significant when events are cast to a different type. This should fix #2146
- ~~move setting the SDL event filter from modinit to pygame.init(), this seems to help with #1238 on my machine~~
- replace misnamed and duplicated maths function with standard library version
- comment why seemingly redundant code is needed
- defensive coding: Check for impossible (according to SDL2) event objects before posting them
- remove redundant code introduced in #2244 

Given the history of #1238, I have no optimism about actually solving this problem for good, but I am confident that none of the changes that remove redundant code are incorrect. I am also confident that all exceptions are raised only in cases that *should* result in an exception, and that all comments are correct. At the very least, more comments will dissuade the next poor soul from trying to refactor this code without understanding what is going on.

PR #2311, opened 2020-11-08
Add K_AC_BACK key constant, fix data_file pruning and "STRIPPED" install.

robertpfeiffer:
fix typo in setup.py, and expose android back button as constant.
(Sorry, I forgot to push this last commit in my android PR before 2.0)


illume:
How can this be tested?

robertpfeiffer:
> How can this be tested?

the typo date_files->data_files, the extensional list of data files when stripped, or the back button constant?

ankith26 (approved):
Nice one :) 

Just a small idea in case you want to make stuff more compact. Of all the icon files you are loading, only pygame_icon.bmp is actually used by pygame (and pygame_icon.tiff is additionally used on Mac). All other files need not be included (which can save about 200 kB), and the tiff file can be conditionally included on Mac only (can save about 60 kB more on non-Mac platforms)

And adding type hints for the new constant you added, will make @MightyJosip happy :)

pmp-p (approved):

MyreMylar:
This PR also needs the new keycode added to `key.rst` for the docs and `locals.py` for the linters as well.

Yes, it's kind of annoying we have keycode stuff in four different places but that's how it is right now.

docs/reST/ref/key.rst
```diff
@@ -181,6 +181,7 @@ that your key handling code works well on both pygame 1 and pygame 2.
       K_MENU                menu
       K_POWER               power
       K_EURO                Euro
+      K_AC_BACK             Android back button
```
illume, line 184: Is it Android specific? I thought those worked on keyboards too?

https://wiki.libsdl.org/SDL_ScancodeAndKeycode
> "AC Back" (the Back key (application control keypad)).

robertpfeiffer: I guess I'll have to try this with a flirc and also a RF multimedia remote. I'll report back later

MyreMylar:
Possibly relatedly, do we also need to be aware of this SDL hint?

```
/**
 * \brief A variable to control whether we trap the Android back button to handle it manually.
 *        This is necessary for the right mouse button to work on some Android devices, or
 *        to be able to trap the back button for use in your code reliably.  If set to true,
 *        the back button will show up as an SDL_KEYDOWN / SDL_KEYUP pair with a keycode of 
 *        SDL_SCANCODE_AC_BACK.
 *
 * The variable can be set to the following values:
 *   "0"       - Back button will be handled as usual for system. (default)
 *   "1"       - Back button will be trapped, allowing you to handle the key press
 *               manually.  (This will also let right mouse click work on systems 
 *               where the right mouse button functions as back.)
 *
 * The value of this hint is used at runtime, so it can be changed at any time.
 */
#define SDL_HINT_ANDROID_TRAP_BACK_BUTTON "SDL_ANDROID_TRAP_BACK_BUTTON"
```

Looks like it may have been set to 1 in p4a here: https://github.com/kivy/python-for-android/pull/1528 though I'm not aware of how that all interacts with pygame.

robertpfeiffer:
> Possibly relatedly, do we also need to be aware of this SDL hint?

AFAIK, p4a sets that setting by default, so there's nothing here to document. I'll look into p4a when I have time. As for multimedia keyboards, I don't know how often you use them with PyGame. I thought the more important use case for that constant is android, but if some people can generate it on PC, we probably should document that you can't rely on this event to only happen on Android. iOS doesn't have a back button, and the "back" button on the xbox gamepad doesn't trigger this event on PC.

Once this PR and #2250 are merged, I'll start working on PyGame+openGL on android, and then I'll write a couple of tutorials that focus on the *code* side of writing PyGame games for android. (Installing the tool chain is a completely different beast. It should be easy enough if you know your UNIX command line, but it is definitely orders of magnitude more difficult than `pip3 install python-for-android` and thus not beginner-friendly.) Having the back button working would be a prerequisite for these tutorials. If this key code can happen on the PC, I'll add that too.

In my own games, I currently write 

```python
import pygame
pygame.init()
pygame.K_AC_BACK=1073742094
```

but that's not a long-term solution and not something I'd want to use in tutorials. 

To fix the build on SDL1.2, can I just declare `pygame.K_AC_BACK=1073742094`, even if it's never used?

robertpfeiffer:
I managed to find a back button on my multimedia remote that indeed creates this keycode/scancode. I don't have a keyboard that has this key, but some people might.

I still think we should mention Android somewhere, as the main reason this is included is Android, and Escape would be used for the same purposes on PC platforms. I'll change the description.

illume (approved):
👍 Danke tausenddreihundertsiebenunddreißig!

ArrowLiu234:
Anyone knew how to get the key code of volume up/down in android? 
I get a little hope when i see the 'K_AC_BACK=1073742094'.

Issue #2191, opened 2020-10-16
default pygame icon not shown on Macos


illume:
- pygame==2.0.0.dev14 from pip shows the icon in the system tray, and when alt+TABing (very pixelated)
- pygame built from git with homebrew SDL 2.0.12 does not show the icon

ps. on windows it shows the default python icon in the task bar, but the little app icon is there in the top left of the window (this is how it was on pygame 1.9.6 too though, so it's not a regression).

ankith26:
https://github.com/pygame/pygame/blob/eca4b24e2765060cda209200eea594d2545af5e6/src_c/display.c#L1414

I’m curious why do you have so some different stuff with Darwin here, what happens if the if statement here is removed

illume:
@ankith26 good catch.

PR #2313, opened 2020-11-09
Fix pyinstaller hooks for Mac OS

ankith26:
Just a minor fix to ensure that the default pygame icon loads with pyinstaller on Mac OS

### DIFF
--- a/src_py/__pyinstaller/hook-pygame.py
+++ b/src_py/__pyinstaller/hook-pygame.py
@@ -26,10 +26,10 @@ def _append_to_datas(file_path):
 
 # First append the font file, then based on the OS, append pygame icon file
 _append_to_datas("freesansbold.ttf")
+_append_to_datas("pygame_icon.bmp")
 if platform.system() == "Darwin":
     _append_to_datas("pygame_icon.tiff")
-else:
-    _append_to_datas("pygame_icon.bmp")
+
 
 if platform.system() == "Windows": 
     from PyInstaller.utils.hooks import collect_dynamic_libs


ankith26:
In case you want to test (I don’t have a mac, so this is all a theory)

1. Get pygame 2.0.0 on a Mac. Make a base pygame app, that does not set any icon. Run the app and you should see pygame pop up with the default icon
2. Now bundle this app with pyinstaller (use latest 4.x), the default icon might not be visible 
3. Build this PR, and the issue should hopefully get fixed

Starbuck5:
I can test this sometime today or tomorrow.

Starbuck5:
I can't replicate the issue you are referring to on 2.0.0.dev1. Did someone report this somewhere?

ankith26:
Nope, this issue was not reported anywhere. It’s something I figured out while going through the code

Starbuck5:
I'm pretty sure the mac version doesn't need the .bmp file, and all this PR does is add the .bmp file to mac pyinstaller builds.

illume:
It's best to only submit PRs if there has been an issue reported or you can test the issue yourself first. Otherwise we use up our limited attention.

ankith26:
I guess we must think a bit, before closing this PR. A while ago, illume noticed an issue on mac #2191 which was related to icon not loading properly, for which I suggested a fix to load the pygame_icon.bmp, and illume merged the fix in #2194 

So pygame_icon.bmp is actually needed, without this PR, #2191 will still persist with pyinstaller exe

MyreMylar:
Is it possible that this test wasn't working:

```
if platform.system() == "Darwin":
    _append_to_datas("pygame_icon.tiff")
```

So mac was already getting the `.bmp` with pyinstaller anyway?

Starbuck5:
I've got to log off for the night.

I'm not at my mac anymore, but I believe the pyinstaller build just had the .tiff and not the .bmp, and it definitely had the icon on the dock. I'm not sure why. Ankith, I see why you would look at the code and think it would fail.

MyreMylar:
I looked over the code just now and found there is a whole seperate path for the mac icon. See:

https://github.com/pygame/pygame/blob/55ac5530bf9790943ad39954e2abc07bf5a21c87/src_py/macosx.py#L9-L22

And:

https://github.com/pygame/pygame/blob/55ac5530bf9790943ad39954e2abc07bf5a21c87/src_c/sdlmain_osx.m#L221-L258


It looks like the code in `display.c` was designed to be skipped on mac and then this mac specific code would take over and stick the .tiff file in as the default icon instead. You can still see this branch in the SDL1 code here:

https://github.com/pygame/pygame/blob/55ac5530bf9790943ad39954e2abc07bf5a21c87/src_c/display.c#L1773-L1784

I'm not sure why it wasn't working before for illume under SDL2, but it might explain why the icon still works even without the bmp file in pyinstaller builds. Perhaps it was this check:

`RunningFromBundleWithNSApplication()`

ankith26:
Ok fine.
If no one is actually facing any problems, then it’s best to close this PR.

PR #2315, opened 2020-11-10
minor bugfix, image.load load jpeg correctly

ankith26:
related to #1516, but does not fix that issue. Just make sure that loading images with jpeg extension should work as expected (when file-like objects are passed)

Of course, this fix will only work when the other underlying issue is solved


illume (changes requested):
Needs tests.

ankith26:
I am closing this PR, because it does not fix anything (yet)

Issue #1778, opened 2020-05-18
Add unit test: image.load_extended()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)
[Draft guide on how to write a unit test for pygame](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/image_test.py#L551-L557
**Docs:** None.
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/imageext.c#L105-L236

bubo-py:
Hello! Is it still open? If so, i'll be happy to work on that one

yac1n3:
Hi ! I'm part of a group project an we'd like to work on this issue.

ankith26:
A unit test has been merged :) 
Closing the issue

PR #2324, opened 2020-11-13
adding unit test for image.load_extended()

kristofszentes:
I added a unit test for image.load_extended(), and some small images (smaller than 32x32 pixels) in the example/data folder to be able to test all the following formats: BMP, PNG, JPG, GIF (non-animated), TGA (uncompressed), TIF, XPM, PPM, PGM, as it was asked in [this previous attempt](https://github.com/pygame/pygame/pull/2014) of a such unit test.
[Here](https://github.com/pygame/pygame/issues/1778) is the issue related to this unit test.


test/image_test.py
```diff
@@ -806,13 +806,55 @@ def test_load_basic(self):
...
 
         # __doc__ (as of 2008-08-02) for pygame.image.load_extended:
 
         # pygame module for image transfer
 
-        self.fail()
+        #We test loading the following file types: bmp, png, jpg, gif (non-animated), tga (uncompressed), tif, xpm, ppm, pgm
+        #All the loaded images are smaller than 32 x 32 pixels
+
+        #test loading a bmp
```
illume, line 818: We only recently added support for using subTest inside the pygame test suite, but they're excellent! With them you can make the messages better if something fails. Also it allows the further parts to be tested if something at the top of the test fails (for example if the png loading fails all the subTests after that will run). Have a look at the docs for subTest: https://docs.python.org/3/library/unittest.html#subtests

Something similar to this... (untested):
```python
with self.subTest("test loading a bmp", filename=filename, expected_color=(...)):
    surf = pygame.image.load_extended(example_path(filename))
    self.assertEqual(surf.get_at((0,0)), expected_color)
```

It's also possible to loop over a data structure of [[filename, expected_color], ...]

illume (approved):
👍 thank you :) It's great to have tests for all those file formats now.

We could merge this in if you're happy with it. I left a note about subTest, which you may want to look into. Another minor thing is that the spacing in the colors could be improved to be pep8 friendly.

kristofszentes:
Thank you for your advice! 
I changed spacing for colors, added the use of subTest and a for loop going through all the tests.

illume:
Brilliant. Thanks again.

PR #2287, opened 2020-10-31
Add mouse.set_color_cursor()

ankith26:
Just adding a new function to set a cursor from a surface. 

Todo:
- [ ] Do some testing
- [ ] Write unittests
- [ ] Update docs (both rst files and header files)
- [ ] type hints??

PR #2331, opened 2020-11-16
Cursor object

### LINKED ISSUE #2098: mouse.get_cursor on SDL2

### LINKED ISSUE #2293: Cursor object
**Description**

I've been thinking about a way to incorporate the new SDL2 cursor features into the existing API, and I think the best way to do it would be to create a Cursor object, and have `mouse.set_cursor()` set up to take this object.

This allows us to add the new cursor types (system cursor and color cursor), without crowding the `mouse` module. (Relevant SDL docs: [SystemCursor](https://wiki.libsdl.org/SDL_CreateSystemCursor), [ColorCursor](https://wiki.libsdl.org/SDL_CreateColorCursor))

Having a Cursor object would also solve the lack of a `get_cursor()` method. We can store the relevant data from the last `set_cursor()` and return an equivalent Cursor object. See issue #2098 

**Backwards Compatibility**

I believe this could be made mostly backwards compatible by making the Cursor object interpretable as a sequence, just like the previous return from `mouse.get_cursor()`. The `set_cursor()` function could also cast its arguments into a Cursor, if it gets the old style of arguments ((w,h), (spotx,spoty), xormask, andmask)

I would recommend removing `mouse.set_system_cursor` from the docs to encourage adopting of this new system.

**Advantages of this system**
`pygame.mouse.get_cursor()` works again
`pygame.mouse.set_cursor(pygame.mouse.get_cursor())` will work regardless of the type of cursor.

**Implementation** 

I've been experimenting with implementing this for a couple days, and I've got it working.
What it does right now is it goes through a python function first, which calls the c function in mouse.c with a keyword argument regarding the type of cursor, allowing one function to handle all three different types.
However I think it would be better to have the python function parse the Cursor object and decide which c function to call. (So there would be a different c function for each type of cursor, i.e. `set_system_cursor`, `set_normal_cursor`, `set_color_cursor`)

**Examples**

    hand = pygame.cursors.Cursor(pygame.SYSTEM_CURSOR_HAND)
    print(hand) >>> Cursor<system_cursor(SYSTEM_CURSOR_HAND)>

    box = pygame.cursors.Cursor((10,10), pygame.Surface((20,20)))
    print(box) >>> Cursor<color_cursor(hotspot: (10, 10), surf: <Surface(20x20x32 SW)>)>

    arrow = pygame.cursors.Cursor(*pygame.cursors.arrow)
    print(arrow) >>> Cursor<normal_cursor(size: (16, 16), hotspot: (0, 0))>

    data, mask = pygame.cursors.compile(pygame.cursors.thickarrow_strings)
    thickarrow = pygame.cursors.Cursor((24,24), (0,0), data, mask)
    print(thickarrow) >>> Cursor<normal_cursor(size: (24, 24), hotspot: (0, 0))>

    pygame.mouse.set_cursor() can take any of these cursor objects, or the raw arguments to construct one.

    pygame.mouse.set_cursor(hand) == pygame.mouse.set_cursor(SYSTEM_CURSOR_HAND), for instance

    After calling that, pygame.mouse.get_cursor() returns a cursor object.
    print(pygame.mouse.get_cursor()) >>> Cursor<system_cursor(SYSTEM_CURSOR_HAND)>

    Since cursor objects emulate a python sequence, code that unpacks the output of pygame.mouse.get_cursor()
    will only break if they use another type of cursor.

    pygame.mouse.set_cursor(arrow)
    size, hotspot, data, mask = pygame.mouse.get_cursor()

Starbuck5:
Closes #2293    - Closes #2098 - Closes #2287 

It works! I tested it with python3 on windows and MacOS (I did my best to write the python code so it would work in python2 as well, but I wasn't able to test that)

Demo code in docs/reST/ref/code_examples. Tested in mouse_test and cursors_test,

Points of interest to improving this PR or testing it or things I'm concerned about:
- It looks like everything in the test for `get_cursor` is also tested in `set_cursor`, so it can probably be removed. (the set_cursor tests have a lot of get_cursor in them).
- Right now there is an undocumented `pygame.cursors.Cursor()` behavior - with 0 arguments it returns a "default" cursor, which is just a system variant arrow cursor.
- Cursor errors are inconsistent, some are caught when you create a cursor, others are caught when you try to set it. Should the Cursor object in python be more aggressive about checking for cursor correctness upfront?

**Edit:** Resolved points:
- I don't know how to test for memory leaks, and I did something that may cause them (I designed it not to, but I'm just scared of memory leaks) (I used Py_XDECREF and Py_INCREF at line 305 in mouse.c)
- `mouse.set_cursor` and `mouse.get_cursor` both are python functions that call into their respective c functions. I wasn't sure where to put these, so for now they are just stuck in \_\_init\_\_.py  **now cursors.py**
- `pygame.cursors` provides some premade cursors, those should probably be converted in `Cursor` objects rather than the tuples they are in currently. **done**


ankith26:
First off, nice work on the PR :) 👍 
Haven’t yet had a good look, but there are a few issues with this PR
1. The whole point of getting this cursors object is backwards compatibility, and I see that you have removed set_cursor and get_cursor from pygame.mouse, which kinda not helping on backward compatibility  
2. You have defined some functions in pygame base file, which might need to be put in some other file

Overall, great idea with the PR, and with a bit of reworking, this will be really good

docs/reST/ref/cursors.rst
```diff
@@ -130,4 +130,100 @@ The following strings can be converted into cursor bitmaps with
...
+
+   .. versionadded:: 2.0.1
+
+   .. ## pygame.cursors.Cursor ##
+   
 .. ## pygame.cursors ##
+
+Example code for creating and settings cursors. (Click the mouse to switch cursor)
+
+.. literalinclude:: code_examples/cursors_module_example.py
```
illume, line 229: I wonder if the example can go inside examples/ instead and still be included?

illume:
Can you please describe how the lifetime of the objects is supposed to work? 
- Especially with xormask/andmask.
- How are they allocated, and how are they freed?

src_c/mouse.c
```diff
@@ -298,6 +300,20 @@ mouse_set_cursor(PyObject *self, PyObject *args)
     SDL_SetCursor(cursor);
     SDL_FreeCursor(lastcursor);
 
+    //To make sure that the mask data sticks around, it has to have its ref count increased
+    //Conversely, the old data stored in cursor_data (if it is there) doesn't need to be around anymore
+    Py_XDECREF(cursor_data.xormask);
```
illume, line 305: Can you please describe how the lifetime of the objects is supposed to work?

- Especially with xormask/andmask.
- How are they allocated, and how are they freed?

src_c/mouse.c
```diff
@@ -390,11 +452,9 @@ static PyMethodDef _mouse_methods[] = {
     {"get_visible", mouse_get_visible, METH_NOARGS, DOC_PYGAMEMOUSEGETVISIBLE},
     {"get_focused", (PyCFunction)mouse_get_focused, METH_VARARGS,
      DOC_PYGAMEMOUSEGETFOCUSED},
-    {"set_cursor", mouse_set_cursor, METH_VARARGS, DOC_PYGAMEMOUSESETCURSOR},
-    {"set_system_cursor", mouse_set_system_cursor, METH_VARARGS, DOC_PYGAMEMOUSESETSYSTEMCURSOR},
-    {"get_cursor", (PyCFunction)mouse_get_cursor, METH_VARARGS,
-     DOC_PYGAMEMOUSEGETCURSOR},
-
+    {"set_system_cursor", mouse_set_system_cursor, METH_VARARGS, "set_system_cursor(constant) -> None\nset the mouse cursor to a system variant"},
```
illume, line 455: We should use the docs from the header files here, not inline ones. eg. DOC_PYGAMEMOUSESETCURSOR as before.

src_c/mouse.c
```diff
@@ -309,73 +325,119 @@ mouse_set_cursor(PyObject *self, PyObject *args)
...
+    if (!cursor)
         return RAISE(pgExc_SDLError, SDL_GetError());
-    }
+
     lastcursor = SDL_GetCursor();
     SDL_SetCursor(cursor);
     SDL_FreeCursor(lastcursor);
-#endif
+
+    cursor_data.type = 2;
```
illume, line 364: What is 2 here?

src_c/mouse.c
```diff
@@ -309,73 +325,119 @@ mouse_set_cursor(PyObject *self, PyObject *args)
...
+
+    if (!PyArg_ParseTuple(args, "i", &constant)) {
+        return NULL;
+    }
+
+    return _set_system_cursor(constant);
 }
-#else /* IS_SDLv1*/
+
+//this function isn't all of mouse.set_cursor, it goes through a python function first
```
illume, line 388: Would be good to point to where the python function lives.

src_c/mouse.c
```diff
@@ -309,73 +325,119 @@ mouse_set_cursor(PyObject *self, PyObject *args)
...
+    else if (w && h) {
+        return _set_bitmap_cursor(w, h, spotx, spoty, xormask, andmask);
+    }
+    else if (surfobj) {
+        return _set_color_cursor(spotx, spoty, surfobj);
+    }
+    return RAISE(PyExc_ValueError, "Invalid cursor format: no valid template found");  
+}
+
+//should the buildvalue 'O's be replaced by 'N's?
```
illume, line 424: What is this about?

src_c/mouse.c
```diff
@@ -309,73 +325,119 @@ mouse_set_cursor(PyObject *self, PyObject *args)
...
+static PyObject*
+mouse_get_cursor(PyObject *self)
+{
+    VIDEO_INIT_CHECK();
 
-    return Py_BuildValue("((ii)(ii)NN)", w, h, spotx, spoty, xordata, anddata);
+    if (cursor_data.type == 0) {
```
illume, line 430: What are these numbers? Maybe they can use CONSTANTS.

docs/reST/ref/code_examples/cursors_module_example.py
```diff
@@ -0,0 +1,42 @@
...
+surf.fill((120, 50, 50))        # and use that as your surface
+color = pygame.cursors.Cursor((20, 20), surf)
+
+cursors = [system, bitmap_1, bitmap_2, color]
+cursor_index = 0
+
+pygame.mouse.set_cursor(cursors[cursor_index])
+
+clock = pygame.time.Clock()
+while True:
```
illume, line 28: I've been trying to get all the examples to use pg, and quit in the following way...
```python
import pygame as pg
...
going = True
while going:
        if event.type == pygame.QUIT or (event.type == pg.KEYDOWN and e.key == K_ESCAPE):
            going = False
```

ankith26: Made a little edit your comment illume, if you don’t mind :)

illume:
Looks like a segfault on python 2 - https://travis-ci.org/github/pygame/pygame/jobs/743901160#L1993

src_c/mouse.c
Starbuck5: Yeah, the cursor_data type attribute just tells get_cursor() what data fields to return. Constants would be a good improvement.

src_c/mouse.c
Starbuck5: The problem is that when the docs in `mouse.rst` are changed, the `setup.py docs` system regenerates the headers, and I removed set system cursor from the docs. Therefore regenerating the mouse docs causes the MOUSE_SETSYSTEMCURSOR constant to be deleted.

ankith26: Probably not removing set_system_cursor from the docs might be a better option?

Starbuck5: set system cursor has been publically available outside of prereleases for only two weeks. I think it should be removed from the docs to so that they more simply guide people towards the new way of doing cursors.

docs/reST/ref/cursors.rst
Starbuck5: I'll give it a shot. Right now there is a cursors example in the examples folder, so the two would need to be merged together.

src_c/mouse.c
illume: Ok, thanks for the explanation. Seems it's good to leave it out of the docs.

docs/reST/ref/cursors.rst
illume: We could have two examples if it makes it a bit clearer. Or the old one can be replaced if it has no value anymore.

src_c/mouse.c
Starbuck5: Yes.

The xormask and andmask are received as pointers to python objects, and I decided to store them that way rather than the parsed data in c itself. (So I could just return the objects as received). The problem was, since python didn't know I was storing these objects, it garbage collected the data away, since if it sometimes saw 0 references to them. 

So the new xormask and andmask have their reference counts get manually increased by 1.
Similarly, the data that is being replaced, the old xormask and andmask pyobject pointers, aren't needed by pygame anymore so they can have their reference count decreased by one. If something else still needs them, they will still stick around since their reference count will reflect other references to them.

Having to explain this makes me realize that surfaces have the same bug, since they are stored similarly.

The issue is when things are written "inline" since then there are no other references to them after the function executes.

    pygame.mouse.set_cursor((10,10), pygame.Surface((20,20)))
    a = pygame.mouse.get_cursor()
    print(a) # fails

Starbuck5:
I don't know what to make of the mac build failure on travis.

src_c/mouse.c
Starbuck5: It's another reference count thing.

Objects can be returned with Py_BuildValue with the format "O" or "N". [https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue](https://docs.python.org/3/c-api/arg.html#c.Py_BuildValue)

Using "O" increases the reference count by 1, using "N" doesn't.
But the docs are technical enough I don't really understand the use cases that they suggest.

docs/reST/ref/cursors.rst
Starbuck5: The old one still has value, since it shows the compiling step from an actual collection of strings. 

I would like to get this merged first, then do that in a subsequent PR.

src_c/mouse.c
illume: Sounds like there is a reference count issue, I haven't had to look into it in detail yet.
Putting this code into a test would be a nice addition.

src_c/mouse.c
illume: Yeah, I guess cursors.py is the best.

If we had a mouse.py, and a _mouse.c then we could combine them into one. However, this has a further complication in that we need to handle deleting old mouse.so/dll files in the setup.py.

illume:
> I don't know what to make of the mac build failure on travis.

Don't worry about it. This seems to be a timeout on either the network or the drive.

docs/reST/ref/cursors.rst
illume: Let's leave it, and perhaps address it later.

illume:
Cool, seems the SDL1 issue and others are fixed.

There's a couple of reference counting issues, and moving the python code left to do.

src_c/mouse.c
Starbuck5: I tested it with "N" and it had intermittent segfaults, so looks like "O" it is.

src_py/cursors.py
```diff
@@ -31,8 +31,90 @@
...
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
+    def __hash__(self):
+        return hash(tuple([self.type] + [data for data in self.data]))
+
+    def __repr__(self):
+        if self.type == "system":
+            id_string = _cursor_id_table.get(self.data[0], "constant lookup error")
+            return "Cursor<system_cursor(" + id_string +")>"
```
illume, line 88: I know the rest of pygame doesn't do this well... but repr(obj) is supposed to try and make an object that is constructable. ie. `eval(repr(obj)) == obj`, failing that it has the "<>" surrounding it to say "this is not how you construct it but sort of is what it's about".

I'm wondering if keyword arguments would make this easier.

Starbuck5: I was trying to imitate other pygame objects like Rect or Surface but I put the "<>" in the wrong place. At the least I'll fix that, move it to this: `<Cursor(type: system, constant: SYSTEM_CURSOR_ARROW)>`

I don't know if a more broad `eval(repr(obj)) == obj` is practical for cursor objects, because they could contain surfaces, and while bringing all the data of a surface into a string and back is possible, it's not very pretty to look at.

illume: > <Cursor(type: system, constant: SYSTEM_CURSOR_ARROW)>
Looks good to me.

Starbuck5: I've made the changes locally, I just don't want to take up build minutes by pushing a commit.
I don't know how to skip ci on a commit.

illume: You can add the text in your commit:
```
[skip ci]
```

docs/reST/ref/mouse.rst
```diff
@@ -183,77 +183,33 @@ scroll, such as ``which`` (it will tell you what exact mouse device trigger the
 
 .. function:: set_cursor
 
-   | :sl:`set the image for the mouse cursor`
-   | :sg:`set_cursor(size, hotspot, xormasks, andmasks) -> None`
+   | :sl:`set the mouse cursor to a new cursor`
+   | :sg:`set_cursor(pygame.cursors.Cursor) -> None`
+   | :sg:`set_cursor(arguments to create a pygame.cursors.Cursor) -> None`
```
illume, line 188: I guess this:
```
    | :sg:`set_cursor(pygame.cursors.Cursor) -> None`
    | :sg:`set_cursor(arguments to create a pygame.cursors.Cursor) -> None`
```

Could be replaced by:
```
   | :sg:`set_cursor(pygame.cursors.Cursor) -> None`
   | :sg:`set_cursor(size, hotspot, xormasks, andmasks) -> Cursor`
   | :sg:`set_cursor(hotspot, surface) -> Cursor`
   | :sg:`set_cursor(constant) -> Cursor`
```

However, `set_cursor(constant)` isn't really clear here in this list what should be used. Perhaps there is a way to make that more clear.

Have you considered python kwargs instead?

```python
set_cursor(arrow=True)
set_cursor(ibeam=True)
set_cursor(wait=True)
set_cursor(crosshair=True)
set_cursor(waitarrow=True)
set_cursor(size_nwse=True)
set_cursor(size_we=True)
set_cursor(size_ns=True)
set_cursor(size=True)
set_cursor(no=True)
set_cursor(hand=True)
```

MyreMylar: I think that many key word arguments would make for a very large function definition in python. The sort of thing pylint would normally cry about I would think. Though maybe it doesn't matter so much in functions defined in C, or kword only arguments. I don't think I've ever seen a function with 11 key word only arguments though. 

Could maybe just change it to:

```
| :sg:`set_cursor(cursor_constant) -> Cursor`
```

So it's more clear what 'constant' means, then there is a list of cursor constants right below theses signatures in the docs so people should be able to connect the two.

illume: The benefit of kwargs:
- they can be documented (so mouse hover on the arg shows the docs for example)
- they work with command line completion
- they are listed in the function and not off in a separate namespace that you need to read the docs to reference.
- constants are more of a C API thing, and not very common in python code.
- NOT_SHOUTING is easier to type 
- easier to read `set_cursor(arrow=True)` VS `set_cursor(pygame.SYSTEM_CURSOR_ARROW))`

We chose the kwarg approach for the Window object after much debate and experimentation.

The other option is set_cursor_arrow() et al. However this makes a proliferation of methods.

illume: Another approach would be:
```
mouse.cursor.arrow = True
```

MyreMylar: Looks like the default pylint warning for too many arguments is at 6, if that is a concern for the linting plans in the future.

Also for the kwargs approach what happens if someone does:

```
set_cursor(arrow=True, ibeam=True, wait=True)
```
?

The Enum type would be perfect for this use case (a limited but large-ish number of options) I think, but sadly it's not in python 2.7 by default so is probably out for pygame:

image

Code:

```
from enum import Enum


class SystemCursor(Enum):
    hand = 0
    arrow = 1
    wait = 2
    crosshair = 3


def set_cursor(system_cursor: SystemCursor):
    print(system_cursor)


set_cursor(SystemCursor.arrow)
```

illume: > Looks like the default pylint warning for too many arguments is at 6, if that is a concern for the linting plans in the future.

It's not a worry, since adding a `#pylint: disable=too-many-arguments` annotation can be done.

pylint doesn't have reasoning for this rule... However, eslint gives a good description:

> "Functions that take numerous parameters can be difficult to read and write because it requires the memorization of what each parameter is, its type, and the order they should appear in. As a result, many coders adhere to a convention that caps the number of parameters a function can take."

This guideline doesn't apply for this function, since you only need to use one of them. Kwargs in python don't require a fixed order, or require all the arguments to be used. X11 famously has lots of arguments for its functions. However, they don't have default arguments, and you have to at the call site use all the arguments. The solution in C of limiting the arguments and putting them into flags is good for C, flags in C let you do something like kwargs.

However, In this case, people only call it with one argument since python has kwarg defaults. So the pylint too-many-arguments does not apply here. No memorization of ordering needed.

```python
set_cursor(arrow=True)
```

> Also for the kwargs approach what happens if someone does:
> set_cursor(arrow=True, ibeam=True, wait=True)

It's not likely that someone would do that, but if they do then it raises an error telling them to use only one option.

> set_cursor(SystemCursor.arrow)

I think this is similar to the constants - you need to find where SystemCursor is and how to use it.

TypeScript allows you to use simple string union/enums like 'arrow' | 'ibeam' | 'wait', but I'm not sure this has made it into python world yet.

Starbuck5: The idea of the cursor object is to standardize the now 3 different ways of making a cursor into one API.

Passing in arguments to create a system cursor, or any type of cursor, is like how you can pass a sequence of numbers to a method expecting a rect. Set_cursor accepts a cursor or a "cursor like object", just like a lot of the draw functions accept a rect or a "rect like object."

I don't think that separating one type of cursors into its own thing ( keyword arguments ) is a good step forward.

I suppose each type could have its own keyword argument, but having only keyword arguments would break backwards compatibility on `set_cursor'. So having 0 kwargs would default to bitmap cursor, but then color cursor and system cursor would need kwargs?

Using a constant from pygame is something you have to do in lots of other places, and I think it is the simplest way to keep each type of cursor on the same playing field.

illume: > I don't think that separating one type of cursors into its own thing ( keyword arguments ) is a good step forward.

Why?

Starbuck5: Because I think having a different style for one type than for the others increases complexity rather than reducing it.

    pygame.mouse.set_cursor((24,24), (8,8), xormasks, andmasks)
    pygame.mouse.set_cursor((5,10), pygame.Surface((20,20)))
    pygame.mouse.set_cursor(pygame.SYSTEM_CURSOR_WAIT)

    pygame.mouse.set_cursor((24,24), (8,8), xormasks, andmasks)
    pygame.mouse.set_cursor((5,10), pygame.Surface((20,20)))
    pygame.mouse.set_cursor(wait=True) #although this does look prettier

Also I think kwargs should be for additional information or arguments, for stuff you can mix and match, not for the entire substance of a call. Having a pattern where you can call one kwarg OR two different function signatures seems weird to me.

Anyways I have to go to sleep now, talk to you later.

illume: [Literal types](https://www.python.org/dev/peps/pep-0586/#abstract) let you specify valid inputs, which provides an elegant interface...

```python
SystemCursors = Literal["arrow", "ibeam", "wait", "crosshair", "waitarrow",
    "size_nwse", "size_we", "size_ns", "size", "no", "hand"]
def set_cursor(cursor: SystemCursors) -> None:
    pass

set_cursor("wait")
```

... except it doesn't seem to work in vscode default python setup (or sublime anaconda default python setup). If you enable mypy it does detect problems in vscode, but it doesn't provide a nice drop down list. I think pycharm has Literal support, but I don't know how good it is.

It's still pretty nice to look at.
```python
set_cursor("arrow")
```

illume:
The docs for cursor here are quite good: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

Should we consider using the names they use? It would be nice to share the same vocabulary, and the naming is a bit better in some cases (what is an 'ibeam' for example?).

Starbuck5:
Here's a comparison between the two terminologies:

| SDL Constant | Mozilla Equivalent | My Suggestion |
|--------------|--------------------|---------------|
| ARROW        | default            | arrow         |
| IBEAM        | text               | text          |
| WAIT         | wait               | wait          |
| CROSSHAIR    | crosshair          | crosshair     |
| WAITARROW    | progress           | waitarrow (arrow_wait?)    |
| SIZENWSE     | nwse-resize        | arrow_nwse    |
| SIZENESW     | nesw-resize        | arrow_nesw    |
| SIZEWE       | ew-resize          | arrow_we      |
| SIZENS       | ns-resize          | arrow_ns      |
| SIZEALL      |                    | arrow_all     |
| NO           | not-allowed        | no            |
| HAND         | pointer            | hand          |

For my suggestion, I usually sided with SDL, because the SDL constants were more clear about what you're getting exactly. For example, SDL's `WAITARROW` is equivalent to Mozilla's `progress`, and I sided with waitarrow because it's clearer what you're getting -- an arrow with the spinning circle of wait next to it.

For all of the SIZE / resize ones, I made new names. SIZENWSE isn't super descriptive, and nswe-resize gives the purpose of the cursor rather than the shape. (Also we don't want to imply that the cursor does anything with resizing). So my suggestions are things like arrow_nwse - it's an arrow pointing from northwest to southeast.

I sided with Mozilla for the IBEAM / text, although I love the I-beam as a shape, because people may think of that shape as a text cursor, so `text` may be the more descriptive term.

illume:
The main benefit of reusing CSS/JS/browser/GL terminology is that the knowledge can be reused. In a teaching context where people teach browser things and python things. We already share similarities in Colors with "#FF0000", and vectors similar to glsl ones. 

For me... (sprinkled with IMHO...) The benefit of using W3C names over SDL names is that they are much more common, whereas SDL C programming is relatively obscure. Using shared terminology over custom names outweighs the benefit of potentially more useful names.  The W3C ones are pretty good IMHO, whilst not describing the cursor appearance they are semantic. Each desktop system is slightly different, however the cursors are not alllll that different (no one is using a bunny rabbit as a wait icon... some use rainbow beach balls or 8th century AD timekeeping technology).

On the other hand, I've seen many times professional web developers not even know that you can change the cursor. I do see changing the cursor as a pretty easy, and fun lesson... but I'm not aware of people actually doing one with cursor changing.

ps. I like how the W3C specification groups the cursors together under different use cases:
https://drafts.csswg.org/css-ui-3/#cursor

illume:
We have these things left to decide...
- A1) using constants (`set_cursor(pygame.SYSTEM_CURSOR_WAIT)`
- A2) using strings with Literal types? (`set_cursor("pointer")`)
- B1) if A2, using the W3C browser terminology? (`set_cursor("pointer")`)
- B2) if A2, using SDL/Starbuck5 terminology (`set_cursor("arrow")`) [see table](https://github.com/pygame/pygame/pull/2331#issuecomment-734194774)

Any opinions @Starbuck5 @MyreMylar @ankith26 ?

docs/reST/ref/mouse.rst
MyreMylar: This is how it looks in PyCharm:

image


Which I would be pretty happy with since it seems fairly close to a Enum's behaviour in terms of autocomplete in editor. Seem to be and idea borrowed from Typescript to replace Enums there which is nice. 

Apparently they are new in the typing module in 3.8 but are in the other versions of python via typing_extensions. Looks like we are already using them in `pixelcopy.pyi` as well so I would endorse this route.

MyreMylar:
I vote A2 having looked into Literal types, seem to be an enum variant that we can actually use and suited to this situation. Plus it works neatly in PyCharm.

I think I would also side with using the web terminology for cursors since they are in W3C standards documentation. Generally I think pygame should be as approachable as possible to people coming from webdev these days as, like it or not, that is where a lot of people cut their teeth in coding these days. That means payign attention to things in javascript, Typescript, HTML and css is a good idea. I tend to lean that way in my own dev when naming stuff unless there is better 'video-games-domain' terminology in wide use already.

ankith26:
A2 looks like a better option because literals seem better than constants.

As to wether we choose B1 or B2, I’m okay with any.
I also like the idea of mixing both.

Starbuck5:
I personally would rather go with A1, but it seems like the consensus is A2 with browser terminology, so I'll get working on implementing that.

BastiHz:
Wouldn't A1 be better for autocompletion? This way the user can start writing pyfame.SYSTEM_CURSOR_ and then see a list of available values in the ide. 
Like with KEYDOWN or MOUSEMOTION.

A string only increases the risk for typos, in my opinion.

MyreMylar:
I think the difficulty with the straight constants under `pygame.` is that we now have a heck of a lot of them for all kinds of different places (keys, blend modes, events) and it's not always clear which ones go in which function.

illume:
As far as I know, none of the editors do tab completion with Literal yet. The type checking ones do check the value though.

MyreMylar:
Yeah it would be hard to typo a Literal in PyCharm unnoticeably because you get a 'yellow' level error popping up that looks like this:

image

`mypy` also picks up this sort of error. 

VSCode's default intellisense python extensions does not though. Confusingly Microsoft now appears to have a better python code type checking/intellisense project called `pylance` (formerly `pyright`) but it is not their default one for complicated internal reasons I don't quite understand.

Anyway, pylance does pick up the Literal type correctly and has a display similar to PyCharm's:

image

And if you enable 'basic' type checking in pylance's extension settings then it will also highlight the typo in red:

image

Starbuck5:
Hi y'all, commenting to keep this active.

I still want to go with the constants option.

MyreMylar mentioned it could be confusing which constants go where, because there are a lot of constants in pygame. Just to be clear, each pygame cursor constant starts with SYSTEM_CURSOR - you type in pygame.SYSTEM_CURSOR_ and it comes up with a list of options to autocomplete.

Also to be clear- the functionality we're trying to add already exists in the `set_system_cursor` function, using constants, that this aims to replace. (I.e. if we switch to literals we still have these constants floating around for `set_system_cursor` backwards compatibility)

I think a good reason to with constants is how often they are used within pygame. Key constants, display modes, etc all use constants.

On the clarity of the terminology used - this could be because I've been working on this - but I don't think the cursor terminology from SDL is hard to understand. There isn't much of a learning curve that using browser terminology would avoid, imo. Maybe a future docs enhancement will be to try to emulate some of the https://developer.mozilla.org/en-US/docs/Web/CSS/cursor keyword value table things, with images next to each cursor entry.

illume:
What does twitter think?

Screenshot 2020-12-13 at 09 39 59

illume:
There's very good arguments for constants, and also people seem to love em.
There's also some love for the browser terminology and strings with teachers.

@Starbuck5 How about we do this?
- merge this in (constants with SDL terminology)
- look at adding string support with W3C terminology and Literal types in a separate issue/PR (I can start this).
- other possible improvements mentioned here can also come later. I can add them to an issue for further discussion/dev.

Starbuck5:
Sounds good!

Other improvements that I will probably work on in the future for this are putting more polish into the documentation and merging the two different examples scripts into one good example of `pygame.cursors` functionality.

Also if we decide to go with deprecation errors I would like to put one of those into `set_system_cursor`.

MyreMylar:
Definitely in favour of adding the deprecation error here 👍

illume (approved):
👍 cool!

Issue #2328, opened 2020-11-14
IME Mac, hold to press accent characters not working


illume:
Hold pressing a key like 'e' brings up a list of accented characters normally. 
This isn't working. (Funnily it used to accidentally work with an older pre-release of pygame 2 that used an old SDLmain).

Probably this needs to be fixed inside SDL.

- There is a bug report on the SDL bugzilla "[macOS ApplePressAndHold IME](https://bugzilla.libsdl.org/show_bug.cgi?id=5071)".
- There's an old discussion here: https://discourse.libsdl.org/t/sdl-2-0-4-text-input-mode-on-osx/21731
- A more recent IME test app posted here: https://discourse.libsdl.org/t/current-status-of-ime-support/28062

robertpfeiffer:
Where did this selection pop up normally? Should this be controlled by `pygame.key.set_text_input_rect()`?

illume:
It's definitely an SDL issue.
> - There is a bug report on the SDL bugzilla "[macOS ApplePressAndHold IME](https://bugzilla.libsdl.org/show_bug.cgi?id=5071)".

robertpfeiffer:
> It's definitely an SDL issue.

I don't doubt you on this, I am just curious about how this used to work when it did.

galexandreg:
> > It's definitely an SDL issue.
> 
> I don't doubt you on this, I am just curious about how this used to work when it did.

This was one of my issues ( #1031 )...
I wonder if calling stop_text_input would have fixed it... But then how to deal with the fact that it's on by default?

robertpfeiffer:
I have configured an emoji IME with iBus and I will test/document how this works for on-screen keyboards and IME input on Linux

robertpfeiffer:
After playing around with ibus, mozc, and anthy, I understand how IME are supposed to work. I think I understand *why* this was deliberately disabled on mac, but I do not understand how this ever used to work properly. If you press and hold A, does it not first generate a text input event with "A"? Or is the text input event only generated when you release A after a short time? Did pressing A and then choosing Ä actually result in two characters being generated ("aä")?

We could enable it again, when the user explicitly starts text input, without breaking too many things, but wouldn't it get in the way of using keys in games for non-text things if pressing and holding popped up a menu?

[...]

PR #2333, opened 2020-11-17
text_input on by default, needed for pygame.TEXTINPUT

galexandreg:
While trying the new `pygame.TEXTINPUT` events, I found some things in the docs that I thought could be improved...
The pygame docs says that "normal" pygame.TEXTINPUT events are not dependent on `start_text_input()` and `stop_text_input()`, but they are according to the SDL docs. 

> This function will start accepting Unicode text input events in the focused SDL window, and start emitting SDL_TEXTINPUT and SDL_TEXTEDITING events.

https://wiki.libsdl.org/SDL_StartTextInput

Testing shows that normal pygame.TEXTINPUT events will stop when pygame.key.stop_text_input() is called, but the whole SDL TextInput thing is on by default, so there is no need to call start_text_input() for them to work.

Everything also works in the textinput.py example file when pygame.key.start_text_input() is removed.

But I don't understand much about the advanced `TEXTEDITING` / IME stuff... perhaps there's something I missed.

Thanks 😄

### DIFF
--- a/docs/reST/ref/key.rst
+++ b/docs/reST/ref/key.rst
@@ -365,11 +365,11 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
 
 .. function:: start_text_input
 
-   | :sl:`start handling IME compositions`
+   | :sl:`start handling Unicode text input events`
    | :sg:`start_text_input() -> None`
 
    Start receiving ``pygame.TEXTEDITING`` and ``pygame.TEXTINPUT``
-   events to handle IME.
+   events.
 
    A ``pygame.TEXTEDITING`` event is received when an IME composition
    is started or changed. It contains the composition ``text``, ``length``,
@@ -378,7 +378,7 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
    When the composition is committed (or non-IME input is received),
    a ``pygame.TEXTINPUT`` event is generated.
 
-   Normal ``pygame.TEXTINPUT`` events are not dependent on this.
+   Text input events handling is on by default.
 
    .. versionadded:: 2.0.0
 
@@ -386,13 +386,13 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
 
 .. function:: stop_text_input
 
-   | :sl:`stop handling IME compositions`
+   | :sl:`stop handling Unicode text input events`
    | :sg:`stop_text_input() -> None`
 
    Stop receiving ``pygame.TEXTEDITING`` and ``pygame.TEXTINPUT``
-   events to handle IME.
+   events.
 
-   Normal ``pygame.TEXTINPUT`` events are not dependent on this.
+   Text input events handling is on by default
 
    .. versionadded:: 2.0.0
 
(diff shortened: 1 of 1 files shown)


robertpfeiffer:
This is probably related to key.unicode emulation, and how we document this depends on whether we want to tell users that **going forward, you shouldn't use key.unicode** and use TEXTINPUT events instead, or whether we want to support not only key.unicode into all eternity, but also the mixing of `key.unicode` and `start_text_input()`. My opinion on this is pretty clear. I wish there was a way to turn backwards compatibility features like key.unicode emulation off, or to issue a warning when backwards compatibility features are used intermingled with new APIs.

If you don't run your games on an Android tablet or a Touchscreen PC with an on-screen keyboard (recent Ubuntu versions work really well in this configuration), you don't need to care about this, but I do so I do.

So I'd propose documenting the current behaviour, and probably also turning off key.unicode emulation once `pygame.key.stop_text_input()` or `pygame.key.stop_text_input()` are called.

robertpfeiffer:
I realise I didn't explicitly comment on this pull request: In its current form, I don't think this _enough_.

We should document what  pygame.key.start_text_input() and pygame.key.stop_text_input() _actually do_ on desktop environments like windows, GNOME, android, and mac os, and what kind of behaviour you can expect with different languages (human languages, not programming languages).

There are two parts to this:

1. When text editing starts and no keyboard is attached (on Windows 10, Android and Gnome 3 (Wayland or X11) at least), an on-screen-keyboard is displayed for the duration of the text editing. The on-screen keyboard might be displayed even when a physical keyboard *is* attached.

2. When your locale is set to a language with complex non-Latin glyphs, and you set the editing area with SDL2, a button might be shown near the text area to open an IME on even desktop platforms with a physical keyboard attached. IME is pretty simple with the latin alphabet, but it's also supposed to allow entering CJK glyphs when your locale is set to such a language, even when your keyboard is non-CJK like QWERTY or Dvorak. I have no idea how such an IME looks, but supposedly it can be a text field rendered by the OS that lets you enter pinyin or romanji with your QWERTY keyboard, which then sends the converted CJK glyphs to the game, or a nested menu to select a gyph, or handwriting recognition. I *think* for Chinese, pinyin is the most common, but this system is meant to be pluggable.

We probably also need some example code and a screenshot of how both this works, and expose all of the necessary functionality properly. I know PyGame currently does the on-screen keyboard part properly (so this should be mentioned in the docs), but may be struggling with the IME part, but in this case we should aim to fix IME instead and mention that IME might currently be broken, not remove all mentions of IME from the docs.

illume (approved):
👍 Thanks.

galexandreg:
I must say I'm a bit confused that the pr was merged, because after reading your comment I agree that some information could be added... oh well 🤔

> I know PyGame currently does the on-screen keyboard part properly (so this should be mentioned in the docs), but may be struggling with the IME part, but in this case we should aim to fix IME instead and mention that IME might currently be broken, not remove all mentions of IME from the docs.

That was not my intention... I mainly wanted to remove the part that said that the normal pygame.TEXTINPUT events were not dependent on `start_text_input()` and `stop_text_input()`, because they clearly are (perhaps the confusion was caused by the fact that it's on by default in the SDL)

But you're right that a lot more information could be added. I can't, unfortunately, test much of the IME stuff for now (I'm using a mac and it's not working... #2328 )

PR #2334, opened 2020-11-17
Add 'None' as a type option when making freetype fonts

MyreMylar:
Noticed this was missing while writing some unit tests using the default font. It matches the pattern used in font.pyi, the behaviour
described in the docs and the actual behaviour.


illume (approved):
👍 thanks

Issue #1951, opened 2020-06-15
Test updating sprite.py to use OrderedDict instead of regular dict 


YariKartoshe4ka:
For testing, i wrote some code:
```python
import pygame

class Sprite(pygame.sprite.Sprite):
    def __init__(self, id):
        super(Sprite, self).__init__()
        self.id = id

group = pygame.sprite.Group()

group.add(Sprite(1))
group.add(Sprite(2))
group.add(Sprite(3))

for sprite in group.sprites():
    print('{0} | {1}'.format(sprite.id, sprite))
```
If python 3 runs this code, we will see:
```
1 | <Sprite sprite(in 1 groups)>
2 | <Sprite sprite(in 1 groups)>
3 | <Sprite sprite(in 1 groups)>
```
Which means that the sprites were added in order, but if this code run python 2, we'll see a different
```
2 | <Sprite sprite(in 1 groups)>
1 | <Sprite sprite(in 1 groups)>
3 | <Sprite sprite(in 1 groups)>
```
Why is this happening, and can it be fixed?
Python-3.6.9 | Python-2.7.17
Pygame 2.0.0-dev10

----------------------------------

TODO

- [ ] see if unit tests run changing the `self.spritedict = {}` and other parts inside src_py/sprite.py works.
- [ ] benchmark results from examples/testsprite.py (with different options... see source)
- [ ] run a few games that use sprite to see if there are breakages.
- [ ] consider only importing collections.OrderedDict for cpython 2.7.

MightyJosip:
This is because sprites are stored in dictionary so I can't guarantee the order (https://github.com/pygame/pygame/blob/4562396964cca81bceb2d12d3f09aa96746cb3fa/src_py/sprite.py#L301). You can also see this in docs (https://www.pygame.org/docs/ref/sprite.html#pygame.sprite.Group). I don't think anything can be done with it, but I will leave this open for now in case someone else wants to share thoughts about it

MyreMylar:
Specifically you will only notice this difference when using python 3.7+ as that was when python changed the basic dictionary to behave like an OrderedDict. See:

https://docs.python.org/3/library/collections.html#ordereddict-objects

It looks like OrderedDict's only exist from 3.1 onwards so if you want to keep supporting python 2 you can't rely on the order of objects in a dictionary/dictionary-like in your code.

[...]

PR #2336, opened 2020-11-17
Use orderedDicts in groups

zoldalma999:
Related to #1951
I will do benchmarks if the tests pass.

### DIFF
--- a/docs/reST/ref/sprite.rst
+++ b/docs/reST/ref/sprite.rst
@@ -213,8 +213,8 @@ Sprites are not thread safe. So lock them yourself if using threads.
        bool    test if any Sprites are contained
        iter    iterate through all the Sprites
 
-   The Sprites in the Group are not ordered, so drawing and iterating the
-   Sprites is in no particular order.
+   The Sprites in the Group are ordered only on python 3.6 and higher.
+   Below python 3.6 drawing and iterating over the Sprites is in no particular order.
 
    .. method:: sprites
 
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -93,14 +93,6 @@ from pygame.rect import Rect
 from pygame.time import get_ticks
 from pygame.mask import from_surface
 
-#check if we need to use orderedDict (see #1951)
-import platform
-import sys
-if sys.version_info < (3, 7, 0) and platform.python_implementation() != 'PyPy':
-    from collections import OrderedDict
-    USE_ORDEREDDICT = True
-else:
-    USE_ORDEREDDICT = False
 
 class Sprite(object):
     """simple base class for visible game objects
@@ -362,10 +354,7 @@ class AbstractGroup(object):
     _spritegroup = True
 
     def __init__(self):
-        if USE_ORDEREDDICT:
-            self.spritedict = OrderedDict()
-        else:
-            self.spritedict = {}
+        self.spritedict = {}
         self.lostsprites = []
 
     def sprites(self):
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -93,6 +93,14 @@ from pygame.rect import Rect
 from pygame.time import get_ticks
 from pygame.mask import from_surface
 
+#check if we need to use orderedDict (see #1951)
+import platform
+import sys
+if sys.version_info < (3, 7, 0) and platform.python_implementation() != 'PyPy':
+    from collections import OrderedDict
+    USE_ORDEREDDICT = True
+else:
+    USE_ORDEREDDICT = False
 
 class Sprite(object):
     """simple base class for visible game objects
@@ -354,7 +362,10 @@ class AbstractGroup(object):
     _spritegroup = True
 
     def __init__(self):
-        self.spritedict = {}
+        if USE_ORDEREDDICT:
+            self.spritedict = OrderedDict()
+        else:
+            self.spritedict = {}
         self.lostsprites = []
 
     def sprites(self):


illume:
Apart from writing some micro benchmarks, one way to do sprite benchmarks is with multiple runs of `python -m pygame.examples.testsprite`. Note, there are different command line options (sorry, it's not using a library for this, so viewing the source is the only way to get them). This lets you test with different sprite and group classes, and with different numbers of sprites.

It's probably good to also test with something else, like with a game that uses sprites.

MyreMylar:
This implementation seems very sensible to me. I believe regular python dicts will be slightly faster (from looking over other benchmarking data) but it's likely that if you care that much about speed _and_ are using sprites you will be using python 3.7+ and if you care about speed to that degree and are using python 2.7 you probably aren't using pygame's sprite module anyway.

Basically I feel like the consistent behaviour across versions will be more widely useful than slightly slowing down performance of older python 3 versions and 2.7 users using sprites.

I expect one day a C version of a sprite group could do some of the looping draw to blit parts a bit faster but that would likely be part of a wider sprite module refactor (that may become dominated by sorting out texture sharing and that sort of thing anyway).

illume:
Let's just benchmark it, and then we can see if the trade off is worth it.

zoldalma999:
I started the benchmarking, will try to finish it today.
Also using python 3.6.8 it seems like it is already sorted (in the original it also says that he used 3.6.9) so it might not be exactly 3.7.0. I will try to find the last version that does not sort it using normal dict.

MyreMylar:
If you read this hackernews thread:

https://news.ycombinator.com/item?id=22266173

it seems that the _implementation_ was changed for dictionaries to be insertion ordered for CPython in 3.6. But the change wasn't made to the actual python language spec until 3.7 So you could still have valid 3.6 python code with unordered dictionaries if it was running on a non-CPython interpreter.

illume:
Cool, thanks.

If you add a unit test for the sorting, then the CI can run on different versions of python to confirm the ordering.

There's some other dict uses in sprite.py that are likely affected by the same issue.

Some microbenchmark timings for dict assignment in python 3.8 on mac.
```python
In [1]: %timeit dict()
126 ns ± 1.16 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [2]: %timeit {}
23.5 ns ± 0.123 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: %timeit {}
23.5 ns ± 0.123 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: from collections import OrderedDict

In [5]: %timeit OrderedDict()
121 ns ± 0.351 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [6]: %timeit OrderedDict()
120 ns ± 0.597 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [7]: %timeit dict()
125 ns ± 0.654 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [8]: %timeit dict()
126 ns ± 1.23 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [9]: %timeit OrderedDict()
121 ns ± 0.64 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)


In [12]: import timeit

In [13]: s= """
    ...: if USE_ORDEREDDICT:
    ...:     a = OrderedDict()
    ...: else:
    ...:     a = {}
    ...: """

In [21]: timeit.timeit(s, setup="from collections import OrderedDict", globals={"USE_ORDEREDDICT":1})
Out[21]: 0.1377905770000325

In [22]: timeit.timeit(s, setup="from collections import OrderedDict", globals={"USE_ORDEREDDICT":0})
Out[22]: 0.04726486300000943

In [23]: timeit.timeit(s, setup="from collections import OrderedDict", globals={"USE_ORDEREDDICT":0})
Out[23]: 0.047848736999981156

In [24]: timeit.timeit("a={}", setup="from collections import OrderedDict", globals={"USE_ORDEREDDICT":0})
Out[24]: 0.02688187400002562

In [25]: timeit.timeit("a={}", setup="from collections import OrderedDict", globals={"USE_ORDEREDDICT":0})
Out[25]: 0.02763441999996985
```

zoldalma999:
- I used `pygame.examples.testsprite`
- I used two python versions(2.7.17, 3.4.4).
- I ran every test three times to be a bit more accurate
- Every number is the fps I got from the test
### Python 2.7.17:
| Flag(s)                      | With orderedDicts       | Without orderedDicts    |
|------------------------------|-------------------------|-------------------------|
|     No flag                   |     1858, 1869, 1857    |     2624, 2368, 2449    |
|     -noupdate_rects          |     1897, 1864, 1902    |     2068, 2063, 2066    |
|     -scaled                  |     1412, 1422, 1394    |     1615, 1586, 1604    |
|     -fullscreen              |     1912, 1856, 1874    |     2452, 2408, 2529    |
|     -alpha                   |     1827, 1839, 1819    |     2293, 2270, 2338    |
|     -scaled   -fullscreen    |     1409, 1422, 1407    |     1537, 1545, 1585    |

### Python 3.4.4
| Flag(s)                      | With orderedDicts       | Without orderedDicts    |
|------------------------------|-------------------------|-------------------------|
|     No flag                   |     1731, 1742, 1716    |     2193, 2199, 2178    |
|     -noupdate_rects          |     1746, 1715, 1695    |     1975, 1974, 1962    |
|     -scaled                  |     1316, 1311, 1307    |     1491, 1489, 1505    |
|     -fullscreen              |     1694, 1713, 1709    |     2130, 2155, 2170    |
|     -alpha                   |     1713, 1715, 1708    |     2086, 2039, 2097    |
|     -scaled   -fullscreen    |     1325, 1327, 1299    |     1479, 1447, 1470    |

MyreMylar:
Looks like the using the basic dicts runs about 10% - 30% faster on these tests. Not super great.

Perhaps if we do go ahead with this we should also add an environment variable that allows users to override USE_ORDEREDDICT. Something like `PYGAME_ALWAYS_USE_ORDERED_SPRITE_DICT = '0'` and default it to '1'?

illume:
Thanks for testing that @zoldalma999.

I'd like to repeat the tests on some slower gfx hardware, and another sprite using game just to be sure. I'll get around to it in a few days.

If it's a similar result, I'd suggest we don't make the change. Instead, people who care could use a mixin class or a monkey patch.

zoldalma999:
If we dont add orderedDicts i would still edit the docs so it tells you which python version you can use to have an ordered group since currently this is whats in the docs:
>The Sprites in the Group are not ordered, so drawing and iterating the Sprites is in no particular order.

Perhaps something like this
>The Sprites in the Group are ordered only on python 3.7 or higher. Below python 3.7 drawing and iterating over the Sprites is in no particular order.

would be nicer.

illume:
I think we should not include this change, just because of the noticeable performance penalty. The doc change is a good idea. I'll also note in the issue that a subclass of the group that uses an ordered dict can be used for people who want to rely on ordering in cpython 2.7.

> The Sprites in the Group are ordered only on python 3.6 and higher. Below python 3.6 drawing and iterating over the Sprites is in no particular order.

Yeah, that text would be nicer. Note however, I think we should use 3.6. Since although it wasn't guaranteed as part of a specification, 3.6 was ordered.

If you agree, we can close this PR and I'll make the notes in the issue about the changes. If you instead think the benefit of ordering outweighs the performance penalty I'm open to including the PR. Can I leave the decision to you @zoldalma999 ?

illume (approved):
👍

zoldalma999:
I think either we should ditch it or make it optional and default to off. What MyreMylar suggested seems like a good way if we want to chose the latter.

illume:
Ok. Well, ditching it seems the easier option for now :)

illume:
Thanks @zoldalma999

PR #2338, opened 2020-11-18
WIP: emulate quit event

robertpfeiffer:
Add an environment variable to generate a QUIT event when alt+f4 are pressed. This has a kind of niche use case: Allowing a user to quit out of a game running without a window manager, directly on the framebuffer or on dispmanx/RPi. This is not needed when a game lets the player quit through the main menu, but quite useful to have in your .profile when you test compatibility of games on raspberry Pi. Maybe there is a hint to get SDL to do this by itself.

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -72,6 +72,7 @@ static int pg_key_repeat_interval = 0;
 static SDL_TimerID _pg_repeat_timer = 0;
 static SDL_Event _pg_repeat_event;
 static SDL_bool  _pg_event_generate_videoresize = SDL_TRUE;
+static SDL_bool  _pg_event_emulate_quit = SDL_FALSE;
 
 static Uint32
 _pg_repeat_callback(Uint32 interval, void *param)
@@ -122,6 +123,10 @@ pgEvent_AutoInit(PyObject *self, PyObject *args)
 #endif /* IS_SLDv2 */
         pg_RegisterQuit(_pg_event_cleanup);
         _pg_event_is_init = 1;
+    }
+
+    if (SDL_getenv("PYGAME_EMULATE_QUIT")){
+        _pg_event_emulate_quit=SDL_TRUE;
     }
 
     return PyInt_FromLong(_pg_event_is_init);
@@ -231,6 +236,13 @@ pg_event_filter(void *_, SDL_Event *event)
 #pragma PG_WARN(Add event blocking here.)
 
     else if (type == SDL_KEYDOWN) {
+
+        if(_pg_event_emulate_quit && event->key.keysym.sym==SDLK_F4
+           && (event->key.keysym.mod  & KMOD_ALT)) {
+            SDL_Event quit_event;
+            quit_event.type = SDL_QUIT;
+            SDL_PushEvent(&quit_event);
+        }
 
         if (event->key.repeat) {
             return 0;


illume:
Sounds useful. Probably something SDL folk would be interested in as well.

ankith26 (approved):
Nice one :)
We can probably merge this from our end. Because if we wait for this to come with SDL 
1) It will take a lot of time
2) Pygame with old versions of SDL will not get this feature

robertpfeiffer:
> Sounds useful. Probably something SDL folk would be interested in as well.

We'd probably need to modify `src/core/linux/SDL_evdev.c`.

In `src/video/directfb/SDL_DirectFB_events.c`, there's already code to generate window close events, which should generate a quit event as well, In the directFB source code it looks like ctrl+C should to the trick.

Not sure yet how kmsdrm fits into this.

MyreMylar:
Just looking at this briefly, I noticed that SDL does generate QUIT events on windows when you press ALT + F4, and has since roughly 2013:

https://github.com/libsdl-org/SDL/commit/5b8b403e3e6cc5a9d180c632ffc01c141f461773

So I guess this would generate two quit events on windows.

dr0id:
Tested on windows 10, on this branch. And indeed, there are two quit events generated as demonstrated with following code when pressing Alt+F4:

```
import os

env_var = 'PYGAME_EMULATE_QUIT'
os.environ[env_var] = "1"

print("env_var", os.environ[env_var])

import pygame

print("init..")

pygame.init()

print("init done")

clock = pygame.time.Clock()
running = True
while running:
	clock.tick(50)
		print(event)

		if event.type == pygame.QUIT:
			running = False

pygame.quit()
```

Pressing Alt+F4 generates this output:

```
...>py -3 quit_test.py
env_var 1
pygame 2.0.1.dev1 (SDL 2.0.12, python 3.9.7)
init..
init done
<Event(4352-AudioDeviceAdded {'which': 0, 'iscapture': 0})>
<Event(4352-AudioDeviceAdded {'which': 0, 'iscapture': 1})>
<Event(512-WindowEvent {'event': 1, 'window': None})>
<Event(32768-ActiveEvent {'gain': 1, 'state': 1, 'window': None})>
<Event(512-WindowEvent {'event': 12, 'window': None})>
<Event(770-TextEditing {'text': '', 'start': 0, 'length': 0, 'window': None})>
<Event(32768-ActiveEvent {'gain': 1, 'state': 0, 'window': None})>
<Event(512-WindowEvent {'event': 10, 'window': None})>
<Event(1024-MouseMotion {'pos': (186, 68), 'rel': (0, 0), 'buttons': (0, 0, 0), 'window': None})>
<Event(32768-ActiveEvent {'gain': 0, 'state': 0, 'window': None})>
<Event(512-WindowEvent {'event': 11, 'window': None})>
<Event(32768-ActiveEvent {'gain': 1, 'state': 0, 'window': None})>
<Event(512-WindowEvent {'event': 10, 'window': None})>
<Event(1024-MouseMotion {'pos': (186, 68), 'rel': (0, 0), 'buttons': (0, 0, 0), 'window': None})>
<Event(32770-VideoExpose {})>
<Event(512-WindowEvent {'event': 3, 'window': None})>
<Event(768-KeyDown {'unicode': '', 'key': 1073742050, 'mod': 256, 'scancode': 226, 'window': None})>
<Event(512-WindowEvent {'event': 14, 'window': None})>
<Event(256-Quit {})>
<Event(256-Quit {})>
<Event(768-KeyDown {'unicode': '', 'key': 1073741885, 'mod': 256, 'scancode': 61, 'window': None})>
```

robertpfeiffer:
Yeah, this is specifically a raspi problem that happens when you run a game on raspi or directly on KMSDRM without a display server that sends window events and suddenly your PyGame app can't be quit because it reacts to quit events that can't be sent. It's probably more appropriate to handle this on the SDL side.
This is not and never was a problem on windows, mac os, or desktop Linux with X11 or wayland.

PR #2340, opened 2020-11-18
test convert()d surfs keep SRCALPHA flag, and get_alpha at 255.

illume:
A compatibility issue found with solarwolf: https://github.com/pygame/pygame/issues/721#issuecomment-729575131

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1198,6 +1198,21 @@ class GeneralSurfaceTests(AssertRaisesRegexMixin, unittest.TestCase):
         finally:
             pygame.display.quit()
 
+    def test_convert_get_alpha(self):
+        """ converted()d surfaces loaded from png transparent style surfaces should return 255.
+        """
+        pygame.display.init()
+        try:
+            pygame.display.set_mode((640, 480))
+
+            s1 = pygame.Surface((100, 100), pygame.SRCALPHA, 32)
+            s1_converted = s1.convert()
+            self.assertEqual(s1_converted.get_flags() & SRCALPHA, SRCALPHA)
+            self.assertEqual(s1_converted.get_alpha(), 255)
+        finally:
+            pygame.display.quit()
+
+
     def test_src_alpha_issue_1289(self):
         """blit should be white."""
         surf1 = pygame.Surface((1, 1), pygame.SRCALPHA, 32)


test/surface_test.py
```diff
@@ -1198,6 +1198,21 @@ def test_convert_alpha_SRCALPHA(self):
         finally:
             pygame.display.quit()
```
dr0id, line 1200: - [ ] line 16 can be removed completely
- [ ] lines 771: 
```
        s.get_view(as_unicode("2"))
        s.get_view(as_bytes("2"))
```
change to: 
`        s.get_view("2")`

test/surface_test.py
```diff
@@ -1198,6 +1198,21 @@ def test_convert_alpha_SRCALPHA(self):
...
+    def test_convert_get_alpha(self):
+        """ converted()d surfaces loaded from png transparent style surfaces should return 255.
+        """
+        pygame.display.init()
+        try:
+            pygame.display.set_mode((640, 480))
+
+            s1 = pygame.Surface((100, 100), pygame.SRCALPHA, 32)
+            s1_converted = s1.convert()
+            self.assertEqual(s1_converted.get_flags() & SRCALPHA, SRCALPHA)
```
dr0id, line 1210:
```suggestion
            self.assertEqual(s1_converted.get_flags() & SRCALPHA, 0)
```

The converted surface does not use SRCALPHA blends anymore or does it?
According to docs:
`SRCALPHA       0x00010000    # Blit uses source alpha blending`

test/surface_test.py
```diff
@@ -1198,6 +1198,21 @@ def test_convert_alpha_SRCALPHA(self):
...
+            self.assertEqual(s1_converted.get_alpha(), 255)
```
dr0id, line 1211: This currently fails using 'pygame 2.1.3.dev5 (SDL 2.0.20, Python 3.9.7)', py x64 on Win10

dr0id (changes requested):
This PR seems rather old. Either finish it or set it to draft. Or close it if there is no intention to merge.

test/surface_test.py
dr0id: I this flag managed by SDL?

test/surface_test.py
dr0id: Is this managed by SDL?

PR #2342, opened 2020-11-18
fix typo in windib driver handler condition

pmp-p:


ankith26 (approved):
Nice catch :)
I wonder how that one got missed.
Thanks for the PR 👍

ankith26:
I think it’s safe to merge this, because it’s a good fix, and no one should have any problems with this fix. 
I will skip CI on this merge, because the CI has aldready passed on this PR

PR #2344, opened 2020-11-18
Make SysFont and match_font accept iterables

### LINKED ISSUE #2341: SysFont names as list documentation
Hi, this is a minor change to the documentation. However, I wanted to first ask here before making a pull request because maybe it's just me who reads it the wrong way.

This concerns both pygame.font.SysFont and pygame.freetype.SysFont. The [freetype.SysFont documentation](https://www.pygame.org/docs/ref/freetype.html#pygame.freetype.SysFont) states "The font name can be a comma separated list of font names to search for.". I interpret this as an actual list object with font name strings as entries. But if I try to use it like that it causes this error:

```py
import pygame.freetype
pygame.init()
f = pygame.freetype.SysFont(["inconsolata", "consolas", "monospace"], 12)
```
```
  File "/home/sebastian/Code/venv/lib/python3.8/site-packages/pygame/freetype.py", line 77, in SysFont
    return _SysFont(name, size, bold, italic, constructor)
  File "/home/sebastian/Code/venv/lib/python3.8/site-packages/pygame/sysfont.py", line 375, in SysFont
    for single_name in name.split(','):
AttributeError: 'list' object has no attribute 'split'
```

The correct way is to use a **string** of comma-separated font names: `"inconsolata, consolas, monospace"`.

I propose changing the sentence to "The font name can be a string of comma-separated font names to search for."

BastiHz:
Previously, the `name` argument to `font.SysFont()`, `freetype.SysFont()` and `font.match_font()` accepted multiple font names in the form of a string or bytes object of comma-separated names. Now it also accepts an iterable of names.

In this example all resulting fonts are the same.
```py
import pygame.freetype
pygame.init()
pygame.freetype.SysFont("inconsolata", 12)  # single string
pygame.freetype.SysFont("inconsolata, consolas", 12)  # comma-separated string
pygame.freetype.SysFont(["inconsolata", "consolas"], 12)  # list of strings
pygame.freetype.SysFont((name for name in ["inconsolata", "consolas"]), 12)  # string generator
pygame.freetype.SysFont(b"inconsolata", 12)  # single bytes object
pygame.freetype.SysFont(b"inconsolata, consolas", 12)  # comma-separated bytes
pygame.freetype.SysFont([b"inconsolata", b"consolas"], 12)  # list of bytes objects
pygame.freetype.SysFont([b"inconsolata", "consolas"], 12)  # mixed list
```

I updated the documentation. Doing that I noticed that there are some small inconsistencies between font and freetype documentation. In freetype the arguments are in italics. I tried to emulate the respective styles.

I also updated the type stubs. It's funny that `SysFont()` in freetype.pyi already specified `List[str]` as an acceptable type for the `name` argument although that was not implemented.

New tests are also included. Tested on Kubuntu 20.04.1 with Python 3.8.5.

closes #2341


BastiHz:
Don't know why that one check failed. If I read the log correctly it doesn't seem to be caused by my changes.

ankith26 (approved):
This PR looks good to me.
The failing tests are most likely not related to this PR
Thanks :) 👍

BastiHz:
Agreed. I will also make the type more generic so it accepts any sequence, not just list.

illume:
I guess this also supported bytes before too?
```python
>>> b'asdf,qwer'.split(',')
['asdf', 'qwer']
>>> isinstance(b'', (bytes, str)), isinstance('', (bytes, str))
(True, True)
```

BastiHz:
@illume Looks like your code snippet works only in Python 2. Because when I run it in Python 3 I get bytes instead of strings:
```py
>>> b'asdf,qwer'.split(b',')
[b'asdf', b'qwer']
```

I now made it that bytes are acceptable, either as a single name, a bytes object of comma-separated names or an iterable of bytes. The commits will be pushed later.

BastiHz:
I updated the top post.

While working on this feature I noticed that most of the logic in `SysFont()` and `match_font()` is duplicated. Maybe it would be a good beginner issue to extract the shared part into a helper function. I did not attempt this because that would be out of scope for this pull request.

test/font_test.py
```diff
@@ -113,6 +113,68 @@ def test_match_font_all_exist(self):
...
+        self.assertTrue(pygame_font.match_font(names) is None)
+
+        # Check generator:
+        names = (name for name in [not_a_font, font, not_a_font])
+        font_path_2 = pygame_font.match_font(names)
+        self.assertEqual(font_path_2, font_path)
+        names = (name for name in [not_a_font, not_a_font, not_a_font])
+        self.assertTrue(pygame_font.match_font(names) is None)
+
+        fonts_b = [f.encode() for f in font]
```
ankith26, line 150: Here, font variable is a string that contains the name of the font. By iterating over it and encoding it, you are encoding each character and putting it as a single element in the list. 
So fonts_b looks like 
`[b's', b'a', b'n', b's']`

Which is why CI is failing (I have no clue how pypy seems to pass)

BastiHz: Thanks! Somehow missed that when testing on my machine.

illume (approved):
👍 great work. Thank you very much.

PR #2345, opened 2020-11-18
Rejigs the alpha blitter path logic to hopefully cover issue on mac

MyreMylar:
This is an experimental, untested PR because I don't have a mac. It was reported that macs had an issue in solarwolf with the shadowed text that wasn't present in 2.0.0. I'm not sure exactly what is the underlying cause but I think rejigging the logic
this way should cover all cases that were caught in 2.0.0 without losing the recent fix. Probably. The surface unit tests pass locally at least.

Hopefully @illume can try this at some point and see if it works.

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -253,13 +253,18 @@ SoftBlitPyGame (SDL_Surface * src, SDL_Rect * srcrect, SDL_Surface * dst,
                             {
                                 alphablit_alpha_sse2_argb_surf_alpha (&info);
                             }
-                            else if (!SDL_ISPIXELFORMAT_ALPHA(dst->format->format))
-                            {
-                                alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(&info);
-                            }
                             else
                             {
-                                alphablit_alpha_sse2_argb_no_surf_alpha (&info);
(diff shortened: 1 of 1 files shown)


MyreMylar:
I linked the solarwolf issue but just to note this won't do anything about the power up `surface.convert()` issue.

illume:
Nice. Can confirm it fixes the `textshadow` solarwolf issue on my Mac.

I haven't gotten around to making a test case for it yet, but will do so.

illume:
I feel weird with changes like this going in without a test. I started on one here: https://github.com/pygame/pygame/issues/721#issuecomment-730614241 but it got a bit tiring and I've ran out of time for now. It's nice that it passes the existing tests, but it won't stop things breaking in the future. But I feel like it's going to take quite a while to nail down all the issues, and merging this in would give a nice fix (and allow another release with the other fixes).

MyreMylar:
If it helps at all, I've tested it locally with both solarwolf, Flyboy (which had the problem with 2.0.0) and my own pygame_gui examples on windows, and everything seems fine.

Perhaps we should just leave the solarwolf issue open so we can figure out what the test should be and then merge this for now anyway.

illume:
> Perhaps we should just leave the solarwolf issue open so we can figure out what the test should be and then merge this for now anyway.

Yeah, seems like a plan :)

PR #2349, opened 2020-11-22
Fix `no-else-break` and `no-else-return` signaled by pylint

Pierre-Sassoulas:
When there is a return, break or raise inside a conditional, the following code can be simplified.


illume (approved):
👍 thanks

Issue #2300, opened 2020-11-05
Travis pricing change, limited build minutes, adjustments needed, github actions?


illume:
https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

TLDR;
- time limit added to builds (1000 minutes)
- mac builds will be very limited
- unsure when the new plan starts for the pygame organization. They say they are introducing it incrementally.
- unsure where to see used minutes in the travis ci interface.

Actions to consider:
- [ ] move mac builds to github actions.
- [ ] move manylinux builds to github actions.
- [ ] move sdist build?
- [ ] pypi uploads working (could be on appveyor or github)
- [ ] github uploads working (could be on appveyor or github)

ankith26:
Well, thats bad news. Ok, I will start using [skip ci] from now

illume:
Committing every change is fine, maybe just consider pushing it only when you're at the end of coding a session (or even in some days after).

ankith26:
Well, now that I’m using git, things will be better, I used to use the github web interface, so even to like change something in another file, I would have to do a separate commit

zoldalma999:
I dont know how many people read it but i think we should add a note in [Hacking](https://www.pygame.org/wiki/Hacking) to use [skip ci] if they are just changing the documentation or do it on their own fork if possible (and possibly redirect to a link which explains how to do that).

ankith26:
Just noticed (in my latest commit), travis has stopped running (even though I forgot to use [skip ci] )

MyreMylar:
Hmm, read through their blog and it does sound bad for pretty much all big open source projects using Travis CI. 

Big shove to move all their free users over to Github Actions I suppose? I'll have a poke around the other similar projects to pygame and see if they are moving off Travis already (if they were on it)

MyreMylar:
Looks like numpy recently made the jump to github actions. See this PR:

https://github.com/numpy/numpy/pull/17738

As did Pillow in this issue and the linked PRs:

https://github.com/python-pillow/Pillow/issues/5028

Looks like a fairly wide consensus to ditch Travis from those projects that were using it.

ankith26:
What’s the current situation with this issue, should we still be doing [skip ci] on our commits??

illume:
Seems like they are shutting things down on December 31st now, and it appears there's still no reporting of used minutes. So I guess we are maybe ok until then (it's still unclear though).

[...]

PR #2352, opened 2020-11-22
Simplify test match font name

Pierre-Sassoulas:
Everything was a comparison to None so using `assertIsNone(x)` directly instead of `assertIsTrue(x is None)` makes for better error messages in case of test failure and also permits to use a list of values.

### DIFF
--- a/test/font_test.py
+++ b/test/font_test.py
@@ -114,8 +114,10 @@ class FontModuleTest(unittest.TestCase):
             self.assertTrue(os.path.isabs(path))
 
     def test_match_font_name(self):
-        """that match_font accepts names of various types"""
+        """That match_font accepts names of various types"""
         font = pygame_font.get_fonts()[0]
+        font_path = pygame_font.match_font(font)
+        self.assertIsNotNone(font_path)
         font_b = font.encode()
         not_a_font = "thisisnotafont"
         not_a_font_b = b"thisisnotafont"
@@ -135,6 +137,15 @@ class FontModuleTest(unittest.TestCase):
             # Check mixed list of bytes and string.
             [font, not_a_font, font_b, not_a_font_b],
         ]
+        for font_name in good_font_names:
+            self.assertEqual(
+                pygame_font.match_font(font_name), font_path, font_name
+            )
+
+    def test_not_match_font_name(self):
+        """match_font return None when names of various types do not exist"""
+        not_a_font = "thisisnotafont"
+        not_a_font_b = b"thisisnotafont"
         bad_font_names = [
             not_a_font,
             ",".join([not_a_font, not_a_font, not_a_font]),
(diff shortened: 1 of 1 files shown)


illume:
Maybe each of the blocks should be in a separate method or subtest? If one of the assert fails, it would be easier to debug and run the test case in isolation.

Pierre-Sassoulas:
Blocked (and inspired) by #2353  :)

Pierre-Sassoulas:
@illume, thank you for your review, I took your remark into account and rebased since the master branch tests are passing again :)

test/font_test.py
```diff
@@ -114,66 +114,50 @@ def test_match_font_all_exist(self):
...
-        self.assertTrue(pygame_font.match_font(names) is None)
-
-        # Check mixed list of bytes and string.
-        names = [font, not_a_font, font_b, not_a_font_b]
-        font_path_2 = pygame_font.match_font(names)
-        self.assertEqual(font_path_2, font_path)
-        names = [not_a_font, not_a_font_b, not_a_font]
-        self.assertTrue(pygame_font.match_font(names) is None)
+    def test_not_match_font_name(self):
+        """match_font return None when names of various types do not exists"""
```
MyreMylar, line 146: Small typo/spelling errors here, should be:

`"""match_font returns None when names of various types do not exist"""`

despite names being plural 'exist' should not be. Don't ask me, English makes no sense.

MyreMylar (changes requested):
Looks good to me, bar that small spelling/plural thing in the comment. The tests run and seem to do what the did before in a more condensed fashion. 👍

test/font_test.py
Pierre-Sassoulas: Fixed :)

MyreMylar (approved):

Pierre-Sassoulas:
The issue with the CI is unrelated to the last changes, I just removed a "s" in a docstring. What should I do ? Wait for something to be merged in `main` and rebase on it so the test are launched again?

ankith26:
If you ever want to re-run the CI, closing and opening the PR helps.
However, temporary, it’s a good idea to use as much less CI as possible, because Travis has limited build-time for free use.
See #2300

Pierre-Sassoulas:
I saw this travis change, very unfortunate for foss... Maybe it's possible to merge without a new pipeline as the pipeline already passed for `4474c84` and the subsequent change can't break the tests ?

illume (approved):
👍 thanks

illume:
Hrmmm... It's crashing at the same place as it was in here: https://github.com/pygame/pygame/pull/2363
```python
test_set_italic (pygame.tests.ftfont_test.FtFontTypeTest) ... ok
test_set_underline (pygame.tests.ftfont_test.FtFontTypeTest) ... ok
test_size (pygame.tests.ftfont_test.FtFontTypeTest) ... ok

Ran 33 tests in 0.023s
OK
WARNING: You are using the SDL disk writer audio driver!
 Writing to file [sdlaudio.raw].

*** Error in `python': corrupted double-linked list: 0x0000ffff84001380 ***
```

Issue #2347, opened 2020-11-21
pygame.time.set_timer() accept numevents argument


ankith26:
Currently, pygame.time.set_timer() has only two options, it can either keep sending events forever, or send once and stop.
IMHO, it would be more useful if we had a way to send a particular amount of events, that can be easily specified by the user.
I know a way to make this work, and I can implement it, if you all like the idea too.
And don’t worry, I won’t break back-compat ;)

PS: Event refactor PR is almost done, took me half an hour to write the description for the PR, so you can expect how long it must be

illume:
What are some use cases for this?

- play four drum beats.
- fire 2 lasers

Have people requested this feature anywhere?

What would the function signatures be?
```python
set_timer(event_id: int, milliseconds: int)
set_timer(event_id: int, milliseconds: int, once=False: bool)
set_timer(event_id: int, milliseconds: int, num_times=None: [int, None])
```


How does it behave if you call it multiple times with `num_times` (maybe not the best name for the argument)?


ps. There's been people requesting being able to specify events as well, rather than just the id, which should maybe be considered at the same time. https://github.com/pygame/pygame/issues/1550

ankith26:
Yeah, accepting an event object is a very useful feature indeed, as you get to post all the event attributes too, and I have already worked on that feature. 

Coming to this issue, there can be many use cases, you might have a game where you want to press a particular key, and pressing another key might be a shortcut to pressing this key n number of times (at regular intervals), or you might have a networking-related app, where networking stuff runs on a separate thread, and you have a sort of a messaging system back to the main thread and you make use of events, which is a good way of moving data around in a thread-safe manner. Here you might have some sort of a requirement to send some data or send some events back to the main thread, and you might have to send a particular amount of events.
Currently, there is no control on how many events are sent. You either send only one, or keep sending forever (although I agree, you can keep a counter variable and suspend the timer after you don’t need it)

On the other hand, this feature might end up making the API a bit more complex, as supporting both the once argument and this numevents argument might be a bit confusing.

So there are both, ups and downs to this idea

illume:
I think keeping the third positional argument as a number (where True == 1) is fine. Renaming the argument to `num_times` (or whatever we call it), and just documenting that. Keeping the `once` param for compatibility, but not documenting it is fine.

[...]

Issue #2325, opened 2020-11-13
Inverse filtering in pygame.event.get()


robertpfeiffer:
It would be useful to filter for some events *except* certain specified ones in pygame.event.get.

The API could either be something like `pygame.event.get(pygame.FINGERDOWN, invert=True)` or `pygame.event.get(reject=(pygame.FINGERDOWN,))`. Both accepting _and_ rejecting in the same function call would make little to no sense.

ankith26:
I quite like the idea, I’m on it

Issue #2319, opened 2020-11-12
event.get(type or sequence) does not work as expected


ankith26:
`pygame.event.get(type or sequence)` is supposed to return a list of all such events from the event queue for the user to iterate over, using a `for` loop.

Ref docs: https://www.pygame.org/docs/ref/event.html#pygame.event.get

Currently broken on pygame 2, with SDL2. Here is some sample code
```python
import pygame
pygame.init()
pygame.display.set_mode((500, 500))

while True:
    pygame.time.delay(1000)
    print(pygame.event.get(pygame.MOUSEMOTION))
    if pygame.event.peek(pygame.QUIT):
        break
pygame.quit()
```

Sample output with pygame 2 (moving the mouse around the screen)
```
[<Event(1024-MouseMotion {'pos': (249, 141), 'rel': (3, 3), 'buttons': (0, 0, 0), 'window': None})>]
[<Event(1024-MouseMotion {'pos': (134, 362), 'rel': (2, 3), 'buttons': (0, 0, 0), 'window': None})>]
[<Event(1024-MouseMotion {'pos': (76, 89), 'rel': (1, 2), 'buttons': (0, 0, 0), 'window': None})>]
[<Event(1024-MouseMotion {'pos': (302, 83), 'rel': (4, 3), 'buttons': (0, 0, 0), 'window': None})>]
[<Event(1024-MouseMotion {'pos': (287, 396), 'rel': (2, 1), 'buttons': (0, 0, 0), 'window': None})>]
```

As you can clearly see, no matter how many MOUSEMOTION EVENTS happened within the call to pygame.event.get(), it returns a list with only one event.

Output with pygame 1.9.x (output gets really huge, so posting just a small part here)
```
[...]

Issue #2214, opened 2020-10-20
reconsider WINDOWEVENT API


illume:
I think we need to look again at WINDOWEVENT. I don't think it recieved enough attention when we added it.

Doing a repr(event) on it, shows stuff like <WindowEvent 'event': 3>, which is meaningless.

Also it's quite verbose to use:
event.type == pygame.WINDOWEVENT and event.event == pygame.WINDOWEVENT_HIDDEN:
VS using attributes:
if event.type == pygame.WINDOWEVENT and event.hidden:
VS using a separate event type for each one:
if event.type == pygame.WINDOWEVENT_HIDDEN:

Having to handle the WINDOWEVENT differently than all the other types is an extra load on the user.

ankith26:
So I Think it would be a better idea to directly check stuff, like
event.type == pygame.WINDOWEVENT_MAXIMISED. This means that each event under the SDL WINDOWEVENT must get a unique event in pygame

ankith26:
But an important thing to consider here is that will the values WINDOWEVENT_HIDDEN, WINDOWEVENT_MINIMISED, WINDOWEVENT_*(whatever) clash with aldready existing constants like pygame.QUIT, etc etc

ankith26:
If we use the constants with the same values of that of SDL, we might get some variable clashes

MyreMylar:
The current `WINDOWEVENT` structure is directly from SDL, it wasn't created by us. 

To change it we would have to filter the events to add pygame versions, but we would likely still have to leave the SDL events on the queue as well in case that creates chaos. This is much as we currently have for events like `pygame.VIDEOEXPOSE`.

**SDL docs:**
https://wiki.libsdl.org/SDL_EventType
https://wiki.libsdl.org/SDL_WindowEvent

robertpfeiffer:
> To change it we would have to filter more of the events to add pygame versions, but we would likely still have to leave the SDL events on the queue as well in case that creates chaos. This is what we currently have for events like `SDL_WINDOWEVENT/SDL_WINDOWEVENT_RESIZED` & `pygame.VIDEORESIZE` and `SDL_WINDOWEVENT/SDL_WINDOWEVENT_EXPOSED` & `pygame.VIDEOEXPOSE`.

Yes. I kept these events for this reason. SDL2 internals need them in the queue.

ankith26:
[...]

PR #2357, opened 2020-11-24
Major event refactor

### LINKED ISSUE #1236: SDL2: KEYUP events not returning unicode value
For some reason KEYDOWN events only return a unicode value some of the time. I know TEXTINPUT is meant to replace it, and that does work well, but this will cause many existing programs to fail. It took me a long time to notice it myself, as it only effects the text input dialog of my menu class, which is rarely use. It wasn't until I started working on new flyboy levels and needed to name an event that I noticed the problem.

It seems to me that KEYDOWN.unicode reports a character about 5% of the time a key is pressed. This seems wierd: one would expect an all or nothing result.

My text input dialog is complicated due to the many conditions handled, such as cursor position, deletion, key repeat, etc. I worked around this problem by ignoring the certatin keydown events if SDL2 is defined, and handling TEXTINPUT events if SDL2 is defined.

	def update_input(self, events, max_length):
		if events == False: return
		for event in events:
				self.cursor_visible = True # So the user sees where he writes

				# If none exist, create counter for that key:
				if not event.key in self.keyrepeat_counters:
					self.keyrepeat_counters[event.key] = [0, event.unicode]

				if event.key == K_BACKSPACE: # FIXME: BACKSPACES NOT REPEATING IN SDL2
					self.input_string = self.input_string[:max(self.cursor_position - 1, 0)] + \
										self.input_string[self.cursor_position:]

					# Subtract one from cursor_pos, but do not go below zero:
					self.cursor_position = max(self.cursor_position - 1, 0)
				elif event.key == K_DELETE:
					self.input_string = self.input_string[:self.cursor_position] + \
										self.input_string[self.cursor_position + 1:]

				elif event.key == K_RETURN or event.key == K_KP_ENTER:
					self.keyrepeat_counters = {}
					return True

				elif event.key == K_RIGHT:
					# Add one to cursor_pos, but do not exceed len(input_string)
					self.cursor_position = min(self.cursor_position + 1, len(self.input_string))

				elif event.key == K_LEFT:
					# Subtract one from cursor_pos, but do not go below zero:
					self.cursor_position = max(self.cursor_position - 1, 0)

				elif event.key == K_END:
					self.cursor_position = len(self.input_string)

				elif event.key == K_HOME:
					self.cursor_position = 0

				# REGULAR KEY PRESSED, HANDLE IT AS KEYDOWN FOR SDL1
				# IGNORE IT IF SDL2 TO AVOID OCCASIONAL REDUNDANT INPUTS
				elif len(self.input_string) <= max_length and not SDL2:
					self.input_string = self.input_string[:self.cursor_position] + \
										event.unicode + \
										self.input_string[self.cursor_position:]
					self.cursor_position += len(event.unicode) # Some are empty, e.g. K_UP

			# GET CHARACTER FROM TEXTINPUT EVENT IF SDL2
			elif SDL2 and event.type == TEXTINPUT:
				if len(self.input_string) <= max_length:
					# If no special key is pressed, add unicode of key to input_string
					self.input_string = self.input_string[:self.cursor_position] + \
										event.text + \
										self.input_string[self.cursor_position:]
					self.cursor_position += len(event.text) # Some are empty, e.g. K_UP

			elif event.type == KEYUP:
				# *** Because KEYUP doesn't include event.unicode, this dict is stored in such a weird way
				if event.key in self.keyrepeat_counters:
					del self.keyrepeat_counters[event.key]

		# Update key counters:
		for key in self.keyrepeat_counters :
			self.keyrepeat_counters[key][0] += self.clock.get_time() # Update clock
			# Generate new key events if enough time has passed:
			if self.keyrepeat_counters[key][0] >= self.keyrepeat_intial_interval_ms:
				self.keyrepeat_counters[key][0] = self.keyrepeat_intial_interval_ms - \
													self.keyrepeat_interval_ms

				event_key, event_unicode = key, self.keyrepeat_counters[key][1]
				pygame.event.post(pygame.event.Event(KEYDOWN, key=event_key, unicode=event_unicode))

		# Rerender text surface:
		self.input_surface = self.body_font.render(self.input_string, 1, self.select_color)

		# Update self.cursor_visible
		self.cursor_ms_counter += self.clock.get_time()
		if self.cursor_ms_counter >= self.cursor_switch_ms:
			self.cursor_ms_counter %= self.cursor_switch_ms
			self.cursor_visible = not self.cursor_visible

		if self.cursor_visible:
			cursor_x_pos = self.body_font.size(self.input_string[:self.cursor_position])[0]
			# Without this, the cursor is invisible when self.cursor_position > 0:
			if self.cursor_position > 0:
				cursor_x_pos -= self.cursor_surface.get_width()
			self.input_surface.blit(self.cursor_surface, (cursor_x_pos, 0))
			self.cursor_off = cursor_x_pos + self.body_font.size('O')[0]

		#self.clock.tick()
		return False

**Related Docs:** https://www.pygame.org/docs/ref/event.html

ankith26:
TLDR;

This PR makes a LOT of fixes and additions to pygame. Here are the issues that this PR can (potentially) close

#1236
#1238
#1524
#1550
#2146
#2214
#2319
#2347
#2378 

This PR also makes fixes outside of these issues.

NOTE: This message is going to be a bit long, I have tried to be less verbose, but its difficult to explain all I have done in a few paragraphs, I have condensed my message to include only the key highlights of this PR.

Starting of with `time.set_timer()`, this function was not updated well to match with SDL2. It did not support the new events of SDL2. With SDL2, it has a very inefficient way of storing data - in an array of length over 8200. And most (almost all) of that memory was never to be used, hence leaked. I put a much better system in place to store all that internal data, so no more memory leaks. Also, `time.set_timer()` now supports the new SDL2 events, and it also accepts its first parameter as an `event.Event` object, which is good because events on timer can have any attributes too, that users will want to put. It also accepts a new `repeat` argument, thats going to indicate how many times the event appears on the queue (ref: #2347).

Coming to `event.post()` issues. This one was quite problematic for many people, it would segfault for some, it would not post some events in some cases, and posted `KEYDOWN/KEYUP` events could not contain the `unicode` feild. All that is now fixed, thanks due to a new system of "proxy events". 
proxy events provide a way to safely post events from pygame side that look like USEREVENTS from the SDL side. Another internal code change I did was to remove code that did pointer checking for posted userevents. While this system was quite good, it kinda seemed redundant to me. Note that the new system of "proxy events" ensures, by design, that all events above a certain event are guarenteed to have been posted by pygame, hence all events are in the format pygame wishes to see. That makes these checks even more redundant.

With SDL2, `KEYDOWN` events did not reliably have the `unicode` attribute. This PR fixes that by putting a system in place to fill the `unicode` attribute, by determining the `unicode` attribute from the `key` attribute. Note that this system is not perfect, and is intended to be used as a last resort when the normal way of determining the `unicode` attribute fails. 
This fix also ensures that the less commonly used key-combinations that do not generate `unicode` in pygame 2.0.0 due to `TEXTINPUT` not supporting them, now work like how they used to. tab, backspace, enter, escape, delete, keypad keys and control characters - all of these now have `unicode` like they used to in pygame 1.9.x.

Also this PR makes the unicode attribute appear on `KEYUP` events as well (This feature is not available with SDL1). There were some disparities on python 2 with unicode attribute on SDL1 and SDL2. Fixed that, and also made the code handling that much more efficient.

Another change this PR makes is to make it such that each subevent in `WINDOWEVENT` gets its own, unique event in pygame. This PR adds new `WINDOW*` constants to pygame. Also, added `x` and `y` attributes for some of the windowevents.

This PR fixed key-repeat getting messed up when KEYDOWN events were posted from `pygame.event.post()`.

There were quite a few issues with event blocking in pygame 2. First, event blocking would not work on `ACTIVEEVENT`, `VIDEORESIZE` and `VIDEOEXPOSE`. Similarly, blocking `WINDOWEVENT` would block those events too. Blocking `MOUSEWHEEL` events would also block `MOUSEBUTTONUP/DOWN` mousewheel events. Blocking `TEXTINPUT` would mean that you will not get the `unicode` attribute on `KEYDOWN` events. And blocking `KEYUP` events would give an infinite loop of `KEYDOWN` events, when key repeat was set on pygame 2 (does not happen on pygame 1). Events posted by `pygame.time.set_timer()` and mixer/music modules callback function never obeyed event blocking.
This PR fixes all these issues.

Also removed experimental function `_set_gen_videoresize()`, because it is the same as doing `set_blocked(pygame.VIDEORESIZE)`

`pygame.event.post()` now returns a boolean on whether the event was posted or not (it still raises an error if things go wrong)

Fixes issues with `pygame.event.get()`, where `pygame.event.get(type or sequence)` was broken with SDL2 (see #2319 for more info)

I have also made a LOT of code cleanups, refactor and updates, and have reduced the "complexity" of a few functions in event.c.

I was planning on working on #2325, but I havent, as there hasn't been any discussion on that issue. Probably after this PR, we can decide on what to do about that (@robertpfeiffer if you want to take that one up, feel free to do so)

Overall, this PR has done quite a few things, I have been working on this over the past month or so. And its not done yet. There's a lot of stuff to test, and probably make in a few more changes. Finally, when this is set to go, I will also update unittests, docs and add type-hints to some newly added constants :)


ankith26:
LGTM is happy, Appveyor is happy, python2 is happy, SDL1 is hopefully happy. Its just travis.

ankith26:
Finally... SDL1 is also happy now. The issue was with SDL_PostEvent having different return codes on error. 

Side note: While doing local SDL1 testing, found that scrap module unittests segfaults occasionally for me.

src_c/mixer.c
```diff
@@ -261,18 +261,45 @@ _format_view_to_audio(Py_buffer *view)
...
 
+static void
+_pg_push_mixer_event(int type, int code)
+{
+    pgEventObject *e;
+    PyObject *dict, *dictcode;
+    SDL_Event event;
+    PyGILState_STATE gstate = PyGILState_Ensure();
+
+    dict = PyDict_New();
```
illume, line 272: There's no error checking.

src_c/mixer.c
```diff
@@ -261,18 +261,45 @@ _format_view_to_audio(Py_buffer *view)
...
+#if IS_SDLv1
+            if (SDL_PushEvent(&event) < 0)
+#else
+            if (SDL_PushEvent(&event) <= 0)
+#endif
+                Py_DECREF(dict);
+            Py_DECREF(e);
+        }
+    }
+    PyGILState_Release(gstate);
```
illume, line 293: Since all branches of endsound_callback use the gil now, maybe this is something that can happen once in endsound_callback?

src_c/music.c
```diff
@@ -51,14 +51,32 @@ mixmusic_callback(void *udata, Uint8 *stream, int len)
...
+    SDL_Event event;
+    PyGILState_STATE gstate = PyGILState_Ensure();
+
+    e = (pgEventObject *)pgEvent_New2(type, NULL);
+    if (e) {
+        pgEvent_FillUserEvent(e, &event);
+#if IS_SDLv1
+        if (SDL_PushEvent(&event) < 0)
+#else
+        if (SDL_PushEvent(&event) <= 0)
```
illume, line 66: https://wiki.libsdl.org/SDL_PushEvent
> Returns 1 on success, 0 if the event was filtered, or a negative error code on failure; call SDL_GetError() for more information. A common reason for error is the event queue being full. 

I'm wondering about the filtered case.

src_c/mixer.c
ankith26: Yeah I did think of that. But then there was the possibility that the function just goes from start to end directly, in which case the GIL will be an extra overhead

src_c/time.c
```diff
@@ -191,68 +255,41 @@ time_wait(PyObject *self, PyObject *arg)
...
-    if (!PyArg_ParseTuple(arg, "ii|i", &event, &ticks, &once))
+    int ticks, once = 0, repeat = 0;
+    PyObject *obj;
+    pgEventObject *e;
+
+    static char *kwids[] = {
+        "event",
+        "millis",
+        "once",
+        "repeat",
```
illume, line 269: I wonder if repeat is the best name?

src_c/music.c
ankith26: Well, if I get what filtering means, it means that the event passed through the event filter and got rejected right.
In that case we do need to decref the dict

src_c/time.c
ankith26: repeat is fine IMHO.
Another alternative would be numevents

src_c/time.c
```diff
@@ -513,18 +553,27 @@ MODINIT_DEFINE(time)
...
     /* type preparation */
     if (PyType_Ready(&PyClock_Type) < 0) {
         MODINIT_ERROR;
     }
 
+    /* register cleanup function for event timer holding structure,
+     * allocate a mutex for this structure too */
+    if (!timermutex && !pg_event_timer) {
+        timermutex = SDL_CreateMutex();
+        if (!timermutex) {
```
illume, line 571: Is this correct? If the timermutex fails, we do anyway `pg_RegisterQuit(_pg_event_timer_cleanup);`?

ankith26: If timermutex fails, then modinit_error will exit the function, right?

src_c/mixer.c
ankith26: It is checking errors, but failing silently. We can’t do anything about errors from callback functions

illume:
Good work. I haven't read through it all yet, but asked a few questions and made some notes.

So what's next?

- We need to make a list of things to test Some would be with unit tests, and others manual(or interactive). That list of issues is a good start, but we need specific things. Which pygame.example apps can we use to test what things, and also which other apps (like solarwolf) can we use to test what features?
- Do you want to continue in one big PR, or try and break it down into small parts that can be merged incrementally? Adding tests first in other PRs could be one way to break this up. I don't have any other ideas on how to break up the changes yet.
- Documentation needs updating (which could also be done in a separate PR).

ankith26:
My initial plans were to do unittests and docs later in this PR. Although those can change if needed :) 
I guess it would be better if I come up with a few blocks of code, that kinda summarise and showcase the changes I made, while also serving as something to test my changes with

src_c/time.c
```diff
@@ -28,76 +28,140 @@
...
+
+    SDL_LockMutex(timermutex);
+
+    if (pg_event_timer) {
+        hunt = pg_event_timer;
+        do {
+            if (hunt->event->type == type) {
+                if (hunt->repeat >= 0)
+                    hunt->repeat--;
+                SDL_UnlockMutex(timermutex);
```
illume, line 100: Better to do cleanup and return in one place at the end.

src_c/time.c
```diff
@@ -28,76 +28,140 @@
...
+_pg_get_event_on_timer(int type)
 {
-    if (SDL_WasInit(SDL_INIT_VIDEO)) {
-        SDL_Event event;
-        memset(&event, 0, sizeof(event));
-        event.type = (intptr_t)param;
-        SDL_PushEvent(&event);
+    pgEventTimer *hunt;
+
+    SDL_LockMutex(timermutex);
```
illume, line 92: SDL_LockMutex can fail. Mostly(only?) if timermutex is NULL. Can that happen? 

Perhaps if the module quits whilst the timer is in mid flight?

src_c/time.c
illume: Indeed, it will.

Probably need to set the python error here with the SDL error?
```c
PyErr_SetString(PyExc_ImportError, SDL_GetError());
```

src_c/time.c
```diff
@@ -28,76 +28,140 @@
...
 }
 
-static Uint32
-timer_callback_once(Uint32 interval, void *param)
+static void
+_pg_event_timer_cleanup(void)
 {
-    return timer_callback(0, param);
+    pgEventTimer *hunt, *todel;
+    SDL_LockMutex(timermutex);
```
illume, line 114: It might not hurt to make _pg_event_timer_cleanup defensive to timermutex being NULL (when SDL_LockMutex fails).

src_c/mixer.c
```diff
@@ -261,18 +261,45 @@ _format_view_to_audio(Py_buffer *view)
...
-        if (channeldata[channel].endevent && SDL_WasInit(SDL_INIT_VIDEO)) {
-            SDL_Event e;
-            memset(&e, 0, sizeof(e));
-            e.type = channeldata[channel].endevent;
-            if (e.type >= PGE_USEREVENT && e.type < PG_NUMEVENTS)
-                e.user.code = channel;
-            SDL_PushEvent(&e);
-        }
+        if (channeldata[channel].endevent && SDL_WasInit(SDL_INIT_VIDEO))
+            _pg_push_mixer_event(channeldata[channel].endevent, channel);
```
illume, line 301: We need to test how adding a GIL lock affects end event timing.

illume:
>  I guess it would be better if I come up with a few blocks of code, that kinda summarise and showcase the changes I made, while also serving as something to test my changes with

Indeed. Unit tests, docs (including code examples) are a great way to communicate to the reviewer what a patch is meant to do.

The python 3.8 build on appveyor runs with a static analyzer, and gets these warnings
https://ci.appveyor.com/project/pygame/pygame/builds/36498456/job/w9mjgpdiu63idpfc?fullLog=true#L2002

```C
src_c/time.c(138): warning C4244: 'function': conversion from 'intptr_t' to 'int', possible loss of data
src_c/time.c(312): warning C4312: 'type cast': conversion from 'int' to 'void *' of greater size
```

src_c/time.c
ankith26: If timermutex was null, we would have failed on module import

src_c/time.c
ankith26: Ah I missed that. Will add

src_c/time.c
ankith26: Will do :)

src_c/music.c
ankith26: Verified, and that’s what filtering means. If the event got filtered then it never entered the event queue.
So we have to check for 0 and decref our dict

src_c/time.c
ankith26: My approach here was to just try to lock the mutex. And if it fails, we just leave it.
Now im thinking maybe it’s not a great idea to do it this way. I will try to add some error handling to this

ankith26:
Hmm I see a few things in the code that needs improvement.
I will probably put in another commit tomorrow that improves a few stuff and also address some issues noted by illume

src_c/time.c
ankith26: Added :)

src_c/time.c
ankith26: done :)

src_c/time.c
ankith26: I just added better error checking in my latest commit

ankith26:
Latest commit does a few fixes and improvements.
Also fixed the appveyor static analyzer warnings, and updated the unit tests for `time.set_timer()` to test the new features of that function

src_c/mixer.c
illume: There's two ways to handle tracebacks in callbacks. 
- 1) log the traceback in the handler
- 2) note the traceback and then raise the traceback somewhere else, perhaps in event get/pump/wait.

Two is more the python way.

src_c/time.c
illume: We need a unit test for this repeat arg.

illume: Let's go with repeat, unless someone objects... what do you think @MyreMylar @Starbuck5 ?

illume:
What's left to do?

- There's a few review items left. Before testing starts in earnest, it might make sense to do these ones first.
- It would be nice if someone else could review this too. I know that's a massive task though.
- The main work left is to write tests/reproducers for various issues. There's already some manual tests/reproducers in some of the issues. It will be easier if these are in unit tests/interactive tests. These tests can be added as separate PRs.

I can help out with writing tests on the weekend (and maybe 1-2 during the week).

src_c/time.c
ankith26: I already added a repeat unit test :)

ankith26:
Hmm yeah. It’s a bit hard to add the unittests for some of new features/fixes. 
This PR makes only 3 additions. pygame.time.set_timer changes which I already wrote unittests for, and pygame.event.post returning a Boolean, which I can add unittests easily. And I removed _set_gen_videoresize(), that does not need any change in the tests.

Rest of all the other changes are fixes, I don’t know how to write unittests for some of those

Thanks for the help, though :) 
There’s one thing I wanted to experiment with. That’s about pygame.event.iget(). But you all can go ahead with reviews, as that won’t change a lot of code

src_c/mixer.c
ankith26: Even the code that was there previously did not do any error handling. It’s much more work to handle those errors. Much easier to fail silently (maybe I’m just being lazy here ;) )

src_c/time.c
ankith26: @robertpfeiffer what do think of the name for the argument

src_c/mixer.c
robertpfeiffer: Asyncio does the first, and since then (in hindsight) it has become the consensus that this was a mistake, and async event handling should probably do the second.

src_c/time.c
MyreMylar: I don't mind repeat. I think I would assume from the name that it was a boolean flag to make the event timer continuously repeat infinitely.

ankith26: Hmm valid point. Do you have any other names in mind

Starbuck5: I agree with MyreMalar - repeat sounds like a flag. I'm partial to illume's proposed `num_times` in the issue. I like that or just plain `times`.

zoldalma999: We could call it `loops` since `mixer` and `mixer.music` uses it in `.play()`.

ankith26: We can discuss this further in #2347

illume:
The items left from here:

- We need to test how adding a GIL lock affects end event timing. If we can still do gap-less playback (as well as before) with end events, that would be fine. Or if there's a way to remove the GIL anyway... even better.
- (semi optional) callback error handling. At minimum the error cases should be handled, and cleaned up. Best would be to raise an error somewhere. Is there an error happening already and we don't know about it? We could just create an issue for this work and risk it.
- (optional) It might not hurt to make `_pg_event_timer_cleanup` defensive to `timermutex` being `NULL` (when `SDL_LockMutex` fails).

ankith26:
1. The GIL is unavoidable. Because we need to support users passing any
events attributes (and thus, python dict)

2. I have a solution for that. And that is creating a new event called
"errorevent". Call back functions will post this event if  things go wrong,
and the receiving function (get, poll, etc) will raise an error. This wont
work for forwarding all errors though, although this can forward most
errors
3. I didn't get what you meant exactly. Does that mean that I have to check
that the mutex is not NULL before going into the function logic??

illume:
> 1. The GIL is unavoidable. Because we need to support users passing any events attributes (and thus, python dict)

Could it be avoided for not-user-event types ?

> 2. creating a new event called "errorevent".

Love it!

ankith26:
> Could it be avoided for not-user-event types ?
In theory, yes. But practically, it is going to involve a **lot** more code reworking, and about a months time (considering I code 1-2 hours per day). And I do not have that much time.
However, I will try to get some benchmarking done on how much of an effect the GIL has, and post the results here. I am 90% sure that the GIL does not have much overhead in this case, but I will benchmark anyways.

The problem with errorevent idea is that if an error happens while posting the error event, then we can’t do anything about that one. 
But I will try to get that one in place, and see how things go

ankith26:
I did some benchmarking, and it turns out, GIL does not have much effect on the efficiency of the callback functions. I did benchmarking on ``pygame.time.set_timer`` because the other two callbacks are one-offs.

Here is the interesting part of the code that does benchmarking
```python
TESTCASES = 100
TIME = 10
x = []

pygame.time.set_timer(pygame.USEREVENT, TIME)
for i in range(TESTCASES):
    start = time.perf_counter()
    pygame.event.wait()
    x.append(((time.perf_counter() - start) * 1000 - TIME) * 1000)
```
My branch averages at `123.79539999528788`
pygame 2.0.0 averages at `143.43869010190247`
Figure_1

illume (approved):
👍 super good. 🎉 🎈

PR #2361, opened 2020-11-27
added unit test for surface.convert()

bwright8:
added unit test for convert


ankith26:
Noticed you added the obsolete test/util/* code back again which got removed recently.
I assume it’s by accident, correct me if I’m wrong

illume:
Yeah, something weird must have happened with those extra files hey?
Might be easiest to make a copy of your change, and then do a fresh git clone.

bwright8:
> Yeah, something weird must have happened with those extra files hey?
> Might be easiest to make a copy of your change, and then do a fresh git clone.

Thanks, I thought I forked just earlier today so that this wouldn’t be an issue. I’ll try again like you say.

PR #2362, opened 2020-11-28
Link to Patreon

Starbuck5:
I noticed that @illume set up a pygame patreon recently. This commit would put a little "sponsor this project" tab on the side of the github repo, which would just serve to increase the visibility of the patreon.

[skip ci]

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -1,0 +1,1 @@
+patreon: pygame


ankith26:
Sounds like a good idea, it’s something I was about to suggest as well.

MyreMylar:
I'm happy for this to go up. 👍

MyreMylar (approved):

PR #2363, opened 2020-11-28
ENH: Updating TYPE_HEAD Macro to use newer API

### LINKED ISSUE #2208: Update C TYPE_HEAD macro in all C files
**Description**

> One thing that came out of this though is changing the TYPE_HEAD macro would make sense. PyVarObject_HEAD_INIT is available in Python 2.7 and 3.x.
> 
> `#if defined(MS_WINDOWS)
> #define PYVAROBJECT_HEAD_INIT(a,b) PyVarObject_HEAD_INIT(0,b)
> #else
> #define PYVAROBJECT_HEAD_INIT(a,b) PyVarObject_HEAD_INIT(a,b)
> #endif
> 
> `
> Then the declarations could be PYVAROBJECT_HEAD_INIT(&PyType_Type,0) which is really what the Python maintainer wants (the &PyType_Type).

This sounds like a good idea to me.

Iamsoto:
Hello,

I'm new to the PyGame community and wanted to help out. 
I followed issue #2208 to update the macro ` TYPE_HEAD` to `PYVAROBJECT_HEAD_INIT` , where the later calls the same named C-API. 

As @MyreMylar suggested, I will let the CI/CD run to see if the conditional on the macro is really needed. If it fails, I will add the conditional in another commit and try again. If it passes then I will go ahead remove the custom Macro and just use the C-API (unless someone tells me this is a bad idea).

I hope this PR ends up helping someone!

fixes #2208 

**UPDATE:**
Appveyor jobs are failing, they appear to all be in windows directories here's an [example output](https://ci.appveyor.com/project/pygame/pygame/builds/36552285/job/5mhe8huasovmm6c8):

So I will make another commit where I add the conditional. And summon the C and Python dieties that be to help.

```
building 'pygame.font' extension
C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\cl.exe /c /nologo /Ox /MD /W3 /GS- /DNDEBUG -Iprebuilt-x86/SDL2-2.0.12/include -Iprebuilt-x86/SDL2_ttf-2.0.15/include -IC:\projects\pygame\pypy2.7-v7.3.2-win32\include /Tcsrc_c/font.c /Fobuild\temp.win32-2.7\Release\src_c/font.obj /GF /Gy
font.c
src_c/font.c(942): error C2099: initializer is not a constant
src_c/font.c(942): warning C4047: 'initializing': 'Py_ssize_t' differs in levels of indirection from 'char [17]'
src_c/font.c(942): warning C4047: 'initializing': 'const char *' differs in levels of indirection from 'unsigned int'
src_c/font.c(942): warning C4047: 'initializing': 'Py_ssize_t' differs in levels of indirection from 'destructor'
src_c/font.c(942): warning C4047: 'initializing': 'PyMappingMethods *' differs in levels of indirection from 'hashfunc'
src_c/font.c(942): warning C4113: 'ternaryfunc' differs in parameter lists from 'hashfunc'
src_c/font.c(942): warning C4047: 'initializing': 'hashfunc' differs in levels of indirection from 'ternaryfunc'
src_c/font.c(942): warning C4113: 'reprfunc' differs in parameter lists from 'ternaryfunc'
src_c/font.c(942): warning C4047: 'initializing': 'PyBufferProcs *' differs in levels of indirection from 'long'
src_c/font.c(942): warning C4047: 'initializing': 'long' differs in levels of indirection from 'char [93]'
src_c/font.c(942): warning C4047: 'initializing': 'richcmpfunc' differs in levels of indirection from 'std::size_t'
src_c/font.c(942): warning C4047: 'initializing': 'iternextfunc' differs in levels of indirection from 'PyMethodDef *'
src_c/font.c(942): warning C4133: 'initializing': incompatible types - from 'PyGetSetDef *' to 'PyMemberDef *'
src_c/font.c(942): warning C4047: 'initializing': 'Py_ssize_t' differs in levels of indirection from 'initproc'
---
```

**UPDATE 2:**
Travis is failing for a reason that appears to be unrelated? 
But Appveyo succeeds.

### DIFF
--- a/src_c/_camera.c
+++ b/src_c/_camera.c
@@ -1745,7 +1745,8 @@ PyObject* camera_getattr(PyObject* self, char* attrname) {
 }
 */
 PyTypeObject pgCamera_Type = {
-    TYPE_HEAD(NULL, 0) "Camera",
+    PyVarObject_HEAD_INIT(NULL,0)
+    "Camera",
     sizeof(pgCameraObject),
     0,
     camera_dealloc,
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -647,7 +647,8 @@ static PyGetSetDef _ftfont_getsets[] = {
 #define FULL_TYPE_NAME MODULE_NAME "." FONT_TYPE_NAME
 
 PyTypeObject pgFont_Type = {
-    TYPE_HEAD(0, 0) FULL_TYPE_NAME,           /* tp_name */
+    PyVarObject_HEAD_INIT(NULL,0)
+    FULL_TYPE_NAME,                           /* tp_name */
     sizeof(pgFontObject),                     /* tp_basicsize */
     0,                                        /* tp_itemsize */
     (destructor)_ftfont_dealloc,              /* tp_dealloc */
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -647,7 +647,7 @@ static PyGetSetDef _ftfont_getsets[] = {
 #define FULL_TYPE_NAME MODULE_NAME "." FONT_TYPE_NAME
 
 PyTypeObject pgFont_Type = {
-    PyVarObject_HEAD_INIT(NULL,0)
+    PyVarObject_HEAD_INIT(0,0)
     FULL_TYPE_NAME,                           /* tp_name */
     sizeof(pgFontObject),                     /* tp_basicsize */
     0,                                        /* tp_itemsize */
--- a/src_c/bufferproxy.c
+++ b/src_c/bufferproxy.c
@@ -654,7 +654,8 @@ static PyBufferProcs proxy_bufferprocs = {
 #endif
 
 static PyTypeObject pgBufproxy_Type = {
-    TYPE_HEAD(NULL, 0) PROXY_TYPE_FULLNAME, /* tp_name */
+    PyVarObject_HEAD_INIT(NULL,0)
+    PROXY_TYPE_FULLNAME,                    /* tp_name */
     sizeof(pgBufproxyObject),               /* tp_basicsize */
     0,                                      /* tp_itemsize */
     (destructor)proxy_dealloc,              /* tp_dealloc */
--- a/src_c/cdrom.c
+++ b/src_c/cdrom.c
@@ -525,7 +525,8 @@ static PyMethodDef cd_methods[] = {
     {NULL, NULL, 0, NULL}};
 
 static PyTypeObject pgCD_Type = {
-    TYPE_HEAD(NULL, 0) "CD", /* name */
+    PyVarObject_HEAD_INIT(NULL,0)
+    "CD",                    /* name */
     sizeof(pgCDObject),      /* basic size */
     0,                       /* itemsize */
     cd_dealloc,              /* dealloc */
--- a/src_c/color.c
+++ b/src_c/color.c
@@ -326,7 +326,8 @@ static PyBufferProcs _color_as_buffer = {
 #define DEFERRED_ADDRESS(ADDR) 0
 
 static PyTypeObject pgColor_Type = {
-    TYPE_HEAD(NULL, 0) "pygame.Color", /* tp_name */
+    PyVarObject_HEAD_INIT(NULL,0)
+    "pygame.Color",                    /* tp_name */
     sizeof(pgColorObject),             /* tp_basicsize */
     0,                                 /* tp_itemsize */
     (destructor)_color_dealloc,        /* tp_dealloc */
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -399,7 +399,8 @@ pg_vidinfo_str(PyObject *self)
 }
 
 static PyTypeObject pgVidInfo_Type = {
-    TYPE_HEAD(NULL, 0) "VidInfo", /*name*/
+    PyVarObject_HEAD_INIT(NULL,0)
+    "VidInfo",                    /*name*/
     sizeof(pgVidInfoObject),      /*basic size*/
     0,                            /*itemsize*/
     pg_vidinfo_dealloc,           /*dealloc*/
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -1170,7 +1170,8 @@ Unimplemented:
 }
 
 static PyTypeObject pgEvent_Type = {
-    TYPE_HEAD(NULL, 0) "Event", /*name*/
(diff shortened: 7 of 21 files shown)


illume:
Thanks for doing this :)

Here is the failing test on the arm64, ubuntue xenial, python 3.8 [build](https://travis-ci.org/github/pygame/pygame/jobs/746389200#L2910).

```
test_set_bold (pygame.tests.ftfont_test.FtFontTypeTest) ... ok
test_set_italic (pygame.tests.ftfont_test.FtFontTypeTest) ... ok
test_set_underline (pygame.tests.ftfont_test.FtFontTypeTest) ... ok
test_size (pygame.tests.ftfont_test.FtFontTypeTest) ... ok

Ran 32 tests in 0.023s

OK

WARNING: You are using the SDL disk writer audio driver!
 Writing to file [sdlaudio.raw].
*** Error in `python': double free or corruption (out): 0x0000ffff5c001e00 ***
/home/travis/.travis/functions: line 112:  9702 Aborted                 $PYTHON_EXE -m pygame.tests.__main__ -v --exclude opengl,timing --time_out 300
```

Iamsoto:
Odd... the previous commit passed Travis. I have clearly summoned the wrong c/python dieties

illume:
heh. 👿 

Could you please try reverting the change in `src_c/_freetype.c`? If it passes, then the issue is probably in that area.

Iamsoto:
I didn't completely remove the change. I just put a '0' instead of the reference `&PyType_Type` 
We'll see...

illume:
Ah, good spot. It's passing on arm64 now :)

I noticed a similar thing in a bunch of other places... like the Camera, and buffer objects. Should these be changed too? Or were they wrong before?

MyreMylar:
I wonder if this failure in the freetype tests (pre-existing this PR) is at all related to why the change didn't work there:

```
FAIL: test_garbage_collection (__main__.FreeTypeFontTest)
Check reference counting on returned new references
  File "C:\Users\dan\Programming\pygame\test\freetype_test.py", line 1654, in test_garbage_collection
    self.assertEqual(getrefcount(item), 2, "refcount fail for item %d" % i)
AssertionError: 4 != 2 : refcount fail for item 0

```

MyreMylar:
Other than the slight weirdness around freetype everything else looks good on windows locally. Ran all the unit tests and tried it out in a few other programs (solarwolf, gui examples, FlyBoy) to no noticeable ill effects.

Maybe @cbxbiker61 who originally suggested this change might want to peek at this PR, if only to see it being brought to fruition :)

MyreMylar (approved):

Iamsoto:
Success!! I think!?
It seems weird to me, the first commit, [it passed travis](https://travis-ci.org/github/pygame/pygame/builds/746386761). And the macro was defined as:
[`#define PYVAROBJECT_HEAD_INIT(a,b) PyVarObject_HEAD_INIT(a,b)` ](https://github.com/pygame/pygame/commit/3e6e624b2ca9472cbf06d4263420c0644ab3fb70#diff-09699f0c3a8440efa47b5e3079577f618ac347378001b05fbeb5c282146b5f4b)meaning it _should_ have been using the reference then...

So I'm wondering if perhaps this is some weird edge case the cpython developers don't check much since they recommend using _[PyVarObject_HEAD_INIT(NULL, 0)](https://docs.python.org/2.7/extending/newtypes.html)_ 

Perhaps it would be better to just ditch the macro, and use the c python api without a reference?

Iamsoto:
But if you guys are cool with this I'm cool with this too!!!

MyreMylar:
I poked around some more and dug up this document:

https://docs.python.org/3.5/extending/newtypes.html


```
Now if you go and look up the definition of PyTypeObject in object.h you’ll see that it has many more fields that the definition above. The remaining fields will be filled with zeros by the C compiler, and it’s common practice to not specify them explicitly unless you need them.

This is so important that we’re going to pick the top of it apart still further:

PyVarObject_HEAD_INIT(NULL, 0)
This line is a bit of a wart; what we’d like to write is:

PyVarObject_HEAD_INIT(&PyType_Type, 0)
as the type of a type object is “type”, but this isn’t strictly conforming C and some compilers complain. Fortunately, this member will be filled in for us by PyType_Ready().
```

And then later:

```
For now, all we want to be able to do is to create new Noddy objects. To enable object creation, we have to provide a tp_new implementation. In this case, we can just use the default implementation provided by the API function PyType_GenericNew(). We’d like to just assign this to the tp_new slot, but we can’t, for portability sake, On some platforms or compilers, we can’t statically initialize a structure member with a function defined in another C module, so, instead, we’ll assign the tp_new slot in the module initialization function just before calling PyType_Ready():

noddy_NoddyType.tp_new = PyType_GenericNew;
if (PyType_Ready(&noddy_NoddyType) < 0)
    return;
All the other type methods are NULL, so we’ll go over them later — that’s for a later section!

Everything else in the file should be familiar, except for some code in PyInit_noddy():

if (PyType_Ready(&noddy_NoddyType) < 0)
    return;
This initializes the Noddy type, filing in a number of members, including ob_type that we initially set to NULL.
```

From other comments (e.g. on stack overflow) it is assumed that the compiler that has the problems with `PyVarObject_HEAD_INIT(&PyType_Type, 0)` was the MSVC one, but perhaps it is occasionally other compilers as well?

MyreMylar:
I think probably changing the first part of the macro to always be NULL is probably safest by the looks of things. Would also simplify the macro further because it could just be the same on all platforms.

Iamsoto:
Is it worth having the pygame defined macro then? the `PYVAROBJECT_HEAD_INIT(a,b)` Or should we just use the built in 
`PyVarObject_HEAD_INIT`  from python.h and pass in NULL (since it will be the same in any system)

cbxbiker61:
> Is it worth having the pygame defined macro then? the `PYVAROBJECT_HEAD_INIT(a,b)` Or should we just use the built in
> `PyVarObject_HEAD_INIT` from python.h and pass in NULL (since it will be the same in any system)

If you are set on passing in NULL... you may as well ditch the macro.

Iamsoto:
Okay, I did it.

Iamsoto:
It seems to pass now. I'm afriad to touch the code again...

illume:
The problem happened here as well: https://github.com/pygame/pygame/pull/2352

It seems to be about a 1 in 4 chance of happening. It's not related to this PR.

illume (approved):
👍 thanks.

Issue #1795, opened 2020-05-18
Add unit test: surface.convert()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)
[Draft guide on how to write a unit test for pygame](http://renesd.blogspot.com/2019/11/draft-2-of-lets-write-unit-test.html)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/surface_test.py#L809-L834
**Docs:** https://www.pygame.org/docs/ref/surface.html#pygame.Surface.convert
**Code to be tested:** https://github.com/pygame/pygame/blob/master/src_c/surface.c#L1688-L1896

elafrenz:
Hello, we would be happy to work on this unit test and should have an update in the coming week.

bwright8:
if yall don't mind, i am up for working on this

MyreMylar:
Sure.

You should likely be aware that I suspect there may be some lurking bugs in this function in the SDL1 version, I saw some oddness testing pygame's solarwolf project where `convert()` was not stripping per-pixel alpha from a surface as the docs claim it should. But that's a good reason to unit test!

bwright8:
Great to know! Thanks I’ll look into it and keep you posted.

illume:
I'll close this thanks to the PR from @bwright8

PR #2368, opened 2020-11-29
Surface.convert tests updated in surface_test.py

bwright8:
added test for Surface.convert() on a fresh clone

For #1795

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1728,32 +1728,58 @@ class GeneralSurfaceTests(AssertRaisesRegexMixin, unittest.TestCase):
 
         self.assertEqual(no_surf_alpha_col, surf_alpha_col)
 
-    def todo_test_convert(self):
+    def test_convert(self):
 
-        # __doc__ (as of 2008-08-02) for pygame.surface.Surface.convert:
+        """Ensure surfaces returned from Surface.convert()
+        have the correct pixel formats"""
 
-        # Surface.convert(Surface): return Surface
-        # Surface.convert(depth, flags=0): return Surface
-        # Surface.convert(masks, flags=0): return Surface
-        # Surface.convert(): return Surface
-        # change the pixel format of an image
-        #
-        # Creates a new copy of the Surface with the pixel format changed. The
-        # new pixel format can be determined from another existing Surface.
-        # Otherwise depth, flags, and masks arguments can be used, similar to
-        # the pygame.Surface() call.
-        #
-        # If no arguments are passed the new Surface will have the same pixel
-        # format as the display Surface. This is always the fastest format for
-        # blitting. It is a good idea to convert all Surfaces before they are
-        # blitted many times.
-        #
-        # The converted Surface will have no pixel alphas. They will be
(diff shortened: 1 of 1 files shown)


MyreMylar (changes requested):
I think the only thing not being tested is this part:

> The converted Surface will have no pixel alphas. They will be stripped if the original had them. See convert_alpha() for preserving or creating per-pixel alphas.

Which is the most interesting part to test at the moment because it currently behaves differently on SDL1 to SDL2 we believe.

With SDL1 we believe that the alpha 'losses' are set to 8 which then affects the rest of the blitting - but the blendmode/SRCALPHA flag is left intact. Meanwhile on SDL2 we believe that the blenmode/SRCALPHA flag is removed and the alpha losses left alone at 0. You can get the alpha losses using:

`surface.get_losses()[3]`

and the SRCALPHA flag using:

`surface.get_flags() & pygame.SRCALPHA`

AFAIK this hasn't been properly tested on all platforms yet so this test would be a good opportunity. To see if it currently does as work as we believe. You can branch to test only on SDL1/SDL2 with:

`if SDL1:` or `if not SDL1:`

bwright8:
> I think the only thing not being tested is this part:
> 
> > The converted Surface will have no pixel alphas. They will be stripped if the original had them. See convert_alpha() for preserving or creating per-pixel alphas.
> 
> Which is the most interesting part to test at the moment because it currently behaves differently on SDL1 to SDL2 we believe.
> 
> With SDL1 we believe that the alpha 'losses' are set to 8 which then affects the rest of the blitting - but the blendmode/SRCALPHA flag is left intact. Meanwhile on SDL2 we believe that the blenmode/SRCALPHA flag is removed and the alpha losses left alone at 0. You can get the alpha losses using:
> 
> `surface.get_losses()[3]`
> 
> and the SRCALPHA flag using:
> 
> `surface.get_flags() & pygame.SRCALPHA`
> 
> AFAIK this hasn't been properly tested on all platforms yet so this test would be a good opportunity. To see if it currently does as work as we believe. You can branch to test only on SDL1/SDL2 with:
> 
> `if SDL1:` or `if not SDL1:`

thanks for the tips and patience :)

illume:
@MyreMylar note there is a test in this PR with some alpha stuff https://github.com/pygame/pygame/pull/2340 Maybe we can add a skipif-sdl2 to the test_convert_get_alpha in #2340 and merge that? (Then keep the alpha functionality of convert separate from this PR, since it's not really clear yet?).

MyreMylar:
@illume Could do, that too - though I guess we'd need to add the `get_losses()[3] == 8` part to it as well to finish off that test. 

I'll leave it up to you @bwright8 you can stop here and we'll fix up the other PR a little to cover the alpha blending behaviour, or you could look at that one too and try and roll it all into this PR if you like. Separating it out into a new `test_convert__alpha_stripping()` function with a skip decorator is a good idea too.

illume:
OK... I'll merge this one in, and fix the other PR up with a skip for now.

Thanks @bwright8 :)

illume (approved):
👍

Issue #2367, opened 2020-11-29
intermittent arm64 crash


illume:
As seen in some PRs:
https://github.com/pygame/pygame/pull/2352#issuecomment-735371983
https://github.com/pygame/pygame/pull/2363#issuecomment-735052736

Perhaps related to this freetype reference counting issue: https://github.com/pygame/pygame/pull/2363#issuecomment-735226242 However, it could potentially be almost anything and not related to the freetype module.

I'm bisecting by re-running old arm64 jobs on travis to see if I can narrow down where the failure first starts occurring.
So far, I can get it to crash so far back as: [improve-get-locks-test](https://github.com/pygame/pygame/pull/2281) PR https://travis-ci.org/github/pygame/pygame/jobs/740447159#L2890 but nothing else can reproduce the crash before it.

~~Very confident the test in #2281 causes the crash. I ran several jobs before that one and none would crash, and could reproduce the crash in that PR multiple times, and ones after that multiple times.~~

ankith26:
If this info helps.
Noticed a same crash on my event-refactor PR. Didn’t give much attention to it, as it happened only once on a commit
https://github.com/pygame/pygame/pull/2357/commits/c853026955ad9a221592f1a3af593513433909f3
My PR is 56 commits behind master (as of now)
So we can safely assume the last 56 commits have nothing to do with this issue

illume:
It crashed even with the lock test changes skipped. https://travis-ci.org/github/pygame/pygame/jobs/746580407#L2910

illume:
Might be something to do with gfxdraw stuff... since those are the tests after the ftfont ones.

However, probably needs to be debugged with an arm64 setup.

MyreMylar:
Would these newish versions of raspberry PI OS work for debugging it do you think?

http://downloads.raspberrypi.org/raspios_arm64/images/

illume:
Yeah, I think so. There's also some newer 64bit ubuntu ones.

ankith26:
Is this issue fixed now? I think it is (can someone please confirm here?)

From what I see the aarch64 builds work consistently

PR #2369, opened 2020-11-29
Multiple locks seem to crash on arm64, skip the test on there.

illume:
For https://github.com/pygame/pygame/issues/2367

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -22,6 +22,7 @@ import weakref
 import ctypes
 
 IS_PYPY = "PyPy" == platform.python_implementation()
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
Does what it says. 👍 

Puzzling why this would happen though.

illume:
> Puzzling why this would happen though.
Yeah.

I ran the arm64 job multiple times, and it didn't fail. So I was pretty confident this stops CI randomly failing... but then...
BAM https://travis-ci.org/github/pygame/pygame/jobs/746580407#L2910

doh.

Issue #1723, opened 2020-05-12
CPU specific modules (neon, SSE4, etc).


illume:
As a user I want to install and distribute pygame where the CPU specific optimizations are included for optimized functions.

As part of the PR [Adds MMX, SSE2 & Arm NEON optimised versions of blit_blend_premultiplied()](https://github.com/pygame/pygame/pull/1711) some more SIMD optimized functions were made. However there's no runtime detection for those functions, so some are disabled by default.

illume:
This is based on some of the discussion in #1711 ...

---

Here's another way we could hack up distutils based on python modules and compilation flags. (obviously these ways would be out of scope for this PR)

Currently Setup.SDLX.in has stuff like this.
```
surface src_c/surface.c src_c/alphablit.c src_c/surface_fill.c $(SDL) $(DEBUG)
```

Adding CPU specific modules could be done by adding flags (`-mfpu=neon`) to the end like this:
```
surface_fpu_neon src_c/surface.c src_c/alphablit.c src_c/surface_fill.c $(SDL) $(DEBUG) -mfpu=neon
```
Then the init machinery (`__init__.py`) can load the correct `surface` (or `surface_fpu_neon`) module based on the CPU capabilities. I guess all the other C based routines would also be sped up because of better compiler optimizations too.

MyreMylar:
Was just quoting that :) 

I think it would require us to:

- [ ] Pull out the SSE2/ARM NEON optimised functions into a separate .c file.
- [ ] Detect support for these intrinsics at compile time (Can we use SDL here?)
- [ ] Test it to see if it works. In theory with the current state of the code the intrinsic functions are never run at runtime anyway if there is no platform support, so that part shouldn't need any changes.

illume:
Another possibility is compiling the whole wheel into a separate package.
Such as `pygame_neon` or `pygame_sse42`. This could potentially work better with other tools. Then the main pygame could try and import that.

Speaking of other tools: things like cx_freeze and pyinstaller need to be considered.

Also, we can distribute our own wheels however we want, but how can other distro packages (such as Debian pygame) take advantage of the solution?

MyreMylar:
A discussion of problems Google had with compiler specific modules using AVX:

https://randomascii.wordpress.com/2016/12/05/vc-archavx-option-unsafe-at-any-speed/

MyreMylar:
I discussed this a little bit on the end of this issue too:

https://github.com/pygame/pygame/issues/1229

[...]

PR #2373, opened 2020-12-01
Disable new alpha blitter on raspberry pi/arm

### LINKED ISSUE #2370: New alpha blitter much slower on Raspberry PI
### **Config**
RaspberyPi 3b+, latest official raspberrypi os (32bits), everything default
pygame installed from pip (and the latest dev compiled to make sure this wasn't fixed since 2.0)

### **Problem**
I was trying some pygame stuff on raspberry pi and I it seems that some strange things are going on with the new alpha blitter #2243 

Anything using per pixel alpha blitting is so much slower than with pygame1 it's sometimes unusable (10 to 20x slower)

It looks like it's related to the new alpha blitter #2243 . 
### **Tests**
Running the tests (#2243) posted by @MyreMylar. (Only the even blit width results because there is not much difference between even and odd)
**pygame 1.9.6**
```
tested Blit no alpha              : 1809.057ms
tested Blit surface alpha         : 5245.058ms
tested Blit pixel to opaque alpha : 2610.661ms
tested Blit pixel to pixel alpha  : 40830.153ms
tested Blit pixel and surf alpha  : 2603.336ms

```
**pygame 2.0.1.dev1 (SDL 2.0.9, python 3.7.3)**
With PYGAME_BLEND_ALPHA_SDL2='1'
```
tested Blit no alpha              : 1791.832ms
tested Blit surface alpha         : 4833.745ms
tested Blit pixel to opaque alpha : 2610.402ms
tested Blit pixel to pixel alpha  : 2667.952ms
tested Blit pixel and surf alpha  : 13076.423ms
```
**pygame 2.0.1.dev1 (SDL 2.0.9, python 3.7.3)**
New blitter
```
tested Blit no alpha              : 1807.6ms
tested Blit surface alpha         : 4887.749ms
tested Blit pixel to opaque alpha : 66528.607ms
tested Blit pixel to pixel alpha  : 59373.177ms
tested Blit pixel and surf alpha  : 66369.716ms
```

The results for no alpha and surface alpha are similar, but the others are much, much worse.
The biggest problem is **pixel to opaque**. It's similar with pygame1 and 2 (SDL2), but **with the new alpha blitter, it's 25 times slower** (on other platforms, like my mac, the new blitter is actually faster than the SDL2 one).

Anyway, this alpha blit optimization is way over my head 🤔  ... but hopefully this makes sense to someone 😄 

Thanks

MyreMylar:
Without NEON SIMD instructions the blitter is 10 times slower than the SDL 2 one. This will mean the blitter on Pi has lots of accuracy bugs, but at least it is fast. Later if we find away to always enable the neon SIMD instructions for arm builds we can turn this back on.

fixes #2370

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -4311,6 +4311,9 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
         }
         /* Py_END_ALLOW_THREADS */
     }
+/* disable alpha blitters on arm for now because we can't use NEON by
+   default in case it isn't available*/
+#ifndef __arm__
     else if(the_args != PYGAME_BLEND_ALPHA_SDL2 &&
             !(pg_EnvShouldBlendAlphaSDL2()) &&
             SDL_GetColorKey(src, &key) != 0 &&
@@ -4326,6 +4329,7 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
             behaved */
         result = pygame_Blit(src, srcrect, dst, dstrect, the_args);
     }
+#endif /* __arm__ */
 #endif /* IS_SDLv2 */
     else {
         /* Py_BEGIN_ALLOW_THREADS */


MyreMylar:
Haven't been able to test this on Pi, but it is a farily straightforward clipping out of the path the new alpha blitter goes down when compiled for Arm CPUs.

MyreMylar:
@galexandreg If you feel like testing this out on pi to see if it works that'd be much appreciated. Work is nuts right now over here and our toilet just sprank a leak at home.

galexandreg:
> @galexandreg If you feel like testing this out on pi to see if it works that'd be much appreciated. Work is nuts right now over here and our toilet just sprank a leak at home.

No problem
I will test it in a few hours.

galexandreg:
I tested this on my 3B+ and it seems it does what's it's supposed to do. But I wonder if it's the best scenario until it's optimized for the pi.
I mean, the SDL2 may not be backwards compatible, but 25 times slower isn't either... The difference is, the way it is now, it's possible to use `PYGAME_BLEND_ALPHA_SDL2` if it's too slow. But if it's SDL2 by default, there is currently no way to turn the pygame blitter back on.

```
tested Blit no alpha              : 1865.207ms
tested Blit surface alpha         : 4917.831ms
tested Blit pixel to opaque alpha : 2732.712ms
tested Blit pixel to pixel alpha  : 2756.407ms
tested Blit pixel and surf alpha  : 13066.596ms
```

illume:
Yeah, it's not really a good option either way. But I kind of agree with @galexandreg. At least there is a work around available with the current situation.

MyreMylar:
I would prefer it all to work nicely on Pi. I mainly made this as a stop gap solution if there was going to be a new release and the Pi stuff was holding it back. Right now, I don't really have a block of time to work on the blitters seriously until probably February/March next year.

So far we are waiting on:

 - updates to sse2neon so that the NEON code can even compile for arm.
 - some sort of solution, involving pulling all the SIMD/SSE/NEON code out into a separate c file which might make it more compatible with PI 1 (I only vaguely recall this idea)
- Obtaining a Pi 1, or Pi 1 equivalent arm testing platform.
- Testing a working built-with-neon wheel on Pi 3/4 and Pi 1 equivalent to see if the same wheel can work on both heading down two different code paths.
- Seeing if the sse2neon ported code is actually fast.
- Finally, if that all works, enabling NEON by default so that most users benefit from it.

illume:
Yeah, it's a nice stopgap. But it's a hard decision of which way to go in the short term.

> some sort of solution, involving pulling all the SIMD/SSE/NEON code out into a separate c file which might make it more compatible with PI 1 (I only vaguely recall this idea)

It might be possible to compile the same file with different flags, and then put it into a different library. So there'd be things like module_neon, and module_sse42. Then python can detect the code, and load that module (`import image_neon as image`). This has the benefit of compiler CPU optimizations also working, not just the asm/intrinsics ones. This would be done in `config_unix.py` as a distutils extension (and maybe shared in config_win/config_darwin/etc). Maybe a weekend project. This would let us use the SSE4.2 pygame.image optimizations too.

illume:
We have an issue open for [CPU specific modules (neon, SSE4, etc).](https://github.com/pygame/pygame/issues/1723), and I think that's the way I'll try to go forward with in pygame the 2.0.3 series. I feel the current slow-but-correct blitters, with SDL2 alternative available are the best of the bad options at the moment.

PR #2374, opened 2020-12-02
Update draw_test.py

ritikdh440:


siggisv:
I'm sorry but I'm going to close this.

The reason: Deleting all 6443 lines of unit tests for this module and replace them with a small interactive game that uses only one of the module's function isn't that useful.

PR #1975, opened 2020-06-23
Display gl_get_attribute unit test

jtoloff:
Please let me know if there is any functionality I failed to test or if my existing tests are insufficient. I am a first-time contributor, and I am looking to improve my skills.

Issue #1746, opened 2020-05-16
Add unit test: display.gl_get_attribute()


MyreMylar:
This function needs a unit test.

[Guide to how to contribute to an open source project on GitHub.](https://github.com/firstcontributions/first-contributions)

**Test stub:** https://github.com/pygame/pygame/blob/master/test/display_test.py#L169-L182
**Docs:** https://www.pygame.org/docs/ref/display.html#pygame.display.gl_get_attribute
**Code to be tested**: https://github.com/pygame/pygame/blob/master/src_c/display.c#L750-L761

jtoloff:
Hello, may I be assigned this issue?

MyreMylar:
Assigned!

jtoloff:
#1975

aminouhammou:
Hi ! I'm part of a group project an we'd be happy to work on this issue if it's still open.

illume:
I had to skip the test because of a failure on raspberry pi https://github.com/pygame/pygame/pull/2415


From issue https://github.com/pygame/pygame/issues/2412
```
pygame 2.0.1.dev1 (SDL 2.0.9, Python 3.7.3)
Hello from the pygame community. https://www.pygame.org/contribute.html

...

======================================================================
ERROR: test_gl_get_attribute (pygame.tests.display_test.DisplayModuleTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/pi/.local/lib/python3.7/site-packages/pygame-2.0.1.dev1-py3.7-linux-armv7l.egg/pygame/tests/display_test.py", line 254, in test_gl_get_attribute
    original_values.append(pygame.display.gl_get_attribute(pygame.GL_STENCIL_SIZE))
pygame.error: OpenGL error: 00000502
```

Harshrj09:
Hello, I would like to contribute. Please assign me.

erinsusanthomas:
Hi,
I would like to contribute, please assign me to this issue

Jumana-22:
Hello. 
I would like to be assigned to this issue.

ankith26:
The function already has tests written for it (see linked PR). It's just that those tests are skipped ATM.

Also glad to hear y'all eager to contribute to pygame :) to get started, you don't need to be assigned to an issue in order to contribute, just opening a PR straight away is fine ;)

Jumana-22:
To clarify, is the issue that the test cases are failing which is why they are skipped ATM?

[...]

PR #2375, opened 2020-12-02
adding unit test for display.gl_get_attribute()

aminouhammou:
I added a unit test for display.gl_get_attribute() by correcting some mistakes in [this previous PR.](https://github.com/pygame/pygame/pull/1975) 
 [Here](https://github.com/pygame/pygame/issues/1746) is the issue related to this unit test.

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -240,7 +240,6 @@ class DisplayModuleTest(unittest.TestCase):
     @unittest.skipIf(
         os.environ.get("SDL_VIDEODRIVER") == "dummy",
         'OpenGL requires a non-"dummy" SDL_VIDEODRIVER',)
-
     def test_gl_get_attribute(self):
 
         # __doc__ (as of 2008-08-02) for pygame.display.gl_get_attribute:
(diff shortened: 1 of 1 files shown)


illume:
Thanks for this. On my Mac, I get this failure.

```python
FAIL: test_gl_get_attribute (__main__.DisplayModuleTest)
  File "test/display_test.py", line 280, in test_gl_get_attribute
    self.assertEqual(pygame.display.gl_get_attribute(pygame.GL_DEPTH_SIZE),24)
AssertionError: 32 != 24

  File "test/display_test.py", line 282, in test_gl_get_attribute
    self.assertEqual(pygame.display.gl_get_attribute(pygame.GL_ACCUM_RED_SIZE),16)
AssertionError: 32 != 16

  File "test/display_test.py", line 283, in test_gl_get_attribute
    self.assertEqual(pygame.display.gl_get_attribute(pygame.GL_ACCUM_GREEN_SIZE),16)
AssertionError: 32 != 16

# continued commenting out tests until line 288


  File "test/display_test.py", line 298, in test_gl_get_attribute
    self.assertEqual(pygame.display.gl_get_attribute(pygame.GL_CONTEXT_RELEASE_BEHAVIOR),1)
pygame.error: OpenGL error: GL_INVALID_ENUM


```

Note this doc from https://wiki.libsdl.org/SDL_GL_SetAttribute

> This function sets the OpenGL attribute attr to value. The requested attributes should be set before creating an OpenGL window. You should use SDL_GL_GetAttribute() to check the values after creating the OpenGL context, since the values obtained can differ from the requested ones. 

This means that setting the attribute is merely a request. So the test can not rely on set actually setting the value for get. Hrmmm... Any ideas? Maybe getting the value first, and then seeing if the get is in (original_value, set_value)?

illume (changes requested):
Try doing the gets after a set_mode call. Try handling that sets are requests, and do not always set the requested value.

aminouhammou:
Hi, I have taken your suggested changes into account and tried to update the unit test. I'm waiting on your feedback :)

illume (approved):
Thanks!

PR #2384, opened 2020-12-06
Remove 'import pygame._view' in packager_imports.

Starbuck5:
Super minor change.

This is spurred by me removing the old pygame hook on the pyinstaller repo.
Since pygame._view doesn't seem to exist anymore, this line doesn't need to exist. (trying to import it throws a ModuleNotFound)

**Edit:** ugh this doesn't need to run CI, but I don't know how to stop it once it starts.

### DIFF
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -323,7 +323,6 @@ def packager_imports():
     import pygame.macosx
     import pygame.bufferproxy
     import pygame.colordict
-    import pygame._view
 
 # make Rects pickleable
 if PY_MAJOR_VERSION >= 3:


ankith26 (approved):
Nice cleanup :) 
Since pygame._view does not exist anymore, we must not be importing it.
Thanks 👍 

CI has failed due to some random issue that occasionally happens, not related to this PR.
BTW. The function we are cleaning up does not get called anywhere. It has been in that state since years. And no one complained about the error that it gives while importing pygame._view.
Maybe remove the function as well (unless it is serving a purpose I am not aware of)

Starbuck5:
The function just exists so that programs that freeze python code into executables will see those specific imports.

This is a more general solution than a hook, because it works for things other than pyinstaller: like py2app/py2exe, as mentioned in the function comment.

Now that pygame doesn't have any _view module that needs to be seen by these programs, this statement can be removed.

ankith26:
I did search the Internet for the usage of this packager_imports function. And i didn’t find any meaningful info.
Probably it’s an obsoleted feature of py2app/py2exe

Starbuck5:
The packager imports function was never supposed to be run. It exists so that programs parsing the code for dependencies can see that “it imports x” without the program actually importing x.

ankith26:
hmm yeah you are right. I just got a bit confused

illume (approved):
👍 thanks

PR #2385, opened 2020-12-08
Drop PyLong_FromUnicode, removed in python 3.10 [skip ci]

ankith26:
PyLong_FromUnicode will be removed in python 3.10 (which is coming soon).
So we should drop it as well.
See
https://docs.python.org/3.10/whatsnew/3.10.html#id4


MyreMylar (approved):
LGTM 👍  We'll also need to upgrade our version of Cython at the time of 3.10 as the current version of the generated files also include this function.

ankith26:
@illume this PR is ready to get merged too. This is a super minor change, that does not affect anything. So this should be an easy one

Thanks :)

illume (approved):
👍

PR #2387, opened 2020-12-08
docs, tests and types for pathlib support

illume:
For https://github.com/pygame/pygame/issues/1129

Updates some docs, tests and types for PathLike support.

- [x] test_image load/save
- [x] image.rst docs
- [x] image.pyi typing
- [x] test Sound 
- [x] mixer.pyi typing
- [x] font.pyi typing
- [x] freetype.pyi typing

### DIFF
--- a/buildconfig/pygame-stubs/font.pyi
+++ b/buildconfig/pygame-stubs/font.pyi
@@ -1,5 +1,11 @@
 from typing import List, Optional, Union, Tuple, IO, Hashable, Iterable
 
+if sys.version_info >= (3, 6):
+    from os import PathLike
+    AnyPath = Union[str, bytes, PathLike[str], PathLike[bytes]]
+else:
+    AnyPath = Union[Text, bytes]
+
 from pygame.color import Color
 from pygame.surface import Surface
 
@@ -30,7 +36,7 @@ class Font(object):
     italic: bool
     underline: bool
 
-    def __init__(self, name: Union[str, IO, None], size: int) -> None: ...
+    def __init__(self, name: Union[AnyPath, IO, None], size: int) -> None: ...
     def render(
         self,
         text: str,
--- a/buildconfig/pygame-stubs/freetype.pyi
+++ b/buildconfig/pygame-stubs/freetype.pyi
@@ -1,5 +1,11 @@
 from typing import Tuple, Optional, Union, List, Text, IO, Sequence, Any, Iterable
 
+if sys.version_info >= (3, 6):
+    from os import PathLike
+    AnyPath = Union[str, bytes, PathLike[str], PathLike[bytes]]
+else:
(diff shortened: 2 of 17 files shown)


ankith26:
Thanks :) 
I was about to start this work tomorrow... but seems like you have taken it up

ankith26:
Nice!

Little nitpick, 
In the docs, you have specified for many functions that it accepts pathlib.Path.
But those functions also accept all other pathlib objects (there are 6 of them)
So maybe you can change that to mention that any pathlib object will work, or you can also mention that os.pathlike objects will work

Otherwise, this PR is looking very good to me :)
Cheers! 👍

illume:
1)  how to describe it?

> In the docs, you have specified for many functions that it accepts pathlib.Path.

Yes, that is a challenge. pathlib is quite complicated, and I couldn't think of a good way to say "a PathLike pathlib *Path* object". The problem with "PathLike" is it's not entirely clear to many python programmers that pathlib even exists.  Probably Path is the most well known ones, and including the namespace might clue people in more than "PathLike".

I chose pathlib.Path instead of the other pathlib.*Path* ones because in the [pathlib docs](https://docs.python.org/3/library/pathlib.html) it starts with: **"If you’ve never used this module before or just aren’t sure which class is right for your task, Path is most likely what you need."** This suggests to me that all the types are confusing, and casual users mostly need to know about pathlib.Path. Explaining all the possibilities is a confusion of types. I think it's not necessary to iterate over all the pathlike types. They are mostly things people don't need to care about. There are other PathLike types outside of pathlib as well, since PathLike is an Abstract Base Class for implementing other versions.

A link to pathlib/PathLike docs in there is useful?

>  So maybe you can change that to mention that any pathlib object will work, or you can also mention that os.pathlike objects will work

PathLike is in the types, which are of course are not all mentioned in the html docs (yet). I'm not sure if we came up with a clever plan to use the type stubs in the docs automatically, or if I should add `param` notes. So... I'll add the param notations?

I expect it's clear to people who know what the other pathlib classes are will know it's PathLike (without looking at the types).


2) The other thing I'm not sure about is the AnyPath type used. That was inspired by the builtin for `open()` on typeshed. It's not the cleanest code to import however. Do you think it's ok @MightyJosip ?

I don't like repeating the AnyPath implementation everywhere... but I don't see a shared spot to put it in the type stubs. Maybe there's a more modern cleaner way to do a "AnyPath". The other thing is I took the int type out of there (`open()` can work with file descriptor integers).

Maybe it should just just include all of them in a Union, rather than making a new AnyPath? Probably cleaner.

illume:
I've added some other tasks to the #2387 issue about param documentation notes.

ankith26 (approved):
LGTM 👍 
Cheers! :)

PR #2389, opened 2020-12-09
Capitalise Python in the import message

TomFryers:


MyreMylar (approved):
This seems to be in tune with the official python style guide:
https://docs.python.org/3.1/documenting/style.html

ankith26 (approved):
Thanks for the contribution 👍

PR #2391, opened 2020-12-10
Add relative mouse movement support

mcpalmer1980:
Added simple getter and setter for relative mouse movement for _sdl2.video

fixes ##2390


ankith26:
Nice!
I’m not much into Cython stuff, so I can’t really review the code. But it’s a good idea to commit the Cython generated C code along with your PR.
The command to do that looks something like
python3 setup.py cython

Issue #2390, opened 2020-12-09
_sdl2.video set_grab() and relative mouse movement


mcpalmer1980:
Calling mouse.set_grab(True) and event.set_grab(True) doesn't enter virtual mouse mode when using sdl2.video

I assume this stems from using video.Window() instead of display.set_mode()

Also, a very small issue, you can set texture.color with a tuple, but you need to use a color object to set image.color.

mcpalmer1980:
I agree that my fix was less than ideal. But pg_GetDefaultWindow() doesn't return a video.Window and I have no idea how to fix that. Since the _sdl2 API is experimental I thought it best to add this hack there instead of messing with the official pygame stuff. Also, there is some sense in it being in Window because it effects how input is received in relation to the window.

When a final sdl2.video API is released, pg_GetDefaultWindow() will probably return the active video.window (if there still is one) and the problem will have fixed itself. Admittedly, I really like the experimental sdl2 API how it stands.

PR #2392, opened 2020-12-10
Added relative mouse motion support to video.Window

mcpalmer1980:
Added simple setter and getter to pygame.src_c/cython/pygame/_sdl2/video.pyx
The new property is pygame._sdl2.video.Window.relative_mouse
Simply calls SDL_GetRelativeMouseMode and SDL_SetRelativeMouseMode
Compiled to video.c

solves #2390


illume (approved):
👍 thanks!

mcpalmer1980:
I don't understand anything in the LGTM C analysis details, and I'm not sure if the c changes should be included in PR or just the pyx changes.

illume:
The LGTM failure seems to be a robot-brainfart on its part - nothing to worry about.

illume:
Looking at this again, it's not really a Window feature. But more a mouse feature.

I'll merge it in to unblock you, but in time the API will change. Also, perhaps the display versions should be fixed to work with Window. Will leave the issue open so it can be investigated still.

PR #2396, opened 2020-12-15
Update README.rst

Starbuck5:
I added a link to the online docs in place of saying "docs/index.html" - I don't think releases or github has a compiled copy of the docs, just the website.

I edited the first paragraph, I think it sounds better now.

I deleted the contributors section because we already link to pygame.org/contribute, but I wanted to keep the word "contribute" in there so I added the sentence "new contributors are welcome."

I pulled everything about building from source out of the install section into its own section, as not to confuse new people who just need to see "pip install pygame".

I removed a reference to pygame.org/downloads.shtml since that seems to be obsolete in the era of PyPI (the pygame 2 release never landed there).

I also updated it to read "commits since 2.0.0" instead of since 1.9.6.

[skip ci]

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,13 +5,14 @@ pygame
 |PyPiVersion| |PyPiLicense| |Python2| |Python3| |GithubCommits|
 |LGTMAlerts| |LGTMGradePython| |LGTMGradeC| |Coverity|
 
-pygame_ is a Free and Open Source cross-platform library
-for the development of multimedia applications like video games on Python.
+pygame_ is a free and open-source cross-platform library
+for the development of multimedia applications like video games using Python.
 It uses the `Simple DirectMedia Layer library`_ and several other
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
Looks good to me 👍 
Cheers :)

illume:
Cool.

> I removed a reference to pygame.org/downloads.shtml since that seems to be obsolete in the era of PyPI (the pygame 2 release never landed there).

Good idea.

ps. I guess we should be pointing most people to: https://www.pygame.org/wiki/GettingStarted
(Because it has platform specific notes in there) Maybe even removing the pip install part altogether? Unfortunately 'pip' isn't often available, and is often different to the pip people are using. If we keep pip instructions in here probably using `python -m pip` would be better (not that python works on many windows installations or is the right python for pycharm people, or on systems with python2 and python3).

illume (approved):
👍 thanks

PR #2404, opened 2020-12-19
Event refactor docs/types

ankith26:
This is a WIP PR to do all docs and types for event-refactor PR stuff


ankith26:
The only things left to do are
1. Docs and types for pygame.time.set_timer after we finalize it’s a API details
2. Make any changes if required to WINDOWEVENT docs, after it gets an API review

illume:
Is there a behaviour change with event.post() returning a bool?

ankith26:
There is no behavior change as such. Places where the function returns None, I have just made it to return True/False wherever appropriate. It’s a useful little addition, if you ask me. Because now people can know that the event did not get posted (because it was blocked)

And where the function used to raise errors before, that one has not been changed

illume:
Very good. Thanks for the clarification.

illume (approved):
👍

Issue #2386, opened 2020-12-08
Adding a rect.scale(percent)


mcpalmer1980:
Being able to easily scale a rect by a percent is vital to the real time scaling possible with sdl2.video. 

I rarely use the rect.inflate() method as is. Whenever I need to scale a rect I want to do it by a percent while maintaining its aspect ratio. To increase performance, I tend to unwrap my helper function into many inner loops to increase performance. Ideally I would like to do it in a single function call from within python, hence a rect.scale(percent) instead of my own scale_rect() that calls rect.inflate().

I am writing a GPU tilemap renderer in python, which will include camera and zoom transforms that must be applied to all Images. This means lots of scaling of sprites (but only once for the tilemap, so that's easy). I already scale particles, and that's enough of a headache(unnecessary performance loss hurts my head).

I could write this myself if adding this feature to the rect class makes since to developers. 

I would also like to write my tilemapper in cython, but as a beginner, it's proven difficult. I can't even get the cython class defs imported into my own pyx module. I don't want to open an issue, but has there been consideration of exposing the sdl cython stuff to external modules? Or perhaps it will be rewritten in C?

mcpalmer1980:
There has been no comment on this yet. Does nobody have an opinion?
I would gladly add this myself if anyone other than me thinks it is a good idea.

illume:
I'm not against `rect.scale(percent)`. I haven't given it much thought.

I guess usually people want to scale keeping the aspect ratio, or just keeping it in one dimension. eg, give me a rect 400 wide, with the height keeping the aspect ratio. Also, where do you scale it from? The center, or the bottom left?

mcpalmer1980:
Keeping aspect ratio is important, yes. But also if you're trying to scale many objects, perhaps based on the scale of a tilemap or some other global zoom value, you're only going to have a percentage, not a value in pixels along the 2 dimensions. And you may have many dozens, hundreds, or thousands of rects that need modifying.

The scale will have to be from the center, of course. Here is the python code I use.

```
def scale_rect(rect, amount): # multiplier, 1=100% or unchanged
	c = rect.center
	w = rect.width * amount
	h = rect.height * amount
	new = pg.Rect(0,0,w,h)
	new.center = c
```

PR #2409, opened 2020-12-21
Adding Rect.scale() and scale_ip() [ED: someone please finish this off?]

mcpalmer1980:
Here is my attempt at scaling Rect objects by a multiplier as suggested in #2386. It works great for me. 

Rect.scale(float scale) -> Rect
Rect.scale_ip(float scale) -> None

2 small issues worth conversation.

**center value** When scaling odd numbered rects the center can be 1 pixel off. Not sure if this is an issue or not. I altered the values used for testing to avoid failed tests. 

**0 and negative scalars** I thought about asserting if scale is 0 or less to avoid divide-by-zero or hilarity. My tests pass with 0 and negative values, however, so I decided not to. The resulting negative scaled rects is wrong though. Not sure if I should throw an error, or refuse to scale, or set the rect to 0 size, or what.


Starbuck5:
I think this is a good idea that can be extended to scaling surfaces.

I had a conversation with mcpalmer about this on discord, and they said that their main use case for scaling rects like this is to enable scaling Textures/Images in an easier and faster way.

```
Texture.draw(srcrect=src_rect, dstrect=dest_rect)
Image.draw(dstrect=dest_rect) # much easier to scale dest_rect to make scaling things with this PR
```

Right now scaling a Surface by a multiplier/percent/scalar is also more complex to implement for the end user than it needs to be.

This is the code I use:
```
# scales an image by a scalar
def scale(pic, scalar):
    x = pic.get_width() * scalar
    x = round(x) if x > 0 else 1
    y = pic.get_height() * scalar
    y = round(y) if y > 0 else 1
    return pygame.transform.scale(pic, (x, y))
```

The only problem with adding a `.scale(float scale)` to the `pygame.transform` module is that `transform.scale` already exists.

**I think it would be elegant to have a unified `Rect`/`Surface` scaling interface, so I have two options to propose.**

**Option 1:**

Add `pygame.transform.scale_percent(Surface, scalar, scale_function = pygame.transform.scale, DestSurface = None)`
Change this PR's `Rect.scale(float scale)` to `Rect.scale_percent(float scale)`
Change this PR's `Rect.scale_ip(float scale)` to `Rect.scale_percent_ip(float scale)`

(Of course it could use a different name than percent, perhaps multiplier)

**Option 2:**

Change `pygame.transform.scale()` and `pygame.transform.smoothscale()` so that they also accept a float/int scalar instead of a sequence of two ints.
Change this PR's `Rect.scale()` and `Rect.scale_ip()` so that they also accept a sequence of two ints instead of just a float/int scalar.

ankith26:
Even I had the idea to have scaling surfaces, while maintaining aspect ratio, but that one did not get any traction. Mainly because that was easy to implement from the users end.
I’m still liking that idea though, and even this rect scaler sounds good to me

illume:
I guess the other thing to decide is if it should be `0-100`, or `0.0 - 1.0`? I tend to find the 0-1 range more common and useful... except scaling images tends to use percent for some reason (so I'd be happy with percent).

illume:
@Starbuck5 good point. It does make sense to think about other scale function APIs so the concepts are unified.

mcpalmer1980:
So I guess the consensus is edging toward using a percent instead of a scalar. I prefer the scalar but will accept and impliment the percentage if it's preferred.

Also, while working on the menu for renderpyg [(my sdl2.video add-on library)](https://github.com/mcpalmer1980/renderpyg), I've discovered a use case for separate x and y scalars when drawing the ninepatch. So I guess it will be better to do something like this. 

```py
def scale_percent(x, y=0):
    xscale = x / 100
    yscale = y / 100 or xscale

   do scaling
```
python->c

illume:
hehe. I love your renderpyg logo. 

Perhaps it should use width, and height percent/scale rather than x,y?

ps. Just to be clear, I usually prefer a 0-1, not percent. But I guess many people also use/prefer percentage. In css for example, you can use various different units (percent, pixels, verticle height, relative to font, etc). It's similar to radians/degrees, in that different people feel more comfy either way.

mcpalmer1980:
@illume 

I'm glad we both prefer the scalar. You usually say 0-1 for simplicity I think, but it can definitely go above 1 (2=200%). It's probably me being stupid, but I want to make sure we're on the same page.

It seems to me 100% is used more for graphic design and 1.0 scalar is more common in real time games. But I could be wrong.

Yeah x and y aren't good names. Not sure what's best, but I have a few ideas.

- scale(scalar, scalary)
- scale(perX, perY)
- scale(pX, pY)
- resize()
- scale_percent()
- percent()
- multiply()

I posted that renderpyg link so you would get an idea how much I've used sdl2.video(most others are aware). Not only is Flyboy using the API exclusively, but I've gotten about 60% of my renderpyg features implemented. I will probably rewrite it partially in c/cython once my API proves itself and the **__** is removed from _sdl2.video. I will probably be filing a few issues regarding the little bugs I've found in the video API soon.

Really just 2: Texure.color supports tuple but Image.color requires pg.Color object, and you can't set Image.origin to None, which SDL2 uses to make origin default to the center, which is used 90% of the time and saves a couple divisions for each rotating object onscreen.

Since I don't know where else to find you:

image

mcpalmer1980:
Without consensus on the other suggested changes, I have updated my PR to include an optional height scalar, as well as the documentation for this.

Starbuck5 (changes requested):
Hi.

I noticed you forgot to change a couple things to `scale` / `scale_ip` language, like the method name in the docs, and the `DOC_` things in the method creation.

Once you've done that, running `setup.py docs` will update the `rect.h` header file, so that `DOC_RECTSCALE` and `DOC_RECTSCALEIP` will actually exist, iirc.

Starbuck5:
Also, if we go with width and height being values instead of scalars, then it would be easy to reconcile this function with `transform.scale` by giving it the option for a single overall scalar.

mcpalmer1980:
Yeah I don't know how I forgot to change the doc binding after doing it right the first time.

I believe using scalars will fit far more use cases then actual widths and heights, which users are unlikely to know and requires an extra step to calculate. I used width and height for the parameter names in error. I changed it to scalex and scaley in the documentation to avoid confusion but not in the source as it might be premature since I'm undecided on the proper names.

**width, height**: potentially confusing
**x, y**: used by inflate but measures pixels, not scalars
**scalex, scaley**: probably the right way.

I offer 2 definitions in the docs
- scale(scalar)
- scale(scalex, scaley)

I neither implemented nor see need to offer keyname arguments, so I don't see much reason to change the width and height variables in code. Perhaps I'm wrong and should offer scalar, scalex, and scaley as keyword arguments, where scalex/scaley will override a single scalar value?

illume:
It's failing because of the generated docs I think.

```
[2021-01-04 17:40:20] [build] x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -Wdate-time -D_FORTIFY_SOURCE=2 -g -fdebug-prefix-map=/build/python2.7-8kFkPd/python2.7-2.7.17=. -fstack-protector-strong -Wformat -Werror=format-security -fPIC -D_REENTRANT -I/usr/X11R6/include -I/usr/include/SDL2 -I/usr/include/python2.7 -c src_c/rect.c -o build/temp.linux-x86_64-2.7/src_c/rect.o
[2021-01-04 17:40:20] [build-err] src_c/rect.c:1138:57: error: ‘DOC_RECTSCALE’ undeclared here (not in a function); did you mean ‘DOC_RECTCOPY’?
[2021-01-04 17:40:20] [build-err]  1138 |     {"scale", (PyCFunction)pg_rect_scale, METH_VARARGS, DOC_RECTSCALE},
[2021-01-04 17:40:20] [build-err]       |                                                         ^~~~~~~~~~~~~
[2021-01-04 17:40:20] [build-err]       |                                                         DOC_RECTCOPY
[2021-01-04 17:40:20] [build-err] src_c/rect.c:1146:6: error: ‘DOC_RECTSCALEIP’ undeclared here (not in a function); did you mean ‘DOC_RECTCLIP’?
[2021-01-04 17:40:20] [build-err]  1146 |      DOC_RECTSCALEIP},
[2021-01-04 17:40:20] [build-err]       |      ^~~~~~~~~~~~~~~
[2021-01-04 17:40:20] [build-err]       |      DOC_RECTCLIP
[2021-01-04 17:40:21] [build] ---
```

The scale(scalar), or scale(scalex, scaley) seem good to me. (I can see how someone would also want to scale in different ways though, and I see the benefit of consistency between different API functions.)

Keyword arguments would be nice if you want to only scale in one direction (but keep the aspect ratio). They could also allow using width/height. `scale(width=250) -> (rect width==250 and height scaled keeping aspect ratio)`.

novialriptide:
Hi, I believe I can finish this off. Unless someone is already working on this issue.

mcpalmer1980:
Now that transform.scale_by() has been implemented in an official build, its convention should probably be followed here and rect.scale_by() can finally be finished. See [#2723]( https://github.com/pygame/pygame/pull/2723)

dr0id:
@novialriptide  you still want to take over? If not I'm happy to try to get it done.

Actually not sure how to take over. Would I have to open a new PR on my branch?

novialriptide:
> @novialriptide  you still want to take over? If not I'm happy to try to get it done.
> 
> Actually not sure how to take over. Would I have to open a new PR on my branch?

Go ahead. :D

dr0id:
See PR #3077

PR #2410, opened 2020-12-21
Fix render_to returning unpositioned rect

### LINKED ISSUE #1182: pygame.freetype.Font.render_to() returns unpositioned rectangle
The docs say [here](https://www.pygame.org/docs/ref/freetype.html#pygame.freetype.Font.render_to) about `render_to()` : 

> The return value is a rectangle giving the size and position of the rendered text within the surface.

But I can see that the rectangle does not have the given position, e.g.:

```
text_rect = self.game_font.render_to(self.game_window, (40, 350), "Hello better World!")
text_rect
<rect(0, 14, 296, 14)>
```

I would have expected <rect(40, 350, 296, 14)>. Did I misunderstand something?

**Related Docs:** https://www.pygame.org/docs/ref/freetype.html#pygame.freetype.Font.render_to

### LINKED ISSUE #1866: freetype.Font.render_to() returns an offset rect rather than a position rect

zoldalma999:
Fixes #1866, Fixes #1182

### DIFF
--- a/src_c/freetype/ft_render.c
+++ b/src_c/freetype/ft_render.c
@@ -393,8 +393,8 @@ _PGFT_Render_ExistingSurface(FreeTypeInstance *ft, pgFontObject *fontobj,
     render(ft, font_text, mode, fgcolor, &font_surf,
            width, height, &surf_offset, underline_top, underline_size);
 
-    r->x = -(Sint16)FX6_TRUNC(FX6_FLOOR(offset.x));
-    r->y = (Sint16)FX6_TRUNC(FX6_CEIL(offset.y));
+    r->x = (Sint16)x;
+    r->y = (Sint16)y;
     r->w = (Uint16)width;
     r->h = (Uint16)height;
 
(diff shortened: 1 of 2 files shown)


illume (approved):
Thanks! test_freetype_Font_render_to passes for me locally with this change.

Issue #2177, opened 2020-10-12
Consider adding sub millisecond Clock to timer module for SDL2


Lobsternator:
- **Operating system** : Widnows 10
- **Python version** : 3.7.3
- **SDL version** : 2.0.12
- **PyGame version** : 2.0.0.dev14

The `Clock.tick_busy_loop` function is consistently too fast for the target fps. `Clock.get_fps()` consistently measures around 62.5 fps when I pass in 60 as the target fps and a manual measurent using `time.perf_counter` is even faster (~65 fps). For clarification, the manual measurement is an average taken over the span of two seconds.

**Test code**

```
import sys, time, pygame
pygame.display.init()

size = width, height = 320, 240
screen = pygame.display.set_mode(size)
clock = pygame.time.Clock()
fps = 60
measures = []

while True:
    start = time.perf_counter()
    for event in pygame.event.get():
        if event.type == pygame.QUIT: sys.exit()
        
    clock.tick_busy_loop(fps)
    end = time.perf_counter()

    measures.append(1 / (end - start))
    if len(measures) > fps * 2:
        del measures[0]

    print(f"manual: {sum(measures) / len(measures)}, pygame: {clock.get_fps()}")
    pygame.display.flip()
```

MyreMylar:
I don't think `busy_loop()` is perfectly accurate, but your start & end are in the wrong place surely? `clock.tick_busy_loop(fps)` delays based on the time since it was last called, but your timing test is excluding the time taken to do all of this:

```
    measures.append(1 / (end - start))
    if len(measures) > fps * 2:
        del measures[0]

    print(f"manual: {sum(measures) / len(measures)}, pygame: {clock.get_fps()}")
    pygame.display.flip()
```

...in the loop. So I would expect the delay to be less than required for 60FPS because `print()` is usually fairly slow at least and is going to eat into some of the frame time, reducing the amount of time `busy_loop()` would need to delay. You'd be better using start and end to capture the whole loop, and perhaps recording your data each loop instead of printing it - and then produce a timing report at the end when you have quit out of the loop.

Lobsternator:
Oh, I see. How about this then? When I do this `Clock.get_fps()` reads the same as the manual measurement.

```
import sys, time, pygame
pygame.display.init()

size = width, height = 320, 240
screen = pygame.display.set_mode(size)
clock = pygame.time.Clock()
fps = 60
measures = []
t = 0
last_t = 1/fps

while True:
    t = time.perf_counter()
[...]

Issue #304, opened 2016-07-12
Pygame Clock inaccurate


illume:
Originally reported by: **Don Polettone (Bitbucket: [DonPolettone](https://bitbucket.org/DonPolettone), GitHub: [DonPolettone](https://github.com/DonPolettone))**

----------------------------------------

Hi all,

this is something that's been making me mad for months, if not years, and I hope someone can finally fix this:

The clock seems not doing well in pygame; animation stutters / jitters. First I thought I was doing something wrong, but when I digged deeper and broke it down to a test as simple as can be, I recognised it must be an issue with pygame itself.

Try THIS out:

```
#!python

import sys
import pygame
pygame.init()

RES = (640, 480)
FPS = 60
BLACK = (0, 0, 0)
WHITE = (255, 255, 255)
BLUE = (0, 0, 255)

display = pygame.display.set_mode(RES, pygame.FULLSCREEN)
clock = pygame.time.Clock()

bg_img = pygame.Surface(RES)
bg_img.fill(WHITE)
for y in (229, 290):
    for x in range(40, RES[0], 40):
        bg_img.set_at((x, y), BLUE)

def motion_test():

    rect = pygame.rect.Rect((0, 240), (40, 40))

    while True:

        display.blit(bg_img, (0, 0))
        
        for e in pygame.event.get():
            if (e.type == pygame.QUIT or
                e.type == pygame.KEYDOWN and e.key == pygame.K_ESCAPE):
                pygame.quit()
                sys.exit()
        
        rect.x += 1
        if rect.x >= RES[0]:
            rect.x = 0
        
        pygame.draw.rect(display, BLUE, rect)
        
        pygame.display.flip()
        clock.tick(FPS)

if __name__ == "__main__":
    motion_test()


```
Every few seconds, the rect stutters a bit. In a pixelated game, this looks awful. Especially when scrolling over pixely tilemaps. I tried it out on both Mac and Win - it is a general issue.

Can someone help/test/agree/decline/fix this ??

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/304

**Related Docs**: https://www.pygame.org/docs/ref/time.html#pygame.time.Clock

MyreMylar:
I was tinkering with this today and observed the same thing as described above - it seems the only way to get completely smooth screen scrolling/movement in pygame 1.9 is to trick the system into enabling vsynch to control the framerate. All the other combinations I've tried using the pygame clock you get a slight hitching every second or so.

[...]

PR #2411, opened 2020-12-22
Add new module pygame._time

ankith26:
continued from #2271 

Related issues: #304 #2177 
**To summerise**
This adds a new experimental module ``pygame._time``, based on ``pygame.time`` with clocks of much better accuracy.

**Internal changes**
It uses the new time-related functions of SDL2 (yup, `SDL_GetPerformanceCounter`), using the extended accuracy provided to give float-millisecond accuracy to the functions in the time module.
This clock uses the most accurate clock available on the system. On most windows platforms, this has microsecond precision and on most Unix-based platforms (Linux, Mac, etc) this gives nanosecond precision (fallback to microsecond precision in rare cases). In worst cases (extremely rare) it may fall back to using millisecond accuracy.

**Usage changes**
1) The function `pygame._time.wait()` now accepts the millisecond argument as an int or float(previously it was only int) and also returns a float, the function is made much more accurate by making it such that it sleeps the processor for some time and does a busy loop to delay for the time requested by the user.
2) The function `pygame._time.delay()` is made an alias to the function `pygame._time.wait()`, now that the function `pygame._time.wait()` is accurate.
3) The `pygame._time.Clock` object now is much more accurate in the method for `tick()` which now uses the time module to do its busy loop to give sub-millisecond accuracy, this means that the game will run at (almost) exactly the FPS you want, rather than rounding it to some arbitrary FPS. The `tick_busy_loop` method is aliased to `tick()` now that it’s accurate.
4) The methods `get_time()` and `get_rawtime()` now return floats, for extended accuracy

**Other Changes**
1) The Unit tests for this new time module are revamped to be more CI friendly, they are also sped up to take half of the original time to run, and they are strictly checking stuff at the same time. 
This is mainly due to the fact that now, if a test fails, it will re-run, just to eliminate any errors due to OS scheduling and CI VM's. Don't worry, the test wont run forever on failure, after a particular point, it gives up and fails
The tests are still blocked from being run on Mac VM, though


ankith26:
How do we proceed further on this one? The only reason I created _time to be separate from time, is because it should be easier for people to test stuff, and merge this into the time module later on (in some version like 2.1) 

Is this a good idea? Many on discord think that introducing a new module in pygame might not be a good idea, and I sort of agree. What else can we do about this?

mcpalmer1980:
I would like to see this come in ASAP no matter the form. I don't mind having a _time module that I can import as time. On the other hand, replacing the existing time module with a more precise method of delay would be great too, as long as anything that may adversely effect performance or compatibility are optional.

**Option 1**
```py
try:
    from pygame import _time as time
except:
    from pygame import time
```
**Option 2**
- The existing time module uses the new more precise method introduced by ankith without any sort of performance chewing or change of return values.
- tick_busy_loop() and similar functions are updated to use the new more precise method introduced by ankith, including a busy loop near the end of the delay and any other code that adversely effects performance.
- A float=False parameter is added to any time methods that can return floats. They continue returning ints by default unless the float=True parameter is set.

ankith26:
IMHO, this PR is the easiest to merge in it’s current form, as a seperate experimental module. So maybe _time is the right way to proceed on this

mcpalmer1980:
I would love to see a Clock.get_ticks() method that returns the total number of ticks it has performed. This would be useful if your game objects hold a clock reference and need to know if it is time to update its state. For instance, if your sprite class (or the one in renderpy) can be drawn as an image as many times as you want but only call update() if Clock.get_ticks() has incremented.

Even if you could just expose the current fps_count variable it would sort of work, but a total frame counter would be better that just 1-10. Most pygame users probably store total_frames in a variable and are forced to pass either the frame counter or a time delta all around their program. The clock probably should have had a frame counter included from the beginning.

Any thoughts?

ankith26:
The idea sure sounds good, but I’m not sure of all it’s usecases. In how many situations will this feature be used?

mcpalmer1980:
Well I certainly can't say how many people would use it or in what situations, but I can speak for places I would have used it.

All of my games have a frames variable incremented in the game loop. 
- I use it for determine time played (it isn't incremented during pauses, menu screens, or dialogue)
```py
	def ShowScore(self):
		if not self.mapdb.last_time:
			return
		time = self.frames * (1000 / 30.0)
```
- I use it for character and item flashing 
I prefer to do blinking  a little differently:
  ```py
    if frames % 30 < 20:
      character.draw()

    if self.player.gloves > 0 and self.frames % 10 < 6:
      self.s_icons.DrawAt(x,y,8)
  ```

These uses require me to set frames as an attribute of my game class, make the class available to all other classes, and manually increment the count during each counted frame. Since renderpyg needs to be design agnostic with minimal assumptions, I have to think differently. My sprite class has an update() method that animates it based on a time delta. Then it is drawn with a separate draw() call.

I would like to offer a way to allow sprites to self animate and be drawn just like a Texture or Image. In order to facilitate this you must assign a reference to your game's clock. Then the sprite will update() whenever draw() is called if that clock reference exists, and the sprite may be passed as an image to a menu class. The menu will then automagically animate the image without any knowledge of the sprite class. But an Image may be drawn() several times throughout the menu, and you don't want it too animate faster for each time it gets drawn, so I need a way to determine if the clock has ticked since the last draw() call.

Here is the code for my current use case:
```py
    def draw(self, dstrect=None):
        """
        Render the sprite at its current position or at the postion of dstrect
        Use set_pos() and set_frame() modify where and how to draw it
    
        Sprites will animate if Sprite.clock references a time.Clock object
        This allows the sprite to be drawn as a normal Image without the caller's
        knowlege of animation
        """
        if self.clock and self.clock.get_ticks() > self.last_update:
             self.last_update = self.clock.get_ticks()
             self.update(self.clock.get_time())
        
        if dstrect:
            self.image.draw(dstrect=dstrect)
        else:
            self.image.draw(self.dest_rect)
```

ankith26:
Interesting ideas. It shouldn’t be hard to implement this one, so I am for it. 
Maybe get suggestions of more people, as it’s bit of an API change. I (or anyone else, if interested) could also do it in a separate PR, after this one has been merged

Starbuck5 (approved):
I say we merge this, on the principle that it doesn't do any harm.

This can be another experimental module like the _sdl2 modules, allowing power users to play with it and hopefully find any issues, should they exist.

I've tested it locally and it works. 
I don't know how to test it against an objective timer, or test relative CPU usage: so "it works" is the best I can do.

The backwards compatibility problems with returning `float` rather than int are only for python 2, because in python 3 float division is the default. So perhaps when we drop python 2 support this module could become the main time module.

Of course its been a while since the PR was introduced, so there are some small changes that are needed to be fully ready for merge, like changing "new in 2.0.1" to "new in 2.0.2."

Starbuck5:
Now that we're on Python 3 we could do some of this stuff incrementally?

Change some return values to floats? Change around some internal arithmetic to round better, without even having to leave the old sleep methods?

PR #2414, opened 2020-12-22
Little doc fix in music docs

### LINKED ISSUE #2413: pygame.mixer.music.stop() triggering music endevent is not documented
This took me way too long to figure out. It's something I had noticed on Windows before and was confused by it, but when it happened on my system too with the Pygame 2 upgrade, I decided it was time to stop putting it off. Basically, I use `pygame.mixer.music.set_endevent()` to play a "loop portion" of music after a "start portion", but I noticed that the "start portion" was being skipped. After some hours investigating this, I was finally able to figure out why it was doing this and put together a simple test case:

[musictest.zip](https://github.com/pygame/pygame/files/5728012/musictest.zip)

In short, `pygame.mixer.music.stop()` triggers the music end event, and this is not documented. What this means is that when my code was first stopping music playback, then playing the "start" music, and then ultimately using a "music queue" system triggered by the Pygame music end event, that event was triggered immediately (rather than at the end of the newly playing track) because of that `pygame.mixer.music.stop()` call. So to reproduce it in this example, I had to start the "start" portion, stop it, and then start it again. If you remove the stop() call there, it works as expected.

I'm not entirely sure if this is a bug in the way the music end event works, or if it's an intentional behavior that's not properly documented. Please do let me know, and if the behavior is indeed intentional, I'd be interested in knowing what the preferred way to work around this is (and that should probably be added to the documentation as well, since catching your own `pygame.mixer.music.stop()` commands probably isn't something anyone is likely to do).

Tested on Debian 10 "Buster" with Pygame 2.0.0 compiled on my own system.

🕵️

ankith26:
closes #2413

### DIFF
--- a/docs/reST/ref/music.rst
+++ b/docs/reST/ref/music.rst
@@ -86,7 +86,8 @@ can crash the program, ``e.g``. Debian Linux. Consider using ``OGG`` instead.
    | :sg:`stop() -> None`
 
    Stops the music playback if it is currently playing.
-   It Won't Unload the music.
+   endevent will be triggered, if set.
+   It won't unload the music.
 
    .. ## pygame.mixer.music.stop ##
 


illume (approved):
Looks good to me :) I'd like to leave the issue open for more investigation though. I haven't read the issue in depth, but if it's a backwards compatibility thing it might need fixing.

PR #2416, opened 2020-12-23
Finalize WINDOWEVENT API, minor fixes

ankith26:
Finalizing the new API, getting rid of the old one. Some more fixes as well. I’m currently testing event stuff a lot, just to make sure I’m not breaking stuff


ankith26:
This PR can close #2214

src_c/event.c
```diff
@@ -598,12 +594,22 @@ _pg_pgevent_deproxify(Uint32 type)
 #if IS_SDLv2
 static SDL_Event *_pg_last_keydown_event = NULL;
 
+static int
+_pg_translate_windowevent(void *_, SDL_Event *event)
+{
+    if (event->type == SDL_WINDOWEVENT) {
+        event->type = PGE_WINDOWSHOWN + event->window.event - 1;
+        return SDL_EventState(_pg_pgevent_proxify(event->type), SDL_QUERY);
```
illume, line 602: What's going on here?

src_c/event.c
```diff
@@ -1751,20 +1753,75 @@ get_grab(PyObject *self)
...
+    while (1) {
+        _pg_event_pump(1); /* Use our custom pump here */
+        switch (PG_PEEP_EVENT_ALL(event, 1, SDL_GETEVENT)) {
+            case -1:
+                return 0; /* Because this never happens, SDL does it too*/
+            case 1:
+                return 1;
+
+            default:
+                if (timeout >= 0 && SDL_GetTicks() >= finish) {
```
illume, line 1791: I went through this function in detail and it looks ok.

src_c/event.c
```diff
@@ -1751,20 +1753,75 @@ get_grab(PyObject *self)
...
+     * might break. So after every event pump, we translate events from
+     * here */
+    SDL_FilterEvents(_pg_translate_windowevent, NULL);
+#endif
+}
+
+static int
+_pg_event_wait(SDL_Event *event, int timeout)
+{
+    /* Custom re-implementation of SDL_WaitEventTimeout, doing this has
```
illume, line 1774: Can we write here what the advantages are? I guess it mainly uses our pump/peek implementations?

test/event_test.py
```diff
@@ -559,23 +558,15 @@ def test_wait(self):
...
 
         # Test case with timeout and an event in the queue.
-        if SDL1:
-            with self.assertRaises(TypeError):
-                pygame.event.wait(250)
-        else:
-            event = pygame.event.Event(EVENT_TYPES[0], **EVENT_TEST_PARAMS[EVENT_TYPES[0]])
-            pygame.event.post(event)
-            wait_event = pygame.event.wait(250)
+        event = pygame.event.Event(EVENT_TYPES[0], **EVENT_TEST_PARAMS[EVENT_TYPES[0]])
```
illume, line 565: Nice the wait works with SDL1 :)

src_c/event.c
ankith26: That’s where we do SDL_WINDOWEVENT translation into respective events. I figured out the old approach to do that from the event filter has its disadvantages, so we do it here now

src_c/event.c
ankith26: That was mostly copied from SDL, just put a bit of twists into it :)

src_c/event.c
ankith26: Yes. We have our own pump function, that we should be using, going forward. This is to trigger the SDL_WINDOWEVENT translation

test/event_test.py
ankith26: That’s another advantage of doing the custom wait implementation :)

illume (approved):
Thanks :)  I did a bit of manual testing, and it LGTM.

src_c/event.c
nimagoli: عالیست

PR #2420, opened 2020-12-24
Docs build fixes + home for the Korean tutorial

Starbuck5:
Brings `setup.py docs` from 8279 warnings to 3 warnings.

All the rst files for the korean tutorial and english variant were encoded in UTF-16, which apparently is incompatible with Sphinx. Building the docs therefore gave pages and pages of this:
example

I also went after warnings generated by the c_api documentation:
capi_warn_example
I changed around the sphinx directives a couple of elements used, and I removed a couple instances of "\\\*". I'm not sure what that is supposed to mean, so I changed it to just "*" because they were emitting warnings.

The next warning was that everything in `tut/en` and `tut/ko` wasn't included in any toctree, so I fixed that.

In the process of that, I made a page for the korean tutorial to be accessible from on the pygame website. (Google translate is a very helpful tool). Now the tutorials section of the docs front page has a link to a korean tutorial page I made, which in turn links to each part/chapter of the tutorial.

I was also able to fix some warnings in `pygame.rst`.

[skip ci] because it's docs


illume (approved):
Thank you! 

It's satisfying all those red lines building the docs are gone now :)

PR #2423, opened 2020-12-26
sdl2: Controller progression

zoldalma999:
#711

- [ ] example script
- [x] examples/eventlist.py
- [x] Tests
- [ ] CONTROLLER_* consts are long, shorter names might be needed
- [x] Better interface for controller mapping
- [x] API docs
- [ ] migration guide
- [x] type hints

I started the documentation but it probably needs improving, as my English is not the best. It is good for a start though.


test/controller_test.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def test_quit__multiple(self):
+        controller.quit()
+        controller.quit()
+        self.assertFalse(controller.get_init())
+
+    def test_get_init(self):
+        self.assertTrue(controller.get_init())
+
+    def test_set_eventstate(self):
```
illume, line 36: I wonder if this needs to be an interactive test? Or maybe split into a part which requires interaction.

"Please move the controller or press some buttons on it"

illume:
Great work so far :)

examples/eventlist.py
```diff
@@ -129,9 +132,13 @@ def main():
 
     # let's turn on the joysticks just so we can play with em
     for x in range(pg.joystick.get_count()):
-        j = pg.joystick.Joystick(x)
-        j.init()
-        txt = "Enabled joystick: " + j.get_name()
+        if pg._sdl2.controller.is_controller(x):
+            c = pg._sdl2.controller.Controller(x)
+            txt = "Enabled contorller: " + c.name
```
robertpfeiffer, line 137: typo

illume:
@zoldalma999 what do you think about merging this in now? The other things in your task list can be done by whoever/whenever at a later time. What we have in this PR is already a nice improvement already.

zoldalma999:
Its fine by me.

But I have question that I could not solve for quite a bit now: I wanted to add the new features from sdl 2.0.14 to the module but could not find a way to `cdef extern` these new functions without it erroring out on previous sdl versions (like 2.0.12). I know there is an `IF` statement but I cant use stuff like `SDL_VERSION_ATLEAST` since that needs to be imported from sdl as well. So is there a way to do it that I missed? Or is there a better way to do it than the `IF` statements?

illume:
> But I have question that I could not solve for quite a bit now: I wanted to add the new features from sdl 2.0.14 to the module but could not find a way to cdef extern these new functions without it erroring out on previous sdl versions (like 2.0.12). I know there is an IF statement but I cant use stuff like SDL_VERSION_ATLEAST since that needs to be imported from sdl as well. So is there a way to do it that I missed? Or is there a better way to do it than the IF statements?

Ah yes. This is annoying. We eventually came to a solution with some of the audio input code(after lots of research and 3 or so PRs getting it wrong first). Since that wasn't in earlier versions either.

The trick in the end was to do it inside C. See here:
- https://github.com/pygame/pygame/blob/main/src_c/pgcompat.c#L4
- https://github.com/pygame/pygame/blob/main/src_c/pgcompat.h#L183

PR #2425, opened 2020-12-28
Update sse2neon header to make it build alpha blitter on arm.

MyreMylar:
As the title. 

It turned out the header we use to convert from sse2 code (used on intel/amd) to neon code (used on arm) didn't have conversions for one of the new intrinsic functions used in the new pygame specific alpha blitter, as such you couldn't compile pygame with SIMD extensions on arm/raspberry pi.

This PR updates the conversion header to the latest version, and comments out a bit at the top of it that doesn't seem to work on the Pi. After this PR you can build pygame on a raspberry pi with neon extensions enabled using this command:

`sudo python3 setup.py install -enable-arm-neon`

### DIFF
--- a/src_c/include/sse2neon.h
+++ b/src_c/include/sse2neon.h
@@ -18,9 +18,14 @@
 //   Malcolm James MacLeod <malcolm@gulden.com>
 //   Devin Hussey (easyaspi314) <husseydevin@gmail.com>
 //   Sebastian Pop <spop@amazon.com>
+//   Developer Ecosystem Engineering <DeveloperEcosystemEngineering@apple.com>
+//   Danila Kutenin <danilak@google.com>
+//   François Turban (JishinMaster) <francois.turban@gmail.com>
+//   Pei-Hsuan Hung <afcidk@gmail.com>
+//   Yang-Hao Yuan <yanghau@biilabs.io>
 
 /*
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
Nice! 👍 
This surely improves things a lot.
Cheers!

ankith26:
Wouldn’t it be better if those arm optimizations are enabled by default now (unless I’m missing some reason here, correct me)
Maybe a -disable-arm-neon could be put in place to disable this default setting

ankith26:
Anyways, the branch compiles for me locally with -enable-arm-neon (on raspi 3b), and the tests run great too!
So good one 👍 I guess it’s safe to merge this one now :)

galexandreg:
> Wouldn’t it be better if those arm optimizations are enabled by default now (unless I’m missing some reason here, correct me)
> Maybe a -disable-arm-neon could be put in place to disable this default setting

The problem is that neon isn't available on raspberry pi 1. So it's a choice between:
- Slow by default (like now), but it's possible to use the SDL2 blitter (PYGAME_BLEND_ALPHA_SDL2) or to compile with neon
- Neon by default, but it stops working on Raspberry pi 1
- Some kind of runtime detection or separate package (#1723)

PR #2426, opened 2020-12-28
Add blend_mode, color tuples and origin=None to _sdl2.Image

galexandreg:
I was playing with the _sdl2 stuff and found some things I thought could be improved. It's the first time I look at cython and I only quickly looked at the pygame.Color stuff so there is probably better ways to do this, but I thought I could try anyway. 🤔 

- **Add blend mode to Image**
There are colors and angles in Image that are simply applied to the texture before drawing. I simply added a blend_mode that works the same as the others.

- **Allow color tuples**
Texture colors can be pygame.Colors, (r,g,b) or (r, g, b, a) (even if the a is ignored) but Images only accept pygame.Color objects. I made a simple change that allows it to work like the Texture colors.

- **Allow None for Image.origin**
It's now possible to set None (this is the default too) as the origin for an Image. This will make the origin at the center, even if the image is scaled with a dstrect. This is how it works in the SDL.

- **Texture default color array size**
I wondered why the defaultColor array size was different in Image and Texture when both use it the same way: `self.color = pgColor_NewLength(defaultColor, 3)`
Since _color_new_internal_length does this:
```
    color->data[0] = rgba[0];
    color->data[1] = rgba[1];
    color->data[2] = rgba[2];
    color->data[3] = rgba[3];
    color->len = length;
```
I guess it should be 4 ints, otherwise the rgba[3] value is read outside the array. Probably not a big deal since the alpha value is not used in Texture and Image, but it made sense to me.

Thanks 😄

### DIFF
--- a/src_c/_sdl2/video.c
+++ b/src_c/_sdl2/video.c
@@ -898,7 +898,7 @@ struct __pyx_opt_args_6pygame_5_sdl2_5video_7Texture_draw {
   int flipY;
 };
 
-/* "pygame/_sdl2/video.pxd":412
+/* "pygame/_sdl2/video.pxd":413
  *     cdef public Rect srcrect
  * 
  *     cpdef void draw(self, srcrect=*, dstrect=*)             # <<<<<<<<<<<<<<
@@ -971,7 +971,8 @@ struct __pyx_obj_6pygame_5_sdl2_5video_Image {
   struct __pyx_vtabstruct_6pygame_5_sdl2_5video_Image *__pyx_vtab;
   pgColorObject *_color;
   float angle;
-  float origin[2];
+  SDL_Point _origin;
+  SDL_Point *_originptr;
   int flipX;
   int flipY;
   float alpha;
@@ -999,7 +1000,7 @@ struct __pyx_obj_6pygame_5_sdl2_5video___pyx_scope_struct__get_drivers {
 
 
 
-/* "pygame/_sdl2/video.pyx":912
+/* "pygame/_sdl2/video.pyx":922
  * 
  * 
  * cdef class Renderer:             # <<<<<<<<<<<<<<
@@ -1711,9 +1712,6 @@ static CYTHON_INLINE Uint32 __Pyx_PyInt_As_Uint32(PyObject *);
 static CYTHON_INLINE SDL_BlendMode __Pyx_PyInt_As_SDL_BlendMode(PyObject *);
 
 /* CIntFromPy.proto */
-static CYTHON_INLINE size_t __Pyx_PyInt_As_size_t(PyObject *);
-
-/* CIntFromPy.proto */
 static CYTHON_INLINE long __Pyx_PyInt_As_long(PyObject *);
 
 /* FastTypeChecks.proto */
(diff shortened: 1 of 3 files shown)


galexandreg:
Do I need to upload the .c files for the cython stuff? 🤔

mcpalmer1980:
I have been aware of this issue for some time as well. I converted an entire 20k line game and am working on a higher level library for _sdl2.video and have been planning to submit issues for the inconsistencies I've discovered.

Not sure if allowing 4-tuples is necessary since Images have their own alpha value. But fixing the missing color-tuple support is important.

For your information, these are this issues I have discovered:
- image.color should accept Tuple
- image.origin should allow None (sdl2 will default to center if origin is None
- image.__init__() should accept srcrect for convenience
- image.draw() should accept arguments like Texture.draw() that effect the draw() call but not the internal attributes
  - srcrect
  - dstrect
  - angle
  - origin
  - color

galexandreg:
> Not sure if allowing 4-tuples is necessary since Images have their own alpha value. But fixing the missing color-tuple support is important.

 It's not, but Textures already work like that, and since pygame.Colors already have 4 values (rgba), it's was more simple this way... without causing any problems, I think... 🤔 

---

I don't know if you wanted to do a pull request for the other stuff, but I suggest that you create an Issue if you want to talk more about them. 😄

>     * image.draw() should accept arguments like Texture.draw() that effect the draw() call but not the internal attributes
>       
>       * srcrect
>       * dstrect
>       * angle
>       * origin
>       * color

Aren't srcrect and dstrect already in there?
As for the others...The idea of the Image is to have a shared Texture with some additional attributes. You can always create another one with the same Texture.
But if you think it can be useful... why not 😄 

>     * image.origin should allow None (sdl2 will default to center if origin is None)
It's already centered by default and you can change it if you want. I guess it could follow the SDL more closely and accept `None`, but I don't have a preference.
>     * image.**init**() should accept srcrect for convenience
I'm not sure i understand...
`def __init__(self, textureOrImage, srcrect=None):`
It already does, that sort of the point of Image 🤔

mcpalmer1980:
Yeah the srcrect and dstrect in that list were sort of a copy/paste error.

I don't believe origin believes as you say. Whenever I scale an object it rotates off center unless I set the origin, and once you set an origin, you can't set it to None.

I never realized you could send Image a srcrect because it errors out when I use a keyword argument. This is still a bug I think since other methods use the srcrect keyword argument. I've been doing this:
```py
img = Image(texture)
img.srcrect = rect
```

galexandreg:
> I don't believe origin believes as you say. Whenever I scale an object it rotates off center unless I set the origin, and once you set an origin, you can't set it to None.

Ah yes... it makes sense when scaling the image... When you don't provide a center in the sdl, it does the division from the dstrect on every RenderCopyEx
```
if (center) {
        real_center = *center;
    } else {
        real_center.x = real_dstrect.w / 2.0f;
        real_center.y = real_dstrect.h / 2.0f;
    }
```
> I never realized you could send Image a srcrect because it errors out when I use a keyword argument. This is still a bug I think since other methods use the srcrect keyword argument. I've been doing this:
> 
> ```python
> img = Image(texture)
> img.srcrect = rect
> ```

I'm still confused about this, because it works fine here... I can do 
`img = Image(texture, rect)` and` img = Image(texture, srcrect=rect)` with rect as a tuple (x, y, h, w) or a pygame.Rect...🤔

galexandreg:
@mcpalmer1980 I was playing with the origin attribute...
Before I try something, since you found the issue, do you want to fix it yourself and make a PR? I don't know if you wanted to fix the issues or simply report them.

mcpalmer1980:
I considered working on this a few weeks ago but decided against it after getting my feet wet in cython only made a mess.  I might get a good grip on it and work it eventually, but with my other projects and lack of experience with cython I doubt it will be any time soon. If you want to look into I'd be much obliged.

galexandreg:
> I considered working on this a few weeks ago but decided against it after getting my feet wet in cython only made a mess. I might get a good grip on it and work it eventually, but with my other projects and lack of experience with cython I doubt it will be any time soon. If you want to look into I'd be much obliged.

This is the first time I do anything with Cython too, so even if I think this is right, it could be totally wrong 😅 
I changed the origin to make it work like it does in the SDL and I think that is all I will put in this PR for now.

Thanks

illume (approved):
👍

illume:
Sorry I didn't get to this sooner. Nice changes :)

PR #2428, opened 2020-12-29
re-init bug fixes

### LINKED ISSUE #2100: pygame.key.set_repeat seems broken on multiple calls to set_mode()
On the newest build of snekoban, holding a key down with set_repeat(500,500) still seems to result in a KEYDOWN event every other frame. Running the game on pygame 1.9.6 produces the expected behaviour.

I fixed this problem in my game by removing the code that inits and de-inits the display, and now it works fine. I don't need to re-init any more, as posting KEYDOWN events is now supported by pygame.

This seems to imply that re-initing breaks key repeat events.

Thanks to @mcpalmer1980 and @jtiai for reporing this bug in snekoban.

### LINKED ISSUE #2421: Re-init breaks a lot of stuff
Whenever we reinit (call pygame.init and pygame.quit) a lot of stuff breaks. 
Just realized that my event-refactor stuff breaks completely after a re-init.
This is an issue robertpfeiffer is also noticing from a long time.

So I am working on a new init-refactor PR that should sort all these issues once and for all.
This PR will come within 4-5 days

### LINKED ISSUE #2612: pg.time.set_timer raises exception without init
Seems like a compatibility bug with pg.time.set_timer with pygame.init being called first.

Before this would work it seems?

From https://github.com/pygame/pygame/pull/2606

```
File "pygame\examples\mask.py", line 175, in main
    pg.time.set_timer(pg.USEREVENT, 33)
pygame.error: pygame is not initialized
```

ankith26:
Re init bug fixes, so init-quit-init works

Closes #2421 
Fixes #2100
Fixes #2612

Sets the event filter at the correct place, so that re-init does not totally break stuff 

And also the init system is reworked a bit, so now the way pygame inits is much more elegant, and controlled.
Basically, pygame.init and pygame.quit directly call the init and quit functions of the respective modules. No more RegisterQuit stuff. Also, we can easily specify the order we want to init stuff, and whether we want to init something as part of something else (like here, we want to call the init and quit functions of event and time modules when we init and quit display module)

A few small cleanups were made to the C-API (see the doc changes).


illume:
Hi,

seems like a much cleaner approach.

How can I reproduce the problems this fixes?
Can you explain why some tests were deleted?

ankith26:
Sure, I should have included my test program at the beginning itself. Here it is, to summerise all the fixes
```python
import pygame
import sys

# Do the dreaded re-init
pygame.init()
pygame.quit()
pygame.init()

win = pygame.display.set_mode((320, 200))
pygame.event.set_blocked([pygame.TEXTINPUT, pygame.MOUSEMOTION])
pygame.key.set_repeat(500, 500)

eventcntr = 0
while True:
    pygame.display.update()
        eventcntr += 1
        print(eventcntr, event)
        if event.type == pygame.QUIT:
            sys.exit()
pygame.quit()
```
We do the re-init, block a few events, set a key-repeat, and read events that appear. 

If you run this code with 2.0.1, you will see that blocked events still appear on the queue (TEXTINPUT and MOUSEMOTION), key-repeat is all messed up, you get KEYDOWN events spammed in the queue, instead of appearing only twice a second (like #2100). Events like ACTIVEEVENT, VIDEORESIZE, VIDEOEXPOSE, MOUSEBUTTONUP/DOWN (for mousewheel) dont appear at all.
The unicode attribute (of KEYDOWN/UP) is broken too, on certain keys only.
And maybe more stuff are broken, that I have not tested yet.

This PR fixes all that.
Also, I removed one test from base_test because that was not relevant anymore. ``pygame.init()`` used to work by calling ``__PYGAMEinit__`` function of **all** modules in sys.modules (yuck). The one unittest I removed was written to take advantage of this ugly/inefficient behaviour. I would also have sped up ``pygame.init()`` by a huge factor

src_c/base.c
```diff
@@ -264,13 +258,86 @@ pg_register_quit(PyObject *self, PyObject *value)
...
-    int success = 0, fail = 0;
+    /* Put all the module names we want to init in this array */
+    const char *modnames[] = {
+        IMPPREFIX "display", /* Display first, this also inits event,time */
+        IMPPREFIX "joystick",
+        IMPPREFIX "font",
+        IMPPREFIX "freetype",
+        IMPPREFIX "mixer",
+#if IS_SDLv2
+        /* IMPPREFIX "_sdl2.controller", Is this required? Comment for now*/
```
ankith26, line 335: I had one question here. Should we init the controller module, when pygame.init is used?
Sinc _sdl2 is an experimental module, it makes sense that it is not implicitly init’d for now
I could change it if required 
Note that the code before this PR would init everything anyway

illume: Nothing in the _sdl2 namespace should be imported/initialized automatically IMHO.

illume:
Thanks for the explanations :)

To manually test re-init I guess it would make sense to test all the examples and some other apps adding the init/quit/init code to the top. Can you think of a way the unit tests could be modified to catch the bugs? A lot of the tests are already doing init/quit/init. So I imagine unit/interactive tests for the features broken by re-init would have caught these specific bugs - if they existed and/or had init/quit/init at the top.

The C-API changes are probably ok. I don't know of any consumers of the pygame 2 C-API. The docs need updating though.

src_c/base.c
ankith26: Nice! So we can leave that line as a comment for now :)

ankith26:
> Can you think of a way the unit tests could be modified to catch the bugs? A lot of the tests are already doing init/quit/init. So I imagine unit/interactive tests for the features broken by re-init would have caught these specific bugs - if they existed and/or had init/quit/init at the top.

Very true.
See this line in the pygame display unit tests
https://github.com/pygame/pygame/blob/d4937b0dd49faaea50c224107c920740b5f90826/test/display_test.py#L415

The code here waits for a WINDOWMINIMIZED event. But an ACTIVEEVENT was also supposed to be generated there, but because of this bug, it would not get generated there. Now that this PR fixes that bug, maybe we can adapt that code too

illume:
This doesn't fix the key repeat issue for me on Ubuntu 20.10. See https://github.com/pygame/pygame/issues/2100#issuecomment-778784758 for a reproducer.

ankith26:
Weird. I just tried this PR on an Ubuntu 20.10 VM (VirtualBox on my windows PC) too, and it works fine, just like it did when I tested it on my Pi3B.

MyreMylar:
Looks like two test failures, one on python 3.6/3.5 on mac & windows:

```
ERROR: test_custom_type__reset (pygame.tests.event_test.EventCustomTypeTest)
Ensure custom events get 'deregistered' by quit().
  File "C:\hostedtoolcache\windows\Python\3.5.4\x64\lib\site-packages\pygame\tests\event_test.py", line 237, in test_custom_type__reset
    before = pygame.event.custom_type()
pygame.error: pygame.event.custom_type made too many event types.

```


and one that looks to be mac & python 2 specific:
```
FAIL: test_register_quit (pygame.tests.base_test.BaseModuleTest)
Ensure that a registered function is called on quit()
  File "/Users/runner/hostedtoolcache/PyPy/2.7.18/x64/site-packages/pygame/tests/base_test.py", line 557, in test_register_quit
    self.assertEqual(quit_count, 1)
AssertionError: 2 != 1

```

That last one at least looks like it might be related to this PR.

ankith26:
The first one was also a result of this PR, and after manually testing with local py3.5 install I have hopefully fixed it

ankith26:
Okay, looks like I fixed the py2 error, the issue was that `Py_XDECREF(PyObject_CallObject(funcobj, NULL));` calls the function twice, the macro uses the variable twice, so we have to do something like `temp =  PyObject_CallObject(funcobj, NULL); Py_XDECREF(temp);`

Starbuck5 (approved):
This PR is difficult to review because of it's size, but I've finally gotten to it.

I can confirm that this fixes #2100 on both Mac and PC

As for the "smoke tests," I put a bunch of `pygame.init()` and `pygame.quit()`s at the beginning of three of my apps, and I did not find any breakages.

illume:
In 5a69e02b4341dc6f799a6ef55fc797cc21cdc625 I tried to fix the merge conflicts (mostly to do with returning bools)... but messed up it seems in display.c

illume:
Fixed the merge mess-ups.

I retested the https://github.com/pygame/pygame/issues/2100#issuecomment-778784758 and it works now. Tried a few other things, and can't find any related issues.

illume (approved):
Massive work! Thanks.

PR #2451, opened 2021-01-14
fix IME placement in SCALED mode

robertpfeiffer:
Fix IME placement in SCALED mode: IME placement with pygame.key.set_text_input_rect was based on physical screen pixels instead of scaled coordinates, now scaling is applied so the IME is in the right place regardless of window size. There is no test case because I see no good, non-interactive way of testing it, but there was a bug before.

### DIFF
--- a/src_c/key.c
+++ b/src_c/key.c
@@ -815,16 +815,20 @@ key_set_text_input_rect(PyObject *self, PyObject *obj)
         return RAISE(PyExc_TypeError, "Invalid rect argument");
 
     if (sdlRenderer!=NULL){
-        SDL_Rect vprect;
+        SDL_Rect vprect, rect2;
+	/* new rect so we do not overwrite the input rect */
 	float scalex, scaley;
 
 	SDL_RenderGetScale(sdlRenderer, &scalex, &scaley);
 	SDL_RenderGetViewport(sdlRenderer, &vprect);
 
-	rect->x = (int)(rect->x * scalex + vprect.x);
-	rect->y = (int)(rect->y * scaley + vprect.y);
-	rect->w = (int)(rect->w * scalex);
-	rect->h = (int)(rect->h * scaley);
+	rect2.x = (int)(rect->x * scalex + vprect.x);
+	rect2.y = (int)(rect->y * scaley + vprect.y);
+	rect2.w = (int)(rect->w * scalex);
+	rect2.h = (int)(rect->h * scaley);
+
+	SDL_SetTextInputRect(&rect2);
+	Py_RETURN_NONE;
     }
 
     SDL_SetTextInputRect(rect);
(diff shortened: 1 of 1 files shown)


robertpfeiffer:
今日は

robertpfeiffer:
There is no way to test this in Python because it is totally transparent to the user. The python code does not see where the IME pops up.

ankith26 (approved):

src_c/key.c
```diff
@@ -804,13 +804,35 @@ key_set_text_input_rect(PyObject *self, PyObject *obj)
...
         Py_RETURN_NONE;
     }
     rect = pgRect_FromObject(obj, &temp);
     if (!rect)
         return RAISE(PyExc_TypeError, "Invalid rect argument");
+
+    if (sdlRenderer!=NULL){
+        SDL_Rect vprect, rect2;
+	/* new rect so we do not overwrite the input rect */
+	float scalex, scaley;
```
ankith26, line 820: This PR is looking good! 👍 
Minor nitpick here, the indentation can be fixed for code cleanliness sake

ankith26:
Oops, looks like I messed up the review comments. I was saying

This PR is looking good 👍 
Maybe the indentation can be fixed for code-cleanliness sake

robertpfeiffer:
What does that tag mean? Needs testing how?

illume:
By that label I mean: it would be good if someone could test this (manually if automated testing is impossible).

robertpfeiffer:
I'll post some screenshots then, before and after on Windows and Linux (X11, ibus, mozc).

robertpfeiffer:
https://cdn.discordapp.com/attachments/772940667231928360/848499918573404200/unscaled.webm

https://cdn.discordapp.com/attachments/772940667231928360/848499940015865856/peek_ime.webm

robertpfeiffer:
AFAICT, this does not pertain to SDL1 anyway.

robertpfeiffer:
Sorry, trying to restart the build

src_c/key.c
```diff
@@ -804,13 +804,35 @@ key_set_text_input_rect(PyObject *self, PyObject *obj)
...
+
     if (obj == Py_None) {
         Py_RETURN_NONE;
     }
     rect = pgRect_FromObject(obj, &temp);
     if (!rect)
         return RAISE(PyExc_TypeError, "Invalid rect argument");
+
+    if (sdlRenderer != NULL) {
+        SDL_Rect vprect, rect2;
```
Starbuck5, line 818: I'm surprised this doesn't crash the python2 compiler...

robertpfeiffer: It's at the start of a block, that should be enough.

Starbuck5 (approved):
I don't have the IME setup to test this locally, but I can approve it.
 
I can see the need for this from the provided videos, and the code seems pretty straightforward.

It even would support renderers created by users, rather than just SCALED created ones.

PR #2456, opened 2021-01-20
docs: explain what TEXTINPUT is

robertpfeiffer:


robertpfeiffer:
Travis is broken, this is a change in the docs only. I fully expect somebody to argue with me because I go into too much detail, but in my opinion it is important to get this right, and to explain what TEXTINPUT is, and why you should enable/disable it.

There are things I left out, for example:
- if you manually start an on-screen keyboard program, this means pygame does not close or minimise the window ever
- the key combinations that do not produce text input are things like CTRL-V or CTRL-A (for obvious reasons) 
- important bugs (IME is broken on Wayland, TEXTINPUT does not work with the voice keyboard on Android)
- some on-screen keyboards produce text but no key presses, others emulate keyboard events
- the way dead keys are handled depends on your OS and settings
- some keys do not generate TEXTINPUT, and some text input does not generate KEYDOWN events, so you can press a special key and pygame will try to DWIM event.unicode and attach the wrong string to a key press. Text input can also appear out-of-order

ankith26 (approved):
I haven’t read too much in-depth, but the idea of explaining more stuff in the docs, sure seems like a very good idea to me. 
Nice one!

galexandreg (approved):
I haven't took the time to test all this as much as you did, but I think the added information makes the docs much easier to understand. It is certainly not too much detail 👍

ankith26:
I did read through properly, now. And I think that things have been explained clearly. It is certainly not excessively detailed, so IMO, this can be merged anytime

robertpfeiffer:
Any glaring formatting errors? If not, I will merge this via command line soon and add my unicode IME example next.

ankith26:
Well this PR looks good to be merged.

But the command line merging likely won’t work (because Travis CI has failed and githubs branch protection rules will not allow the merge)
The only way you can merge it now, is to either wait for illume or close and open this PR (re-run CI) until it passes.

Starbuck5:
I just closed and re opened to re-run on the new CI (even though it's just a documentation change...)

ankith26:
The CI is not running because it is a doc change.
But I had hoped that it should give green for these doc changes, rather than a yellow and an "expected" tag.

illume (approved):
👍 thanks

PR #2457, opened 2021-01-21
Fix set_mods() type hint

w-v:
Fixes the signature type hint of the set_mods function to match the C code and the documentation


ankith26 (approved):
Nice fix!
Indeed the typehints were wrong before, and this PR fixes that.
Thanks 👍

ankith26:
Don’t mind the close/open. It’s just done to run the CI again :)

PR #2460, opened 2021-01-25
Fix leaving fullscreen creates window at (0,0)

### LINKED ISSUE #2360: Pygame 2.0.0 puts window at (0,0) when toggling back from full-screen, hiding the top bar. (worked fine in 1.9.6)
**Environment:**

- **Operating system** Windows 7
- **Python version** 3.7.9
- **SDL version** 2.0.12
- **PyGame version** 1.9.6 & 2.0.0 

**Current behavior:**

When toggling back from fullscreen, the window gets placed at coördinates (0,0), the uttermost top left of the screen. This hides the top bar, making it hard/impossible to grab the window. 

**Expected behavior:**

The 1.9.6 behavior: when toggling between fullscreen and non-fullscreen, the position of the non-fullscreen stays the same. If that is impossible for some reason: at least display the new non-fullscreen window a topbar height below the top, so users can adjust the position themselves. 

**Screenshots**

Before Fullscreen toggle 1.9.6:
before1 9 6
After Fullscreen toggle 1.9.6:
after1 9 6
Before Fullscreen toggle 2.0.0:
before2 0 0
After Fullscreen toggle 2.0.0:
after2 0 0

**Steps to reproduce:**

pip install pygame==1.9.6
run example code, press F twice
pip install pygame==2.0.0
run example code, press F twice

**Test code**

```
import sys
import pygame
from pygame.locals import *
 
class PygameProblem():
    def __init__(self):
        pygame.init()
        
        pgdi=pygame.display.Info()
        self.fullScreenSize=(pgdi.current_w, pgdi.current_h)
        self.nonFullScreenSize=(int(pgdi.current_w/2.0),int(pgdi.current_h/2.0))
        self.isFullScreen=False
        
        self.setDisplay()
        pygame.display.update()
        
        self.mainLoop()        
        
    def setDisplay(self):
        if self.isFullScreen:
            self.windowSurface = pygame.display.set_mode(self.fullScreenSize, pygame.FULLSCREEN ,0)
        else:
            self.windowSurface = pygame.display.set_mode(self.nonFullScreenSize, pygame.RESIZABLE ,0)
        
    def mainLoop(self):
        while True:
                if event.type == QUIT:
                    pygame.quit()
                    sys.exit()
                elif event.type== KEYDOWN:
                    if event.key== K_f: # F toggles fullscreen
                        self.isFullScreen=not self.isFullScreen
                        self.setDisplay()
                        pygame.display.update()
if __name__=='__main__':
    PygameProblem()
```

Starbuck5:
Targeted to fix #2360.

Over the past day I've been investigating issue #2360.
The window created inside `set_mode()` is supposed to get its position from the last window, if it exists.
The problem is that if the last window is a fullscreen window, `SDL_GetWindowPosition()` returns (0, 0).
Perhaps in SDL1 the window position values were saved in fullscreen windows, despite not being "real" - fullscreen windows don't have a position on the screen - they are the screen.

This solution works on Windows 10 and Mac.

Further things:
- Is there a more elegant solution than adding special backup ints to be got/set in random circumstances?
- Does this work on ubuntu? Other linux distros?

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1218,6 +1218,13 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                 }
             }
 
+            // SDL doesn't preserve window position in fullscreen mode
+            // However, windows coming out of fullscreen need these to go back into the correct position
+            if (sdl_flags & (SDL_WINDOW_FULLSCREEN | SDL_WINDOW_FULLSCREEN_DESKTOP)) {
+                state->fullscreen_backup_x = x;
+                state->fullscreen_backup_y = y;
+            }
+
             if (!win) {
                 /*open window*/
                 win = SDL_CreateWindow(title, x, y, w_1, h_1, sdl_flags);
@@ -1247,13 +1254,6 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                         win, sdl_flags & (SDL_WINDOW_FULLSCREEN |
                                           SDL_WINDOW_FULLSCREEN_DESKTOP))) {
                     return RAISE(pgExc_SDLError, SDL_GetError());
-                }
-
-                // SDL doesn't preserve window position in fullscreen mode
-                // However, windows coming out of fullscreen need these to go back into the correct position
-                if (sdl_flags & (SDL_WINDOW_FULLSCREEN | SDL_WINDOW_FULLSCREEN_DESKTOP)) {
-                    state->fullscreen_backup_x = x;
-                    state->fullscreen_backup_y = y;
                 }
 
                 assert(surface);
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -65,6 +65,8 @@ typedef struct _display_state_s {
     Uint8 scaled_gl;
     int scaled_gl_w;
     int scaled_gl_h;
+    int fullscreen_backup_x;
+    int fullscreen_backup_y;
     SDL_bool auto_resize;
 } _DisplayState;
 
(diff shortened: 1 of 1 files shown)


src_c/display.c
```diff
@@ -1208,6 +1218,13 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                 }
             }
 
+            // SDL doesn't preserve window position in fullscreen mode
+            // However, windows coming out of fullscreen need these to go back into the correct position
+            if (sdl_flags & (SDL_WINDOW_FULLSCREEN | SDL_WINDOW_FULLSCREEN_DESKTOP)) {
```
robertpfeiffer, line 1223: What about switching from fullscreen to fullscreen?

Starbuck5: It seems to preserve the information through multiple fullscreen set_mode calls in a row. (If that's what you're asking).

However I just realized if you start in fullscreen mode and exit, you will still go to the top corner, which is annoying. I thought I had that working, maybe the MAC changes broke that.

robertpfeiffer:
> Does this work on ubuntu? Other linux distros?

Depending on the window manager, it already worked (but only when you compile from source and link against system SDL), or it can't be made to work (when you have a non-compliant window manager). I'll test it on Ubuntu's flavour of Gnome Shell under X and Wayland and report back...

Starbuck5:
I think everything is fixed now.

I had a lot of trouble getting it to work until I moved the "Set_WindowPosition" under the "SDL_SetWindowFullscreen" - which makes sense. The whole cause of this trouble was fullscreen windows not holding locations, so windows coming out of fullscreen need to be repositioned after they leave fullscreen.

illume (approved):
:+1: nice one :)

illume:
Looks like the crash in Travis is unrelated, and seems to be crashing in lots of other PRs.

illume:
> Does this work on ubuntu? Other linux distros?

Works for me on the gorilla Ubuntu with default settings, with pygame compiled from source.

CodeGuru12:
Not sure what changed from Pygame 2.0.0 to 2.0.1, but there's still an issue with this on windows 10. My version is:  pygame 2.0.1, using SDL 2.0.14, python 3.9.4 with windows 10.

Issue #1126, opened 2019-06-30
The gfxdraw functions do not support mapped colors


charlesej:
The gfxdraw functions do not support surface mapped colors as stated in the [documentation](https://www.pygame.org/docs/ref/gfxdraw.html).
> The color argument can also be an integer pixel value that is already mapped to the Surface's pixel format.

The following code gives an example of a mapped color value being rejected by a gfxdraw function.
```python
# File: gfxdraw_color.py
import pygame
import pygame.gfxdraw

pygame.display.init()
screen = pygame.display.set_mode((150, 150))
RED = pygame.Color('red')
rect = pygame.Rect((0, 0), (40, 40))

for color in ((255, 0, 0), (255, 0, 0, 255), RED, screen.map_rgb(RED)):
    print('Drawing using color:', type(color), color)
    pygame.draw.rect(screen, color, rect, 1)
    pygame.gfxdraw.rectangle(screen, rect, color)

pygame.display.quit()
```

The above code gives the following output. It shows that a mapped color value is accepted by the `pygame.draw.rect()` function, but causes the `pygame.gfxdraw.rectangle()` to raise a `TypeError`.
```
pygame 2.0.0.dev2 (SDL 2.0.9, python 3.7.2)
Hello from the pygame community. https://www.pygame.org/contribute.html
Drawing using color: <class 'tuple'> (255, 0, 0)
Drawing using color: <class 'tuple'> (255, 0, 0, 255)
Drawing using color: <class 'pygame.Color'> (255, 0, 0, 255)
Drawing using color: <class 'int'> 16711680
Traceback (most recent call last):
  File "gfxdraw_color.py", line 13, in <module>
    pygame.gfxdraw.rectangle(screen, rect, color)
TypeError: invalid color argument
```

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit)
- pygame: 2.0.0.dev2 (SDL: 2.0.9) at aba2da716c9bd265a6e1a110b2f38604a9d0ec66

JLkp:
I realised this is a very old issue but for anyone looking for this. 
The `gfxdraw.rectangle` function only supports a `Color` or `tuple (int, int, int, [int]))` as the color value. In order to make your code work you need to add the `unmap_rgb()` function which convert the mapped integer value back to a color tuple. Your code would look like this:  

```python
import pygame
import pygame.gfxdraw

pygame.display.init()
screen = pygame.display.set_mode((150, 150))
[...]

PR #2462, opened 2021-01-28
Add str and RGBA tuple types to freetype/font type hints

MyreMylar:
As per the title. Likely these were missed at some point when the font modules were updated.

Anyway it looks like these two modules use the same `pg_RGBAFromFuzzyColorObj()` function as the draw module and it has support for string colour types.

I suspect at least some, if not all of the other five modules (mask, display, gfxdraw, pixelarray & transform) that take a `_ColorValue` type hint could also have their C code updated to use this same fuzzy loader function so that the `_ColorValue` type could be consistent across all modules rather than the current split with some accepting the string colour types and some not.

### DIFF
--- a/buildconfig/pygame-stubs/font.pyi
+++ b/buildconfig/pygame-stubs/font.pyi
@@ -10,7 +10,7 @@ from pygame.color import Color
 from pygame.surface import Surface
 
 _ColorValue = Union[
-    Color, Tuple[int, int, int], List[int], int, Tuple[int, int, int, int]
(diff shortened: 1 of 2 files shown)


zoldalma999:
> I suspect at least some, if not all of the other five modules (mask, display, gfxdraw, pixelarray & transform) that take a _ColorValue type hint could also have their C code updated to use this same fuzzy loader function

That would also fix issue #1126

illume (approved):
Thanks 🎉

PR #2467, opened 2021-02-05
clamp HSV saturation, fixes #2465

### LINKED ISSUE #2465: Color.hsva can return invalid values
**Environment:**
- **Python version**: 3.8.3
- **SDL version**: 1.2.16, 2.0.14
- **PyGame version**: 1.9.6, 2.0.1

**Description**
With at least one rgb value (128, 196, 0), pygame.Color.hsva returns a value (100.00000000000001) that is outside of the accepted [0, 100] range for saturation. The docs do note that the hsva values might be slightly different than expected due to rounding errors, but Color.hsva should still provide a valid value.

It is possible this bug also effects the hue/value of hsva, and it is also possible that it effects the other color conversions (cmy, hsla, i1i2i3).

One use case of Color.hsva would be to change and wrap the hue value, leaving the other values the same. If a value returned by Color.hsva cannot be set back into Color.hsva, this use case can't work reliably without additional rounding.

**Test code**
```
import pygame

pygame.init() # Optional

color = pygame.Color(128, 196, 0)

print(color.hsva)

color.hsva = color.hsva
```

**Output**
```
pygame 2.0.1 (SDL 2.0.14, Python 3.8.3)
(80.81632653061226, 100.00000000000001, 76.86274509803923, 100.0)
  File ".\test_color.py", line 9, in <module>
    color.hsva = color.hsva
ValueError: invalid HSVA value
```

robertpfeiffer:
I can't test this, on my FPU it already worked...

### DIFF
--- a/src_c/color.c
+++ b/src_c/color.c
@@ -1122,6 +1122,13 @@ _color_get_hsva(pgColorObject *color, void *closure)
     /* Calculate S */
     hsv[1] = 100. * (maxv - minv) / maxv;
 
+    /* Clamp S, needed on some but not all FPUs */
+    if (hsv[1] < 0) {
+        hsv[1] = 0.f;
+    } else if (hsv[1] > 100) {
+        hsv[1] = 100.f;
+    }
+
     /* Calculate H */
     if (maxv == frgb[0]) {
         hsv[0] = fmod((60 * ((frgb[1] - frgb[2]) / diff)), 360.f);


Starbuck5 (approved):
This fixes the bug locally for me.

Should we just clamp every output value so that it fixes any other similar bugs?

robertpfeiffer:
> Should we just clamp every output value so that it fixes any other similar bugs?

I think we had a quick chat about this already on Discord. Most values don't need to be clamped. Hue values don't need to be clamped, but wrapped around. Many values are calculated in a way that makes this kind of floating point error impossible.

I have more changes committed locally, but I am not sure they are needed. Prime candidates for bugs are HSLA and i1i2i3.

Another way to go would be to be more permissive in the setter functions.

Starbuck5:
I did some testing and I found that saturation is the only value that needs to be clamped on HSV.
I ran my computer through all ~16 million rgb combinations and all of the failures were a slightly high saturation value.
There were 12625 failures.

I also found this at the top of color.c:
```
/* The follow bug was reported for the pygame.math module:
 *
 * Adjust gcc 4.4 optimization for floating point on x86-32 PCs running Linux.
 * This addresses bug 52:
 * http://pygame.motherhamster.org/bugzilla/show_bug.cgi?id=52
 *
 * Apparently, the same problem plagues pygame.color, as it failed the
 * test_hsva__all_elements_within_limits and
 * test_hsva__sanity_testing_converted_should_not_raise test cases due
 * to slight, on the order of 10e-14, discrepencies in calculated double
 * values.
 */
```

The relevant tests don't through every rgb value though, so they missed this one.


I'll repeat my brute force testing with the other color formats and get back to you on those.

illume (approved):
👍 thanks

PR #2471, opened 2021-02-06
Fix rotate

### LINKED ISSUE #2464: Segfault in pygame.transform.rotate
**Environment:**
- **Python version**: 3.8.3
- **SDL version**: 1.2.16, 2.0.14
- **PyGame version**: 1.9.6, 2.0.1

**Description**
When attempting to rotate a surface which has a size of 0 in either dimension, Pygame throws a segmentation fault.

**Test code**
```
import pygame

pygame.init() # Optional

surface = pygame.Surface((0, 0))

surface = pygame.transform.rotate(surface, 45)

print("Success!") # This never runs
```

**Stack trace**
```
pygame 2.0.1 (SDL 2.0.14, Python 3.8.3)
Fatal Python error: (pygame parachute) Segmentation Fault

Current thread 0x000003b0 (most recent call first):
  File ".\test_rotation.py", line 7 in <module>
```

MightyJosip:
This should fix #2464.

### DIFF
--- a/docs/reST/ref/transform.rst
+++ b/docs/reST/ref/transform.rst
@@ -60,9 +60,6 @@ Instead, always begin with the original image and scale to the desired size.)
    transparent. Otherwise pygame will pick a color that matches the Surface
    colorkey or the topleft pixel value.
 
-   If the width or the height of the surface is 0, this function won't do
-   anything
-
    .. ## pygame.transform.rotate ##
 
 .. function:: rotozoom
--- a/docs/reST/ref/transform.rst
+++ b/docs/reST/ref/transform.rst
@@ -60,6 +60,9 @@ Instead, always begin with the original image and scale to the desired size.)
    transparent. Otherwise pygame will pick a color that matches the Surface
    colorkey or the topleft pixel value.
 
+   If the width or the height of the surface is 0, this function won't do
+   anything
+
    .. ## pygame.transform.rotate ##
 
 .. function:: rotozoom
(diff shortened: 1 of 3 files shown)


Starbuck5:
It still fails for me locally, when I try to use the returned surface for anything.

```
s = pygame.Surface((0,0))
s = pygame.transform.rotate(s, 20)
print(s)
```

This segfaults ^

I have no idea why. Reference count issues? Something about the dimensions being 0?

illume (changes requested):
Seems there's a segfault still with the print of the return value. This code as a unit test would be helpful.

MightyJosip:
Don't know how the optimal solution for this, but discovered this:
image
This code segfaults
image
And this one doesn't (both of this changes are made on the current master branch.
The only difference (it think it is only) is that in one case the function `newsurf_fromsurf` gets run, and in other case it doesn't.
It is completely useless in my code, but if it doesn't run it segfaults.

Starbuck5:
I found another working solution, but this is surprisingly perplexing.

```
    if (surf->w < 1 || surf->h < 1) {
        Py_INCREF(surfobj);
        return surfobj;
    }
```

The Py_INCREF isn't needed if the function is called like this:
```py
surface2 = pygame.transform.rotate(surface, 45)
surface = surface2
print(surface)
```

But it is needed if the function is called like this:
```py
surface = pygame.transform.rotate(surface, 45)
print(surface)
```

ankith26:
Yes, indeed, `Py_INCREF` is required here, and there is no memory leak, in both cases of code you mentioned

ankith26 (approved):
LGTM, looks good now! 👍

Starbuck5 (changes requested):
Ok, I've tested this locally, with sys.getrefcount, and I'm confident it works as expected.

I think the sentence added to the docs is unnecessary, though

And this should probably also have a unit test just to avoid future regressions.

Starbuck5 (approved):
Let's finally get this merged! Thanks 👍

Issue #1663, opened 2020-04-24
Arcs with width values greater than 1 being drawn with tiny holes/gaps


MightyJosip:
Arcs drawn with width values greater than 1 may contain tiny holes/gaps. This is shown by the following code as the resulting arc contains tiny gaps where it should be a solid color.

    import pygame
    import time

    pygame.display.init()
    screen = pygame.display.set_mode((100, 100))
    rect = pygame.Rect((0, 0), (80, 70))
    rect.center = screen.get_rect().center

    pygame.draw.arc(screen, pygame.Color('yellow'), rect, 1, 2, 35)
    pygame.display.flip()

    time.sleep(10)
    pygame.display.quit()

Resulting arc with tiny gaps.

![image](https://user-images.githubusercontent.com/29337110/80218691-cab3d200-8641-11ea-9f4f-0ede7984b472.png)

System details:

    os: windows 10 (64bit)
    python: 3.7.4 (64bit)
    pygame: 2.0.0.dev7 (SDL: 2.0.12)

**Related docs:** https://www.pygame.org/docs/ref/draw.html#pygame.draw.arc

Mushguys:
This looks to me like the resolution of the arc is too low. I can't be sure without seeing the code.
Where is the code for drawing arcs?

MightyJosip:
Another example

```
import pygame
pygame.display.init()
screen = pygame.display.set_mode((200, 200))
rect = pygame.Rect((0, 0), (180, 170))
rect.center = screen.get_rect().center
pygame.draw.arc(screen, pygame.Color('yellow'), rect, 1, 2, 70)
pygame.display.flip()
while True:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            quit()
```

Source code for drawing arc:
front-end: https://github.com/pygame/pygame/blob/master/src_c/draw.c#L498-L574
algorithm: https://github.com/pygame/pygame/blob/master/src_c/draw.c#L1326-L1371

Mushguys:
The problem seems to be here:
[https://github.com/pygame/pygame/blob/master/src_c/draw.c#L556-L560](https://github.com/pygame/pygame/blob/master/src_c/draw.c#L556-L560)
This code draws multiple arcs to make up a single, wider arc. The radii of the arcs are drawn at one-pixel intervals, which is large enough for holes in between the arcs. A smaller interval should fix it, but I'm not sure just how small.

Loupau38:
[...]

PR #2480, opened 2021-02-15
Fix incomplete arcs

mhollands:
Fixes issue where arcs of large radius may not be completely drawn, or may not be drawn at all. Issue occurs due to minimum angular step size in draw loop resulting in draw loop overshooting the end of the arc. My function ensures that the draw loop is completed at least once.

(Before and after shown in YouTube video https://www.youtube.com/watch?v=1PBd3VJrfDw&feature=youtu.be)

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1616,10 +1616,10 @@ draw_arc(SDL_Surface *surf, int x, int y, int radius1, int radius2,
 
     x_last = (int)(x + cos(angle_start) * radius1);
     y_last = (int)(y - sin(angle_start) * radius2);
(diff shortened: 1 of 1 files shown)


mhollands:
Not sure why travis-ci is failing. Seems like it could be caused by something other than my changes?

Starbuck5:
Hello mhollands, thanks for the contribution!

I've also ran into this problem, when I was playing around with draw.arc trying to fix #1663 
Don't worry about travis-ci, I it's just broken.

mhollands:
Good to know - thank you!

Starbuck5 (approved):
Looks good!

illume (approved):
Nice video too! Thanks.

PR #2482, opened 2021-02-15
Correct blits typehint sequence arg to cover blit inputs

### LINKED ISSUE #2481: Surface.blits parameter type annotation is inaccurate
In [surface.pyi (line 51)](https://github.com/pygame/pygame/blob/97668d157beac3d0e04f161cb251db8b3e021c5c/buildconfig/pygame-stubs/surface.pyi#L51), currently has

`sequence: Sequence[Union[Surface, Rect]]`

Shouldn't that be a `Tuple` instead of a `Union`?

`sequence: Sequence[Tuple[Surface, Rect]]`

Because according to `Union._doc` 

> Union type; Union[X, Y] means either X or Y.

But each item in `sequence` should be `X and Y`, not `X or Y`, right?

MyreMylar:
Previously this typehint incorrectly asked for a Rect or a Surface in the sequence argument but each blit requires at least both. It now takes one of four tuples each of which requires a Surface and a Rect and three of which cover the permutations of the optional arguments in order.

fixes #2481

### DIFF
--- a/buildconfig/pygame-stubs/surface.pyi
+++ b/buildconfig/pygame-stubs/surface.pyi
@@ -48,7 +48,12 @@ class Surface(object):
         special_flags: int = ...,
     ) -> Rect: ...
     def blits(
-        self, sequence: Sequence[Union[Surface, Rect]], doreturn: Union[int, bool]
+        self,
+        sequence: Sequence[Union[Tuple[Surface, Rect],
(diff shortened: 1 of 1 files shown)


nthykier (approved):

illume (approved):
Looks good to me too. Thanks.

PR #2490, opened 2021-02-17
Adds return parameter to mixer.set_reserved()

### LINKED ISSUE #2488: mixer.set_reserved() should return the number of reserved channels
`mixer.set_reserved()` calls `Mix_ReserveChannels`, which returns the number of channels reserved.

But this value is not returned by `mixer.set_reserved()`. It would be nice if it would simply return that value.

MyreMylar:
Includes docs plus a new test. This functionality just mirrors the SDL function we are wrapping.

fixes #2488


nthykier (approved):

illume (approved):
👍

PR #2494, opened 2021-02-20
Fix segfault in mixer.pre_init()

### LINKED ISSUE #2493: Fatal Python error: (pygame parachute) when calling pygame.mixer.pre_init with a device name.
**Environment:**
I am on macOS Mojave and installed the latest version of PyGame. I am running the code in the native Mac terminal.

- **Operating system** MacOS Mojave.
- **Python version** 3.7.6

When I run the following code to change the output device with a device name as follows:

```
import pygame
import pygame._sdl2 as sdl2
pygame.init()
num = sdl2.get_num_audio_devices(0)
names = [str(sdl2.get_audio_device_name(i, 0), encoding="utf-8") for i in range(num)]
dev_name = names[0] # device number 
pygame.mixer.pre_init(devicename= dev_name)
```
It returns the following error with the mac error message to quit python as in the screenshot below. I tried this with the string literals such as:  "Sennheiser USB headset" or b'"Sennheiser USB headset" which are values returned from list "names" in above code. All of them return the same error as below.

image

Any help would be much appreciated.

Starbuck5:
Fixes #2493 

I'm not sure why this bug didn't occur on Windows.

This bug happened because the devicename parameter was using the "z#" format string, which goes into (string / NULL) and (length of string). However, it was grabbing the length of the string as an integer, rather than a Py_ssize_t, which is why it segfaulted, since that has been deprecated. (And since we defined PY_SSIZE_T_CLEAN before including Python.h).

I fixed this by changing the format string to just "z" - since the length of the string is irrelevant for our case.

I also changed the format string of mixer.init() devicename to "z" - since setting devicename = None should translate into NULL.

### DIFF
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -575,7 +575,7 @@ init(PyObject *self, PyObject *args, PyObject *keywds)
 
     static char *kwids[] = {"frequency", "size",       "channels",
                             "buffer",    "devicename", "allowedchanges", NULL};
-    if (!PyArg_ParseTupleAndKeywords(args, keywds, "|iiiisi", kwids, &freq,
+    if (!PyArg_ParseTupleAndKeywords(args, keywds, "|iiiizi", kwids, &freq,
                                      &size, &channels, &chunk, &devicename,
                                      &allowedchanges)) {
         return NULL;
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
LGTM 👍

ankith26 (approved):
LGTM!
Nice one :)

ankith26:
Close-Open to get travis re-building

illume (approved):
👍

Issue #1178, opened 2019-07-19
SDL2: scrap module


illume:
seems the scrap module isn't being compiled with SDL2.

Related: https://github.com/pygame/pygame/pull/1175

illume:
It's been compiled now thanks to these PRs:
- https://github.com/pygame/pygame/pull/1290
- https://github.com/pygame/pygame/pull/1284
- https://github.com/pygame/pygame/pull/1175

But there is still work to do.

mcpalmer1980:
Would redefining pygame.SCRAP_TEXT and decoding the byte string be enough?

robertpfeiffer:
I wrote the current SDL2-based scrap only as a stopgap.

In the long term, we should probably expose the SDL2 clipboard handling via _sdl2 and get the old win32/mac/x11 scrap stuff for binary image data working again.

In that case, I would propose deleting most of the existing scrap code base and re-implementing it in pure-ish python and ctypes.

illume:
Removing the critical label, because I don't think it should block a pygame 2 release.

I personally have a couple of apps that use this, so I'd like to get to it eventually... but if anyone else wants to work on it, that'd be fine by me. Because I might not ever get to it with everything else left to do.

PR #2495, opened 2021-02-21
Fixes scrap on Windows.

Starbuck5:
**Why was it broken?**

The way to grab a window HWND changed in SDL2.
Also I don't think the compiler liked that the HWND was named SDL_Window, which is now a type name.

Additionally, the order of the `#includes` at the top meant that PY_SSIZE_T_CLEAN wasn't defined properly before python was included, which meant every call to `scrap.put` was printing a deprecation warning. To fix this, I also changed `scraplen` in `scrap.c` to a `Py_ssize_t` as requested. I don't believe this will affect Python 2.7 compatibility because the PEP for this, 353, was implemented in 2.5.

**Other scrap stuff:**

I couldn't get scrap building on my mac, even for pygame 1.9.6. Whatever I do `pygame.scrap` is `MissingModule`.

### DIFF
--- a/src_c/scrap.c
+++ b/src_c/scrap.c
@@ -37,8 +37,6 @@
 
 #include "doc/scrap_doc.h"
 
-
-
 /**
  * Indicates, whether pygame.scrap was initialized or not.
  */
@@ -66,7 +64,7 @@ static PyObject *
 _scrap_set_mode(PyObject *self, PyObject *args);
 
 /* Determine what type of clipboard we are using */
-#if IS_SDLv2
+#if IS_SDLv2 && !defined(__WIN32__)
 #define SDL2_SCRAP
 #include "scrap_sdl2.c"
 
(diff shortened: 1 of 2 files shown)


robertpfeiffer:
Does this PR actually link the platform-specific scrap module again on windows instead of the sdl2-based clipboard shim?

Starbuck5:
Yes, it does. I changed the “if SDL2” to “if SDL2 and not Windows”

I wasn’t sure if I should though, because this will make it more inconsistent across platforms.

Starbuck5:
I should also mention the main issue tracking scrap, #1178.

robertpfeiffer:
Shouldn't your PR also touch the Setup_Win or Setup_SDL2 files?

Starbuck5:
I don't think the Setup_Win file was changed when you added the SDL2 text compatibility shim. `scrap_win.c` just wasn't being included in `scrap.c`.

It seems like it should, but it built fine on my computer. I don't think the specific platform module being included needs to be explicitly set in one of the Setup files... it doesn't do that on mac either.

illume (approved):
👍 Nice one @Starbuck5

PR #2497, opened 2021-02-22
Add namehint argument to music.load()

### LINKED ISSUE #2407: file-like object mp3 doesn't work but file.mp3 works
System: Ubuntu 20.04
Python: 3.8
pygame: just installed
Problem: playing file-like object from gtts
I'm using this code
```
from gtts import gTTS
import pygame
from io import BytesIO

pygame.init()

def say(text):
    tts = gTTS(text=text, lang='en')
    fp = BytesIO()
    tts.write_to_fp(fp)
    fp.seek(0)
    pygame.mixer.init()
    pygame.mixer.music.load(fp)
    pygame.mixer.music.play()
    while pygame.mixer.music.get_busy():
        pygame.time.Clock().tick(10)

text = "This is an example"
say(text)
```
But there is no sound. If I store the gtts output to file.mp3 and do load file.mp3 the output can be heard. But I don't want to write and read "real" files. What can I do. The error message is:
```
  File "/home/jz/Skripte/TTS/googletts-4.py", line 20, in <module>
    say(text)
  File "/home/jz/Skripte/TTS/googletts-4.py", line 14, in say
    pygame.mixer.music.load(fp)
pygame.error: Module format not recognized
```

Starbuck5:
Fixes #2407  (they would add namehint="mp3" to their program)

This was relatively simple to add, since there is a function `Mix_LoadMUSTypeRw()`, and since the SDL_Mixer code already had a list of file extensions to music type conversions.

### DIFF
--- a/src_c/music.c
+++ b/src_c/music.c
@@ -297,8 +297,10 @@ music_get_endevent(PyObject *self, PyObject *args)
 Mix_MusicType
 _get_type_from_hint(char *namehint)
 {
-    // Adjusts namehint into a mere file extension component
+    Mix_MusicType type = MUS_NONE;
     const char *dot;
+
+    // Adjusts namehint into a mere file extension component
     if (namehint != NULL) {
         dot = strrchr(namehint, '.');
         if (dot != NULL) {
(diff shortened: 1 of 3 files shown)


Starbuck5:
I just realized that I should probably also add this to `mixer.music.queue`.

src_c/music.c
```diff
@@ -323,7 +390,10 @@ music_load(PyObject *self, PyObject *args)
 #if IS_SDLv1
             new_music = Mix_LoadMUS_RW(rw);
 #else  /* IS_SDLv2 */
-            new_music = Mix_LoadMUS_RW(rw, SDL_TRUE);
+            if (namehint)
+                new_music = Mix_LoadMUSType_RW(rw, _get_type_from_hint(namehint), SDL_TRUE);
```
illume, line 394: if _get_type_from_hint returns MUS_NONE, should we still call Mix_LoadMUSType_RW?

illume (approved):
Thanks 🎈

PR #2499, opened 2021-02-23
Get scrap building on Mac

Starbuck5:
This only builds the SDL2 clipboard based version of scrap, because scrap_mac has its own problems, but at least it upgrades `pygame.scrap` from the dreaded `pygame.MissingModule`.

It took me a long time to figure out why building scrap wasn't even being attempted, and in the end it was a one line fix.

I'm not sure why this fixes it, because I'm unfamiliar with these `.in` files, but this does fix it.

With this is in place it will also be possible to get scrap_mac.c up to speed. (Because it actually builds!)

I also added `.DS_Store` to the gitignore as a convenience for anyone committing from a mac.

Again, relevant to #1178

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,9 @@
 /prebuilt-x??
 /prebuilt_downloads
 
+# macOS generated files
+.DS_Store
+
 # Sphinx generated files: makeref.py
 /docs/\.buildinfo
 /docs/objects.inv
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -8,6 +8,9 @@
 /prebuilt-x??
 /prebuilt_downloads
 
+# macOS generated files
+.DS_Store
```
illume, line 12: 🎉

illume (approved):
Nice :) Thanks.

illume:
> I'm not sure why this fixes it, because I'm unfamiliar with these .in files, but this does fix it.

These files are parsed and then used as a list of which files to build. So you adding it on the Darwin file builds it on Darwin again.

PR #2506, opened 2021-03-03
document get_desktop_sizes

robertpfeiffer:
This function has been in PyGame 2.0.0, and does what people want list_modes to do, so it's high time we document it on pygame.org

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -315,6 +315,26 @@ required).
 
    .. ## pygame.display.get_wm_info ##
 
+.. function:: get_desktop_sizes
+
+   | :sl:`Get sizes of active desktops`
+   | :sg:`get_desktop_sizes() -> list`
+
+   This function returns the sizes of the currrently configured
+   virtual desktops as a list of (x, y) tuples of integers.
+
+   The length of the list is not the same as the number of attached monitors,
+   as a desktop can be mirrored across multiple monitors. The desktop sizes
+   do not indicate the maximum monitor resolutions supported by the hardware,
+   but the desktop size configured in the operating system.
+
+   In order to fit windows into the desktop as it is currently configured, and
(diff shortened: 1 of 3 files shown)


docs/reST/ref/display.rst
```diff
@@ -315,6 +315,26 @@ required).
...
+   as a desktop can be mirrored across multiple monitors. The desktop sizes
+   do not indicate the maximum monitor resolutions supported by the hardware,
+   but the desktop size configured in the operating system.
+
+   In order to fit windows into the desktop as it is currently configured, and
+   to respect the resolution configured by the operating system in fullscreen
+   mode, this function *should* be used to replace many use cases of
+   ``pygame.display.list_modes()`` whenever applicable.
+
+   .. versionchanged:: 2.0.0
```
zoldalma999, line 336: I think this should be `versionadded`

zoldalma999:
There is a part in the display docs that look like this:

> Pygame can only have a single display active at any time. Creating a new one with pygame.display.set_mode() will close the previous display. If precise control is needed over the pixel format or display resolutions, use the functions pygame.display.mode_ok(), pygame.display.list_modes(), and pygame.display.Info() to query information about the display.

It should probably be changed to mention `get_desktop_sizes`.

Apart from these it looks good to me!

robertpfeiffer:
Thanks for the feedback. Today, precise control over pixel format and display resolution is almost never the right idea, because we do not use CRTs any more. As for having a single display active at a time, that is also not true any more if you use `pygame._sdl2`.

I'll try to find a diplomatic way to phrase this, but the easiest way to explain all this would be mentioning that SDL1.2 was made in the age of CRTs, before flat screens. With SDL2, you can just assume that every platform supports at least 32bit RGBA.

illume (approved):
👍 🎉 thanks!

PR #2507, opened 2021-03-04
Upgrade to github actions, ditch travis/appveyor

ankith26:
I noticed that Travis CI has been failing on all PRs recently. So I got Github Actions CI working for pygame, it is currently deployed on my fork of pygame. Also, while migrating Mac/Linux CI from travis, I thought why not move Appveyor CI too.

And yeah, I have tried to replicate very closely what the previous CI scripts were doing. Github release uploads, twine uploads to PyPI (this might need a few changes to make it work), SDL1 builds and sdist tests, and also got builds with same configurations.

**OS:** Windows x86, Windows x64, MacOS 10.15, manylinux
**Python Versions:** 2.7, 3.5 - 3.9, pypy-2.7, pypy-3.6, pypy-3.7 (this wasn't there before, I added it in this commit)

**Pros of Github Actions:**
- Much better startup time of jobs, faster build times too (thanks to parallelism of jobs). For instance, appveyor would take about 20 mins on average to do the windows jobs. With Github actions, it took about 8 mins on the latest commit. Travis would take about 40-50 mins of build time on the Mac Build, and a long startup time (which, together would be about 1-2 hours of wait) and with Github Actions it took about 14 mins on the latest build. Manylinux build is a bit faster too, although I haven't yet refactored it to take advantage of parallelism, once I do that it will be even more faster)
- After every build, the dist files are uploaded to github under the `Github Actions` section in Github Web interface. I feel this is really useful, incase we want bug reporters to test out a fixed patch to see if it works for them.

**Cons of Github Actions:**
- Github Actions does not provide a good variety of machines to run, like ARM or PPC.

Overall I think we should move to Github Actions fast, because Travis is about to go down soon.


zoldalma999:
Related to #2300

illume:
Cool.

I think we should do this incrementally, rather than disabling the other ones at this point.

Having a look at the current builds:
- conan mac build is broken (we can't release with homebrew files... I think)
- arm build is broken, because of a segfault
- manylinux build is because of some new docker rate limiting issue (https://travis-ci.org/github/pygame/pygame/jobs/761785948#L799)
- Windows on appveyor is working.

This is what I reckon:
- Leave travis, and appveyor in there.
- disable failing test on travis arm build (but leave ARM build)
- remove manylinux build on travis (it should run on github actions... lets see if the docker issue appears later)
- disable/comment out mac build on travis (might need for later).
- remove SDL1 build on travis (covered by new actions build).
- comment out all the release parts of github actions for now.

With this we should be building on many platforms again(ya for merging in PRs again), but not be able to release. Then we can see how the github actions go, and maybe disable more things. Additionally, the next release can maybe be done with the old appveyor/travis if needed.


What do you reckon?

illume (changes requested):
Left some notes in a comment about doing it incrementally. Let's get this in, and leave some other parts there for now.

ankith26:
Sounds good

robertpfeiffer:
Does the mac build use Apple clang? Which compilers do have have covered now? I assume MSVC (64-bit), clang, and GCC?

ankith26:
Yes, mac uses clang

ankith26:
Latest commit gets Appveyor back, and Travis ARM build back. The rest of travis builds are deleted out. Also removed travis releases, because we have removed all the wheel build stuff on travis. Commented out Github actions releases for now, we can get those in later

illume:
I fixed an ARM/Xenial issue, but arm is still crashing on the freetype_test (I think). https://travis-ci.org/github/pygame/pygame/jobs/762819676#L6620 It would be good to skip that module entirely on arm for now (if tracking down the issue is not possible).

ankith26:
Yeah. I wanted to track down the ARM issue, but since I don't have an actual ARM64 local setup, tracking the specific bug is really hard. Maybe skipping the whole module for now is a good idea.

ankith26:
Very interestingly, the ARM tests seem to run without segfaults here on the latest merge commit. But there are many errors being reported in the tests

buildconfig/ci/appveyor/install.ps1
```diff
@@ -1,46 +1,46 @@
...
-   & 7z x $prebuilt_zip
-}
-
-
-function main () {
-    DownloadPrebuilt
-}
-
-main
+# For downloading windows prebuilt dependencies.
```
illume, line 1: I don't know what happened here... Perhaps they were readded with some different line ending.

ankith26: Probably yes, I think that accidentally happened when I readded those. Thankfully, nothing seems to break with this

ankith26:
I have still not come up with a solution for the ARM64 issues. 
But in the meantime, I have tested some of the wheels that github actions gives, the windows and linux wheels seem to work well, and I haven't tested Mac ones because I don't have a mac.
I think those might need a bit of attention though, because the build log shows a few complaints regarding a version mismatch between 10.15 and 10.14.

Anyways, what I feel is that we need not wait for ARM64 to pass to merge this PR. We can fix those issues in subsequent PRs once GH actions are in place.

.github/workflows/linux.yml
```diff
@@ -0,0 +1,67 @@
+name: ManyLinux
+
+# Run CI only when a release is created, on changes to main branch, or any PR 
+# to main. Do not run CI on any other branch. Also, skip any non-source changes 
+# from running on CI
+on:
```
Pierre-Sassoulas, line 6: I think there is an optimization to be made here. In pylint we cached a step called "prepare base" used by other jobs later, by using a key based on [the requirements, setup.py and setup.cfg](https://github.com/PyCQA/pylint/blob/master/.github/workflows/ci.yaml#L34). Disregard this comment if not applicable here. also it's probably a next step once this MR is merged.

illume: Good idea.

illume:
I continued on this PR here: https://github.com/pygame/pygame/pull/2600 (2600... nice).

Issue #2500, opened 2021-02-25
[pygame_gui] NotImplementedError on Apple M1


BUCKFAE:
### Operating system
- macOS Big Sur (11.2.1) - Apple M1

### Python Version
- Python 3.8.2

### Creating a new venv
I create a new venv using PyCharm (same results if I do this via Terminal)

### Python Code
```python
import pygame
import pygame_font

pygame.init()
```

### Installing dependencies
```
(venv) (base) julianschubert@MBP-von-Julian DependencyCheck % pip install pygame
Collecting pygame
  Downloading pygame-2.0.1.tar.gz (5.5 MB)
     |████████████████████████████████| 5.5 MB 2.7 MB/s 
Using legacy 'setup.py install' for pygame, since package 'wheel' is not installed.
Installing collected packages: pygame
    Running setup.py install for pygame ... done
Successfully installed pygame-2.0.1
(venv) (base) julianschubert@MBP-von-Julian DependencyCheck % pip install pygame_gui
Collecting pygame_gui
  Using cached pygame_gui-0.5.7.tar.gz (1.0 MB)
Requirement already satisfied: pygame>=1.9.5 in ./venv/lib/python3.8/site-packages (from pygame_gui) (2.0.1)
Using legacy 'setup.py install' for pygame-gui, since package 'wheel' is not installed.
Installing collected packages: pygame-gui
    Running setup.py install for pygame-gui ... done
Successfully installed pygame-gui-0.5.7
(venv) (base) julianschubert@MBP-von-Julian DependencyCheck % 
```

### Running the program
```
/Users/julianschubert/PycharmProjects/DependencyCheck/venv/lib/python3.8/site-packages/pygame_gui/core/interfaces/font_dictionary_interface.py:16: RuntimeWarning: use font: cannot import name 'Font' from partially initialized module 'pygame.font' (most likely due to a circular import) (/Users/julianschubert/PycharmProjects/DependencyCheck/venv/lib/python3.8/site-packages/pygame/font.py)
(ImportError: cannot import name 'Font' from partially initialized module 'pygame.font' (most likely due to a circular import) (/Users/julianschubert/PycharmProjects/DependencyCheck/venv/lib/python3.8/site-packages/pygame/font.py))
  bold: bool = False, italic: bool = False) -> pygame.font.Font:
Traceback (most recent call last):
  File "main.py", line 2, in <module>
    import pygame_gui
[...]

Issue #2452, opened 2021-01-16
TODO: M1 Mac Tasks


robertpfeiffer:
- [ ] Add detailed instructions for M1 mac to the wiki here: https://www.pygame.org/wiki/MacCompile
- [x] find at least one contributor who has a shiny new M1 mac and is willing to test/troubleshoot the build system
- [ ] investigate up-to-dateness of SDL2 (and possibly pygame) in macports/fink/homebrew
- [ ] manually upload arm64 binary wheels for mac os 11
- [ ] better error messages when SDL2/sdl2-config is not installed/not found
- [ ] (optional) update build script for building with SDL2 Framework (instead of sdl2-config)
- [ ] maybe warn during build when an older SDL2 version is found
- [ ] investigate code paths and build flags for fast blitting on aarch64 (RPi4 and M1)
- [ ] have CI on mac OS again

related: https://github.com/libsdl-org/SDL/issues/4181

Carlosted:
@robertpfeiffer I do have a new M1 mac and would love to help. I'm probably not the best for this job because I'm very new to programming and don't know a lot about git. If you ever need help testing things you can contact me. Also, english is not my first language and I still go to school so I might not answer quickly.

johnson7788:
> Is there help available for getting pygame to work on M1s as this thread implies? My friend tried installing it, but it couldn't find pygame.mixer (and some other packages), which sounds like the issues that others have had with SDL2.

same issue, Mac M1

```
/Users/admin/git/rl-flappybird/src/flappy.py:156: RuntimeWarning: use mixer: No module named 'pygame.mixer'
(ModuleNotFoundError: No module named 'pygame.mixer')
  SOUNDS["die"] = pygame.mixer.Sound("data/assets/audio/die" + soundExt)
Traceback (most recent call last):
  File "/Users/admin/git/rl-flappybird/src/flappy.py", line 711, in <module>
    main()
  File "/Users/admin/git/rl-flappybird/src/flappy.py", line 156, in main
    SOUNDS["die"] = pygame.mixer.Sound("data/assets/audio/die" + soundExt)
  File "/Users/admin/miniforge3/envs/py38/lib/python3.8/site-packages/pygame/__init__.py", line 59, in __getattr__
    raise NotImplementedError(missing_msg)
NotImplementedError: mixer module not available (ModuleNotFoundError: No module named 'pygame.mixer')
```

devinsuttles:
[...]

PR #2508, opened 2021-03-05
Update ftfont.py to fix circular dependency

nelsonlove:
Moving the pygame.sysfont import to immediately before the SysFont function fixes the circular dependency issue with the font module on my M1 Mac. See https://github.com/pygame/pygame/issues/2500 for details.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ dist
 .vagrant
 *.egg-info/
 *.so
+Setup.bak
--- a/README.rst
+++ b/README.rst
@@ -1,4 +1,4 @@
-Minor fork for M1 Macs. If you're having trouble using, e.g., pygame_font or the extended image library and you installed SDL2 and companion libraries with Homebrew, this might work for you.
+Minor fork for M1 Macs. If you installed SDL2 and companion libraries with Homebrew and Pygame is giving you errors when you use, i.e., pygame_font, this might work for you.
 
 .. image:: https://raw.githubusercontent.com/pygame/pygame/main/docs/pygame_logo.gif
   :width: 200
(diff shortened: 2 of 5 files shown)


illume:
I don't understand this one... is someone able to explain how this fixes what it fixes?

llindstrom (approved):
The pull request has been tested on linux with python 3.9 and python 2.7. It solves the circular import problem when the SDL2_ttf based **pygame.font** extension module was not built, but the **pygame.freetype** module is available.

In this case, the font extension module is replaced with font.py, a copy of ftfont.py, as **pygame.ftfont** is designed as a **pygame.freetype** based backwards compatible alternative to **pygame.font**. The replacement module font.py imports **sysfont**, which in turn imports Font from font.py. If the Font class is not already present in the partially instantiated font.py an ImportError is raised with the potential circular import message.

Starbuck5:
@nelsonlove It looks like you added some commits to this PR unintentionally.

illume:
Yeah, It would be good to just include commits up to 0f86c84 in this PR.

ubertao:
@nelsonlove Homebrew can be configured to install to locations other than `/opt/homebrew`.

Can we use the output of `brew config` to get the homebrew prefix dir instead of hardcode it as `/opt/homebrew`?

andrewdstrain:
This patch fixes the circular dependency with regards to fonts on my M1 MacBook Air. The `pygame.examples.aliens` app works as expected.

In other news, `pygame.tests` still does not pass 100% - but that should probably be put as another issue once this one has been fixed.

andrewdstrain:
This pull request does work... please read the comments I made in issue #2452.

Thank you!

Starbuck5:
I believe the problem this patch solves was solved more simply in #2636 , so I'm going to close this.

lemon-sh:
I know this is ancient, but this fixed the issue on my RPi, so maybe the issue is not really fixed?
```
>>> from pygame.font import Font
pygame 2.1.3.dev8 (SDL 2.26.1, Python 3.9.2)
>>> Font
<class 'pygame.font.Font'>
```

PR #2510, opened 2021-03-06
Vector projection

dr0id:
Add 'project' methods to the Vector2 and Vector3 classes to project a vector onto another.

### DIFF
--- a/buildconfig/makeref.py
+++ b/buildconfig/makeref.py
@@ -14,15 +14,15 @@ c_header_dir = os.path.join('src_c', 'doc')
 
 def Run():
     try:
-        path_to_sphinx_build = os.path.join(os.path.dirname(sys.executable), "Scripts", "sphinx-build")
-        print("executing sphinx-build:", path_to_sphinx_build)
-        return subprocess.run([path_to_sphinx_build,
+        subprocess_args = [sys.executable, '-m', 'sphinx',
                             '-b', 'html',
                             '-d', rst_doctree_dir,
                             '-D', 'headers_dest=%s' % (c_header_dir,),
                             '-D', 'headers_mkdirs=0',
                             rst_source_dir,
-                            html_dir,]).returncode
+                            html_dir,]
+        print("executing sphinx in subprocess with args:", subprocess_args)
+        return subprocess.run(subprocess_args).returncode
     except:
         print('---')
         print('Have you installed sphinx?')
--- a/buildconfig/makeref.py
+++ b/buildconfig/makeref.py
@@ -7,9 +7,8 @@ import subprocess
 
 rst_dir = 'docs'
 rst_source_dir = os.path.join(rst_dir, 'reST')
-rst_build_dir = rst_dir
+rst_build_dir = os.path.join('docs', 'generated')
 rst_doctree_dir = os.path.join(rst_build_dir, 'doctrees')
-html_dir = 'docs'
 c_header_dir = os.path.join('src_c', 'doc')
 
 def Run():
@@ -20,7 +19,7 @@ def Run():
                             '-D', 'headers_dest=%s' % (c_header_dir,),
                             '-D', 'headers_mkdirs=0',
                             rst_source_dir,
-                            html_dir,]
+                            rst_build_dir,]
         print("executing sphinx in subprocess with args:", subprocess_args)
         return subprocess.run(subprocess_args).returncode
     except:
--- a/buildconfig/makeref.py
+++ b/buildconfig/makeref.py
@@ -14,7 +14,9 @@ c_header_dir = os.path.join('src_c', 'doc')
 
 def Run():
     try:
-        return subprocess.run(['sphinx-build',
+        path_to_sphinx_build = os.path.join(os.path.dirname(sys.executable), "Scripts", "sphinx-build")
+        print("executing sphinx-build:", path_to_sphinx_build)
+        return subprocess.run([path_to_sphinx_build,
                             '-b', 'html',
                             '-d', rst_doctree_dir,
                             '-D', 'headers_dest=%s' % (c_header_dir,),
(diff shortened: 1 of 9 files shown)


dr0id:
Not sure what went wrong with the travis build. Can someone help me figure out how to pass the checks? Thanks.

Starbuck5:
Don't worry about the travis build, travis is just broken.

See #2507.

test/math_test.py
```diff
@@ -2140,6 +2140,102 @@ def test_swizzle_set_oob(self):
         with self.assertRaises(AttributeError):
             v.xz = (1, 1)
 
+    def test_project_v2_onto_x_axis(self):
```
dr0id, line 2143: move the Vector2 tests to the Vector2 test class

test/math_test.py
```diff
@@ -2140,6 +2140,102 @@ def test_swizzle_set_oob(self):
...
+        other = Vector2(3, 5)
+
+        # act
+        actual = v.project(other)
+
+        # assert
+        expected = v.dot(other) / other.dot(other) * other
+        self.assertEqual(expected.x, actual.x)
+        self.assertEqual(expected.y, actual.y)
+
```
dr0id, line 2182: add tests for error cases

illume:
There's a precision error on the pypy 2.7 Travis build.
```
FAIL: test_project_v3_onto_other (pygame.tests.math_test.Vector3TypeTest)
Project onto other vector.
  File "/opt/python/pp27-pypy_73/site-packages/pygame/tests/math_test.py", line 2237, in test_project_v3_onto_other
    self.assertEqual(expected.y, actual.y)
AssertionError: 2.951807228915663 != 2.9518072289156625
```

buildconfig/makeref.py
```diff
@@ -14,7 +14,9 @@
 
 def Run():
     try:
-        return subprocess.run(['sphinx-build',
+        path_to_sphinx_build = os.path.join(os.path.dirname(sys.executable), "Scripts", "sphinx-build")
```
illume, line 17: This isn't x-platform. 

Maybe it's worth making this happy by default on windows too. It could be changed to first look in the Scripts folder, and if there use that - otherwise use "sphinx-build" as before. Or can sphinx-build be called as a python module with -m?

illume (changes requested):
Cool. This is much requested.

I made a note about sphinx-build (maybe for another PR?), and about the precision issue on pypy27.

The docs could maybe be improved with more of a description, and perhaps a note of some reasons why it's useful.

buildconfig/makeref.py
dr0id: Thanks for the comment. I was not sure if this would work on linux and similar platforms. The problem that I encountered was that this subprocess used a sphinx-build from another python installation but not the one I war running setup.py

I could try to check if exists and fallback to the way as it was before, but it I would prefer a more elegant/portable solution. I'll look into the -m options of sphinx.

dr0id:
I overlooked the precision error. Not sure what to do about it. Any hints as what could be a possible solution?


I'll try to document it better.

buildconfig/makeref.py
dr0id: @illume updated to use the module directly

src_c/doc/math_doc.h
```diff
@@ -359,4 +369,4 @@ pygame.math.disable_swizzling
  disable_swizzling() -> None
 globally disables swizzling for vectors.
 
-*/
```
dr0id, line 362: not sure what happened here.... wrong line ending?

src_c/math.c
```diff
@@ -1574,6 +1577,44 @@ vector_str(pgVector *self)
...
+    if (b_dot_b < self->epsilon) {
+        PyErr_SetString(PyExc_ValueError,
+                        "Cannot project onto a vector with zero length");
+        return NULL;
+    }
+
+    factor = a_dot_b / b_dot_b;
+
+    for (i = 0; i < self->dim; ++i)
+        ret->coords[i] = other_coords[i] * factor;
```
dr0id, line 1612: with the tests in place now maybe this could be optimized for speed. But I will leave it as is for now.

illume:
The precision issues are still there: https://travis-ci.org/github/pygame/pygame/jobs/762156458#L13110

Either the algorithm can be improved to have the same results on this platform(maybe impossible/very hard)... or the tests can be changed to check against an epsilon. See [assertAlmostEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual) and see the math_test.py module for how it's used.

illume:
@Starbuck5 what do you think about the doc generation changes in here?

Starbuck5 (approved):
These docs changes make the whole process a lot more tidy! I like it.

It's easy to delete output files and regenerate the entire thing from scratch now, and the gitignore and docs file structure look way better now.

This touches different bits of the docs system than my sphinx3 commit, so I don't foresee any merge conflicts with that.

The biggest effect of this will probably be on you, @illume. If you have scripts to automatically upload generated html to the website you'll have to change the filepaths that goes on, of course.

I am curious @dr0id, why did you add an extra underscore to links?

dr0id:
> I am curious @dr0id, why did you add an extra underscore to links?

There was a warning about links with the same name. And I read somewhere that using double underscores makes the warning go away by using anonymous links. See [this issue](https://github.com/sphinx-doc/sphinx/issues/3921)

buildconfig/makeref.py
illume: Thanks @dr0id

illume (approved):
👍 🎉 🎈 Thanks.

illume:
> The biggest effect of this will probably be on you, @illume. If you have scripts to automatically upload generated html to the website you'll have to change the filepaths that goes on, of course.

Yeah, I'll have to get to that. Hopefully it doesn't affect too many others. The cleanup is worth it I reckon... it's bugged me for yeeeears :)

Note:
- note, the Appveyor failure was because of an unrelated intermittent compilation failure with the MSVC compiler to do with the buffer code I think
- the static analyzer running on the python3.8 builds on appveyor show some unrelated-to-this-change warnings https://ci.appveyor.com/project/pygame/pygame/build/job/65c525vdyvddgipd?fullLog=true#L2545

illume:
The manylinux build failed because of some docker rate limit thing, so waiting for that to finished before merging.

PR #2511, opened 2021-03-08
Hide old modules and warn users (Overlay, cdrom)

### LINKED ISSUE #1187: SDL2: pygame.cdrom package was not found
When I was trying to use pygame 2.0.0.dev3 on the latest Raspbian Buster I got the following exception:
```
>>> import pygame.cdrom
pygame 2.0.0.dev3 (SDL 2.0.9, python 3.7.3)
ModuleNotFoundError: No module named 'pygame.cdrom'

```
It would be very helpful to get that package back.

**Related Docs:** https://www.pygame.org/docs/ref/cdrom.html

Starbuck5:
This commit does two things:

1) Gets rid of links to the cdrom and Overlay docs so people won't stumble on to them in the present day and think pygame has that capability
2) Inserts a warning message at the top of the module docs that pygame 2.0+ doesn't support it, except for when compiled with SDL1, and that pygame isn't going to support those modules in the future.

This allows power users on older versions to still access the documentation online (by linking directly), but prevents new users from falling into holes where the docs are totally wrong.

### DIFF
--- a/docs/reST/index.rst
+++ b/docs/reST/index.rst
@@ -94,9 +94,6 @@ Reference
 :doc:`ref/bufferproxy`
   An array protocol view of surface pixels
 
-:doc:`ref/cdrom`
-  How to access and control the CD audio devices.
-
 :doc:`ref/color`
   Color representation.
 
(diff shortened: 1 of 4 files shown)


illume (approved):
Good idea.

illume:
I don't think this should close #1187 though. Instead we should leave that open and make it clear that we'd accept PRs but no one is currently planning on working on it.

PR #2512, opened 2021-03-08
inverse filtering in event.get, closes #2325

### LINKED ISSUE #2325: Inverse filtering in pygame.event.get()

robertpfeiffer:
For https://github.com/pygame/pygame/issues/2325


illume (changes requested):
Seems like a useful feature. 

Would it be possible to add a unit test? If not, we can add a task item into the issue instead.

MightyJosip (changes requested):
Hi, can you also add the typehints for the new argument here in this file: https://github.com/pygame/pygame/blob/main/buildconfig/pygame-stubs/event.pyi

MightyJosip (approved):

src_c/event.c
```diff
@@ -1960,6 +1960,112 @@ _pg_event_append_to_list(PyObject *list, SDL_Event *event)
...
+
+        do {
+            ret = PG_PEEP_EVENT(&event, 1, SDL_GETEVENT, type);
+            if (ret < 0) {
+                PyErr_SetString(pgExc_SDLError, SDL_GetError());
+                goto error;
+            }
+            else if (ret > 0) {
+                if (filtered_index==filtered_events_len) {
+                    SDL_Event *new_filtered_events = malloc(sizeof(SDL_Event)*filtered_events_len*4);
```
illume, line 2001: Sorry, I'm a bit daft... Why is this *4 here?

test/event_test.py
```diff
@@ -405,6 +405,23 @@ def test_get_type(self):
...
+        self.assertEqual(queue[0].type, pygame.USEREVENT)
+
+        pygame.event.post(pygame.event.Event(pygame.KEYUP))
+        pygame.event.post(pygame.event.Event(pygame.USEREVENT))
+        queue = pygame.event.get(exclude=(pygame.KEYDOWN, pygame.KEYUP))
+        self.assertEqual(len(queue), 1)
+        self.assertEqual(queue[0].type, pygame.USEREVENT)
+
+        queue = pygame.event.get()
+        self.assertEqual(len(queue), 2)
```
illume, line 423: Can we please have a test for this part from the documentation?
> If  an ``exclude`` parameter is passed, the ``eventtype`` parameter *must* be None.

Testing some other error conditions could be useful.

docs/reST/ref/event.rst
```diff
@@ -218,17 +218,24 @@ Most these window events do not have any attributes, except ``WINDOWMOVED``,
...
    | :sg:`get(eventtype=None) -> Eventlist`
    | :sg:`get(eventtype=None, pump=True) -> Eventlist`
+   | :sg:`get(eventtype=None, pump=True, exclude=None) -> Eventlist`
 
    This will get all the messages and remove them from the queue. If a type or
    sequence of types is given only those messages will be removed from the
-   queue.
+   queue and returned.
+
+   If a type or sequence of types is passed in the ``exclude`` argument
```
illume, line 227: > The optional ``exclude`` parameter lets you get events excluding the ones you specify - leaving the excluded events on the event queue.

illume:
Looking good. I'd be happy to merge, but made some questions and notes.

- All the tests all passed.
- The static analyzer on the python 3.8 builds on appveyor has no new warnings (but seems some have crept in previously).
https://ci.appveyor.com/project/pygame/pygame/builds/38246038/job/6tvuvi5r1061nrps?fullLog=true#L2191
- It was a bit hard for me to track how the malloc and free were working in there. In the time I had, I don't think I understood it entirely. If it could avoid malloc/free in there I'd be happier. But I couldn't spot any errors in the C code. It would be nice if someone else could review the C code. @MightyJosip did you already review the C code?
- Would this feature be useful in any of the examples? Adding it to an example will show people why it's a useful function, and also provide a second way of testing the feature. Additionally, more people can find out about the feature if it's in an example. This could come later of course.

illume (approved):
👍 🎉

src_c/event.c
robertpfeiffer: Growing the allocated array. Could also be filtered_events_len*2 or filtered_events_len*3, or even filtered_events_len+1.

test/event_test.py
robertpfeiffer: Sure, we can test this.

Theoretically, we can even relax this requirement in a later release (at the expense of code complexity), but in practice, I see no reason to have a type in both the eventtype and the exclude parameter.

robertpfeiffer:
> It was a bit hard for me to track how the malloc and free were working in there. In the time I had, I don't think I understood it entirely. If it could avoid malloc/free in there I'd be happier. But I couldn't spot any errors in the C code. It would be nice if someone else could review the C code. MightyJosip did you already review the C code?

Should I use realloc instead of malloc+free?

robertpfeiffer:
> Would this feature be useful in any of the examples? Adding it to an example will show people why it's a useful function, and also provide a second way of testing the feature. Additionally, more people can find out about the feature if it's in an example. This could come later of course.

I made this because I wanted to work on a touchscreen example and found the code to cumbersome to put into an example, with all the pumping and pump=False. As it stands, get() with a type and get() with no types are a kind of race condition.

Also there is an old bug report that turned out to be not a bug, but a race condition in the "proof of concept" caused by event.get(RESIZE) followed by event.get() sometimes returning RESIZE events from the second function call.

illume:
> Should I use realloc instead of malloc+free?

It would be best to not do any allocations in there in normal runtime use if possible. So perhaps allocating a big chunk on the first go that can be used.

robertpfeiffer:
> It would be best to not do any allocations in there in normal runtime use if possible. So perhaps allocating a big chunk on the first go that can be used.

Ah, I see. I made all this effort making a "proper" implementation, but just allocating a fixed size and raising an error would have been much easier. I considered this out of laziness, but I chose not to do it because I couldn't decide on the "right" number of events ;)

PR #2513, opened 2021-03-09
Move to sphinx 3

### LINKED ISSUE #1298: Documentation build issues
Here are some issues that exist (once #1278 is resolved) when using Sphinx v2.2.0 (the latest version) to build the pygame docs.

- [ ] Bulleted lists are not being formatted correctly. The text appears on the line after the bullet indicator.
- [ ] A warning is displayed in the doc build output.
    ```
    running docs
    Running Sphinx v2.2.0
    WARNING: html_static_path entry '_static' is placed inside outdir
    loading pickled environment... done
    building [mo]: targets for 0 po files that are out of date
    building [html]: targets for 0 source files that are out of date
    updating environment: 0 added, 0 changed, 0 removed
    looking for now-outdated files... none found
    no targets are out of date.
    build succeeded, 1 warning.
    
    The HTML pages are in docs.
    ```

System details:
- os: windows 10 (64bit)
- python: 3.7.4 (64bit)
- pygame: 2.0.0.dev3 (SDL: 2.0.10) at c256992573554c137a9770d7d8c3e099172bc8c6

Starbuck5:
Since the beginning of time, pygame has been stuck on sphinx 1.x. But no more!

I've managed to fix **almost** every issue with upgrading to the latest versions of sphinx.

Issue 1) Bulleted lists were super wrong
list_gone_wrong
Fixed by getting rid of `list-style-position: inside;` for unordered lists in the css.

Issue 2) Spacing gone wonky
weird_spacing
Fixed by changing `paragraph` docutils nodes to `Text` and `line` nodes. Newer docutils versions (from newer sphinx versions) put more `<p>` tags in the `paragraph` nodes, leading to the spacing problem.

Issue 3)
comparison
I wasn't able to fix this. Left is before, using sphinx 1.8: each table header has a ":" and is inline with its contents. In the right picture, using sphinx 3.5, it doesn't do that, and I'm at a loss to why. I think it is in the css, and the generated html changed, making stylings that previously applied to it not apply anymore.

Why the effort?
Besides the general fact that being up to date on libraries is good, this also allows us to use fancy features added in sphinx 2 and 3. Like many of the features listed around [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#directive-py-method).

Other random fixes I did:
Update copyright to 2021
Removed duplicate `dt.module` css.

Final things:
The docs no longer build on sphinx 1.8.x, these changes break it. It does work on sphinx 2.x however.
The output is not 100% percent the same, but the error is small enough that I think it's acceptable. But if this doesn't get merged, at least it can guide someone else to fixing these problems in the future.

Also fixes #1298!

### DIFF
--- a/docs/reST/conf.py
+++ b/docs/reST/conf.py
@@ -41,7 +41,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'pygame'
-copyright = u'2000-2020, pygame developers'
+copyright = u'2000-2021, pygame developers'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
--- a/docs/reST/ext/boilerplate.py
+++ b/docs/reST/ext/boilerplate.py
@@ -14,7 +14,7 @@ from sphinx.domains.python import PyClasslike
 from docutils.nodes import (section, literal, reference, paragraph, title,
                             document, Text, TextElement, inline,
                             table, tgroup, colspec, tbody, row, entry,
-                            whitespace_normalize_name, SkipNode)
+                            whitespace_normalize_name, SkipNode, line)
 import os
 import re
 from collections import deque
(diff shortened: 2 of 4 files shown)


ankith26 (approved):
Nice changes!

illume (approved):
👍 nice one

PR #2514, opened 2021-03-09
Add "touch" attribute to the mouse events (motion, buttonup, buttondown, wheel), fix horizontal scroll

dr0id:
The reason to add this is the comment on the sdl docs:

`the mouse instance id, or SDL_TOUCH_MOUSEID; see Remarks for details`

and in the remarks:

`may be SDL_TOUCH_MOUSEID, for events that were generated by a touch input device, and not a real mouse. You might want to ignore such events, if your application already handles SDL_TouchFingerEvent. `

I also changed the code such that horizontal scroll now works (only available in the MOUSEWHEELEVENT).


ankith26 (approved):
LGTM!
Nice :)

Starbuck5:
Do we expose SDL_TOUCH_ID? I guess it's always `1`?

We could have the same functionality by adding a `touch` attribute and setting that to `True` or `False` depending on whether it's a "touch" mouseevent.

dr0id:
> 
> 
> Do we expose SDL_TOUCH_ID? I guess it's always `1`?
> 
> We could have the same functionality by adding a `touch` attribute and setting that to `True` or `False` depending on whether it's a "touch" mouseevent.

The SDL_TOUCH_ID is not exposed AFAIK. I guess a 'touch' attribute could be introduced, but then what value would instance_id have? (leave it as it is now?)

`event.touch = event.instance_id == SDL_TOUCH_ID;`

The other way would be to filter them completely out since I'm not sure why SDL introduced this `synthetic` event anyway (will have to investigate).

Starbuck5:
I'm thinking that if the only value `instance_id` gives is saying whether it was a touch mouse event, then we should expose it as a boolean attribute like `touch`.

Because doesn't `instance_id` say 0 for all mice?

dr0id:
> Because doesn't `instance_id` say 0 for all mice?

As far as I can see, yes. But it could change in future. 
If its only this, then I agree, it does not make much sense and `touch` attribute would make more sense.

Starbuck5:
Then my 2 cents is that we should do a `touch` boolean attribute.

If SDL changes in the future we can always ad an `instance_id` attribute to provide that functionality.

dr0id:
> 
> 
> Then my 2 cents is that we should do a `touch` boolean attribute.
> 
> If SDL changes in the future we can always ad an `instance_id` attribute to provide that functionality.

Changed to 'touch' attribute as suggested.

Another question: would it make also sense to map the 'timestamp' attribute? (I'm thinking here about detecting double clicks....)

Starbuck5:
Alright, I've built this locally and it works.

It would be great to get a tester with some touch screen capable device, like a Microsoft Surface, to test this too. But the code correctly detects that my mouse is not a touchscreen and it is completely in line with the SDL docs, so it shouldn't block merging.

I do have some notes about the documentation though:
- The two  `versionchanged` sections can be collapsed into one.
- The link to the SDL docs doesn't highlight `SDL_TOUCH_MOUSEID` as intended (the first %29 needs to be %28)
- We shouldn't reference SDL_TouchFingerEvent in the use case, we should reference the pygame way of getting touch events. `pygame.FINGERMOTION`, `pygame.FINGERDOWN`, `pygame.FINGERUP`?

> Another question: would it make also sense to map the 'timestamp' attribute? (I'm thinking here about detecting double clicks....)

It might in come in handy to someone, but I'm not sure how. For double clicks you don't want the time it was emitted, you want the difference between the two events. So a normal program could just have a `last_click_time` variable populated by `pygame.time`. But it can't hurt for us to wrap more SDL Event information...

If you want to add this you could do it in a separate pull request.

dr0id:
@Starbuck5 I have a touch screen and verified that the touch attribute works as expected:

`<Event(1026-MouseButtonUp {'pos': (734, 445), 'button': 1, 'touch': True, 'window': None})>
<Event(1793-FingerUp {'touch_id': 131159, 'finger_id': 19, 'x': 0.716796875, 'y': 0.5794270634651184, 'dx': 0.0, 'dy': 0.0, 'pressure': 0.0})>
<Event(1024-MouseMotion {'pos': (731, 441), 'rel': (-3, -4), 'buttons': (0, 0, 0), 'touch': False, 'window': None})>
`

Starbuck5 (approved):
Alright, looks good!

Starbuck5:
@dr0id, I've just realized that this fails when built on SDL1, which we still support.
I said not to worry about the Travis build, but in this case it has uncovered a real issue, since SDL1 doesn't have `SDL_TOUCH_MOUSEID`.

I suppose the simplest solution is to wrap every change for the `touch` attribute in `#ifdef SDL2` stuff, but you can see what other SDL2 specific event things have done.

dr0id:
@Starbuck5 I have a touch screen and verified that the touch attribute works as expected:

`<Event(1026-MouseButtonUp {'pos': (734, 445), 'button': 1, 'touch': True, 'window': None})>
<Event(1793-FingerUp {'touch_id': 131159, 'finger_id': 19, 'x': 0.716796875, 'y': 0.5794270634651184, 'dx': 0.0, 'dy': 0.0, 'pressure': 0.0})>
<Event(1024-MouseMotion {'pos': (731, 441), 'rel': (-3, -4), 'buttons': (0, 0, 0), 'touch': False, 'window': None})>
`

> I've just realized that this fails when built on SDL1, which we still support.

How or where did you see this? I guess you are right, but I'm unable to interpret the travis output correctly.
I will try to find a solution for this.

Starbuck5:
The first Travis build is on SDL 1.2, and scrolling around near the bottom shows the source of the compiler error.

Your new commit will probably fix it.

illume:
The SDL1 build is indeed passing again.

illume (approved):
Looks good to me, thanks.

👍 🎉

illume:
If you want to use the `touch` attribute somewhere in an existing example/ that would be a nice improvement.

PR #2519, opened 2021-03-11
Update snippet in demo page: ChimpLineByLine.rst

deek80:
super minor: use `==` for int comparison in the example (I see this was already fixed in the source)


illume (approved):
Thanks :)

PR #2524, opened 2021-03-14
Fix typo in display.set_gamma() that causes it to fail.

Starbuck5:
Bug reported [on Stackoverflow](https://stackoverflow.com/questions/63326377/set-gamma-with-pygame-2-0-x).

Relevant Docs: [pygame.display.set_gamma](https://www.pygame.org/docs/ref/display.html#pygame.display.set_gamma)

Unlike the SO user, on my end, `display.set_gamma()` always returned `False`. But this does fix it for me.


illume (approved):
Looks correct to me (g and b can't use the same offset). Thanks.

(I didn't test it by comparing it to pygame 1.9.x though)

PR #2525, opened 2021-03-14
Guard against MUS_OPUS not being defined on Xenial era SDL

illume:
Unfortunately I didn't notice before merging, that there was a change that seems to break builds that don't include opus support.
https://travis-ci.org/github/pygame/pygame/builds/761959165#L1279

```c
src_c/music.c:326:16: error: ‘MUS_OPUS’ undeclared (first use in this function); did you mean ‘MUS_MP3’?
         type = MUS_OPUS;
                ^~~~~~~~
                MUS_MP3
```

Related: https://github.com/pygame/pygame/pull/2497

### DIFF
--- a/src_c/music.c
+++ b/src_c/music.c
@@ -322,9 +322,11 @@ _get_type_from_hint(char *namehint)
     else if (SDL_strcasecmp(namehint, "OGG") == 0) {
         type = MUS_OGG;
     }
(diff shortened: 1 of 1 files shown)


Starbuck5:
Thanks illume!

It's weird that it's needed, since I grabbed that code directly from the SDL_Mixer source.

PR #2520, opened 2021-03-11
Add draw_blend_mode to _sdl2 renderers

galexandreg:
Small pr to add renderer draw blend modes as requested by @queueseven in Issue #2518.
I thought it could be useful too...

Thanks!

PR #2526, opened 2021-03-15
Add SDL_ComposeCustomBlendMode

queueseven:
pygame#2522 

Adds `SDL_ComposeCustomBlendMode` as a static method `compose_custom_blend_mode` to the `Renderer` class.

TBH I have no idea if the C files should be included and why, since they generated by Cython.


Starbuck5:
Thanks for the contribution!

> TBH I have no idea if the C files should be included and why, since they generated by Cython.

It's just because we don't have the tooling set up to generate them automatically on build or whenever. Although it means that this might have a merge conflict with #2520, because you both add things to video.pyx which changes the generated files.

> Since pygame is sort of at a higher level of abstraction than the SDL itself, i'm not sure how I feel about having different supported features on different platforms, and I don't know how much the team want the new sdl2 things to be compatible with the old stuff in the end...
But I don't see why it shouldn't be available somewhere anyway for people who want it.

I agree with @galexandreg on this.

I think it's a good thing for us to include this, but it will be challenging to document for that reason. Also it is definitely a function for more advanced use cases.

Did you have a specific application for this or did you just feel like it should be added? (Just curious).

PR #2527, opened 2021-03-16
Fix pylance autocomplete

zoldalma999:
Before this pr pylance autocomplete looked like this:
image

As seen from the image it didnt support submodules. The first suggestion should be `pygame.display`.

After the pr the autocomplete looks like this for me:
image

Before merging I would ask to test if it does not mess up any other tools (though shouldnt).

### DIFF
--- a/buildconfig/pygame-stubs/__init__.pyi
+++ b/buildconfig/pygame-stubs/__init__.pyi
@@ -69,3 +69,29 @@ def encode_file_path(
 ) -> bytes: ...
 def register_quit(callable: Callable) -> None: ...
 def __getattr__(name) -> Any: ...  # don't error on missing stubs
+
+surface = pygame.surface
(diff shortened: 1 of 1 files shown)


MightyJosip (commented):
Ok it looks good, it doesn't break anything and you say it helps with VSCode.
Just a small note. Because of this change when you write `display = pygame.display` than for the PyCharm that display is treated as the variable:
image
and right now it is treated as a module:
image
This also has another consequence. With your changes you don't get error for the code like this:
image
And in current version you do (which is correct):
image
This is just a few notes to consider before merging it

zoldalma999:
Also related to microsoft/pylance-release#758

jakebailey:
I'd probably suggest taking #2537 instead; this assignment is not the proper way to re-export modules.

Issue #2515, opened 2021-03-10
Mask should take Vector2s


Starbuck5:
**Description**

From the docs:

> The methods overlap(), overlap_area(), overlap_mask(), draw(), erase(), and convolve() use an offset parameter to indicate the offset of another mask's top left corner from the calling mask's top left corner. The calling mask's top left corner is considered to be the origin (0, 0). Offsets are a tuple or list of 2 integer values (x_offset, y_offset). Positive and negative offset values are supported.

A convenient way of dealing with locations and points in pygame is the `math.Vector2` class. However, since mask operation `offset` can only be integers, and Vectors use floats, it raises a TypeError.

`TypeError: integer argument expected, got float`

It would be cool if `pygame.mask` worked with Vector2s, which would be accomplished by letting the `offset` variable be two floats, which would just be rounded or truncated internally.

Edit:
What if the `pg_TwoIntsFromObj` macro was updated to take sequences of two floats and round them? That would add support for float sequences (and therefore vectors) throughout a lot of pygame code.
Edit Edit:
`pg_TwoIntsFromObj` already supports floats, its just that the mask code doesn't use that macro much.

<details>
  <summary>Potential spots to implement `pg_TwoIntsFromObj` macro gathered using grep</summary>
 
```
src_c/display.c:    if (!PyArg_ParseTupleAndKeywords(arg, kwds, "|(ii)iiii", keywords, &w, &h,
src_c/display.c:    if (!PyArg_ParseTupleAndKeywords(args, kwds, "(ii)|iii", keywords,
src_c/display.c:    if (!PyArg_ParseTupleAndKeywords(arg, kwds, "|(ii)iii", keywords, &w, &h,
src_c/display.c:    if (!PyArg_ParseTupleAndKeywords(args, kwds, "(ii)|iii", keywords, &w, &h,
src_c/image.c:    if (!PyArg_ParseTuple(arg, "O!(ii)s|i", &Bytes_Type, &string, &w, &h,
src_c/image.c:    if (!PyArg_ParseTuple(arg, "O(ii)s|i", &buffer, &w, &h, &format))
src_c/mask.c:    if (!PyArg_ParseTuple(args, "(ii)", &x, &y))
src_c/mask.c:    if (!PyArg_ParseTuple(args, "(ii)|i", &x, &y, &value))
src_c/mask.c:    if (!PyArg_ParseTuple(args, "O!(ii)", &pgMask_Type, &maskobj, &x, &y))
[...]

PR #2530, opened 2021-03-19
Add keyword arg and Vector2 support to Mask module

Starbuck5:
As I noticed in #2515, the Mask module threw a fit whenever it got a `pygame.Vector2` argument, so my goal was to use the `pg_TwoIntsFromObj` macro to fix this.

I also decided to add support for keyword arguments while I was at it, as per #808.

I added keyword arguments and vector2 testing to `mask_test.py`, not through new cases, but by sprinkling them throughout. This is because `mask_test.py` is already 6400 lines long; it's scary enough already. (Also I was lazy). I also ran `mask_test.py` through the `black` code formatter, which introduced some other changes throughout.

I added two version changed notifications at the top of the Mask docs, to announce these two changes. I thought it would be repetitive to announce the change in every function. (Although the draw docs do this). I also removed several `type` param modifiers in the docs, because now the sequences of two numbers can be input in so many formats I thought it would be unhelpful to list them. The terminology I've used in the TypeError error messages is "a sequence of two numbers."

This also updates the `mask.pyi` type stubs to reflect the change, using the preexisting `_Coordinate = Union[Tuple[float, float], List[float], Vector2]`. I tried to test these new definitions, but I couldn't get mypy to fail at all, even when I gave it bogus like a Vector2 made of strings or a mask offset of a single number.


Starbuck5:
I think I've fixed everything that was my fault...

When this is ready for merge, feel free to squash merge.

illume:
Cool. Keyword arguments used in the tests really do make the code much more readable!

(I'm a bit afraid of that strange commit at the beginning of this PR.)

> I tried to test these new definitions, but I couldn't get mypy to fail at all, even when I gave it bogus like a Vector2 made of strings or a mask offset of a single number.

This reminds me, I wonder how or if we could start adding some sort of tests for the type definitions?

test/mask_test.py
```diff
@@ -2034,7 +2031,7 @@ def test_convolve__with_output(self):
         o.clear()
         test.clear()
 
-        m.convolve(k, o, (10, 10))
+        m.convolve(othermask=k, outputmask=o, offset=Vector2(10, 10))
```
illume, line 2034: I wonder if these should be `other, output`?

illume (approved):
I reviewed everything, and it looks good to me :) I'm a bit afraid of that commit at the beginning... (I could squash merge it, and I hope that'd fix it).

There was one naming question, but I think your choice is good.

test/mask_test.py
Starbuck5: `other` and `output` are good. Having the type name in isn't necessary. I'll fix that up.

Starbuck5:
Thanks.

I think I'm going to continue this in another PR, just like my other change, to get rid of the first commit and to squash all the other commits together.

PR #2532, opened 2021-03-21
Start docs for SDL2 video

Starbuck5:
This adds the `_sdl2.video` module to the documentation.

Each method or attribute is just a stub, without the usual implementation notes and guides for users. It just has a direct reference -- what you can access and what you can pass in / get out.

It uses the method I started in #2511 to hide the module in the top bar. This would only be accessible through direct link. The module also has a warning at the top clarifying that the module is in development and that the API will change.

It doesn't get everything defined in `video.pyx`, but it completely defines the 4 big video classes, so I think it's a good start. (I think the messagebox system should be documented separately).

I also included the auto generated doc.h file, despite it not being used by anything right now. I didn't want to put in a system to block generating those.


illume:
> I also included the auto generated doc.h file, despite it not being used by anything right now. I didn't want to put in a system to block generating those.

If we do go ahead with a video.c implementation eventually, this could be used for that.

illume:
> Merge pull request #1 from pygame/main 

This is a weird commit ^. I saw this on another PR too by someone else... I wonder how it happened.
Probably this PR could be rebased to get rid of it.

illume (changes requested):
Very good. Thanks for doing this!

Could you please rebase to get rid of that suspicious commit talking about PR number 1?

galexandreg:
This is great! I often see people online wondering if pygame supports hardware acceleration... Having docs will make the new features a lot easier to find / try.

I took a quick look at it and there is a few things that I think could be added:

- In Renderer.draw_blend_mode, I feel like `Gets and sets the blend mode used by the drawing functions.`, like it is for draw_color would be better.
- In Image, the origin can be set as None ( and it is the default ). This will put the origin at the center.
- Also in Image, blend_mode is missing.

Thanks again, I think this will be really helpful for people who want to try the new features.🙂

Starbuck5:
> This is a weird commit ^. I saw this on another PR too by someone else... I wonder how it happened.
Probably this PR could be rebased to get rid of it.

I normally update my pygame fork by deleting and re cloning, but this time I tried to merge it in a more "normal" fashion.

> If we do go ahead with a video.c implementation eventually, this could be used for that.

I've actually been working on that. You can see my progress at https://github.com/Starbuck5/pygame/tree/testing2. (At least before I delete my fork again and put the changes somewhere more permanent, like a branch on pygame). That's why I made the docs. I needed them to reimplement the API, and I figured they would be a lot easier to merge too.

> This is great! I often see people online wondering if pygame supports hardware acceleration... Having docs will make the new features a lot easier to find / try.

Eventually... Right now I don't want these docs to be very "user facing," since it's still in development.

Thanks for the notes galexandreg!

> Could you please rebase to get rid of that suspicious commit talking about PR number 1?

I tried... but it looks like that is beyond my git competency at the moment. Since it's only 3 files, I think it'll be easiest to make another PR.

Issue #2380, opened 2020-12-04
display.toggle_fullscreen does not work when toggling a maximized window to fullscreen


QuetzalQatl:
**Environment:**
pygame 2.0.0 (SDL 2.0.12, python 3.7.9)

- **Operating system** windows 7
- **Python version** 3.7.9
- **SDL version** 2.0.12
- **PyGame version** 2.0.0:

**Current behavior:**

When toggling fullscreen from a windowed, maximized window, it tries to do it, fails, gives this warning:
`Warning: re-creating window in toggle_fullscreen` 
then a windowed screen is displayed: that is not re-sizable

**Expected behavior:**

The same as non-maximized windows

**Steps to reproduce:**

Inspect the sample code: set isFullSreen to True if you want to start with a fullscreen (same behaviour)
Toggle fullscreen with F a couple of times: that works just fine. 
then maximize the window, and press F to see the problem arise. 

**Test code**

```
#!/usr/bin/env python3
import os
import pygame

pygame.init()

# find out & store desktopSize
pgdi=pygame.display.Info()
desktopSize=(pgdi.current_w, pgdi.current_h)

FPS = 100
clock = pygame.time.Clock()

screenWidth=160 # initial screen width when windowed
screenHeight=120

hor=0 # ball position on screen 
ver=0

# will make the first intance of the window appear in the middle of the screen.
# (consecutive instances will appear in the upper most left corner)
os.environ['SDL_VIDEO_CENTERED'] = '1' 

# flag to keep track of the question: is the window maximized or not
# should remain False at startup
maximized=False 

# choose to startup fullscreen or not.
# both options will work fine
isFullScreen=True 

# set up initial display:

# always make the non-fullscreen screen first, 
# so the fullscreen has something to toggle back to that is placed properly (as in not at 0,0 )
screen = pygame.display.set_mode((screenWidth,screenHeight), pygame.RESIZABLE) 

if isFullScreen:
    # set size to desktop 
    # not doing this takes much longer (about 4 seconds)
    # but doing this is near instantanious
    pygame.display.set_mode(desktopSize, pygame.RESIZABLE) 
    
    pygame.display.toggle_fullscreen() # now toggle fullscreen


done=False
while not done:
    for event in pygame.event.get():        
        if event.type == pygame.QUIT: 
            # like ctrl-break or the cross in the title bar
            done = True
        elif event.type == pygame.KEYDOWN and event.key == pygame.K_q: 
            # pressing q will also quit the program
            done = True
[...]

PR #1673, opened 2020-04-26
Correct return type (int -> bool) to match documentation

nthykier:
Related discussion: #1667

PR #2534, opened 2021-03-22
Update functions to return bools rather than ints

Starbuck5:
Someone on [stackoverflow](https://stackoverflow.com/questions/66739461/pygame-colliderect-function-returning-0-zeros-and-1-ones-instead-of-true-a) pointed out that some `Rect` functions return `1`  or `0` instead of the `True` or `False` that the documentation says.

I don't think many people noticed, because of how close `1` and `0` are to `True` and `False`, but these changes up our docs consistency a bit.

**Summary of changes:**

I started this by fixing the inconsistency pointed out about Rect, and then I moved on to other modules. I used grep to find all instances of `return PyInt_FromLong(...)`, and replaced them with `PyBool_FromLong` when appropriate.

One change that might be noticed the most is `key.get_pressed()` returning a list of booleans rather than integers.

~~The most extensive changes were to `display.toggle_fullscreen()`. The docs say that it returns an int, but not what that int means. Looking at the SDL1 toggle_fullscreen function, it should return 0/False on failure, 1/True on success. I edited the SDL2 toggle_fullscreen function to match this, rather than returning `-1` on failure as it seemed to be doing. I also edited the `toggle_fullscreen()` docs to say it returns a boolean, True on success and False on failure.~~

While I had the `toggle_fullscreen()` docs right there, I added a notice about #2380.

Fun fact, I did some digging in the commit history, and some of these lines were written before Python had booleans. 😮

Edit: related to #1673


robertpfeiffer:
Yeah, that was undocumented on purpose...

docs/reST/ref/display.rst
```diff
@@ -514,6 +515,10 @@ required).
     * wayland (Linux/Unix)
     * cocoa (OSX/Mac)
 
+   .. Note:: :func:`toggle_fullscreen` doesn't work on Windows
+             unless the window size is in :func:`pygame.display.list_modes()`.
+             See `issue #2380 <https://github.com/pygame/pygame/issues/2380>`_.
+
    .. ## pygame.display.toggle_fullscreen ##
```
robertpfeiffer, line 522: That only goes for non-SCALED modes, right?

Starbuck5: I hadn't tested with SCALED modes before, but you're right. I'll update that.

Starbuck5:
> Yeah, that was undocumented on purpose... I wasn't sure if toggle_fullscreen should perhaps return the resolution or the display surface in the future, and previously it returned nothing.

The SDL1 one looked like it returned 1 on success and 0 on failure, so I changed them both to return True on success and False on failure. @robertpfeiffer is that alright? If it needs more discussion and I can pull it out of this PR.

robertpfeiffer:
It probably needs a *little* more discussion. Some of these cases should issue a warning or a message of some kind, but you know how it is with warnings and backward compatibility...

On some platforms, toggling fullscreen is just not a thing, and if that is the case, this should be handled by the user. Maybe we need a can_toggle_fullscreen() function (for Android and framebuffer and RPi and KMSDRM), but since that would be retrofitted _after_ togglr_fullscreen was a thing, it would not be backward-compatible either.

Starbuck5:
I see.

The previous output, although undocumented, did give 3 distinct pieces of information; `-1`, unsupported; `0` , failure; or `1`, success.

Adding warnings or adding a `can_toggle_fullscreen()` doesn't block this though, so I'm going to keep my change in for now, unless you would like me to take it out.

I could see a `can_toggle_fullscreen()` function being useful. For example, an options screen that only presents a fullscreen / not fullscreen option if that is an actual option.

robertpfeiffer:
> The previous output, although undocumented, did give 3 distinct pieces of information; `-1`, unsupported; `0` , failure; or `1`, success.

I'll be happy if you add comments like
```C
Py_RETURN_FALSE; //unsupported

Py_RETURN_FALSE; //failed
```

Starbuck5:
I realized that it raised an error in lots of cases, and I didn't want to change those all to return `False`, because that isn't very transparent, so this commit is another stab at making everything consistent.

The `versionchanged` comment assumes pygame 1.9.x people are using SDL1 and pygame 2.x people are using SDL2, which is true for almost everyone.

illume (changes requested):
If I'm understanding correctly, there's still some open discussion on the toggle stuff? If so, maybe we could take the toggle stuff out of this PR so it doesn't block getting merged? Feels like the toggle stuff needs more before it's documented?

Starbuck5:
I took all the toggle stuff out, except for the warning about it not working properly on windows.
I think that's important to keep in.

robertpfeiffer:
I'd be happy if this was commented, or if the toggle stuff was kept out for now, or if we printed a warning. Either is fine.

robertpfeiffer:
So for now, all my objections are withdrawn or have been addressed.

Starbuck5:
> So for now, all my objections are withdrawn or have been addressed.

Is that an approving review @robertpfeiffer? 😉

illume (approved):
Thanks :)

illume:
Sounds like robertpfeiffer approved, so I'll merge.

PR #2535, opened 2021-03-22
Implement pygame.display.get_window_handles

szabolcsdombi:
Hi,

First of all Thank You for this great project.
I am writing in favor of the ease of use with external libraries.

I would like to have a platform independent way to return all the handles necessary to create a vulkan surface.
Currenty this is supported by the `get_wm_info`.

I have a working code to render into a pygame window with glnext (a vulkan based renderer).

[pygame_window.py](https://github.com/glnext/glnext/blob/fdf94a4ae92552937de2ef8a8b4ddae8cf6b5887/examples/pygame_window.py)

Unfortunately this does not work out of the box if I change to Linux.
There is a "display" key available but it is wrapped into a PyCapsule and the pointer is retrievable in pure python.

I think it would be awesome to get a tuple with the necessary values on all platforms.
I think on windows it is (HINSTANCE, HWND), with X11 it is (Display, Window), on Mac it may be just a surface pointer I guess.
For the other platforms I assume the (Display, Window) pair is necessary.

I made the PR just to make an example of what sort of function I am talking about. I tried to apply the same code style as above. I did not manage to build it yet.

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -3112,6 +3112,7 @@ static PyMethodDef _pg_display_methods[] = {
     {"Info", pgInfo, METH_NOARGS, DOC_PYGAMEDISPLAYINFO},
     {"get_surface", pg_get_surface, METH_NOARGS, DOC_PYGAMEDISPLAYGETSURFACE},
     {"get_window_size", pg_window_size, METH_NOARGS,
+     DOC_PYGAMEDISPLAYGETWINDOWSIZE},
     {"get_window_handles", pg_get_window_handles, METH_NOARGS,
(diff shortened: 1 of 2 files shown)


robertpfeiffer:
So you would like a the handle on the window from SDL2? I think we used to have something like this.

Or rather, I know I wrote this, but it was an internal API, and then I deleted it again. But maybe you mean something else.

You want a pointer to the the xlib-specifc window data structure?

Starbuck5:
I'm going to close this to try to tidy up the PR list, since this has been stale for a while.

PR #2536, opened 2021-03-23
Add keyword arg and Vector2 support to Mask module

Starbuck5:
Continued from #2530, with the name changes suggested and no suspicious merge no. 1.


illume (approved):
Thanks :) Sorry the othermask part was so much!

PR #2537, opened 2021-03-25
Explicitly export modules from stubs, fix type errors reported by pylance/pyright

jakebailey:
Pylance recently dropped its own stubs for pygame (https://github.com/microsoft/pylance-release/issues/758).

This PR fixes a number of issues with the stubs that ship with pygame (which are generally high-quality otherwise):

- Modules that are intended to be exported from `pygame` directly are now explicitly exported. PEP 484 lays out how to do this `from . import X as X` form to indicate "yes I really mean to export it". This is the main issue I was fixing before it snowballed into fixing the rest. 🙂 
- Generic types used in annotations now include the missing type params. `IO` is the big offender; it likely should really be `SupportsFileNo`, as all the C code does is get the `fileno`.
- All functions, classes, etc, now have fully known types. I looked at some of the C and Python code to do this.
- Statements that don't have any effect in stubs have been removed. I think these were copy-paste leftovers.
- `self` has been added to a few methods that were missing it.
- An invalid TypeVar use to try and capture "bytes on Python 3, str on Python 2" has been switched to a plain `sys.version_info` check that's more accurate and not uncommon in stubs.

I'm not sure what testing procedure exists for these stubs; a number of these issues I think should have been reported by mypy. pyright is a bit stricter.

I'm happy to answer any questions about this PR, or split it up if you believe it's too big. Hopefully it's fine for me to have just sent the code rather than reporting each of these as individual bugs.


buildconfig/pygame-stubs/__init__.pyi
```diff
@@ -1,53 +1,53 @@
...
-import pygame.sprite
-import pygame.transform
-import pygame.bufferproxy
-import pygame.pixelarray
-import pygame.pixelcopy
-import pygame.sndarray
-import pygame.surfarray
-import pygame.math
-import pygame.fastevent
+from pygame import cursors as cursors
```
illume, line 22: Aaaaah, that's how it's done.

buildconfig/pygame-stubs/cursors.pyi
```diff
@@ -46,28 +46,35 @@ def compile(
...
+    List[int],
+    Tuple[int, ...],
+    Tuple[int, ...]
+]: ...
+
+
+class Cursor(Iterable[object]):
     @overload
-    def __init__(constant: int) -> None: ...
+    def __init__(self, constant: int) -> None: ...
```
illume, line 63: 👍

buildconfig/pygame-stubs/mask.pyi
```diff
@@ -40,7 +40,7 @@ class Mask:
     ) -> None: ...
     def copy(self) -> Mask: ...
     def get_size(self) -> Tuple[int, int]: ...
-    def get_rect(self, **kwargs) -> Rect: ...  # Dict type needs to be completed
+    def get_rect(self, **kwargs: Any) -> Rect: ...  # Dict type needs to be completed
```
illume, line 43: I wonder if the comment makes sense now? Has the dict type been completed here?

illume:
Thanks Jake!

> I'm not sure what testing procedure exists for these stubs; a number of these issues I think should have been reported by mypy. pyright is a bit stricter.

So far, the test procedure has been asking MightyJosip to have a look :) I think it's been manually testing them with different type checkers/IDEs to see what happens. We'd love to hear suggestions on how we can automate some of this testing.

ps. We still have a lot of C functions that need to move towards keyword arguments. Also I recall we had some issues with buffers/memory views and how to type them in python (I think there wasn't a way to do that last few times we looked).

buildconfig/pygame-stubs/mask.pyi
jakebailey: From my reading of the C code, this code really can accept anything and is a helper to set basically any property of this class.

Potentially the comment has been removed, but I'm uncertain that this typing can do much better unless someone takes the time to unravel all possible args and then turns them into named arguments. I'm certainly no expert there.

Adding `Any` makes the type not unknown, which makes some type checkers (like pyright) unhappy. We did this to typeshed a bit ago and that's kept clean. Whatever works here, I just narrowed these stubs down to as few issues as possible (the only problems remaining are the presence of default values, but I'd rather not remove them if they are being kept up to date; a new PEP outlines that default values are acceptable in some cases and they are useful for editors).

MightyJosip (approved):
Hi, sry for late reply. Thank you for your contribution. Your fixes look great. I mostly did this code to get the code completion, so it is possible it has some errors, but you fixed a solid amount of them 👍 . About testing sadly I didn't find how could I automate it, so right now I test it manually with pretty big file with a lot of examples of different function calls etc. in PyCharm. Since you know a lot about type hints and have solid amount of experience can you give me a good source where I could read how to set up automated tests for it: for example I would like to make sure `pygame.Rect("a")` gives a warning, and `pygame.Rect((10, 10), (10, 10))` is ok. But anyways back to the topic of my review, this PR looks good, doesn't break anything and fixes some things so I am OK with merging (also just a note you have 2 conflicts, one is because we added a new argument for the event.get function (so you need to add something like `exclude: Optional[_EventTypes]`, and also pygame.Mask.get_at now accepts Vector2 so you need to use `pos: _Coordinate`. But I think it is only because you used somewhat old pygame version when you were doing this. Otherwise 👍

buildconfig/pygame-stubs/mask.pyi
MightyJosip: > the only problems remaining are the presence of default values

This is because we get some extra information because of that default values (more in this comment: https://github.com/pygame/pygame/pull/2050#pullrequestreview-483058627 (3rd point)). But if it is problem for some tools maybe we can consider removing them

jakebailey: Yes, I totally understand, which is why I left them. Usually it's a bad idea because you have to ensure they are properly synced, but I have no problem leaving them and figuring that out later. Like I said, a new PEP about stub conventions may change this up.

jakebailey:
>  Since you know a lot about type hints and have solid amount of experience can you give me a good source where I could read how to set up automated tests for it:

This will be tricky due to the somewhat unique code layout; you have `pygame-stubs` in a random build folder that are copied over, so the tooling has to run there. Most repos just put them in the same location as the other source, but in your case that'd be worse due to that being `src_py`. If that folder were named `pygame`, imports would work correctly, but at the moment the layout implies that the module is called `src_py`. Take a look at pytorch or pandas or numpy for an example (they have each of their folders in the root as the module names).

I'm not sure about mypy, but pyright has some docs: https://github.com/microsoft/pyright/blob/master/docs/ci-integration.md

Pyright (and probably other tooling) will likely need some more configuration due to that aforementioned code layout.

You can also look typeshed (https://github.com/python/typeshed), as I set up pyright there and they have quite a few GHA workflows for type checking. We can take this to another issue to have a longer discussion, if you'd prefer.

> also just a note you have 2 conflicts

No worries, I can fix it.

> But I think it is only because you used somewhat old pygame version when you were doing this

I based it on whatever was in master at the time I sent this, master's changed so I'll merge and fix it up.

jakebailey:
Merged main, and fixed the issues introduced in #2423:

- Don't use `str | None` syntax yet, that's not available in more recent builds of Python and stubs generally use syntax available in Python 3.6.
- `Dict` needs type arguments. I read the `pyx` files, and they are strs.
- `Dict` is usually a code smell as an argument; all the code needs is any `Mapping`, not specifically a `dict` instance.

illume (approved):
👍 🎉

buildconfig/pygame-stubs/event.pyi
```diff
@@ -13,11 +13,11 @@ class Event:
 _EventTypes = Union[SupportsInt, Tuple[SupportsInt, ...], List[SupportsInt]]
```
brccabral, line 13: Can I suggest to add key as a class object?
```
class Event:
    type: int
    key: int
```

Issue #2543, opened 2021-03-30
Segfault in example aliens script if ran in loop


Pycryptor10:
**Environment:**

You can get some of this info from the text that pops up in the console when you run a pygame program.

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): MacOS 11.2.3
- **Python version** (e.g. 3.7.9, 3.8.5) : 3.9.2
- **SDL version** (e.g. SDL 2.0.12): 2.0.14
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.0.1
- **Relevant hardware** (e.g. if reporting a bug about a controller, tell us the brand & name of it): 

**Current behavior:**

After ~5 minutes of playing, python will give
`zsh: segmentation fault  python3 -m pygame.examples.aliens`

**Expected behavior:**

The game does not segfault

**Screenshots**

If applicable, copy and paste screenshots to help explain your problem.

**Steps to reproduce:**

1. Download pygame
2. Apply the change from https://github.com/pygame/pygame/compare/main...Pycryptor10:patch-1.diff
2. run `python3 -m pygame.examples.aliens`
3. Keep playing for ~5 minutes

**Test code**

https://github.com/pygame/pygame/compare/main...Pycryptor10:patch-1.diff

**Stack trace/error output/other error logs**
See files attached to https://bugs.python.org/issue43671 as I thought it could be a python issue first

Pycryptor10:
```
Call to __init__ on line 142 of /Users/cern/vevns/epq2/lib/python3.9/site-packages/pygame/examples/aliens.py from line 384 of /Users/cern/vevns/epq2/lib/python3.9/site-packages/pygame/examples/aliens.py
Call to __init__ on line 112 of /Users/cern/vevns/epq2/lib/python3.9/site-packages/pygame/sprite.py from line 143 of /Users/cern/vevns/epq2/lib/python3.9/site-packages/pygame/examples/aliens.py
Call to add on line 117 of /Users/cern/vevns/epq2/lib/python3.9/site-packages/pygame/sprite.py from line 115 of /Users/cern/vevns/epq2/lib/python3.9/site-packages/pygame/sprite.py
Call to add on line 117 of /Users/cern/vevns/epq2/lib/python3.9/site-packages/pygame/sprite.py from line 133 of /Users/cern/vevns/epq2/lib/python3.9/site-packages/pygame/sprite.py
[...]

PR #2548, opened 2021-04-03
Fix issue #2543 segfault in font dealloc after reinit

REW1L:
TTF_Quit frees all the resources allocated by freetype. However pointers to FT_Face from TTF_Font are not set to NULL after this. Therefore when font_dealloc tries to call TTF_CloseFont, FT_Done_Face is called on already freed object and segmentation fault appears.

It can be so when Font module quits and then GC is called. (test is provided)

Solution provided is a little hack for setting font->face to NULL. It will tell TTF_CloseFont that face has been already freed and all that's left is to free allocated SDL resources.

Also I had to introduce a field current_ttf_generation to the PyFontObject to understand if object passed to the font_dealloc function is being destructed after TTF_Quit.

### DIFF
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -65,6 +65,7 @@ PyFont_New(TTF_Font *);
 #define PyFont_Check(x) ((x)->ob_type == &PyFont_Type)
 
 static int font_initialized = 0;
+static uint current_ttf_generation = 0;
 static const char font_defaultname[] = "freesansbold.ttf";
 static const char pkgdatamodule_name[] = "pygame.pkgdata";
 static const char resourcefunc_name[] = "getResource";
@@ -188,6 +189,7 @@ font_autoquit(void)
     if (font_initialized) {
         font_initialized = 0;
         TTF_Quit();
+        current_ttf_generation++;
     }
 }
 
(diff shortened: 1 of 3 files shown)


REW1L:
This should fix #2543

src_c/font.c
```diff
@@ -762,9 +764,15 @@ static void
 font_dealloc(PyFontObject *self)
 {
     TTF_Font *font = PyFont_AsFont(self);
-
-    if (font && font_initialized)
+    if (font && font_initialized) {
+        if (self->ttf_init_generation != current_ttf_generation) {
+            // Since TTF_Font is a private structure
+            // it's impossible to access face field in a common way.
+            int** face_pp = font;
```
illume, line 771: I wonder if this is enough to stop the problem instead?
```c
TTF_CloseFont(font);
font=NULL;
```

REW1L: Because of tp_free we are able not to aware about font. Therefore setting font to NULL is just redundant. Also font_dealloc is called only once after TTF_Quit to get a segfault inside a function TTF_CloseFont.

illume: Ok, thanks for the explanation :)

Pycryptor10:
@illume: Is there a timeline for this being released? I'm using pygame as part of a project and want to be able to update the supervisor as it's approaching deadlines.

Thanks for the quick patch @REW1L

Starbuck5:
Thanks for looking into this @REW1L, sorry it's been dormant on our end for so long.

Perhaps this should be abstracted into a more abstract "font object invalidation system," since something like this also segfaults:
```py
import pygame
import gc
 
pygame.font.init()
font = pygame.font.Font(None, 20)
pygame.font.quit()
pygame.font.init()

font.render("hello", True, "white")
print("rendered") # never happens, segfaults on the render
```

But I still have some experimenting around this to do, I think. This is a weird situation / bug.

Issue #2551, opened 2021-04-09
Typos in README.rst: SDL_tff -> SDL_ttf


NKID00:
Two typos in [the dependencies section of README.rst](https://github.com/pygame/pygame/blob/main/README.rst#dependencies):

1.  ... The font module relies on **SDL_tff**, which is dependent on freetype. ...
     ... The font module relies on **SDL_ttf**, which is dependent on freetype. ...
2.  **SDL_tff** >= 2.0.11
     **SDL_ttf** >= 2.0.11

See https://www.libsdl.org/projects/SDL_ttf/.

Starbuck5:
Want to have your name in the contributors tab? 😄 

I could put in a pull request for this, but that would be no fun.

PR #2559, opened 2021-04-11
Fix #2550 - Make it easier to find Colour names

### LINKED ISSUE #2550: pygame.colordict is (almost) undocumented
**Environment:**

- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.0.1

**Current behavior:**

I found `pygame.colordict` by snooping in the source code. It doesn't have specific documentation, and is only mentioned once in `pygame.Color`'s docs:

> - color name str: name of the color to use, e.g. 'red' (all the supported name strings can be found in the colordict module)

**Expected behavior:**

All public Pygame bits and pieces show up in the documentation.

**Steps to reproduce:**

Please explain the steps required to duplicate the issue, especially if you are able to provide a sample application.
if the bug is caused by a specific file (image, font, sound, level, please upload it as an attachment

1. Be a Pygame noob, and have no idea what's available.
2. Get very close to implementing `colordict.THECOLORS` yourself because you don't know about it.
3. Be looking at the source code because you're curious about how the docs are generated with Sphinx.

Taffer:
This makes it easier (hopefully) to find `pygame.Color()`'s ability to use named colours, and also provides a page listing each, with a sample colour swatch.

Tested with Sphinx 3.5.3 and Firefox 87.0.

This addresses #2550.

### DIFF
--- a/docs/reST/ref/color.rst
+++ b/docs/reST/ref/color.rst
@@ -11,7 +11,6 @@
    | :sg:`Color(r, g, b) -> Color`
    | :sg:`Color(r, g, b, a=255) -> Color`
    | :sg:`Color(color_value) -> Color`
-   | :sg:`Color(color_name) -> Color`
 
    The ``Color`` class represents ``RGBA`` color values using a value range of
    0 to 255 inclusive. It allows basic arithmetic operations — binary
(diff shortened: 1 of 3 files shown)


Starbuck5 (changes requested):
Cool work on the color swatches page, that's great!

I think it would be best to keep `color.rst` more similar to how it was, with the named color being one of the possible color formats under `Supported color_value formats` like every other format.

That being said, you could replace the link to the colordict source code with the link to your color swatches page.

I think it would also be nice to have a link to the color swatches page in the colordict source code and vice versa, so that any future editors to those files remember the existence of the other. 

Also: the mystery in this PR is why the new page doesn't show up in the pygame documentation top of the page header, because new files are added to that automatically... Something in the format?

**Note to anyone building the documentation (illume)** This page seems to need sphinx 3 and up to date docutils, although I haven't tested it intensively.

Starbuck5:
Ah, sorry I was unclear. I was happy that this doesn't show up in the top bar, I just wondered **why**. Since I've dabbled in hiding modules before, like in #2551.

I'd also like to have the link to `pygame/pygame/blob/master` be changed to `pygame/pygame/blob/main` since we switched default branch naming.

Starbuck5 (approved):
Cool! Hopefully this will be helpful to people.

Starbuck5:
On second look, I see some colors are incorrect.

For example all the golds are just black, same as all the darkoranges, and the greens, for example.

Taffer:
> On second look, I see some colors are incorrect.
> 
> For example all the golds are just black, same as all the darkoranges, and the greens, for example.

That's extremely weird, since Python wrote most of the file; I'll look into it though! Which browser are you using?

Starbuck5:
I'm using Chrome, but I don't think it would matter?

The one's that I'm talking about, on examination of the rst file, only have `background-color`: missing the `color` that most other ones have.

Taffer:
The issue was incorrectly translating colours with values less than 16 in one or more components; a lot of colours were affected. I inspected the entire rendered page and everything looks good now.

illume (approved):
Fun :) Thanks.

PR #2560, opened 2021-04-13
Update stars.py

jonozw:
add pg.quit() to exit gracefully

### DIFF
--- a/examples/stars.py
+++ b/examples/stars.py
@@ -91,6 +91,7 @@ def main():
             elif e.type == pg.MOUSEBUTTONDOWN and e.button == 1:
                 WINCENTER[:] = list(e.pos)
         clock.tick(50)
+    pg.quit()
 
 
 # if python says run, then we should run


zaphodikus (approved):
font_viewer example could do with the same cleanliness fix if someone does not mind fixing other places with the same omission.

ankith26 (approved):
Well technically `pg.quit()` is redundant, because python automatically calls it on exit conditions.
But an explicit `pg.quit()` is also fine
So this PR has no problems it may be merged 👍

ankith26:
Close and open to restart CI

Starbuck5 (approved):
As @zaphodikus says, there are others that are missing the pg.quit(), which is mainly important for IDLE-friendliness.

Maybe those can be compiled in an issue, which would be a good first issue.

But this is a good contribution, that's been waiting since April, so let's merge it.

AnudeepGunukula:
Hi i have made a pull request #2643 which will solve this freezing issue, 
i have added pg.quit() to the examples in necessary places for idle friendliness
I have tested in python IDLE and visual studio as well, its working perfect now.
Please review my pr guys, thanks in advance.

PR #2564, opened 2021-04-17
sdl2 video c

Starbuck5:
This PR aims to almost completely replace the _sdl2.video subsystem with an implementation in C.

This PR follows dlon's work in #1066 and on the `sdl-video-c` branch, which is now over 2000 commits behind. I would not have been able to do this without dlon's great work on the original `sdl-video-c` and the `_sdl2` module itself.

I learned a lot about the Python C API through this process, so regardless of whether it is merged, I've gained some skill.

Why?
- It brings this system more in line with the rest of pygame (lots of examples of how to do things)
- It makes it easier to interface with other pygame systems
- C is more common knowledge than Cython, for current and future contributors

This shouldn't change any aspect of the API for people already using _sdl2.video. The only intentional change I made was to `Window.borderless`, setting that to True should make it borderless, which is the opposite of what happens in the Cython version.

Not to say I don't have some ideas about API changes, but that is beyond the scope of this PR.

**What is `cyvideo.pyx`?**

I left `messagebox` and `get_drivers` alone, I didn't want to have to port those. cyvideo.pyx, or "cython video" is automatically added to the `video` module in `src_py/_sdl2/__init__.py`

**Video C-API?**

I did start to add a Video C API, but I managed to get through everything without needing one. I figured it would be a shame to remove what I did, since somebody will need to implement it in the future. Right now the C-API is one slot, and it's the integer 42.

**video.pyx_old?**

I preserved video.pyx and video.pxd by adding `_old` to the file extensions. The Cython code still has information in it, and I don't want to make that harder to access than necessary. I did transfer lots of comment information over, and I've started documentation, but the Cython code is still gospel.

**Testing?**

Since there is only a single unit test for the video subsystem, I relied mostly on integration testing from complete _sdl2.video programs.
@mcpalmer1980 was nice enough to provide 4 different programs that helped me a lot.

Here are 3 of them: [sdl2_examples_s.zip](https://github.com/pygame/pygame/files/6330511/sdl2_examples_s.zip)

The 4th is Flyboy (from source), https://mcpalmer1980.itch.io/flyboy (the source zip didn't seem to have all the assets so I grabbed the rest from one of the exe downloads)

These programs work perfectly in the new system, as well as the single unittest and the video program in `examples`.


MightyJosip (approved):
Tested this with some examples I could find and something I wrote and everything looks great. Tho I can't guarantee that there are no bugs because this is a pretty huge PR, but I am all for merging this because is experimental and undocumented anyways, and imo this is the right direction for the future of _sdl2 module

Thanks. 👍

Starbuck5:
@galexandreg, as one of the other people I've seen working on this system, I'm curious what your take on this is?

We could stay with Cython, and try to find a way to communicate between the Cython code and C code, and vice versa.

I feel that the Cython code is less clear about how it behaves, whereas the C code is more verbose, but you can see every step of Python-C conversion. But on the other hand, merging something like this seems very difficult. And the Cython code is faster, by maybe 10%.

galexandreg:
> @galexandreg, as one of the other people I've seen working on this system, I'm curious what your take on this is?
> 
> We could stay with Cython, and try to find a way to communicate between the Cython code and C code, and vice versa.
> 
> I feel that the Cython code is less clear about how it behaves, whereas the C code is more verbose, but you can see every step of Python-C conversion. But on the other hand, merging something like this seems very difficult. And the Cython code is faster, by maybe 10%.

Well, I really dont think I have enough C or Cython skills (or even general programming at this point 🤷‍♂️) to have a useful opinion on this. Using (and contributing to) pygame was a really good learning experience, but I am still very far from an expert in anything 😄. I feel like people think I am way better at this than I really am, and while I still follow the project, I have not used it much in the last months.

I don't really have an opinion on this anyway... I guess it feels strange to have different languages for different parts of the library. But in the end, there is not a lot of regular developers on the project so I feel like it should be what the main contributors are the most comfortable with. For an occasional contributor like me, it doesn't really matter.

Starbuck5:
> I feel like people think I am way better at this than I really am

Maybe you're better than you think then :smile:. I thought you were a pro. Don't sell yourself short. Even though pygame is arguably the most popular game library for the most popular language in the world, we've got plenty of amateurs here. I'm a student, ankith26 is a student, I think MightyJosip is a student as well.

Plus, you're a member of the pygame organization, which means you have the power to review things like this. And besides me and MightyJosip and you, I don't know any other members who are active/interested in this system.

illume:
Marked this for pygame 2.1. Sound like a good time to merge it?

PR #2567, opened 2021-04-18
The enter key code is len = 1

zaphodikus:
On Win32, the ENTER key handling completely was defeated for some reason, so chatbox entering of text does not do anything since the unicode for ENTER is '\r' and is length 1, not zero.

### DIFF
--- a/examples/textinput.py
+++ b/examples/textinput.py
@@ -103,10 +103,10 @@ def main():
                     _IMETextPos = max(0, _IMETextPos - 1)
                 elif event.key == pg.K_RIGHT:
                     _IMETextPos = min(len(_IMEText), _IMETextPos + 1)
-
+                # Handle ENTER key
                 elif (
                     event.key in [pg.K_RETURN, pg.K_KP_ENTER]
-                    and len(event.unicode) == 0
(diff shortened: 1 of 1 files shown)


Starbuck5:
The example works on pygame 2.0.0, but not on 2.0.1.

With this example script, you can see that KEYDOWN events like enter or ctrl-c have filled out `unicode` attributes in 1.9.6, and in 2.0.1, but not in 2.0.0.

```
import pygame
pygame.init()

screen = pygame.display.set_mode((400,400))

clock = pygame.time.Clock()
while True:
    clock.tick(100)

        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit

            print(event)
```

So that's the deal with **why** this is broken now.

@zaphodikus Should this test for `unicode` field length at all?

ankith26:
I consider this change as a "fix" from my event refactor PRs, because the output is same as that in 1.9.6

As to this PR, it is looking good, but I feel that completely removing the `event.unicode` check might also be a good option

robertpfeiffer:
I have made a new textinput example to supplant this one after my IME fix and IME docs are merged. I'll take a look at this soon.

PR #2572, opened 2021-04-26
i have add colision detection funciton 

Prabesh-Shrestha:
first pull request

### DIFF
--- a/src_py/colision.py
+++ b/src_py/colision.py
@@ -1,0 +1,14 @@
+from math import sqrt
+
+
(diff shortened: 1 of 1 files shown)


ankith26:
Hello! Thank you for your first time contribution.

Sadly, this PR will most likely not be merged, because it is an API change, and this new API does not need to be in pygame, because pygame users can easily include these functions in their own code (BTW there are a few bugs in the functions).

So closing this PR

PR #2579, opened 2021-05-04
Fix latest warnings for pylint 2.8.2 (Consider using with, Consider using max)

Pierre-Sassoulas:
Hello, this is a small merge request following the release of pylint 2.8.2. One is using a builtin max instead of an if, and the other can be ignored.


ankith26 (approved):
Looks Good! 
Thanks 👍

illume (approved):
:+1:

PR #2580, opened 2021-05-05
Compatibility fix for GraalPython

fniephaus:
[GraalPython](https://github.com/oracle/graalpython) defines [`PYPY_VERSION`](https://github.com/oracle/graalpython/blob/39e26642d7c4a9f490c19db9270753a657b97f30/graalpython/com.oracle.graal.python.cext/include/Python.h#L51) for compatibility with some other extensions. pygame seems to work fine on GraalPython, but only without PyPy-specific code. This makes sure `PYPY_VERSION` is not defined when running on GraalPython.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -33,9 +33,12 @@
 */
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
+
+/* Ensure PyPy-specific code is not in use when running on GraalPython (PR #2580) */
 #if defined(GRAALVM_PYTHON) && defined(PYPY_VERSION)
 #undef PYPY_VERSION
 #endif
+
 #include <SDL.h>
 
 /* IS_SDLv1 is 1 if SDL 1.x.x, 0 otherwise */
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -33,6 +33,9 @@
 */
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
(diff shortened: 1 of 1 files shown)


Starbuck5:
Cool! My only request is for you to add the rationale into the source code too, just so it's easy for someone looking at it potentially years from now to see the purpose.

fniephaus:
Of course, just pushed https://github.com/pygame/pygame/pull/2580/commits/3c4f5de5db6d17511cecb36260a25e57a8cd821d. PTAL.

Starbuck5 (approved):
I don't know much about GraalPython, but I'll take your word (and patch) for it. LGTM.

fniephaus:
Thanks, @Starbuck5!

illume:
Pretty cool that is the only change needed.

Just out of curiosity, I wonder why PYPY_VERSION is defined on GraalPython?

fniephaus:
> Just out of curiosity, I wonder why PYPY_VERSION is defined on GraalPython?

According to [this commit](https://github.com/oracle/graalpython/commit/d86d630d67960e258727078129495913e383dc44), GraalPython defines `PYPY_VERSION` to avoid compilation issues. I assume they actually try to mimic PyPy whenever they can for performance reasons. After all, PyPy and GraalPython deploy JITs that work best when most code is directly written in Python, low-level C code needs to be avoided. Maybe @fangerer or @timfel can elaborate?

fangerer:
We still need `PYPY_VERSION` for other packages like Pandas. In this case, it's about compatibility and not performance. But we plan to get rid of it in near future.

PR #2581, opened 2021-05-07
Add keyword arguments to transform functions

Starbuck5:
This was mostly straightforward, I think.

`set_smoothscale_backend` and `threshold` already had keyword arguments, which I unfortunately had to change.

- `set_smoothscale_backend`  used `type` as its keyword, which we shouldn't do because it's a python keyword.

- `threshold` used `dest_surf` and `surf`, which is fine and all, but `surface` is what all the draw and mask module functions use for their keyword argument needs, so I decided to change these to `surface` and `dest_surface` to be consistent.

Rather than adding dedicated tests in `transform_test.py`, I just made sure at least one call with keywords was happening for each function, and marked them all with comments, so they won't get refactored away unknowingly.

I added Vector2 / float argument support to the sizes passed to `scale` and `smoothscale`, which wasn't on my agenda for this PR, but happened almost by accident. This actually makes the `scale` and `smoothscale` docs more accurate, because they both talk about the size as a sequence of two numbers, not a sequence of two ints.

Related to #808


illume:
Very good. Since there are new names in here, we should validate them with a few people.

If someone else does a review, can you please take note that the argument names are the best they could be?

illume (changes requested):
Very good.

It would be good if someone else can review the names introduced.
Also, types are needed.

Starbuck5:
Alright, I think I've gotten this up to merge quality again.

I fixed a python2 build failure, I fixed some typing inconsistencies this introduced, and I fixed the merge conflict.

ankith26:
The PR is looking good!

Minor nitpick (because illume asked opinions on keyword names), I'd personally prefer `surf` over `surface`, because that seems more "programming style", IDK. And maybe `dest_surface` is too verbose for a keyword name, we could shorten it to `dest_surf` too.

But either way is fine actually

Starbuck5:
Thanks for the feedback.

I went with `surface` and `dest_surface` exclusively for consistency with the draw keywords (done by someone else), and the mask keywords (done by me).

Starbuck5:
The old legends say that one day this PR will pass Appveyor without a random seeming failure on a random Python version 🌟

illume (approved):
Thanks :)

illume:
The reasons for the argument naming seem sound to me.

I'll wait for another person to review this before merge.

ankith26 (approved):
LGTM! 🎉 👍

PR #2582, opened 2021-05-08
Resolve dead link in setup.py seen by debian users

### LINKED ISSUE #2531: Setup.py refers users to nonexistent help URL
Seen on StackOverflow: https://stackoverflow.com/questions/66692760/im-getting-errors-while-installing-pygame-on-my-chromebook-and-to-make-matters

Debian users are directed to https://www.pygame.org/wiki/CompileDebian, by [this line in setup.py](https://github.com/pygame/pygame/blob/main/setup.py#L53)

But the link doesn't exist.

Pierre-Sassoulas:
Closes #2531 and relates to https://github.com/pygame/pygameweb/issues/113

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,11 +54,10 @@ def compilation_help():
         'Darwin': 'https://www.pygame.org/wiki/MacCompile',
         'RedHat': 'https://www.pygame.org/wiki/CompileRedHat',
         # TODO There is nothing in the following pages yet
-        # 'Suse': 'https://www.pygame.org/wiki/CompileSuse',
-        # 'Python (from pypy.org)': 'https://www.pygame.org/wiki/CompilePyPy',
-        # 'Free BSD': 'https://www.pygame.org/wiki/CompileFreeBSD',
-        # TODO This page does not exists yet:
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
LGTM! 👍 
Thanks for the PR 🥳

illume (changes requested):
Cool. Thanks for looking into this.

Do you think you could add the Debian one back? Also uncomment those other ones you've added (and we can add wiki pages for them too).

ps. I've added a wiki page for CompileDebian just linking to the CompileUbuntu one for now.

Pierre-Sassoulas:
Thanks for the review, I updated the list.

> (and we can add wiki pages for them too).

I don't know where the wiki code is but let me know if can do that too :)

Starbuck5:
@Pierre-Sassoulas the wiki can be edited on pygame.org like wikipedia, I believe.

https://www.pygame.org/wiki/

But you need an account on pygame.org, and new account creation has been stopped for a while because of spam.

illume (approved):
👍 🎈 thanks :)

PR #2586, opened 2021-05-09
Python 3.10 support

Starbuck5:
With the release of Python 3.10 beta 1 recently, I was inspired to see if I could build pygame on 3.10.

The first problem I faced was that some of the Cython autogenerated code was using old unicode systems that have been deprecated. I fixed this by completely regenerating two of them on the latest version of Cython. Going over the Cython release notes, I can see that they did this work for 0.29.21.

The other problem I faced was `src_c/rwobject.c(665): error C2065: 'ssize_t': undeclared identifier`, which confused me for a while until I found a numpy PR where they replaced some `ssize_t` with `Py_ssize_t` to prepare for 3.10. I suppose Python is getting more strict about moving people over to ``Py_ssize_t``.

I also wrapped a call to `PyEval_InitThreads()`, which was emitting a deprecation warning when built, in a version tag so it won't get touched on Python 3.7 or above. I've never done anything like that before, so tell me if I'm doing anything weird/wrong there.

This builds for me locally [Windows] and passes the test suite.

---

edit: related: there's now a [Python 3.10 issue](https://github.com/pygame/pygame/issues/2599) tracking some other tasks.


ankith26:
LGTM! 👍

ankith26 (approved):
LGTM 👍 🥳

src_c/rwobject.c
```diff
@@ -662,7 +667,7 @@ _pg_rw_read(SDL_RWops *context, void *ptr, size_t size, size_t maxnum)
 {
     pgRWHelper *helper = (pgRWHelper *)context->hidden.unknown.data1;
     PyObject *result;
-    ssize_t retval;
+    Py_ssize_t retval;
```
illume, line 670: This looks like a bug? The function returns size_t, but the retval is ssize_t. Should they both be size_t instead?

src_c/rwobject.c
```diff
@@ -518,9 +518,14 @@ pgRWops_FromFileObject(PyObject *obj)
     rw->write = _pg_rw_write;
     rw->close = _pg_rw_close;
 
+/* https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads */
```
illume, line 521: 👍

illume (changes requested):
Cool. I left a question about the return type.

src_c/rwobject.c
illume: ps. As a side note - I'm not sure if this is the kind of thing that should be reported to python as breakage? I'm guessing this type will come back before the official release since that type is fairly common.

Starbuck5:
I've been doing a bit more reading about it and I'm not gaining any new understanding.

Here's the numpy PR I found for reference: https://github.com/numpy/numpy/pull/18890

The function uses `retval` as the variable for `read()`, which returns `size_t`, and also for `PyBytes_GET_SIZE` / `PyString_GET_SIZE` (depending on py2 or 3), which returns `Py_ssize_t`.

Perhaps the solution is to split those cases into two separate variables and forget about it? I don't have a good understanding of all the different numeric types in C / CPython.

src_c/rwobject.c
ankith26: AFAIK, `Py_ssize_t` is just an alias for `ssize_t` when `ssize_t` is available, and when it is not, defines an equivalent implementation of the same. So basically, `Py_ssize_t` should always be defined

illume:
Thanks for looking into it. Seems like Py_ssize_t would cover both requirements.

PR #2587, opened 2021-05-12
Fixes window minimum size when leaving SCALED on Windows.

### LINKED ISSUE #2327: Window size can't be reduced going from a SCALED window to one that's not SCALED
Hello,

**Environment:**
- **Operating system** : Windows 10
- **Python version** : 3.7.7
- **SDL version** : 2.0.12
- **PyGame version** : 2.0.0

**Current behavior:**

When switching from a SCALED window to one that's not SCALED, if the width or the height is lowered, it would still be the value of the previous SCALED window.

**Expected behavior:**

I would expect my new window to have the size I set in the parameters.

**Test code**

```
import pygame as pg
pg.init()

screen = pg.display.set_mode((500, 500), pg.SCALED)
print("Initial screen (500x500 SCALED) : " + repr(screen))

while True:
    for e in pg.event.get():
        if e.type == pg.KEYDOWN:
            if e.key == pg.K_a:
                screen = pg.display.set_mode((500, 500))
                print("Same screen (500x500) : " + repr(screen))
            if e.key == pg.K_z:
                screen = pg.display.set_mode((600, 600))
                print("Bigger screen (600x600) : " + repr(screen))
            elif e.key == pg.K_e:
                screen = pg.display.set_mode((400, 400))
                print("Smaller screen (400x400) : " + repr(screen))
            elif e.key == pg.K_r:
                screen = pg.display.set_mode((600, 400))
                print("Wider, shorter screen (600x400) : " + repr(screen))
    pg.display.update()

```

Log of a test execution of the previous code :

```
Initial screen (500x500 SCALED) : <Surface(500x500x32 SW)>
Bigger screen (600x600) : <Surface(600x600x32 SW)>
Smaller screen (400x400) : <Surface(500x500x32 SW)>
Wider, shorter screen (600x400) : <Surface(600x500x32 SW)>
Same screen (500x500) : <Surface(500x500x32 SW)>
```

Starbuck5:
Fixes #2327 

`SCALED` mode sets a minimum size for its window: the requested user size.
When leaving `SCALED` mode with another call to `display.set_mode()`, different things can happen on different operating systems.
I'm not totally sure why, but on Windows, the window size is changed on the same `SDL_Window` object. On other platforms, it ends up that a new window of a different size is created.

That's why this bug only affected Windows. Because other platforms ended up with new `SDL_Window` objects, those new objects did not carry over the minimum size set by `SCALED` in the previous call.

There isn't an "unset" minimum size, so I just changed it to set a minimum size of `(1,1)` in the relevant part of the code. I was worried about this causing an edge case with user requested `(0, 0)` sized windows, but 0-dimensions are treated as requests to use the display resolution dimension, so we're good.

The original issue has a good manual test code, for anyone who wants to test this locally.

I also added an automated test!

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1240,6 +1240,12 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                 init_flip = 1;
             }
             else {
+                /* set min size to (1,1) to erase any previously set min size
+                 * relevant for windows leaving SCALED, which sets a min size
+                 * only relevant on Windows, I believe.
+                 * See https://github.com/pygame/pygame/issues/2327 */
+                SDL_SetWindowMinimumSize(win, 1, 1);
+
                 /* change existing window.
                  this invalidates the display surface*/
                 SDL_SetWindowTitle(win, title);
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -559,6 +559,16 @@ class DisplayModuleTest(unittest.TestCase):
             winsize[0] / surf.get_size()[0], winsize[1] / surf.get_size()[1]
         )
 
+    def test_set_mode_unscaled(self):
+        """ Ensures a window created with SCALED can become smaller. """
+        # see https://github.com/pygame/pygame/issues/2327
+
+        screen = pygame.display.set_mode((300,300), pygame.SCALED)
+        self.assertEqual(screen.get_size(), (300,300))
+
+        screen = pygame.display.set_mode((200,200))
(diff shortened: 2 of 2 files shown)


robertpfeiffer:
Let me look through the SDL2 code before we merge this. As is, this is better than nothing, but maybe we can fix it in SDL. The test is good in any case

robertpfeiffer:
After studying the SDL2 side, I see no problem with this patch.

illume (approved):
Thanks!

+100 for including a SCALED test

PR #2590, opened 2021-05-14
Lots of docs fixes

Starbuck5:
In order of importance:

- Stole some css from Sphinx's `basic.css` to fix the last major Sphinx 3 bug.
- Got rid of two warnings coming from `event.rst`. To do this, I had to move around some of the information about the new `touch` attributes on `MOUSE` events. (Apparently, we can't have more than one line in a `.. versionchanged:`)
- Since the documentation build process is python 3 only, I took the opportunity to remove python 2 support from something in pygame. I also ran `black` on it for good measure. So the code in `docs/reST/ext` is now prettier than ever.
- Added the `allowedchanges` argument to the `pre_init()` docs, also clarified that to get a default for this you have to enter `-1`. Someone on the pygame discord found that `allowedchanges` was undocumented but worked in `pre_init()`.
- Linked to `event.custom_type()` when it talks about `USEREVENT` in `event.rst`
- Corrected a few random typos
- Corrected a couple errors in pygame.css_t that hadn't been noticed

It's so beautiful: (See point 3 of #2513 for context)
beautiful

Through my research on this PR, I learned that themes are supposed to import Sphinx's `basic.css`. This is a file that gets added to the generated docs by Sphinx, but we don't do anything with it. I think this would help maintainability in the future, as when Sphinx or docutils changes Sphinx will update their `basic.css` along with it. But that's beyond the scope of this PR.

Also, Sphinx 4 is out now. And it breaks more stuff :cry: (In fairness to Sphinx, it might be docutils that has changed). So far I've noticed that something is funky with the codeblocks that breaks the image positioning in the tutorials, when built on Sphinx 4.

**For people building the docs:** The latest version of Sphinx I think will work 3.5.4.


docs/reST/ext/indexer.py
```diff
@@ -57,21 +55,31 @@ def prep_document_info(app, env, docname):
...
+
 class CollectInfo(Visitor):
 
     """Records the information for a document"""
-    
-    desctypes = set(['data', 'function', 'exception', 'class',
-                     'attribute', 'method', 'staticmethod', 'classmethod'])
+
+    desctypes = set(
+        [
```
ankith26, line 72: Minor nitpick here
Why declare a list and make it a set, when you can directly declare it as a set?

ankith26 (changes requested):

docs/reST/ext/indexer.py
Starbuck5: Good catch.
I didn't even know the `dict`-esque syntax to do that existed.

illume:
I left a note about the sphinx version here: https://www.pygame.org/wiki/Hacking#Generating%20docs

illume (approved):
👍 🎉

illume:
I added the doc TODOs found in this PR and made an issue from them here: https://github.com/pygame/pygame/issues/2598

PR #2592, opened 2021-05-18
Build configuration for MSYS2

dciabrin:
Support MSYS2 as an additional Windows-native target.
The new build config is adapted from config_windows, and
only supports pre-built dependencies in the form of
MSYS2 packages available in the MSYS2 distribution.

config_msys2 can be configured to bundle dll dependencies
in the target installation (e.g. to build wheels) or to
rely on external dlls (e.g. to build a MSYS2 package for PyGame).

### DIFF
--- a/buildconfig/config.py
+++ b/buildconfig/config.py
@@ -33,6 +33,12 @@ def print_(*args, **kwds):
     msysio.print_(*args, **kwds)
 
 
+def is_msys2():
+    """Return true if this in an MSYS2 build"""
+    return ('MSYSTEM' in os.environ and
+            re.match(r'MSYS|MINGW.*|CLANG.*|UCRT.*', os.environ['MSYSTEM']))
+
+
 def is_msys_mingw():
     """Return true if this in an MinGW/MSYS build
 
@@ -175,6 +181,14 @@ def main(auto=False):
             import config_conan as CFG
         except ImportError:
             import buildconfig.config_conan as CFG
+
+    elif (sys.platform == 'win32' and
+        (sys.version_info >= (3, 8) and is_msys2())):
+        print_('Using WINDOWS MSYS2 configuration...\n')
+        try:
+            import config_msys2 as CFG
+        except ImportError:
+            import buildconfig.config_msys2 as CFG
 
     elif (sys.platform == 'win32' and
         # Note that msys builds supported for 2.6 and greater. Use prebuilt.
--- a/buildconfig/config_msys2.py
+++ b/buildconfig/config_msys2.py
@@ -1,0 +1,520 @@
+"""Config on MSYS2"""
+
+# The search logic is adapted from config_win.
+# This assumes that the PyGame dependencies are resolved
+# by MSYS2 packages.
+
+try:
+    from setup_win_common import get_definitions
+except ImportError:
+    from buildconfig.setup_win_common import get_definitions
+
+import os, sys
+import re
+import logging
+import subprocess
+from glob import glob
+from distutils.sysconfig import get_python_inc
+
+
+def get_ptr_size():
+    return 64 if sys.maxsize > 2**32 else 32
+
+def as_machine_type(size):
+    """Return pointer bit size as a Windows machine type"""
+    if size == 32:
+        return "x86"
+    if size == 64:
+        return "x64"
+    raise BuildError("Unknown pointer size {}".format(size))
+
+def get_machine_type():
+    return as_machine_type(get_ptr_size())
+
+def get_absolute_win_path(msys2_path):
+    output = subprocess.run(['cygpath', '-w', msys2_path],
+                            capture_output=True, text=True)
+    if output.returncode != 0:
+        raise Exception("Could not get absolute Windows path: %s"%msys2_path)
+    else:
+        return output.stdout.strip()
+
+class Dependency(object):
+    huntpaths = ['..', '../..', '../*', '../../*']
+    inc_hunt = ['include']
+    lib_hunt = ['lib']
+    check_hunt_roots = True
+    def __init__(self, name, wildcards, libs=None, required=0, find_header='', find_lib=''):
+        if libs is None:
+            libs = []
+        self.name = name
+        self.wildcards = wildcards
+        self.required = required
+        self.paths = []
+        self.path = None
+        self.inc_dir = None
+        self.lib_dir = None
+        self.find_header = find_header
+        if not find_lib and libs:
+            # Windows dllimport libs (.lib) are called .dll.a in MSYS2
+            self.find_lib = r"lib%s\.dll\.a" % re.escape(libs[0])
+        else:
+            self.find_lib = find_lib
+        self.libs = libs
+        self.found = False
+        self.cflags = ''
+        self.prune_info = []
+        self.fallback_inc = None
+        self.fallback_lib = None
+
+    def hunt(self):
+        parent = os.path.abspath('..')
+        for p in self.huntpaths:
+            # for w in self.wildcards:
+            # found = glob(os.path.join(p, w))
+            found = glob(p)
+            found.sort() or found.reverse()  #reverse sort
+            for f in found:
+                if f[:5] == '..'+os.sep+'..' and \
+                   os.path.abspath(f)[:len(parent)] == parent:
+                    continue
+                if os.path.isdir(f):
+                    self.paths.append(f)
+
+    def choosepath(self, print_result=True):
+        if not self.paths:
+            if self.fallback_inc and not self.inc_dir:
+                self.inc_dir = self.fallback_inc[0]
+            if self.fallback_lib and not self.lib_dir:
+                self.lib_dir = self.fallback_lib[0]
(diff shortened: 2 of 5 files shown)


robertpfeiffer:
So to be perfectly clear, can you confirm/answer this?

1. This is *not* a cross compilation setup, yet. You have to run it on windows.
2. This searches the msys installation for SDL DLLs
3. The SDL needs to be compiled with GCC/msys and glibc, with no msvcrt.dll linkage
4. This always downloads the same prebuilt toolchain at the moment

dciabrin:
>     1. This is _not_ a cross compilation setup, yet. You have to run it on windows.

I'd say that's correct (which a twist explained in 3.)
This config can only be used on windows, and assumes it is being run in a MSYS2 environment (i.e. from a MSYS2 shell like bash.exe). The autoconfig step won't select the MSYS2 config if the setup.py script is not run from a MSYS2 shell.

> 
>     2. This searches the msys installation for SDL DLLs

That's correct.
Reasoning behind it is - unlike the original MSYS - MSYS2 is an environment that comes with a package manager (arch's pacman), and it just so happens that all the PyGames and SDL2 dependencies are already available in MSYS2 packages.
So when building native libs for MSYS2, a quick means of linking against the right dependencies is to use the DLL from those MSYS2 packages. 

> 
>     3. The SDL needs to be compiled with GCC/msys and glibc, with no msvcrt.dll linkage

I'd say that's slightly incorrect.
MSYS2 comes with two different environments/subsystems/runtimes [1]. One that targets a cygwin-like runtime with glibc, and one that strictly depends on the native windows runtime (i.e. it links against msvcrt.dll rather than glibc).
The buildconfig targets the native windows runtime and uses the gcc cross-compiler [2]  to target that windows runtime only.

The same is true for all the MSYS2 packages that the buildconfig scans dll from: they all target the native windows runtime [3,4], so this is safe to link against them, and the resulting PyGame only depends on the windows runtime.

>     4. This always downloads the same prebuilt toolchain at the moment
That's correct.
Reason being, people wanting to use PyGame on the MSYS2 environment will most likely want to MSYS2 packages for the dependencies, as they are already available and each dependency is updated regularly in MSYS2 [5] to track upstream.
So I figured the quickest way to download the right PyGame dependencies was to consume those packages in the buildconfig.


[1] https://www.msys2.org/docs/environments/
[2] https://packages.msys2.org/package/mingw-w64-x86_64-gcc
[3] e.g. https://packages.msys2.org/package/mingw-w64-x86_64-SDL2?repo=mingw64
[4] e.g. https://packages.msys2.org/package/mingw-w64-x86_64-libpng?repo=mingw64
[5] https://github.com/msys2/MINGW-packages

illume:
Amazing. So far we don't have up to date dependencies on Windows... and it's kind of a problem since we use ancient ones. It's especially with freetype because of font differences.

I wonder if binaries made with the pygame this produces will have any dependencies on msys2? If not, maybe we can release pygame with this toolchain.

I'll try it now and see if I can get it building for me locally...

dciabrin:
> I wonder if binaries made with the pygame this produces will have any dependencies on msys2? If not, maybe we can release pygame with this toolchain.

My understanding of the MSYS2 docs [1] is that when using the cross-compiler that targets a native windows environment (mingw64 or ucrt), the generated binaries only contains runtime dependencies to the windows runtime. The few cases that require additional glue/shim functions (e.g. the improvements to msvcrt's printf for C99 compatibility) are implemented via static libraries.

That being said, if you meant "building wheels for Windows using DLL dependencies from MSYS2 packages", that sure looks like a bold move, or a maintenance nightmare :) 

[1] https://www.msys2.org/docs/environments/

> I'll try it now and see if I can get it building for me locally...

Thanks for your time!

mljli:
Great work! This allows us to use/develop pygame with msys and mingw on Windows. 

As for wheels, binaries generated by mingw do have dependencies on msys2 packages (as long as it's dynamic linked). Check out the ldd output of an SDL program below.

```sh
$ ldd hello.exe
        ntdll.dll => /c/WINDOWS/SYSTEM32/ntdll.dll (0x7ffefd290000)
        KERNEL32.DLL => /c/WINDOWS/System32/KERNEL32.DLL (0x7ffefb9f0000)
        KERNELBASE.dll => /c/WINDOWS/System32/KERNELBASE.dll (0x7ffefab60000)
        msvcrt.dll => /c/WINDOWS/System32/msvcrt.dll (0x7ffefbff0000)
        SHELL32.dll => /c/WINDOWS/System32/SHELL32.dll (0x7ffefc790000)
        msvcp_win.dll => /c/WINDOWS/System32/msvcp_win.dll (0x7ffefafe0000)
        ucrtbase.dll => /c/WINDOWS/System32/ucrtbase.dll (0x7ffefae80000)
        USER32.dll => /c/WINDOWS/System32/USER32.dll (0x7ffefb500000)
        win32u.dll => /c/WINDOWS/System32/win32u.dll (0x7ffefaa80000)
        GDI32.dll => /c/WINDOWS/System32/GDI32.dll (0x7ffefc090000)
        gdi32full.dll => /c/WINDOWS/System32/gdi32full.dll (0x7ffefb080000)
        SDL2.dll => /mingw64/bin/SDL2.dll (0x7ffec1470000)
        ADVAPI32.dll => /c/WINDOWS/System32/ADVAPI32.dll (0x7ffefb890000)
        sechost.dll => /c/WINDOWS/System32/sechost.dll (0x7ffefbb20000)
        RPCRT4.dll => /c/WINDOWS/System32/RPCRT4.dll (0x7ffefcf80000)
        IMM32.dll => /c/WINDOWS/System32/IMM32.dll (0x7ffefc270000)
        ole32.dll => /c/WINDOWS/System32/ole32.dll (0x7ffefb370000)
        combase.dll => /c/WINDOWS/System32/combase.dll (0x7ffefbc20000)
        OLEAUT32.dll => /c/WINDOWS/System32/OLEAUT32.dll (0x7ffefb6a0000)
        SETUPAPI.dll => /c/WINDOWS/System32/SETUPAPI.dll (0x7ffefc320000)
        cfgmgr32.dll => /c/WINDOWS/System32/cfgmgr32.dll (0x7ffefae30000)
        bcrypt.dll => /c/WINDOWS/System32/bcrypt.dll (0x7ffefaab0000)
        VERSION.dll => /c/WINDOWS/SYSTEM32/VERSION.dll (0x7ffef4320000)
        WINMM.dll => /c/WINDOWS/SYSTEM32/WINMM.dll (0x7ffef0360000)
```

dciabrin:
> Great work! This allows us to use/develop pygame with msys and mingw on Windows.
> 
> As for wheels, binaries generated by mingw do have dependencies on msys2 packages (as long as it's dynamic linked). Check out the ldd output of an SDL program below.
> 

I think there are two sides to the discussion here.
Bundling MSYS2 dll in a wheel that is meant to be consumed in a MSYS2 env should work. The only complication that I see is that once compiled, MSYS2 dll lost the notion of a POSIX root path (i.e. "/"). They only know Windows absolute path. Luckily, for what I saw and tested so far, it seems that the dll built for MSYS2 packages only depend on dll names, i.e. they do not depend on a library's absolute path. I could verify that building a pygame wheel in a MSYS2 env installed in e.g. "C:\foo" could be consumed in a MSYS2 env installed in "C:\bar". So building wheels with .dll from MSYS2 packages seems to work fine.  

The other side of the question is "would it work to bundle .dll build for MSYS2 into the existing windows wheel?" I don't see why it wouldn't work because in theory MSYS2 .dll don't have any runtime dependencies to external MSYS2 libraries. However, I suspect this is potentially fragile and my gut feeling is that we might see surprises when bundling dll built for MSYS2 in a wheel that is meant to be used outside of MSYS2. In fact, any advise from a MSYS2 developer would certainly be well advised :)

mljli:
@dciabrin Correct. My point is we have to bundle the DLLs if the wheel is supposed to be used outside of MSYS2. And I see that you add an environment variable to control the behavior.

dciabrin:
@mljli oh ok I didn't get the point..

For the record, i made the bundling of DLLs conditional because I was originally planning to provide a MSYS2 package for PyGame, and in this scenario the MSYS2 pygame package shouldn't ship dlls already provided by other MSYS2 packages.

Having the conditional allows to build a wheel, in case some users would be interested in "pip install pygame" from their MSYS2 env.

I once tried to force install in my MSYS2 env wheels that were built for official cpython from python.org, but I never managed to make it work. At this stage, I think the opposite is probably not going to work either: installing wheels made for MSYS2 in an official cpython's environment seems too much of a stretch to be useful.

Now I wonder if @illume will have success in building a wheel for windows with dll coming from MSYS2 packages.

illume (approved):
Thanks! 🎈 🎉 

I didn't try building a wheel with this yet, but don't think that should stop this very useful PR being merged in.

dciabrin:
Heya folks,

It's been some time since I last logged to an MSYS2 env, so I didn't check if this revision of the PR is still up to date, but from the automatic CI tests, it doesn't look bad (Travis failed on a different issue).

Let me know if you feel I should push another revision or if I can do anything else, thanks!

Issue #2553, opened 2021-04-09
Excessive time to load font with pygame.font.SysFont on Mac


IrvKalb:
- **Operating system**: Mac OSX 11.2.3
- **Python version** : 3.9.1
- **SDL version**: 2.0.14
- **PyGame version**: 2.0.1

**Current behavior:**

The first font loaded by name using pygame.font.SysFont(), is taking over 7 seconds on my Mac to load.  

**Expected behavior:**

I would expect that a font loaded by name would be almost instantaneous - as it is when using pygame.font.font()

**Steps to reproduce:**

See sample program below.  The first font loaded using a call to pygame.font.SysFont() takes a a long time.  After the first font, additional calls to pygame.font.SysFont() are almost instanteous.  Sample output:

Loading System font with pygame.font.Font took 0.0007171630859375
Loading Arial with pygame.font.SysFont took 7.108750104904175
Loading Courier with pygame.font.SysFont took 0.0005779266357421875

But the problem appears to be with pygame.font.get_fonts().  If I add in a call to that after the first call load the System font, I get this result:

Loading System font with pygame.font.Font took 0.0006759166717529297
Getting fonts with pygame.font.get_fonts() took 7.203648805618286
Loading Arial with pygame.font.SysFont took 0.000492095947265625
Loading Courier with pygame.font.SysFont took 0.0006482601165771484

This does not seem to be a problem on Windows, only Mac

**Test code**
```
import sys
import pygame
import time
 
pygame.init()
screen = pygame.display.set_mode((400, 400))
clock = pygame.time.Clock()

now = time.time()
font1 = pygame.font.Font(None, 24)
print('Loading System font with pygame.font.Font took', time.time() - now)

#Uncomment these lines to see that the slowdown is related to getting a list of fonts.
#now = time.time()
#fontList = pygame.font.get_fonts()
#print('Getting fonts with pygame.font.get_fonts() took', time.time() - now)

now = time.time()
font2 = pygame.font.SysFont('Arial', 24)
print('Loading Arial with pygame.font.SysFont took', time.time() - now)

now = time.time()
font3 = pygame.font.SysFont('Courier', 24)
print('Loading Courier with pygame.font.SysFont took', time.time() - now)


while True:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            pygame.quit()
            sys.exit()
     
    screen.fill((255, 255, 255))
     
    pygame.display.flip()
    clock.tick(30)
```

[...]

PR #2594, opened 2021-05-20
Overhaul SysFont on Mac

### LINKED ISSUE #2549: Font loading does not handle missing system_profiler command from PATH
**Environment:**

- **Operating system**: macOS Mojave 10.14.6
- **Python version**: 3.9.1
- **SDL version**: 2.0.14
- **PyGame version**: 2.0.1
- **Relevant hardware**: n/a

**Current behavior:**

If you run in an environment where your path may be incorrect and `system_profiler` is not found, the font loading module of pygame will crash with a seemingly unrelated exception (`xml.etree.ElementTree.ParseError`).

In my case I'm running the `fish` shell on a computer I bought and set up recently, and `/usr/sbin/` was not in my path. Adding it in fixed the problem, but I had to debug the pygame library to find out what was happening.

**Expected behavior:**

My expectation would be that if the subprocess for loading fonts fails to run the command, that an error detailing as such will be raised. This is the offending function:

```python
def _system_profiler_darwin():
    fonts = {}
    flout, _ = subprocess.Popen(
        ' '.join(['system_profiler', '-xml', 'SPFontsDataType']),
        shell=True,
        stdout=subprocess.PIPE,
        stderr=subprocess.PIPE,
        close_fds=True
    ).communicate()

    for font_node in ET.fromstring(flout).iterfind('./array/dict/array/dict'):
        _add_font_paths(font_node.iter("*"), fonts)

    return fonts
```

stderr is assigned to `_` and contains information regarding the error `Unknown command: system_profiler`, but because the error isn't accounted for the code continues. `flout`, which is now an empty binary string `b''`, is then passed to `fromstring()` which then produces the exception shown in my screenshot.

**Screenshots**

**Steps to reproduce:**

This code will immediately fail if your path is missing `/usr/sbin/`, since pygame can't find `system_profiler`:

```python
import pygame
pygame.init()
font = pygame.font.SysFont('Arial', 12)
```

**Test code**

Listed above

N/A

Starbuck5:
This fixes #2549 and #2553.

`system_profiler` is slowwwwwwwww
It hasn't been a problem in the past because pygame SysFont prefers to use the X11 installation that exists on MacOS to report fonts. (TIL there's an X11 install on some Macs). But newer Macs / Mac OS versions don't have this, so their users are starting to grate against the ~10 seconds `system_profiler` takes to give results.

To fix this, I wrote the `_font_finder_darwin()` function. It grabs fonts from all the locations I saw `system_profiler` grabbing fonts.

However, I discovered another problem. The fonts were not being "grouped" properly before. All the sysfont init functions return a dictionary with the format `{simple_font_name: {(bold, italic): font_file}, ...}`. But the `system_profiler` derived font names weren't being parsed properly. For example, Arial bold would be represented as a different font, not bold or italic.

It would look something like this:
```
fonts = {"arial": {(False, False): "/System/Library/Arial.ttf"}, 
         "arialbold": {(False, False): "/System/Library/Arial-Bold.ttf"}}
```

When it should look like:
```
fonts = {"arial": {(False, False): "/System/Library/Arial.ttf", (True, False): "/System/Library/Arial-Bold.ttf"}}
```

I made a "fidelity test" to demonstrate this, which you can download if you want - [fidelitytest.py.zip](https://github.com/pygame/pygame/files/6514855/fidelitytest.py.zip)

TLDR: It requests Arial and also requests Bold

The old (`system_profiler` path based) output was:
fake

The new output is:      (this is the same as the X11 path output, as well)
legit

This is because the old version didn't categorize the fonts properly, so the font module had to fake the bold. Whereas in the second version, it has a premade bold style to work from, making it much prettier on the eyes.

**Note for reviewers:**
Although this does "overhaul" the system, macs that were using the X11 based Sysfont will still use it, so this won't change compatibility on anyone who had a working SysFont module on MacOS. This makes this PR safer to merge than it would be otherwise.

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -177,7 +177,7 @@ def _add_font_paths(sub_elements, fonts):
 def _system_profiler_darwin():
     fonts = {}
     flout, _ = subprocess.Popen(
-        ' '.join(['system_profiler', '-xml', 'SPFontsDataType']),
+        ' '.join(['/usr/sbin/system_profiler', '-xml', 'SPFontsDataType']),
         shell=True,
         stdout=subprocess.PIPE,
         stderr=subprocess.PIPE,
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -27,10 +27,6 @@ from os.path import basename, dirname, exists, join, splitext
 from pygame.font import Font
 from pygame.compat import xrange_, PY_MAJOR_VERSION, unicode_
 
-if sys.platform == 'darwin':
-    import xml.etree.ElementTree as ET
-
-
 OpenType_extensions = frozenset(('.ttf', '.ttc', '.otf'))
 Sysfonts = {}
 Sysalias = {}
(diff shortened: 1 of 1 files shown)


illume (approved):
👍 🎉 Thanks, this is a really nice improvement.

Looks like you've done a lot of testing. For older systems, and ones with fc-list, I think it's a pretty good idea to use that. Since it does caching. Newer systems have SSD fast storage usually, and it seems like you've done lots of testing too.

PR #2600, opened 2021-05-22
Add github actions, disable travis builds except arm64

illume:
Continued on from [PR by Ankith](https://github.com/pygame/pygame/pull/2507) (Mostly this is by Ankith, with a few tweaks).

- windows python 2 build fails because someone at MS removed compiler-for-py2 download. We use appveyor for these instead.
- Disabled all windows python except 3.5. We use appveyor for others instead.
- windows, pypy stopped distributing 32bit builds, and now does 64bit only. We use appveyor for these.
- release uploads are disabled for now
- all builds on Travis are disabled, except arm64 (which is allowed to fail).
- gave the jobs different names like `build-windows`, rather than all of them being `build`.

There's more discussion below, and further work possible... but this is good for now.

With everything passing here, we will move to requiring github actions to pass before allowing a merge.

Closes issue: https://github.com/pygame/pygame/issues/2300

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -28,7 +28,7 @@ on:
     - '.github/workflows/sdl1-sdist.yml'
 
 jobs:
-  build:
+  build-macos:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false  # if a particular matrix build fails, don't skip the rest
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,87 @@
+name: MacOS
+
+# Run CI only when a release is created, on changes to main branch, or any PR 
+# to main. Do not run CI on any other branch. Also, skip any non-source changes 
+# from running on CI
+on:
+  release:
+    types: [created]
+  push:
+    branches: main
+    paths-ignore:
+    - 'docs/**'
+    - 'examples/**'
+    - '.gitignore'
+    - 'README.rst'
+    - '.github/workflows/linux.yml'
+    - '.github/workflows/windows.yml'
+    - '.github/workflows/sdl1-sdist.yml'
+  pull_request:
+    branches: main
+    paths-ignore:
+    - 'docs/**'
+    - 'examples/**'
+    - '.gitignore'
+    - 'README.rst'
+    - '.github/workflows/manylinux.yml'
+    - '.github/workflows/windows.yml'
+    - '.github/workflows/sdl1-sdist.yml'
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false  # if a particular matrix build fails, don't skip the rest
+      matrix:
+        # TODO: when Github actions supports M1 Mac, add `macos-11.0` to this
+        os: [macos-10.15]
+        pyversion: ['pypy-3.7', 'pypy-2.7', '2.7', '3.5', '3.6', '3.7', '3.8', '3.9']
+
+    steps:
+    - uses: actions/checkout@v2.3.4
+    
+    - name: Set up Python ${{ matrix.pyversion }}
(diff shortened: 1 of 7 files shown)


ankith26:
Looking very good! 👍
Just a few stuff need to be looked into.
- In the middle-term, might be worth having windows builds on GH actions too. In my initial PR, all windows builds worked, but that should not stop this PR from being merged, I could look into it sometime in the future ;)
- The mac build gives off a lot of warnings. I am not so sure about the wheels it gives, because that is the only thing I have not manually tested out with the wheels. Among the things it is complaining, I see a "using 10.15 SDL for a 10.14 build" message, which does not look good. Maybe the wheels should be tested out by someone on a mac, so that we'd feel safer

ankith26:
Found why 3.6 and above win builds were failing on GH Actions. The error is
`src_c/SDL_gfx/SDL_gfxPrimitives.c(3981): error C2169: 'lrint': intrinsic function, cannot be defined`

ankith26:
I hope me commiting to this branch is not a problem for you, illume :)

Basically, I wanted to get 3.6+ and PyPy back on Windows GH actions CI again.

Getting 3.6+ working was a easy fix, newer versions of MSVC would fail at building a file, because we were defining an intristic (well, this was techically supposed to go in another PR, because it is an unrelated fix, but mehh, I wanted GH actions to run on it). But the good news is, 3.6 and above windows builds works now on Github Actions!

Coming to the PyPy situation on windows, basically, we have access to PyPy 7.3.3 win32 builds (on python 2.7, 3.6 and 3.7) and the latest PyPy 7.3.4 does not give win32 builds, but win64 ones (on python 2.7 and 3.7).

So I made it such that we do 7.3.3 builds for win32, and v7.3.4 for win64 builds. This is almost working, I guess (except for the usual python2 being a jerk, in it's natural habitat ;) )

illume:
> I hope me commiting to this branch is not a problem for you, illume :)

No, that's the benefit of using a branch on the pygame repo :)

But I think probably we can merge it if this latest commit works, and do other fixes in other PRs. So we can take advantage of the extra test coverage.

Some notes for later:

- The mac builds are not usable for release, but usable for testing. We can't use homebrew for releasing because they use arch specific stuff. Future work would be to use another dependency system. Probably not conan... since the SDL related packages are not used by many people and not maintained well.
- cibuildwheel is maybe the future for some of our builds? https://github.com/pypa/cibuildwheel
- the windows wheels from github actions would also need to be tested manually (especially with older machines). 
- I still think disabling all but one windows build is best for now (because they are redundant with the appveyor ones). Keeping one is nice just to make sure gh actions windows is still working.  But hey... it caught one bug already.
- Maybe the new msys2 buildchain is the future for windows, because updated decencies (freetype etc).
- the gh windows ones are pretty slow at up to 9m for each of them VS 1-2m for appveyor builds.

ankith26:
Good points you mentioned. I agree with them

But I am still in favour of windows builds on GH Actions.
- They are faster. Basically, each windows build takes about 1-2 mins for cpython versions, and 8-9 mins for PyPy builds (which we should be able to greatly improve, by caching deps, PyPy builds happen to build stuff like numpy from source RN)
But because each build is parallelized, the overall time is much less, like about 10 mins per commit. Appveyor on the other hand, has a bit of a startup delay (not much) and each build again takes 1-2 mins. But they happen in a sequence, and overall take bout 15-20 mins
- Another big advantage of Github actions is that the build files are uploaded for everyone to see. This means issue reporters can test the wheels for themselves, so that they can confirm that the fix for their problem works
- Upgrading python versions is very easy, especially PyPy, on github actions. No need to manually write a script to pull PyPy, and keep updating it everytime PyPy releases a new version, with GH actions, it is much easier

ankith26:
Or maybe it can be something for future PRs. This one should be all set to go! :)

Issue #2601, opened 2021-05-22
arm64 build broken: display.c arm64 failure ( ‘last_height’ may be used uninitialized in this function )


illume:
```c
src_c/display.c: In function ‘pg_list_modes’:
src_c/display.c:1646:34: error: ‘last_height’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
         if (last_width == mode.w && last_height == mode.h && last_width != -1) {
```

PR #2602, opened 2021-05-23
Fix display.list_modes()

MightyJosip:
This actually has no end result, because the first time last_height gets used it will be initialized, but this PR is about fixing compiler warning (#2601)


ankith26 (approved):
LGTM :)
👍

illume (approved):
👍 cool, thanks

PR #2604, opened 2021-05-23
Fix pyinstaller issue with default font

### LINKED ISSUE #2603: PyInstaller + pygame default font
I thought it would be handy to have an issue to track this.

I had heard people talking about how pygame default font crashes when a game is packaged with PyInstaller, and they're right, sadly.

This is something we should be able to fix in the pygame pyinstaller hook right?

Here's a small test script:
```py
import pygame

pygame.init()

screen = pygame.display.set_mode((400,240))
font = pygame.font.Font(None, 80)
#font = pygame.font.Font("freesansbold.ttf", 80)
#font = pygame.font.SysFont(None, 80)
text = font.render("PYGAME", True, (255,255,255))

clock = pygame.time.Clock()

while True:
    screen.fill("purple")
    screen.blit(text, (10,10))
    
        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit

    pygame.display.flip()
    clock.tick(144)
```

Test procedure:
```
Terminal: PyInstaller testfont.py
Terminal: run generated exe
```

When frozen, the Font based ones emit:
```
  File "testfont.py", line 6, in <module>
    font = pygame.font.Font(None, 80)
TypeError: expected str, bytes or os.PathLike object, not _io.BytesIO
```

And the SysFont one emits:
```
  File "testfont.py", line 8, in <module>
    font = pygame.font.SysFont(None, 80)
  File "pygame\sysfont.py", line 428, in SysFont
  File "pygame\sysfont.py", line 346, in font_constructor
TypeError: expected str, bytes or os.PathLike object, not _io.BytesIO
```

Interestingly, when you go into the `pygame` folder in `dist`, and remove `freesansbold.ttf`, it emits:
```
  File "testfont.py", line 6, in <module>
  File "pygame\pkgdata.py", line 84, in getResource
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\Users\\Starbuck\\Desktop\\ftest\\dist\\testfont\\pygame\\freesansbold.ttf'
```

@ankith26 Got any ideas?

ankith26:
Quickfix, closes #2603 
Please ignore the formatting changes, my editor does that automatically :/

### DIFF
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -784,6 +784,7 @@ font_init(PyFontObject *self, PyObject *args, PyObject *kwds)
         return -1;
     }
 
+    /* Incref obj, needs to be decref'd later */
     Py_INCREF(obj);
 
     if (fontsize <= 1) {
@@ -791,6 +792,7 @@ font_init(PyFontObject *self, PyObject *args, PyObject *kwds)
     }
 
     if (obj == Py_None) {
+        /* default font */
         Py_DECREF(obj);
         obj = font_resource(font_defaultname);
         if (obj == NULL) {
@@ -809,6 +811,7 @@ font_init(PyFontObject *self, PyObject *args, PyObject *kwds)
     /* SDL accepts UTF8 */
     oencoded = pg_EncodeString(obj, "UTF8", NULL, NULL);
     if (!oencoded || oencoded == Py_None) {
+        /* got a file object, or an error */
         Py_XDECREF(oencoded);
         oencoded = NULL;
         PyErr_Clear();
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -305,7 +305,6 @@ font_setter_bold(PyObject *self, PyObject *value, void *closure)
     return 0;
 }
 
-
 /* Implements get_bold() */
 static PyObject *
 font_get_bold(PyObject *self, PyObject *args)
@@ -382,7 +381,6 @@ font_set_italic(PyObject *self, PyObject *args)
     Py_RETURN_NONE;
 }
 
-
 /* Implements getter for the underline attribute */
 static PyObject *
 font_getter_underline(PyObject *self, void *closure)
@@ -414,7 +412,6 @@ font_get_underline(PyObject *self, PyObject *args)
 {
     return _font_get_style_flag_as_py_bool(self, TTF_STYLE_UNDERLINE);
 }
-
 
 /* Implements set_underline(bool) */
 static PyObject *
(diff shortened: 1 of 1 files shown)


Starbuck5:
I tested this locally, and it fixes the issue, good job 👍

robertpfeiffer:
> Please ignore the formatting changes, my editor does that automatically :/

then use `git add -p` in the future, or run clang-format. Doing this automatically messes with merges and git blame, and makes it hard to see what was actually changed.

robertpfeiffer:
Can you explain what you did here and what the problem was a bit more?

ankith26:
Yea sure!
Basically I changed only a small part of the code in the constructor for the `Font` object, the old code used to look like this 
```c
if (obj == Py_None) {
    Py_DECREF(obj);
    obj = font_resource(font_defaultname);
    if (obj == NULL) {
        if (PyErr_Occurred() == NULL) {
            PyErr_Format(PyExc_RuntimeError,
                            "default font '%.1024s' not found",
                            font_defaultname);
        }
        goto error;
    }
    fontsize = (int)(fontsize * .6875);
    if (fontsize <= 1) {
        fontsize = 1;
    }

    oencoded = obj;
    Py_INCREF(oencoded);
    filename = Bytes_AS_STRING(oencoded);
} else {
    /* SDL accepts UTF8 */
    oencoded = pg_EncodeString(obj, "UTF8", NULL, NULL);
    if (!oencoded || oencoded == Py_None) {
        Py_XDECREF(oencoded);
        oencoded = NULL;
        PyErr_Clear();
        goto fileobject;
    }
    filename = Bytes_AS_STRING(oencoded);
}
```
The line `obj = font_resource(font_defaultname);` is supposed to load the default font by the name, into `PyObject`, `obj`, which is expected to be a bytes object. But for some reason, in the pyinstaller case, the function started returning a BytesIO object, a file-like object. Naturally, this would cause problems, because this case is not handled (but file-like objects are handled in the else block). So I simply made the else-part run on all cases (even when `obj` is None), so that all kinds of input from the font resource loading functions are handles normally

robertpfeiffer:
Are you sure the refcounting still works the way it should? Was the refcounting broken before? I think this needs to be commented/refactored a bit more, because font loading is very brittle.

ankith26:
In the bit of code I dealt with, I am pretty sure I did not break refcounts. But overall, font module is a bit complicated, and needs simplification. We also need more uniform default resource loading functions

Starbuck5:
> Py_INCREF(oencoded);

This is the suspect line about ref counting.

But it's weird, because why increase the ref count of oencoded right before it gets encoded into a filename?

ankith26:
I checked that, and yea, `oencoded` gets decref'd later in the function. This incref is required to compensate for that decref

illume (approved):
👍 thanks!

Starbuck5 (approved):
Alright, I finally feel that I understand this enough to approve it.

The condition that checks for file objects was only run when the argument was a non default font. This just changes it so it runs regardless.

PR #2606, opened 2021-05-24
Fixed pygame.examples.mask crashing

MightyJosip:
Thanks to discord user Nikorasu#2462 who pointed out that there is an error with this example that says 
```
File "pygame\examples\mask.py", line 175, in main
    pg.time.set_timer(pg.USEREVENT, 33)
pygame.error: pygame is not initialized
```
This simple fix should make the example work (tested locally)

### DIFF
--- a/examples/mask.py
+++ b/examples/mask.py
@@ -133,6 +133,7 @@ def main(*args):
     if len(args) == 0:
         raise ValueError("Require at least one image file name: non given")
     print("Press any key to quit")
+    pg.init()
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
LGTM! 👍

illume:
Sounds like this is a compatibility issue? I opened an issue for that here: https://github.com/pygame/pygame/issues/2612

illume (approved):
Thanks

PR #2607, opened 2021-05-25
Add python 3.6, 3.7 and 3.9 wheel build support for aarch64

odidev:
Add python 3.6, 3.7 and 3.9 wheel build support for aarch64
@illume, Could you please review this PR?

### DIFF
--- a/.github/workflows/manylinux.yml
+++ b/.github/workflows/manylinux.yml
@@ -49,6 +49,43 @@ jobs:
       with:
         path: dist/*.whl
 
+  build-aarch64:
+    name: "Build aarch64 wheel"
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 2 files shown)


ankith26:
Well, travis is shutting down soon (see #2300), maybe by the end of this month (4-5 more days left), and we have already migrated to github actions. So sadly, this PR will not work out.

Also, github actions too does not have aarch64 machines, so after travis closes, we'd have no setup for aarch64 testing

odidev:
@ankith26, On github actions, support for aarch64 can be added with the help of qemu. 
I have raised the PR for the same. Please have a look: https://github.com/pygame/pygame/pull/2610

ankith26:
Okay, so that means that this PR can be closed now, I will look into the other one 👍

PR #2608, opened 2021-05-25
Adjust shot explosion position in the aliens example

mljli:
Currently, when a shot hits an alien, explosion will be created at the position of the shot because `groupcollide(shots, aliens, 1, 1).keys()` returns `shots`. This PR swaps the position of `shots` and `aliens` which makes explosion looks more natural.

```py
# See if shots hit the aliens.
for alien in pg.sprite.groupcollide(shots, aliens, 1, 1).keys():
    if pg.mixer:
        boom_sound.play()
   Explosion(alien)
```

### DIFF
--- a/examples/aliens.py
+++ b/examples/aliens.py
@@ -378,7 +378,7 @@ def main(winstyle=0):
             player.kill()
 
         # See if shots hit the aliens.
-        for alien in pg.sprite.groupcollide(shots, aliens, 1, 1).keys():
+        for alien in pg.sprite.groupcollide(aliens, shots, 1, 1).keys():
             if pg.mixer:
                 boom_sound.play()
             Explosion(alien)


MightyJosip (approved):
Nice change, thanks

Starbuck5 (approved):
This makes it look cooler, confirmed.

mljli:
@Starbuck5 Why does the CI take so long? It builds fast on my laptop. And sorry that I didn't see #2300 before I created this PR. Should have used [skip ci] for this change.

ankith26:
The CI does not start because this is an examples change. I had configured the new CI to skip running on non-core changes, but I had hoped that it would pass, rather than halting like this. This is probably due to the fact that the builds are marked as "required" builds.

PR #2610, opened 2021-05-28
Add aarch64 wheel build support

odidev:
Added aarch64 wheel build support.
@illume, Could you please review this PR?

### DIFF
--- a/.github/workflows/manylinux.yml
+++ b/.github/workflows/manylinux.yml
@@ -35,10 +35,14 @@ jobs:
     steps:
     - uses: actions/checkout@v2.3.4
 
+    - name: Set up QEMU
+      id: qemu
+      uses: docker/setup-qemu-action@v1
+      
     - name: Build manylinux wheels
       run: |
         cd buildconfig/manylinux-build
-        make pull pull-manylinux wheels wheels-manylinux
+        make base-image-aarch64 pull pull-manylinux wheels wheels-manylinux
         cd ../..
         mkdir -p dist/
         cp buildconfig/manylinux-build/wheelhouse/*.whl dist/
--- a/buildconfig/manylinux-build/build-wheels.sh
+++ b/buildconfig/manylinux-build/build-wheels.sh
@@ -1,7 +1,11 @@
 #!/bin/bash
 set -e -x
 
-export SUPPORTED_PYTHONS="cp27-cp27mu cp35-cp35m cp36-cp36m cp37-cp37m cp38-cp38 cp39-cp39"
+if [ `uname -m` == "aarch64" ]; then
+   export SUPPORTED_PYTHONS="cp36-cp36m cp37-cp37m cp38-cp38 cp39-cp39" 
+else
+   export SUPPORTED_PYTHONS="cp27-cp27mu cp35-cp35m cp36-cp36m cp37-cp37m cp38-cp38 cp39-cp39" 
+fi
 
 if [[ "$1" == "buildpypy" ]]; then
 	export SUPPORTED_PYTHONS="pp27-pypy_73 pp36-pypy36_pp73 pp37-pypy37_pp73"
--- a/buildconfig/manylinux-build/docker_base/Dockerfile-aarch64
+++ b/buildconfig/manylinux-build/docker_base/Dockerfile-aarch64
@@ -1,0 +1,75 @@
+ARG BASE_IMAGE=manylinux2014_aarch64
+FROM quay.io/pypa/$BASE_IMAGE
+ENV MAKEFLAGS="-j 4"
+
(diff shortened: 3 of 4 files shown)


illume:
Hi ya. 

have a look at https://github.com/pypa/manylinux (I see you have already) and in the pygame repo at buildconfig/manylinux

It's quite well documented both on the pygame side, and with manylinux. Basically we'd need to create a pygame image using the scripts which download and compile libraries (like sdl_image, and libpng). There's a manylinux2014_arch64 image that can be used as a base. This is a standard that python uses for fairly compatible linux binaries. These are the ones that we'd upload to pypi, since they don't rely on a distro specific libraries and let us use more modern versions of the libraries (manylinux includes very old versions of the dependencies).

Additionally, check out the piwheels project. It pre-compiles many python binaries (including pygame) based on the raspberry pi OSen versions. The difference for this is that it uses the system libraries, and depends on them, rather than compiling and including the libraries. Both approaches have their benefits and downsides.

It would be nice to add manylinux2014 arm64 builds to the buildconfig/manylinux, and not to github actions or travis. Maybe from copy-pasting some of the existing config in there.

odidev:
@illume, Could you please have a look at this PR? Thanks in advance.

odidev:
@illume, Could you please have a look at this PR? Thanks in advance.

buildconfig/manylinux-build/build-wheels.sh
```diff
@@ -1,7 +1,11 @@
 #!/bin/bash
 set -e -x
 
-export SUPPORTED_PYTHONS="cp27-cp27mu cp35-cp35m cp36-cp36m cp37-cp37m cp38-cp38 cp39-cp39"
+if [ `uname -m` == "aarch64" ]; then
+   export SUPPORTED_PYTHONS="cp36-cp36m cp37-cp37m cp38-cp38 cp39-cp39"
```
illume, line 5: I guess because the manywheel images don't support python 2? It's something we can improve on later if anyone wants to contribute it.

buildconfig/manylinux-build/docker_base/Dockerfile-aarch64
```diff
@@ -0,0 +1,75 @@
+ARG BASE_IMAGE=manylinux2014_aarch64
```
illume, line 1: Why does this need a separate Dockerfile? The base image can be passed in as an argument, so I don't see what the need is for a separate file.

illume (approved):
👍 thanks 🎈 🎉

illume:
@odidev sorry for the delay in looking at it. Looks good to me :) Note, I didn't try the wheels themselves out on any arch64 arm boards.

I requested a couple of other devs to review it because we need two people to review each PR (for knowledge sharing, and to make sure people have a chance for their input).

.github/workflows/manylinux.yml
```diff
@@ -35,10 +35,14 @@ jobs:
     steps:
     - uses: actions/checkout@v2.3.4
 
+    - name: Set up QEMU
```
ankith26, line 38: We need to setup QEMU here only on the manylinux aarch64 build, not all builds

ankith26 (changes requested):
This PR looks good. I have not reviewed the docker stuff in great detail because I don't have much clue about it. But on the github actions end, we need to parallelize the builds. manylinux build has been running since about an hour now, and shows no sign of finishing. I have already done some parallelization work using github matix builds in #2605. 

I guess we should wait for that PR to be merged, then update this PR branch with main, and add a new matrix for this build

ankith26:
Hmmm the manylinux build is running since about 3 hours now. I guess something went really wrong with this PR, because ideally builds should not take more than 1 hour at the worst case. If it continues to run like this, we might have to force close it

illume:
It finished in the end...
> Successful in 188m
https://github.com/pygame/pygame/pull/2610/checks?check_run_id=3149579773

I'll try it again to see how long it takes.

.github/workflows/manylinux.yml
illume: In this workflow it is doing everything within the same job. There's not a separate one for it. (I think you had a separate PR for that maybe?)

illume:
I see... it's making the base image as part of the PR workflow. Whereas we usually make the base image once, and upload it. 
I'll merge this into another branch and make a few tweaks.

.github/workflows/manylinux.yml
ankith26: Yes, I had a PR for it to make matrix (parallel builds) for many linux, which I feel should be merged before this PR, and this PR can take add a new matrix for the aarch64 builds

Issue #2598, opened 2021-05-22
Some sphinx docs improvements


illume:
- Themes are supposed to import Sphinx's basic.css. This is a file that gets added to the generated docs by Sphinx, but we don't do anything with it. I think this would help maintainability in the future, as when Sphinx or docutils changes Sphinx will update their basic.css along with it. But that's beyond the scope of this PR.
- Sphinx 4 is out now. And it breaks more stuff 😢 (In fairness to Sphinx, it might be docutils that has changed). So far I've noticed that something is funky with the codeblocks that breaks the image positioning in the tutorials, when built on Sphinx 4.
- allowedchange in mixer.pre_init needs a versionedchanged note.

PR #2615, opened 2021-05-30
Some small changes I've been meaning to do

### LINKED ISSUE #2620: Small fix to document for mouse wheel event for Pygame 2.0+
When reading the document for mouse and mouse wheel my eyes are drawn to the title:

"Mouse Wheel Behavior in pygame 2" and so I didn't think anything above that line was relevent to me because I was using the most recent version of pygame.

What would have helped me out would have been to mention that the MOUSEBUTTON things described above that line were still happening even in the most recent version.  It took me a while to figure out that the bugs I was seeing were really not bugs at all.  :)

Thanks!  Having fun learning.  :)

### LINKED ISSUE #2633: Rotation direction is incorrect
Hi there, 

I was working on a collision detection/simultation system when i noticed it wasn't working correctly - all the angles seemed to be flipped in the wrong direction. A minimal working example is here:

```py
import pygame
import math

WIDTH = 800
HEIGHT = 800
FPS = 60


## initialize pygame and create window
pygame.init()
pygame.mixer.init()  ## For sound
screen = pygame.display.set_mode((WIDTH, HEIGHT))

pygame.display.set_caption("Name")
clock = pygame.time.Clock()     ## For syncing the FPS


## group all the sprites together for ease of update
all_sprites = pygame.sprite.Group()

## Game loop
running = True
while running:

    #1 Process input/events
    clock.tick(FPS)     ## will make the loop run at the same speed all the time
    for event in pygame.event.get():        # gets all the events which have occured till now and keeps tab of them.
        ## listening for the the X button at the top
        if event.type == pygame.QUIT:
            running = False


    #2 Update
    all_sprites.update()
    v = pygame.math.Vector2(1,0) * 50

    center = pygame.math.Vector2(WIDTH/2, HEIGHT/2)

    pygame.draw.line(screen, (255, 0, 0), center, center + v)
    pygame.draw.line(screen, (0, 255, 0), center, center + v.rotate_rad(math.tau/8))
    pygame.draw.line(screen, (0, 0, 255), center, center + v.rotate_rad(-math.tau/8))


    #3 Draw/render
    #screen.fill(pygame.color.THECOLORS['black'])

    all_sprites.draw(screen)
    ########################

    ### Put code here

    ########################

    ## Done after drawing everything to the screen
    pygame.display.flip()       

pygame.quit()
```

Which produces:
image

We can see that the green should be rotated counter clockwise (because that's what the [docs](https://www.pygame.org/docs/ref/math.html#pygame.math.Vector2.rotate_rad) say) by tau/8 with respect to the red vector, but instead it's rotated clockwise by that much. 

If you're not familiar with tau, just know that tau/8 is 45 degrees.

I am using `rotate_rad` but Salvak in the discord group saw the same behavior with `rotate`. I'm assuming this issue probably extends into all the rotation functions for vectors.

Here is a link to our discussion: https://discord.com/channels/772510943811731499/861874127315927041/862765375942950952

I am aware that one way to fix this is to simply change the docs - but I think rotation should be counter clockwise to adhere to the mathematical convention.

Starbuck5:
This is just a collection of small changes that would be hard pressed to fill a PR on their own.

- In `buildconfig/download_win_prebuilt.py`, I replaced a call to `distutils` with an equivalent function in `os`. (Because distutils is slowly being deprecated, and this function had an easy equivalent)
- In `mixer.rst`, I specified in `init()` and `pre_init()` that the allowechanges and devicename arguments are new in pygame 2. (As per #2598)
-  The recent commit fixing explosion location in the aliens example made me realize that the bomb was being drawn with a rectangular background behind it, so I just removed the background of `bomb.gif` in an image editor.
- I switched around the conditionals about SDL version in `_pygame.h`, so that VScode now shades out the SDL1 regions as unused code, rather than the SDL2 regions.
- In `__init__.py`, I put `Cursor` in the pygame namespace directly, like the other classes, and I fixed a theoretical problem where `mask` wouldn't be set to `MissingModule` when it failed to import.
- Adds documentation fixes requested: fixes #2633, fixes #2620


illume:
> The recent commit fixing explosion location in the aliens example made me realize that the bomb was being drawn with a rectangular background behind it, so I just removed the background of bomb.gif in an image editor.

Is this a compatibility issue/regression?

Starbuck5:
> Is this a compatibility issue/regression?

You would think so, but it's the same way in 1.9.6. And I checked the code and didn't see any colorkeying / paletting jazz.

buildconfig/download_win_prebuilt.py
```diff
@@ -117,10 +117,9 @@ def get_urls(x86=True, x64=True, sdl2=True):
 def download_prebuilts(temp_dir, x86=True, x64=True, sdl2=True):
     """ For downloading prebuilt dependencies.
     """
-    from distutils.dir_util import mkpath
     if not os.path.exists(temp_dir):
         print("Making dir :%s:" % temp_dir)
-        mkpath(temp_dir)
+        os.makedirs(temp_dir)
```
ankith26, line 122: There is a small behavioral change here, `mkpath` does not fail if directory exists, while `os.makedirs` can raise `FileExistsError` error. We need to do try-except handling for that (there is an `exist_ok` flag that can be passed to `os.makedirs` but that is not on py2, so we need to do the more verbose try-except handling)

ankith26 (changes requested):

buildconfig/download_win_prebuilt.py
Starbuck5: Good spot. But the os.makedirs is inside a check of whether the path exists, so it shouldn't be an issue. Also I think it's a bit funny that there's a `FileExistsError`. (Seems a bit niche).

ankith26: ohh yea my bad 😄 
I just saw the 2 line change diff and did not see the check for path exists

illume (approved):
Thanks. 🎉 🎈 

> The recent commit fixing explosion location in the aliens example made me realize that the bomb was being drawn with a rectangular background behind it, so I just removed the background of bomb.gif in an image editor.

Does this mean there is a bug in pygame? If so, there should be an issue made for it (labeled critical/backwards compat) rather than fix the image so that it works.

> I deleted the unused scrap_qnx code, since we don't appear to support qnx

A lot of people maintain pygame outside of here(it's packaged in hundreds of distros for example). It's difficult to know if anyone is still maintaining pygame on QNX externally or not, mainly because it's mostly used commercially. I suspect so. My guess is that there will be users for some time of it for a long time and know that there are some embedded GUIs on QNX still going using pygame. Probably they will keep using SDL1, and older versions of pygame however. Personally, I feel that keeping such code in is OK as long as it's not too much of a maintenance burden - because then it makes it ever so slightly easier for those maintainers to get a newer version of pygame working. Some developers/maintainers of these esoteric OS have provided valuable contributions to pygame and SDL in the past - so even if the developer base is tiny it's worthwhile to pygame-the-project just for that reason.

PR #2617, opened 2021-06-02
Import parent theme stylesheet for docs

Grimmys:
I added the import of Sphinx's basic.css in our stylesheet, and removed all our CSS rules that were coming from this parent stylesheet, as suggested in #2598.

This import broke the display of images next to a code block in the tutorial pages, so I added new rules to fix it.

I also made small changes on buildconfig/makeref.py file to be more PEP-8 compliant.

Importing the stylesheet of the base theme seems to fixed some issues, as you can see below (left part is the online version / right part is the docs after my changes).

add_scrollbar
responsive_code_block

About the first screenshot, an horizontal scroll bar is now added to any code block / text block (actually, any pre tag) if the text can't fill the box.

For the second screenshot, it was taken for a very small window width, so it is about responsive issue for code block with floating image.
This new version seems better but can still be improved I think.

Beside these things, many minor changes can be observed, such as more emphasis on method documentation title (bold), one more level for breadcrumbs, highlightement for keyword(s) in search results and for anchors  (on the first screenshot, I accessed the page by searching "bayer"). 
I think they are ok, but we can discuss about that.

I'm not very experienced with the pygame doc, so I may have missed some issues...

PS: This is my first PR on pygame, don't hesitate to tell me if I did anything wrong in the contributing process.


Starbuck5:
Thank you for tackling this!

PEP-8 wants a capitalized function name?

One thing I noticed is that modules that are classes have a different font size for the "class" component of their name.
theme1

I think sticking with the Sphinx default of bolded functions will look good, but I think the entire name needs to be bolded.
theme2

This doesn't look good in my opinion.

Also, if we're bolding function names, should we also bold the function names in the top link section?
This section:
theme3

I see you changed the content width from 600px to 700px, a subtle change. But it preserves the look and feel while making it slightly more modern / functional, so I'm pro.

I see the generated `pygame.css` has a couple css warnings: incomplete style, unrecognized argument and the like. I don't think you caused these, but it would be fit in this PR to fix them.

Grimmys:
> PEP-8 wants a capitalized function name?

I don't think so... See my changes, the function name **was** capitalized.

> One thing I noticed is that modules that are classes have a different font size for the "class" component of their name.
> theme1
> 
> I think sticking with the Sphinx default of bolded functions will look good, but I think the entire name needs to be bolded.
> theme2
> 
> This doesn't look good in my opinion.

Hmmmm I don't know. Visually it might be weird, but that makes a bit of sense: this section is under `pygame.draw`, so you should already be aware that `rect()` is a method of this module. The only important part of the title is the method name, not the common prefix.
I will try to check this in more detail, but maybe someone else can give his opinion about it too.

> Also, if we're bolding function names, should we also bold the function names in the top link section?
> This section:
> theme3

Good remark, I'm going to check it and see if we should do it.

> I see you changed the content width from 600px to 700px, a subtle change. But it preserves the look and feel while making it slightly more modern / functional, so I'm pro.

Actually, it was to fixed the example that we can see in the second screenshot I provided: the theme import style seems to have slightly increased the width of a codeblock, so it was required to increase the content width too for keeping the ball next to the code.
Modernize the style is not in the scope of this PR don't worry. 😄 

> I see the generated `pygame.css` has a couple css warnings: incomplete style, unrecognized argument and the like. I don't think you caused these, but it would be fit in this PR to fix them.

Ok, I will try to make a bigger clean up.
I removed the invalid properties/selectors with no properties I encountered but it was not my goal to be exhautive.
However, you are right, it would perfectly fit in this PR so I'm going to do it. 👍 

Thank you for your serious checking. 😃

Starbuck5:
> I don't think so... See my changes, the function name was capitalized.

Oh of course! I read the diff wrong, my bad.

When we resolve the font size difference in those specific places, and the function bolding, I think this PR will be ready to merge.

Grimmys:
And here we go, I think all is ok now.

Starbuck5 (approved):
Alright, this looks good to me!

I'm happy to our Sphinx theme based on Sphinx's own theme, which allows us to simplify `pygame.css_t`.

And I think the bolded function names look good.

Do you think the bolded function names at the top look too busy? I'm going back and forth on that now.

Grimmys:
Yes, it seems too much for me too. 

I think it is because of the combo bold + underline, a combo that should generally be avoid imo. 

So maybe we have to make a choice for the function names at the top: bold, or underline?

Starbuck5:
> I think it is because of the combo bold + underline, a combo that should generally be avoid imo.

Ah, good point.

Then lets stay with just underlined, since they are links.

illume (approved):
🎉 Thanks!

PR #2619, opened 2021-06-03
Added SVG to supported image formats

kaiec:
Since 2.0.2, SDL Image supports SVG, see here: https://www.libsdl.org/projects/SDL_image/

### DIFF
--- a/docs/reST/ref/image.rst
+++ b/docs/reST/ref/image.rst
@@ -20,40 +20,50 @@ supports any extended file formats. By default it can only load uncompressed
 ``BMP`` images. When built with full image support, the ``pygame.image.load()``
 function can support the following formats.
 
-   * ``JPG``
-
-   * ``PNG``
+   * ``BMP``
 
    * ``GIF`` (non-animated)
 
-   * ``BMP``
+   * ``JPEG``
+
+   * ``LBM`` (and ``PBM``, ``PGM``, ``PPM``)
 
    * ``PCX``
 
+   * ``PNG``
+   
+   * ``PNM``
+
+   * ``SVG`` (limited support, using Nano SVG)
+
    * ``TGA`` (uncompressed)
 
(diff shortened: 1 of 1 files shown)


Starbuck5:
Thanks for the catch! I didn't know we could load svg images now, that's cool.

I would appreciate it you could test out some of the other discrepancies between what we say and what SDL_Image says.

Looking at the SDL_Image website, I see there might be a few more discrepancies. Most importantly, I see that SDL_Image and therefore us support webp now, which is cool.

Also, it would be good to mark things new in pygame 2.0 as new explicitly, using a `..versionadded` tag in the docs source.

kaiec:
Ok, I added all formats, sorted the list alphabetically (like at the SDL Image page, which makes comparison easier). I merged the lines LBM and PBM, the additional formats are not listed on the SDL Image page, though. I also consistently use JPEG and TIFF, but extended the note that TIF and TIFF are the same. I am not very familiar with the internals of pygame and sdl, so please let me know if the discrepancies had a technical reason.

Starbuck5 (approved):
Alright, this looks good to me.

If you want to add a small note about font loading not working, feel free. Is text a custom thing in svg? It's not just the lines, like everything else?

kaiec:
Thanks. Let's keep it as is. Yes, font rendering is common in SVG, but for interoperability, you often have to convert text to paths. Maybe it even works on other systems or with other fonts. The hint with limited and Nano SVG should be ok to point everyone in the right direction, if problems arise.

ankith26:
This is sorta related to this PR, but when we are explicitly mentioning in the docs that we support those formats, it might be a good idea to also include in the unit tests, some test cases related to these new formats.
Maybe this could be done in a seperate PR, or if you are interested, you could do it in this PR itself 😄

illume (approved):
Thanks 👍 🎈

Issue #2618, opened 2021-06-03
Interrogations/Suggestions about docs package


Grimmys:
**Description**

After a fresh onboarding on the docs package to modify the rendering of the pygame documentation, I have a few remarks to make, which may or may not be relevant.

- [x] There is a file about the release process of pygame in this package (`docs/howto_release_pygame.txt`), I think this file would be better elsewhere since this does not seem to be about the release process for the docs in particular. 
- [x] `docs/__main__.py` file should display the generated docs (in local), and the online docs if there is none generated in local. But it is looking for the `index.html` file in the docs package while the docs are generated in a different sub-package (`docs/generated/index.html`), so it always start the online docs (https://www.pygame.org/docs/). I think `__main__.py` should look for `generated/index.html`.
- [x] For the onboarding of a new contributor like me, I found this page: [Generating docs](https://www.pygame.org/wiki/Hacking?parent=#Generating%20docs), but it looks hard to find it in my opinion. And there seems to be new contributions guidelines that will be added to the README (see #2539). I suggest to add too a more specific README in the docs package to explain how to build the docs, how to access them (via `docs/__main__.py`?), what guidelines should be followed for editing documentation content or theme (if any) and maybe add a link to the sphinx documentation.
- [x] Strangely, `pygame.css_t` file seems to not be "watched" by sphinx builder. If the docs are already built, and the only changes made are in this css file, the sphinx generation command will **not** take care of this changes. The only ways to apply these changes are to either remove the cache (i.e. the `docs/generated` folder) or made a irrevelant change on another file.
According to sphinx-build documentation ([sphinx-build](https://www.sphinx-doc.org/en/master/man/sphinx-build.html)), the `-E` option (and maybe the `-a` option too, not sure of its meaning) can solve this issue, but I'm not sure it is what we would like because it will totally ignores the cached files...

Starbuck5:
`docs/howto_release_pygame.txt` is more of a historical document. The real procedure is stored inside illume's head.

Recently, another contributor (dr0id) made the generated docs go into the `generated` folder so it would be more contained and easier to delete and regen. Both of us missed that the change requires changing `__main__.py`.

Adding a README.md to the docs folder sounds like a great idea!

It's surprising to me how you're unsatisfied with the css theme not being watched (expressed here and on discord). I had just accepted it as a fact of life. We could have an explicit `cleandocs` command that would regenerate with the `-E` option? We don't manually define what files Sphinx watches, and I don't know if there is a way to do that.

[...]

PR #2624, opened 2021-06-13
Docs improvements

Grimmys:
According to #2618, this PR makes the following changes:

- Remove historical `docs/howto_release_pygame.txt` file
- Fixed start up of local documentation via `docs/__main__.py`
- Add the optional flag `--fullgeneration` (alias `--f`) to the `docs` command to be able to fully rebuild the documentation
- Create a first version for the docs README

Regarding the optional flag, I added a new entry in `user_options` of the docs command, and a new attribute directly linked to this option, following the inner workings of setuptools/distutils.  
If the flag is present, `buildconfig/makeref.py` will be called with the arg `full_generation`, and the `-E` option will be added to sphinx-build command.
To sum up:
`python setup.py docs` -> unchanged behavior
`python setup.py docs --fullgeneration` -> call sphinx-build command with `-E` as an additional option

About the README, it's really a first draft.
Many other things should be added, I have not so much experience with the docs package and [Starbuck said he has some ideas to add](https://github.com/pygame/pygame/issues/2618#issuecomment-854572757).
Moreover, I'm not a native: it would be really great if a native takes the time to verify the grammar and to reformulate the weird sentences!


Starbuck5:
Sorry I haven't gotten to this PR in so long.

I like the changes!

The organization of the README.md strikes me as a little bit convoluted, I think changing the first main header to something like "Generating the docs" would help. Just since there is an "Access documentation locally" and then an "Access to documentation" header.

The only English mistake I found was "advice" should be "advise" in the contribution section.

But I think the contribution section should be changed:
- It needs a message encouraging people to contribute to the docs
- I think the "check out the sphinx docs" request is intimidating to newcomers. Maybe it could be phrased in a more informational way... like "pygame uses Sphinx to generate html documentation from restructured text (rst) files. Sphinx docs website is a good resource".

docs/README.md
```diff
@@ -0,0 +1,46 @@
+## Access documentation locally
```
illume, line 1: There are two audiences here:
- 1) people looking to contribute/develop/write docs
- 2) people trying to read pygame docs.

Doesn't have to be in this PR, but we probably also need something for people looking into the docs folder looking for pygame docs.

illume:
Note we have: https://www.pygame.org/wiki/Hacking

There's a docs section in there - feel free to take from that. Probably when this README is complete for contributing to docs, then we should link to it from the Hacking page (DRY) so we only have one to maintain.

docs/__main__.py
```diff
@@ -2,21 +2,24 @@
...
     path, last = os.path.split(path)
     if last:
         for p in iterpath(path):
             yield p
         yield last
 
+
 pkg_dir = os.path.dirname(os.path.abspath(__file__))
-main_page = os.path.join(pkg_dir, 'index.html')
+main_page = os.path.join(pkg_dir, 'generated/index.html')
```
illume, line 22: oops. Good catch.

illume (approved):
👍 thanks.

I personally think we should merge this and improve it in separate PRs.

Starbuck5 (changes requested):
docs/README.md isn't merge ready, in my opinion,

@Grimmys you could take it out of this PR and then we can push everything else through, or stay the course and do it all as one PR.

Grimmys:
Thanks for your reviews! 

I don't know if I should split this PR... 
In my opinion, I should fix the grammar and formulation issues, and it will be ready to be merge. 

The README doesn't have to be complete (everyone can have things to add) for this PR, but should be at least in a minimal "valid shape".

Grimmys:
I removed the commits about the README from this branch/PR, we should now be able to merge it.

docs/__main__.py
ankith26: minor code nitpick here, `os.path.join(pkg_dir, 'generated/index.html')` can also be `os.path.join(pkg_dir, 'generated', 'index.html')`

ankith26 (approved):
This PR LGTM! 🎈 🥳 
After the merge conflict is fixed, this PR should be good to go

docs/__main__.py
Grimmys: oh yeah you are right, definitively better to rely on the default directory separator provided by `os.path.join`, thanks!

Starbuck5 (approved):
Let's get this merged!

Just have to wait for the tests to run on the merge.

PR #2627, opened 2021-06-20
Remove restriction on rendering 4-byte characters on `pygame.font`

AvaxarXapaxa:
Some time ago, SDL_TTF gave support on rendering 4-byte characters (Which are emojis and some CJK characters). Although not documented in the official documentation, it was indeed implemented. It seems pygame hasn't removed this restriction on rendering 4-byte characters yet.

Test run:
image

### DIFF
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -514,12 +514,12 @@ font_render(PyObject *self, PyObject *args)
             return RAISE(PyExc_ValueError,
                          "A null character was found in the text");
         }
-#if !SDL_VERSION_ATLEAST(2, 0, 15)
+#if !SDL_TTF_VERSION_ATLEAST(2, 0, 15)
         if (utf_8_needs_UCS_4(astring)) {
             Py_DECREF(bytes);
             return RAISE(PyExc_UnicodeError,
                          "A Unicode character above '\\uFFFF' was found;"
-                         " not supported with SDL version below 2.0.15");
+                         " not supported with SDL_ttf version below 2.0.15");
         }
 #endif
         if (aa) {
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -514,12 +514,6 @@ font_render(PyObject *self, PyObject *args)
             return RAISE(PyExc_ValueError,
                          "A null character was found in the text");
         }
-        if (utf_8_needs_UCS_4(astring)) {
-            Py_DECREF(bytes);
-            return RAISE(PyExc_UnicodeError,
-                         "A Unicode character above '\\uFFFF' was found;"
-                         " not supported");
-        }
         if (aa) {
             if (bg_rgba_obj == NULL) {
                 surf = TTF_RenderUTF8_Blended(font, astring, foreg);
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -514,6 +514,14 @@ font_render(PyObject *self, PyObject *args)
             return RAISE(PyExc_ValueError,
                          "A null character was found in the text");
         }
+#if !SDL_VERSION_ATLEAST(2, 0, 15)
+        if (utf_8_needs_UCS_4(astring)) {
+            Py_DECREF(bytes);
+            return RAISE(PyExc_UnicodeError,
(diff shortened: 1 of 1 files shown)


AvaxarXapaxa:
> Some time ago, SDL_TTF gave support on rendering 4-byte characters

Probably should give some sources, [written here](https://github.com/libsdl-org/SDL_ttf/blob/main/CHANGES.txt):

> 2.0.15:
> Sam Lantinga - Fri Oct 26 13:26:54 PDT 2018
>  * Updated to FreeType version 2.9.1
> Sam Lantinga - Sun Sep 10 00:18:45 PDT 2017
>  * Text rendering functions now use the alpha component of the text colors
> Sam Lantinga - Sat Sep  9 22:21:55 PDT 2017
>  * ***Added support for characters greater than 0xFFFF (e.g. emoji) in the UTF-8 APIs***

src_c/font.c
```diff
@@ -514,12 +514,6 @@ font_render(PyObject *self, PyObject *args)
             return RAISE(PyExc_ValueError,
                          "A null character was found in the text");
         }
-        if (utf_8_needs_UCS_4(astring)) {
```
ankith26, line 517: This PR is looking good! Having `pygame.font` being able to handle more unicode is a good thing. Now for review, rather than completely removing this block, it would be better if you put some handling for SDL_TTF < 2.0.15 (maybe keep displaying this error message in that case). Also, another minor code nitpick, `utf_8_needs_UCS_4` function defined in this file is not used anywhere now, and can be removed. This change needs docs.

Also, the `ftfont` module will become irrelevant after this change, so we have to see what to do about it

AvaxarXapaxa:
Oh shoot, I pressed "Conversation Resolved" accidentally @ankith26

AvaxarXapaxa:
oh nvm. Hope that didn't mess up anything.

src_c/font.c
AvaxarXapaxa: Okay, I made a preprocessor condition which only "removes" the UCS-4 restriction if SDL_ttf is [above] 2.0.15

AvaxarXapaxa: Anyway, I will mark this convo as resolved. Since this change only take part on SDL_ttf >= 2.0.15, which pygame that uses SDL_ttf versions below 2.0.15 might still want to use `utf_8_needs_UCS_4` and `ftfont` for reverse compatibility sake (Which I guess is a common theme with illume keeping Python 2, kek). You can unresolve this convo if you have something to add.

AvaxarXapaxa:
It seems the unittest failed in `pygame/blob/main/test/font_test.py` line 390. I'm not sure on how to handle that, especially with the SDL_ttf version check macro I put.

Starbuck5:
Thanks for working on this Avaxar!

I think the best way to handle this would be to set the "ucs4" attribute to `True` on the font objects when they are created (given the version supports it, of course).

This would support an "internal API" put out by `ftfont`, which is an implementation of the `Font` API on top of freetype.

See ftfont.py: https://github.com/pygame/pygame/blob/main/src_py/ftfont.py#L52

This would prevent the failing test bit from running.

illume:
It would be fine with me if the test was simply commented out for now :)

```
        # If the font module is SDL_ttf based, then it can only supports  UCS-2;
        # it will raise an exception for an out-of-range UCS-4 code point.
        if UCS_4 and not hasattr(f, "ucs4"):
            ucs_2 = as_unicode(r"\uFFEE")
            s = f.render(ucs_2, False, [0, 0, 0], [255, 255, 255])
            ucs_4 = as_unicode(r"\U00010000")
            self.assertRaises(
                UnicodeError, f.render, ucs_4, False, [0, 0, 0], [255, 255, 255]
            )
```

Perhaps change the comment to this though:
>         # If the font module is SDL_ttf < 2.0.15 based, then it only supports UCS-2;

Maybe add an issue or note about adding a ucs4 attribute... but that's not required for this PR IMHO.

illume (approved):
👍 thanks

PR #2634, opened 2021-07-14
Close a ``subprocess.Popen`` automatically by using a context (and other pylint 2.9 lint fixes)

Pierre-Sassoulas:
The main thing is the with used with subprocess, which help make sure the subprocess is closed, I included a number of small fixes too as to not open 5 obnoxious merge requests.

### DIFF
--- a/src_py/_camera_opencv_highgui.py
+++ b/src_py/_camera_opencv_highgui.py
@@ -7,15 +7,16 @@ import opencv
 from opencv import highgui
 
 
-
 def list_cameras():
     """
     """
     # -1 for opencv means get any of them.
     return [-1]
 
+
 def init():
     pass
+
 
 def quit():
     pass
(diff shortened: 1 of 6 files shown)


Pierre-Sassoulas:
Turn out popen is not a context manager in all target python version, so it was a false positive.

ankith26 (approved):
LGTM! Thanks 👍 🥳

Starbuck5 (approved):
Looks good to me as well. I hope we will actually be able to merge, now that Travis is down (it's still marked as required).

Pierre-Sassoulas:
If you have no more check to do on Travis, you can remove Travis from the required status check in Github's settings / branches / then click "edit" on the "main" branch's protection rules. I had this problem when removing Travis and Appveyor a month or so ago. I had to do 10 mn of research right now for this comment even though I knew it was possible to do, and approximately where 😄

PR #2636, opened 2021-07-14
Add paths to /opt/homebrew

Muxelmann:
I had the issue that Font and Sound would not load on my M1 Mac, despite having installed all sdl2 libraries using homebrew. It appears that by adding the paths to where home-brew installs the libraries to the config file fixes the issue. At least, when I now install Pygame using:

```
pip install git+https://github.com/Muxelmann/pygame.git@patch-1
```

the following lines no longer raise exceptions/errors:

- `pygame.mixer.Sound(...)` used to result in the error `cannot import name 'Font' from partially initialized module 'pygame.font' (most likely due to a circular import)`
- `pygame.font.Font(...)` used to result in the error `mixer module not available (ModuleNotFoundError: No module named 'pygame.mixer')`
- `pygame.image.load(...)` used to result in the error `File is not a Windows BMP file`

### DIFF
--- a/buildconfig/config_darwin.py
+++ b/buildconfig/config_darwin.py
@@ -158,11 +158,11 @@ def main(sdl2=False):
     ])
 
     print ('Hunting dependencies...')
-    incdirs = ['/usr/local/include']
+    incdirs = ['/usr/local/include', '/opt/homebrew/include']
     if sdl2:
-        incdirs.append('/usr/local/include/SDL2')
+        incdirs.extend(['/usr/local/include/SDL2', '/opt/homebrew/include/SDL2'])
     else:
-        incdirs.append('/usr/local/include/SDL')
+        incdirs.extend(['/usr/local/include/SDL', '/opt/homebrew/include/SDL'])
 
     incdirs.extend([
        #'/usr/X11/include',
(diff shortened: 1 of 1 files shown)


Starbuck5:
Thanks for contributing!

I was wondering why this is, so I found this reddit thread that explains the problem: https://www.reddit.com/r/MacOS/comments/jw9guu/why_did_homebrew_move_from_usrlocalto_opthomebrew/

With that info, would this PR mess with anything else?

Theoretically it would still check /usr/local first, and then /opt?

Universal2 python vs emulated x86 python?

I can test this on my old 10.12 intel Mac and make sure it doesn't break that, at least.

Muxelmann:
> I can test this on my old 10.12 intel Mac and make sure it doesn't break that, at least.

Good idea! I also installed this PR on my old (2014) intel MacBook running 11.2.3 and can report that both the install and the `aliens.py` example worked fine.

Starbuck5 (approved):
Alright, I tested this locally and it didn't change my install process on an older MacOS.

I'm not much of an expert on "mac stuff," but this PR seems very straightforward. Homebrew changed their install location, so we should add that new location to be searched.

Does this PR close #2500 / replace #2508?

Muxelmann:
> Does this PR close #2500 / replace #2508?

I had similar issues in that the font (and image and sound) modules were not properly loaded / found on my M1 Mac. So I guess - since loading any of the modules works fine for me after adding the changed Homebrew library path - that the `NotImplementedError` for all these modules is addressed and the PRs may be closed, too.

illume (approved):
👍 Thanks! 🎉 🎈

redrover9:
Thank you to all concerned. This fixed it for me as well.

PR #2643, opened 2021-07-18
added pg.quit() to make examples IDLE friendly

AnudeepGunukula:
Previously  the pygame window will hang around until you close the IDLE output tab as well.

I added pg.quit() in necessary places  to make all examples idle friendly
 I have tested in both python IDLE and visual studio , they are working fine now.

The freezing of window is now fixed.


Starbuck5 (changes requested):
Hi, thanks for doing this.

In `sound_array_demos.py` and `sound.py` is there a need to explicitly quit the mixer? pygame.quit() would do that.

This is kinda nitpicky, but I think the `#quit` comments should be removed. It's the same as what the line below says.

Also, in the past we've used the `black` autoformatter to format examples. This PR would be a good opportunity to make sure we're up to par on style. Just pip install black and use it on the examples folder. `black examples` in a terminal.

AnudeepGunukula:
Hi @Starbuck5 , ok i will start working on it

Starbuck5 (approved):
Alright, thanks. I'm happy with this now.

AnudeepGunukula:
Thanks, @Starbuck5 for approving.

examples/music_drop_fade.py
```diff
@@ -170,7 +170,8 @@ def main():
     pg.scrap.init()
     pg.SCRAP_TEXT = pg.scrap.get_types()[0]  # TODO remove when scrap module is fixed
     clipped = pg.scrap.get(pg.SCRAP_TEXT).decode(
-        "UTF-8")  # store the current text from the clipboard TODO remove decode
+        "UTF-8"
+    )  # store the current text from the clipboard TODO remove decode
```
ankith26, line 174: minor code cleanliness nitpick, the comment can go on a new line, so that the line of code stays on the same line

AnudeepGunukula:
Hi @ankith26 ,thanks for reviewing . 
i have made the changes, comment goes on next line and code remains on same line in music_drop_fade.py

examples/music_drop_fade.py
```diff
@@ -170,7 +170,8 @@ def main():
     pg.scrap.init()
     pg.SCRAP_TEXT = pg.scrap.get_types()[0]  # TODO remove when scrap module is fixed
     clipped = pg.scrap.get(pg.SCRAP_TEXT).decode(
-        "UTF-8")  # store the current text from the clipboard TODO remove decode
+        "UTF-8")
```
ankith26, line 173: Another small nitpick, now that the comment is on the new line, `clipped = pg.scrap.get(pg.SCRAP_TEXT).decode("UTF-8")` can probably be on a single line rather than being broken like this

AnudeepGunukula: Hi @ankith26 , yeah now i got it completely and made an another commit after updating it
please check now is it fine?

ankith26: Yes, it is looking good now

ankith26 (approved):
This PR looks good to me! 🎉 
Thanks for your contribution 👍

AnudeepGunukula:
Thanks for reviewing guys, this is my first contribution . i am very happy

AnudeepGunukula:
Hi @Starbuck5  please merge this pr, as it pass the reviews and checks.

Starbuck5:
We're waiting a bit because another contributor said there might be a problem with this. I think it looks fine, but maybe I missed something.

AnudeepGunukula:
ok @Starbuck5 i understand ,
i have tested each example running in python idle and made the changes.
 after re-reviewing again, 
if there is any changes needed then please mention them , i will try to fix them.

illume (approved):
👍 thanks.

examples/setmodescale.py
```diff
@@ -17,7 +17,7 @@
 FPS = 30
 clock = pg.time.Clock()
 
-print ("desktops", pg.display.get_desktop_sizes())
+print("desktops", pg.display.get_desktop_sizes())
```
robertpfeiffer, line 20: Sorry I just got around to this @Starbuck5. This line breaks py2.7 compat. Are we officially dropping it?

examples/setmodescale.py
```diff
@@ -29,8 +29,13 @@
 
 r_name, r_flags = pg.display._get_renderer_info()
 print("renderer:", r_name, "flags:", bin(r_flags))
```
robertpfeiffer, line 31: Shoot, here I am not having py2.7 compat myself...

examples/vgrade.py
```diff
@@ -95,6 +95,8 @@ def main():
         elif event.type == pg.USEREVENT:
             DisplayGradient(screen)
 
+    pg.quit()
```
robertpfeiffer, line 98: Should most of these examples just be try-finally constructions around the call to `main()`?

examples/setmodescale.py
Starbuck5: I'll put in a PR that fixes this by doing `from __future__ import print_function`

examples/vgrade.py
Starbuck5: Not sure what you mean...

```
try:
    main()
finally:
    pg.quit()
```
So that way, if it crashes halfway through it's guaranteed to call pg.quit()?

examples/setmodescale.py
robertpfeiffer: Yeah, that would fix it, too. This was the one thing I noticed going through the diffs that I told you about. Not a blocker though :)

examples/vgrade.py
robertpfeiffer: Is consistency here good in general? Is this a pattern we want to promote? As far as I can tell, the PR adds pg.quit() in the correct place, which is sometimes after the call to main, sometimes at the end of main, sometimes at the end of a script with no main, and sometimes right before raising sysexit. It works, but it isn't pretty.

Issue #2642, opened 2021-07-17
Typing is too strict for AbstractGroup.add()


nathanielconroy:
The type annotation here is Sprite, but this method also accepts objects of type Group as well in practice. Could we change the type to Union[Sprite, Group] so that this doesn't produce type checker warnings?

https://github.com/pygame/pygame/blob/551ee32106787a503fe30df7e27bfabc4e153b2b/buildconfig/pygame-stubs/sprite.pyi#L55

Starbuck5:
The Group documentation says

> Each sprite argument can also be a iterator containing Sprites.

So it should be a Union[Sprite, sequence[Sprite], AbstractGroup] right? (Not sure if that is the correct syntax for a sequence of sprites).

nathanielconroy:
@Starbuck5 I would suggest Iterable rather than sequence, so `Union[Sprite, Iterable[Sprite], AbstractGroup]`

The only problem with this is that technically an Iterable passed in here could legally contain AbstractGroups or other Iterables as well, and I don't think there's any way to perfectly capture all of the possibilities here. For example, you could pass something like `[some_sprite, some_group, [another_sprite, another_group, [yet_another_sprite, and_another_sprite]]]` and I believe the `add` method could handle this just fine, but the type of the whole thing would be `Iterable[Union[Sprite, AbstractGroup, Iterable[Union[Sprite, AbstractGroup, Iterable[Sprite]]]]]` and even that doesn't exhaust the possibilities.

I don't think we're going to be able to fully capture all possibilities in the type hint.

Something like `Union[Sprite, AbstractGroup, Iterable]` might be the best compromise. It isn't perfect, since it wouldn't warn if someone tried to pass in an iterable of some other type, but I think not strict enough is better than too strict. I could see `Union[Sprite, AbstractGroup, Iterable[Union[Sprite, AbstractGroup, Iterable]]]` being ok as well. That would protect us against something like `[1, 2, 3]` even if it doesn't protect us against something like `[[1, 2, 3], [4, 5, 6]]`.

The actual contract is something like "sprites, groups or iterables that may be nested but must contain sprites or groups as their leaf elements" but I don't think there's a way to exhaustively define this contract in a type annotation.

Starbuck5:
Good points all.

I checked the source and you're right, it recursively adds the contents of iterators. But that seems like more of an implementation quirk than defined / intended behavior.

[...]

PR #2646, opened 2021-07-21
less strict type checking for AbstractGroup.add

nathanielconroy:
For https://github.com/pygame/pygame/issues/2642


illume (approved):
👍 Thanks! 🎈

nathanielconroy:
@illume thanks for the review!

MightyJosip (changes requested):
OK, after checking your https://github.com/pygame/pygame/issues/2642#issuecomment-883827605 I can agree solution without allowing type hints to accept classes that have add_internal method is correct. However I would just require a small extra change. You didn't cover the 
`group.add((group, group))`
 possibility. So I would go with something like
`def add(self, *sprites: Union[Sprite, AbstractGroup, Iterable[Union[Sprite, AbstractGroup]]]) -> None: ...`
Then I think it is ok.

And thank for your work on this 👍

nathanielconroy:
@MightyJosip , I agree. I amended my change. Thanks!

ankith26 (approved):
LGTM! 👍 🥳 
Thanks for the PR!

MightyJosip (approved):
👍

PR #2651, opened 2021-07-28
Update chimp example, other examples fixes

Starbuck5:
Updated chimp example:
- Upped the window size to something more reasonable
- made it use SCALED to it will even be reasonable on 4k+ displays
- Replaced `0`s and `1`s with `True` and `False` when appropriate
- New pixelart chimp, rather than an aliased downscale. (I actually found the original image chimp.bmp came from, and made the pixelart from that)
- New arm picture, also less aliased.

Other fixes:
- ran black on everything
- moved incorrect `pg.quit()` location in the aliens example
- added `from __future__ import print_function` to setmodescale to fix a py2.7 incompatibility noticed by @robertpfeiffer


ankith26 (approved):
LGTM! 👍

ankith26:
Also, another minor thing, maybe more example files can use `from __future__ import print_function`, seeing that there are many examples that use print function. OFC, we should not be bothering with py2 compat that much anyways, because it is gonna be removed soon(tm)

Starbuck5:
A lot of prints already work in python2. For example, `print("hello", "world")` works. But if it's getting a value from a function call, it doesn't. So rather than go through exhaustively evaluating every print, I just chose to hit the ones that have been already pointed out.

Starbuck5:
I just realized there is a ChimpByLine example that I will need to revamp with this as well.

illume (approved):
👍 thanks

PR #2654, opened 2021-07-29
De-emphasize HWSURFACE

Starbuck5:
This PR:
- Marks display flags ``HWSURFACE``, ``DOUBLEBUF``, and ``ASYNCBLIT`` obsolete on pygame 2. ([proof that they actually are](https://github.com/pygame/pygame/blob/main/src_c/_pygame.h#L58-L77))
- Removes several small references to them in other documentation / tutorials.
- ~Removes the ``SW`` in the surface repr, because all surfaces are software surfaces now.~ (The tests would have to get changed, and then it would be SDL1/SDL2 conditional, and that's too much trouble)
- Realize `transform_doc.h` change that was missed by my transform keywords PR 
- Remove Travis build from README, switch to "commits since 2.0.1" rather than "since 2.0.0".

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -148,26 +148,18 @@ required).
 
    The flags argument controls which type of display you want. There are
    several to choose from, and you can even combine multiple types using the
-   bitwise or operator, (the pipe "|" character). If you pass ``0`` or no flags
-   argument it will default to a software driven window. Here are the display
+   bitwise or operator, (the pipe "|" character). Here are the display
    flags you will want to choose from:
 
    ::
 
       pygame.FULLSCREEN    create a fullscreen display
-      pygame.DOUBLEBUF     recommended for HWSURFACE or OPENGL
-      pygame.HWSURFACE     hardware accelerated, only in FULLSCREEN
+      pygame.DOUBLEBUF     (obsolete in pygame 2) recommended for HWSURFACE or OPENGL
+      pygame.HWSURFACE     (obsolete in pygame 2) hardware accelerated, only in FULLSCREEN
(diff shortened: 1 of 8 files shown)


ankith26 (approved):
LGTM! 🥳

docs/reST/ref/display.rst
```diff
@@ -148,26 +148,18 @@ required).
...
-      pygame.DOUBLEBUF     recommended for HWSURFACE or OPENGL
-      pygame.HWSURFACE     hardware accelerated, only in FULLSCREEN
+      pygame.DOUBLEBUF     (obsolete in pygame 2) recommended for HWSURFACE or OPENGL
+      pygame.HWSURFACE     (obsolete in pygame 2) hardware accelerated, only in FULLSCREEN
       pygame.OPENGL        create an OpenGL-renderable display
       pygame.RESIZABLE     display window should be sizeable
       pygame.NOFRAME       display window will have no border or controls
-
-
-   Pygame 2 has the following additional flags available.
```
illume, line 165: I think we still need to say which flags are new in pygame 2. (We're still at a point for some people where they want their games to run on pygame 2 and 1, and pygame 1 is still more widely installable)

illume (approved):
🎉 🎈 thanks.

(I left one note for your consideration. Feel free to merge.)

docs/reST/ref/display.rst
Starbuck5: Yes, definitely! I knew I considered that but I had forgotten how. There was pre existing ..versionchanged that enumerates which ones are new in pygame 2, I just though the two tables would look much prettier as one 😄

PR #2655, opened 2021-07-30
Improve font_viewer example script

njbradley:
Hello! Here I fixed a couple bugs / inconsistencies in the font_viewer example script. This was recommended to me by Starbuck5 as a good first contribution, hopefully it is all good! 

One thing I changed was instead of the previous scrolling controls, where you would click once to enter "scroll mode" and your mouse would disappear, I changed it so it is click and drag. This is really similar in code and I thought it was more intuitive. I also made it so that the scroll wheel scrolls the page too.

I also made it so that textboxes off the screen are not rendered, and fixed a bug where the text would disappear before it actually left the screen.

Let me know if anything should be changed or if I forgot something!

### DIFF
--- a/examples/font_viewer.py
+++ b/examples/font_viewer.py
@@ -246,10 +246,14 @@ class FontViewer:
                     return False
             elif e.type == pg.MOUSEWHEEL:
                 self.y_offset += e.y * self.MOUSE_SCROLL_SPEED * -1
-            elif e.type == pg.MOUSEBUTTONDOWN or e.type == pg.MOUSEBUTTONUP:
-                # enter or exit virtual mouse mode for click and drag
-                self.grabbed = not self.grabbed
-                pg.event.set_grab(self.grabbed)
+            elif e.type == pg.MOUSEBUTTONDOWN:
+                # enter dragging mode on mouse down
+                self.grabbed = True
+                pg.event.set_grab(True)
+            elif e.type == pg.MOUSEBUTTONUP:
+                # exit drag mode on mouse up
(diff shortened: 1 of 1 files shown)


Starbuck5 (approved):
Looks good to me.

Thanks for the contribution!

examples/font_viewer.py
```diff
@@ -241,11 +244,12 @@ def handle_events(self):
             elif e.type == pg.KEYDOWN:
                 if e.key == pg.K_ESCAPE:
                     return False
-            elif e.type == pg.MOUSEBUTTONDOWN:
-                # enter or exit virtual mouse mode for scrolling
+            elif e.type == pg.MOUSEWHEEL:
+                self.y_offset += e.y * self.MOUSE_SCROLL_SPEED * -1
+            elif e.type == pg.MOUSEBUTTONDOWN or e.type == pg.MOUSEBUTTONUP:
+                # enter or exit virtual mouse mode for click and drag
```
robertpfeiffer, line 250: It's a minor thing, but I'd probably handle up and own separately. Since this is an example script, I think it's a good idea to set a good example.

ankith26 (approved):
LGTM! 🥳 
 After you fix that review by @ robertpfeiffer, this PR should be good to go! Thanks for the first PR! 👍

examples/font_viewer.py
njbradley: Ah I see, that makes sense! I separated them out, hopefully that is what you were thinking of.

Issue #2649, opened 2021-07-26
Test newly supported image formats


Starbuck5:
Recently, new contributor kaiec added the new image formats pygame.image can load (because of the update to SDL2). (https://github.com/pygame/pygame/pull/2619)

However, we don't have tests for this, unlike the old image formats.

Relevant code is `test_loaded_extended` in `image_test.py`, I believe.

duthades:
Hi @Starbuck5, I would like to work on this issue. Can you assign me?

PR #2656, opened 2021-07-31
Add tests for svg, pnm and webp image format

duthades:
Resolves: https://github.com/pygame/pygame/issues/2649
#### Description:
 - Add new image format files in `examples/data/`
 - Add test for `svg`, `xcf`, `pnm` and `webp` image format in `image_load_extended` in `image_test.py`
 - Fix extracting format from filename


Starbuck5 (changes requested):
Thanks for this.

The other PR documents support for lots of new formats, not just SVG. For example, WEBP support.

It would be good to test all new formats.

duthades:
> Thanks for this.
> 
> The other PR documents support for lots of new formats, not just SVG. For example, WEBP support.
> 
> It would be good to test all new formats.

Oh I see. I will add tests for all new formats.

duthades:
@Starbuck5 I have added tests for svg, xcf, pnm and webp image formats. I was not able to test xv format. It was giving me `pygame.error: Unsupported image format` error.

Starbuck5:
Cool, cool.

`xv` files seem really old and obsolete now, so we can drop that easily. It would be good to fix the documentation for it though, in this PR.

```
FAIL: test_load_extended (__main__.ImageModuleTest) [Test loading a xcf] (filename='examples/data/aquamarine.xcf', expected_color=(127, 255, 212, 255))
can load different format images.
  File "C:\Users\Starbuck\Downloads\pygame-add_svg_load_test\pygame-add_svg_load_test\test\image_test.py", line 862, in test_load_extended
    self.assertEqual(surf.get_at((0, 0)), expected_color)
AssertionError: (0, 0, 0, 0) != (127, 255, 212, 255)

Ran 33 tests in 1.142s

FAILED (failures=1)
```
On my system at least, loading xcf files fails I guess. Maybe you have to have Gimp installed? If so, that's weird.

One other thing I forgot to mention: We still support building with SDL1, so this PR needs to only run the new tests if running on SDL2. You can do this easily with `pygame.get_sdl_version()[0] >= 2`

Starbuck5 (changes requested):
Remaining items for this PR:

- [ ] Loading xcf files fails for me. What's up with that?
- [ ] These image formats tests should only run on SDL2 based pygame, since we still technically support building with SDL1.

illume:
Seems we have to manually approve github actions to run for the first time.

These tests are great, they've found some bugs on some platforms it seems :)

duthades:
> Remaining items for this PR:
> 
> * [ ]  Loading xcf files fails for me. What's up with that?
> * [ ]  These image formats tests should only run on SDL2 based pygame, since we still technically support building with SDL1.

Sorry for delay in reply. I made the changes but forgot to reply. Now the new tests run only for SDL2. I am not sure what is happening with xcf files. xcf test is working on my system. Gimp was indeed installed. But later I uninstalled it, still the xcf test were passing.

Starbuck5:
> Sorry for delay in reply. I made the changes but forgot to reply. Now the new tests run only for SDL2. I am not sure what is happening with xcf files. xcf test is working on my system. Gimp was indeed installed. But later I uninstalled it, still the xcf test were passing.

Ah, cool. The tests are failing on all our CI builds though. I wonder if xcf is the reason it's failing, or another image format.

Also I found this little bit in an SDL_Image documentation PDF
Capture

@duthades, how about you drop the xcf test and see if everything passes then. If that's it, we can just remove that from the docs. If it's something that's frequently unreliable, there's no use saying we support it.

duthades:
@Starbuck5 Yeah make sense. I removed xcf from tests and docs. All tests are passing now.

ankith26 (approved):
LGTM! 👍 
Thank you for your first PR! 🥳 🎈

Starbuck5 (approved):
Looks good to me as well. Thanks @duthades!

PR #2659, opened 2021-08-03
Docs readme

Grimmys:
This PR is the result of the split of this one: https://github.com/pygame/pygame/pull/2624,
concerning the README part as stated on the issue #2618.

This is only a first draft, and @Starbuck5 should add other things to it.


Grimmys:
#2667 covers this

PR #2661, opened 2021-08-04
Check SDL version earlier, report known incompat (closes #2657)

### LINKED ISSUE #2657: pygame built against SDL 2.0.14 is incompatible with older SDL2 versions
Hi, I'm opening this issue to collate some information discovered while helping a user in [#python](https://www.pound-python.org) on the Libera IRC network. ~**I do not believe this to be a bug in `pygame`, this issue is meant to assist others the same problem.**~ My initial belief was wrong, this issue represents a flaw in error reporting, see remarks below.

---

### The problem
The user is running Raspbian Lite (`buster`) on a Raspberry Pi, and ran into a problem with an undefined SDL symbol `SDL_RWclose` when running the examples:

```
pi@raspberrypi:~/pygameLearn $ python3.9 -m pygame.examples.alien
ImportError: /home/pi/.local/lib/python3.9/site-packages/pygame/rwobject.cpython-39-arm-linux-gnueabihf.so: undefined symbol: SDL_RWclose
```

As I understand it, the problem comes from [a change introduced in SDL 2.0.14](https://github.com/libsdl-org/SDL/commit/316ff3847b4d9d87d9b0aab15321461db0e8ae0b) which changes `SDL_RWclose`[1] from a macro into a function, creating a new exported symbol.

After having the user re-install with verbosity turned up[2], I was able to see that the package was coming [from `piwheels`](https://www.piwheels.org/project/pygame/), and apparently they build `pygame` for Python 3.9 against `bullseye`, the next Debian release after `buster`, which [provides libsdl2 2.0.14](https://packages.debian.org/bullseye/libsdl2-2.0-0). **The resulting wheel is not compatible** with the version of SDL [provided in `buster` (2.0.9)](https://packages.debian.org/buster/libsdl2-2.0-0), leading to the error seen above.

[1] used in  [`src_c/image.c`](https://github.com/pygame/pygame/blob/6746053d1cda4c194a617e696b6eca06414e80b7/src_c/image.c), [`src_c/imageext.c`](https://github.com/pygame/pygame/blob/6746053d1cda4c194a617e696b6eca06414e80b7/src_c/imageext.c), and [`src_c/rwobject.c`](https://github.com/pygame/pygame/blob/39c15bffa052237534d6559faa4ee08b7f1f06a9/src_c/rwobject.c)
[2] `python3.9 -m pip install --force-reinstall --log install.log -vvv pygame`

---

### Resolution

The user decided to use Python 3.7 for their `pygame` project. If a user wants to use `pygame` with Python 3.9 on this platform, they will either need to find another source for a binary wheel, install `libsdl2-2.0-0` from `testing` (`bullseye`), or [build `pygame` from source](https://www.pygame.org/wiki/CompileUbuntu?parent=#Python%203.x). To build the latest `pygame` compatible with their Python, they can also install the packages from the previous link and run `python3.9 -m pip install --no-binary=pygame pygame` (if I understand `pygame`'s build process correctly, that is!)

SnoopJeDi:
pygame built against SDL 2.0.14 is known to be incompatible with runtime 2.0.9. These changes check for this edge case and raise an appropriate exception.


src_c/base.c
```diff
@@ -2295,6 +2302,11 @@ MODINIT_DEFINE(base)
 #endif
 #endif
     }
+
+    if (!pg_CheckSDLVersions()) {
```
ankith26, line 2306: Can this check come somewhere top in the module init code? That would help to error faster

robertpfeiffer:
This checking should probably be a bit more general than this. You can always use a newer version of SDL2 than compiled, but if you use an older one, it will break. Either we merge this and I whip something more future-proof up later, or we do it here.

Also, how does this interact with: https://github.com/libsdl-org/SDL/blob/main/docs/README-dynapi.md

I'd assume it does the right thing, but I am not 100% sure.

Now that I think about it, it sounds weird that SDL would run at all with a downgraded libSDL at runtime. Should probably ask the SDL devs about that...

src_c/base.c
jgerityneurala: Possibly. I did try putting it as early as possible in the init code and hit some error that I don't recall (I ran it in a container in a pretty ad-hoc way, don't have the output, sorry :disappointed:)

SnoopJeDi: Possibly. I did try putting it as early as possible in the init code and hit some error that I don't recall (I ran it in a container in a pretty ad-hoc way, don't have the output, sorry :disappointed:)

SnoopJeDi:
> This checking should probably be a bit more general than this. You can always use a newer version of SDL2 than compiled, but if you use an older one, it will break. Either we merge this and I whip something more future-proof up later, or we do it here.

I'm happy to change it to a hard error if the runtime SDL is behind the compiled version in any of these three fields, but tried to make the minimum change that addresses the issue that brought me to it.

> 
> Also, how does this interact with: https://github.com/libsdl-org/SDL/blob/main/docs/README-dynapi.md
> 
> I'd assume it does the right thing, but I am not 100% sure.

I believe [`SDL_GetVersion`](https://github.com/libsdl-org/SDL/blob/cb1fd30e9a2edf0bce6e86f5aae33dc28fe1822b/include/SDL_version.h#L109-L123) should always come from whatever's been linked against, but it's not clear to me from that document if the result is still valid if you load some other version using the dynamic API. Taking a quick peek [at how the dynamic switcharoo works](https://github.com/libsdl-org/SDL/blob/c59d4dcd38c382a1e9b69b053756f1139a861574/src/dynapi/SDL_dynapi_procs.h#L522) it seems to me like the answer is a qualified "yes," but this is something worth asking the SDL devs about.

> 
> Now that I think about it, it sounds weird that SDL would run at all with a downgraded libSDL at runtime. Should probably ask the SDL devs about that...

Definitely worth reaching out to have a conversation with them to figure out exactly what can be expected from patch versions. On the basis of this compatibility, I'd say the most intuitive conclusion is that all three version fields are important, and it should probably be an error if compile version is ahead of runtime.

robertpfeiffer:
I should have thought about it earlier. Obviously you can't compile this check into SDL, because the binary SDL library does not have access to the version your game is compiled against.

ankith26:
While this PR is looking good, in the long term, we would need a more robust solution. SDL may break ABI compatability between patch versions, we cannot really make assumptions that ABI does not change (SDL does keep adding many good features in between patch releases). I would say, we need to error if the linked and compiled versions don't match at all

SnoopJeDi:
Fine by me, let me know if this PR needs anything else to be ready for inclusion. Seems like the questions raised above are concerns for the maintenance team.

illume:
I'm a bit worried if we fix/warn this ourselves that it might conflict with other patches that people make to SDL.

To fix it, I guess we'd need to make a .so with the symbols/functions which we load only when SDL >= 2.0.0 < 2.0.14. I think that'd work (but not 100%). Ankith had an idea how to as well.

SnoopJeDi:
> Where and with what circumstances do you encounter this problem?

Discovered while helping a user that stumbled into this incompatibility on Raspbian. See #2657 for details.

> Is there a link on the SDL issue tracker about this ABI breakage? The warning is nice, but I wonder if we can fix the issue somehow rather than warning about it?

I don't recall seeing information about it, but I didn't look super hard, either.

> I'm a bit worried if we fix/warn this ourselves that it might conflict with other patches that people make to SDL.
> 
> To fix it, I guess we'd need to make a .so with the symbols/functions which we load only when SDL >= 2.0.0 < 2.0.14. I think that'd work (but not 100%). Ankith had an idea how to as well.

This sounds like it might be a separate issue, I don't have an opinion about it either way.

dr0id (approved):
LGTM although it probably needs merging head into it.

robertpfeiffer (approved):

src_c/base.c
```diff
@@ -192,6 +192,34 @@ pg_CheckSDLVersions(void) /*compare compiled to linked*/
...
+        /* SDL 2.0.14 replaces some macros with symbols, see
+         * https://github.com/libsdl-org/SDL/commit/316ff3847b4d9d87d9b0aab15321461db0e8ae0b
+         */
+        PyErr_Format(PyExc_RuntimeError,
+                     "Known SDL incompatibility detected! (compiled with "
+                     "version %d.%d.%d, linked to %d.%d.%d)",
+                     compiled.major, compiled.minor, compiled.patch,
+                     linked.major, linked.minor, linked.patch);
+        return 0;
+    }
```
ankith26, line 207: Is this block required? My understanding is that it is merely a special case of the more generic error raised in the next `else if` block.

src_c/base.c
```diff
@@ -192,6 +192,34 @@ pg_CheckSDLVersions(void) /*compare compiled to linked*/
...
+    else if ((linked.major == compiled.major &&
+              linked.minor == compiled.minor &&
+              linked.patch < compiled.patch) ||
+             (linked.major == compiled.major &&
+              linked.minor < compiled.minor)) {
```
ankith26, line 212: These two lines can probably be removed, `linked.minor != compiled.minor` check happens in the first `if` block

ankith26 (commented):
Revisiting this PR after quite some time in which this PR was buried under a load of PRs (sorry about that!)

I believe this is an important PR so it should not wait for some version like 2.2.0, lets try to get this one in 2.1.3

src_c/base.c
ankith26: hmmm... I think this check would be useless if it is called late. Atleast, it should be called before any other SDL API is called. In this case, `SDL_RegisterApp` is called before this function on windows, so this check should happen before that.

Additionally, `pg_CheckSDLVersions` is also called in the `pg_init` function, which will be redundant and should be removed, if this check already happens in the module init function

src_c/base.c
robertpfeiffer: Yes, it is a special case. It is the special case that has been *known* to cause a problem and reported in the wild, but other downgrades will be problematic too.

src_c/base.c
robertpfeiffer: This was just for clarity.

src_c/base.c
robertpfeiffer: The original PR was *only* the special case.

src_c/base.c
robertpfeiffer: It's still useful even if late, because there could be many bugs that happen in actual usage of pygame after all the init. Merging in this state (and making another issue to address later) would still be an improvement on the status quo.

src_c/base.c
robertpfeiffer: This is something the compiler will optimise, unless you pass -O0, and I think it's more readable like this. If you think the other way round is more readable, then by all means go ahead! If you are concerned about performance, this is just a one-off check using a few cycles, I wouldn't worry about it!

src_c/base.c
robertpfeiffer: Now that I think about it, SDL 2.0.5 also has a lot of incompatibilities, but 2.0.5 is ancient

src_c/base.c
ankith26: I was not worried about performance, but more about readability. But eh, anyways it's a minor thing, should be fine

ankith26 (approved):
Lets get this one in! 👍 🎉 
Any extra improvements/changes can happen in future PRs, instead of keeping this one blocked for a longer time, since this PR is nearly a year old now.

Thanks for the PR @SnoopJeDi and again, sorry for the delays!

PR #2663, opened 2021-08-05
Fix syntax for typing AbstractGroup attributes

### LINKED ISSUE #2662: Wrong syntax for typing AbstractGroup attributes
In sprite.pyi there is the following definition of AbstractGroup:

https://github.com/pygame/pygame/blob/8e3706b8d506379825ac90acfad98b5a27f48d67/buildconfig/pygame-stubs/sprite.pyi#L48-L51

However the attribute declarations are wrong: they should use `:` to declare the type annotation, instead of `=`!

This lead to Mypy errors when using those attributes.

vnmabus:
Closes #2662


Starbuck5:
Did you encounter this in the wild or did you happen across it when looking at the pyi file?

Because neither of those attributes are documented, as far as I can tell.

vnmabus:
I was typing a project which subclassed `AbstractGroup` and used those. But I think we should refactor away that subclass.

bitcraft (approved):
LGTM

ankith26 (approved):
LGTM as well! Thanks for your contribution! 👍 🎈

Starbuck5 (approved):
Thanks for the PR.

I see you're doing a lot of work on Tuxemon, maybe you ran into this when working on that.

vnmabus:
> I see you're doing a lot of work on Tuxemon, maybe you ran into this when working on that.

Yeah, that is the case. I didn't know you were familiar with that project.

PR #2670, opened 2021-08-14
Move to SDL 2.0.16 for Windows

Starbuck5:
This is a pretty simple PR that updates the windows buildconfig to use the new SDL 2.016, rather than SDL 2.0.14.

I also had to change some mixer tests, because SDL changed their audio sampling systems on Windows, leading to a change in frequency that is new behavior but within specification.

I fixed this by setting the `allowedchanges` to 0.

I raised an issue with SDL about this and they said it was fine, which it is, and it actually led to some commits under this issue: https://github.com/libsdl-org/SDL/issues/4608

Edit: Wow this "simple" PR failed epically. The renderer_set_viewport doesn't happen for me locally, but maybe it's a software backend thing, so I'll try to reproduce. - It actually looks like an x64 vs x86 thing, which is really odd.

### DIFF
--- a/test/mixer_test.py
+++ b/test/mixer_test.py
@@ -61,7 +61,7 @@ class MixerModuleTest(unittest.TestCase):
 
     def test_init__keyword_args(self):
         # note: this test used to loop over all CONFIGS, but it's very slow..
-        mixer.init(**CONFIG)
+        mixer.init(**CONFIG, allowedchanges=0)
         mixer_conf = mixer.get_init()
 
         self.assertEqual(mixer_conf[0], CONFIG["frequency"])
@@ -71,7 +71,7 @@ class MixerModuleTest(unittest.TestCase):
 
     def test_pre_init__keyword_args(self):
         # note: this test used to loop over all CONFIGS, but it's very slow..
-        mixer.pre_init(**CONFIG)
+        mixer.pre_init(**CONFIG, allowedchanges=0)
         mixer.init()
 
         mixer_conf = mixer.get_init()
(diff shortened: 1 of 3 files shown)


illume:
Seems there's another test failure on one of the windows machines...
```
FAIL: test_renderer_set_viewport (pygame.tests.video_test.VideoModuleTest)
works.
  File "C:\hostedtoolcache\windows\Python\3.5.4\x86\lib\site-packages\pygame\tests\video_test.py", line 21, in test_renderer_set_viewport
    self.assertEqual(renderer.get_viewport(), (0, 0, 1920, 1080))
AssertionError: <rect(0, 0, 1917, 1077)> != (0, 0, 1920, 1080)
```

test/mixer_test.py
```diff
@@ -86,7 +87,7 @@ def test_pre_init__zero_values(self):
         # default values. No way to check buffer size though.
         mixer.pre_init(22050, -8, 1)  # Non default values
         mixer.pre_init(0, 0, 0)  # Should reset to default values
-        mixer.init()
+        mixer.init(allowedchanges=0)
```
illume, line 90: How do we ensure that the original pre_init values are not used? By testing not equal.

```python
        self.assertNotEqual(mixer.get_init()[0], 22050)
        self.assertNotEqual(mixer.get_init()[1], -8)
        self.assertNotEqual(mixer.get_init()[2], 1)
```

But even that could result with weird hardware giving back those values as default. Imagine a 22050 only sound device.

Yeah, the `allowedchanges=False` is probably the best way to do this. But with our imaginary 22050 only device this test would mistakenly fail.

We'll see with wider testing, but I think this should be fine.

illume:
I suggest adding a skip on this video_test.py for windows 32bit for now.

```
    @unittest.skipIf(("Windows" in platform.platform() and not (sys.maxsize > 2**32)), "Windows 32 bit SDL 2.0.16 has an issue.")
```

test/mixer_test.py
Starbuck5: That's a good point, but yes I think this will be fine.

Starbuck5:
Alright, hopefully this is good to go now.

I'll also link the relevant issue: #2672

ankith26 (approved):
LGTM! 👍 🎈

illume (approved):
🎉  Thanks

PR #2673, opened 2021-08-15
Remove redundant python includes, fix compiler warnings

### LINKED ISSUE #2613: Init functions specified multiple times on windows builds
Getting lots of these warnings on windows builds.

```
newbuffer.obj : warning LNK4197: export 'PyInit_newbuffer' specified multiple times; using first specification
```

It might be the cause of an intermittent compilation error on windows:
Here is one example: https://ci.appveyor.com/project/pygame/pygame/builds/39301792/job/y216gt96xpyw8f6p#L800

It's not just limited to newbuffer, but this is where it seems to rarely crash.

ankith26:
Super minor PR, to hopefully fix #2613 

Since pygame header already includes `<Python.h>` re-including those is redundant. 
I am not a 100% sure that this fix works, because I believe python.h is header guarded anyways, but this PR is worth a shot

### DIFF
--- a/src_c/freetype.h
+++ b/src_c/freetype.h
@@ -26,7 +26,6 @@
 
 #include "pgcompat.h"
 #include "pgplatform.h"
-#include <Python.h>
 
 #include <ft2build.h>
 #include FT_FREETYPE_H
--- a/src_c/newbuffer.c
+++ b/src_c/newbuffer.c
@@ -19,7 +19,6 @@
 */
 
 #include "pygame.h"
-#include <Python.h>
 
 #include "pgcompat.h"
 


ankith26:
Okay, I realise this PR does not fix the root cause of the issue, but I think I now know how to fix it

illume (approved):
Thanks :) 🎉

ankith26:
Yeah sure, this can totally wait for 2.0.3. I agree the fix is a bit hacky, and this sort of thing is best not implemented just before a release. Also the issue that it fixes shows up rarely, so it should be fine :)

illume:
Yeah, let's merge it in now so it can get some testing.

PR #2674, opened 2021-08-15
Update Vector3 overloaded constructor signature

bydariogamer:
Update the typing of the overloaded constructor of `Vector3` in `buildconfig/pygame-stubs/math.pyi` so it accepts both integer and float numbers.


ankith26 (approved):
LGTM! 🥳 
Thanks for the first PR! 👍

Starbuck5:
We had a discussion about this over on discord, but the type hints for Vector2 and Vector3 init are still not correct, and it would be great to get those sorted in this PR.

This is what Ankith said:
Capture

Although I think the first overload case, for `Vector2()` (no args) is covered by the second overload case.

One of the problems is that the old ones use `Optional[stuff]`, which is the same as `Union[None, stuff]`. But Vectors don't take `None`.

ankith26:
> Although I think the first overload case, for `Vector2()` (no args) is covered by the second overload case.
Uhh actually no, the second overload typehint still expects an argument, so we need the no arg overload

Starbuck5:
Is typing `float` the same as `Union[int, float]`, because every int is compatible with float?

ankith26:
Yes, according to some typing PEP, `Union[int, float]` is redundant for `float` because any value that can take float, should also take int without any problems

Starbuck5 (approved):
Alright, this looks good to me. Thanks for the contribution bydario 👍 

Or should I call you freeduck?

PR #2675, opened 2021-08-15
Continue on aarch64 wheels

illume:
Continues on https://github.com/pygame/pygame/pull/2610 by @odidev 

This leaves building the images to another process, and just pulls them instead. Even with that, the arm64 build takes 4-5x as long (but at least not 3 hours) because of the qemu emulation.

This is still useful for releases, and maybe in the future we can split it up so the arm64 ones are done on arm hardware (either when github actions adds arm support or we use travis arm64 support). Additionally, people can make their own wheels locally if they want.

ps. this guide worked for setting up qemu on docker + ubuntu 20.04 https://www.stereolabs.com/docs/docker/building-arm-container-on-x86/

### DIFF
--- a/.github/workflows/manylinux.yml
+++ b/.github/workflows/manylinux.yml
@@ -42,7 +42,7 @@ jobs:
     - name: Build manylinux wheels
       run: |
         cd buildconfig/manylinux-build
-        make base-image-aarch64 pull pull-manylinux wheels wheels-manylinux
+        make pull pull-manylinux wheels wheels-manylinux
         cd ../..
         mkdir -p dist/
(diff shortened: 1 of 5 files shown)


ankith26 (approved):
LGTM! 👍 🥳

Starbuck5 (approved):
I approve, and I even sort of understand

Ankith really put the screws on me for the second approval 😅

ankith26:
hehe... This PR is pretty useful for the upcoming release, nice work on this illume, and also nice first contribution @odidev, thankies! 🥳

PR #2678, opened 2021-08-16
Fix Optional typehint usage in typehint files

ankith26:
Fixed `Optional` typehint usage. According to docs (https://docs.python.org/3/library/typing.html#typing.Optional), an `Optional` typehint signifies a typehint that can be `None`, but not an argument that is optional. This PR is a quickfix to this typing issue.

Also runs `black` on all modified files, for consistent code formatting with the rest of the project that uses the same

Co-authored with @zoldalma999 in a code collab


buildconfig/pygame-stubs/mask.pyi
```diff
@@ -59,24 +55,24 @@ class Mask:
...
+    def outline(self, every: int = 1) -> List[Tuple[int, int]]: ...
     def convolve(
         self,
         other: Mask,
         output: Optional[Mask] = None,
-        offset: Optional[_Coordinate] = (0, 0),
+        offset: _Coordinate = (0, 0),
     ) -> Mask: ...
     def connected_component(
         self, pos: Union[List[int], Tuple[int, int]] = (-1, -1)
```
Starbuck5, line 66: You didn't change this or anything, but this looks fishy to me.

It's not that the pos is a default argument, it really is just optional. https://www.pygame.org/docs/ref/mask.html#pygame.mask.Mask.connected_component.

Would passing in (-1,-1) really be the same as passing nothing? The docs don't say so.

Maybe this is a good spot for a good old `@overload` instead.

buildconfig/pygame-stubs/mixer.pyi
```diff
@@ -12,19 +13,19 @@ from . import music as music
...
-    frequency: Optional[int] = 44100,
-    size: Optional[int] = -16,
-    channels: Optional[int] = 2,
-    buffer: Optional[int] = 512,
-    devicename: Optional[Union[str, None]] = None,
+    frequency: int = 44100,
+    size: int = -16,
+    channels: int = 2,
+    buffer: int = 512,
+    devicename: Optional[str] = None,
```
Starbuck5, line 28: pre_init also needs the allowedchanges argument. The docs didn't have that correct until recently, but the docs source has been fixed.

buildconfig/pygame-stubs/mixer.pyi
```diff
@@ -81,5 +86,5 @@ class Channel:
     def get_busy(self) -> bool: ...
     def get_sound(self) -> Sound: ...
     def get_queue(self) -> Sound: ...
-    def set_endevent(self, type: Optional[Union[int, Event]] = None) -> None: ...
+    def set_endevent(self, type: Union[int, Event] = 0) -> None: ...
```
Starbuck5, line 89: Same concern here as with `mask.connected_component`. Is a default arg of zero the same as not passing an arg at all?

https://www.pygame.org/docs/ref/mixer.html#pygame.mixer.Channel.set_endevent

buildconfig/pygame-stubs/color.pyi
```diff
@@ -15,7 +15,7 @@ class Color:
     i1i2i3: Tuple[float, float, float]
     __hash__: None  # type: ignore
     @overload
-    def __init__(self, r: int, g: int, b: int, a: Optional[int] = ...) -> None: ...
+    def __init__(self, r: int, g: int, b: int, a: int = ...) -> None: ...
```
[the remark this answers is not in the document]
ankith26: Ohh yeah right

buildconfig/pygame-stubs/mask.pyi
ankith26: Yes, going through the code, I see that passing `(-1, -1)` would actually error, but if no value is passed, the internal functions get `(-1, -1)`

In any case, yeah you are right, we can make the typehint something like `pos: Union[List[int], Tuple[int, int]] = ...` and not `(-1, -1)`

buildconfig/pygame-stubs/mixer.pyi
ankith26: Ooh nice to know, will fix the typehint in this PR

buildconfig/pygame-stubs/mixer.pyi
ankith26: In this case, I had gone through the code, and can confirm that passing 0 indeed does the same thing as not passing an arg at all

Starbuck5 (approved):
Alright, this looks good to me! 🎉

MightyJosip (approved):
Nice. Thank you 👍

PR #2682, opened 2021-08-18
Change Surface input type from List to Sequence.

### LINKED ISSUE #2680: Use Sequence instead of List for read-only parameters
The following definition used to initialize Surface should be made more flexible, allowing `Sequence[float]` instead of `List[float]`.
https://github.com/pygame/pygame/blob/cbf4c674f568b5a17e13b0a697b43503936a883f/buildconfig/pygame-stubs/surface.pyi#L28
Note that as `List[float]` is invariant instead of covariant, the current implementation does not even allow to pass a `List[int]`!!

vnmabus:
Closes #2680.

### DIFF
--- a/buildconfig/pygame-stubs/_common.pyi
+++ b/buildconfig/pygame-stubs/_common.pyi
@@ -1,0 +1,4 @@
+from typing import Sequence, Tuple, Union
+from pygame.math import Vector2
+
+_Coordinate = Union[Tuple[float, float], Sequence[float], Vector2]
--- a/buildconfig/pygame-stubs/display.pyi
+++ b/buildconfig/pygame-stubs/display.pyi
@@ -4,10 +4,9 @@ from typing_extensions import Protocol
 from pygame.color import Color
 from pygame.surface import Surface
 from pygame.rect import Rect
-from pygame.math import Vector2
 from pygame.constants import FULLSCREEN
+from pygame._common import _Coordinate
 
-_Coordinate = Union[Tuple[float, float], List[float], Vector2]
 _CanBeRect = Union[
     Rect,
     Tuple[int, int, int, int],
--- a/buildconfig/pygame-stubs/draw.pyi
+++ b/buildconfig/pygame-stubs/draw.pyi
@@ -4,9 +4,8 @@ from typing_extensions import Protocol
 from pygame.color import Color
 from pygame.rect import Rect
 from pygame.surface import Surface
-from pygame.math import Vector2
+from pygame._common import _Coordinate
 
-_Coordinate = Union[Tuple[float, float], List[float], Vector2]
 _ColorValue = Union[
     Color, str, Tuple[int, int, int], List[int], int, Tuple[int, int, int, int]
 ]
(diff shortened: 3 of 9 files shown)


ankith26 (approved):
This PR is looking good! 🎉 

Are there any other places where a `List` typehint is used, but a `Sequence` typehint should have been used? If so, this PR would be a good opportunity to fix those.

Thanks for the PR 👍

vnmabus:
> Are there any other places where a `List` typehint is used, but a `Sequence` typehint should have been used?

The only case were `List` should be used instead of `Sequence` is when the list is going to be modified afterwards, for example appending elements, or if some list-only methods or attributes are used. In other cases one should use `Sequence` if possible.

I will review the other stubs and change them, and we can discuss the changes afterwards.

vnmabus:
Well, I think that most uses of `List` could be replaced by `Sequence`, but I do not know the Pygame API enough to be comfortable changing them.

Also, the `_CanBeRect` type, which makes use of `List`, is being repeated everywhere, instead of imported. That probably should be changed before in another PR.

Starbuck5:
You mentioned in one of your issues that there are type unions floating around the pygame stubs rather than centrally imported, sometimes with different definitions.

This just makes that worse, because it changes one of the `_Coordinate` definitions and not the others.

Also, I guess `Sequence` is theoretically better than list, but I can't think of many practical times someone wouldn't just pass in a list, and lists / tuples are much easier to assure actually work.

Also also, `pygame.Surface([20,20])` does type check, tested using `mypy`, which seems to undermine the reason for this PR in the first place.

vnmabus:
Please, double-check that Mypy accepts that as it didn't for me, and it shouldn't. It would be a type violation to accept a `List[int]` were a `List[float]` is required, as the code could legally append a float to the list, changing its type.

Starbuck5:
key.py:
```py
import pygame
pygame.Surface([20,20])
```

terminal:
```
C:\Users\Starbuck\Desktop>mypy key.py
Success: no issues found in 1 source file
```

Python 3.8, mypy 0.910, if it matters.

vnmabus:
Your example works because Mypy infers the type of the argument as `List[float]` because it is used directly. The following example does not typecheck:

```python
import pygame

a = [20, 20]
pygame.Surface(a)
```

Mypy says `Mypy: Argument 1 to "Surface" has incompatible type "List[int]"; expected "Union[Tuple[float, float], List[float], Vector2]"
"List" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
Consider using "Sequence" instead, which is covariant`.

Starbuck5:
You could've opened with that!

Ok, do you agree we should centralize the `_Coordinate` and other similar macros?

We could put them in `__init__.pyi` perhaps, or in modules that fit with them. Like `_CanBeRect` could be in `Rect.pyi`?

vnmabus:
As long as every function accepts the same kind of objects (e.g. that there isn't a case in which a function accepts `((4, 3), (1, 2))` as a rectangle and other doesn't) I think it would be better to centralize them, yes. I, however, do not know the Pygame API well enough to be comfortable doing a large manipulation.

Starbuck5:
How about you do a "small" large manipulation, and just put all the `_Coordinate`s together. I don't really want to merge this if it just changes one of them, when the same issue almost certainly exists for the rest.

Then reviewers (and/or you) can do some thorough testing to make sure it all works properly.

vnmabus:
Ok, I centralized the `_Coordinate` type.

Starbuck5:
Thanks for this, @vnmabus 🚀

Issue #2684, opened 2021-08-19
Fix type hint of get_pressed() in mouse.pyi


shingo-kagami:
Since the parameter `num_buttons` of `pygame.mouse.get_pressed` is optional with the default value 3, the correct type hint for it would be something like `int = 3` or `typing.Optional[int]`.

https://github.com/pygame/pygame/blob/34acc0edff5ce6dee6126d6cf07113ac5547fb37/buildconfig/pygame-stubs/mouse.pyi#L6

Starbuck5:
It actually shouldn't be `Optional`, see #2678. `Optional` is shorthand for `Union[None, stuff]` rather than for a default argument. Confusing naming by the typing folks, though.

`int = 3` would work great though.

Since it's a simple fix I'm assigning you to the issue, because if you can point out the exact line you can also fix it. But of course if you don't want to I'll unassign you. 😄

PR #2685, opened 2021-08-19
Fix type hint of get_pressed()

shingo-kagami:
Fix for Issue #2684


Starbuck5 (approved):
Thanks! This looks good to me.

You could also run `black` on the pyi file to auto format it, but plenty of the other files aren't formatted either.

ankith26 (approved):
Thanks for your first PR! 🎉 🥳

PR #2686, opened 2021-08-21
Better deprecation for swizzle, change up repr functions

Starbuck5:
This PR does 2 separate things:

### Thing 1

Prepares the `math.enable_swizzling()` and `math.disable_swizzling()` for removal. I removed the functionality of the functions, and made them emit deprecation warnings. I also removed calls to those functions from the tests.

### Thing 2

According to the [python datamodel docs](https://docs.python.org/3.8/reference/datamodel.html#object.__repr__) for __repr__

> If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned.

So, the `<...>` syntax used by most pygame repr functions is not quite correct. Don't get me wrong, some pygame classes definitely should use it, such as `pygame.Surface`, which cannot be recreated with the data printed out.

I went through the pygame classes and identified places where I thought the repr function could lose the brackets or be otherwise improved.

`<rect(1, 2, 3, 4)>` -> `Rect(1, 2, 3, 4)`

`<Vector2(1, 2)>` -> `Vector2(1, 2)`

`<Vector3(1, 2, 3)>` -> `Vector3(1, 2, 3)`

`(255, 0, 120, 255)` -> `Color(255, 0, 120, 255)`

I also removed the Vector `str()` functionality, because Vectors should be represented as Vectors.
```
>>> str(pygame.Vector2(1,2))
'[1, 2]'
```
Not as lists!

The Vector test suite has a test where it puts the output of `__repr__()` back into a Vector constructor, and it works. There's an undocumented way to construct a Vector using a string. So I had to change that undocumented way to accept the new string format rather than the old one.

I'm not sure why this functionality exists. The only thing I could think of is for pickling, but the Vector classes each have a reduce() method to handle that. It seems useless and undocumented, although it is tested for, and so I would like to remove it. I left it in for this PR though, because I'm not sure if it has a use, and I want to keep this PR simpler to review.

I wanted to do `clang-format` on the files, but I put it as a separate commit so it would be easier to see changes by clicking on the other two commits.

Other math.c things that could be improved: the vectors have an undocumented `epsilon` attribute, the code has a bunch of commented out buffer protocol stuff that hasn't been touched since at least 2012, so that should just be deleted.

### DIFF
--- a/src_c/color.c
+++ b/src_c/color.c
@@ -45,9 +45,9 @@
 
 #include <ctype.h>
 
-
-#if (!defined(__STDC_VERSION__) || __STDC_VERSION__ < 199901L) && !defined(round)
-#define pg_round(d) (((d < 0) ? (ceil((d)-0.5)) : (floor((d)+0.5))))
+#if (!defined(__STDC_VERSION__) || __STDC_VERSION__ < 199901L) && \
+    !defined(round)
+#define pg_round(d) (((d < 0) ? (ceil((d)-0.5)) : (floor((d) + 0.5))))
 #else
 #define pg_round(d) round(d)
 #endif
@@ -190,7 +190,6 @@ pg_RGBAFromColorObj(PyObject *color, Uint8 rgba[]);
 static int
 pg_RGBAFromFuzzyColorObj(PyObject *color, Uint8 rgba[]);
 
-
 /**
  * Methods, which are bound to the pgColorObject type.
  */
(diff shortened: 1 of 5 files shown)


illume:
We probably can't change repr output easily, especially because many tests rely on them. If we do decide this is valuable enough a change to do, then it would have to be a major version change.

Starbuck5:
This is definitely low priority, just housekeeping things.

However, it does pass the tests, and I didn't have to edit them all too much.

Starbuck5:
Zoldalma got the swizzle stuff through.

I’d still like to change the Reprs, I think the number of people impacted will be very minimal. But I don’t think it will happen soon, so I’ll close this.

PR #2688, opened 2021-08-22
2.0.2.dev2

illume:


illume:
... before this is merged, we should change the version to "2.0.2.dev3"?

Starbuck5:
Ok! I pushed the tag.

dev3 is the way to go. It allows us more flexibility about when to release then going into rc (release candidate) territory.

Maybe we should let the wheels generate first before going to dev3 though? Just for convenience.

PR #2694, opened 2021-08-25
More informative FileNotFoundError

Starbuck5:
This PR adds a more informative FileNotFoundError to pygame's RWops system.

Take this, for example:
```py
pygame.mixer.Sound("data/nonexistent.ogg")
```

Normally, if there's a problem, pygame will say:
```
FileNotFoundError: No such file or directory.
```

Which is accurate, but not super helpful, especially for newbies. And I see newbies struggle with the concept of working directories a lot.

This PR uses the os module to produce this error message instead:
```
FileNotFoundError: No file 'data/nonexistent.ogg' found in working directory 'C:\Users\Starbuck\Desktop'.
```

This could obviously still be misinterpreted, but the term "working directory" will hopefully point people in a good direction with Google searches.

And this information is omitted for absolute paths, because for those it doesn't matter what your working directory is. And if the `os` module is unavailable for some reason, it will just fall back to the old error message.

However, there's a problem. Most of pygame's loading functions make really complex use of the pg_RWops API, rather than using the more high level functions. So, right now, this FileNotFound will only be emitted by Sound objects failing to load. I will broaden this a lot in a companion PR, kept separate for ease of review.


src_c/rwobject.c
```diff
@@ -736,8 +738,50 @@ _rwops_from_pystr(PyObject *obj)
...
+
+                    PyObject* path = PyObject_GetAttrString(os_module, "path");
+                    if (path == NULL) {
+                        Py_DECREF(cwd);
+                        PyErr_SetString(PyExc_FileNotFoundError,
+                                        "No such file or directory.");
+                        return NULL;                        
+                    }
+
+                    PyObject* abs = PyObject_CallMethod(path, "isabs", "O", obj);
```
ankith26, line 759: The `abs` variable needs a new name, because `abs` is already a function in C stdlib

ankith26 (approved):
Yes, I like the idea behind this PR. It should be really helpful for many newbies when they get FileNotFound errors in their code, and this info will help them debug better

src_c/rwobject.c
Starbuck5: Thanks, good catch.

Starbuck5:
I also updated this so it absolute paths will do the normal python thing and say what file isn't found. I can't believe it didn't do that before!

`pygame.mixer.Sound("C:/Users/Starbuck/desktop/not_a_sound.ogg")` Now emits

`FileNotFoundError: No such file or directory: 'C:/Users/Starbuck/desktop/not_a_sound.ogg'.`

instead of just

`FileNotFoundError: No such file or directory.`

illume (approved):
Really great usability improvement. Thanks! 🎉 

LGTM

robertpfeiffer:
So... this could probably be slightly inaccurate on some platforms where SDL_RWOps loads assets differently, but at least on Android (the only relevant one because PyGame officially aims to support it) the APK asset loading via JNI is done in addition to regular fopen() in the current directory, so it works out, and you can still put your assets into your P4A code directory.

Just something to keep in mind in the future...

PR #2695, opened 2021-08-25
Simplify pygame resource loaders

Starbuck5:
In my journey of #2694, I realized that almost every resource loading function in pygame duplicates a lot of the same code, and they are very complex.

But I noticed there was one function provided by rwobject.c that could simplify a lot of this down, `pgRWops_FromObject`. It can handle string filenames and file like objects.

The reason it hasn't been used much is that lots of the resource loaders either need to pass the filepath to SDL, or the type / extension of the file. And you can't get the extension out of an `SDL_RWops`.

So all our resource loaders would re-do all the functionality nicely wrapped up in `pgRWops_FromObject` just to have the actual name of the file, and use the non-RW loader to load directly from a file name.

To simplify this, I made resource loaders that use `SDL_RWops` unconditionally, and changed rwobject.c to save the extension if loading an `SDL_RWops` from a real file. I then added the `pgRWops_GetFileExtension` function to allow the resource loaders to access the information.

Here are the 7 resource loading functions in pygame's C code:
```
mixer.Sound              - used pgRWops_FromObject already
image.load_basic         - converted in this PR (easily, because it didn't need the file type, saved 15 lines)
image.load_extended      - converted in this PR (saved 130 lines)
mixer.music.load         - converted in this PR (saved 19 lines)
mixer.music.queue        - can't mess with this one, because #2558 is pending
freetype.Font            - scared to mess with this one
font.Font                - scared to mess with this one
```

I would like to eventually hit freetype.Font and font.Font, along with perhaps image.save (looks really complex), but this PR will be hard enough to review with just the changes I've done already.

Oh, and this will make the errors from #2694 be emitted by mixer.music and image, alongside mixer.

### DIFF
--- a/src_c/music.c
+++ b/src_c/music.c
@@ -370,22 +370,24 @@ _get_type_from_hint(char *namehint)
 }
 
 Mix_Music * 
-_load_music(PyObject *args) {
-    PyObject *obj;
+_load_music(PyObject *obj, char *namehint) {
     Mix_Music *new_music = NULL;
     char* ext = NULL;
-    char *namehint = NULL;
     SDL_RWops *rw = NULL;
-
-    if (!PyArg_ParseTuple(args, "O|s", &obj, &namehint)) {
-        return NULL;
-    }
+    PyObject* _type = NULL;
+    PyObject* error = NULL;
+    PyObject* _traceback = NULL;
 
     MIXER_INIT_CHECK();
 
     rw = pgRWops_FromObject(obj);
-    if (rw == NULL) /* stop on NULL, error already set */
+    if (rw == NULL) { /* stop on NULL, error already set is what we SHOULD do */
+        PyErr_Fetch(&_type, &error, &_traceback);
+        PyErr_SetObject(pgExc_SDLError, error);
+        Py_XDECREF(_type);
+        Py_XDECREF(_traceback);
         return NULL;
+    } 
     if (namehint) {
         ext = namehint;
     } else {
--- a/test/mixer_music_test.py
+++ b/test/mixer_music_test.py
@@ -26,9 +26,6 @@ class MixerMusicModuleTest(unittest.TestCase):
         if pygame.mixer.get_init() is None:
             pygame.mixer.init()
 
-    @unittest.skipIf(
-        "Darwin" in platform.system(), "SDL2_mixer not loading mp3 on travisci"
-    )
     def test_load_mp3(self):
         "|tags:music|"
         self.music_load("mp3")
@@ -182,7 +179,7 @@ class MixerMusicModuleTest(unittest.TestCase):
 
     def test_queue__invalid_filename(self):
         """Ensures queue() correctly handles invalid filenames."""
-        with self.assertRaises(IOError):
+        with self.assertRaises(pygame.error):
             pygame.mixer.music.queue("")
 
     def todo_test_stop(self):


src_c/rwobject.c
```diff
@@ -722,22 +737,43 @@ _rwops_from_pystr(PyObject *obj)
...
         PyObject *oencoded;
+        char* encoded = NULL;
+        char* ext = NULL;
+        char* extension = NULL;
+
+        //longest extension is 4 chars, plus NULL at end
+        extension = malloc(sizeof(char)*5);
+        if (extension == NULL) {
+            return (SDL_RWops*)PyErr_NoMemory();
+        }
```
ankith26, line 748: hmmmm this part smells like a potential segfault magnet. By allocating only 5 bytes, we are not handling any wider extensions (and those do exist, just they are not common enough). Now, when a user tries to load such a file with a large extension, and exceeds 5 bytes, the `strcpy` will probably try to copy that into this small allocated memory space, and the rest of the data overflows and potentially segfaults

A cleaner solution here is to have just enough memory allocated to store the entire extension string (with some API functions of PyUnicode/PyString)

ankith26 (changes requested):

src_c/rwobject.c
Starbuck5: Thanks for the feedback, I'll try and implement that.

src_c/rwobject.c
```diff
@@ -722,22 +737,43 @@ _rwops_from_pystr(PyObject *obj)
...
             return NULL;
         }
         if (oencoded != Py_None) {
-            rw = SDL_RWFromFile(Bytes_AS_STRING(oencoded), "rb");
+            encoded = Bytes_AS_STRING(oencoded);
+            rw = SDL_RWFromFile(encoded, "rb");
+            ext = strrchr(encoded, '.');
+            if (ext) {
+                ext++;
+                strcpy(extension, ext);
```
illume, line 760: Perhaps move the malloc to something like this down here (untested):
```c
        if (ext && strlen(ext)) {
            extension = malloc(strlen(ext));
            if (extension == NULL) {
                return (SDL_RWops*)PyErr_NoMemory();
            }
            strcpy(extension, ext);
        }
```

src_c/rwobject.c
```diff
@@ -722,22 +737,43 @@ _rwops_from_pystr(PyObject *obj)
...
+            }
         }
         Py_DECREF(oencoded);
         if (rw) {
+            /* adding the extension to the hidden data for RWops from files */
```
illume, line 765: +1 for great comment

illume (approved):
Apart from the issue ankith26 found, LGTM.

🎉 thanks!

src_c/rwobject.c
Starbuck5: Thanks! I thought that this system would be very confusing to figure out for anyone new coming to look at the code.

src_c/rwobject.c
Starbuck5: This kind of c code still makes me extremely cautious.

I've gone with this:
```c
ext = strrchr(encoded, '.');        
if (ext && strlen(ext) > 1) {
    ext++;
    extension = malloc(strlen(ext)+1);
    if (extension == NULL) {
        return (SDL_RWops*)PyErr_NoMemory();
    }
    strcpy(extension, ext);
}
```

strlen() > 1 because of the case where the filename ends with a period
malloc with a + 1 to leave space for the null terminator of the string

illume: That's looking good now :)

ankith26 (approved):
Yeah, looks good now! 🎈

illume:
I'll fix the conflict, because I merged in the other PR which caused it.

Starbuck5:
I want this to happen in 2.0.3, not 2.0.1.

The only API changed is mixer.music.load and mixer.music.queue, so I suppose they could be "fixed up" to return the content as a `pygame.error` instead.

Also, IOError is python 2 only, I think. FileNotFoundError is the way now.

Starbuck5:
@illume, I fixed it so it can go in with no API change!

Rather than passing along the error it gets from the files, it reads that error and retransmits it as a pygame.error.

If this is merged we should raise an issue so we don't forget to change this in 2.1.

I also removed the macOS test mp3 skip, because it is no longer necessary.

PR #2697, opened 2021-08-26
Update annotation for pygame.Surface.blits() (and some other tidy ups)

rethanon:
Updated annotation for pygame.Surface.blits() to correct the keyword used and expected parameters.


rethanon:
> Uh, that's weird.
> 
> setup.py docs should just build the docs, not pygame.

I've taken a closer look at the error I'm getting on my Windows machine and the first line starts with
`WARNING, No "Setup" File Exists, Running "buildconfig/config.py"`

I assume it's then trying to build pygame rather than just the docs.

rethanon:
Added a few minor updates to annotations in mixer.pyi, also corrected a couple of minor typos in the mixer docs.

docs/reST/ref/surface.rst
```diff
@@ -134,8 +134,8 @@
 
       | :sl:`draw many images onto another`
       | :sg:`blits(blit_sequence=(source, dest), ...), doreturn=1) -> [Rect, ...] or None`
```
Starbuck5, line 136: This one still needs another parenthesis.

buildconfig/pygame-stubs/mixer.pyi
```diff
@@ -19,7 +19,7 @@ def init(
     buffer: int = 512,
     devicename: Optional[str] = None,
     allowedchanges: int = 5,
-) -> None: ...
+) -> True: ...
```
Starbuck5, line 22: This probably returns whether init was successful, not just True.

Also, if you're going to fix that you've got to fix the docs too, because they say it returns None.

docs/reST/ref/surface.rst
rethanon: Somehow I didn't even notice that  one! Thanks.

buildconfig/pygame-stubs/mixer.pyi
rethanon: My fault, getting mixed up with which changes I'd made as I was looking at my fork on 2 different machines, think I committed the wrong one as the other one I'd changed to bool, not True. Also updated docs, apologies. I feel like a noob!

Starbuck5:
Alright, I opened a broader issue about the mixer.init() thing, #2702, so I think it would be best to take the mixer.init() docs and types changes out of this PR.

Thanks for pointing that out though. I think it should be fixed across several modules, by changing the code to match the docs, rather than the other way around.

And the parentheses should be after `blit_sequence=`, since the sequence is the value of the keyword argument.

Also I guess I need to fix `setup.py docs` so that people who don't have the full setup to build pygame can build the docs.

rethanon:
Reverted changes for mixer docs and types and fixed parentheses in surface.blits docs. Thanks.

ankith26 (approved):
Looks good to go! Thanks for the PR! 🎈

Starbuck5 (approved):
This looks good to me as well.  👍 

Thanks for contributing!

PR #2700, opened 2021-08-28
Try export MACOSX_DEPLOYMENT_TARGET

Starbuck5:
99% of this is stolen from Ankith,

I just want to see if `export MACOSX_DEPLOYMENT_TARGET=10.13` solves everything, which it possibly will.

*"everything" does not include ARM-based Macintosh systems, offers not valid after 9/01/2021, terms and conditions apply.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -61,7 +61,7 @@ jobs:
           # CIBW_ARCHS_MACOS: "x86_64 universal2 arm64"
 
           # Install some dependencies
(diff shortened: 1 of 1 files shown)


Starbuck5:
Why is this so painful?

Homebrew doesn't support setting deployment target: https://github.com/Homebrew/legacy-homebrew/issues/16285 (At least they didn't 9 years ago, and I couldn't find any indication it has changed)

Starbuck5:
I guess this plan isn’t going to work.

PR #2703, opened 2021-08-30
added test case for gl_get attribute

shreyabandhakavi:


test/display_test.py
```diff
@@ -398,29 +332,27 @@ def todo_test_gl_set_attribute(self):
         'iconify is only supported on some video drivers/platforms'
     )
     def test_iconify(self):
```
ankith26, line 334: why is this function being changed here? I remember having updated this unit test a few months ago, and now you are reverting it?

ankith26 (changes requested):
First of all, thanks for contributing to pygame! Sorry it took a while for this PR to get some attention.

I went through the code changes, and I don't understand a lot of the changes being made here? Some changes seems redundant, and in some places it seems like you are reverting work commits before this PR did already. Maybe something went wrong while fixing merge conficts? I suggest you start cleanly, start from the latest `main` branch updated on your work branch!

MyreMylar:
I'm closing this PR. 

It seems to be a bit of a mess code wise (beyond the scope of the description) and there is no response from the original poster after six months.

PR #2705, opened 2021-08-30
checking pip version

anukaal:


setup.py
```diff
@@ -10,6 +10,20 @@
 
 with io.open('README.rst', encoding='utf-8') as readme:
     LONG_DESCRIPTION = readme.read()
+    
+if sys.version_info < (3, 6):
```
robertpfeiffer, line 14: why is this here, and not with the other version checks?

setup.py
```diff
@@ -10,6 +10,20 @@
...
+     pip_message = 'This may be due to an out of date pip. Make sure you have pip >= 9.0.1.'
+     try:
+         import pip
+         pip_version = tuple([int(x) for x in pip.__version__.split('.')[:3]])
```
robertpfeiffer, line 18: why is pip supposed to be a dependency when building a wheel?

robertpfeiffer (changes requested):

Starbuck5:
I think I'm going to close this PR.

From reading the comments on another repository you added this exact PR to, I learned that this is so users on older versions of pip don't get confused if it attempts to install a version that isn't compatible with their python. Since the python_requires tag isn't supported on older pip: https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires.

However, we don't need this patch, because we still support older Pythons, like 2.7 and 3.5.

PR #2708, opened 2021-09-04
remove `_numpysndarray` and `_numpysurfarray` modules for being useless

bydariogamer:
`_numpysndarray` and `_numpysurfarray` were removed as the only supported array type is `numpy`.
All functions were moved to `sndarray` or `surfarray`, replacing the useless try/except calls.
The docstrings were updated to make clear numpy is the only supported module.
Deprecation warnings where added to functions about  the arraytype.


Starbuck5:
Thanks for taking the initiative to help modernize the pygame codebase!

My first review concern is the difference with how lack of numpy is handled, since this imports numpy unconditionally. I tested it, and if numpy is not found (or any of the surfarray imports fail), `surfarray` becomes a `MissingModule` object. Previously, pygame would supply the module, but it would have an ugly crash when called:

```
>>> s = pygame.Surface((20,20))
>>> pygame.surfarray.pixels_blue(s)
  File "C:\Users\Starbuck\AppData\Local\Programs\Python\Python38\lib\site-packages\pygame\surfarray.py", line 288, in pixels_blue
    return numpysf.pixels_blue(surface)
AttributeError: 'NoneType' object has no attribute 'pixels_blue'

During handling of the above exception, another exception occurred:

  File "C:\Users\Starbuck\AppData\Local\Programs\Python\Python38\lib\site-packages\pygame\surfarray.py", line 290, in pixels_blue
    import pygame._numpysurfarray as numpysf
  File "C:\Users\Starbuck\AppData\Local\Programs\Python\Python38\lib\site-packages\pygame\_numpysurfarray.py", line 49, in <module>
    import numpy
ModuleNotFoundError: No module named 'numpy'
```

I think having it be `MissingModule` is better, overall. It's just a non obvious change to consider.

So, the array_ methods were exposed already, just undocumented. That's not too surprising (did you know that Vectors have an epsilon attribute for equality comparisons?). The difference seems to be that the `pixel_` functions share the memory, whereas the `array_` functions copy the data. I'm not sure how useful this distinction really ends up being, but it's good to have documentation up to spec. The documentation updates you provided are up to date with the docstrings of the functions, but I still think that the wording is quite misleading about the actual distinction.

Starbuck5:
Newly documented functions also need types.

Starbuck5:
Alright, I'm thinking this is pretty much ready to go.

However, why do the array_ ones say they can take any Surface format, and the pixel_ ones say they can only take 24/32 bits?

I thought they worked nearly the same under the hood, so one of those is probably wrong.

bydariogamer:
> However, why do the array_ ones say they can take any Surface format, and the pixel_ ones say they can only take 24/32 bits?
> I thought they worked nearly the same under the hood, so one of those is probably wrong.

The pixel_ functions return an array which directly points to the surface values. Only 24 or 32 bits depths can be mapped to 3 uint8 in `pixels3d`. In the other hand, pixels
Trying to use other surface depths with `pixels3d` functions will raise `ValueError`:
```
>>> pixels_16_3d = pygame.surfarray.pixels3d(surface_16)
  File "~/.local/lib/python3.9/site-packages/pygame-2.0.2.dev3-py3.9-linux-x86_64.egg/pygame/surfarray.py", line 185, in pixels3d
    return numpy_array(surface.get_view('3'), copy=False)
ValueError: unsupported bit depth 16 for 3D reference array
```
The same applies to the pixels_\<color\> ones:
```
>>> pixels_16_b = pygame.surfarray.pixels_blue(surface_16)
  File "~/.local/lib/python3.9/site-packages/pygame-2.0.2.dev3-py3.9-linux-x86_64.egg/pygame/surfarray.py", line 310, in pixels_blue
    return numpy.array(surface.get_view('B'), copy=False)
ValueError: unsupported colormasks for blue reference array
```
The only pixels_ function that accepts any depth is `pixel2d`. As an example, this will work fine:
```
>>> pixels_8_2d = pygame.surfarray.pixels2d(surface_8)
>>> pixels_8_2d
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=uint8)

```
In the other hand, all array_ functions accept any pixel depth, as they simply copy data and the array values are not linked to anything:
```
>>> array_8_b = pygame.surfarray.array_blue(surface_8)
>>> array_8_b
array([[0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]], dtype=uint8)
```
Or in the `array3d` case:
```
>>> array_8_3d = pygame.surfarray.array3d(surface_8)
>>> array_8_3d
array([[[0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0]],

       [[0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],

    <more output, elided for readability, was here>

        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0],
        [0, 0, 0]]], dtype=uint8)
```

Starbuck5 (approved):
Alright, this one looks good to me. Thanks @bydariogamer! 🎮

illume (approved):
Great improvements :) Thanks! 🎉

PR #2709, opened 2021-09-09
Added doc for Cursor object

fortwoone:
The pygame.cursors.Cursor object is a brand new feature that I have just discovered, and I find sad that is is not documented already, so I tried to make a decent one

### DIFF
--- a/src_py/cursors.py
+++ b/src_py/cursors.py
@@ -29,6 +29,8 @@ Here is a list of available cursors:
 
 There is also a sample string cursor named 'thickarrow_strings'.
 The compile() function can convert these string cursors into cursor byte data.
+
+Alternately, you can also use Cursor objects to use colored images ingame.
 """
 
 import pygame
@@ -49,6 +51,17 @@ _cursor_id_table = {
 
 class Cursor(object):
     def __init__(self, *args):
+        """Cursor()->Cursor
+        Cursor(constant)->Cursor
+        Cursor(Cursor)->copies the Cursor object passed as an argument
+        Cursor(hotspot, Surface)->Cursor
+        
+        pygame object for representing cursors
+        
+        You can initialize a cursor from a system cursor or use the constructor on an existing Cursor object,
+        which will copy it. 
+        Providing a Surface instance will transform the cursor image into the Surface.
+        Said Surfaces may use other colors than black and white."""
         if len(args) == 0:
             self.type = "system"
             self.data = (pygame.SYSTEM_CURSOR_ARROW,)
@@ -222,6 +235,7 @@ tri_right = Cursor(
 # Here is an example string resource cursor. To use this:
 #    curs, mask = pygame.cursors.compile_cursor(pygame.cursors.thickarrow_strings, 'X', '.')
 #    pygame.mouse.set_cursor((24, 24), (0, 0), curs, mask)
+# Be warned, though, that cursors created from compiled strings do not support colors.
 
 # sized 24x24
 thickarrow_strings = (


Starbuck5:
It is documented, it's just that the docs on the website haven't been updated since 2.0.0. If you generate the docs locally you'll be able to see it, or if you snoop around the `.rst` files themselves.

It is a shame though, so I wrote a reddit post that will hopefully explain it to people before the docs get regenerated: https://www.reddit.com/r/pygame/comments/octscc/definitive_guide_to_pygame_cursors/.

With that in mind, do we need docstrings in the source? Maybe the docstrings would should up on some editors, which could help people out? But they should probably be a close copy of what is in the documentation itself.

fortwoone:
Sure, but I noticed there was one __init__ overload that was not documented, and I added it to the docstring inside the cursor object. It is not documented in the corresponding .rst file either sadly...

Starbuck5:
The `Cursor(cursor)` overload is arguably not important to document, since it's assumed behavior from most python objects.

But if you want to document it you should change it in the documentation, not in a docstring.

Also, your docstring misses an `__init__` case, for the old school 4 argument cursors.

fortwoone:
Oh thanks for noticing. Do you think I should document it too ?

fortwoone:
Ok so now I am finished

src_py/cursors.py
```diff
@@ -80,6 +94,11 @@ def __eq__(self, other):
 
     def __ne__(self, other):
         return not self.__eq__(other)
+    
+    def __copy__(self):
+        return self.__class__(self)
+    
+    copy=__copy__
```
Starbuck5, line 101: Thought you could sneak an API change under my nose? :smile:

You want to add a function, you got to add it to the docs.

src_py/cursors.py
```diff
@@ -49,6 +51,18 @@
 
 class Cursor(object):
     def __init__(self, *args):
+        """Cursor() -> Cursor
+        Cursor(constant) -> Cursor
+        Cursor(Cursor) -> copies the Cursor object passed as an argument
+        Cursor(hotspot, Surface) -> Cursor
+        Cursor(size, hotspot, xormasks, andmasks) -> Cursor
```
Starbuck5, line 58: This should be in the same order, and the exact same as it is in the documentation

src_py/cursors.py
```diff
@@ -29,6 +29,8 @@
 
 There is also a sample string cursor named 'thickarrow_strings'.
 The compile() function can convert these string cursors into cursor byte data.
+
+Alternately, you can also use Cursor objects to use colored images ingame.
```
Starbuck5, line 33: This kinda implies that Cursor objects are just for the surfaces, when it's really more of a unifying API.
Also, should be "in game"

docs/reST/ref/cursors.rst
```diff
@@ -138,6 +138,8 @@ The following strings can be converted into cursor bitmaps with
    | :sg:`Cursor(size, hotspot, xormasks, andmasks) -> Cursor`
    | :sg:`Cursor(hotspot, surface) -> Cursor`
    | :sg:`Cursor(constant) -> Cursor`
+   | :sg:`Cursor(Cursor) -> copies the Cursor object as a parameter`
```
Starbuck5, line 141: I still don't think this needs documentation

src_py/cursors.py
fortwoone: I did not want to sneak an API change. If you look into the commits normally I mentioned that change

src_py/cursors.py
fortwoone: Ok thanks I will correct this

docs/reST/ref/cursors.rst
fortwoone: But you did document a similar constructor for pygame.Color objects, why not do the same here ?

src_py/cursors.py
fortwoone: Look into the ecf2fc72 commit

src_py/cursors.py
fortwoone: This just got corrected

src_py/cursors.py
Starbuck5: I'm just joking with you. Although it's more conventional to have changes in comments than buried in the commit explanations themselves.

Still needs to be added to the docs.

docs/reST/ref/cursors.rst
Starbuck5: Fine. But it should be `-> Cursor` just like everything else.

src_py/cursors.py
fortwoone: Oh yes, I added it in a docstring... but not in the rst file, silly me

Starbuck5: For the rst file, I meant as a method, not a note. Like Rect.copy().

src_py/cursors.py
Starbuck5: >  You can initialize a cursor from a system cursor or use the constructor on an existing Cursor object, which will copy it. 
Providing a Surface instance will transform the cursor image into the Surface. Said Surfaces may use other colors than black and white.

Why mention 2 of the ways (+copying) to construct a cursor, but not the third?

Also "transform the cursor image into the Surface" sounds backwards to me. "Set the cursor to the image in the Surface" ?

fortwoone: Oh, no, I just sucked while writing, sorry

docs/reST/ref/cursors.rst
```diff
@@ -174,6 +176,12 @@ The following strings can be converted into cursor bitmaps with
       pygame.SYSTEM_CURSOR_NO          slashed circle or crossbones
       pygame.SYSTEM_CURSOR_HAND        hand
 
+   **Creating a cursor without passing arguments**
+   
+   In addition to the cursor constants available and described above,
+   you can also call ``pygame.cursors.Cursor()``, and your cursor is ready.
+   Doing this call actually creates a system cursor using the default native image.
```
Starbuck5, line 183: maybe this should just say it's the same as pygame.SYSTEM_CURSOR_ARROW, because that's what it actually does.

docs/reST/ref/cursors.rst
```diff
@@ -199,6 +207,14 @@ The following strings can be converted into cursor bitmaps with
...
    Width and height must be a multiple of 8, and the mask arrays must be the 
    correct size for the given width and height. Otherwise an exception is raised.
+   
+   .. method::copy
+      | :sl:`copy the current cursor`
+      | :sg:`copy() -> Cursor`
+      
+      Returns a new Cursor object with the same data and hotspot as the original.
+   .. ## pygame.cursors.Cursor.copy ##
+
```
Starbuck5, line 217: this doesn't show up like at all when I generate the docs locally. Maybe it's because of the spacing? You're going to have to generate the docs locally, I think. It also updates some header files with the docstrings, so that change should also be committed. You can try `setup.py docs`, and if that doesn't work you could rebase this branch against the current main, where it should be easier to generate them. If you're newer to git (like me), the easiest way to do that would just be to close this PR and make a new one with the changes off of current main.

src_py/cursors.py
```diff
@@ -28,7 +28,8 @@
     arrow, diamond, ball, broken_x, tri_left, tri_right
 
 There is also a sample string cursor named 'thickarrow_strings'.
-The compile() function can convert these string cursors into cursor byte data.
+The compile() function can convert these string cursors into cursor byte data that can be used to
+create Cursor objects (which can also be instantiated from surfaces).
```
Starbuck5, line 32: I think this change makes things more confusing.

docs/reST/ref/cursors.rst
fortwoone: Done

docs/reST/ref/cursors.rst
fortwoone: I actually figured it out, a spacing mistake

src_py/cursors.py
fortwoone: Ok, I will try to make it clearer

docs/reST/ref/cursors.rst
```diff
@@ -174,6 +176,13 @@ The following strings can be converted into cursor bitmaps with
       pygame.SYSTEM_CURSOR_NO          slashed circle or crossbones
       pygame.SYSTEM_CURSOR_HAND        hand
 
+   **Creating a cursor without passing arguments**
+   
+   In addition to the cursor constants available and described above,
+   you can also call ``pygame.cursors.Cursor()``, and your cursor is ready (doing that is the same as
+   calling ``pygame.cursors.Cursor(pygame.SYSTEM_CURSOR_ARROW)``.
+   Doing one of thoses calls actually creates a system cursor using the default native image.
```
Starbuck5, line 184: "thoses" is a grammatical error.

Also this whole sentence can be deleted.

Starbuck5:
Also, it would be good if you could build the docs locally, because that would forward your changes to `src_c/doc/cursors_doc.h`

It should just be `setup.py docs`, but if you're on an outdated version of main, it requires installing the pygame build dependencies, which is difficult on Windows.

fortwoone:
Ok I will check

Issue #2419, opened 2020-12-23
MP3 does not work after upgrade of Pygame 1.9.6 -->  2.0.0


dakshsriv:
**Environment:**

You can get some of this info from the text that pops up in the console when you run a pygame program.

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac):
```
$ cat /etc/os-release
NAME="elementary OS"
VERSION="5.1.7 Hera"
ID=elementary
ID_LIKE=ubuntu
PRETTY_NAME="elementary OS 5.1.7 Hera"
LOGO=distributor-logo
VERSION_ID="5.1.7"
HOME_URL="https://elementary.io/"
SUPPORT_URL="https://elementary.io/support"
BUG_REPORT_URL="https://github.com/elementary/os/issues/new"
PRIVACY_POLICY_URL="https://elementary.io/privacy-policy"
VERSION_CODENAME=hera
UBUNTU_CODENAME=bionic
```

- **Python version** (e.g. 3.7.9, 3.8.5) :
```
$ python3 -V
Python 3.6.9
```
- **SDL version** (e.g. SDL 2.0.12):
```
$ sudo apt search python3 |grep -i sdl
python3-sdl2/bionic,bionic,now 0.9.3+dfsg2-1 all [installed]
  Python bindings to the SDL2 C-library (Python 3 build)
$ pip3 list
pip (9.0.1)
pkg-resources (0.0.0)
pygame (2.0.0)
PySDL2 (0.9.7)
PySimpleGUI (4.32.1)
setuptools (39.0.1)
```

- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6):
2.0.0

- **Relevant hardware** (e.g. if reporting a bug about a controller, tell us the brand & name of it):
Lenvo Thnkpad bult-in audio. Works in other programs.
I even made a program using PySDL library to play sound and that works.
Only PyGame fails to play.

**Current behavior:**

My environment  had Pygame 1.9.6 before.  upgraded to Pygame 2.0.0 and after the upgrade mp3 playback is not working.
```
$ python3 audiot.py
pygame 2.0.0 (SDL 2.0.12, python 3.6.9)
Hello from the pygame community. https://www.pygame.org/contribute.html
Traceback (most recent call last):
  File "audiot.py", line 5, in <module>
    mixer.music.load(file)
pygame.error: Unrecognized audio format
```

**Expected behavior:**

```
$ python3 audiot.py
[...]

PR #2712, opened 2021-09-12
Changed linux build options to support mp3 

njbradley:
Hello!
This is an attempt to fix mp3 support on linux with Starbuck5, based on the suggestion by llindstrom in issue #2419. I'm making this pull request to generate wheels to test.


illume:
Unfortunately the Dockerfile is duplicated (because reasons I can't remember). Sorry. So the mpg123 needs to be added to the other one as well.

PR #2714, opened 2021-09-12
Removed necessity to pass by the whole build loop for generating the docs

Grimmys:
I was not able to verify if I didn't broke the standard installation process because I uninstalled my Microsoft build tools for testing the docs generation without it, and I'm not able to install it again (well, it's installed, but not for distutils apparently...).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -313,182 +313,186 @@ if len(sys.argv) == 1 and sys.stdout.isatty():
     if not reply or reply[0].lower() != 'n':
         sys.argv.append('install')
 
+only_docs_generation = len(sys.argv) == 2 and sys.argv[1] == 'docs'
+extensions = []
+data_files = []
 
-# make sure there is a Setup file
-if AUTO_CONFIG or not os.path.isfile('Setup'):
-    print ('\n\nWARNING, No "Setup" File Exists, Running "buildconfig/config.py"')
-    import buildconfig.config
+# make sure there is a Setup file if compilation is needed
+if not only_docs_generation:
+    if AUTO_CONFIG or not os.path.isfile('Setup'):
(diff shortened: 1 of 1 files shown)


Starbuck5:
The CI has spoken, and I also tested locally - but this works.

It does seem like quite a bit of churn and adds a maintenance burden to setup.py.

What if the docs command worked like the cython command? Pseudocode:

```py
if consume_arg(`docs`):
    fullgeneration = consume_arg('fullgeneration')
    subprocess.call docs stuff

    if not sys.argv: # if there are no more arguments, stop execution so it doesn't get to the SETUP file reading parts
        sys.exit()
```

robertpfeiffer:
This changes around quite a lot, and some of these changes feel very *wrong*

robertpfeiffer:
I'd rather you checked for docs only and quit with sys.exit early than putting the rest of the build in a conditional.

Grimmys:
> What if the docs command worked like the cython command? Pseudocode:
> I don't see any reason it needs to be a DocsCommand class through distutils.

Ok if you want we can try this, I just don't know why we have two different systems for parsing commands...

> This changes around quite a lot, and some of these changes feel very wrong

I definitely agree with you, but imo it's the whole `setup.py` file that feel very _wrong_
As you said following the "return early pattern" would be best, but it's might be hard because code reorganization would be needed: some statements after the command should be skipt, but others souldn't.

I will first try Starbuck's solution - that should work - and then yours if needed.
Thank you for your reviews.

Starbuck5 (approved):
I'm digging it. 🎉 

Thanks as well to @rethanon for pointing out this problem.

It will be great to have it so easy for new contributors to build the docs.

ankith26 (approved):
Yeah this PR looks good, thanks! 👍 🎈

PR #2718, opened 2021-09-13
[Needs taking over]: Improve newbie guide

davidpendergast:
- Removed link to Python 2.0 Reference guide
- Removed reference to (nonexistent?) module 'Userrect'
- Removed advice about dirty-rect tracking (this advice is pretty outdated)
- Added new section about outdated / obsolete advice that tends to float around the web (dirty-rect tracking, HWSURFACE)
- Removed section about HWSURFACE
- Updated code examples to python 3 syntax
- Added note about how you have to clear the event queue even if you aren't using it
- Added links to python profilers, and more a more up-to-date source about pygame/python performance


Starbuck5 (commented):
Line 22, supported OS comment: What even is BeOS? Maybe we should just claim Windows, Linux, and Mac.
Line 80, on multiple windows: It's not an SDL limitation now, so maybe we should just delete that whole sentence.

/docs/reST/index.rst needs to be updated, since it describes this as "A list of thirteen helpful tips for people to get comfortable using pygame." - and now it's actually 12.

docs/reST/tut/newbieguide.rst
```diff
@@ -60,8 +62,8 @@ Recognize which parts of pygame you really need.
 Looking at the jumble of classes at the top of the pygame Documentation index
```
Starbuck5, line 62: Documentation doesn't need to be capitalized.

docs/reST/tut/newbieguide.rst
```diff
@@ -386,11 +356,14 @@ good that if your code is still slow, and you've done the things I've mentioned
...
-unwieldy, chances are its speed can be improved, too.  Read over `Python
-Performance Tips`_ for some great advice on how you can improve the speed of
-your code.  That said, premature optimisation is the root of all evil; if it's
-just not fast enough, don't torture the code trying to make it faster.  Some
-things are just not meant to be :)
+unwieldy, chances are its speed can be improved, too.  Read over `Why Pygame is 
+Slow`_ for some deeper insight into why pygame might be considered slower than 
+other frameworks/engines, and what that actually means in practice.  
+And if you're truly stumped by performance problems, profilers like cProfile_ or 
+SnakeViz_ can help identify bottlenecks (they'll tell you which parts of the
```
Starbuck5, line 363: Technically Snakeviz is a visualizer for the profiles generated by cProfile.

davidpendergast:
Thanks, I'll make those changes.

One other thing I was thinking might be valuable is a section on distribution? One thing I notice a lot is people not considering cross-OS compatibility or packaging until they're completely finished with a project, at which point they fire up pyinstaller, a bunch of things don't work (e.g. reading asset files), they try and fail to fix them, and then get frustrated and give up.

It might be nice to suggest some things like:
- using relative paths to reference all asset files
- using os.path / pathlib to keep filepaths os-independent
- (trying to) explain the quirks of pyinstaller --onefile mode

And maybe some other stuff like that?

Starbuck5:
There definitely could be a section about paths. I would recommend relative paths with the slashes like this -> `"/"`. That will work on all operating systems. I frequently see newbies who have heard to use os.path.join, but do end up doing something completely useless with it, and since it isn't actually necessary, I think we shouldn't bring it up here.

PyInstaller is a bit more advanced, and it's something that needs a lot of explanation, so maybe it shouldn't be in the newbie guide either.

If we're brainstorming ideas for sections, it could be good to have some "hello world" code. I remember when I was learning pygame I found the game loop on some random website, and it really helped me. I also see people who don't know `Clock` exists, or who don't understand they have to `flip()`, or who just don't understand a basic pygame loop.

This is my personal brand of minimum viable pygame program:
```py
import pygame

pygame.init()

screen = pygame.display.set_mode((1280,720))

clock = pygame.time.Clock()

while True:
    screen.fill("purple")
    
        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit

    pygame.display.flip()
    clock.tick(144) #framerate
```

docs/reST/tut/newbieguide.rst
robertpfeiffer: "with snakeviz" then.

robertpfeiffer:
> Line 22, supported OS comment: What even is BeOS?

Has anybody tried compiling on Haiku?

Starbuck5:
> Has anybody tried compiling on Haiku?

Let's keep it simple for the newbies guide.

davidpendergast:
I added a new section about software architecture and design patterns, which I think fits really nicely in this and gives a good opportunity to talk about the pygame "base template" and game loops in general in a more abstract way. Also made most of the suggested changes from comments - all that's left now (I think) is:
- removing mention of SDL / multiple windows limitations (missed that one~)
- linking function calls to docs
- running setup.py docs to make sure it all looks right
And I decided to just add a sentence to the 'learn python' section about paths, rather than adding a standalone section. I agree that distribution is probably too complex for this.

robertpfeiffer:
> > Has anybody tried compiling on Haiku?
> 
> Let's keep it simple for the newbies guide.

I'll make this its own issue then. I have no idea how well current Pygame runs on FreeBSD or Haiku.

davidpendergast:
Alrighty, I've made another update, which should address those few remaining issues I mentioned on the last diff and fix the HTML formatting issues (of which there were many).

Starbuck5 (approved):
Looks good to me. 👍 

Thanks for this @davidpendergast!

docs/reST/tut/newbieguide.rst
```diff
@@ -123,93 +127,61 @@ output file had the same pixel format as the input file.  If you're writing a
...
 
- * It only works on some platforms. Windows machines can usually get hardware surfaces if you ask for them. Most other platforms can't. Linux, for example, may be able to provide a hardware surface if X4 is installed, if DGA2 is working properly, and if the moons are aligned correctly. If a hardware surface is unavailable, SDL will silently give you a software surface instead.
+That being said though, there are still plenty of ways to accidentally tank your 
+game's performance with poorly optimized rendering logic.  For example, even on 
+modern hardware it's probably too slow to call ``set_at`` once per pixel on the 
+display surface.  Being mindful of performance is still something you'll have to 
+do.
 
- * It only works fullscreen.
+**HWSURFACE and DOUBLEBUF**
```
illume, line 164: Let's not mention these at all?

docs/reST/tut/newbieguide.rst
```diff
@@ -123,93 +127,61 @@ output file had the same pixel format as the input file.  If you're writing a
...
-background jumping along too much.
-
-A final note -- not every game requires high framerates. A strategic wargame
-could easily get by on just a few updates per second -- in this case, the added
-complexity of dirty rect animation may not be necessary.
-
-
-There is NO rule six.
----------------------
+Be wary of outdated, obsolete, and optional advice.
```
illume, line 130: Against removing this part. Especially since dirty rect optimization is still useful.

davidpendergast: Do you think it's useful to newbies though? Here are my main problems with the section:
- Keeping track of dirty rects is error prone and beyond the capabilities of most newbies, from what I've seen. I worry that telling them to try it will lead them down a path of frustration (e.g. "my game is running is slow, I should try using dirty rects to fix it" -> they struggle to implement it properly and the FPS doesn't even improve).
- Most beginner games can't benefit from dirty rect tracking anyways, due to having a moving camera + background image  (e.g. flappy bird, top-down survival, platformers). And simpler games that could benefit (like snake, puzzle games, pacman etc.) likely don't need it to achieve high performance in the first place.
- The way the section was written made it sound like pygame _can't_ run well without it, which makes pygame sound limited, slow & outdated. Prospective pygame users have asked about the display.update() -> 28 FPS maximum thing (this could be fixed with a rewrite)

I guess I'm just not sure who the target audience of this section would be, and what advice we'd want to convey beyond "this exists, but you probably don't need it?"

docs/reST/tut/newbieguide.rst
ankith26: IMHO, this section does do a good job at clearing any potential confusion, many-a-times newbies learning from older tutorials come across these flags, and are confused as to whether these should be used or not (not just newbies, even some more experienced peeps may sometimes misunderstand these flags as well)
So I think this bit should probably stay, IG

docs/reST/tut/newbieguide.rst
robertpfeiffer: Dirty rect optimisation is not that useful. What might be useful is not re-drawing the whole screen, and instead only erasing and re-drawing a small section, but specifically the dirty rects parameter to update seems like a micro-optimisation these days.

robertpfeiffer:
Now that I think about it, a lot of perf advice like "avoid quadratic collision detection" or "don't use scrolling backgrounds to keep blits low" is situational, and the most important advice is "please benchmark instead of blindly guessing" and "move loading and rendering out of the game loop if you can".

Starbuck5:
Alright, I’d like this to move forward, since it is a needed modernization, so I talked to illume about how to get this ready to go.

The biggest sticking point is dirty rect optimization, illume says it’s still valid in some cases, and doesn’t want it presented as old advice. I think that it should be contextualized as an advanced technique, because beginners often mess it up when they try to do it.

Additionally, it was brought up in that the stuff about HWSURFACE doesn’t particularly belong in a newbies guide. I personally like the way Ghast put it in, but if we’re changing dirty rect that section might just go.

Maybe the dirty rect stuff could be more of a footnote in a performance section, that expands more on profiling.

I think illume and I both have an idea or two about how to get this fully up to par and merged, so either of us might take it over. But I’m also happy to leave it to Ghast, with those notes.

davidpendergast:
Feel free to take this over - it seems like you two have slightly different visions for this guide than I do so I'm probably not the best person to write up those ideas. I've also never personally used (or seen anyone use) dirty-rect optimization so my insight on that is limited. I hope my attempt at this was helpful at least though!

MyreMylar:
Updated the title to indicate that this needs taking over. I think some good progress was made here but it needs somebody argumentative to get it over the line :)

If you want to take this over, either close this PR when you open a new one - or let me know that you want to take it over and I will close it.

**Also:** Please credit the author of this PR in your takeover PR.

MyreMylar:
I will have a go at taking this over.

Issue #2702, opened 2021-08-30
Module init() consistency


Starbuck5:
@rethanon noticed that `mixer.init()` returns `True`, which is undocumented.

I checked some other module inits, and
`display.init()` returns `None`, documented
`freetype.init()` returns `True`, undocumented
`joystick.init()` returns `None`, documented
`scrap.init()` returns `None`, documented (even documents a pygame.error will be raised on fail)
`midi.init()` returns `None`, documented
`font.init()` returns `True`, undocumented

I checked the mixer code, and it can only return `True`. If it fails, it errors, rather than returning `False`. This makes sense. Python programs don't handle errors in this way, C programs do.

Can we bring the errant modules back in line without breaking compatibility too much? Most people just use pygame.init() to do them all at once.

Should we just grit our teeth and correct the docs and types to match the functionality? Or maybe wait to change this in 2.1?

My vote is to change it sooner rather than later, since this "feature" wasn't documented. Instead of changing the docs and types to match the code, we can change the code to match the docs and types.

Starbuck5:
@ankith26 I did some digging, and I'm pretty sure this is a regression from https://github.com/pygame/pygame/commit/740896a2cb2d32df1f8a177345576b20ce1d9b43.

ankith26:
Hmm, are you sure? I tried hard to not break stuff in that commit. I kept the return types the same as how they were before. Can you tell me which init function exactly is broken, and/or a piece of code that behaves differently before and after that commit?

Starbuck5:
```
>>> import pygame
pygame 2.0.1 (SDL 2.0.14, Python 3.8.6)
Hello from the pygame community. https://www.pygame.org/contribute.html
>>> pygame.font.init()
```
prints nothing, i.e. returns `None`

```
>>> import pygame
pygame 2.0.2.dev2 (SDL 2.0.16, Python 3.8.6)
Hello from the pygame community. https://www.pygame.org/contribute.html
>>> pygame.font.init()
True
```

I think it's because you consolidated some functions together. font.c used to have font_autoinit and fontmodule_init, now it is just fontmodule_init, but with the return from font_autoinit.

ankith26:
Ooh yeah I just noticed that, yeah my bad, that is a regression from my commit. Are there any other regressions apart from that? (I hope and think not)

Starbuck5:
> Are there any other regressions apart from that?

It's font.init, mixer.init, and freetype.init

[...]

PR #2719, opened 2021-09-13
Fix regression of init functions returning bool

ankith26:
This was a regression introduced in my older PR, #2428

Fixes issue #2702


Starbuck5 (approved):
Thanks 👍

MightyJosip (approved):
Good job

PR #2723, opened 2021-09-17
Add scale_by, smoothscale_by functions

Starbuck5:
For years, pygamers around the world have been unnecessarily inconvenienced by how it is to just scale a surface **by** some multiple, by some scalar.

This was the best anyone could do:
```py
scale = 2.5
surf2 = pygame.transform.scale(surf, (int(surf.get_width() * scale), int(surf.get_height() * scale)))
```

In 2.0.2.dev2, I changed it so `transform.scale` could take float arguments, so it got simplified to this:
```py
scale = 2.5
surf2 = pygame.transform.scale(surf, (surf.get_width() * scale, surf.get_height() * scale))
```
OR
```py
# This one even fits under Python's line limit suggestion! (As long as it isn't indented at all)
scale = 2.5
surf2 = pygame.transform.scale(surf, pygame.Vector2(surf.get_size())*scale) 
```

But it still seems unnecessarily complex / inconvenient for this common use case.

So after this PR, you can do
```py
scale = 2.5
surf2 = pygame.transform.scale_by(surf, scale)
```

You can also scale_by a different horizontal and vertical scalar by putting in two arguments. And it continues to support the `dest_surface` argument of scale. I also added the same thing for `smoothscale`.

This includes tests, docs, and type stubs.


src_c/transform.c
```diff
@@ -593,6 +592,53 @@ surf_scale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+{
+    PyObject *surfobj;
+    PyObject *surfobj2 = NULL;
+    PyObject *new_args = NULL;
+    PyObject *scale_obj = NULL;
+    PyObject *scaley_obj = Py_None;
+    float scale, scaley;
+    SDL_Surface *surf;
+    int width, height;
+    static char *keywords[] = {"surface", "scale", "scale_y", "dest_surface", NULL};
```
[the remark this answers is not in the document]
Starbuck5: Ah, I forgot to respond 😅

Thank you for the feedback.

If there was `scale_x` and `scale_y`, users might interpret them as defaulting to 1. So `scale_y=5` should stretch vertically 5x and do nothing to the x.

I would use them positional-only, but I was trying to fit with a big push to get pygame to accept kwargs in #808.

I guess anything could be potentially confusing, it would be up to the docs to set things straight.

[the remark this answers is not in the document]
lordmauve: > If there was scale_x and scale_y, users might interpret them as defaulting to 1.

Ok, I agree.

> I was trying to fit with a big push to get pygame to accept kwargs in #808.

Sure, but this seems like a good place to deviate and go positional-only.

Firstly because `scale` is overloaded as discussed. Secondly because the function name is `scale_by(...)`, the question is "scale by WHAT". `scale_by(factor)` reads as English, so does `scale_by(xscale, yscale)`. `scale_by(scale=scale)` doesn't read as well. `scale_by(dest_surface=screen, scale=5)` is right out.

MyreMylar:
Could you not have either a single 'overloaded' keyword argument:

 - `scale` : a single float or int.  
 - `scale`: a tuple of (scale_x, scale_y) or a Vector2

Then based on what comes in, the surface is scaled in the appropriate way?

I had a think, but couldn't come up with a better keyword than scale for this though. So I say we stick with that. 

We are never going to get a great english line read for a function existing in the transform module, the awkwardness is built in there - `scale_by(surface, scale)` is as good as it gets with these foundations.

MyreMylar (changes requested):
Gonna put this on request changes until someone tells me why I'm wrong about merging the keywords into one.

MyreMylar:
shower thought this morning - would `factor` be another potential keyword for scale_by?

```
pygame.transform.scale_by(surface=surf, factor=5.0)
pygame.transform.scale_by(surface=surf, factor=(5.0, 3.0))
```

I've heard people say "scale that up by a factor of 10"

Starbuck5:
Alrighty, I like the sound of that one, I'll make it happen.

docs/reST/ref/transform.rst
```diff
@@ -116,6 +132,23 @@ Instead, always begin with the original image and scale to the desired size.)
...
+
+   Same as :func:`smoothscale()`, but scales by some factor, rather than
+   taking the new size explicitly. For example,
+   :code:`transform.smoothscale_by(surf, 3)` will triple the size of the
+   surface in both dimensions. Optionally, the scale factor can be a sequence
+   of two numbers, controlling x and y scaling separately. For example,
+   :code:`transform.smoothscale_by(surf, (2, 1))` doubles the image width but
+   keeps the height the same.
+
+   .. versionadded:: 2.1.1
```
zoldalma999, line 148: Same here

docs/reST/ref/transform.rst
```diff
@@ -49,6 +49,22 @@ Instead, always begin with the original image and scale to the desired size.)
...
+   | :sg:`scale_by(surface, factor, dest_surface=None) -> Surface`
+
+   Same as :func:`scale()`, but scales by some factor, rather than taking
+   the new size explicitly. For example, :code:`transform.scale_by(surf, 3)`
+   will triple the size of the surface in both dimensions. Optionally, the
+   scale factor can be a sequence of two numbers, controlling x and y scaling
+   separately. For example, :code:`transform.scale_by(surf, (2, 1))` doubles
+   the image width but keeps the height the same.
+
+   .. versionadded:: 2.1.1
```
zoldalma999, line 64: Version number is outdated

src_c/transform.c
```diff
@@ -593,6 +617,45 @@ surf_scale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    SDL_Surface *surf;
+    int width, height;
+    static char *keywords[] = {"surface", "factor", "dest_surface", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "OO|O", keywords, 
+                                     &surfobj, &factorobj, &surfobj2))
+        return NULL;
+
+    if (!_get_factor(factorobj, &scale, &scaley)) {
+        printf("returning here\n");
```
zoldalma999, line 637: Debug print left in

zoldalma999 (changes requested):
Left some notes of some minor stuff, that should be fixed, but apart from those, this pr seems good
to me. Will be a really handy function to have!

docs/reST/ref/transform.rst
Starbuck5: Thanks

src_c/transform.c
Starbuck5: Oops, glad you caught that.

Starbuck5:
I rebased it, fixed the issues Zoldalma pointed out, and now it's failing everywhere. I'll fix it tomorrow.

MyreMylar (approved):
LGTM 👍

MyreMylar:
Just needs @zoldalma999 now!

zoldalma999 (approved):
LGTM!

Matiiss:
I'm not sure how useful a `dest_surface` argument is, you'd need to know the size of the scaled surface to be able to use the destination surface since the sizes have to match.

Starbuck5:
>  I'm not sure how useful a dest_surface argument is, you'd need to know the size of the scaled surface to be able to use the destination surface since the sizes have to match.

I've thought similar, but I think it's worth it to be consistent with `transform.scale`. One other idea: perhaps the dest surface shouldn't need to be the exact size of the scale, just have enough space on the `x` and `y`.

PR #2728, opened 2021-09-18
Changed angle_to() and added angle_between()

Muxelmann:


Muxelmann:
As discussed [here](https://github.com/pygame/pygame/issues/2729#issuecomment-922385145), it makes more sense to not alter the existing functionality, but add new one - if necessary.

PR #2731, opened 2021-09-23
Fix playmus example

cassiobotaro:
Issue: pygame.error: pygame is not initialized

Add pygame initialization when create window.

### DIFF
--- a/examples/playmus.py
+++ b/examples/playmus.py
@@ -39,6 +39,7 @@ class Window(object):
         if Window.instance is not None:
             return Window.instance
         self = object.__new__(cls)
+        pg.init()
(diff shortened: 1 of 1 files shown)


Starbuck5:
Thanks for coming in and opening a pull request!

However, the underlying issue is fixed in 2.0.2.dev2, so pygame.init() is not needed.

Now that I'm looking at this though, there are some other problems. There's a typo - "faid" should be "fade." There also seems to be problems with rewinding and pausing. On 2.0.2.dev2, pausing just closes the window- that shouldn't happen. In 2.0.0, where pausing does work, rewinding doesn't seem to actually work.

Issue #2734, opened 2021-09-26
Playmus example is broken


Starbuck5:
@cassiobotaro put in a PR, #2731, to fix the playmus example being broken in 2.0.1.

The fix they proposed is actually necessary for 2.0.1, but it looks like the underlying issue has been fixed by now in 2.0.2.dev.

But there seem to be no shortage of other things wrong with this example:

> Now that I'm looking at this though, there are some other problems. There's a typo - "faid" should be "fade." There also seems to be problems with rewinding and pausing. On 2.0.2.dev2, pausing just closes the window- that shouldn't happen. In 2.0.0, where pausing does work, rewinding doesn't seem to actually work.

It just needs some TLC, some tender loving care.

rethanon:
@Starbuck5 
The issue with the program exiting when the music is paused is due to a change in pygame 2.0.1 which updated `pygame.music.mixer.get_busy()` to return False when music is paused, in previous versions it returned True (and playmus.py closes when no music is playing).

I should be able to give this a bit of love to get it working properly.

PR #2738, opened 2021-09-28
Fix playmus.py example (broken in 2.0.1)

rethanon:
This PR is to fix an issue with the bundled example file playmus.py which stopped working correctly as of pygame 2.0.1 due to a change in returning False for `pygame.mixer.music.get_busy()` when music is paused (previously returned True).

Also tidied up the wording a little bit and fixed a typo.

Issue was raised [#2734 ](https://github.com/pygame/pygame/issues/2734)

### DIFF
--- a/docs/reST/ref/music.rst
+++ b/docs/reST/ref/music.rst
@@ -80,6 +80,7 @@ can crash the program, ``e.g``. Debian Linux. Consider using ``OGG`` instead.
    | :sg:`rewind() -> None`
 
    Resets playback of the current music to the beginning.
+   .. note:: rewind does not support all music file formats. This is a limitation in the SDL_mixer, which states rewind only supports ``MOD``, ``OGG``, ``MP3`` and Native MIDI files. 
 
    .. ## pygame.mixer.music.rewind ##
 
--- a/docs/reST/ref/music.rst
+++ b/docs/reST/ref/music.rst
@@ -80,7 +80,9 @@ can crash the program, ``e.g``. Debian Linux. Consider using ``OGG`` instead.
    | :sg:`rewind() -> None`
 
    Resets playback of the current music to the beginning.
-   .. note:: rewind does not support all music file formats. This is a limitation in the SDL_mixer, which states rewind only supports ``MOD``, ``OGG``, ``MP3`` and Native MIDI files. 
+   .. note:: rewind does not support all music file formats. This is a limitation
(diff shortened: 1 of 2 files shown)


Starbuck5 (approved):
Thanks, this is great! 

Think it would be prettier with some different colors? The blue on white isn't the prettiest on the eyes.

Starbuck5:
If I play a .wav I can't rewind, but if I play a .mp3 I can?

Also another thing I've found that isn't ideal- you can't close the window when in a paused state.

rethanon:
Thanks I'll resubmit with updated color scheme and I've fixed the being able to quit when a track is paused. I'll have a look at not being able to rewind a wav but I'm getting the same, could be a deeper issue causing that one.

rethanon:
@Starbuck5 I checked the SDL_Mixer docs and it states rewind only works for MOD, OGG, MP3, and Native MIDI. Their website seems to indicate the latest version of SDL_Mixer is version 2.0.4 yet the docs don't appear to have been updated since 2009 (they refer to SDL_mixer version 1.2.10, dated November 2009!)

Starbuck5:
Thanks for all the investigation! The SDL_Mixer docs are old but it has a very stable interface throughout the years. However, sometimes you have to dig into the SDL_Mixer source directly to see how things fit together. Maybe we should add a note that WAV doesn't support rewind to the mixer.music docs.

rethanon:
@Starbuck5 I've fixed this so that playmus.py will use rewind for MOD, MP3 and WAV files, and for anything else will restart (using `pygame.mixer.music.play()`. SDL_mixer docs say Native MIDI files are supported but I can't seem to find what this means. I tested with a standard `.mid` file and rewind wasn't working for that so I thought it better to just use rewind for the 3 main file types.

I had a go at updating the docs for `pygame.musix.mixer.rewind()` but I still have that issue where I can't build the docs locally to test them so could you check them as well please?

Starbuck5:
> I still have that issue where I can't build the docs locally to test them so could you check them as well please?

That's because you're using an old version of main, from like August.

rethanon:
> > I still have that issue where I can't build the docs locally to test them so could you check them as well please?
> 
> That's because you're using an old version of main, from like August.

Ah yeah, I was able to check the docs built correctly on my Mac machine instead.

rethanon:
Closing this so I can rebase code and will then raise a new PR

PR #2740, opened 2021-10-02
Make windows builds function again?

Starbuck5:
Integrates 3 files from download that are needed for windows build, probably.

Problem is that the download keeps failing because the checksum doesn't match. But when I download the file myself, the checksum is fine. But if the download is automated, the checksum is wrong. WTH.

This problem has happened in the past with the pygame.org downloads, and it went away randomly after a week, but I want to release soon, so we can't just have broken windows builds for no determinable reason for an indeterminate amount of time.


illume:
I added a mirror in this PR instead: https://github.com/pygame/pygame/pull/2743

Starbuck5:
Thanks, I'm glad windows builds function again!

Hopefully it doesn't have any intermittent problems with pygame.org like has happened in the past.

PR #2741, opened 2021-10-02
Add `rumble` and `stop_rumble` methods for joysticks and controllers

zoldalma999:


illume:
Very cool!

Would it be possible to add it to one (or some) of the examples/ ?

PR #2742, opened 2021-10-03
Build mac dependencies using our manylinux scripts

illume:
A little bugfix for building mac dependencies with the manylinux scripts.

Things left to package:
- [x] glib (for fluidsynth) illume
- [x] libsndfile (for fluidsynth) illume
- [ ] brotli (for freetype) illume
- [ ] *harfbuzz (for freetype optional/later)
- [ ] bzip2 (for freetype) illume
- [x] libtiff (for webp, and sdl_image) illume
- [x] zlib (because apple zlib is too old for libpng) ankith26
- [x] readline (use apple OS provided one not homebrew/other)
- [ ] libmodplug (for sdl_mixer)
- [ ] *libjpegturbo (for sdl_image optional/later)


Starbuck5:
These are some weird test failures.

One dependency thing I just learned about is https://github.com/rbdannenberg/portmidi, a somewhat maintained offshoot of portmidi by someone who claims to be the original author. See https://github.com/mixxxdj/portmidi/issues/8#issue-1013516723

This probably contains what we need, which is a portmidi that doesn't try to link against MacOS 10.5, and with other maintenance fixes we might not know we need yet.

illume:
Good to see some portmidi activity, and good to see it trying to move to the github repo :) I guess we wait for a new release from there.


---

Warnings from the build:

```
ld: warning: dylib (/usr/local/lib/libtiff.dylib) was built for newer macOS version (10.15) than being linked (10.9)
ld: warning: dylib (/usr/local/lib/libbrotlidec.dylib) was built for newer macOS version (10.15) than being linked (10.9)
ld: warning: dylib (/usr/local/opt/gettext/lib/libintl.dylib) was built for newer macOS version (10.15) than being linked (10.9)
ld: warning: dylib (/usr/local/Cellar/libsndfile/1.0.31/lib/libsndfile.dylib) was built for newer macOS version (10.15) than being linked (10.9)

ld: warning: dylib (/usr/local/Cellar/glib/2.70.0/lib/libgthread-2.0.dylib) was built for newer macOS version (10.15) than being linked (10.9)
ld: warning: dylib (/usr/local/Cellar/glib/2.70.0/lib/libglib-2.0.dylib) was built for newer macOS version (10.15) than being linked (10.9)

/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/zlib.h:1917:32: note: 'inflateValidate' has been marked as being introduced in macOS 10.13 here, but the deployment target is macOS 10.9.0
```

Build these ourselves?
- glib (for fluidsynth)
- libsndfile (for fluidsynth)
- brotli (for freetype)
- *harfbuzz (for freetype optional/later)
- bzip2 (for freetype)
- libtiff (for webp, and sdl_image)
- zlib (because apple zlib is too old for libpng)
- readline (use apple OS provided one not homebrew/other)
- libmodplug (for sdl_mixer)
- *libjpegturbo (for sdl_image optional/later)

Testing for homebrew linked libraries could be possible (`find . -name \*.so -exec otool -L {} \; | grep Cellar`).
Somehow removing homebrew include paths might be useful (maybe rm-ing them).


Some configure options? Could add a CONFIG_ARGS env var so linux/mac can do different things.
- `sdl_image --disable-imageio --disable-jpg-shared --disable-png-shared --disable-tif-shared --disable-webp-shared`
- freetype `--without-harfbuzz` (because harfbuzz is not supported by pygame yet).


To detect linux/mac? `uname -a | grep Darwin`, or pass an ENV var down from the top.

Starbuck5:
I would also like libmpg123 for SDL_mixer mp3 support.

SDL_Image works with libjpegturbo? I assumed it would be linked to normal libjpeg.

I'm still bummed that installing from brew with the build from source flag didn't recognize the export targets. That would've been SUCH an easy way to do this.

ankith26:
Heh, was trying to review but because of the force pushed it could not be submitted.

(in reference to the CI env part)
This part should be replicated a few lines below in the CIBuildWheel2 section as well

This is my bad, when I wrote the original code for this, I made it repetitive, I did not know that one `env:` section at the top of the section would suffice. RN, it needs the same bit of code in 2 places, it probably needs to be fixed ;)

But I see that now you have fixed it :)

illume:
@ankith26 thanks, yeah spotted that one after I saw it fail :) I put my name next to the deps I'm going to work on. I'll be rebasing/forcing this branch... so if you want to add the deps in another branch maybe go for it. Otherwise I'll probably get through this list in some hours.

ankith26:
Yeah sure! :) while I don't exactly understand how this works, I can still try to help out and see how it is done in other places, and learn my way from there ;)

Yeah I will try opening a new branch from this branch and go through the deps you are not doing RN

illume:
I'm working on libtiff

ankith26:
Wow, nice progress so far! I am thinking imma take on libmodplug and zlib

illume:
The generated wheel ran on a couple of test machines.

ankith26:
Hmm, I can also try adding harfbuzz and libjpegturbo as well, I guess.

ankith26:
I can fix that in my next commit

illume:
Still trying to get fluidsynth going. Got libsndfile working. I used an older glib that provides an autotools configure script, so we don't depend on meson and ninja. But glib requires gettext... so now trying to add that even though it's a real pain. 

fluidsynth is really nice for playing midi on mac, so I guess it's worth it? lol.

ankith26:
Wow, nice! 🥳 🎉 

Glad to see a release blocker being cleared, this was a really tedious PR to make, so cheers for that, illume! :) 

But I still see auditwheel erroring on the wheel repair part, maybe that needs to be fixed?

> the mac deps do take 22+ minutes or so of build time... oooof. So I added that to the list of 'future work' in the buildconfig/macdependencies/README.rst We could also enable homebrew deps for testing normally.

IMHO this should be fixed right in this PR before getting merged. But DW, after this PR is done from your end, I could make a commit onto this PR with a simple fix for this, that is caching `/usr/local`, so that the builds are not built every time (unless something in the buildconfig folder changes, in which case it will have to rebuild stuff). Generally this sort of thing would be a not-so-good idea, but in this case, doing this makes sense IG
Like I already said, this is an easy change, and will not block this PR in any way, so I would like to do it in this PR itself, unless you strongly wish otherwise ;) 

Also, something I feel which is a nice to have for v2.0.2 is universal2 wheels, considering the fact that arm macs are getting much more common these days. But that OFC, is something for another PR, and is something I can try taking on.

illume:
Giving up on harfbuzz for now. It's taking too long, and it's not a regression... so now it's in the do later pile.

ankith26:
Also I'd like to give dep caching a shot soon (maybe in another PR, because I don't wanna block this)

Also if the caching works as I think it will, we would not need to return to homebrew for deps for non-release versions, IMHO.

illume:
> Also I'd like to give dep caching a shot soon (maybe in another PR, because I don't wanna block this)
> 
> Also if the caching works as I think it will, we would not need to return to homebrew for deps for non-release versions, IMHO.

Yeah, cool. Caching of this is definitely needed!

Here's some notes on caching thoughts I've had (not sure anything is super useful). 
I guess you have a plan already.

- use actions/cache (see README.md in `buildconfig/macdependencies` for notes
- modify scripts to allow them to skip download if the folder is there (see build-freetype.sh)
- have them build in a separate folder (something like .pygame-dependencies)
- we use cmake, and autoconf for building currently... but both have a '--prefix' way of installing... maybe worth looking at.
- autotools configure has a way to skip configure if it's there already. It's a bit fragile, because it uses dependency caching too.
- some deps could be built in parallel. However some rely on each other. Maybe putting build script into a makefile would allow it to be fairly easy done in parallel (with `make -j4`).
- /usr/local caching has been fragile when tried in the past. But I guess you've tried something with the previous PR way of doing it.

PR #2743, opened 2021-10-03
Use pygame.org link for jpeg, because it's blocked on appveyor

illume:


illume:
It works here: https://ci.appveyor.com/project/pygame/pygame/builds/41001256/job/i4981wku6rcyu6u5

PR #2744, opened 2021-10-04
Add a requirements-dev.txt

bydariogamer:
Added a `requirements-dev.txt` file to the root directory to help developers to setup python packages for building `pygame`:
- `numpy` is optional for a basic pygame install, but it is probably needed for development
- `setuptools` is needed for setup.py
- `wheel` is needed for building wheels

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,4 +2,4 @@ setuptools
 wheel
 numpy >= 1.6.2
 requests
-Sphinx
+Sphinx==3.5.4
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,3 +1,5 @@
 setuptools
 wheel
 numpy >= 1.6.2
+requests
+Sphinx
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,0 +1,3 @@
+setuptools
+wheel
+numpy >= 1.6.2
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -3,3 +3,4 @@ wheel
 numpy >= 1.6.2
 requests
 Sphinx==3.5.4
+cython
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,6 @@
-setuptools
-wheel
-numpy >= 1.6.2
-requests
-Sphinx==3.5.4
-cython
+setuptools      # needed to setup
+wheel           # needed to build wheels
+numpy >= 1.6.2  # optional in a basic pygame install, probably needed for development
+requests        # might be needed to download some stuff
+Sphinx==3.5.4   # needed to build docs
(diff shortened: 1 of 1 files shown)


illume:
I like the spirit of this PR, but I'm not for adding other files to the root of the repository.

Is there some other way you could do this?

Starbuck5:
It could be hosted in the buildconfig folder maybe?

Added to the README?

Other resources will definitely need to be altered to point to this as well.

ankith26:
I understand that keeping the root dir of a project clean is a really good idea, but there are some files that need to be at the root for more clarity of the project. IMHO `requirements-dev.txt` is one such file, many projects in the python community have such a file, and it helps a tiny bit for new contributors getting started when they see the file right at the base dir.

bydariogamer:
I though it was an important file now that pygame has a growing number of contributors and I think the root of the project is the right one, but tell me where to move it if you really don't like it here.

Also, Cython missing! I will add it inmediately.

bydariogamer:
Also, @Grimmys might want Sphinx version updated now that #2666 got merged.

Grimmys:
Yes, true, it would be best to specify a more recent version of Sphinx (i.e. 4.1.2, it was the last version some weeks ago but damn Sphinx is already at 4.2).

In fact, it's ok for people to use Sphinx 3.X if they just need to have **access** to the docs.
However, if they want to **modify** the docs, by editing CSS or even just a RST file, I highly recommend them to have the last version supported by pygame.

Also, I want to add my vote on the fact that is not bad to add this file at the root of the project.
I mean, it's a really common file that can be found at the root of many popular projects, and I think that keeping a root directory clean is different than having a strict policy of not adding a single more file in it.
Being able to launch the command `pip install -r requirements-dev.txt` directly at the root folder is really sweet imo. 🙂

Starbuck5:
@Grimmys all my PR testing on your Sphinx 4 PR was on Sphinx 4.2. I think we're fine inside of major versions.

MyreMylar:
What about adding this:

```
        'extras_require': {
            'dev': [
                'setuptools'      # needed to setup
                'wheel'           # needed to build wheels
                'numpy >= 1.6.2'  # optional in a basic pygame install, probably needed for development
                'requests'        # might be needed to download some stuff
                'Sphinx~=4.1'     # needed to build docs
                'cython' 
            ]
        }
```

in one of the `setup.py` blocks so you could install development dependencies with something like:

```
pip install -e .[dev]
````

I've seen something similar bandied around in a few places and we could add some documentation on it to the wiki as well if it works out.

ankith26:
I'm +1 for the idea posted above

Starbuck5:
Thanks for putting this in bydario, and thanks for the discussion everyone. ❤️ 

I'm going to close this PR now, because I don't think it's going anywhere.

PR #2745, opened 2021-10-04
Fix playmus.py example (broken in 2.0.1)

rethanon:
This PR is identical to PR [#2738](https://github.com/pygame/pygame/pull/2738) which I closed as I needed to update base code to prevent automated build test failures.

=====

This PR is to fix an issue with the bundled example file playmus.py which stopped working correctly as of pygame 2.0.1 due to a change in returning False for pygame.mixer.music.get_busy() when music is paused (previously returned True). 

Color scheme updated to be easier on the eye
Docs updated to reflect that rewind() is only supported for certain file types
Tidied up the wording in playmus.py a little bit and fixed a typo

Related to a closed issue that was raised under [#2734](https://github.com/pygame/pygame/issues/2734)

### DIFF
--- a/docs/reST/ref/music.rst
+++ b/docs/reST/ref/music.rst
@@ -81,8 +81,8 @@ can crash the program, ``e.g``. Debian Linux. Consider using ``OGG`` instead.
 
    Resets playback of the current music to the beginning.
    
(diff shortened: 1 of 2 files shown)


illume (approved):
🎉 Thanks.

Starbuck5 (approved):
Thanks for figuring this one out @rethanon!

PR #2746, opened 2021-10-05
Remove restriction on rendering 4-byte characters on pygame.font part 2 of 2

illume:
continued on from https://github.com/pygame/pygame/pull/2627

Disable old test.


illume:
The SDL1 build fails to compile src_c/surface_fill.c but doesn't say why. Doesn't seem related to this PR. It doesn't give a reason. https://github.com/pygame/pygame/pull/2746/checks?check_run_id=3800737378#step:6:407

ankith26 (commented):
This PR is very useful indeed!
Hehe, Avaxar had asked me on discord to continue on his PR and I got some work done locally on this a while back, but forgot to finish up and push. Anyways, no probs! We can go ahead with this PR, and the `#define` of `SDL_TTF_VERSION_ATLEAST` when it is missing is nice!

Before this PR is merged, it would need a doc update to reflect this change, and also I was thinking, a way to check at runtime if `pygame.font` module supports ucs4 range would be something useful to people

illume:
How about I add the doc update, and for now detecting usc4 can be done with a pygame version check of >= 2.0.2?

ankith26:
Well uhm... There is a chance that someone has pygame 2.0.2 but not SDL_ttf v2.0.15. So users checking for `2.0.15` is a not-so-good idea. Maybe this feature can wait for 2.0.3? but IDK

I would certainly prefer to have like an attribute check, something like
```
if hasattr(pygame.font, "ucs4"):
    do_ucs4_stuff()
else:
    do_non_ucs4_stuff()
```

Also this way, the checking code remains compatible with older pygame versions, because introducing a new check function would ofc not work on older pygame versions

ankith26:
Mind if I push to this branch and finish this PR? Or @illume do you want to continue on this?

PR #2749, opened 2021-10-06
Temporary hack to setup mac deps

ankith26:
This is intended to be a temporary solution to our mac deps situation, uploading locally installed dependencies from @Starbuck5  's mac


illume (changes requested):
Why?

ankith26:
Oh, and if your manylinux scripts mac dep PR works, then that would totally be a much better solution than this ;)

illume:
Do you maybe want to help packaging some of these dependencies instead?
(as the other `buildconfg/manylinux/docker_base` ones are done)
https://github.com/pygame/pygame/pull/2742

Let me know if you want to pick a couple so we don't do the same ones?

Starbuck5:
Yeah, I guess we should just close this.

I was thinking the one advantage of this strategy would be easier universal2 binaries, but I suppose the other strategy can be souped up to support that.

PR #2762, opened 2021-10-10
GH Actions make /usr/local cache for mac deps

ankith26:
Cache the deps built in /usr/local because they can take really long build times


ankith26:
After like a million force pushes, this PR is finally ready for review now! 

Also to explain a bit on what this PR does, mac deps take a long while to build from source. This PR adds caching of the built deps so that we don't waste build time building these again and again. The deps are built only when something in the `buildconfig/macdependences` or `buildconfig/manylinux-build` dir changes, and this dep build runs on a separated job, because we don't want the deps to be built over and over again in the build matrices running on parallel. 

The original idea of this PR was to cache the entire `/usr/local` dir, but it quickly became clear that it would be problematic. So this PR takes a slightly different approach, the manylinux build scripts also install the built files onto a separate cached dir. This is better than caching `/usr/local`  because `/usr/local` has tons of stuff we don't need to cache, whereas this cached dir will only have the built deps. Also, this way the cache is more resilient, and would not break when mac versions change (because cached deps are built to work on any 10.9+ mac)

This PR saves about 25-35 minutes of build time on mac runners

Issue #2326, opened 2020-11-13
Pygame 2 Deprecation: Tracking stats


MyreMylar:
I mentioned a few times in various other issues related to things we can't do while we maintain support for python 2 that it would be useful to begin tracking download stats to given an idea of how many of pygame's users are still using python 2. After all, there would be no point us continuing to support python 2 if almost no-one was still using it.

I think a reasonable usage proxy for python 2 usage is downloads from PyPi (the place where packages come form when you do `pip install pygame`). Not everyone gets pygame this way, but a lot of our users do and probably a decent cross section as pip is available on every platform and seems to be widely used these days.

Luckily for us there is a site that specifically tracks pypi download stats already. And it also has nice graphs which we can look at and paste into this issue periodically. 

Here is the relevant URL:
https://pypistats.org/packages/pygame

And here is the relevant graph:

![image](https://user-images.githubusercontent.com/13382426/99096645-6d2ba680-25ce-11eb-984f-28aaf474b4af.png)

As you can see it looks like our current python 2 downloads are around 20% of the total. It looks like basically all python 2 downloads are 2.7 in case anyone was concerned about that. Python 2.7 is the second most used minor version of python (it competes with several well used minor versions of python 3).

Some of these downloads will be people, including us, downloading pygame on 2.7 for their CI tests (or just local testing  in my case) but that definitely doesn't account for all of them. I would propose that we don't think about dropping python 2 support until the graph above has python 2 usage hovering below 5% for at least three months.

I will try to post a new version of the graph in this thread once every three months (or more regularly if it seems interesting), but anyone else should feel free to update it as well. There are some changes afoot in the Linux and OSX worlds that should start to see python 2 installs declining in 2021 and beyond.

ankith26:
Fun fact: A good number of python 2 pip installs are either typo (missing pip3 with pip) or newbies who don’t quite get the difference between pip and pip3 (I was in that category for a while :) I had done pip install of many stuff and they didn’t work, then I got to know)

ankith26:
I guess the title should have been **python 2** deprecation??
**pygame 2 deprecation** kinda gives the impression that pygame 2 is getting deprecated

MyreMylar:
Yes, that was the intended title. Oops.

illume:
Good idea.

[...]

PR #2764, opened 2021-10-10
setup.py: Add metadata, including python_requires 2.7

illume:
@takluyver made a good point in https://github.com/pygame/pygame/issues/2326 about python_requires.

Took the opportunity to add a few more things, looking at the pillow and numpy setup.py meta data.


illume:
For python 2.7+ and 3.5+ I think it's like this:
`python_requires='>=2.7,>=3.5'`

https://stackoverflow.com/questions/44660448/using-python-requires-to-require-python-2-7-or-3-2

https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

Seems to be supported in setuptools from Jul26 2016: https://setuptools.pypa.io/en/latest/history.html?highlight=python_requires#v24-2-0

illume:
```
$ twine check dist/pygame-2.0.2.dev3.tar.gz 
Checking dist/pygame-2.0.2.dev3.tar.gz: PASSED
```

takluyver:
Yes, unfortunately there's no concise syntax (as far as I know) for '2.7 or 3.5 and above'. The version restrictions are effectively AND-ed together. So you have to say '2.7 or above, but not 3.0, 3.1, 3.2, 3.4'. Obviously this gets much cleaner when you drop Python 2.7.

PR #2767, opened 2021-10-10
Fix case of CoreMIDI

jmroot:
This could cause build failures when using a case-sensitive filesystem.

### DIFF
--- a/buildconfig/config_darwin.py
+++ b/buildconfig/config_darwin.py
@@ -147,7 +147,7 @@ def main(sdl2=False):
 
 
     DEPS.extend([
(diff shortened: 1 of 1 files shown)


jmroot:
Using CoreMIDI is preferable on macOS anyway, not least because unlike portmidi it's actively maintained.

jmroot:
Adding CoreMIDI directly is an odd way of fixing a problem due to what I assume is the lack of a libporttime though. I'll see if there's a better way.

jmroot:
It's not actually much complexity: 2 lines added, 2 lines changed, 1 line removed. I'll update the other PR as you suggest; feel free to close this one.

PR #2768, opened 2021-10-10
config_darwin.py improvements

jmroot:
The actual code of frameworks shipped with the OS is no longer present as a file in the filesystem (it's only in the shared cache).


jmroot:
Updated to remove unnecessary framework dependencies, prompted by discussion in #2767.

PR #2782, opened 2021-10-20
Build arm64 mac wheels on CI

ankith26:
This PR adds `arm64` wheels for the Apple Silicon M1 chips that run MacOS 11.0 and above. It is to be noted that these wheels should still be considered beta quality, and need more testing.

### The gist of this PR
- Adds arm64 mac wheels on python versions 3.8, 3.9 and 3.10 (older python versions don't support arm64 macs themselves)
- The above is achieved by cross compiling our deps to arm64 on CI intel runners (in the future it is a good idea to port this to use native mac arm runners when github actions CI adds those. That will probably take a few months, or maybe even a year or two, so this PR is a good temporary solution) 
- Adds `zlib-ng` and `libjpegturbo` to serve as drop-in replacements for `zlib` and `libjpeg` respectively. The rationale behind this change is that both libraries have much better performance than their predecessors, and also `libjpeg` has arm64 compile issues
- Fixes `SDL_mixer` build to actually use `fluidsynth`

### Things to discuss
Stuff I need opinions on, and things that need some more discussing
- `zlib-ng` and `libjpegturbo` are only added on mac builds RN. Maybe they should be added on manylinux builds as well?
- I had to switch some builds that run on mac to use `cmake` to easily solve some build problems on mac. To not risk breaking any `manylinux` stuff, I put conditionals to keep using `autoconf` on that. Should this be removed and manylinux builds use cmake on these? Either way, I have tried to make sure not to break any manylinux build config stuff in this PR, but maybe it would still need testing just to be extra sure

### Things to do
- Get M1 mac users help in testing the wheels. RN the CI cannot even run tests on arm64 builds, so we need people to manually verify that the wheels work
- Investigate more optimisation flags, stuff like arm_neon that can also be used on M1 macs (something for future PRs)
- Maybe release pygame 2.0.3 with these wheels, while also explicitly marking these as beta quality, to get wider testing and more bug reports that can be worked on. Even if the wheels are not fully working, it is still better having almost working wheels than not having wheels at all


illume (approved):
👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍👍

Lots of thumbs for all the pain this must have been to do.

Issue #2796, opened 2021-10-24
New layer groups


fortwoone:
**Description**

I suggest we add new types of sprite groups :
- LayeredFilter : would be able to block sprite addition based on their type ;
- AutomatedLayered : each subclass of this group would have a class attribute `sprite_table` (a dict containing pairs of str and ints, each key being the name of a sprite class and the mapped integer the corresponding layer) to which it would refer when adding sprites.

Moreover, it could also be interesting to create a `NoSpriteLayer` singleton in case someone makes sprites for certain uses and don't want these sprites to be rendered (e.g., adding a sprite of that kind into a group to prevent the player from moving).

fortwoone:
I used some subclasses of current LayeredUpdates that behave similarly in my game, definition here : [http://github.com/fortwoone/Oracle-Project/blob/main/oracle_project/game.py](url)

PR #2801, opened 2021-10-27
Added two new sprites groups, reworked LayeredUpdates

fortwoone:
cf. issue #2796 to see what it's about


illume (changes requested):
Nice one.

So what's next? ...

Could you please add an example of how to use this?

This way it would be easier for people to see what this does. So I think the example would be the next step before doing other things required to get this merged.

- example in examples/
- docs in the .rst file
- types in the sprite.pyi file
- tests in sprite_test.py

fortwoone:
> 
> 
> To make a new style class in python2, you need to inherit from object explicitly.

Oh I didn't know this



> 
> 
> Nice one.
> 
> So what's next? ...
> 
> Could you please add an example of how to use this?
> 
> This way it would be easier for people to see what this does. So I think the example would be the next step before doing other things required to get this merged.
> 
>     * example in examples/
> 
>     * docs in the .rst file
> 
>     * types in the sprite.pyi file
> 
>     * tests in sprite_test.py

Well, for the example, it's very complicated, and I have no idea on how to explain this. The docs could be a better way to describe my additions. Still, I'm sure that somehow I will work out a way to make an example for my additions.

MyreMylar:
I think the changes to LayeredUpdates will have a performance impact when you have a lot of sprites as it is filtering the sprite list every frame in `sprites_to_draw()`. 

I think the current approach is just to use 0 width surfaces for invisible surfaces, IIRC. So ideally any change would beat that, but I think the list for drawing should at least be filtered on adding/removing rather than on every frame - especially as visibility is determined by Sprite class, not dynamically.

fortwoone:
Well, OK, I will fix that

Le dim. 7 nov. 2021 à 21:54, Dan Lawrence ***@***.***> a
écrit :

PR #2802, opened 2021-10-30
Move to generate Cython code at build time

Starbuck5:
I don't expect this to be really correct, since I don't know what I'm doing with this stuff.

But this does work, generating wheels with _sdl2 stuff without having thousands of lines of pregenerated C sitting around.

Of course, the rationale for this is to reduce merge conflicts and huge diffs of Cython generated code. It will also stop the need to have PRs explicitly just regenerating the Cython generated code when new versions of Python come out, for example.


Starbuck5:
Nothing to see here

PR #2805, opened 2021-10-30
Fix selection of manylinux to build

illume:
So that more recent manylinux2010 and 2014  wheels are made in
addition to the older manylinux1 builds.

### DIFF
--- a/.github/workflows/manylinux.yml
+++ b/.github/workflows/manylinux.yml
@@ -33,12 +33,14 @@ jobs:
       fail-fast: false  # if a particular matrix build fails, don't skip the rest
       matrix:
         arch: ['x64', 'x86', 'aarch64']
-        tag: ['-', '-manylinux-'] # one tag for CPython builds, and another for PyPy
+        tag: ['-manylinux1-', '-manylinux2010-', -manylinux2014-]
 
         exclude:
-          # no pypy builds on aarch64 for now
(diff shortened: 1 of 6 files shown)


.github/workflows/manylinux.yml
```diff
@@ -33,12 +33,14 @@ jobs:
       fail-fast: false  # if a particular matrix build fails, don't skip the rest
       matrix:
         arch: ['x64', 'x86', 'aarch64']
-        tag: ['-', '-manylinux-'] # one tag for CPython builds, and another for PyPy
+        tag: ['-manylinux1-', '-manylinux2010-', -manylinux2014-]
```
ankith26, line 36: minor nitpick here, maybe the leading and trailing hyphen `-` must not be in the `matrix.tag` rather it can come a few lines below, where `make pull${{ matrix.tag }}${{ matrix.arch }} wheels${{ matrix.tag }}${{ matrix.arch }}` should become `make pull-${{ matrix.tag }}-${{ matrix.arch }} wheels-${{ matrix.tag }-}${{ matrix.arch }}`

This is so that the build name looks neater :)

ankith26 (approved):
Nice work, this PR is good! 🎉 🥳

.github/workflows/manylinux.yml
illume: Good point. I'll fix that after the release :)

PR #2811, opened 2021-10-31
Fix autoimported classes having different types in type hints

### LINKED ISSUE #2809: Autoimported classes have different types in type hints
It seems that pygame is subclassing the autoimported types in stubs instead of just importing them:

https://github.com/pygame/pygame/blob/0a238d271146f2f503790e251a70023434286e89/buildconfig/pygame-stubs/__init__.pyi#L38-L45

Some people already have problems with this:

> hey folks, I'm running into some typing issues but I'm not sure if this is the right place to ask for help. I use `mypy` and it is complaining when I use `pygame.Surface` instead of `pygame.surface.Surface` as my types. Are these meant to be different types or is `pygame.Surface` supposed to be a shortcut to `pygame.surface.Surface`? I have the same issue with `pygame.Vector2` vs `pygame.math.Vector2`.
> 
> ```
> Incompatible types in assignment (expression has type "pygame.math.Vector2", variable has type "pygame.Vector2")
> Incompatible return value type (got "pygame.surface.Surface", expected "pygame.Surface")
> ```
> 
> I think the issue is because `pygame.Surface` seems to be defined as a class that extends `pygame.surface.Surface` instead of just being defined as the same class. So things like `pygame.display.set_mode()` will always return instances of `pygame.surface.Surface` and not `pygame.Surface`.
> 
> Can anyone help me better understand what I might be doing wrong? Are these meant to be different classes or is this a bug?

@Zeelot I am having the same issue. (maybe create an issue?)

_Originally posted by @Gaming32 in https://github.com/pygame/pygame/issues/839#issuecomment-922524901_

MightyJosip:
This will fix #2809 
Con of this is that PyCharm thinks `pygame.Rect` is variable (instead of class)
image


ankith26 (approved):
LGTM! 👍 🎉

Zeelot:
hey @MightyJosip I think if you want pycharm to behave, we can use `__all__` here instead of creating new variables. Should we try switching to that?

ankith26:
I don't see how that would work, but I have another idea here, what if we do something like
```py
from pygame.rect import Rect
from pygame.color import Color
from pygame.pixelarray import PixelArray
from pygame.math import Vector2, Vector3
from pygame.cursors import Cursor
```
would this work in both cases well?

MightyJosip:
> 
> 
> I don't see how that would work, but I have another idea here, what if we do something like
> 
> ```python
> from pygame.rect import Rect
> from pygame.surface import Surface
> from pygame.color import Color
> from pygame.pixelarray import PixelArray
> from pygame.math import Vector2, Vector3
> from pygame.cursors import Cursor
> ```
> 
> would this work in both cases well?

@ankith26 Sadly not, if you do 
`from pygame.surface import Surface`
than you get 
`error: Name "pg.Surface" is not defined`
with mypy/PyCharm also doesn't give me pygame.Surface autocomplete. The working expression would look like
`from pygame.surface import Surface`
but that produces the same result as this PR

ankith26:
Wait, what's the difference between the working and the erroring line, again?

MightyJosip:
> 
> 
> Wait, what's the difference between the working and the erroring line, again?

Oh sorry I made a typo. The eorking line is `from pygame.surface import Surface as Surface`

Zeelot:
I don't know how `pyi` files work but can we not use `__all__` in them? usually it looks a little like this:

```python
from pygame.rect import Rect
from pygame.color import Color
from pygame.pixelarray import PixelArray
from pygame.math import Vector2, Vector3
from pygame.cursors import Cursor

__all__ = [
    "Rect",
    "Surface",
    "Color",
    "PixelArray",
    "Vector2",
    "Vector3",
    "Cursor",
]
```

MyreMylar (changes requested):
I'd like to try the method suggested by @Zeelot using imports and `__all__` in the `__init__.pyi` file.

In my quick test it at least seemed to produce better type hinting information for `pygame.Surface` in PyCharm than the current state of affairs. Not sure how mypy will cope with it though.

MightyJosip:
Yea that is even better. Thanks @Zeelot

buildconfig/pygame-stubs/__init__.pyi
```diff
@@ -35,16 +35,24 @@ from pygame import fastevent as fastevent
 # Other
 from pygame import scrap as scrap
 
-from ._common import _AnyPath
+_AnyPath = Union[str, bytes, PathLike[str], PathLike[bytes]]
```
ankith26, line 38: Perhaps `from ._common import AnyPath` would be better here?

buildconfig/pygame-stubs/rect.pyi
```diff
@@ -84,7 +84,7 @@ class Rect(object):
         ],
     ) -> None: ...
     @overload
-    def clamp(self, rect: Union[_CanBeRect, Rect]) -> Rect: ...
+    def clamp(self, rect: Union[CanBeRect, Rect]) -> Rect: ...
```
ankith26, line 87: `Union[CanBeRect, Rect]` can be replaced with `RectValue` here and many other places in this file. Sorry I forgot to fix it in my last PR but it'd be great if you could fix it in this one. Also there are more places where `CanBeRect` is used but `RectValue` can be used in place of that

ankith26 (changes requested):

buildconfig/pygame-stubs/__init__.pyi
MightyJosip: If I do `from ._common import` in \_\_init\_\_ file than PyCharm gives me typehint for pygame._common.
image
But importing from _common in other files works well

ankith26 (approved):
LGTM! Neat typehint fixes 🎉

illume:
I think the all should include all the things? Currently some are missing.

ankith26:
Yeah I initially suspected that would break stuff too. But seems like `mypy` nor `pylance` don't complain about `__all__` missing stuff, but yeah maybe it can be fixed for better clarity

buildconfig/pygame-stubs/__init__.pyi
```diff
@@ -35,16 +35,24 @@ from pygame import fastevent as fastevent
...
+
+__all__ = [
+    "Rect",
+    "Surface",
+    "Color",
+    "PixelArray",
+    "Vector2",
+    "Vector3",
+    "Cursor",
+]
```
illume, line 55: Can you please add all the things in here?

illume (changes requested):
Left a note about all the things in dunder all... since maybe it could break some other tool?

MightyJosip:
I changed too much stuff here so I am gonna split it up in few PRs

Issue #2814, opened 2021-11-03
Deprecation warnings from surfarray.py after numpy version 1.20.0


MyreMylar:
Specifically this section:

https://github.com/pygame/pygame/blob/14fbf614b9c1d040cd15b267a527ae50c9b90528/src_py/surfarray.py#L50-L54

It appears that `numpy.float` has been deprecated in Numpy 1.20.0 (along with some other numpy types that were just aliases to python builtin types) so we should now just be using the standard python `float` type. See:

https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

I'm not 100% that we should just swap out `numpy.float` for `float` in this list but from a quick skim that seems to be the recommendation.

This just popped up for me after upgrading a bunch of python libraries and running through my test suite.

Full Deprecation warning looks like this:

```
surfarray.py:54: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this wi
ll not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
  Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
    numpy_floats.append(getattr(numpy, type_name))

```

PR #2815, opened 2021-11-03
Simple fix to swap numpy.float type out in favour of float

MyreMylar:
This is because numpy.float is now deprecated.

See #2814


Starbuck5:
It could also be good to have a comment linking to the issue, for curious people in the future.

ankith26 (approved):
Looks good! 👍 🎉

Starbuck5 (approved):
Looks good. Thanks!

PR #2816, opened 2021-11-05
Release GIL on SDL_Quit, fix deadlock

### LINKED ISSUE #2813: pygame process hangs on SDL_Quit
**Environment:**

- **Operating system**: Windows, Mac (possibly any)
- **Python version**: 3.7.4
- **SDL version**: 2.0.16
- **PyGame version**: 2.0.2+

**Current behavior:**
#2428 introduces an issue where `pygame` process will hang after trying to quit, either via closing the window, calling `sys.exit()` or calling `pygame.quit()` (note that the submodules like display or mixer seem to close gracefully). Problem did not occur before the change was introduced (I checked it with the same python/SDL mix). After some investigation, it turns out to be stuck on the following line: https://github.com/pygame/pygame/blob/14fbf614b9c1d040cd15b267a527ae50c9b90528/src_c/base.c#L389
Removing `SDL_Quit` line makes the process shut as expected.

**Steps to reproduce:**
This happens in an application of significant complexity, I did not run into this issue on simple ones. I cannot share the source code but can assist in any other meaningful way.

ankith26:
A PR to fix #2813

The thread that is responsible for posting event timers runs in the background, and that thread needs to acquire GIL to post events. When `pygame.quit()` calls `SDL_Quit()` the GIL is held here, and the SDL function probably waits internally for the timer thread to complete and this creates a deadlock that hangs the app during `pygame.quit()`

The only fix for this is to release GIL while calling `SDL_Quit()`, hopefully this does not have any other unintended side effects, so this PR needs more testing before merging.

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -474,7 +474,11 @@ _pg_quit(void)
 
     pg_is_init = 0;
 
+    /* Release the GIL here, because the timer thread cleanups should happen
+     * without deadlocking. */
+    Py_BEGIN_ALLOW_THREADS; 
     pg_atexit_quit();
+    Py_END_ALLOW_THREADS;
 }
 
 static PyObject *
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -536,8 +536,8 @@ quit(PyObject *self)
 
         Py_BEGIN_ALLOW_THREADS;
         Mix_CloseAudio();
-        Py_END_ALLOW_THREADS;
         SDL_QuitSubSystem(SDL_INIT_AUDIO);
+        Py_END_ALLOW_THREADS;
     }
     Py_RETURN_NONE;
 }


Starbuck5:
Do you have a program that creates the issue, so I can test before/after locally?

ankith26:
Any program that sets a timer with `pygame.time.set_timer()` with a short delay (I tested with 32 milliseconds) should reproduce this issue. Here is a basic reproducer (yes, no "complex" code stuff ;) )
```py
import pygame

pygame.init()
win = pygame.display.set_mode((400, 300))
pygame.display.set_caption('Hello World!')

# This line triggers the bug with pygame latest stable release
pygame.time.set_timer(pygame.event.custom_type(), 32)

while True: # main game loop
        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit()
    pygame.display.update()
```

Starbuck5 (approved):
This is total black magic, but it works.

Did it release the GIL before your refactor and it was just missed while transitioning?

illume (approved):
👍 Magical indeed.

PR #2817, opened 2021-11-05
Officially drop python2 (and 3.5) and SDL1 support

Starbuck5:
This is a lot lighter than my other one, so it should hopefully be easy to merge.

I dropped SDL1, python2, and python3.5 CI runs and claims of support in the README.

It still needs an sdist workflow, which will happen in a separate PR.


illume (approved):
👍 thanks

PR #2818, opened 2021-11-05
neon optimisations default on arm64

ankith26:
ARM neon optimisations are default on arm64 (things like M1 mac and pi 3/4 running 64-bit OS, and not older pis and/or ones that run 32-bit OSes), so enable them in our code as well

This PR also makes neon optimised SDL builds on M1 mac

### DIFF
--- a/buildconfig/manylinux-build/docker_base/sdl_libs/build-sdl2-libs.sh
+++ b/buildconfig/manylinux-build/docker_base/sdl_libs/build-sdl2-libs.sh
@@ -27,8 +27,13 @@ tar xzf ${SDL2}.tar.gz
 # this is for renaming the tip.tar.gz
 # mv SDL-* ${SDL2}
 
+if [[ "$MAC_ARCH" == "arm64" ]]; then
+    # Build SDL with ARM optimisations on M1 macs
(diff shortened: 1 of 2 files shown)


Starbuck5:
This is cool stuff. I wonder if we could get someone with that hardware to test and see what kind of performance difference it gives.

ankith26:
> This is cool stuff. I wonder if we could get someone with that hardware to test and see what kind of performance difference it gives.

All the good stuff was done by MyreMylar in this PR #1711 , where the speedups that are made by `--enable-arm-neon` are discussed. So I'd expect similar speedups that happen by default without the flag, in this PR. 

ARM neon is a default thing on arm64, and does not even need any flag like `-mfpu=neon` (which even errors on arm64). 

As said earlier, this PR has no affect on arm32 (the flag can still be passed in optionally as before), and therefore should be safe to merge in

illume (approved):
👍 Thanks :)

Starbuck5 (approved):
Looks good to me.

Also, I had heard about MyreMylar's work on this, but until I read the PR thread I didn't realize how much of a performance gain it actually is.

PR #2823, opened 2021-11-06
Remove most SDL1/SDL2 compatibility c code

Starbuck5:
Same deal as #2819 

Mostly automatic, but with manual edits and verification.

This still includes the source for SDL1 only modules, like cdrom and overlay.

IS_SDLv1 and IS_SDLv2 still exist and are still defined, but there are much less of them.

Numbers:
IS_SDLv1: 320 -> 13
IS_SDLv2: 556 -> 13


illume (approved):
So many thankses (a gang of thanks)!

So good. So many lines. So many files.

Bye bye SDL1 😢

Issue #519, opened 2018-09-04
docs not being included


illume:
From https://twitter.com/pygame_org/status/1036699852334411781
```
Ubuntu based distro.

In fact, the error I get when I run "python -m http://pygame.docs " on my terminal is precisely "(~home_folder)/.local/lib/python3.5/site-packages/pygame/docs/index.html: No such file or directory".
```

The tar.gz and .whl files have the docs. https://pypi.org/project/Pygame/#files

Maybe it has to do with how it was installed? pip(with what options?) or apt-get?

algorithmpub:
Thank you for creating the issue. I don't have the exact details of my pip install, except for the fact I used pip3, since I use pygame with my python 3 version.

Details:
OS: Lubuntu 16.04 LTS 32-bits
Installation: via pip3 (additionally, I don't remember having updating the library via pip3 but I guess there's a slight chance that I did so)
Python: 3.5.2
Pygame: 1.9.3

Problem description: As mentioned above, when I run the "python -m pygame.docs" command on my terminal I get precisely the following error message: "(~home_folder)/.local/lib/python3.5/site-packages/pygame/docs/index.html: No such file or directory". After taking a good look in the folder mentioned on the error message, I noticed not only the index.html file is missing, but the folder doesn't seem to be completed. The exact contents were:
__init__.py
logos.html
__main__.py
__pycache__
pygame_logo.gif
pygame_powered.gif
pygame_small.gif
pygame_tiny.gif
ref/docscomments.json

And nothing more. Perhaps this was a problem on installation? I can't really tell since the "python -m pygame.docs" command was something I only tried recently, after more than a year of using the library in this same computer. And despite all that, all my pygame applications work fine, never had a single issue besides this one. Even, the "python -m pygame.tests" command runs all the tests nicely.

algorithmpub:
obs.: somehow github displayed the init and main files and the pycache folder without the leading and trailing double underscores, but they have them.

MyreMylar:
This is exactly what happens if the pygame code is built from source (rather than a pypi package). The docs are built separately from the setup.py file with:

`python setup.py docs`

When you are in the pygame source directory. So are not be included if you just do:

`python setup.py build`
`pip install . -U`

Or something similar. 

The solution for the average user who ends up building pygame from source (which can be triggered in lots of edge installation cases):

a) Simplest:  read the docs online at pygame.org/docs
b) Slightly more complex: Build the docs from source after grabbing the source from GitHub as well with the command above.

[...]

PR #2826, opened 2021-11-08
Add docs to the CI-generated wheels

### LINKED ISSUE #2690: Distribute local copy of pygame docs with installs
Currently, you can use the command `python -m pygame.docs` to launch the pygame.org/docs website.

It would be cool if pygame distributions came with their own documentation, and the command launched that instead.

This would help users with poor internet, or users programming without internet, mitigate the impact of website outages, and provide a canonical documentation version for every release.

We can't do this now (easily), because our docs build process only runs on Python 3.6 and above, but once py2.7 and py3.5 support is dropped, this will be possible

Starbuck5:
Closes #2690 

blubberquark said that he thought this used to work, and he's right! https://renesd.blogspot.com/2018/02/pygame-documentation-updates.html

Oh wow I hadn't seen #519. And the fact that local docs are documented in https://www.pygame.org/wiki/FrequentlyAskedQuestions.


ankith26 (approved):
This PR is looking good! Nice additions, local docs are sure really useful! 👍 

I went through CI generated wheels and sampled wheels from each configuration, and saw that indeed all wheels have generated html pages, except the sdist wheel which does not have those for some reason, so IG this would need fixing

Also it would be a good idea to have some sort of unit test for testing that the docs exist in the wheels, because we don't want a future regression removing docs in wheels going unnoticed again 😉

illume:
Very good!

Eventually it'd be good to generate these something like how we will do with Cython. Then we could also pin the version of the sphinx dependency to the one that works for us.

ankith26 (approved):
👍 🥳

Starbuck5:
Damn I messed up my squash and push when I amended the commit to add coauthor info.

Ankith, feel free to mess with it (the commit history) if you want.

MyreMylar (approved):
LGTM 👍

Issue #2830, opened 2021-11-09
Spritedict is wrongly typed


vnmabus:
The `spritedict` attribute of `AbstractGroup` is typed as `Dict[Sprite, int]` when it should be typed as `Dict[Sprite, Rect]`.

PR #2831, opened 2021-11-10
changed spritedict type from [Sprite, int] to [Sprite, Rect]

novialriptide:
This is a really simple pull request, but yeah.

In response to this: https://github.com/pygame/pygame/issues/2830


ankith26:
Thanks for the PR! 👍 🥳 
I am not so familiar with `pygame.sprite` well enough to be giving this PR a formal review, but from the code I see an integer still being used here
https://github.com/pygame/pygame/blob/a054dcf8686e009bcf0a4a9d81dc158a8d98ca5c/src_py/sprite.py#L383

Maybe someone with more experience with sprite can provide some clarity here?

MyreMylar:
I think this PR is correct. 

The intended value of spritedicts is a rect. The dictionary's rect values are updated every time the group is drawn it just starts out as a 0 for the AbstractGroup, I assume to save a teeny bit of time creating a Rect that will be shortly replaced anyway. 

Perhaps we could replace this line in AbstractGroup with the similar line from `LayeredUpdates` which creates a zero'd out Rect instead. It would be a little bit slower to add sprites to groups but might stop the linters from complaining.

```
self.spritedict[sprite] = self._init_rect
```

Where _init_rect is defined as:

```
_init_rect = Rect(0, 0, 0, 0)
```

Alternately we could switch the line to `self.spritedict[sprite] = None` And set the type as `[Sprite, Optional[Rect]]` ?

MyreMylar (changes requested):
Gonna set this as Request Changes. I don't feel that strongly about either of the two options I put above, but we should probably pick one and be done with it.

vnmabus:
Please, if `None` cannot be obtained in practice, don't type it as `Optional`. Otherwise we would need to check that it is not `None` everywhere to keep Mypy happy.

Starbuck5:
Yeah, it seems like we should create a zeroed-out Rect to maintain consistency with the newly correct typing.

On another note we haven't been putting any (AFAIK) issue # comments in the type hints, so we probably shouldn't do that here.

novialriptide:
It seems the forked repository is gone. I'll close then...

PR #2833, opened 2021-11-11
remove old SDL1 and PY2 tests

bydariogamer:
- All tests that were "python3 only" or "SDL2 only" are now run always.
- All tests that were "python2 only" or "SDL1 only" have being deleted.
- Tests from deprecated modules (`pygame.cdrom` and `pygame.compat`) have being removed.
- Removed unneeded tags files.
- Some minor style fixes in tests files
- The `test_get_focused` in `key_test.py` is now skipped on GNU/Linux.


test/base_test.py
```diff
@@ -585,14 +585,8 @@ def test_set_error(self):
...
-        if sys.version_info.major > 2:
-            pygame.set_error(u"你好")
-            self.assertEqual(u"你好", pygame.get_error())
-        else:
-            # no unicode objects for now
-            pygame.set_error(u"你好")
-            encstr = u"你好".encode("utf8")
-            self.assertEqual(encstr, pygame.get_error())
+        pygame.set_error(u"你好")
+        self.assertEqual(u"你好", pygame.get_error())
```
Starbuck5, line 589: u prefix is unnecessary in python3

test/key_test.py
```diff
@@ -25,25 +24,29 @@ def setUp(cls):
...
 
-    @unittest.skipIf(SDL1, "SDL1 always thinks it has keyboard focus.")
+    # fixme: test_get_focused always fails in linux
+    @unittest.skipIf(
+        sys.platform.startswith("linux"),
+        "This test fails in all tested GNU/Linux"
+    )
     def test_get_focused(self):
+        # Note: this test fails in SDL2 and it was skipped in SDL1.
+        # There might be something wrong in this function
```
Starbuck5, line 37: I don't think it's always broken on Linux, and it works fine on two other platforms. I think this is a problem with your system. 

I'm not sure this test should be skipped.

Starbuck5:
I read the whole thing, this yields some good simplifications, good job!

A couple things:
- there were some tests you commented out fully, they could just be unconditionally skipped instead.
- 1 or 2 detailed Surface tests were deleted because they break in SDL2. I'm not sure those should be deleted quite yet.
- Did you run this through a code formatter? I've found it's best to do manual code changes in one commit, then do the auto formatting in another commit, so reviewers can see exactly what **you** changed

test/base_test.py
bydariogamer: I fixed these two and other one in 6ca0a1011f70fdc83dc10ba6dabced4d7cbde4b3 and a3192431f88a04b78b31bd12eef6ed0e57bff15d

test/key_test.py
bydariogamer: I removed the "skip". However, I left the `# fixme` because the reason why it fails it is still unknown.

bydariogamer:
>  There were some tests you commented out fully, they could just be unconditionally skipped instead.

Now they are uncommented and skipped.

>  Did you run this through a code formatter? I've found it's best to do manual code changes in one commit, then do the auto formatting in another commit, so reviewers can see exactly what **you** changed

I didn't run a code formater. Although, I was thinking of using `black` after I finish all changes. Some files doesn't follow any code style.

illume (approved):
👍 Epic work. Thanks! 🎉

PR #2834, opened 2021-11-11
Remove ability to reference joysticks by index (breaking change)

lordmauve:
This removes the compatibility hacks that enabled SDL1-style joystick device indexes to work under SDL2.

This is a deliberate breaking change which will force users to use the (correct) SDL2 API i.e. switch to instance IDs and abandon device indexes.

SDL2 does not use indexes at any point after initialising a device. This is because hotplugging events remap the device indexes. Instead SDL2 provides unique instance IDs. The hacks we added allowed us to synthesise a `.joy` attribute that behaves like a device index, based on the instance ID.

But: these hacks do break if joystick devices are added or removed. You can have multiple connected devices that have the same device index, making it impossible to distinguish events. This bug affects any game using PyGame/SDL2 but with the SDL1-compatible API.

Removing the hacks forces users to adopt correct utilisation of the API and therefore eliminates these bugs.

### DIFF
--- a/docs/reST/ref/event.rst
+++ b/docs/reST/ref/event.rst
@@ -55,7 +55,7 @@ their own new events with the :func:`pygame.event.Event()` function.
 The event type identifier is in between the values of ``NOEVENT`` and
 ``NUMEVENTS``. User defined events should have a value in the inclusive range
 of ``USEREVENT`` to ``NUMEVENTS - 1``. User defined events can get a custom
-event number with :func:`pygame.event.custom_type()`. 
+event number with :func:`pygame.event.custom_type()`.
 It is recommended all user events follow this system.
 
 Events support equality and inequality comparisons. Two events are equal if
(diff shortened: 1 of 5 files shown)


lordmauve:
I'm wondering if we should skew the instance_ids by a constant (say 1000), so that device indexes and instance IDs are pretty much guaranteed never to match up.

Starbuck5:
I'm very unfamiliar with the joystick module, but do the `joy` attributes emit deprecation warnings?

I see they're deprecated in the docs, but people could easily pay that no mind.

We could make the joy attribute into a property and have it emit a deprecation warning on the getter.

MyreMylar (changes requested):
Referencing @Starbuck5 above, currently using `get_id()` in pygame does not emit any deprecation warnings.

How about an intermediate stage, before we remove `get_id()` entirely, where we make `get_id()` a wrapper that calls `get_instance_id()` and emits a deprecation warning that says this function is now just a temporary wrapper for `get_instance_id()` which is the new way to uniquely identify controller-like devices. 

That way we are basically removing it now - but we are also less likely to break anyone's code immediately as I imagine in many projects `get_instance_id()` will do the exact same thing as `get_id()`, and anyone who does get any weirdness will see the deprecation warning and have a good idea where to start.

lordmauve:
There's barely any point in emitting deprecation warnings for get_id() because it just returns the value that users pass into the constructor. Almost no need to call it.

Much existing code does this:

```
sticks = [Joystick(n) for n in get_count()]

for ev in pygame.event.get():
    if ev.type == JOYBUTTONDOWN:
        stick = sticks[ev.joy]
```

Issuing a DeprecationWarning for `Event.joy` is probably the high value thing to do, as well as adding the fixed offset to the instance IDs. And updating the docs.

illume:
Cleaning up.

PR #2837, opened 2021-11-12
Fix for python3+

penggrin12:
Old version was not working for python3+
It was because of print function

Old has that feature "print 'text'"
New one does not.
Now code look like this "print('text')"

### DIFF
--- a/docs/reST/tut/MoveIt.rst
+++ b/docs/reST/tut/MoveIt.rst
@@ -11,6 +11,7 @@ Help! How Do I Move An Image?
 
 .. rst-class:: docinfo
 
+:Edited: Akim Sidorov
 :Author: Pete Shinners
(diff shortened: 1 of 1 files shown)


Starbuck5:
Hey, thanks for the spot.

I don't think adding your name as an editor is necessary though, other people have edited in the past without doing that.

MyreMylar (changes requested):
There is a missed python 2 style print on line 84.

I would also change the credit to 'Updated code to Python 3: Akim Sidorov' if we are going to have it so it's clear what was changed.

Starbuck5:
@novialriptide Can you take this over?

I.e. go through the tutorials in the docs and make sure things are using Python 3 syntax (like the print function)

Starbuck5:
Fixed in #2984

PR #2838, opened 2021-11-12
add missing typehints to Vector2 and Vector3

bydariogamer:
Added `__neg__` typehint to `Vector2` and removed `Optional` from the typing imports


ankith26 (approved):
LGTM! Nice PR 👍 🎈

buildconfig/pygame-stubs/math.pyi
```diff
@@ -124,6 +124,15 @@ class Vector2:
     def __rmul__(self, other: float) -> Vector2: ...
     def __truediv__(self, other: float) -> Vector2: ...
     def __floordiv__(self, other: float) -> Vector2: ...
+    def __neg__(self) -> Vector2: ...
+    def __pos__(self) -> Vector2: ...
+    def __bool__(self) -> bool: ...
```
Starbuck5, line 129: is it necessary to type this? Python's `object` has `__bool__` already, which is presumably typed

MyreMylar: I don't think it will do any harm in the type stub,

MyreMylar (approved):
LGTM 👍

MyreMylar:
Not sure how to kick appveyor from my end to get this merged.

Starbuck5:
I have the appveyor password, so usually I can kick things into motion to re-build around weird failures.

But this time it errors when I try to do that, probably because the branch was deleted on bydariogamer's fork.

illume (approved):
👍  thanks

illume:
If the branch is deleted it seems we can’t merge it. So it’ll need to be done again (or the branch restored). It says there’s a conflict, and that’s why it can’t be merged (just a bad error message I guess)

Starbuck5:
We can merge deleted branches in general (one of novial's the other day was a deleted branch), but just not if it needs more work.

bydariogamer:
I added the same changes in #2908 and now it shouldn't conflict anything (and the branch exists, so it should be no problem if any change is needed). I will close this PR, the other one has the exact same additions so reviewing it shouldn't take long.

PR #2840, opened 2021-11-13
fix filenotfound errors on pathlib loading

ankith26:
Fixes bug where `FileNotFoundError`s are not displayed right when file loading fails from pathlib objects. Also does a bit of code cleanup in the function handling the errors

### DIFF
--- a/src_c/rwobject.c
+++ b/src_c/rwobject.c
@@ -48,9 +48,6 @@ static const char pg_default_errors[] = "backslashreplace";
 
 static PyObject* os_module = NULL;
 
-#define PATHLIB "pathlib"
-#define PUREPATH "PurePath"
-
 static Sint64
 _pg_rw_size(SDL_RWops *);
 static Sint64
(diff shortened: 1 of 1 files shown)


src_c/rwobject.c
```diff
@@ -700,90 +662,83 @@ _pg_rw_read(SDL_RWops *context, void *ptr, size_t size, size_t maxnum)
...
+                        " '%S'.", obj, cwd);                       
+
+    Py_DECREF(cwd);
+    Py_DECREF(isabs);
+    return NULL;
+
+error:
+    Py_XDECREF(cwd);
+    Py_XDECREF(isabs);
+    PyErr_Format(PyExc_FileNotFoundError, "No such file or directory: '%S'.", obj);
```
Starbuck5, line 741: It's a bit strange to have a state labeled error that is fine to go to in normal execution.

Maybe it should be named like `simple_case` or something?

Starbuck5:
This is a lot of code change to fix the issue, but I see it also simplifies some code and removes from python <3.6 compatibility, which is cool.

Just in case, for the future, I would also like a test in the pathlib loading test that asserts it raises a FileNotFound error when the file is nonexistent.

src_c/rwobject.c
ankith26: yeah, true

ankith26:
The old code has unnecessary nesting which was painful on readability and that's why it was harder to catch the bug. Once I unnested and simplified the code it was an easy fix

> Also, could you please highlight to me what of this exactly fixes the problem?

The issue was that the code was doing `PyUnicode_Check(obj)` before, so when `obj` is pathlib object or bytes object, instead of erroring cleanly, it goes into the file object handling side, which then later fails with `TypeError: Not a file object`

The check is not required at all because we are sure that `obj` should be `str | bytes | os.PathLike[str] | os.PathLike[bytes]` at that stage

MyreMylar (approved):
Tested locally and can confirm it fixes the error with Path() objects error messages.

👍

illume (approved):
👍 thanks

PR #2843, opened 2021-11-14
Deprecate fastevent, compat shim based on event

ankith26:
Deprecates `pygame.fastevent`, a module that is no longer relevant as pygame uses SDL2.

For now, this PR removes old fastevent code and puts a `pygame.event` based shim in place for more compatibility (this also fixes a few `fastevent` regressions introduced in 2.0.1 that I don't think anyone noticed yet xD)


illume (approved):
👍

PR #2845, opened 2021-11-15
remove pygame.compat and black tests and src_py

### LINKED ISSUE #2820: Remove Python2 and SDL1 compatibility from Python code.
Now that we have officially dropped support for these two things, there is obsolete python code in `src_py` and in `tests` (at least).

I was wondering if @rethanon, @bydariogamer, or @Grimmys wanted to do some code cleaning on this one. I think it would be up your collective wheelhouses. :smile:

Breaking this into 4 distinct chunks for separate pull requests:
- [x] first half of the tests (start through mask?)
- [x] second half of tests (math through video?)
- [ ] src_py code
- [ ] examples

There is a pygame.compat module, anytime you see that being invoked should be replaced with the correct call for python3. (Don't actually delete the pygame.compat module yet, that will have to come at the very end).

The tests commonly have a constant at the top about whether it's SDL2 or not, and then skips certain stuff on SDL2 or on SDL1. Those conditionals and constants can be removed.

bydariogamer:
- Remove `pygame.compat` use from all directories (i.e.: `src_py`, `test` and `examples`)
- `black test` and `black src_py` (as @illume had already done in 0e67476d9b39d31ccf75add1afc1a02365f17843)
- Closes #2820


examples/fonty.py
```diff
@@ -74,8 +73,8 @@ def main():
 
     # Get some metrics.
     print("Font metrics for 'Fonty':  %s" % a_sys_font.metrics(text))
-    ch = unicode_("%c") % 0x3060
-    msg = unicode_("Font metrics for '%s':  %s") % (ch, a_sys_font.metrics(ch))
+    ch = str("%c") % 0x3060
+    msg = str("Font metrics for '%s':  %s") % (ch, a_sys_font.metrics(ch))
```
Starbuck5, line 77: this can be further simplified. You can put a character explicitly in the string with \u, and use an f string for the message.

examples/midi.py
```diff
@@ -42,6 +42,9 @@ def _print_device_info():
 
 def input_main(device_id=None):
     pg.init()
+    pg.fastevent.init()
+    event_get = pg.fastevent.get
+    event_post = pg.fastevent.post
```
Starbuck5, line 48: This shouldn't be re added?

src_py/colordict.py
```diff
@@ -686,4 +684,4 @@
 }
 
 for k, v in THECOLORS.items():
-    THECOLORS[unicode_(k)] = v
+    THECOLORS[str(k)] = v
```
Starbuck5, line 687: Isn't this a non operation now?

src_py/ftfont.py
```diff
@@ -32,19 +31,19 @@ class Font(_Font):
     __get_default_resolution = staticmethod(get_default_resolution)
     __default_font = encode_file_path(get_default_font())
 
-    __unull = as_unicode(r"\x00")
-    __bnull = as_bytes("\x00")
+    __unull = "\x00"
+    __bnull = b"\x00"
```
Starbuck5, line 35: This probably works, but I just want to confirm. __unull was a raw string and now it isn't.

src_py/sysfont.py
```diff
@@ -436,7 +427,7 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
     gotbold = gotitalic = False
     fontname = None
     if name:
-        if isinstance(name, (str, bytes, unicode_)):
+        if isinstance(name, (str, bytes)):
             name = name.split(b"," if str != bytes and isinstance(name, bytes) else ",")
```
Starbuck5, line 431: This looks like a py2 compat thing, since then str would equal bytes. Same on line 503 below.

test/color_test.py
```diff
@@ -290,7 +289,7 @@ def test_length(self):
         self.assertRaises(ValueError, c.set_length, 5)
         self.assertRaises(ValueError, c.set_length, -1)
         self.assertRaises(ValueError, c.set_length, 0)
-        self.assertRaises(ValueError, c.set_length, pow(2, long_(33)))
+        self.assertRaises(ValueError, c.set_length, pow(2, int(33)))
```
Starbuck5, line 292: does this need to be casted to an int?

test/font_test.py
```diff
@@ -542,6 +536,7 @@ def test_load_from_file_bytes(self):
             os.path.split(pygame.__file__)[0], pygame_font.get_default_font()
         )
         filesystem_encoding = sys.getfilesystemencoding()
+        filesystem_errors = "replace" if sys.platform == "win32" else "surrogateescape"
```
Starbuck5, line 539: where did this come from?

test/freetype_test.py
```diff
@@ -1018,7 +1013,7 @@ def test_freetype_Font_render_raw(self):
 
         # bug with non-printable characters: this would cause a crash
         # because the text length was not adjusted for skipped characters.
-        text = unicode_("").join([unichr_(i) for i in range(31, 64)])
+        text = str("").join([chr(i) for i in range(31, 64)])
```
Starbuck5, line 1016: no need for explicit string here

test/image_test.py
```diff
@@ -25,7 +24,7 @@ def test_magic(f, magic_hexes):
     if len(data) != len(magic_hexes):
         return 0
     for i, magic_hex in enumerate(magic_hexes):
-        if magic_hex != ord_(data[i]):
+        if magic_hex != data[i]:
```
Starbuck5, line 27: it's fine without any `ord()`?

test/image_test.py
```diff
@@ -369,8 +368,8 @@ def test_save_colorkey(self):
     def test_load_unicode_path(self):
         import shutil
 
-        orig = unicode_(example_path("data/asprite.bmp"))
-        temp = os.path.join(unicode_(example_path("data")), u"你好.bmp")
+        orig = str(example_path("data/asprite.bmp"))
+        temp = os.path.join(str(example_path("data")), u"你好.bmp")
```
Starbuck5, line 372: example_path probably returns string all the time now?

test/font_test.py
bydariogamer: It was in the `compat.py`. It is a function that, once removed py3 alternatives, does essentially what you see in that line, so I replaced the only case I saw in the project with the code and deleted the function from `compat.py` (and eventually deleted the file after it was empty).

test/image_test.py
bydariogamer: Yes, the `ord_` compatibility function wasn't just like the regular `ord` in python3. You can try to run the test with `ord` if you want, but you will get an error:
```Traceback (most recent call last):
  File "/code/pygame/test/image_test.py", line 254, in test_save
    (1, fmt), (test_magic(handle, magic_hex[fmt.lower()]), fmt)
  File "/code/pygame/test/image_test.py", line 27, in test_magic
    if magic_hex != ord(data[i]):
TypeError: ord() expected string of length 1, but int found```

File: src_py/sysfont.py (hunk 5 above)
Comment (bydariogamer) [reply src_py/sysfont.py]: I removed it from more places in the same file too. 
c37e1b5  should remove it.

File: test/image_test.py (hunk 10 above)
Comment (bydariogamer) [reply test/image_test.py]: Fixed in 768fda1 (and same fix applied in many lines in many files).

**Comment by bydariogamer** [comment]
If this PR is merged, it closes #2820 (all four tasks done).

File: src_py/__init__.py [hunk 11]
```diff
@@ -77,13 +77,12 @@ def warn(self):
 
 # we need to import like this, each at a time. the cleanest way to import
 # our modules is with the import command (not the __import__ function)
-# isort: skip_file
```
Starbuck5, line 80: why is this removed?

src_py/ftfont.py
Starbuck5: It's a pattern all over the codebase, it makes sense that a pygame compatibility method like `as_unicode` would want a raw string, to control encoding.

test/font_test.py
Starbuck5: Yeah, but the diff doesn't show it replacing anything, just popping up out of nowhere.

src_py/__init__.py
bydariogamer: I am not sure. Maybe someone changed that line in the two days before I made more changes and I overrode it.
I will add it back later.

src_py/fastevent.py
```diff
@@ -1,84 +0,0 @@
-"""
```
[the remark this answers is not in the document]
bydariogamer: Readded.
I also added the right number of spaces between function declarations.

ankith26 (approved):
This PR looks great! Super nice cleanups, nice PR dario! 🥳 👍

illume (approved):
🎉 sparks joy! Thanks.

PR #2854, opened 2021-11-20
C cleanups to remove more old compat code

ankith26:
Removes old compat code, to make the codebase easier to work with. This PR might be harder to review, so I have split the cleanups this PR does in multiple commits that describe what is being cleaned


MyreMylar (changes requested):
After this PR this test:

```
    def test_set_gamma(self):
        pygame.display.set_mode((1, 1))
        gammas = [0.0, 0.25, 0.5, 0.88, 1.0]
        for gamma in gammas:
            self.assertEqual(pygame.display.set_gamma(gamma), True)
        gammas = [(0.5, 0.5, 0.5), (1.0, 1.0, 1.0), (0.22, 0.33, 0.44), (0.0, 0.0, 0.0)]
        for gammaTuple in gammas:
            self.assertEqual(
                pygame.display.set_gamma(gammaTuple[0], gammaTuple[1], gammaTuple[2]),
                True,
            )
```

fails locally for me on windows. No idea why.

Everything looks good otherwise, tests pass and I'm excited to see the main branch codebase with all the python 2 stuff fully stripped out. 👍

ankith26:
I rebased this branch against main, fixing conficts and such, and now that test passes! I think it's best if this PR can be merged in fast, otherwise it will keep conflicting with other C code change PRs

MyreMylar:
The gamma test is still failing locally for me, but after a bit of tinkering I think it is just a test that should **never** pass locally so I'm not sure why it sometimes was when running the full suite of tests.

Basically the underlying function, at least on windows, will not set gamma ramps that make the monitor too dark to read (anything below 0.25ish will always fail). Since the test currently has three values that violate that principle I'm gonna refactor it with subtests and allowed values and submit that in a separate PR.

This one is good to go.

MyreMylar (approved):
Approved 👍

ankith26:
Ah, that makes sense! Yeah it fails for me too now, IDK maybe before I was running the wrong python version that was on another branch

src_c/_sdl2/touch.c
```diff
@@ -140,13 +140,13 @@ MODINIT_DEFINE(touch)
...
     import_pygame_base();
     if (PyErr_Occurred()) {
-        MODINIT_ERROR;
+        return NULL;
     }
 
     /* create the module */
     module = PyModule_Create(&_module);
     if (module == NULL) {
-        MODINIT_ERROR;
```
Starbuck5, line 149: does this need a module decref?

ankith26: no, because `module` is already `NULL` here

src_c/cdrom.c
```diff
@@ -88,7 +88,7 @@ static PyObject *
 get_count(PyObject *self)
 {
     CDROM_INIT_CHECK();
-    return PyInt_FromLong(SDL_CDNumDrives());
+    return PyLong_FromLong(SDL_CDNumDrives());
```
Starbuck5, line 91: cdrom and overlay should probably just be removed at this point.

ankith26: yeah, but that's probably for another PR hehe

src_c/scrap.c
```diff
@@ -113,10 +113,6 @@ _scrap_init(PyObject *self, PyObject *args)
     /* In case we've got not video surface, we won't initialize
      * anything.
      */
-#if IS_SDLv1
-    if (!SDL_GetVideoSurface())
-        return RAISE(pgExc_SDLError, "No display mode is set");
-#endif
```
Starbuck5, line 119: The thing with this is that we actually would need this check if we used the old windows scrap code, which we might do in the future, to get image support back. Because the windows scrap code needs a window handle iirc.

Also the comment is still there.

I say leave this code but have it commented out?

src_c/transform.c
```diff
@@ -45,30 +45,11 @@ struct _module_state {
...
-static void
-filter_expand_X_ONLYC(Uint8 *, Uint8 *, int, int, int, int, int);
-static void
-filter_expand_Y_ONLYC(Uint8 *, Uint8 *, int, int, int, int, int);
-
-static struct _module_state _state = {
-    "GENERIC", filter_shrink_X_ONLYC, filter_shrink_Y_ONLYC,
-    filter_expand_X_ONLYC, filter_expand_Y_ONLYC};
-#define GETSTATE(m) PY2_GETSTATE(_state)
-#define smoothscale_init(st)
```
Starbuck5, line 69: shouldn't some of this still be around?

ankith26: No, I refactored this file a bit because it was not making proper use of module state storage mechanism. Now it's all fixed up and tested (feel free to do more testing :smile:)

Starbuck5: I mean, if MMX is not defined, it's supposed to have filter_ methods defined here. And now it doesn't.

Unless I'm missing something else in the bigger picture.

ankith26: > I mean, if MMX is not defined, it's supposed to have filter_ methods defined here. And now it doesn't.

Actually it still does, the onlyc filter implementation is a few lines below, I just removed the forward declaration because it was redundant

Starbuck5: It's 1000 lines below 😄

But ok

Starbuck5 (approved):
This looks good to me.
I read through everything, and ran the unit tests locally.

It's great to have pygame C code more standard, like using PyLong_ and PyUnicode_ stuff directly.

src_c/scrap.c
ankith26: Hmm, aight!

PR #2857, opened 2021-11-21
Pylinting, format and lint check with setup.py lint

illume:
## What is added

- Fixes pylint errors and code formatting issues (thanks to black and clang-format)
- Adds a `setup.py format` command, that formats code with [clang-format](https://clang.llvm.org/docs/ClangFormat.html) (.c) and [black](https://black.readthedocs.io/en/stable/) (.py).
- Adds a `setup.py lint` command, that runs [pylint](https://pylint.org/), and that checks that code is formatted (without formatting the code).
- Adds a separate github actions workflow for lint/format/doc checks
- runs `setup.py lint` on CI, which should stop formatting and lint issues being introduced
- runs `setup.py docs` on CI, and checks if it generates any files that aren't committed (like .h files)
- setup.py docs is a distutils command again (rather than using consume_arg)
- configure doesn't run for lint, docs, and format commands
- The src_c and src_py folders are formatted again

## To Test

People need black, pylint and clang-format in order to run:

```bash
# python3 -m pip install black pylint clang-format
python3 setup.py lint
```

Which checks no formatting would be done in the src_c, and src_py code and then runs pylint to check for issues.

Then to format the code, run this:
```bash
python3 setup.py format
```

Then check if docs runs without a Setup file.

```bash
cp Setup /tmp/ #backup
rm Setup
python3 setup.py docs
```

- There should be a good error if you don't have clang-format, black, or sphinx.
- It should show you the error in the CI job when checking the docs, or linting.
- The CI job should show `setup.py lint` command in the title of it, to make it easier for newbies.

## Notes

- https://pypi.org/project/clang-format/ can be had with pip
- A future PR will add black, pylint, and clang-format to build-system.requires, currently this requires people to have the tools installed locally.
- objc files are not formatted yet
- setup.py, tests, examples and buildconfig are not yet formatted or linted (for a separate PRs)
- Compiler warnings are not checked, but we have an issue for that open already
- Hacking wiki page, and the contribution guide should be updated when this is merged

### DIFF
--- a/.github/workflows/format-lint.yml
+++ b/.github/workflows/format-lint.yml
@@ -40,3 +40,14 @@ jobs:
 
       - name: Check code Formatting and Linting
         run: python3 setup.py lint
+
+      - name: Check docs changes are checked in
+        run: |
+          python3 setup.py docs
+          if [[ `git status --porcelain` ]]; then
+            echo "Generating docs caused changes. Please check them in."
+            echo "You may need to run: python3 setup.py docs --fullgeneration"
+            # Run git status again, so people can see what changed.
+            git status --porcelain
+            exit 1
+          fi
--- a/.github/workflows/format-lint.yml
+++ b/.github/workflows/format-lint.yml
@@ -1,0 +1,42 @@
+name: python3 setup.py lint
+
+# Run CI only when a release is created, on changes to main branch, or any PR
+# to main. Do not run CI on any other branch. Also, skip any non-source changes
+# from running on CI
+on:
+  release:
+    types: [created]
+  push:
+    branches: main
+    paths-ignore:
+      - 'docs/**'
+      - 'examples/**'
+      - '.gitignore'
+      - 'README.rst'
+      - '.github/workflows/build.yml'
+      - '.github/workflows/manylinux.yml'
+      - '.github/workflows/ubuntu-sdist.yml'
+
+  pull_request:
+    branches: main
+    paths-ignore:
+      - 'docs/**'
+      - 'examples/**'
+      - '.gitignore'
+      - 'README.rst'
+      - '.github/workflows/build.yml'
+      - '.github/workflows/manylinux.yml'
+      - '.github/workflows/ubuntu-sdist.yml'
+
+jobs:
+  format-lint-code-check:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2.3.4
+
+      - name: Install deps
+        run: python3 -m pip install pylint black clang-format sphinx
+
+      - name: Check code Formatting and Linting
+        run: python3 setup.py lint
--- a/setup.py
+++ b/setup.py
@@ -254,8 +254,9 @@ if consume_arg('cython'):
         kwargs['progress'] = '[{}/{}] '.format(i + 1, count)
         cythonize_one(**kwargs)
 
+
 if consume_arg('docs'):
-    """ 
+    """
     For building the pygame documentation.
 
     This generates html, and documentation .h header files.
@@ -288,7 +289,8 @@ if consume_arg('docs'):
         sys.exit()
 
 
-AUTO_CONFIG = False
+no_compilation = any(x in ['lint', 'format', 'docs'] for x in sys.argv)
+AUTO_CONFIG = not os.path.isfile('Setup') and not no_compilation
 if consume_arg('-auto'):
     AUTO_CONFIG = True
 
@@ -385,7 +387,7 @@ if len(sys.argv) == 1 and sys.stdout.isatty():
 
 
 # make sure there is a Setup file
-if AUTO_CONFIG or not os.path.isfile('Setup'):
+if AUTO_CONFIG:
     print ('\n\nWARNING, No "Setup" File Exists, Running "buildconfig/config.py"')
     import buildconfig.config
     try:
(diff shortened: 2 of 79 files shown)


src_py/threads/__init__.py
```diff
@@ -106,15 +102,15 @@ def doit(x):
 
         wq = WorkerQueue(num_workers)
         t1 = time.time()
-        for xx in range(20):
-            print("active count:%s" % threading.activeCount())
+        for _ in range(20):
+            print(f"active count:{threading.activeCount()}")
```
ankith26, line 106: unrelated to this PR but... `threading.activeCount` is a deprecated function and `threading.active_count` is the new name for it

src_py/sprite.py
```diff
@@ -1560,6 +1558,7 @@ def __repr__(self):
         """
         Turn the class into a string.
         """
+        # pylint: disable=consider-using-f-string
         return "<{klass} @{id:x} {attrs}>".format(
```
ankith26, line 1562: Minor detail but, can we change this to use f-strings (and also change some similar code above in this file) for more consistency? To not clutter the f-string we can always use temporary variables outside the string :)

ankith26 (approved):
Overall I really like the changes this PR does! 🥳 the use of f-strings over old style formatting makes the codebase more pleasant to work with, also lint on CI is a really good idea to enforce code quality better :) left a few minor reviews

src_py/sprite.py
illume: I tried that, but it turned out looking like an evil monster.

src_py/threads/__init__.py
MyreMylar: I think we could probably do a separate PR pass on cleaning up places where pygame itself uses functions it has deprecated (there are a bunch in the tests if you run the test files one by one)

MyreMylar (approved):
LGTM 👍 

some nice linting cleanups.

MyreMylar:
I'm gonna let somebody who has dealt with these YAML files before resolve that conflict. 😄

ankith26:
seems like there might be some more work to be done on this PR, better not merge it in just yet

illume:
Currently the configure code is run when running either the format or lint commands. It also shouldn't run configure when we run the `setup.py docs` command.

setup.py
```diff
@@ -778,25 +782,50 @@ def run(self):
...
             if shutil.which(linter) is None:
-                msg = "Please install '%s' in your environment. (hint: 'pip install %s')"
+                msg = "Please install '%s' in your environment. (hint: 'python3 -m pip install %s')"
                 warnings.warn(msg % (linter, linter))
                 sys.exit(1)
 
-        c_files_unfiltered = glob.glob("src_c/**/*.[ch]")
-        c_file_disallow = ["_sdl2", "pypm", "SDL_gfx", "sse2neon.h", "src_c/doc/"]
+        c_files_unfiltered = glob.glob("src_c/*.[ch]")
+        c_files_unfiltered += glob.glob("src_c/**/*.[ch]")
```
ankith26, line 790: There two lines should become
`c_files_unfiltered = glob.glob("src_c/**/*.[ch]", recursive=True)`
That way, not only it handles the base of the directory in the same function call, it also actually recurses into nested sub-directories according to the [docs](https://docs.python.org/3/library/glob.html#glob.glob) of the function

illume: This is done, thanks.

.github/workflows/format-lint.yml
```diff
@@ -0,0 +1,57 @@
...
+    steps:
+      - uses: actions/checkout@v2.3.4
+
+      - name: Install deps
+        run: |
+          sudo apt-get update --fix-missing
+          sudo apt-get upgrade
+          sudo apt-get install libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev \
+            libfreetype6-dev libportmidi-dev libjpeg8-dev python3-setuptools \
+            python3-dev python3-numpy libsdl2-ttf-dev
```
ankith26, line 44: we don't need any of these apt lines here because we don't build anything here :)

.github/workflows/format-lint.yml
```diff
@@ -0,0 +1,58 @@
...
+          python3 -m pip install pylint black clang-format sphinx
+
+      - name: Check code Formatting and Linting
+        run: python3 setup.py lint
+
+      - name: Check docs changes are checked in
+        run: |
+          python3 setup.py docs
+          git status --porcelain
+          if [[ `git status --porcelain` ]]; then
```
ankith26, line 54: why is the command being called twice here?

illume: Once to print out the result. So people can see what files changed in git.

.github/workflows/format-lint.yml
illume: Indeed. I will remove them.

illume: Runs lots faster without them!

ankith26 (approved):
Super nice PR! 🥳 👍 🎉 
This was much needed, and improves things a ton!

setup.py
```diff
@@ -811,25 +785,85 @@ def run(self):
...
                 sys.exit(1)
 
-        c_files_unfiltered = glob.glob("src_c/**/*.[ch]")
-        c_file_disallow = ["_sdl2", "pypm", "SDL_gfx", "sse2neon.h", "src_c/doc/"]
+        c_files_unfiltered = glob.glob("src_c/**/*.[ch]", recursive=True)
+        c_file_disallow = ["_sdl2", "pypm", "SDL_gfx", "sse2neon.h", "src_c/doc/", "_sprite.c"]
         c_files = [x for x in c_files_unfiltered if not any([d for d in c_file_disallow if d in x])]
+
+        # Other files have too many issues for now. setup.py, buildconfig, etc
         python_directories = ["src_py", "test"]
```
ankith26, line 797: I wonder if the `examples` directory should be lint checked too? Perhaps that could come in another PR too, it should not block this :)

illume: Yeah, I think setup.py, tests, examples and buildconfig can be done in separate PRs.

PR #2858, opened 2021-11-22
Improve MacOS icon

### LINKED ISSUE #2791: Improve MacOS default image
Currently, on MacOS, the default image isn't exactly pretty on the eyes
dock

I did some investigating, and we have explicit code to put in a much higher resolution image, but it is overridden by the default call to `SDL_SetWindowIcon`. 

On Windows, that puts the image in the top left corner of the window. A tiny space, where `pygame_icon.bmp` (32x32) is appropriate.

However, on Mac, the icon goes to the dock at the bottom of the screen, which has more screen real estate. So it should get `pygame_icon.tiff` (128x128).

The pygame code actually tries to give out `pygame_icon.tiff` on Mac! base.c calls VideoAutoInit in macosx.py, which calls _NSBundleApplication to set the ApplicationIconImage with the bytes of `pygame_icon.tiff`. But then this is quickly overridden by  the `SDL_SetWindowIcon` in `display.set_mode()`, I believe.

Tasks
- [x] figure out if this is a regression from 1.9.6. It is. (Maybe the logic in set_mode changed, because SDL_SetWindowIcon still seems to work the same way when called with `display.set_icon`)
regression

- [ ] change default icon (in display.c, I believe) to point to the higher res one on Mac
- [ ] remove macosx.py and _InstallNSApplication (sdlmain_osx.m)?

Starbuck5:
Fixes #2791

In 1.9.6, the icon looks like this:
regression

In 2.0+, the icon looks like this:
Screen Shot 2021-11-21 at 11 29 39 PM

After this patch, the icon looks like this:
Screen Shot 2021-11-21 at 11 22 11 PM

I'm not 100% sure on this, but my hypothesis is that setting the window icon in SDL1 didn't set the icon on the dock, which is where the window icon should go on MacOS. So the pygame developers of the day put custom code into SDLmain_osx.m that would automatically put a logo on if it wasn't running as an app bundle. It used pygame_icon.tiff. Now, I believe that happens but it is instantly overwritten by the default icon that display.c sets.

The regression icon is the icon set by display.c. It is very pixelated, but that actually looks fine on Windows, since the logo field is much smaller than a space on the MacOS dock. So this PR just changes the default icon filename in display.c for MacOS. (And it adds an attribute for the colorkeying because it can't colorkey away black unconditionally now).

I deleted the outdated / unused logo svg and tiff files.

I also deleted the obsolete macosx.py / SDLmain_osx.m code path, although it is some cool code.

Thanks to @TomSchimansky for bringing up to Mac standard and putting the snake on a rounded corners white background, providing the image files for this PR.

On a side note: why do we have .icns and .ico files in src_py anyway? I updated this with the new icns, but it doesn't seem like it would be used anywhere.


MyreMylar (approved):
LGTM 👍  

Even better if it also fixes a crash.

ankith26 (approved):
Yeah, this PR looks good! 👍 🎉 
I'd still like if we could get more reviews from people who can test on a wider range of macs because this change seems like it could break something unexpected?

MyreMylar:
I actually think that perhaps the best way to get this tested on a wide range of macs is to merge it in and release it. 

On the one hand the icon quality is not a show stopping issue, on the other almost everyone who uses pygame on mac will see it immediately so by releasing it we will likely get lots of feedback if it looks worse on some subset of macs. I'd certainly rather pygame 2 had a modern and approachable icon on relatively modern macs, and perhaps a slightly worse icon on really old macs (if that is the case). One of pygame's issues on the wider internet (sadly) is people seeing it as old fashioned and a  pixelated, artefact-y icon likely reinforces that perception.

Starbuck5:
I don't think this is very high risk, so I'm going to merge it in

But if @tomschimansky or @rethanon wanted to test this, the mac wheels for this branch are in https://github.com/pygame/pygame/actions/runs/1489122638

rethanon:
> I don't think this is very high risk, so I'm going to merge it in
> 
> But if @TomSchimansky or @rethanon wanted to test this, the mac wheels for this branch are in https://github.com/pygame/pygame/actions/runs/1489122638

Tested on my Intel MacBook Pro running python 3.10 and MacOS 12.0.1, working great, thanks.

Issue #818, opened 2019-02-12
Some fonts are not fully displayed


me2beats:
only the bottom of a text displayed:

```
import pygame, sys
from pygame.locals import *
pygame.init()
windowSurface = pygame.display.set_mode((500, 400), 0, 32)
pygame.display.set_caption('Hello world!')
WHITE = (255, 255, 255)
basicFont = pygame.font.Font("New York Deco.ttf", 48)
text = basicFont.render('AR', True, WHITE)
textRect = text.get_rect()
textRect.centerx = windowSurface.get_rect().centerx
textRect.centery = windowSurface.get_rect().centery
windowSurface.blit(text, textRect)
pygame.display.update()

while True:
    for event in pygame.event.get():
        if event.type == QUIT:
            pygame.quit()
            sys.exit()
```
font: https://www.fontpalace.com/font-download/New+York+Deco/

The font is displayed fine for example in Godot Engine 3

**To do to close this:**

- [x] Test code above works on OSX in 2.0.0.dev10 or above.
- [x] Test code above works on Linux in 2.0.0.dev10 or above.

**Related Docs:**

https://www.pygame.org/docs/ref/font.html

markph0204:
Do you have other font examples as well?

I'm unsure if this is truly a pygame issue and more of a font styling issue, here's why...

I'm opening the font on my native OS (Mac OS X) with a tool provided named Font Book.

![screen shot 2019-02-17 at 8 50 43 am](https://user-images.githubusercontent.com/210862/52913977-25f57200-3291-11e9-9f57-8263857e9d1b.png)

Note how the baseline's look fine, but the cap height seems to be incorrectly identified in the font itself.  When I select a row of glyphs -- the highlight is not covering the entire glyph AND the rows are writing on top of one another.

Reference for font terms I am not an expert in:

![image](https://user-images.githubusercontent.com/210862/52913999-62c16900-3291-11e9-9e52-30019c3c1c7b.png)

Now notice another random font I grabbed off same website:

![screen shot 2019-02-17 at 8 54 06 am](https://user-images.githubusercontent.com/210862/52914024-9a301580-3291-11e9-833b-d4cdf798e36b.png)

This font displays properly AND when selected, captures the entire glyph.

Open to other suggestions on pygame problem or font problem on this one.
CC: @illume for input/guidance too.

illume:
This definitely seems a font issue.

[...]

PR #2859, opened 2021-11-22
Fix weird MacOS display surf weirdness

Starbuck5:
The contributors have been aware of this for a long time, but it doesn't impact normal pygame programs, so it has been low priority.

It does bite people occasionally though, so it's worth fixing.

While I was digging around display.c for something else I saw a comment `* ensure window is initially black */`. Basically, the perfect place to ensure the window is initially black.

I don't really understand the problem, display.c is very complex, but this is a very simple solution that does solve it.

One reproducible time this issue has gotten me is in testing #818

On Mac (pygame 2.0 and above probably), the example code produces
Screen Shot 2021-11-21 at 11 56 27 PM

But now it (correctly) produces
Screen Shot 2021-11-21 at 11 55 21 PM

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1302,8 +1302,10 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
         Py_DECREF(surface);
 
         /* ensure window is initially black */
-        if (init_flip)
+        if (init_flip) {
+            SDL_FillRect(surf, NULL, SDL_MapRGB(surf->format, 0, 0, 0));
             pg_flip_internal(state);
+        }
     }
 
     /*set the window icon*/


src_c/display.c
```diff
@@ -1303,6 +1303,7 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
 
         /* ensure window is initially black */
         if (init_flip)
+            SDL_FillRect(surf, NULL, SDL_MapRGB(surf->format, 0, 0, 0));
             pg_flip_internal(state);
```
ankith26, line 1307: The `if` condition needs flower brackets `{}`😄 without it only the first line is a part of the condition and the second line would execute regardless of the value of `init_flip`

Starbuck5: Damn, I knew a one line fix was too good to be true.

ankith26 (approved):
👍 nice PR

Starbuck5:
Also, just to be clear, this fixes the "weird green lines" thing. I just didn't take screenshots of it. But the "weird green lines" thing is included in "display weirdness" :smile:

MyreMylar (approved):
LGTM 👍

PR #2861, opened 2021-11-24
Added attributes to surface.

novialriptide:


novialriptide:
I have a quick question, for documentation, should I add onto the description of the existing methods? 

For example,
```
   .. method:: get_width

      | :sl:`get the width of the Surface`
      | :sg:`get_width() -> width`
      | :sg:`width -> width`

      Return the width of the Surface in pixels.

      .. ## Surface.get_width ##
```

novialriptide:
Are we still set about this feature? There was some debate on this being added in the Discord server. @ankith26

Starbuck5:
I don't think this should go in.

We haven't fully considered the system, and putting in this API means we would need to support it forever.

I really don't want to be discouraging to @novialriptide though, it's super cool that you've gotten this to work! And I'm happy to have you working on the deprecation/naming issue.

novialriptide:
> I don't think this should go in.
> 
> We haven't fully considered the system, and putting in this API means we would need to support it forever.
> 
> I really don't want to be discouraging to @novialriptide though, it's super cool that you've gotten this to work! And I'm happy to have you working on the deprecation/naming issue.

Thanks, I'll close this pr now.

Issue #2775, opened 2021-10-16
Rename suggestion: pygame.math.Vector2.rotate_ip_rad to rotate_rad_ip and so on


algorithmpub:
**Rename in-place methods so they end with '_ip'**

Just a suggestion to rename some methods in pygame.math to improve clarity and keep consistency.

Some relatively new methods which perform changes in-place in pygame.math were introduced. Such methods would be:

- pygame.math.Vector2.rotate_ip_rad();
- pygame.math.Vector3.rotate_ip_rad();
- pygame.math.Vector3.rotate_x_ip_rad();
- pygame.math.Vector3.rotate_y_ip_rad();
- pygame.math.Vector3.rotate_z_ip_rad();

However, I believe it would be clearer and more consistent if they actually ended in '_ip' instead of having '_ip' in the middle of their names. Having to remember if an in-place method ends in '_ip' or has '_ip' in the middle of the function name seems like a potential source of typos/mistakes.

Additionally, if I recall correctly, up until recent times, all in-place methods in pygame.math ended in '_ip'.

Finally, if this change is to be implemented, now would be the best time, since there's not much time since the methods were added, so the change would affect almost no codebase.

Starbuck5:
Perhaps we should do the same thing we did with `mouse.set_system_cursor`:
Undocument old way, document / implement the new way, give the old way a deprecation warning with a long window.

PR #2863, opened 2021-11-25
Renamed some Vector methods to keep consistency

novialriptide:
This is in response to https://github.com/pygame/pygame/issues/2775


Starbuck5:
Hey, thanks for this.

I see that you treated the old methods as aliases, which works.

However, I would like them to also emit deprecation warnings. Here's an example of a deprecation warning from c code: https://github.com/pygame/pygame/blob/main/src_c/mouse.c#L379-L381. And then you should be able to forward the PyObject args and kwargs (potentially) into the proper function.

novialriptide:
@Starbuck5 It looks my pr is ready

ankith26 (commented):
A mention of the deprecated functions in the docs might be a good thing to do :)

src_c/math.c
```diff
@@ -3054,24 +3101,32 @@ static PyMethodDef vector3_methods[] = {
...
+    {"rotate_y_rad_ip", (PyCFunction)vector3_rotate_y_rad_ip, METH_O,
+     DOC_VECTOR3ROTATEYIPRAD},
     {"rotate_y_ip_rad", (PyCFunction)vector3_rotate_y_ip_rad, METH_O,
      DOC_VECTOR3ROTATEYIPRAD},
     {"rotate_z", (PyCFunction)vector3_rotate_z, METH_O, DOC_VECTOR3ROTATEZ},
     {"rotate_z_ip", (PyCFunction)vector3_rotate_z_ip, METH_O,
      DOC_VECTOR3ROTATEZIP},
     {"rotate_z_rad", (PyCFunction)vector3_rotate_z_rad, METH_O, DOC_VECTOR3ROTATEZRAD},
+    {"rotate_z_rad_ip", (PyCFunction)vector3_rotate_z_rad_ip, METH_O,
+     DOC_VECTOR3ROTATEZIPRAD},
```
ankith26, line 3129: the doc macro name on this line should be fixed. Also the same fix must be applied on other functions

ankith26 (commented):
Ok left some more reviews. But overall this PR is looking great! Infact we were just discussing on discord whether this feature should make it to pygame 2.1.1, with Starbuck5 supporting this feature to make it in, at the earliest! I'm fine with this coming in any version really, so perhaps we need more people to review and weigh in their thoughts to decide which version this gets added

MyreMylar (approved):
These changes all look good to me, the new naming is more consistent with the current pattern in the math module. It is also a good work out for deprecation warnings which I expect us to occasionally need to exercise in the future of pygame. 

EDIT: I say get it in ASAP as the faster it is in the less time there is for people to use the old API.

👍

src_c/doc/math_doc.h
```diff
@@ -21,7 +21,7 @@
 #define DOC_VECTOR2ROTATE "rotate(angle) -> Vector2\nrotates a vector by a given angle in degrees."
 #define DOC_VECTOR2ROTATERAD "rotate_rad(angle) -> Vector2\nrotates a vector by a given angle in radians."
 #define DOC_VECTOR2ROTATEIP "rotate_ip(angle) -> None\nrotates the vector by a given angle in degrees in place."
-#define DOC_VECTOR2ROTATEIPRAD "rotate_ip_rad(angle) -> None\nrotates the vector by a given angle in radians in place."
+#define DOC_VECTOR2ROTATEIPRAD "rotate_rad_ip(angle) -> None\nrotates the vector by a given angle in radians in place."
```
ankith26, line 24: The doc macro should read `DOC_VECTOR2ROTATERADIP` here :)

PS: These doc headers are autogenerated by running `python3 setup.py docs`, you don't have to manually change these

src_c/math.c
```diff
@@ -2059,6 +2059,15 @@ vector2_rotate_ip_rad(pgVector *self, PyObject *angleObject)
     Py_RETURN_NONE;
 }
 
+static PyObject *
+vector2_rotate_ip_rad(pgVector *self, PyObject *angleObject)
+{
+    if (PyErr_WarnEx(PyExc_DeprecationWarning, "vector2_rotate_rad_ip() now has all the functionality of vector2_rotate_ip_rad(), so vector2_rotate_ip_rad() will be deprecated in pygame 2.2", 1) == -1) {
```
ankith26, line 2065: the version number here needs updating. And similar changes needed in other functions

ankith26 (approved):
Looks good to go now! Thanks for the PR! 🎉 👍

Starbuck5:
Alright, this is really close to ready.

The old ones could be untyped, but we're leaving it documented as deprecated, so it makes sense to leave the types as well, I suppose.

The final thing is the the _doc.h files are auto generated on the `python setup.py docs`. It looks like you edited these ones manually, since it changes when the docs are regenerated. If you generate them automatically the old and new method docstrings are both there, so I suppose the C code should use the correct docstring for each.

Also maybe whoever merges this should do a squash and merge, since there are lots of small commits?

novialriptide:
I squashed all the commits into one and regenerated the doc files.

MyreMylar (approved):
LGTM 👍

ankith26 (approved):
:+1: Thanks for the PR!

Issue #2850, opened 2021-11-17
colordict.py is missing a few CSS standard colors


tswast:
I noticed a few missing colors when I compare https://www.w3.org/wiki/CSS/Properties/color/keywords with https://www.pygame.org/docs/ref/color_list.html / https://github.com/pygame/pygame/blob/main/src_py/colordict.py

Since the pygame list is almost a superset of the CSS/HTML colors, I assume this is just an oversight.

**Description**

I'd like the following color names added to pygame:

* aqua (0, 255, 255, 255)
* crimson (220, 20, 60, 255)
* fuchsia (255, 0, 255, 255)
* indigo (75, 0, 130, 255)
* lime (0, 255, 0, 255)
* olive (128, 128, 0, 255)
* silver (192, 192, 192, 255)
* teal (0, 128, 128, 255)

**Resources**

Manual merging of the two lists of colors: https://docs.google.com/spreadsheets/d/18FombCm_PIGn63NBhB007SzH29qgs27Ci6Uh_iPKiWE/edit?usp=sharing

tswast:
P.S. I also noticed that the RGBA value for purple doesn't match the CSS standard, but perhaps that was intentional? It would be a breaking change to fix it now anyway.

* pygame: (160, 32, 240, 255)
* CSS: (128, 0, 128, 255)

Starbuck5:
Cool!

I feel like I've actually tried to use "teal" and "fuschia" and been surprised they don't exist in pygame.

If you want to add these, go ahead and open a pull request. You've already identified the source file for the colors, there's also a color_list.rst somewhere in `docs/` that would need to be updated as well.

Jumana-22:
Hello, 
I would like to be assigned to this issue.

erinsusanthomas:
Hi, 
I would like to contribute, please assign me to this issue

ankith26:
It's nice to see that both of you are eager to contribute to pygame :) maybe you can decide among yourselves one of you who wants to put up a PR for this issue, and the other can take up any other issue ;)
(PS: you don't need to be assigned to a issue to work on it, just opening a PR directly should be fine)

PR #2865, opened 2021-11-25
colordict.py is missing a few CSS standard colors #2850

GustavoQuintero99:
Added the missing css colors requested on the #2850 also added them into the color_list_rst (can undo if needed)


Starbuck5:
Thanks for this.

I don’t think purple should be changed though, because it would change the output of people’s existing programs.

GustavoQuintero99:
Sure, it was suggested on the issue, that's why I changed it, should I put back the original color?

Starbuck5:
Yes, please put back the original color.

ankith26 (approved):
LGTM! Thanks for the PR! 🎉 

PS: It might be worth adding a note somewhere in the docs stating that these new colours were added in pygame version `2.1.1`

Starbuck5 (approved):
LGTM, thanks.

I'm not sure how this would get a note in the docs, but it should be okay without, it's pretty minor. Definitely a note in the release notes though.

PR #2870, opened 2021-11-27
Fix misnamed class aliasing in init file

### LINKED ISSUE #2869: pygame sometimes will not start: possible error in __init__.py ?
I'm using pygame 2.1.0 on Windows 10 and Python 3.8.8.
On some computers, the line number #91 in pygame/__init__.py crashes with the message "color is not defined".
Are the lines 91 and 94 correct ?
Shouldn't you write :

import pygame.color
color = pygame.color.Color
# instead of
# color = color.Color

(same remark for bufferproxy on line 94)

Cheers,

Jerome

ankith26:
A quickfix to fix #2869 

I have no clue how pygame even imports, before this bugfix. It should straight up error while importing pygame but it does not. Anyways, this PR should fix this

### DIFF
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -88,10 +88,10 @@ from pygame.rwobject import encode_string, encode_file_path
 import pygame.surflock
 import pygame.color
 
-Color = color.Color
+Color = pygame.color.Color
 import pygame.bufferproxy
 
-BufferProxy = bufferproxy.BufferProxy
+BufferProxy = pygame.bufferproxy.BufferProxy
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
This seems harmless from my testing. Perhaps it will make pygame work for the issue reporter.

👍

Starbuck5 (approved):
Weird stuff

But as MyreMylar said, it doesn't hurt anything as far as I can tell, and there seems to be a good chance it will fix things for the issue reporter.

Issue #2681, opened 2021-08-17
Rect is iterable but not typed or documented


vnmabus:
So, it seems that `Rect` is iterable and you can do

```python
r = Rect(0, 1, 2, 3)

x, y, w, h = r
```
However I do not see this documented anywhere, nor exposed in the type hints. Thus, Mypy complains but the code works.

Is this an oversight, or is this feature not supposed to be used?

Starbuck5:
I knew about this, I don't remember how though.

Rect objects are just supposed to behave like sequences of 4 numbers, and iteration is part of that.

They can also be directly indexed, is getitem typed?

It would be great to have a small note in the docs about these two behaviors.

Anyways, I think `__iter__` can be typed, in fact I believe I did that for the new Cursor object.

ankith26:
Related to this issue, `Color` is iterable but that is not typed as well. I am opening a misc typehints fix PR RN, but I will leave this issue for anyone else interested to tackle, maybe @Jumana-22 is interested to take this up (if not, I or someone else could do it as well)

robertpfeiffer:
is there a way to do destructuring assignment that isn't exposing an iterator?

bkwabiad:
Hi, I would like to contribute to this issue, I can submit a PR with a misc typehints fix within 2 days if this is fine with @ankith26 and others?

ankith26:
> is there a way to do destructuring assignment that isn't exposing an iterator?

AFAIK that's not possible ATM with python

> Hi, I would like to contribute to this issue, I can submit a PR with a misc typehints fix within 2 days if this is fine with @ankith26 and others?

Yeah, that would be cool for me! But I had recommended this issue to @Jumana-22 as well, so could you decide with them on who would work on this issue, so that we don't have 2 people working on the same issue :)

bkwabiad:
Hi @Jumana-22, wondering if you would like to take on this issue. I believe I have a fix for this issue but if you would like to take this on I will contribute in other ways!

novialriptide:
It looks like this issue hasn't moved anywhere in a while. I'll start working on it.

lordmauve:
Did we reject the idea of making Rect not iterable, not quacking like a 4-item sequence? I think maybe there was a discussion on Discord about

```
(x, y) in rect
```

being allowed but always returning `False`, which can trip people up.

If people want the coordinates out we could add an attribute like `.xywh`

illume:
[...]

PR #2556, opened 2021-04-10
Add unit test: display.gl_set_attribute() #1748

anesantos:

PR #2874, opened 2021-11-29
Added unit cases for gl_set_attribute()

Jumana-22:


ankith26:
I'm afraid I'd have to close this issue because it's a duplicate of #2556
Thanks for considering contributing to pygame, but please don't open PRs for stuff that already have PRs open for them, because we can't merge two PRs doing similar things as they'd conflict.
But since you've shown interest in contributing to pygame, I'd suggest looking into a good first issue related to typehints, #2681 😄

Issue #2829, opened 2021-11-09
AbstractGroup draw return type breaks Liskov substitution principle


vnmabus:
Currently `AbstractGroup`'s `draw` method returns `None`, but it has subclasses that return a different type (list of rects). I propose to change `AbstractGrop`'s `draw` method type hints to return `object` instead.

PR #2877, opened 2021-12-02
Changed AbstractGroup.draw() to return Rect_list

novialriptide:
In response to: https://github.com/pygame/pygame/issues/2829


ankith26 (changes requested):
Some changes would be needed in the type stubs file for sprite, [this](https://github.com/pygame/pygame/blob/main/buildconfig/pygame-stubs/sprite.pyi) is the file that you'd need to update

ankith26 (commented):
Ah I just realised, the docs of `sprite.Group.draw` method needs updating because it inherits from `AbstractGroup`

novialriptide:
Sorry for this.. I accidentally deleted the repository so I need to reopen another pull request.

PR #2880, opened 2021-12-03
Fix drawing of big ellipses

### LINKED ISSUE #2879: Ellipse drawing issue
Placeholder issue for this:
https://stackoverflow.com/questions/70209526/large-ellipse-draws-as-a-diamond

MightyJosip:
This fixes #2879
Before this patch with big enough value calculation can exceed the int max value on the line 1912 and 1914 (same can be said about the function that draws ellipse with line width)
https://github.com/pygame/pygame/blob/40991f75e8b8c14a91acf43d4b16788c409abbf4/src_c/draw.c#L1912-L1914
And because of that you can end up with that weird shape (won't get into much details about why exactly do you get "rotated square")
Edge value was if the width was 1024 and height was 1024; or anything that gives you the bigger value of (width^2)*height (because that is 2^31, and int supports up to 2^31 - 1)

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1891,7 +1891,8 @@ static void
 draw_ellipse_filled(SDL_Surface *surf, int x0, int y0, int width, int height,
                     Uint32 color, int *drawn_area)
 {
-    int dx, dy, x, y, x_offset, y_offset;
+    long long dx, dy, x, y;
+    int x_offset, y_offset;
     double d1, d2;
     if (width == 1) {
         draw_line(surf, x0, y0, x0, y0 + height - 1, color, drawn_area);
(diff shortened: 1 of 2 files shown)


ankith26 (approved):
LGTM, nice PR! 🥳 👍 
Not something that should block this PR from being merged, but maybe a unit test for drawing large ellipses might be a good idea?

MyreMylar:
Before PR:

image


After PR:

image


Test code used:

```
import pygame

pygame.init()


screen = pygame.display.set_mode((1920, 1080))

pygame.draw.ellipse(screen, pygame.Color('#FF0000'), pygame.Rect((0, 0), (2048, 2048)), 2)
clock = pygame.time.Clock()
running = True

while running:

    clock.tick(60)

        if event.type == pygame.QUIT:
            running = False
            running = False

    pygame.display.update()

```

LGTM 👍 


I think adding a test case is a good idea. I would suggest adding one large size to `test_ellipse` here:

https://github.com/pygame/pygame/blob/c15cc0dd8b4e8b808e658ec89adb02d105e14697/test/draw_test.py#L502

Then adding an additional test after this:

https://github.com/pygame/pygame/blob/c15cc0dd8b4e8b808e658ec89adb02d105e14697/test/draw_test.py#L512-L514

if the `border_width = 1` that instead checks four points, inwards 2 pixels and rotated 45 degrees inside the ellipse are *not* the ellipse colour. 

I'm not sure of the maths of how to calculate those points though... maybe this could help:

https://math.stackexchange.com/questions/3200370/rotate-a-point-on-an-ellipse-by-an-angle-and-calculate-the-distance-between-them

Orrrr you could bodge it a little because with the current test values of 4 & 5 and the proposed new one of `(2048, 2048)` the ellipse is basically a circle which makes it a lot easier to calculate?

MyreMylar (changes requested):
Changing to request changes to ask for a test please :)

MyreMylar (approved):
LGTM 👍

Issue #2862, opened 2021-11-24
pygame.draw.rect() with width > 1 has weird corners


lordmauve:
**Environment:**

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Linux (Ubuntu)
- **Python version** (e.g. 3.7.9, 3.8.5) : 3.9.7
- **SDL version** (e.g. SDL 2.0.12): 2.0.16
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.1.0
**Current behavior:**

![rect](https://user-images.githubusercontent.com/617224/143325878-37b13319-d33e-4c83-9af9-407ddc6d5f59.png)

On the left is the output of 

```python
pygame.draw.rect(screen, (255, 255, 0), Rect((20, 20), (100, 100)), width=9)
```

The right is what I expect. I implemented this by drawing 4 filled rects.

Starbuck5:
I think this has been brought up before, but it would be a compatibility break?

Although I’d be fine with fixing it, in theory.

Iirc the rect method actually calls the polygon draw routine internally for most cases, which may be why this effect is there in the first place.

lordmauve:
I doubt it's an important compatibility break. The output is strictly different but so is line rendering in every Pygame release. I don't think the results of pygame.draw should be guaranteed to be identical across releases. They just need to do the thing that they are specified to do.

This would be fixed if the polygon draw routine produced mitred corners.

ankith26:
+1, I too don't think fixing this would break backwards compatibility much, because it could be considered a bug from older versions. OFC, the fully enclosed rect seems much neater and correct

MyreMylar:
I also agree that squared off corners are a better result from this function if it can be made to happen. Especially now that there are various rounded corner options in pygame anyway, no real need to preserve this sort of crappy one for the future.

I also doubt it will hugely impact many users negatively in a backwards incompatible fashion. I suspect a lot of the mono-colour shape drawing code is *mostly* used by people starting out and then for initial testing code & debugging where users are less likely to care about little inconsistencies between versions.

Could just add a small note to the docs about the change and point users towards alternatives, in the unlikely case anyone is furious about gaining some sharper corners on their thick outlined rectangles. If somebody can thing of a good use case or has strong feelings I guess there could also be a boolean to toggle back to the 'gappy corners' approach of the present day.

novialriptide:
I would like to add that the corners come out inconsistent in some cases.
[...]

PR #2881, opened 2021-12-03
Fix rect drawing with width

MightyJosip:
Not much explanation needed, this is a fix for #2862

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -39,11 +39,11 @@
 
 /* Declaration of drawing algorithms */
 static void
-draw_line_width(SDL_Surface *surf, Uint32 color, int x1, int y1, int x2,
-                int y2, int width, int *drawn_area);
+draw_line_width(SDL_Surface *surf, Uint32 color, int x1, int y1, int x2, int y2, int width,
+                         int *drawn_area);
 static void
 draw_line(SDL_Surface *surf, int x1, int y1, int x2, int y2, Uint32 color,
-          int *drawn_area);
+         int *drawn_area);
 static void
 draw_aaline(SDL_Surface *surf, Uint32 color, float startx, float starty,
             float endx, float endy, int blend, int *drawn_area);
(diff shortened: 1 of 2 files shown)


ankith26:
whoops, seems like the manylinux1 builds are not happy. Might have to do something with the fact that `centos5` uses an old gcc

MyreMylar (approved):
Tested it locally and it does what it says on the tin.

Looks good. I like the new, properly rectangular, corners.

illume (changes requested):
Thanks. Lgtm too.

A test case would be nice if possible. If you don’t have time, let me know… I can whip one up?

src_c/draw.c
```diff
@@ -67,21 +67,23 @@ static void
...
 draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
                 int width, Uint32 color, int top_left, int top_right,
                 int bottom_left, int bottom_right, int *drawn_area);
 
 // validation of a draw color
 #define CHECK_LOAD_COLOR(colorobj)                                         \
-    if (PyLong_Check(colorobj))                                            \
-        color = (Uint32)PyLong_AsLong(colorobj);                           \
+    if (PyInt_Check(colorobj))                                             \
+        color = (Uint32)PyInt_AsLong(colorobj);                            \
```
ankith26, line 86: This change breaks stuff, because the `PyInt_*` macros no longer exist

src_c/draw.c
```diff
@@ -511,7 +509,7 @@ arc(PyObject *self, PyObject *arg, PyObject *kwargs)
 {
     pgSurfaceObject *surfobj = NULL;
     PyObject *colorobj = NULL, *rectobj = NULL;
-    SDL_Rect *rect = NULL, temp;
+    GAME_Rect *rect = NULL, temp;
```
ankith26, line 512: use `SDL_Rect` in place of `GAME_Rect`

src_c/draw.c
```diff
@@ -2473,21 +2360,21 @@ MODINIT_DEFINE(draw)
...
     }
     import_pygame_rect();
     if (PyErr_Occurred()) {
-        return NULL;
+        MODINIT_ERROR;
     }
     import_pygame_surface();
     if (PyErr_Occurred()) {
-        return NULL;
+        MODINIT_ERROR;
```
ankith26, line 2375: do `return NULL;` instead of `MODINIT_ERROR;`

ankith26 (commented):
Whoops looks like a merge conflict fix went bad here 😅 
Left some reviews that should be fixed
Also once your PR is done you should do a `python3 setup.py format` to fix lint

MightyJosip:
Closing because this branch is in a weird state after rebasing, continued in #2927

PR #2893, opened 2021-12-07
Added .copy() to Vector2 and Vector3

novialriptide:
Response to this issue: https://github.com/pygame/pygame/issues/2892


novialriptide:
Closed because somehow Github made me download an older version.

PR #2894, opened 2021-12-07
Added .copy() to Vector2 and Vector3

novialriptide:
https://github.com/pygame/pygame/issues/2892


bydariogamer:
Really useful PR!
A method for copying vectors was needed and this does its work perfect.
The tests run OK, and using `vector_to_copy.copy()` is faster than `pygame.Vector2(vector_to_copy)`:
```
>>> timeit("pygame.Vector2(pygame.Vector2(1.0, 2.0))","import pygame",number=10000000)
2.9492202239998733
>>> timeit("pygame.Vector2(1.0, 2.0).copy()","import pygame",number=10000000)
2.4121558550000373
```
The docs are right too.

bydariogamer (approved):

src_c/math.c
```diff
@@ -779,6 +781,17 @@ vector_nonzero(pgVector *self)
     return 0;
 }
 
+static PyObject *
+vector_copy(pgVector *self, PyObject *args)
```
ankith26, line 785: Here the `PyObject *args` argument is redundant because the copy method is a `METH_NOARGS` anyways

test/math_test.py
```diff
@@ -105,6 +105,11 @@ def testSequence(self):
         self.assertEqual(v.x, 9.1)
         self.assertEqual(v.y, 11.12)
 
+        v_copy0 = Vector2(2004.0, 2022.0)
+        v_copy1 = v_copy0.copy()
+        self.assertEqual(v_copy0.x, v_copy1.x)
+        self.assertEqual(v_copy0.y, v_copy1.y)
```
ankith26, line 111: perhaps these tests must be in their own test function, something like `test_copy`?

ankith26 (approved):
LGTM, thanks for the PR! 🎉 👍 
Apart from the minor reviews, this PR is good to go

ankith26 (commented):
Oh also, these newly added functions need type stubs

novialriptide:
Once I've received 2 approving reviews, I can squash all of these commits into one.

ankith26 (approved):
LGTM, thanks for the PR! :+1: 🎉

MyreMylar (approved):
LGTM 👍 

Tested it locally and it all works and performs as I expected.

PR #2896, opened 2021-12-08
Changed AbstractGroup.draw() to return List[Rect]

novialriptide:


ankith26 (changes requested):
Continuing my review from the other PR, the docs of `sprite.Group.draw` method needs updating because it inherits from `AbstractGroup`

ankith26 (approved):
Oh, seems like the docs were already updated and I somehow missed that while making my previous review.
Now this PR looks good, thanks! 🎉 :+1:

MyreMylar (approved):
LGTM

PR #2897, opened 2021-12-08
Add __contains__ method to the Rect class

### LINKED ISSUE #2795: Some lil' things to add in next Pygame releases
(@illume said I could put this here instead of #2794 )

**Description**
I have some things I'd like to add into future releases :

- copy and flip methods for the Vector objects ;
- a copy method for Font objects (font.Font and freetype.Font) ;
- a crop method for pygame.Surface objects (would work in a similar way as Surface.subsurface but without linking the two surfaces together) ;
- a `__contains__` method for Rect objects, which would make typical syntaxes like this : rect1 in rect2 OK and would work as an alternative to rect1.contains(rect2);
- some other kinds of sprite Groups that can filter the sprite types they can include and have an automated way of managing layers for those sprites

(I may update this list if I have other ideas)

michaelasp:
Implements the __contains__ method with rect objects. Can now use `rect in other_rect` syntax for checking rectangle containment. 

See tests implemented in rect_test.py for how this works and for working unit tests. 

@bkwabiad worked with me on this and has been the one conversing with contributors on this issue. Partially resolves #2795 

We have discussed with contributors on the discord about some of these features and decided that rectangle containment is feasible and seems to work here as implemented. Suggestions welcome!


src_c/rect.c
```diff
@@ -1131,6 +1132,23 @@ pg_rect_length(PyObject *_self)
...
+    if (!(self = pgRect_FromObject(self_uncast, &temp))) {
+        return -1;
+    }
+
+    contained = (self->x <= argrect->r.x) && (self->y <= argrect->r.y) &&
+                (self->x + self->w >= argrect->r.x + argrect->r.w) &&
+                (self->y + self->h >= argrect->r.y + argrect->r.h) &&
+                (self->x + self->w > argrect->r.x) &&
+                (self->y + self->h > argrect->r.y);
+    return contained;
```
ankith26, line 1149: seems like there's some code repetition here. The function `pg_rect_contains` above does almost exactly the same thing, except it returns a python bool object. Perhaps this function can be renamed to something like `_pg_rect_contains`, and then `pg_rect_contains` can call this function, something like 
```c
static PyObject *
pg_rect_contains(pgRectObject *self, PyObject *args)
{
    return PyBool_FromLong(_pg_rect_contains(self, args);
}
```

michaelasp: Makes sense. Will go ahead and make it more modular!

michaelasp: The other issue stems from the fact that __contains__ passes the argument as the first function. I'm still a little unskilled with using c and python together so if theres a way to typecast this properly to be able to easily flip arguments I will do this. Leaving as is for now!

michaelasp: All done, added a reverse boolean to make sure that the reversal works properly!

src_c/rect.c
```diff
@@ -956,25 +957,51 @@ pg_rect_clipline(pgRectObject *self, PyObject *args)
     return Py_BuildValue("((ii)(ii))", x1, y1, x2, y2);
 }
 
-static PyObject *
-pg_rect_contains(pgRectObject *self, PyObject *args)
+static int
+_pg_rect_contains(pgRectObject *self, PyObject *argrect, bool rev)
```
ankith26, line 961: Why is the `rev` required here? Actually we don't need this reversal, IMHO `big_rect.contains(arg_rect)` should be identical to `arg_rect in big_rect`. That way no reversal is needed, also it makes more logical sense

src_c/rect.c
```diff
@@ -956,25 +957,51 @@ pg_rect_clipline(pgRectObject *self, PyObject *args)
...
+    return _pg_rect_contains(self, arg, true);
+}
 
-    return PyBool_FromLong(contained);
+static PyObject *
+pg_rect_contains(pgRectObject *self, PyObject *arg)
+{   
+    int ret = _pg_rect_contains(self, arg, false);
+    if (ret < 0) {
+        RAISE(PyExc_TypeError, "Argument must be rect style object");
```
ankith26, line 998: The error must be returned here, like `return RAISE(...);` and not just `RAISE(...);`

michaelasp: Woops that was a misstep when converting. Fixed! Been working too much in c++ :(

src_c/rect.c
michaelasp: Ah yep that does make more sense, and then that means no reversal is needed. I was going off the assumption that the 1st rect would be the the one containing the second but actually the second argument should be the one that contains the 1st to make logical sense. Fixed!

src_c/rect.c
```diff
@@ -956,25 +956,36 @@ pg_rect_clipline(pgRectObject *self, PyObject *args)
...
+    if (!(argrect = pgRect_FromObject((PyObject *)arg, &temp_arg))) {
+        return -1;
     }
 
     contained = (self->r.x <= argrect->x) && (self->r.y <= argrect->y) &&
                 (self->r.x + self->r.w >= argrect->x + argrect->w) &&
                 (self->r.y + self->r.h >= argrect->y + argrect->h) &&
                 (self->r.x + self->r.w > argrect->x) &&
                 (self->r.y + self->r.h > argrect->y);
+    return contained;
```
ankith26, line 973: minor nitpick here, you can probably return `contained` directly without storing it in a temporary variable :)

test/rect_test.py
```diff
@@ -704,16 +704,28 @@ def test_contains(self):
...
         )
+        self.assertTrue(
+            Rect(2, 3, 0, 0) in r, "r does not contain an empty rect within its bounds",
+        )
         self.assertFalse(r.contains(Rect(0, 0, 1, 2)), "r contains Rect(0, 0, 1, 2)")
         self.assertFalse(r.contains(Rect(4, 6, 1, 1)), "r contains Rect(4, 6, 1, 1)")
         self.assertFalse(r.contains(Rect(4, 6, 0, 0)), "r contains Rect(4, 6, 0, 0)")
+        self.assertFalse((Rect(0, 0, 1, 2) in r), "r contains Rect(0, 0, 1, 2)")
+        self.assertFalse((Rect(4, 6, 1, 1) in r), "r contains Rect(4, 6, 1, 1)")
+        self.assertFalse((Rect(4, 6, 0, 0) in r), "r contains Rect(4, 6, 0, 0)")
```
ankith26, line 728: some more minor nitpicks, there are some extra and redundant parenthesis here, perhaps those could be removed

ankith26 (approved):
Thanks for the PR! 🎉 👍 
Overall this is looking good! But I've left some minor reviews that could be addressed before the merge

src_c/rect.c
michaelasp: Done!

test/rect_test.py
michaelasp: Fixed!

Starbuck5:
I haven't looked at this at all, but I just want to bring this up:

I'm pretty sure something like `5 in pygame.Rect(1,2,5,3)` works now, and that behavior needs to be preserved.

michaelasp:
Makes sense, I can make sure that behavior stays to not break anything

michaelasp:
Done! Now we have backwards compatibility for that use case too.

buildconfig/pygame-stubs/rect.pyi
```diff
@@ -181,6 +181,16 @@ class Rect(object):
     def fit(self, left: float, top: float, width: float, height: float) -> Rect: ...
     def normalize(self) -> None: ...
     @overload
+    def __contains__(self, rect: Union[_RectStyle, Rect]) -> bool: ...
```
ankith26, line 184: hmm since an int is allowed for a contains check, we can probably do `Union[_RectStyle, Rect, int]` here

src_c/rect.c
```diff
@@ -956,25 +956,42 @@ pg_rect_clipline(pgRectObject *self, PyObject *args)
...
-
-    if (!(argrect = pgRect_FromObject(args, &temp))) {
-        return RAISE(PyExc_TypeError, "Argument must be rect style object");
+static int
+_pg_rect_contains(pgRectObject *self, PyObject *arg)
+{
+    GAME_Rect *argrect, temp_arg;
+    if (!(argrect = pgRect_FromObject((PyObject *)arg, &temp_arg))) {
+        // Case where argument is one of (x,y,w,h)
+        if (PyLong_Check(arg)) {
```
ankith26, line 965: I don't think this part of the code should run when `pg_rect_contains` calls this function because `Rect.contains` is not supposed to handle integer coordinates.

michaelasp: Alright, I think then the best way is to make a base helper function which both the explicit and implicit contains methods call to make sure the behavior is slightly different for each. Pushed a commit like that, let me know what you think!

buildconfig/pygame-stubs/rect.pyi
michaelasp: Done!

buildconfig/pygame-stubs/rect.pyi
```diff
@@ -181,6 +181,16 @@ class Rect(object):
     def fit(self, left: float, top: float, width: float, height: float) -> Rect: ...
     def normalize(self) -> None: ...
     @overload
+    def __contains__(self, rect: Union[_RectStyle, Rect, int]) -> bool: ...
+    @overload
+    def __contains__(
+        self,
+        left_top: Union[List[float], Tuple[float, float], Vector2],
+        width_height: Union[List[float], Tuple[float, float], Vector2],
+    ) -> bool: ...
```
ankith26, line 190: Ah, seems like there's a merge conflict in here, sorry it happened because one of my typehint fix PRs got merged. To fix this on this PR, replace any `_RectStyle` with `_CanBeRect`

Also `Union[List[float], Tuple[float, float], Vector2]` can be replaced with `_Coordinate` which is imported from `_common` typestub file

src_c/rect.c
ankith26: Yes, good idea. And nice fix!

buildconfig/pygame-stubs/rect.pyi
michaelasp: Done

MyreMylar:
I like this. 

I tested it with this test program:

```
import pygame

pygame.init()


rect_1 = pygame.Rect((300, 200), (200, 200))
rect_2 = pygame.Rect((100, 100), (50, 50))

print("Back compat works") if 300 in rect_1 else print("Nope!")
print("Back compat works") if 200 in rect_1 else print("Nope!")
print("Back compat works") if 50 in rect_2 else print("Nope!")
print("Back compat works") if 100 in rect_2 else print("Nope!")
print("Back compat works") if 99 not in rect_2 else print("Nope!")

red = pygame.Color("red")
green = pygame.Color("green")

clock = pygame.time.Clock()
running = True

while running:

    clock.tick(60)

        if event.type == pygame.QUIT:
            running = False
            running = False

    rect_2.center = pygame.mouse.get_pos()

    screen.fill((0, 0, 0))
    pygame.draw.rect(screen, red, rect_1, width=1)
    pygame.draw.rect(screen, green if rect_2 in rect_1 else red, rect_2, width=1)

    pygame.display.update()
```

and all behaved as expected.

MyreMylar (approved):
I considered if we needed to mention anything about this in the docs, but it is a fairly small change that doesn't add, or take away any overall functionality to the submodule - it just adds a more pythonic way to access a part of it that I enjoy.

Thanks for making this PR 👍

Issue #2786, opened 2021-10-23
pygame.mixer.music.get_pos() incorrect when called immediately after unpausing music


nfearnley:
**Environment:**

You can get some of this info from the text that pops up in the console when you run a pygame program.

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Windows
- **Python version** (e.g. 3.7.9, 3.8.5): 3.10
- **SDL version** (e.g. SDL 2.0.12): SDL 2.0.16
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.0.2
- **Relevant hardware** (e.g. if reporting a bug about a controller, tell us the brand & name of it):

**Current behavior:**
When unpausing music using `pygame.mixer.music.unpause()` the next call to `pygame.mixer.music.get_pos()` is incorrect. It includes the time that the music was paused.

**Expected behavior:**
After unpausing music using `pygame.mixer.music.unpause()` the next call to `pygame.mixer.music.get_pos()` should return the correct time since the beginning of the song.

**Screenshots**

N/A

**Steps to reproduce:**

Play an mp3 file using `pygame.mixer.music`, pause the playback, wait a short time, unpause the playback. An immediate call to `pygame.mixer.music.get_pos()` will return incorrect information.

**Test code**

If possible add a simple test program that shows the problem described in this report.
```
from time import sleep
import pygame
import pygame.mixer

def test_pgmusic():
    pygame.init()
    pygame.mixer.init()
    pygame.mixer.music.load(str(Path(__file__).parent / "test_music.mp3"))
    pygame.mixer.music.play()

    # Wait 0.5s, then pause
    sleep(0.5)
    pygame.mixer.music.pause()
    # Wait 0.5s, get position, unpause, then get position again
    sleep(0.5)
    before_unpause = pygame.mixer.music.get_pos()
    pygame.mixer.music.unpause()
    after_unpause = pygame.mixer.music.get_pos()
    # Wait a single 60fps frame, then get position again
    sleep(1/60)
    after_wait = pygame.mixer.music.get_pos()

    assert before_unpause < after_unpause
    assert after_unpause < after_wait
```
[test_music.zip](https://github.com/pygame/pygame/files/7402341/test_music.zip)

PR #2899, opened 2021-12-09
Fix for paused music and get_pos()

JoshuaShequin:
Setting music_pos_time after an unpause to adjust for time spent in a pause for get_pos()

fixes this issue: [https://github.com/pygame/pygame/issues/2786](https://github.com/pygame/pygame/issues/2786)

The test script did need to be modified to allow for:
assert before_unpause <= after_unpause
instead of 
assert before_unpause < after_unpause

This is because the after_unpause is created so quickly that the song has not progressed yet and they are the same value.

### DIFF
--- a/test/mixer_music_test.py
+++ b/test/mixer_music_test.py
@@ -4,6 +4,7 @@ import os
 import sys
 import platform
 import unittest
+import time
 
 from pygame.tests.test_utils import example_path
 import pygame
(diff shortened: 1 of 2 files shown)


ankith26 (commented):
Thanks for the PR! 🎉 
Perhaps this change needs a unit test

MyreMylar (changes requested):
This seems to work comparing the before and after behaviour with the test code.

As @ankith26 says, it really needs a unit test. The mixer module needs a bunch of unit tests actually, so we should definitely get one in now while there is the opportunity.

I've prepared a simple one below that fails on the current main branch, but passes for me locally with this PR - you can use or adapt it, (some work may be required to make it through the CI tests on multiple platforms I'm not sure):

```
    def test_music_pause__unpause(self):
        """Ensure music has the correct position immediately after unpausing"""
        filename = example_path(os.path.join("data", "house_lo.mp3"))
        pygame.mixer.music.load(filename)
        pygame.mixer.music.play()

        # Wait 0.5s, then pause
        time.sleep(0.5)
        pygame.mixer.music.pause()
        # Wait 0.5s, get position, unpause, then get position again
        time.sleep(0.5)
        before_unpause = pygame.mixer.music.get_pos()
        pygame.mixer.music.unpause()
        after_unpause = pygame.mixer.music.get_pos()

        self.assertEqual(before_unpause, after_unpause)
```

Stick it into `mixer_music_test.py` at the bottom of the existing completed tests (under the queue ones). 

Otherwise seems a nice simple change.

test/mixer_music_test.py
```diff
@@ -178,6 +179,26 @@ def test_queue__invalid_filename(self):
...
+    def test_music_pause__unpause(self):
+        """Ensure music has the correct position immediately after unpausing
+
+        |tags:music|
+        """
+        filename = example_path(os.path.join("data", "house_lo.mp3"))
+        pygame.mixer.music.load(filename)
+        pygame.mixer.music.play()
+
+        # Wait 0.5s, then pause
```
illume, line 191: Is it possible to change these sleeps to a lower value like maybe 0.05? So the tests run a bit quicker.

JoshuaShequin: Yes and done

illume:
Thank you kindly :)

ankith26 (approved):
LGTM! Thanks for the PR 👍 🎈

PR #2901, opened 2021-12-10
Reform colors across the examples

### LINKED ISSUE #2860: Reform colors across the examples
In pygame 2.0, you can now put the arguments you would put into a pygame.Color object into any color field.

Despite this, I still see lots of code where people have a block at the top like:
```py
WHITE = (255,255,255)
BLACK = (0,0,0)
BLUE = (0,0,255)
```

Even though that is now completely unnecessary, because pygame has named colors built in easily pass-able to functions.

So I thought it would be a good idea to at least set our examples straight on the matter.

I read through some examples and aacircle, aliens, stars, and freetype_misc all could use some "color reform." (There might be more, these were just the ones I saw)

For example, this is unnecessary: https://github.com/pygame/pygame/blob/main/examples/freetype_misc.py#L36-L42

And these don't need to turn into pygame.Color objects explicitly: https://github.com/pygame/pygame/blob/main/examples/aacircle.py#L23-L24

maxsteep:
Thank you for allowing us to work on this fantastic project, please let us know if anything needs improvement/alterations.

Edit Starbuck: closes #2860


examples/scroll.py
```diff
@@ -28,7 +28,7 @@
 DIR_RIGHT = 4
 
 zoom_factor = 8
-
+ARROWCOLOR="black"
```
novialriptide, line 31: Spaces between the = sign would be nice.

novialriptide:
Nice PR! I just have 1 non-binding review about this.

maxsteep:
Thank you very much! Just added the whitespace, sorry for not putting there!

novialriptide (approved):
:)

examples/scroll.py
Starbuck5: Pygame uses black for formatting. I believe `setup.py lint` will run it over everything, but that's a very new workflow. You can `black` individual files or folders.

Starbuck5 (changes requested):
Thanks for putting this in, it's really cool that a student team is working on this.

Several bits of this aren't exactly what I had in mind.

If the constant at the top is only referenced once, or is mainly just the name of a color, it probably shouldn't be there.

For example, in aacircle.py
it could just be `pygame.draw.circle(screen, "green", (50, 100), 10)` on that line.

Or in aliens.py, why have WHITESPRITE when it could just be `self.color = "white"`. It saves a line and its easier to modify, since if you changed it to something else now you would also have to change the variable name, since it wouldn't be a white sprite anymore.

When mauve suggested the CONSTANT = "color name" pattern, I believe he meant it for THINGNAME = "color of the thing", not for COLORNAME = "color specified by name."

Like the BACKGROUNDCOLOR one in midi.py is good.

ARROWCOLOR in scroll.py follows the right pattern, but does it need to be there if it only occurs once?

Lastly, for the freetype_misc person presumably, it's cool that you figured out how to modify colordict to make the colors match perfectly (I wouldn't have been able to do that on my first contribution). But it's alright if the colors (of the result in freetype_misc) change a little on your change this time. I'd rather not change colordict.py though, since that is it's own whole deal.

lordmauve:
Agreed, it's really bad practice to write

```python
BLACK = "black"
WHITE = "white"
```

because the only reason to do this is to allow them to change, and it would be insane to change them to

```python
BLACK = "blue"
WHITE = "yellow"
```

Fuller thoughts on what constitutes best practices for string-valued constants: http://mauveweb.co.uk/posts/2014/12/dont-name-strings.html

maxsteep:
Absolutely - I want to apologize, I completely understand - we were, albeit naively, trying to follow the semantic proposed linking, but we implemented it on too simplistic of examples, both adding to the complexity and as Daniel Pope said definitely string-valued constants we proposed are logically wrong. We will make changes immediately, thank you for the feedback and for the helpful article!

maxsteep:
We removed the color defines for the 'joystick' example, made requested changes to 'aacircle' and 'aliens' examples, fully rolled back colordict changes, ensured freetype_misc using only pre-defined colordict profiles with the colors that match closest, removed black/white piano key color defines from midi as they are used once and represent real-life colors, semantically linked examples in the prevent_display_stretching example, and removed the semantic linking from scroll example as requested. Please let us know if any other changes must be made, and we apologize for defining everything, we were a little overzealous in semantic linking.

novialriptide (commented):
Also, once the PR has been approved by 2 members, can you squash all of the commits into 1?

Starbuck5:
This looks way better now!

Remaining notes:
I don't think it's necessary to call out colordict.py above the constants, that's more of an implementation detail than anything.
In prevent_display_stretching.py, the constants should probably go under the imports.

Pygame python code is formatted with black, you can `pip install black` and `black examples` from the project root, that will make sure everything is formatted properly.

And if you don't want to worry about squashing we can probably do a squash and merge through github while merging.

maxsteep:
Thank you!
Just implemented all the requested changes, for squashing we were thinking to squash and merge at the same step - or whatever you prefer, let us know please.

Starbuck5 (approved):
Thanks, this looks good to me!

IDK where the merge conflict came from but I resolved that.

maxsteep:
Thank you very much!

examples/camera.py
```diff
@@ -91,7 +91,9 @@ def main(self):
 
             self.get_and_flip()
             self.clock.tick()
-            pygame.display.set_caption("CAMERA! (" + str(round(self.clock.get_fps())) + " FPS)")
+            pygame.display.set_caption(
+                "CAMERA! (" + str(round(self.clock.get_fps())) + " FPS)"
```
ankith26, line 95: This could become an f-string for better readability, something like `f"CAMERA! ({self.clock.get_fps():.2f} FPS)"` where `.2f` rounds the float to 2 decimal places

illume (changes requested):
Looks good to me, thanks.

There’s only the f string thing already mentioned to fix. Squashing and rebasing the commits would be nice.

maxsteep:
Thank you very much, we implemented the f-string and we are figuring out squashing and rebasing.
We had to sync our fork to the current status of the pygame branch and hope that the recent 3 commits were not in error, only the f-string fix was expected to be seen by us.

ankith26 (approved):
LGTM! 👍 
Thanks for the PR 🎈

illume (approved):
Thanks. I'll "squash and merge" this in so it makes it in time for the next pygame release.

Issue #2889, opened 2021-12-05
Pygame causes GIL error when smooth scaling alpha surfaces


quasar098:
**Environment:**

- **Operating system**: Windows
- **Python version**: 3.9.1
- **SDL version**: 2.0.14
- **PyGame version**: 2.0.1
- **Relevant hardware**: Graphics card is a GTX 1660 SUPER

**Current behavior:**

An error is sometimes raised when trying to smooth scale a transparent object (SRCALPHA flag)

**Expected behavior:**

This should not happen
also this only happens with alpha surfaces for some reasons

**Screenshots**

![image](https://user-images.githubusercontent.com/70716985/144763310-476e520a-5032-4528-88c6-b5245ee56250.png)
![image](https://user-images.githubusercontent.com/70716985/144763333-1467148f-a7ee-4822-b8ed-b9eb1d205a93.png)
caused by line 67

**Steps to reproduce:**

1. Make a surface with SRC
2. Use Smooth scale
3. It crashes occasionally

Other details:
Process finished with exit code -1073740791 (0xC0000409)

ankith26:
Thanks for the bug report! Can you please post a minimal code example (in text) that reproduces this issue?

robertpfeiffer:
Is this just what a segfault looks like with the GIL released, or is this an actual threading problem?

MyreMylar:
Might be related to old issues with threading and smooth scaling seen here:

https://github.com/pygame/pygame/issues/1515

As some GIL changes were made.

I wonder if the small sizes of the surfaces used here are helping cause a race condition. i.e. something is finishing before it would normally because the surface is so small and the GIL is still released (after the changes mentioned in the PR above perhaps?) 

To the OP @quasar098 - it looks like you are doing something similar to what I was back in 2019 and I found then that the `transform.rotozoom` function was more stable while achieving the same result, so perhaps try that and see if your issue is resolved.

quasar098:
i can try to use rotozoom 
thanks for looking into to it

ankith26:
I still cant reproduce the original issue even though I am doing stuff similar to the code in the screenshot. Can you please paste (in text) another minimal reproducer of the problem?

yoyoberenguer:
Did you try to set the smoothscale backend to generic ? to see if this helps

quasar098:
I was reading some docs and read something about it once.
I ended up just using rotozoom and it seemed to not throw any errors

PR #2904, opened 2021-12-12
fix GIL release in transform module

ankith26:
I was playing around with #2889 looking in the transform module for GIL issues, and I did find some, although I'm not sure whether it can fix the issue, and I cannot test stuff without any reproducer OP has provided.

However I do find other issues, where the GIL is released in calls to `rotate90` and `chop` C helper functions, which in turn call a function that calls python API error reporting functions which are not supposed to be called while the GIL is released

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -200,6 +200,8 @@ rotate90(SDL_Surface *src, int angle)
     dst = newsurf_fromsurf(src, dstwidth, dstheight);
     if (!dst)
         return NULL;
+
+    Py_BEGIN_ALLOW_THREADS;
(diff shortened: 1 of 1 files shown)


illume:
Would a thread using test catch this error? 

Long term, maybe we should have a thread using test for every single pygame method.

ankith26:
> Would a thread using test catch this error?

Well no, this bug could be caught from any thread (even the main one) provided there is an internal SDL error that happens while creating the surface

> Long term, maybe we should have a thread using test for every single pygame method.

Ah, that sounds like a good idea

illume (approved):
👍 thanks

PR #2906, opened 2021-12-12
Working gamma test on windows

MyreMylar:
On windows at least, gamma values that are too dark to read are forbidden by the OS.

The set gamma test was failing locally and as they are not run on CI I suspect it slipped by for a while. There also appeared to be some odd behaviour when running the whole test suite where sometimes this test would pass and sometimes not - not sure on that, but running the display_test file on it's own and using subtests made the problem clear.
 
I added subtests and split up the tests just so it is easier to see which values are failing on any other platforms/ or in the future.

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -480,15 +480,25 @@ class DisplayModuleTest(unittest.TestCase):
     )
     def test_set_gamma(self):
         pygame.display.set_mode((1, 1))
-        gammas = [0.0, 0.25, 0.5, 0.88, 1.0]
+
+        gammas = [0.25, 0.5, 0.88, 1.0]
         for gamma in gammas:
-            self.assertEqual(pygame.display.set_gamma(gamma), True)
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
LGTM! 👍 🎉

Starbuck5:
This didn't fail in the past, and I suspected it was something in the great py2 and sdl1 cleanups that changed it.

Here's what it used to be:
https://github.com/pygame/pygame/blob/2.1.0/test/display_test.py#L501-L513

So this suggests it used to test without an active display on SDL2, and that that works.

And I tested it out in a command line repl, and it does work.

Although if you set the gamma super low and then try to open the display, it fires back with
```
pygame.error: SetDeviceGammaRamp(): The operation completed successfully.
```

And the gamma doesn't take effect until a window is open.

illume:
I think this is ok to merge but we should create a new issue about the last comment? 
https://github.com/pygame/pygame/pull/2906#issuecomment-992183049

illume (approved):
👍

illume:
Started an issue stub here: https://github.com/pygame/pygame/issues/2947 for more investigation in 'the future'.

Issue #2903, opened 2021-12-11
Ensure docs can be built without fully working build environment


Starbuck5:
Added recently to make it easier for newbies to contribute to docs: https://github.com/pygame/pygame/pull/2714

Regression from merge: https://github.com/pygame/pygame/commit/30637d3a3ee4d6b37a4925f5f7eb53c4f9861262#

Hoping @Grimmys can fix this one.

PR #2907, opened 2021-12-12
Fix docs building without config

Grimmys:
This style has to be privileged over `distutils` command pattern to avoid the checking of dependencies not needed to generate the docs.

See #2903 

I also took the opportunity to refactor a bit the process:

- Replace some bare `except` statements by `except Exception` because it's a code smell: https://www.flake8rules.com/rules/E722.html
- Check return code by sub process to throw an exception if needed, so if there is an error like Sphinx not installed it's printing the help message. Throw and catch a custom/more specific exception would be even better.

And I'm wondering if we need the except branch in `makeref.py` with the "Have you installed sphinx?" message. I tried to run the command without Sphinx installed because of fresh setup but I didn't got the message, just a message from the `subprocess` telling me that the given package is not installed. This branch seems unreachable for me, because `subprocess` is not throwing any exception but returning a code != 0 instead.

### DIFF
--- a/buildconfig/makeref.py
+++ b/buildconfig/makeref.py
@@ -10,6 +10,7 @@ rst_build_dir = os.path.join('docs', 'generated')
 rst_doctree_dir = os.path.join(rst_build_dir, 'doctrees')
 c_header_dir = os.path.join('src_c', 'doc')
 
+
 def run():
     full_generation_flag = False
     for argument in sys.argv[1:]:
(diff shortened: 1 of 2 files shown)


Starbuck5 (approved):
Thanks!
I tested that it works by deleting my SETUP file and seeing if it could run without generating a new one (since I obviously have a full pygame development environment).

Yeah, the code you're talking about in makeref.py does seem unreachable, if subprocess works how I think.

Issue #1677, opened 2020-04-27
pygame.font.Font constructor fails to properly open all file streams of type typing.BinaryIO


MyreMylar:
**Context:**

I am experimenting with generalised improvements to the speed and future proofing of loading resources in the [Pygame GUI](https://pygame-gui.readthedocs.io/en/latest/) package. 

From a fair bit of reading over the slightly muddled state of parallel programming in python, and after profiling a few other options - I determined that using a few python Threads `from threading import Thread` was likely a reasonable way to get some speed up when loading a long-ish list of files (five threads seems to be the optimum number on my dev machine).

My initial explorations were mainly done using pygame.Surfaces where, when loading about 20 of them, I saw roughly a 3x speed up when using 5 Threads over simple sequential loading - without any obvious (to me) negative side effects.

**Problem:**

While this multi-threaded loading approach works fine for pygame.Surfaces, it doesn't appear to work for pygame.font.Font objects, the other main resource used in Pygame GUI projects.

**Desired outcome**

pygame.font.Font objects are loadable on Threads other than the main thread - or somebody has an idea how I could achieve the same broad goal under the current system.

**Example code**

    import pygame
    
    from typing import Union, Any, Dict
    
    from collections import namedtuple
    from threading import Thread, Lock
    from importlib.resources import open_binary
    
    from queue import Queue
    
    from fps_counter import FPSCounter
    
    
    PackageResource = namedtuple('PackageResource', ('package', 'resource'))
    
    
    class FontResource:
        def __init__(self, font_id: str, size: int, style: Dict[str, bool],
                     location: Union[PackageResource, str]):
            self.font_id = font_id
            self.size = size
            self.style = style
            self.location = location
            self.loaded_resource = None
    
    
    class ImageResource:
        def __init__(self, image_id: str, location: Union[PackageResource, str]):
            self.image_id = image_id
            self.location = location
            self.loaded_resource = None
    
    
    def complete_image_load(image_resource: ImageResource) -> ImageResource:
        with open_binary(image_resource.location.package,
                         image_resource.location.resource) as open_resource:
            image_resource.loaded_resource = pygame.image.load(open_resource)
            return image_resource
    
    
    def complete_font_load(font_resource: FontResource) -> FontResource:
        with open_binary(font_resource.location.package,
                         font_resource.location.resource) as open_resource:
            font_resource.loaded_resource = pygame.font.Font(open_resource, font_resource.size)
[...]

PR #2911, opened 2021-12-14
Font rwops

Starbuck5:
This PR is a bugfix, and enhancement, and code cleanup.

Code cleanup: It follows #2695's quest to simplify pygame resource loaders by rewriting `font_init` to use higher level pygame rwobject constructs. This about halves the lines of code in `font_init`.

Enhancement: the higher level loading systems allow for better error messages from #2694. I also prettied up the docs, cleaning up some misleading statements and putting some more restructured text function linking in there.

Bugfix: despite the pygame default font + pyinstaller issue being officially fixed, people have still been running into it. Ankith and I neglected to test `--onefile` executables, and the issue persisted there. This fixes it, tested both with and without onefile.

Fun fact!
Pygame will load freesansbold.ttf if the user passes `None` or `"freesansbold.ttf"`. 
But if the user passes `"freesansbold.ttf"`, pygame actually looks locally first, then loads the default font if it can't find something called `"freesansbold.ttf"` locally.

Here's my testing file if anyone wants to use it.
```py
import sys
import pygame

pygame.init()

screen = pygame.display.set_mode((400, 200))

clock = pygame.time.Clock()

num = 0
if len(sys.argv) > 1:
    num = int(sys.argv[1])

# loads default font
if num == 0:
    font = pygame.font.Font(None, 100)

# loads a font file
if num == 1:
    font = pygame.font.Font("Oswald-Regular.ttf", 100)

# looks in current dir, if it can't find it, loads default font
# HOWEVER, does not scale the default font down like None does
if num == 2:
    font = pygame.font.Font("freesansbold.ttf", 100)

# loads a weird IO object
if num == 3:
    font = pygame.font.Font(open("Oswald-Regular.ttf", "rb"), 100)

# fails to load a font file
if num == 4:
    font = pygame.font.Font("not-your-font.ttf", 100)

im = font.render("Hello", True, "black")
print(im)

while True:
    screen.fill("purple")
    screen.blit(im, (0,0))

        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit

    pygame.display.flip()
    
    clock.tick(144)
```

### DIFF
--- a/docs/reST/ref/font.rst
+++ b/docs/reST/ref/font.rst
@@ -25,7 +25,7 @@ accessed by passing ``None`` as the font name.
 Before pygame 2.0.3, pygame.font accepts any UCS-2 / UTF-16 character
 ('\\u0001' to '\\uFFFF'). After 2.0.3, pygame.font built with SDL_ttf
 2.0.15 accepts any valid UCS-4 / UTF-32 character 
-(like emojis, if the font has them) ('\\U00000001' to '\\UFFFFFFFF')).
+(like emojis, if the font has them) ('\\U00000001' to '\\U0010FFFF')).
 More about this in :func:`Font.render`.
 
 Before pygame 2.0.3, this character space restriction can be avoided by
--- a/docs/reST/ref/font.rst
+++ b/docs/reST/ref/font.rst
@@ -8,29 +8,31 @@
 
 | :sl:`pygame module for loading and rendering fonts`
 
-The font module allows for rendering TrueType fonts into a new Surface object.
-It accepts any UCS-2 character ('\u0001' to '\uFFFF'). This module is optional
-and requires SDL_ttf as a dependency. You should test that :mod:`pygame.font`
-is available and initialized before attempting to use the module.
+The font module allows for rendering TrueType fonts into Surface objects.
+This module is built on top of the SDL_ttf library, which comes with all
+normal pygame installations.
 
-Most of the work done with fonts are done by using the actual Font objects. The
-module by itself only has routines to initialize the module and create Font
-objects with ``pygame.font.Font()``.
+Most of the work done with fonts are done by using the actual Font objects.
+The module by itself only has routines to support the creation of Font objects
+with :func:`pygame.font.Font`.
 
-You can load fonts from the system by using the ``pygame.font.SysFont()``
-function. There are a few other functions to help lookup the system fonts.
+You can load fonts from the system by using the :func:`pygame.font.SysFont`
+function. There are a few other functions to help look up the system fonts.
 
-Pygame comes with a builtin default font. This can always be accessed by
-passing None as the font name.
+Pygame comes with a builtin default font, freesansbold. This can always be
+accessed by passing ``None`` as the font name.
 
-To use the :mod:`pygame.freetype` based ``pygame.ftfont`` as
-:mod:`pygame.font` define the environment variable PYGAME_FREETYPE before the
-first import of :mod:`pygame`. Module ``pygame.ftfont`` is a :mod:`pygame.font`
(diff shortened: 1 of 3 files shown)


ankith26:
Can the last commit be reverted now (`IOError` check changing back to `FileNotFoundError`) because the freetype PR is merged now? After this branch has been rebased against main OFC ;)

If you need my help with rebasing or anything of that sort feel free to ask :smile:

illume (changes requested):
There's a request from @ankith26 about rebasing and such.

ankith26:
Perhaps this PR also fixes #1677 ? Although I've not done much testing yet, will update this comment with test results later

docs/reST/ref/font.rst
```diff
@@ -8,29 +8,31 @@
...
+You can load fonts from the system by using the :func:`pygame.font.SysFont`
+function. There are a few other functions to help look up the system fonts.
+
+Pygame comes with a builtin default font, freesansbold. This can always be
+accessed by passing ``None`` as the font name.
+
+Before pygame 2.0.3, pygame.font accepts any UCS-2 / UTF-16 character
+('\\u0001' to '\\uFFFF'). After 2.0.3, pygame.font built with SDL_ttf
+2.0.15 accepts any valid UCS-4 / UTF-32 character 
+(like emojis, if the font has them) ('\\U00000001' to '\\UFFFFFFFF')).
```
ankith26, line 28: minor doc nitpick here, the largest valid unicode codepoint is `"\U0010FFFF"`, trying to input strings like `"\U00110000"` and above is a `SyntaxError` in python.

src_c/font.c
```diff
@@ -745,128 +718,55 @@ font_init(PyFontObject *self, PyObject *args, PyObject *kwds)
...
-            }
-            if (test == NULL) {
-                if (!PyErr_Occurred()) {
-                    PyErr_Format(PyExc_IOError,
-                                 "unable to read font file '%.1024s'",
-                                 filename);
+    if (rw == NULL && PyUnicode_Check(obj)) {
+        PyObject *font_defaultname_string =
+            PyUnicode_FromString(font_defaultname);
+        if (!PyUnicode_Compare(font_defaultname_string, obj)) {
```
ankith26, line 728: perhaps it's better to use [PyUnicode_CompareWithASCIIString](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_CompareWithASCIIString) here, rather than converting to python string and comparing with `PyUnicode_Compare`. Not only is this more elegant code wise, it also has the advantage of not erroring, whereas the old code here can error, and that is not handled correctly

docs/reST/ref/font.rst
Starbuck5: Thank you, I saw that in testing but didn't know the actual thing to put. Is there a reason python doesn't support `"\U00110000"` to `"\U0010FFFF"` of unicode?

ankith26: According to the [unicode standard](https://en.wikipedia.org/wiki/Unicode), `"\U0010FFFF"` is the limit on unicode (this limit was chosen keeping in mind the max size of UTF-16 surrogate pair encoded characters

illume (changes requested):
Looking good... but ever watchful Ankith found another issue (or 2?) that needs addressing.

Starbuck5:
Thanks for the feedback Ankith, I finally fixed it.

ankith26 (approved):
LGTM! 👍  Thanks for the PR 🥳

illume:
Sorry, there's a conflict now.

illume (approved):
👍 🎉 Thanks

Starbuck5:
I decided to fix a conflict by just rebasing the whole thing.

PR #2916, opened 2021-12-15
Generate Cython code at build time

Starbuck5:
This hopes to reduce merge conflicts and huge diffs of Cython generated code. It will also stop the need to have PRs explicitly just regenerating the Cython generated code when new versions of Python come out, for example.

How does it work?
-runs `setup.py cython` on every build run of setup.py
-adds a pyproject.toml to tell build system that we need cython, requests, wheel, etc.
-has to run install through a python command now, `pip install .` instead of `setup.py install`, because the python command will do the build system as specified. (ubuntu-sdist isn't possible to run through a python command, so it explicitly installs cython)

Gets rid of the generated c files on the repo and gitignores them.


Starbuck5:
The remaining case to cover is the appveyor analyze case.

It should work with `--install-option="-enable-msvc-analyze"`, but it doesn't work on my computer, for a weird reason, I'll have to figure that out.
```
      LINK : fatal error LNK1104: cannot open file 'build\temp.win-amd64-3.9\Release\Users\charl\AppData\Local\Temp\pip-install-pt724jb5\cython_b24fa9b9d75640daac6f3710dd1e7e3c\Cython\Compiler\Scanning.cp39-win_amd64.exp'
      error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2019\\Community\\VC\\Tools\\MSVC\\14.29.30037\\bin\\HostX86\\x64\\link.exe' failed with exit code 1104
```

illume (commented):
Thanks. Made a few notes to consider.

dr0id:
> link.exe' failed with exit code 1104

Maybe that [Stackoverflow](https://stackoverflow.com/questions/49239036/dynamic-on-the-fly-compile) entry will help. Or the [hint (see last comment)](https://youtrack.jetbrains.com/issue/PY-29520) that the path is too long?

Starbuck5:
I'm going to give up on this for now, again.

PR #2919, opened 2021-12-16
Fix warnings, make warnings error on CI

### LINKED ISSUE #1316: Compiler warnings should fail on CI.

ankith26:
PR to close #1316

Fixes some compiler warnings (and also skips some warnings) and make warnings error on CI

Also fixes most MSVC static analyzer warnings (all except ones in `pygame.mask`) and those too error on CI

### Some skipped warnings (todos for future)
- [ ] GCC on linux complains about `unused-but-set-variable` while compiling freetype. There is a complex macro system here that caused the warning, that seemed hard to fix so I skipped it for now
- [ ] MSVC static analyzer complains while compiling `pygame.mask` about possible buffer overflows, and I'm not sure whether it's a false report or not here, so I skipped it here for now

### DIFF
--- a/.github/workflows/cppcheck.yml
+++ b/.github/workflows/cppcheck.yml
@@ -1,0 +1,32 @@
+name: cppcheck Static Analysis
+
+# Run cppcheck on src_c changes to main branch, or any PR to main.
+on:
+  push:
+    branches: main
+    paths:
+      - 'src_c/**'
+
+  pull_request:
+    branches: main
+    paths:
+      - 'src_c/**'
+
+# TODO: Any more static checkers can be added here
+jobs:
+  run-cppcheck:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - uses: actions/checkout@v2.3.4
+
+      - name: Install deps
+        run: |
+          sudo apt-get update --fix-missing
+          sudo apt-get upgrade
+          sudo apt install cppcheck
+
+      - name: Run Static Checker
+        # skip cppcheck on SDL_gfx and scrap for now
+        run: cppcheck src_c --force --enable=performance,portability,warning \
+          --suppress=*:src_c/SDL_gfx/* --suppress=*:src_c/scrap*
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -1205,7 +1205,7 @@ pgGetArrayInterface(PyObject **dict, PyObject *obj)
     if (!PyDict_Check(inter)) {
         PyErr_Format(PyExc_ValueError,
                      "expected '__array_interface__' to return a dict: got %s",
-                     Py_TYPE(dict)->tp_name);
+                     Py_TYPE(inter)->tp_name);
         Py_DECREF(inter);
         return -1;
     }
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -1171,10 +1171,20 @@ _pg_buffer_is_byteswapped(Py_buffer *view)
     if (view->format) {
         switch (view->format[0]) {
             case '<':
-                return SDL_BYTEORDER != SDL_LIL_ENDIAN;
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+                /* Use macros to make static analyzer happy */
+                return 0;
+#else
+                return 1;
+#endif
             case '>':
             case '!':
-                return SDL_BYTEORDER != SDL_BIG_ENDIAN;
+#if SDL_BYTEORDER == SDL_BIG_ENDIAN
+                /* Use macros to make static analyzer happy */
+                return 0;
+#else
+                return 1;
+#endif
         }
     }
     return 0;
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -40,7 +40,7 @@
 #define VC_EXTRALEAN
 #include <windows.h>
 extern int
-SDL_RegisterApp(char *, Uint32, void *);
+SDL_RegisterApp(const char *, Uint32, void *);
 #endif
 
 #if defined(macintosh)
@@ -374,7 +374,7 @@ pg_get_sdl_byteorder(PyObject *self)
 static void
 _pg_quit(void)
 {
-    int num, i;
+    Py_ssize_t num, i;
     PyObject *quit, *privatefuncs, *temp;
 
     /* Put all the module names we want to quit in this array */
@@ -755,7 +755,7 @@ _pg_new_capsuleinterface(Py_buffer *view_p)
     cinter_p->inter.two = 2;
     cinter_p->inter.nd = ndim;
     cinter_p->inter.typekind = _pg_as_arrayinter_typekind(view_p);
-    cinter_p->inter.itemsize = view_p->itemsize;
+    cinter_p->inter.itemsize = (int)view_p->itemsize;
     cinter_p->inter.flags = _pg_as_arrayinter_flags(view_p);
     if (view_p->shape) {
         cinter_p->inter.shape = cinter_p->imem;
@@ -1589,7 +1589,7 @@ _pg_values_as_buffer(Py_buffer *view_p, int flags, PyObject *typestr,
                         "'shape' and 'strides' are not the same length");
         return -1;
     }
-    view_p->ndim = ndim;
+    view_p->ndim = (int)ndim;
     view_p->buf = PyLong_AsVoidPtr(PyTuple_GET_ITEM(data, 0));
     if (!view_p->buf && PyErr_Occurred()) {
         return -1;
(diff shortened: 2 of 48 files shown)


illume:
Very good! It's my understanding that static analysis errors
- often cause bugs when the warnings are fixed
- often point to areas of code that are smelly in some way other than the warning they fix

So I want to take a bit more time than usual to review these by testing out the code in a debugger and such, as well as having a look at code/docs/tests around the areas fixed.

Probably the two open tasks could be turned into two new issues?

ankith26:
For the most part I'm fairly confident this PR is safe, but yeah I'd appreciate if you and/or anyone else could do a bit more checking around pixelarray and buffer stuff, as I'm not too sure about those fixes

src_c/base.c
```diff
@@ -1160,10 +1160,20 @@ _pg_buffer_is_byteswapped(Py_buffer *view)
     if (view->format) {
         switch (view->format[0]) {
             case '<':
-                return SDL_BYTEORDER != SDL_LIL_ENDIAN;
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+                /* Use macros to make static analyzer happy */
```
illume, line 1164: This is I guess also a bit faster :)

src_c/base.c
```diff
@@ -1578,7 +1588,7 @@ _pg_values_as_buffer(Py_buffer *view_p, int flags, PyObject *typestr,
                         "'shape' and 'strides' are not the same length");
         return -1;
     }
-    view_p->ndim = ndim;
+    view_p->ndim = (int)ndim;
```
illume, line 1591: This is surprising that ndim is an int. I guess it's not possible to create such large dimensions.

src_c/base.c
```diff
@@ -2028,15 +2038,6 @@ pg_uninstall_parachute(void)
 
 /* bind functions to python */
 
-static PyObject *
-pg_do_segfault(PyObject *self, PyObject *args)
```
illume, line 2032: heh. I didn't even realize this function was here, and can't remember what it was for.

Compiler: "I found a function pg_do_segfault that will segfault." Thanks compiler. Yes, probably we should get rid of it.

src_c/bitmask.c
```diff
@@ -39,7 +39,8 @@
 static INLINE unsigned int
 bitcount(BITMASK_W n)
 {
-    if (BITMASK_W_LEN == (32)) {
+    const int bitmask_len = BITMASK_W_LEN;
```
illume, line 42: I'm not entirely sure why this wasn't/isn't an `#if BITMASK_W_LEN == 32`?

src_c/camera_v4l2.c
```diff
@@ -177,8 +177,8 @@ v4l2_xioctl(int fd, int request, void *arg)
    currently two step processes. */
 /* TODO: Write single step conversions where they may actually be useful */
 int
-v4l2_process_image(pgCameraObject *self, const void *image,
-                   unsigned int buffer_size, SDL_Surface *surf)
+v4l2_process_image(pgCameraObject *self, const void *image, int buffer_size,
```
illume, line 180: A size really only makes sense as an unsigned int, because a size can't be negative. But I don't see images getting that big.

src_c/display.c
```diff
@@ -1915,29 +1920,29 @@ pg_set_caption(PyObject *self, PyObject *arg)
...
         SDL_SetWindowTitle(win, title);
-    /* TODO: icon title? */
 
-    PyMem_Free(title);
-    PyMem_Free(icontitle);
+    /* TODO: icon title? */
     Py_RETURN_NONE;
-
-error:
-    PyMem_Free(title);
```
illume, line 1938: I just had to check this was correct... it is.

https://docs.python.org/3/c-api/arg.html
> In general, when a format sets a pointer to a buffer, the buffer is managed by the corresponding Python object, and the buffer shares the lifetime of this object. You won’t have to release any memory yourself. The only exceptions are es, es#, et and et#.

src_c/event.c
```diff
@@ -1804,7 +1800,8 @@ _pg_flush_events(Uint32 type)
 static PyObject *
 pg_event_clear(PyObject *self, PyObject *args, PyObject *kwargs)
 {
-    int loop, len, type;
+    Py_ssize_t len;
```
illume, line 1803: I know it wouldn't because that many events would never be returned... but for this loop, it would theoretically be an issue: `for (loop = 0; loop < len; loop++) {`

Because loop would wrap around into negative when it gets past sizeof(int). But there's never going to be that many events, so this isn't an issue.

src_c/rwobject.c
```diff
@@ -536,7 +536,7 @@ _pg_rw_seek(SDL_RWops *context, Sint64 offset, int whence)
     PyGILState_STATE state;
 
     if (helper->fileno != -1) {
-        return lseek(helper->fileno, offset, whence);
+        return lseek(helper->fileno, (long)offset, whence);
```
illume, line 539: Ah. This is an error. We should be using `lseek64(helper->fileno, (long)offset, whence)`

src_c/base.c
ankith26: yeah, true. `ndim` would usually be from 1 to 3 for most practical purposes, right?

src_c/base.c
ankith26: hehe yeah, compiler was going crazy over a segfaulting function, which for some reason, was *made* to segfault

src_c/bitmask.c
ankith26: This was weird to me too at start. But then I figured out that `BITMASK_W_LEN` uses `sizeof` which, weirdly enough, is not a macro. So you can't really use preprocessor macros with it, and if you use it in regular `if` statement checks, then compiler warns about constant comparison. To make the compiler happy I had to store the value first in a variable and then do regular comparisons with it

src_c/camera_v4l2.c
ankith26: Yeah I remember the compiler crying about stuff here, and if I left `size` as an unsigned int, I'd have to do more changes around the code to make the compiler happy. This was the easier solution, but if it has the be other way, then I'm okay for doing it that way too

src_c/event.c
ankith26: Yeah hehe. But just to know how safe we are here, I just did some quick maphs. The largest int is `2147483647` and size of `pygame.event.Event` is 72 bytes, and that turns out to be around 154 GB, so we are quite safe, no one would get 154 GB of events, the average RAM device would give up much before that

src_c/rwobject.c
ankith26: Oh right, nice catch!

.github/workflows/cppcheck.yml
```diff
@@ -0,0 +1,32 @@
...
+
+  pull_request:
+    branches: main
+    paths:
+      - 'src_c/**'
+
+# TODO: Any more static checkers can be added here
+jobs:
+  run-cppcheck:
+    runs-on: ubuntu-20.04
```
illume, line 18: cppcheck! Very nice.

I wonder though if we should use the latest release? https://github.com/danmar/cppcheck#cmake

The one on ubuntu 20.04 is a bit old.

illume: I guess there's no easy to install package available... so building it from source would take too long. I looked for a pypi package too, and couldn't find one.

Probably sticking to the ubuntu one is fine IMHO.

ankith26: Yeah, it should be fine for now. We can always update this runner to use Ubuntu 22.04 later when it starts rolling to GH actions runners in a few months

ankith26: Wait actually... CPPCheck provides windows installers on their [github releases](https://github.com/danmar/cppcheck/releases).
Ubuntu 20.04 one is still at 1.9.0 while the latest version is 2.6

Perhaps for some extra safety, that could be an additional check too

src_c/camera_v4l2.c
illume: Cool, let's leave it as it is now then.

illume (approved):
👍 TMNT 👀🥷🐢

thanks heaps

src_c/rwobject.c
```diff
@@ -31,6 +31,15 @@
...
 
+#if defined(_WIN32)
+#define PG_LSEEK _lseeki64
+#elif defined(__APPLE__)
+/* Mac does not implement lseek64 */
+#define PG_LSEEK lseek
+#else
+#define PG_LSEEK lseek64
+#endif
+
```
saper, line 42: Proposed https://github.com/pygame/pygame/pull/3898 to improve this on non-glibc platforms.

PR #2922, opened 2021-12-17
Freetype FileNotFound

Starbuck5:
This is a pretty small PR, just some small improvements around the freetype code.

Most importantly, making it return FileNotFound instead of IOError on one of its loading paths.

This allows the changing of the font test too, because ftfont uses those tests.

But we don't have to worry about backwards compatibility, because FileNotFound is a subclass of IOError, so `except`s that caught the old one will catch the new one.


ankith26 (approved):
LGTM! Nice PR and cleanups, also nice it does not break back compat 😉

illume (approved):
👍 thanks. Lgtm too.

PR #2920, opened 2021-12-16
Cython code modernizations

Starbuck5:
Moved all _sdl2 cython to language level 3str, the default in upcoming Cython 3.0. It was an easy modernization to do, and I just did it to remove any possible python 2 related confusions in the future of developing those modules.

I also replaced the functions `pygame._sdl2.audio.get_num_audio_devices` and `pygame._sdl2.audio.get_audio_device_name` with the `pygame._sdl2.audio.get_audio_device_names` function.

To get a list of strings of the audio devices the old way, you had to do:
```py
count = pygame._sdl2.audio.get_num_audio_devices(False)
names = [pygame._sdl2.audio.get_audio_device_name(i, False).decode("utf-8") for i in range(count)]
```

Now you can just do:
```py
names = pygame._sdl2.audio.get_audio_device_names(False)
```

I hope my build time cython PR can get to a working state and be merged, so I didn't include the regenerated cython in this PR.

PR #2925, opened 2021-12-18
Writing fixes, corrected audio device mistake

Starbuck5:
I made a couple mistakes in #2920 about how the new `get_audio_device_names()` function would interact with the _sdl2.audio system. Which I fixed by making everything strings, not bytes, which seems like a decent way to go in late 2021.

I also included the new generated C file this time because I'm still having trouble with the "build time cython" PR.

I also made the docs inclusion test only run on CI, and corrected some stuff in the docs and in math.c.


ankith26 (approved):
👍 nice fixups!

illume:
I wonder where device name is documented to be utf8?

illume:
The pygame compatibility issues here should be listed as bugs.

examples/audiocapture.py
```diff
@@ -26,10 +22,9 @@
 pg.init()
 
 # init_subsystem(INIT_AUDIO)
-names = [get_audio_device_name(x, 1) for x in range(get_num_audio_devices(1))]
+names = get_audio_device_names(True)
```
illume, line 25: The keyword arg here would make it readable. Which makes me wonder if a separate function for capture devices would be better? Because then people would use it rather than supply True/1 without the keyword.

illume (commented):
Thanks.

Misc PRs make it difficult to discuss things and to understand what's changed. If there's an issue with one item it blocks them all. But a misc pr is probably better than no PR... not sure a good way to do these small changes really.

Can you please update the title to summarize what's in here?

examples/audiocapture.py
Starbuck5: I was thinking about that too, but I wanted to keep the PR small, just making sure the audio device system works with the new function from a previous PR.

The new function is a lot nicer because it constructs the list, and it doesn't return byte strings.

Maybe `get_playback_device_names()` and `get_capture_device_names()`? It probably shouldn't be just `get_capture_devices` because that suggests it returns the AudioDevice itself, rather than just the name.

Starbuck5:
> The pygame compatibility issues here should be listed as bugs.

The ones that I updated the docs about can't be fixed, SDL doesn't support the `icontitle` system anymore. And it was never guaranteed to work, I just wanted to clarify that it will never work in the present day.

> I wonder where device name is documented to be utf8?

https://wiki.libsdl.org/SDL_OpenAudioDevice
> device: a UTF-8 string reported by SDL_GetAudioDeviceName() or a driver-specific name as appropriate. NULL requests the most reasonable default device.

illume (approved):
👍 thanks 🎈🌲

brccabral:
Hi @Starbuck5 , I am getting an import error. Any idea?
`ImportError: cannot import name 'get_audio_device_names' from 'pygame._sdl2'`

Starbuck5:
@brccabral 

```py
>>> import pygame
pygame 2.1.2 (SDL 2.0.18, Python 3.9.5)
>>> from pygame._sdl2 import get_audio_device_names
>>> get_audio_device_names
<built-in function get_audio_device_names>
>>> pygame.init()
(5, 0)
>>> get_audio_device_names()
['Speaker (Realtek(R) Audio)', 'Headset (Soundcore Life Tune XR Hands-Free AG Audio)', 'Headphones (Soundcore Life Tune XR Stereo)']
>>>
```

Are you on the latest version of pygame?

brccabral:
I was on 2.0.2, upgrading to 2.1.2 solved the issue. Thank you @Starbuck5

PR #2929, opened 2021-12-19
Add .move_towards() for Vector2 and Vector3

novialriptide:
Fix https://github.com/pygame/pygame/issues/2905


novialriptide:
I noticed that `move_toward` is not grammatically correct, but I think some other people will have a better name than `move_towards`, so I'll leave it for now until there are some better name suggestions.

novialriptide:
Alright, I'm done with the `move_towards()`.
I did some performance testing with the Python version.
```
import pygame
import time

start = time.perf_counter()

def vector2_move_towards(
    origin,
    target,
    speed: float
):
    """Moves towards the target Vector2 by the movement speed.

    Must be put in a loop until its reached its target.

    Parameters:
        origin: The original position
        target: The target position.
        speed: The movement speed.
    
    """
    delta = target - origin
    dist = delta.magnitude()

    if (dist <= speed or dist == 0):
        return target
    
    return origin + delta / dist * speed

vec1 = pygame.Vector2(10, 20)
vec2 = pygame.Vector2(30, 30)
speed = 1
vec3 = vector2_move_towards(vec1, vec2, speed)

print(vec3)

end = time.perf_counter()
print(end - start)
```

Here are the results:
```python
C: 0.00005339999
Python: 0.00011449999874457717
```

Time to work on `move_towards_ip()`!

docs/reST/ref/math.rst
```diff
@@ -195,6 +195,36 @@ Multiple coordinates can be set using slices or swizzling
 
       .. ## Vector2.distance_squared_to ##
 
+   .. method:: move_toward
+
+      | :sl:`returns a vector that moves toward to the given vector at the given speed.`
```
lordmauve, line 200: Everywhere you have written `speed` you mean `distance`. Speed is distance over time and you don't know how often this will be called.

docs/reST/ref/math.rst
```diff
@@ -195,6 +195,36 @@ Multiple coordinates can be set using slices or swizzling
 
       .. ## Vector2.distance_squared_to ##
 
+   .. method:: move_towards
+
+      | :sl:`returns a vector that moves toward to the given vector at the given speed.`
+      | :sg:`move_towards(Vector2, float) -> Vector2`
+
+      Similar to ``lerp()``, this returns a Vector which moves self to the
+      given Vector at a given speed. The first parameter determines the target
```
lordmauve, line 204: Still lots of mentions of speed that should be distance.

novialriptide: Sorry, it looks like my editor didn't catch that.

src_c/math.c
```diff
@@ -1287,6 +1293,89 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+    pgVector *ret;
+    ret = (pgVector *)pgVector_NEW(self->dim);
+    if (ret == NULL) {
+        return NULL;
+    }
+    for (i = 0; i < self->dim; ++i)
+        ret->coords[i] = self->coords[i];
+
+    if (!PyArg_ParseTuple(args, "Od:move_towards", &target, &distance)) {
+        return NULL;
```
ankith26, line 1346: hmmm this can potentially segfault if the first argument passed is not a Vector object, see other functions for how the conversion is done. Also this arg handling bit should come at the top, before the `pgVector_NEW` call, because it should error quickly when args are not right.

src_c/math.c
```diff
@@ -1287,6 +1293,89 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+    }
+    if (!_vector_move_towards_helper(self->dim, ret->coords, target->coords, distance)) {
+        return NULL;
```
ankith26, line 1349: There's a memleak here, `ret` is already allocated, so before a `NULL` return it has to be `Py_DECREF`d

src_c/math.c
```diff
@@ -1287,6 +1293,89 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+static PyObject *
+vector_move_towards_ip(pgVector *self, PyObject *args)
+{
+    Py_ssize_t i;
+    pgVector *target;
+    double distance;
+    pgVector *ret;
+    ret = (pgVector *)pgVector_NEW(self->dim);
+    if (ret == NULL) {
+        return NULL;
```
ankith26, line 1363: we don't need `ret` at all, here, these lines can go. We can always use `self->coords` in place of `ret->coords`

src_c/math.c
```diff
@@ -1287,6 +1293,89 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+    }
+    for (i = 0; i < self->dim; ++i)
+        ret->coords[i] = self->coords[i];
+
+    if (!PyArg_ParseTuple(args, "Od:move_towards_ip", &target, &distance)) {
+        return NULL;
```
ankith26, line 1369: again, this bit should be at the top, with proper argument handling

docs/reST/ref/math.rst
```diff
@@ -195,6 +195,36 @@ Multiple coordinates can be set using slices or swizzling
...
+      self.
+
+      .. versionadded:: 2.1.2
+
+      .. ## Vector2.move_towards ##
+
+   .. method:: move_towards_ip
+
+      | :sl:`returns a vector that moves toward to the given vector at the given distance.`
+      | :sg:`move_towards_ip(Vector2, float) -> Vector2`
```
ankith26, line 216: `_ip` does not return a `Vector2`, it returns `None` and modifies the vector inplace.

The docs below this line need updating too

ankith26 (changes requested):

ankith26 (changes requested):
This also needs unittests and type stubs

novialriptide:
> This also needs unittests and type stubs

Done!

buildconfig/pygame-stubs/math.pyi
```diff
@@ -272,6 +274,8 @@ class Vector3:
     def as_spherical(self) -> Tuple[float, float, float]: ...
     def from_spherical(self, spherical: Tuple[float, float, float]) -> None: ...
     def copy(self) -> Vector3: ...
+    def move_towards(self) -> Vector3: ...
+    def move_towards_ip(self) -> None: ...
```
ankith26, line 278: the function signatures need to indicate the first 'target' argument and the 'distance' argument

docs/reST/ref/math.rst
```diff
@@ -195,6 +195,36 @@ Multiple coordinates can be set using slices or swizzling
 
       .. ## Vector2.distance_squared_to ##
 
+   .. method:: move_towards
+
+      | :sl:`returns a vector that moves toward to the given vector at the given distance.`
+      | :sg:`move_towards(Vector2, float) -> Vector2`
```
ankith26, line 201: Rather than have having args as `Vector2` and `float` which are actually typenames, we need to specify the parameter names here. I'm thinking something like "target" and "distance" will do

docs/reST/ref/math.rst
```diff
@@ -195,6 +195,36 @@ Multiple coordinates can be set using slices or swizzling
...
+
+      Similar to ``lerp()``, this returns a Vector which moves self to the
+      given Vector at a given distance. The first parameter determines the target
+      Vector, while the second parameter determines the distance. The distance value
+      must be greater than 0, but if the distance value is equal to 0, it will return
+      self.
+
+      .. versionadded:: 2.1.2
```
ankith26, line 209: needs update :)

docs/reST/ref/math.rst
```diff
@@ -195,6 +195,36 @@ Multiple coordinates can be set using slices or swizzling
...
+
+      .. ## Vector2.move_towards ##
+
+   .. method:: move_towards_ip
+
+      | :sl:`returns a vector that moves toward to the given vector at the given distance.`
+      | :sg:`move_towards_ip(Vector2, float)`
```
ankith26, line 216: Same review as above, but also additionally needs a `-> None` at the end to show that `None` is returned

docs/reST/ref/math.rst
```diff
@@ -195,6 +195,36 @@ Multiple coordinates can be set using slices or swizzling
...
+
+      Similar to ``lerp()``, this returns a Vector which moves self to the
+      given Vector at a given distance. The first parameter determines the target
+      Vector, while the second parameter determines the distance. The distance value
+      must be greater than 0, but if the distance value is equal to 0, it will return
+      self.
```
ankith26, line 222: the docs def need some updating here, why is there no mention of the inplace update?

src_c/math.c
```diff
@@ -1287,6 +1293,99 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+        return 0;
+    }
+
+    for (i = 0; i < dim; ++i)
+        delta[i] = origin_coords[i] - target_coords[i];
+
+    /* Get magnitude of Vector */
+    dist = sqrt(_scalar_product(delta, delta, dim));
+
+    if (dist <= distance || dist == 0)
```
ankith26, line 1315: would make more logical sense if `target` Vector is returned when `dist <= distance`, but when `dist == 0`, the `origin` vector should be left as is, and returned

src_c/math.c
```diff
@@ -1287,6 +1293,99 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+    Py_ssize_t i;
+    pgVector *target;
+    double distance;
+    pgVector *ret;
+    double self_coords[VECTOR_MAX_SIZE];
+
+    if (!PyArg_ParseTuple(args, "Od:move_towards", &target, &distance))
+        return NULL;
+    
+    if (!PySequence_AsVectorCoords(self, self_coords, self->dim))
```
ankith26, line 1341: the arg handling is still not right 😅 
```suggestion
    if (!PySequence_AsVectorCoords(target, self_coords, self->dim))
```

src_c/math.c
```diff
@@ -1287,6 +1293,99 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+        origin_coords[i] = origin_coords[i] - delta[i] / dist * distance;
+
+    return 1;
+}
+
+static PyObject *
+vector_move_towards(pgVector *self, PyObject *args)
+{
+    Py_ssize_t i;
+    pgVector *target;
```
ankith26, line 1333:
```suggestion
    PyObject *target;
```

src_c/math.c
```diff
@@ -1287,6 +1293,99 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+        return NULL;
+    }
+    return (PyObject *)ret;
+}
+
+static PyObject *
+vector_move_towards_ip(pgVector *self, PyObject *args)
+{
+    Py_ssize_t i;
+    pgVector *target;
```
ankith26, line 1366:
```suggestion
    PyObject *target;
```

src_c/math.c
```diff
@@ -1287,6 +1293,99 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+    double distance;
+    double self_coords[VECTOR_MAX_SIZE];
+    
+    if (!PyArg_ParseTuple(args, "Od:move_towards_ip", &target, &distance))
+        return NULL;
+    
+    if (!PySequence_AsVectorCoords(self, self_coords, self->dim))
```
ankith26, line 1373:
```suggestion
    if (!PySequence_AsVectorCoords(target, target_coords, self->dim))
```

src_c/math.c
```diff
@@ -1287,6 +1293,99 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+    return (PyObject *)ret;
+}
+
+static PyObject *
+vector_move_towards_ip(pgVector *self, PyObject *args)
+{
+    Py_ssize_t i;
+    pgVector *target;
+    double distance;
+    double self_coords[VECTOR_MAX_SIZE];
```
ankith26, line 1368:
```suggestion
    double target_coords[VECTOR_MAX_SIZE];
```

src_c/math.c
```diff
@@ -1287,6 +1293,99 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+        return NULL;
+    
+    if (!PySequence_AsVectorCoords(self, self_coords, self->dim))
+    {
+        PyErr_SetString(PyExc_TypeError, "Argument 1 must be a vector.");
+        return NULL;
+    }
+
+    for (i = 0; i < self->dim; ++i)
+        self->coords[i] = self->coords[i];
```
ankith26, line 1380: why is this even required? lol. Just remove this bit

src_c/math.c
```diff
@@ -1287,6 +1293,99 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+
+    if (!_vector_move_towards_helper(self->dim, self->coords, target->coords, distance))
+        return NULL;
+
+    memcpy(self->coords, self->coords, self->dim * sizeof(self->coords));
```
ankith26, line 1385: even this bit is not required, you cannot have same source and dest pointer for memcpy in C

ankith26 (changes requested):
Lots of changes 😅

docs/reST/ref/math.rst
ankith26: Oh wait NVM, we need to indicate types here because everywhere else in the math module it's done this way

novialriptide:
> Lots of changes 😅 

Whoops, looks like I rushed things lol.

docs/reST/ref/math.rst
novialriptide: What should I change the version to?

src_c/math.c
novialriptide: I get an error if I would change target from a pgVector to PyObject because I'm trying to access it coords as well

src_c/math.c
```diff
@@ -1287,6 +1293,77 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+
+static PyObject *
+vector_move_towards(pgVector *self, PyObject *args)
+{
+    double *target_coords[VECTOR_MAX_SIZE];
+    double distance;
+    pgVector *ret;
+
+    if (!PyArg_ParseTuple(args, "Od:move_towards", &target_coords, &distance))
+        return NULL;
```
ankith26, line 1334:
```c
    PyObject *target;
    pgVector *ret;
    double target_coords[VECTOR_MAX_SIZE];
    double distance;
    pgVector *ret;

    if (!PyArg_ParseTuple(args, "Od:slerp", &target, &distance)) {
        return NULL;
    }

    if (!PySequence_AsVectorCoords(target, target_coords, self->dim)) {
        PyErr_SetString(PyExc_TypeError, "Argument 1 must be a vector.");
        return NULL;
    }
```
The first step is to extract user passed object into a `PyObject *target`. Then, extract contents of that pyobject into C double array `target_coords`, and incase there are errors in this step, a python error is set

src_c/math.c
```diff
@@ -1287,6 +1293,77 @@ vector_scale_to_length(pgVector *self, PyObject *length)
...
+}
+
+static PyObject *
+vector_move_towards_ip(pgVector *self, PyObject *args)
+{
+    Py_ssize_t i;
+    double *target_coords[VECTOR_MAX_SIZE];
+    double distance;
+    
+    if (!PyArg_ParseTuple(args, "Od:move_towards_ip", &target_coords, &distance))
```
ankith26, line 1358: *We still need proper arg handling here* 😅 
See my comment above for code and explanation

ankith26 (changes requested):

src_c/math.c
```diff
@@ -1287,6 +1293,95 @@ vector_scale_to_length(pgVector *self, PyObject *length)
     Py_RETURN_NONE;
 }
 
+static int
+_vector_move_towards_helper(Py_ssize_t dim, double *origin_coords, double *target_coords, double distance)
```
ankith26, line 1297: perhaps this can become a `static void` function. Because all codepaths seem to return `1`, so it is redundant to be checking function output here

ankith26 (commented):
This PR is almost looking good to go! Before merge, you must run `python3 setup.py format` and commit the autoformat changes, and all these commits can be squashed and merged

illume (approved):
🎉 👍 thanks :)

docs/reST/ref/math.rst
```diff
@@ -195,6 +195,35 @@ Multiple coordinates can be set using slices or swizzling
...
+      .. versionadded:: 2.1.3
+
+      .. ## Vector2.move_towards ##
+
+   .. method:: move_towards_ip
+
+      | :sl:`moves the vector toward its target at a given distance.`
+      | :sg:`move_towards_ip(Vector2, float) -> None`
+
+      Noves itself toward the given Vector at a given distance.
```
Starbuck5, line 218: Typo: Moves

docs/reST/ref/math.rst
```diff
@@ -195,6 +195,35 @@ Multiple coordinates can be set using slices or swizzling
 
       .. ## Vector2.distance_squared_to ##
 
+   .. method:: move_towards
+
+      | :sl:`returns a vector moved toward the target by a given distance.`
+      | :sg:`move_towards(Vector2, float) -> Vector2`
+
+      Returns a Vector which moves itself to the given Vector at a given
+      distance.
```
Starbuck5, line 204: I thought we talked about this?
`Returns a Vector which moves itself to the given Vector at a given distance.` is awkward phrasing

`Returns a Vector which is moved towards the given Vector by a given distance`?

Starbuck5:
There's a problem with input parsing

```py
>>> a = pygame.math.Vector2((0,0))
>>> a.move_towards((10,10), 2)
<Vector2(1.41421, 1.41421)>
>>> b = pygame.math.Vector3(0)
>>> b
<Vector3(0, 0, 0)>
>>> b.move_towards(a, 2)
<Vector3(0, 0, 1.42421e-306)>
>>> a.move_towards((1,2,3,4,5,6,7,8), 10)
TypeError: Expected Vector as argument 1
```

If the dimensions aren't equal, it should fail, presumably

And if the input is rejected, it shouldn't say it expected a Vector, since it can take vector-like. Pygame rect functions would call this "Rect-style," so that could be used here as "Vector2-style" or "Vector3-style"


Also I'm putting in a commit that fixes a couple things with the example.
It adds it to the examples listing, and it makes it a size that will work on people's 125% scaled displays (like mine), and it does pygame.quit() properly.

examples/go_over_there.py
```diff
@@ -0,0 +1,72 @@
...
+while running:
+    for event in pg.event.get():
+        if event.type == pg.QUIT:
+            running = False
+
+        if event.type == pg.MOUSEBUTTONUP:
+            target_position = pg.mouse.get_pos()
+
+        if event.type == pg.KEYUP:
+            if event.key == pg.K_r:
```
dr0id, line 58: Suggestion: I like to provide a keyboard exit shortcut too like pressing escape (e.g. as on windows 'alt+F4' to close the program). AFAIK `pg.QUIT` is only triggered when pressing the 'x' button on the window frame so its mouse only.

Resolve at will.

test/math_test.py
```diff
@@ -88,6 +88,100 @@ def testCopy(self):
...
+        origin = Vector2(7.22, 2004.0)
+        target = Vector2(12.30, 2021.0)
+        change_ip = Vector2(7.22, 2004.0)
+
+        change = origin.move_towards(target, 3)
+        change_ip.move_towards_ip(target, 3)
+
+        self.assertEqual(round(change.x, 2), expected.x)
+        self.assertEqual(round(change.y, 2), expected.y)
+        self.assertEqual(round(change_ip.x, 2), expected.x)
```
dr0id, line 102: suggestion: next time maybe consider to separate move_towards and move_towards_ip into two different test cases

test/math_test.py
```diff
@@ -88,6 +88,100 @@ def testCopy(self):
...
+        self.assertEqual(round(change_ip.y, 2), expected.y)
+
+    def test_move_away(self):
+        expected = Vector2(6.36, 2001.13)
+        origin = Vector2(7.22, 2004.0)
+        target = Vector2(12.30, 2021.0)
+        change_ip = Vector2(7.22, 2004.0)
+
+        change = origin.move_towards(target, -3)
```
dr0id, line 139: feels like contradicting: move_towards but actually it moves away with the negative 

resolve at will

dr0id (approved):
I left some suggestions. No need to change them in this PR or at all. Approving anyway.

novialriptide:
> I left some suggestions. No need to change them in this PR or at all. Approving anyway.

Thanks! I'll be looking to resolving these later today.

test/math_test.py
novialriptide: Will do.

ankith26 (approved):
Finally approving this PR 🥳 
There is a CI fail but it's unrelated.
Also, side note to the person merging the PR, squash and merge, and remember to remove co-author tags while doing so, so that me and starbuck don't come as PR authors here :)

test/math_test.py
illume: It's a pretty good point. Even though we've gone over naming this method more than anything else in the story of pygame, and perhaps any method named anywhere ever...

Did we consider "to"? `v.to(v2, -1.1)`.

illume (approved):
👍 thanks for this, and for being such a good sport about all the naming conversations.

Issue #2912, opened 2021-12-14
Discrepancy in c_api/base.c docs


Starbuck5:
The docs: https://www.pygame.org/docs/c_api/base.html

It claims that `pg_TwoFloatsFromObj` returns with an error set, when it doesn't.

I'd really appreciate it if someone were to go through and make sure the python error status is correct for all these functions, especially the numberFROMthing ones, because I see a couple other errors too I think.

PR #2934, opened 2021-12-20
Addressing set error inconsistencies in certain src_c/{base.c, rect.c} functions

anmallie:
Addressing issue #2912. 

I took a thorough look through c_api/base.c and cross referenced with the python C API and as expected, the issue seems to be limited to the functions like `pg_FloatFromObjIndex`, `pg_TwoIntsFromObj`, etc. There are no explicit exceptions raised in any of these, but if I would guess it was thought that `PyTuple_GET_ITEM` and `PySequence_GetItem` raise exceptions like `PyTuple_GetItem` when they do not. 

Regarding the unsigned int varieties, they will not raise an exception unless `PyLong_AsUnsignedLong` is passed values outside of unsigned long range, so I thought it appropriate to document the `OverflowError` from that specifically. While `pg_RGBAFromObj` also uses `PyLong_AsUnsignedLong`, the doc specifically mentions the expected 0-255 restriction so I figured documenting the `OverflowError` was inappropriate. 

Please let me know if there is anything I should change.

Corrections:
- `pg_IntFromObjIndex`: No Python exceptions raised.
- `pg_TwoIntsFromObj`: No Python exceptions raised.
- `pg_FloatFromObjIndex`: No Python exceptions raised.
- `pg_TwoFloatsFromObj`: No Python exceptions raised.
- `pg_UintFromObj`: Documented OverflowError
- `pg_UintFromObjIndex`: Documented OverflowError


illume (approved):
Very nice 👍 Thank you for this.

ankith26 (changes requested):
Thanks for the PR! 🥳 

Unfortunately this issue is harder than it seems and a doc update is not enough to fix stuff. I did some poking around the code the other day, and surely `pg_IntFromObj` does not raise any python errors errors, but the `pg_IntFromObjIndex` *can* raise error when a call to `PySequence_GetItem` fails, but when this function is calling the former function it silences the error. This is/are bug(s) that needs fixing in the source, because these functions have really mixed uses, in some calls to these functions errors are being set at the callers end, and in a few more places there is a plain `NULL` return that assumes this function set the error.

Perhaps these functions even need updating to use `PyNumber_*` API so that a larger variety of objects are handles and taken care at python API level

anmallie:
Hello, thanks for taking a look!
> the pg_IntFromObjIndex _can_ raise error when a call to PySequence_GetItem fails, but when this function is calling the former function it silences the error.


My bad, `PySequence_GetItem` definitely does raise error on failure. I naively assumed that it wouldn't based on Python's API documentation. However, the ___FromObjIndex functions check for `PySequence_GetItem` failure, and so shouldn't the ___FromObj functions not be called and therefore not silence the error? In this case, the original documentation for the ___FromObjIndex functions would be correct.

From c_api/base.c:
```c
static int
pg_FloatFromObjIndex(PyObject *obj, int _index, float *val)
{
    int result = 0;
    PyObject *item = PySequence_GetItem(obj, _index);

    if (item) {
        result = pg_FloatFromObj(item, val);
        Py_DECREF(item);
    }
    return result;
}
```
The other ___FromObjIndex functions are practically identical.

Small test:
```c
#define PY_SSIZE_T_CLEAN
#include <Python/Python.h>

int main() {
  Py_Initialize();
  PyObject *tup = Py_BuildValue("iii", 3, 4, 5);
  PyObject *item = PySequence_GetItem(tup, 3);

  if (item)
    printf("item is not null\n");

  if (PyErr_Occurred())
    PyErr_Print();

  Py_XDECREF(item);
  Py_XDECREF(tup);
  Py_Finalize();
  return 0;
}
```
The `printf` is not reached, and `IndexError` is given. Hopefully it is safe to assume `pg_FloatFromObjIndex` has the same behavior.

As for fixing bugs in the source, are you thinking of appropriately adding and removing explicit error sets at the caller end to bring in line with the true behavior, or is the issue deeper than that? In addition, should the pg_Two___FromObj functions be given explicit error sets to match the docs instead of changing the docs?

Since this is deeper than simply changing documentation, this may be a task better suited for someone with more experience than me, but if my help would still be welcomed I would be glad to dive in.

ankith26:
Yes! Your help is still very much welcome and appreciated! :)

I believe this can be fixed by making `*_FromObjIndex` functions do `PyErr_Clear()` so that they completely adhere to the new docs, likewise fixing other functions to clear errors in all codepaths seems like a good idea. Because usually when these functions are called, the caller sets a custom user friendly error message (and if there are places where caller does not set the error, then there needs to be fixes there too to set an error message)

anmallie:
>  likewise fixing other functions to clear errors in all codepaths seems like a good idea

Are there other functions I should take a look at specifically, or was this just referring to the other `pg_*From*` functions?

I added `PyErr_Clear` to the `*FromObjIndex` functions and took a look at all the calls for these `pg_*From*` functions in the source code, and almost all of them handle failure and set their own error messages as you mentioned. 

The only place I found any issue was in src_c/rect.c which checked for set error with `PyErr_Occurred` before setting its own message, which is unnecessary since the `pg_IntFromObjIndex` call should never set an error with the change.

The only other questionable area is `pgRect_FromObject` which has `pg_IntFromObjIndex` calls returning `NULL` on failure without setting an error, but all calls to this function set their own errors, similar to how the other functions are being handled, so it seems to be fine as is.

Let me know if there is anything else I can do or change, and thanks for the guidance/review!

ankith26:
I did some more poking around the code in this PR, and now for the most part this PR is looking good to go!

Some notes
- `pg_UintFromObj` and `pg_UintFromObjIndex` must not raise any errors too, because the caller seems to be handing those in all cases
- It is documented that `pgRect_FromObject` must set python errors on failure. But looking in the code, in most code paths python errors are not raised, but there are some codepaths where errors can be raised. And most cases, the caller of this function seems to set a custom error, and there are some places (like `_pg_rect_contains`) that pass without custom errors. This functionality was added recently and has not seen a release yet, and is broken in the error case (whoops I reviewed that PR and did not notice this break)
```py
import pygame
"not rect type" in pygame.Rect(20, 40, 60, 80)
```
Fails with 
```
  File "E:\test.py", line 2, in <module>
    "not rect type" in pygame.Rect(20, 40, 60, 80)
SystemError: error return without exception set
```
Rather than a `TypeError` with a descriptive message, so this is definitely a bug that needs fixing
I propose `pgRect_FromObject` must not error on any codepath, and all callers of this function must do error handling on their side.
You could take it up in this PR if you want to because it is related, and if you don't then no worries we can always do it in later PRs :)

anmallie:
Thanks for taking a look!

> `pg_UintFromObj` and `pg_UintFromObjIndex` must not raise any errors too, because the caller seems to be handing those in all cases

Should I clear any `OverflowError` from `PyLong_AsUnsignedLong` calls in `pg_UintFromObj` too then, and make them all consistent for future use?

And I will definitely take a look at `pgRect_FromObj`, it is quite a related issue.

anmallie:
It seems to be failing on this one test:
```
FAIL: test_contains (__main__.RectTypeTest)
  File "test/rect_test.py", line 707, in test_contains
    self.assertTrue(Rect(2, 3, 1, 1) in r, "r does not contain Rect(2, 3, 1, 1) 2")
AssertionError: False is not true : r does not contain Rect(2, 3, 1, 1) 2

Ran 143 tests in 0.010s

FAILED (failures=1)
```
But unless I misunderstand how the tests work (which I very well may) this seems to be an issue that was already present on my end, as this test failure occurs on my untouched branch as well. Perhaps someone else should take over from here on. I am also second guessing the change I proposed to pg_rect_contains_seq (also neglecting the plain int return for numeric values), but it does look like this function may be where the lack of a descriptive error stems from.

illume:
> But unless I misunderstand how the tests work (which I very well may) this seems to be an issue that was already present on my end, as this test failure occurs on my untouched branch as well. Perhaps someone else should take over from here on. I am also second guessing the change I proposed to pg_rect_contains_seq (also neglecting the plain int return for numeric values), but it does look like this function may be where the lack of a descriptive error stems from.

Oh!

This test is passing on our CI, but it is a new change that was merged in very recently... so it's very possible that it crashes on your system and not on our test systems. OR... maybe a merge messed up somehow on your end? Could you please let me know if main branch is passing for you? If it's broken, it would be good to know what your system is so then maybe someone can reproduce it/debug it more easily.

ankith26:
> Should I clear any OverflowError from PyLong_AsUnsignedLong calls in pg_UintFromObj too then, and make them all consistent for future use?

Yes, please. That gives more consistency for the function, otherwise caller would have to handle 2 cases of error/noerror which is more messy

src_c/rect.c
```diff
@@ -980,15 +971,20 @@ pg_rect_contains(pgRectObject *self, PyObject *arg)
     return PyBool_FromLong(ret);
 }
 
-static int
+static PyObject *
```
ankith26, line 974: the return type should not change here, because `pg_rect_contains_seq` is a "slot" function so the expected return type is an integer itself. The `ret` from `_pg_rect_contains` must be returned directly without `PyBool_FromLong`

anmallie:
My bad, that makes sense. I reverted the changes to `pg_rect_contains_seq` and used `PyErr_SetString` for the error instead, since returning the RAISE macro would be returning `(PyObject *) NULL` instead of `-1`.

I also added `PyErr_Clear` to `pg_UintFromObj` to silence any potential `OverflowError` and updated the docs accordingly.

I doubt this will pass any build checks though, I'll be looking into that more soon.

src_c/base.c
```diff
@@ -557,6 +561,10 @@ pg_UintFromObj(PyObject *obj, Uint32 *val)
         }
```
ankith26, line 561: a `PyErr_Clear();` is required here too just before `return 0;` because `PyNumber_Long` sets python error. Actually it would be pretty rare for flow to pass through this codepath because of the initial `PyNumber_Check`, but we still need to clear error for more robust code purposes

src_c/rect.c
```diff
@@ -307,6 +297,7 @@ pgRect_FromObject(PyObject *obj, SDL_Rect *temp)
             PyObject *rectresult = PyObject_CallObject(rectattr, NULL);
```
ankith26, line 297: a few lines above here, `PyObject_GetAttrString` returns `rectattr`, and this can be `NULL` on error with python error set. We need to handle that case by clearing the error and returning `NULL` before doing a `PyCallable_Check` here

ankith26 (commented):
Just a few more bits to do, overall this PR is really nice, especially like that more descriptive typeerror, that makes the error more apparent in a membership test fail

ankith26 (approved):
Thanks a ton for the PR! 🥳 
LGTM now 👍

anmallie:
Thanks so much for the reviews and all the help!

PR #2935, opened 2021-12-21
Fix jiggling problem in pygame.transform.rotate()

nickd4:
### Issue

My son has been working on a game with an overhead-view playfield. He has drawn some characters in overhead view, and what we are trying to do is to call `pygame.transform.rotate()` to make the character face in the correct direction, and then draw onto the screen surface with `Surface.blit()`. We are calculating the (x, y) position of the blit using code similar to this:
```
screen.blit(rotated_image, math.floor(x - rotated_image.get_width() / 2), math.floor(y - rotated_image.get_height() / 2))
```
So basically the idea is to examine the size of the image returned from `pygame.transform.rotate()` and use this to centre the image over the (x, y) which is supposed to represent the current centre of the character.

The problem is that this does not yield a smoothly rotating character. As a test we made the character rotate by 1 degree every game tick, and what we see is the character seems to jiggle slightly to the left/right and up/down at various times in its rotation.

### Cause

Examining the code of `pygame.transform.rotate()` in `src_c/transform.c` I can more or less see the problem: the original code does not seem to have good control over rounding vs truncation and whether the rotation would be performed about the centre of a pixel or the boundary line between pixels.

I must say I do not want to be too critical of the original code, as there could be subtleties of the code which I am not appreciating, but as there are few comments I did not spend too long trying to analyze what it did before. Instead I went with an experimental approach to see if rewriting the code would fix the issue, and it did seem to.

### Proposed fix

I rmodified both the code which determines the size of the rotated image, in the function `surf_rotate()`, and also the inner loop code that generates the pixels of the rotated image, in the function `rotate()`. The modified code is following almost the same procedure as the original but with some slight refinements and more comments.

As I have a slightly different way of thinking about the mathematics than the original author, I renamed some of the variables, for instance `cx` seemed to mean `cos(angle) * x` and I renamed it to `xc` to mean `x * cos(angle)`. I also introduced a number of new variables for clarity (some optimization would be possible but I felt it not worth messing up the code).

On a side note, the previous code used 16-bit fixed point arithmetic in the inner loop for speed, but floating point arithmetic in other calculations such as deciding the width and height of the output image. For consistency I decided to change it to 16-bit fixed point arithmetic always. So if there is a tiny distortion then this will also be reflected in the sizing of the output image. This means that `sangle` changes to `isin` and `cangle` changes to `icos` everywhere. I also sneakily swapped the order of the last 2 parameters to the `rotate()` function to `icos, isin` instead of `sangle, cangle`.

### Half-pixel issues

After thinking carefully about how it needs to work, I came to the conclusion that in order for a rotated image to be able to be blitted onto something interchangeably with the original image, the rotation centre has to be the exact centre of both the original and rotated images.

This implies that if a dimension is odd (e.g. width is 201) then the rotation centre is the centre of the middle pixel (e.g. the pixel with x = 100) whereas if a dimension is even (e.g. with is 200) then it's the boundary between the 2 middle pixels (e.g. the pixels with x = 99 and x = 100).

It's also important that if the rotation centre was at the centre of a pixel (rather than a boundary) in the original image then it needs to be at the centre of a pixel (rather than a boundary) in the rotated image, and vice versa. This applies separately in the horizontal and vertical dimensions.

So it's rather strange when you think about it: suppose I have an image of 201 x 100 and then I rotate it 90 degrees, I'd expect to get a new image of 100 x 201. But this will not be able to maintain a consistent centre when drawn into a playfield, because I cannot draw the 100 x 201 image at a half-pixel offset to have it centred with respect to where the 201 x 100 image used to be. So the right way in my opinion is to return a new 101 x 200 (or 99 x 202 or similar) image for in this example.

Perhaps in the future we could have a way of passing a subpixel offset into the rotate function, or a more advanced rotate function. But this would be too hard to explain to my son, and in any case we are caching the rotated image until the next time the player changes direction, so we don't want to have the subpixel offset of player's current position built into it.

### Discussion

If people think the issue of a 201 x 100 image being rotated 90 degrees not returning a 100 x 201 image is a dealbreaker, then I am open to ideas like a new function name or more advanced API with subpixel level control. But at any rate we need a solution of some sort, because his game was not working well without this fix and now it's much smoother.

Edit: Thinking about it some more, I realize that we probably need a Boolean flag or flags to send into the rotate() function that would force the x or y dimension to be odd or even in the output image. And if the flag were not sent and defaulted to None, then it would size the image in a similar way to the original code. This would also avoid breakage for existing games. Thoughts? Note I haven't implemented this yet, so the PR might not be suitable to merge at the present time.

### Sample animations and code

Here is an animation without the fix: 
https://user-images.githubusercontent.com/21211188/146869618-6309a64b-7934-4111-9be4-3be229fab8d5.mp4

Here is an animation with the fix:
https://user-images.githubusercontent.com/21211188/146869652-c1a06d85-ab45-4faa-bea0-81be530e67be.mp4

The animations were produced with the following utility that I call `make_webp.py` that I whipped up for sample purposes:
```
#!/usr/bin/env python3

import PIL.Image
import PIL.ImageDraw
import pygame
import pygame.image
import pygame.transform
import sys

if len(sys.argv) < 6:
  print(f'usage: {sys.argv[0]:s} in.png deg_(start,stop,step) out_size_(x,y) bg_color_(r,g,b) out.webp')
  sys.exit(1)
in_png = sys.argv[1]
[deg_start, deg_stop, deg_step] = [float(i) for i in sys.argv[2].split(',')]
[out_size_x, out_size_y] = [int(i) for i in sys.argv[3].split(',')]
[bg_color_r, bg_color_g, bg_color_b] = [int(i) for i in sys.argv[4].split(',')]
out_webp = sys.argv[5]

in_image = pygame.image.load(in_png)

frames = []
deg = deg_start
while deg < deg_stop:
  rotated_image = pygame.transform.rotate(in_image, deg)

  out_image = pygame.surface.Surface((out_size_x, out_size_y), depth = 24)
  out_image.fill((bg_color_r, bg_color_g, bg_color_b))
  out_image.blit(
    rotated_image,
    (
      (out_size_x - rotated_image.get_width()) // 2,
      (out_size_y - rotated_image.get_height()) // 2
    )
  )

  frames.append(
    PIL.Image.frombytes(
      'RGB',
      (out_size_x, out_size_y),
      pygame.image.tostring(out_image, 'RGB')
    )
  )

  deg += deg_step

frames[0].save(
  out_webp,
  format = 'webp',
  append_images = frames[1:],
  save_all = True,
  duration = 100,
  loop = 0
)
```

To reproduce, here is the source image drawn by my son:
person

And the command line used to generate the animations was:
```
./make_webp.py person.png 0,360,1 300,300 128,128,128 out.webp
```

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -624,28 +624,38 @@ surf_rotate(PyObject *self, PyObject *args, PyObject *kwargs)
     pgSurfaceObject *surfobj;
     SDL_Surface *surf, *newsurf;
     float angle;
+    PyObject *even_odd = Py_None;
+    int even_odd_x, even_odd_y;
 
     double radangle;
     int icos, isin;
     int x0, y0, xc, yc, xs, ys;
     int nxmax, nymax;
     Uint32 bgcolor;
-    static char *keywords[] = {"surface", "angle", NULL};
+    static char *keywords[] = {"surface", "angle", "even_odd", NULL};
 
-    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O!f", keywords,
-                                     &pgSurface_Type, &surfobj, &angle))
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O!f|O", keywords,
+                                     &pgSurface_Type, &surfobj, &angle,
+                                     &even_odd))
         return NULL;
     surf = pgSurface_AsSurface(surfobj);
     if (surf->w < 1 || surf->h < 1) {
         Py_INCREF(surfobj);
-        return surfobj;
+        return (PyObject *)surfobj;
     }
 
     if (surf->format->BytesPerPixel == 0 || surf->format->BytesPerPixel > 4)
         return RAISE(PyExc_ValueError,
                      "unsupport Surface bit depth for transform");
 
-    if (!(fmod((double)angle, (double)90.0f))) {
+    // check if evenness/oddness tuple (for both dimensions) is given
+    // yes => extract vales for each dimension and use full routine
+    // no  => we can use optimized routine for 90-degree rotations
+    if (even_odd != Py_None) {
+        if (!PyArg_ParseTuple(even_odd, "pp", &even_odd_x, &even_odd_y))
+            return NULL;
+    }
+    else if (!(fmod((double)angle, (double)90.0f))) {
         pgSurface_Lock(surfobj);
 
         Py_BEGIN_ALLOW_THREADS;
@@ -675,29 +685,26 @@ surf_rotate(PyObject *self, PyObject *args, PyObject *kwargs)
     //   (x0, -y0) => (x0 icos + y0 isin, x0 isin - y0 icos)
     // max of all the x coordinates gives right side x relative to centre
     // max of all the y coordinates gives top side y relative to centre
-    // we will then round to integer (this can lose a small amount of image)
+    // then round to a multiple of one-half (can lose a small amount of image)
     xc = (x0 * icos + 1) >> 1; // in 16:16 fixed point
     yc = (y0 * icos + 1) >> 1; // in 16:16 fixed point
     xs = (x0 * isin + 1) >> 1; // in 16:16 fixed point
     ys = (y0 * isin + 1) >> 1; // in 16:16 fixed point
     nxmax = (
-      MAX(MAX(MAX(xc - ys, -xc - ys), -xc + ys), xc + ys) + 0x8000
-    ) >> 16;
+        MAX(MAX(MAX(xc - ys, -xc - ys), -xc + ys), xc + ys) + 0x4000
+    ) >> 15;
     nymax = (
-      MAX(MAX(MAX(xs + yc, -xs + yc), -xs - yc), xs - yc) + 0x8000
-    ) >> 16;
+        MAX(MAX(MAX(xs + yc, -xs + yc), -xs - yc), xs - yc) + 0x4000
+    ) >> 15;
 
-    // we now have (nxmax, nymax) = half the resulting image size, in integer,
-    // to get the full image size we'll double and add the lowest bit of the
-    // original x and y sizes, this is because for odd x size we have to rotate
-    // around the middle of the centre pixel whereas for even x size we have
-    // to rotate around the border between the two centre pixels, similarly y
-    // (by symmetry an even x size can only result in even x size, similarly y)
-    newsurf = newsurf_fromsurf(
-      surf,
-      (nxmax << 1) + (x0 & 1),
-      (nymax << 1) + (y0 & 1)
-    );
+    // now (nxmax, nymax) = half the resulting image size in 31:1 fixed point
+    // (i.e. in units of one-half) or the full resulting image size in integer,
+    // next may need padding to meet the user's evenness/oddness specification
+    if (even_odd != Py_None) {
+        nxmax = ((nxmax + even_odd_x + 1) & ~1) - even_odd_x;
+        nymax = ((nymax + even_odd_y + 1) & ~1) - even_odd_y;
+    }
+    newsurf = newsurf_fromsurf(surf, nxmax, nymax);
     if (!newsurf)
         return NULL;
 
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -267,37 +267,47 @@ rotate90(SDL_Surface *src, int angle)
 }
 
 static void
-rotate(SDL_Surface *src, SDL_Surface *dst, Uint32 bgcolor, double sangle,
-       double cangle)
+rotate(SDL_Surface *src, SDL_Surface *dst, Uint32 bgcolor, int icos, int isin)
 {
-    int x, y, dx, dy;
+    int x0, y0, x1, y1, x2, y2, tx, ty, dx, dy, i, j;
 
     Uint8 *srcpix = (Uint8 *)src->pixels;
     Uint8 *dstrow = (Uint8 *)dst->pixels;
     int srcpitch = src->pitch;
     int dstpitch = dst->pitch;
 
-    int cy = dst->h / 2;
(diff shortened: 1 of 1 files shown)


MyreMylar:
Just want to quickly say that this is an impressive dive straight into pygame's C code base for a first time contributor 😃 

I expect it will take a while to review, largely because it is an existing widely used function - but partly because it is the Christmas holidays.

Though we are not generally opposed to changing small-ish visual things in pygame that have been slightly broken for a long time if the end-result is better. If I understand the change correctly from a skim read, then I expect a lot of users use sprite images that happen to rotate well with the current code.

I've kicked off the existing CI tests to run for now.

nickd4:
hi MyreMylar and thanks for the kind words. Well I work as a professional developer and I spend a good part of my day writing C code and github PR's so hopefully some of that translates to an ability to contribute to open source projects. Sadly I cannot work on open source as much as I would like, due to lack of time, you can thank my son and his game for the contribution :)

### Improvement

I have modified things now so that the `pygame.transform.rotate()` function now takes a new (keyword or positional) parameter called `even_odd`, which can be passed as either `None` or a tuple of 2 Booleans such as `(True, True)`. What this does is to control whether the output image will be even (`False') or odd (`True') in the x and y dimensions respectively. So it will now compute the output image size slightly more accurately (relative to the previous fix) such that it should be more compatible with existing pygame games. And then it will add up to 1 pixel of padding in each dimension to meet the user's specification if any.

### Evaluation

I have not done a detailed comparison of the old behaviour vs the new behaviour. I believe the new behaviour, with the `even_odd` parameter passed as `None`, is a slight improvement on the old behaviour, due to better control of the output image size (uses rounding) and the subpixel issues (samples in the centre of each pixel in both original and output images). You might be able to see this by comparing the original movie file from the last comment (before the fix) with the below one.

I do know that when I use the `even_odd` parameter as intended (I pass `(True, True)` because my son's original image is odd in both dimensions and I want the output image to be the same) then I get a smooth rotation, as I can centre it consistently.

### Sample animations and code

Here is an animation with the fix, and `even_odd` passed as `None`, i.e. compatible with original pygame:
https://user-images.githubusercontent.com/21211188/147011227-927bfc0b-9f53-4593-95a7-d52ef1974277.mp4

Here is an animation with the fix, and `even_odd` passed as `(True, True)`, i.e. the smoothly rotated and centred version:
https://user-images.githubusercontent.com/21211188/147011266-1f4b5a5c-e211-4dc8-b4f7-a32c953501c6.mp4

They are produced with this updated version of `make_webp.py`:
```
#!/usr/bin/env python3

import PIL.Image
import PIL.ImageDraw
import pygame
import pygame.image
import pygame.transform
import sys

if len(sys.argv) < 6:
  print(f'usage: {sys.argv[0]:s} in.png deg_(start,stop,step) out_size_(x,y) bg_color_(r,g,b) out.webp [even_odd_(x,y)]')
  sys.exit(1)
in_png = sys.argv[1]
[deg_start, deg_stop, deg_step] = [float(i) for i in sys.argv[2].split(',')]
[out_size_x, out_size_y] = [int(i) for i in sys.argv[3].split(',')]
[bg_color_r, bg_color_g, bg_color_b] = [int(i) for i in sys.argv[4].split(',')]
out_webp = sys.argv[5]
even_odd = None
if len(sys.argv) >= 7:
  [even_odd_x, even_odd_y] = [i == 'odd' for i in sys.argv[6].split(',')]
  even_odd = (even_odd_x, even_odd_y)

in_image = pygame.image.load(in_png)

frames = []
deg = deg_start
while deg < deg_stop:
  # note: we can pass even_odd as None in the fixed version of pygame (the same
  # as not passing it), but we cannot pass it for the original pygame, so test
  rotated_image = (
    pygame.transform.rotate(in_image, deg)
  if even_odd is None else 
    pygame.transform.rotate(in_image, deg, even_odd)
  )
  #print(rotated_image.get_width(), rotated_image.get_height())

  out_image = pygame.surface.Surface((out_size_x, out_size_y), depth = 24)
  out_image.fill((bg_color_r, bg_color_g, bg_color_b))
  out_image.blit(
    rotated_image,
    (
      (out_size_x - rotated_image.get_width()) // 2,
      (out_size_y - rotated_image.get_height()) // 2
    )
  )

  frames.append(
    PIL.Image.frombytes(
      'RGB',
      (out_size_x, out_size_y),
      pygame.image.tostring(out_image, 'RGB')
    )
  )

  deg += deg_step

frames[0].save(
  out_webp,
  format = 'webp',
  append_images = frames[1:],
  save_all = True,
  duration = 100,
  loop = 0
)
```

The command lines were:
```
./make_webp.py person.png 0,360,1 300,300 128,128,128 new_none.webp
./make_webp.py person.png 0,360,1 300,300 128,128,128 new_odd_odd.webp odd,odd
```

MyreMylar:
Hello,

I've had a quick look at this and noticed a small oddity in your initial test case that I wanted to ask you about. In it you specify the blitting position for the rotated sprite with this code:

```
((out_size_x - rotated_image.get_width()) // 2,  (out_size_y - rotated_image.get_height()) // 2)
```

Whereas probably the more standard pygame-y way to get a centre'd position for the sprite would be to do something like:

```
rotated_rect = rotated_image.get_rect()
rotated_rect.center = (out_size_x // 2, out_size_y // 2)
blit_pos = rotated_rect.topleft
```

If I swap in the pygame-y way - using a rect centre to position the sprite - into your original test case then the jittering goes away and it looks (to my eyes at least) like the version produced with the changed C code. The positions produced are different between the two ways of centering the sprite.

I'm not sure if this is just the way the pygame Rect class calculates the centre point synching up neatly with the way rotate works or something else wrong with the maths somewhere.

MyreMylar:
I'm going to close this for now as I don't think these changes are necessary - but I'm not 100%. If anybody else wants to look into it then feel free to investigate and re-open.

PR #2937, opened 2021-12-21
Allow software rendering on _sdl2 Windows

Starbuck5:
This adds `get_surface()` and `flip()` functions to the `Window` class, finally allowing people to use multiple software rendered windows at the same time.

Before, it was possible to have 1 software rendered window controlled by a `Window`, if you created it with display first and then used `Window.from_display_module()`. But you couldn't do "normal pygame stuff" with Surfaces and `pygame.draw`, etc, with multiple Windows.

I've thought of this as the "one big missing feature" for the `Window` class for several months now. So hopefully this gets it closer to a full release :smile:

- I was unable to add update() since I wanted to put things in base.c with the get/set default window functions, and base.c can't access the rect c api, I believe.
- It should be doable to have a Window created when there is no pygame.display SDL_Window set the default window. That way Window could actually replace set_mode() in a meaningful way, with convert() and other functions around the codebase recognizing the default window.
- Lots of state should be centralized / moved around in display and base? Display has a display state struct, which doesn't include the SCALED specific info, and the actual window and surface are stored statically in base.

Oh! And I also changed the default window title to match pygame.display.


illume (changes requested):
Looking good.

Can we have some tests? Seems like Window is going for stabalization so for it we should start adding tests.

Also using the methods inside some example/ would be nice.

Starbuck5:
> Can we have some tests? Seems like Window is going for stabalization so for it we should start adding tests.

I've put in a very small amount of testing. But it's a start!

> Also using the methods inside some example/ would be nice.

Good idea, I put it in the `video` example.

illume:
Looks like there are some incompatibilities with linux and opacity.

```
ERROR: test_window_opacity (pygame.tests.video_test.VideoModuleTest)

  File "/opt/python/cp36-cp36m/lib/python3.6/site-packages/pygame/tests/video_test.py", line 40, in test_window_opacity
    window.opacity = 0.4
  File "src_c\cython\pygame\_sdl2\video.pyx", line 513, in pygame._sdl2.video.Window.opacity.__set__
pygame._sdl2.sdl2.error: That operation is not supported
```


In the ubuntu 18.04 (old SDL build) there's an additional error:
```
ERROR: test_window_software_render (pygame.tests.video_test.VideoModuleTest)

The above exception was the direct cause of the following exception:

  File "/home/runner/.local/lib/python3.6/site-packages/pygame/tests/video_test.py", line 33, in test_window_software_render
    window.flip()
SystemError: <built-in method flip of pygame._sdl2.video.Window object at 0x7f7181e55348> returned a result with an error set
```

ankith26:
This is a fairly large PR, I don't think we can resolve this in any way in time for the 2.1.3 release, pushing this to 2.1.4 milestone

PR #2938, opened 2021-12-21
Fix pygame.version.SDL and pygame.version.vernum

MightyJosip:
On current version of pygame.version.SDL (pun intended 😅) you get the error if you write 
`print(pygame.version.SDL)`

`NameError: name 'major' is not defined`

This should fix it


MyreMylar (approved):
LGTM 👍

PR #2943, opened 2021-12-22
audio: Use null audio driver instead of disk for tests

illume:
Hopefully faster.
Hopefully can stop using the word dummy in another commit.

Note, that this only works with SDL2, which is the reason it wasn't done before.


illume:
Not sure if this is actually faster... need to test that before merging. Can not see by comparing run times of github actions... because they vary by several minutes.

illume:
This seems slower. So... meh.

PR #2944, opened 2021-12-22
freetype: Fix intermittent crash with test_font_file_not_found

illume:
On mac homebrew this was crashing inside freetype.
Not sure why, so made a hackaround of using SDL first.

To reproduce: 
```bash
python3 test/ftfont_test.py -k test_font_file_not_found
```

### DIFF
--- a/src_c/freetype/ft_wrap.c
+++ b/src_c/freetype/ft_wrap.c
@@ -417,6 +417,20 @@ _PGFT_TryLoadFont_Filename(FreeTypeInstance *ft, pgFontObject *fontobj,
     char *filename_alloc;
     size_t file_len;
 
+    /* There seems to be an intermittent crash with opening
+       a missing file and freetype 2.11.1 on mac homebrew.
+       python3 test/ftfont_test.py -k test_font_file_not_found
+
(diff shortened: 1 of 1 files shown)


Starbuck5:
Freetype has two separate loading paths, one through the Freetype library and one through SDL_RWops. These paths are not both tried, one of them, RWops I think, I’d only tried on Windows, Freetype loading is only tried on not Windows.

I wanted to have them both use the SDL way, but I was trying to keep the last PR very simple.

But anyways, getting it all on rwops through the pygame C api would be a better solution if it worked. Would also get consistent error messages with other things.

illume:
@Starbuck5 can you recommend a way forward?

It sounds like future work could fix this properly... but this PR could fix the error in the meantime. If so, we could keep this PR until the future rework is done?

Starbuck5:
It depends on how eager you are to release 2.1.1 full. If it’s going to be this weekend, I can come up and test an alternate patch by then. But if you want to release sooner, then you should merge in this as a stopgap.

illume:
I'd like to release, so will merge this one in... and then there's no rush to get the other font changes in.

Issue #2851, opened 2021-11-17
Logos page not working anymore


Grimmys:
The logos page is currently not working: https://www.pygame.org/docs/logos.html

The reason why is that the file (logos.html) is not in the generated folder because it's not part of the Sphinx generation: it's a static page.

Solution could be to move this "static" file to the generated folder (this would mean to add the folder in git I guess), or alternatively to let this page follow the same model as others and be in RST format to let Sphinx generate it.

I'm going to try fixing this issue, and I will try first to switch to the RST format because I like clean solutions.

ankith26:
Related to this issue, there are some other static pages apart from the logos pages that have been missed out too, like the license link for `LGPL.txt`, and potentially others

Grimmys:
Thanks @ankith26 for the additional data, I will try to check these things this week. 👍

PR #2945, opened 2021-12-23
Add lowres_pygame_powered.png

novialriptide:


Starbuck5:
This shouldn't just be added as a file, it needs to be added to the pygame logos page.

But the pygame logos page doesn't work anymore, and Grimmys might be rewriting it in rst. See #2851.

And talk to Grimmys on discord about it, probably.

novialriptide:
This PR is going to be merged with @Grimmys 's website fix PR

PR #2952, opened 2021-12-26
clean modinit code, fix bugs, use recommended API

### LINKED ISSUE #2951: Clean up some module initialization code
Port our uses of `PyModule_GetDict` and `PyDict_SetItemString` at the end of the modules to `PyModule_AddObject`.

Mixer, event, surface, rect, all use the older style, certainly more do as well. Font looks like it does it correctly, for an example of good behavior.

Additionally, we could investigate using macros to "port" new python C api functions introduced in 3.9 and 3.10 that could have further convenience for the code. PyModule_AddObjectRef and PyModule_AddType, from this page: https://docs.python.org/3/c-api/module.html.

ankith26:
PR to close #2951

Does some code cleanups around modinit code, fixing a few bugs in the process, and migrating away from `PyModule_GetDict` which is discouraged by the [docs](https://docs.python.org/3/c-api/module.html#c.PyModule_GetDict)

I originally planned to make use of newer API like `PyModule_AddObjectRef` (new in py3.10) and `PyModule_AddType` (new in py3.9) because that makes the modinit code even more elegant, but to do this we'd have to put some hacks on `pyver < 3.10` and I figured this could create problems if we ever intend to use `Py_LIMITED_API`, so I left this for now (but we can still do this when the minimum supported python version is 3.10, and this would happen in 5-6ish years)

### DIFF
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -2205,10 +2205,9 @@ MODINIT_DEFINE(_freetype)
     FREETYPE_MOD_STATE(module)->cache_size = 0;
     FREETYPE_MOD_STATE(module)->resolution = PGFT_DEFAULT_RESOLUTION;
 
-    Py_INCREF((PyObject *)&pgFont_Type);
-    if (PyModule_AddObject(module, FONT_TYPE_NAME, (PyObject *)&pgFont_Type) ==
-        -1) {
-        Py_DECREF((PyObject *)&pgFont_Type);
+    Py_INCREF(&pgFont_Type);
+    if (PyModule_AddObject(module, FONT_TYPE_NAME, (PyObject *)&pgFont_Type)) {
+        Py_DECREF(&pgFont_Type);
         Py_DECREF(module);
         return NULL;
     }
@@ -2239,13 +2238,8 @@ MODINIT_DEFINE(_freetype)
     c_api[1] = &pgFont_New;
 
     apiobj = encapsulate_api(c_api, "freetype");
-    if (!apiobj) {
-        Py_DECREF(module);
-        return NULL;
-    }
-
-    if (PyModule_AddObject(module, PYGAMEAPI_LOCAL_ENTRY, apiobj) == -1) {
-        Py_DECREF(apiobj);
+    if (PyModule_AddObject(module, PYGAMEAPI_LOCAL_ENTRY, apiobj)) {
+        Py_XDECREF(apiobj);
         Py_DECREF(module);
         return NULL;
     }
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -2069,11 +2069,9 @@ static PyMethodDef _base_methods[] = {
 
 MODINIT_DEFINE(base)
 {
-    static int is_loaded = 0;
-    PyObject *module, *dict, *apiobj;
-    PyObject *atexit_register = NULL;
+    PyObject *module, *apiobj, *atexit;
+    PyObject *atexit_register;
     PyObject *pgExc_SDLError;
-    int ecode;
     static void *c_api[PYGAMEAPI_BASE_NUMSLOTS];
 
     static struct PyModuleDef _module = {PyModuleDef_HEAD_INIT,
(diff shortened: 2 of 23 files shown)


src_c/_freetype.c
```diff
@@ -2205,10 +2205,10 @@ MODINIT_DEFINE(_freetype)
     FREETYPE_MOD_STATE(module)->cache_size = 0;
     FREETYPE_MOD_STATE(module)->resolution = PGFT_DEFAULT_RESOLUTION;
 
-    Py_INCREF((PyObject *)&pgFont_Type);
+    Py_INCREF(&pgFont_Type);
     if (PyModule_AddObject(module, FONT_TYPE_NAME, (PyObject *)&pgFont_Type) ==
         -1) {
```
Starbuck5, line 2210: This should just test the value directly, rather than comparing to `-1`, to be consistent with the other ones. Especially since `-1` on failure is not documented.

ankith26: oh, nice catch! :) I made this change in many other places but forgot to do it here

Starbuck5:
Looking at this code, it looks like there should be a function in base that encapsulates the c api for a module into an object, and adds that object to the module.

It looks like a lot of this takes out error handling for the return of encapsulate_api macro. Which goes into PyCapsule_New, which has error reporting stuff. https://docs.python.org/3/c-api/capsule.html#c.PyCapsule_New

ankith26:
> Looking at this code, it looks like there should be a function in base that encapsulates the c api for a module into an object, and adds that object to the module.

hmmm, yeah maybe that could be done here

> It looks like a lot of this takes out error handling for the return of encapsulate_api macro. 

Yeah that's an intentional change. I only removed that bit because `PyModule_AddObject` simply returns `NULL` when `obj` is `NULL`, so I unified the error handling of `encapsulate_api` and `PyModule_AddObject` to get rid of some code repetition

ankith26:
Ok, that did not go so well, there were issues with that approach, a function cannot be used because `encapsulate_api` is a macro and uses macro string stuff, and there seem to be no elegant ways to make the entire block of code a macro, because that would make error handling more painful, so I guess I'll leave that for now, maybe we can figure something in future PRs ;)

Starbuck5:
> I only removed that bit because PyModule_AddObject simply returns NULL when obj is NULL, so I unified the error handling of encapsulate_api and PyModule_AddObject to get rid of some code repetition

Ah, smart.

src_c/_freetype.c
```diff
@@ -2205,10 +2205,9 @@ MODINIT_DEFINE(_freetype)
     FREETYPE_MOD_STATE(module)->cache_size = 0;
     FREETYPE_MOD_STATE(module)->resolution = PGFT_DEFAULT_RESOLUTION;
 
-    Py_INCREF((PyObject *)&pgFont_Type);
-    if (PyModule_AddObject(module, FONT_TYPE_NAME, (PyObject *)&pgFont_Type) ==
-        -1) {
-        Py_DECREF((PyObject *)&pgFont_Type);
+    Py_INCREF(&pgFont_Type);
```
Starbuck5, line 2208: `(PyObject *)`?

src_c/base.c
```diff
@@ -2173,62 +2153,45 @@ MODINIT_DEFINE(base)
...
 #ifdef MS_WIN32
-        SDL_RegisterApp("pygame", 0, GetModuleHandle(NULL));
-#endif
-#if defined(macintosh)
-#if (!defined(__MWERKS__) && !TARGET_API_MAC_CARBON)
-        SDL_InitQuickDraw(&pg_qd);
-#endif
+    SDL_RegisterApp("pygame", 0, GetModuleHandle(NULL));
+#elif defined(macintosh) && !defined(__MWERKS__) && !TARGET_API_MAC_CARBON
+    SDL_InitQuickDraw(&pg_qd);
```
Starbuck5, line 2188: QuickDraw is obsolete, so this can be removed.

src_c/joystick.c
```diff
@@ -671,16 +670,20 @@ MODINIT_DEFINE(joystick)
     }
     joy_instance_map = PyObject_GetAttrString(module, "_joy_instance_map");
     Py_DECREF(module);
```
Starbuck5, line 672: This is kinda strange, do you understand why it is there? I don't.

src_c/_freetype.c
ankith26: Yes this is intentional. Because the incref/decref API are macros, they cast the argument to `(PyObject *)` already, so doing it before the macro call is redundant

src_c/base.c
ankith26: aight, good idea

src_c/joystick.c
ankith26: It's for joystick events to have the old style ID attribute, IIRC (which is deprecated now)

Starbuck5: I mean the `Py_DECREF(module)`

(I thought Github would highlight my selection)

ankith26: A few lines above `module = PyImport_ImportModule("pygame.event");` returned a new reference to `module`, and after we have extracted the dict attribute from the event module, we have to decref it.

Perhaps you got confused thinking the joystick module object is being freed, which is not

Starbuck5 (approved):
Alrighty. Thanks Ankith!

I've read through the whole thing and it looks good.

I also ran the test suite locally and it's still good there.

src_c/base.c
```diff
@@ -2086,59 +2084,41 @@ MODINIT_DEFINE(base)
                                          NULL,
                                          NULL};
 
-    if (!is_loaded) {
```
illume, line 2089: Why was the !is_loaded guard removed?

illume (approved):
Looks ok to me.

I didn't go through each call in detail, didn't look for new warnings, and didn't test locally.

src_c/base.c
ankith26: Because that was a local variable, lol, and was always set to 0 when the init function is called. So the conditional was useless because it will get executed in all cases.

PR #2953, opened 2021-12-26
Add some more events

ankith26:
Adds some new SDL events to pygame, also adds `precise_x` and `precise_y` fields to `MOUSEWHEEL` events, this was implemented in SDL in 2.0.18


illume (approved):
Nice cleanups :)

- Some of these new events could use more docs.
- Also, it would be good if each event could be used within an example/. 

If not in this PR, then an issue should be made for each of these items.

CAPTAIN1947:
I would like to make an example with the android events. I think there is not any example related to android there. And I have a question to @ankith26 are the events APP_DIDENTERBACKGROUND, etc only relevant to android or we can get those events in some other cases?

ankith26:
> And I have a question to @ankith26 are the events APP_DIDENTERBACKGROUND, etc only relevant to android or we can get those events in some other cases?

Nope, these are mobile specific events so they will only work on Android and iOS (SDL supports this). But pygame4ios is not a thing yet, so yeah these events are only on android for now

CAPTAIN1947:
Oh. Thank you.

buildconfig/pygame-stubs/constants.pyi
```diff
@@ -20,6 +20,12 @@ ANYFORMAT: int
 APPACTIVE: int
 APPFOCUSMOUSE: int
 APPINPUTFOCUS: int
+APP_DIDENTERBACKGROUND: int
+APP_DIDENTERFOREGROUND: int
+APP_LOWMEMORY: int
+APP_TERMINATING: int
+APP_WILLENTERBACKGROUND: int
+APP_WILLENTERFOREGROUND: int
```
Starbuck5, line 28: The previous `APP` events don't have underscores between `APP` and `TYPE`. Readability vs consistency?

The previous ones also don't seem to be documented anywhere.

src_py/locals.py
```diff
@@ -83,6 +83,13 @@
     "BUTTON_WHEELUP",
     "BUTTON_X1",
     "BUTTON_X2",
+    "APP_TERMINATING",
+    "APP_LOWMEMORY",
+    "APP_WILLENTERBACKGROUND",
+    "APP_DIDENTERBACKGROUND",
+    "APP_WILLENTERFOREGROUND",
+    "APP_DIDENTERFOREGROUND",
+    "CLIPBOARDUPDATE",
```
Starbuck5, line 92: It looks like these are mostly alphabetical, so that should be preserved I guess?

Starbuck5:
As Illume said, it would be good to document the new events a bit more. I didn't find anything on the SDL wiki, but the source header is very informative on the events I checked: https://github.com/libsdl-org/SDL/blob/main/include/SDL_events.h.

buildconfig/pygame-stubs/constants.pyi
ankith26: The new ones have underscore because the respective SDL constants have them. not sure whats up with the older `APP` constants

src_py/locals.py
ankith26: true

buildconfig/pygame-stubs/constants.pyi
illume: What's the consistent way here? We should probably change them to that.

ankith26: So far, pygame has always followed SDL constants directly, usually only stripping the initial `SDL_` before exporting an SDL constant. SDL initially did not use `_`s in the event constant names, but looks like nowadays, in newer constants they add, they seem to use `_`s more often. The easiest way for pygame would be to stick to SDL event names for consistency, also IMHO the names with underscores are more readable, but older constants added previously do not make use of them

illume: Hrmm, I can't find `SDL_APPACTIVE` inside SDL (nor SDL_APPINPUTFOCUS... didn't look for the others). What am I making a mistake with?

illume: Also tried SDL_APP_ACTIVE, and SDL_APP_INPUTFOCUS

ankith26: They are defined for compatibility with SDL1 `ACTIVEEVENT`, the enum is defined like 
```c
typedef enum {
    SDL_APPFOCUSMOUSE,
    SDL_APPINPUTFOCUS,
    SDL_APPACTIVE
} PygameAppCode;
```
These are not events, but values that `event.state` can take when event is of type `ACTIVEEVENT`

MyreMylar:
Looks like[ this page](https://github.com/libsdl-org/SDL/blob/120c76c84bbce4c1bfed4e9eb74e10678bd83120/src/test/SDL_test_common.c) of SDL tests is also conveniently informative on what attributes events have.

Looks like the new events being added are (with some notes by me):
```
APP_DIDENTERBACKGROUND
APP_DIDENTERFOREGROUND
APP_LOWMEMORY
APP_TERMINATING
APP_WILLENTERBACKGROUND
APP_WILLENTERFOREGROUND
CLIPBOARDUPDATE - Add docs: The clipboard changed
CONTROLLERSENSORUPDATE
CONTROLLERTOUCHPADDOWN
CONTROLLERTOUCHPADMOTION
CONTROLLERTOUCHPADUP
KEYMAPCHANGED no attributes, add docs from SDL: "Keymap changed due to a system event such as an input language or keyboard layout change."
KSCAN_AC_BACK - I assume this is just the android back button?
LOCALECHANGED - usually used in concert with `SDL_GetPreferredLocales` (I don't think pygame supports this yet)
RENDER_DEVICE_RESET
RENDER_TARGETS_RESET
WINDOWDISPLAYCHANGED
WINDOWICCPROFCHANGED
```

I guess we need a few notes for many of them in the docs. I think any examples/example updates should probably be in a separate PR - but an issue should be created to update examples to use the new events (where it makes sense to do so)

Starbuck5:
Its worth pointing out that `CLIPBOARDUPDATE` will likely only fire on clipboard updates SDL picks up, which would only be text clipboard events? So this could be misleading for pygame.scrap users, since we haven't quite given up on image support yet.

ankith26:
I added docs for some events in the latest commit, but I intentionally left out documenting some stuff like `RENDER_DEVICE_RESET` and `RENDER_TARGETS_RESET` because these are `_sdl2` related. 
Also I undocumented an old feature about pygame 1.9.2 sending userevent for file drops on mac, this seems to be broken from a long time, and from what I can tell, this is quite unused and redundant (SDL2 has dropfile event that works on all platforms). There's probably no need to fix this to maintain backwards compatibility, because if people widely used this feature, we would have bug reports flowing in by now

MyreMylar (approved):

PR #2954, opened 2021-12-26
fix segfault in display.update

ankith26:
A regression of the latest 2.1.1 release, display.update can segfault on some kinds of inputs


Starbuck5:
some kinds of inputs = list of rects, also incorrect input will segfault it

This means the aliens example segfaults in 2.1.1

regression from https://github.com/pygame/pygame/commit/17298e430e8639fd483ad3eed1c05bf8614bb186#diff-e6e90037fae974fd4cb565f9dcde4d003062eb172fb5c4132e2b2d994c29d8ad

PR #2955, opened 2021-12-26
Quick fix to get 'python -m pygame.docs' working

ankith26:
Nothing much to explain here, just another quickfix


illume (approved):
Great :)

PR #2958, opened 2021-12-27
Add test for pygame.locals

ankith26:
While doing #2953 I realised that it is quite easy for someone to forget to update `pygame.locals` when they add new constants into `pygame.constants`
So I wrote a quick test to catch this issue, and also fixed a missing constant that someone in the past forgot to add


MyreMylar (approved):
LGTM. A simple test that should prevent these constants getting out of synch in the future.

Starbuck5 (approved):
Good idea!

PR #2959, opened 2021-12-27
Update the image doc, delineate bytes and str

Starbuck5:
This PR was started to improve the image documentation, on a lark, but it led to more change than I expected.

`pygame.image` has `tostring` and `fromstring` methods, but they don't actually take strings, they take bytes. So I updated the function descriptions, and added a note about Python 2 strings. I also changed the error messages from the source to reflect this, and changed the typehints, which (wrongly) typed it as string.

In the future, we could deprecate this. Pillow actually deprecated their tostring/fromstring for tobytes/frombytes about 6 years ago. But that's out of the scope of this PR.

The other doc changes in this PR are mainly rephrasing things to be clearer and marking functions and classes so Sphinx can link them directly.


ankith26 (approved):
LGTM👍

docs/reST/ref/image.rst
```diff
@@ -81,18 +82,19 @@ following formats.
...
-   pygame may not always be built to support all image formats. At minimum it
-   will support uncompressed ``BMP``. If ``pygame.image.get_extended()``
-   returns 'True', you should be able to load most images (including PNG, JPG
+   Pygame may not always be built to support all image formats. At minimum it
+   will support uncompressed ``BMP``. If :func:`pygame.image.get_extended()`
+   returns ``True``, you should be able to load most images (including PNG, JPG
    and GIF).
 
-   You should use ``os.path.join()`` for compatibility.
+   You should use :func:`os.path.join()` for compatibility.
```
MyreMylar, line 97: Should we recommend pathlib here instead?

`path = Path("data", "images", "player.png")`

Motivations:

 - Pathlib is available since Python 3.4 and we support 3.6+ in pygame.
 - https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/
 - https://towardsdatascience.com/dont-use-python-os-library-any-more-when-pathlib-can-do-141fefb6bdb5
 - https://betterprogramming.pub/should-you-be-using-pathlib-6f3a0fddec7e

I feel like generally the python community is leaning toward pathlib being the better library for path stuff now.

docs/reST/ref/image.rst
```diff
@@ -184,27 +187,34 @@ following formats.
...
+             but they deal in bytes. In Python 2, str = bytes, so this is
+             "from string" as in a Python 2 string.
 
    .. ## pygame.image.fromstring ##
 
 .. function:: frombuffer
 
    | :sl:`create a new Surface that shares data inside a bytes buffer`
-   | :sg:`frombuffer(bytes, size, format) -> Surface`
+   | :sg:`frombufferbuffer, size, format) -> Surface`
```
MyreMylar, line 212: This looks like a typo.

src_c/doc/image_doc.h
```diff
@@ -4,9 +4,9 @@
...
 #define DOC_PYGAMEIMAGEGETSDLIMAGEVERSION "get_sdl_image_version() -> None\nget_sdl_image_version() -> (major, minor, patch)\nget version number of the SDL_Image library being used"
 #define DOC_PYGAMEIMAGEGETEXTENDED "get_extended() -> bool\ntest if extended image formats can be loaded"
-#define DOC_PYGAMEIMAGETOSTRING "tostring(Surface, format, flipped=False) -> string\ntransfer image to string buffer"
-#define DOC_PYGAMEIMAGEFROMSTRING "fromstring(string, size, format, flipped=False) -> Surface\ncreate new Surface from a string buffer"
-#define DOC_PYGAMEIMAGEFROMBUFFER "frombuffer(bytes, size, format) -> Surface\ncreate a new Surface that shares data inside a bytes buffer"
+#define DOC_PYGAMEIMAGETOSTRING "tostring(Surface, format, flipped=False) -> bytes\ntransfer image to string buffer"
+#define DOC_PYGAMEIMAGEFROMSTRING "fromstring(bytes, size, format, flipped=False) -> Surface\ncreate new Surface from a string buffer"
+#define DOC_PYGAMEIMAGEFROMBUFFER "frombufferbuffer, size, format) -> Surface\ncreate a new Surface that shares data inside a bytes buffer"
```
MyreMylar, line 9: again this looks like something has gone wrong in the 'frombufferbuffer' part.

MyreMylar (changes requested):
Just spotted a couple of small things while reading through,

docs/reST/ref/image.rst
Starbuck5: I would recommend using neither (for basic usage) If you do “path/to/file.extension” it works everywhere.

I sometimes see pygame beginners where every load image is a bazillion characters long because they do a bunch of path stuff in every single one.

docs/reST/ref/image.rst
Starbuck5: Thanks

docs/reST/ref/image.rst
MyreMylar: I mean really, all files should be in python packages so they can be loaded from memory blobs as well as actual file systems. ..

I'm not bothered if it is left as is, just thought I should mention it since I saw one of these `pathlib` is better than `os.path` articles the other day.

Starbuck5: Let's keep as it is for now, just because I'm not sure what to replace it with.

docs/reST/ref/image.rst
```diff
@@ -15,10 +15,11 @@ Note that there is no Image class; an image is loaded as a Surface object. The
 Surface class allows manipulation (drawing lines, setting pixels, capturing
 regions, etc.).
 
-The image module is a required dependency of pygame, but it only optionally
-supports any extended file formats. By default it can only load uncompressed
-``BMP`` images. When built with full image support, the ``pygame.image.load()``
-function can support the following formats.
+In the vast majority of installations, including those from pip, pygame is
+built to support extended formats with SDL2_Image. However, installations are
```
illume, line 19: Nice improvement mentioning that most installations will have extended formats supported.

IMHO "SDL2_Image" is jargon for most readers, and "pip" is noise here.

docs/reST/ref/image.rst
```diff
@@ -81,18 +82,19 @@ following formats.
...
-   For alpha transparency, like in .png images, use the ``convert_alpha()``
-   method after loading so that the image has per pixel transparency.
+   For alpha transparency, like in .png images, use the
+   :func:`pygame.Surface.convert_alpha()` method after loading so that the
+   image has per pixel transparency.
 
-   pygame may not always be built to support all image formats. At minimum it
-   will support uncompressed ``BMP``. If ``pygame.image.get_extended()``
-   returns 'True', you should be able to load most images (including PNG, JPG
+   Pygame may not always be built to support all image formats. At minimum it
```
illume, line 92: Always, always use lower case pygame.

This section does seem to be a repeat. Remove the whole sentence "pygame may not always be built to support all image formats."?

illume (approved):
Looks like a good improvement to me. Thanks :)

Left some comments, if you feel to address or not is up to you!

docs/reST/ref/image.rst
Starbuck5: I reworded it a bit.

docs/reST/ref/image.rst
Starbuck5: I think "the start of a sentence" > normal pygame casing rules.

It is a repeat, but it's relevant information for the function, and it links to another relevant function, so I'm inclined to leave it.

MyreMylar (approved):
LGTM 👍

PR #2961, opened 2021-12-28
fix some bugs in camera.py

### LINKED ISSUE #2936: pygame.camera members not defined before init?
I tried to call pygame.camera.list_cameras() and it said 

`AttributeError: module 'pygame.camera' has no attribute 'list_cameras'`

When it should have complained that I hadn't initialized the camera module yet, which was the real problem.

This bug is 100% something I caused, but I'm putting up an issue in case someone wants to take a crack at it before I do. Code in src_py/camera.py

ankith26:
Initially set out to fix #2936 then noticed a few more issues, got sidetracked as usual and now I make this PR with a few more fixes 

Now all functions are defined at all times regardless init state, and error correctly. After a quit, they are reset back to error. Make config envvar accept both `vidcapture` and `videocapture` (case insensitive) and fix `get_backends` returning `VidCapture` when it should be `VideoCapture`, and some more code cleanups around that. Also errors with a more helpful message if a particular backend is unavailable

### DIFF
--- a/src_py/camera.py
+++ b/src_py/camera.py
@@ -1,82 +1,132 @@
 import os
-import sys
 import platform
+import sys
 import warnings
 from abc import ABC, abstractmethod
 
-_is_init = 0
+from pygame import error
+
+_is_init = False
 
 
-def _setup_opencv_mac():
-    global list_cameras, Camera, colorspace
+class AbstractCamera(ABC):
+    @abstractmethod
+    def __init__(self, device=0, size=(640, 480), mode="RGB"):
+        """ """
 
-    from pygame import _camera_opencv
+    @abstractmethod
+    def start(self):
+        """ """
 
-    try:
-        from pygame import _camera
-    except ImportError:
-        _camera = None
+    @abstractmethod
+    def stop(self):
+        """ """
 
-    list_cameras = _camera_opencv.list_cameras_darwin
-    Camera = _camera_opencv.CameraMac
-    if _camera:
-        colorspace = _camera.colorspace
+    @abstractmethod
+    def get_controls(self):
+        """ """
+
+    @abstractmethod
+    def set_controls(self, **kwargs):
+        """ """
+
+    @abstractmethod
+    def get_size(self, dest_surf=None):
+        """ """
+
+    @abstractmethod
+    def query_image(self, dest_surf=None):
+        """ """
+
+    @abstractmethod
+    def get_image(self, dest_surf=None):
+        """ """
+
+    @abstractmethod
+    def get_raw(self):
+        """ """
 
 
-def _setup_opencv():
-    global list_cameras, Camera, colorspace
+def _pre_init_placeholder(*_, **__):
+    if not _is_init:
+        raise error("pygame.camera is not initialized")
 
-    from pygame import _camera_opencv
-
-    try:
-        from pygame import _camera
-    except ImportError:
-        _camera = None
-
-    list_cameras = _camera_opencv.list_cameras
-    Camera = _camera_opencv.Camera
-    if _camera:
-        colorspace = _camera.colorspace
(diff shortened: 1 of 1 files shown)


Starbuck5:
Thanks Ankith, but yeesh, this diff is something else. This changes 110 lines of a <200 line file.

I don't want to expand environment variables for vidcapture, because that thing is 100% deprecated. In fact, in my last camera PR, illume told me I could remove it, but I kept it, in a role reversal for the ages. Actually, vidcapture could use the deprecation warning. The biggest problem for me is that I've **never** been able to get it working, even though there are third party python 3 wheels (vidcapture is py2) that supposedly work.

Starbuck5:
Ok, I've read through everything, it looks good.

I'm curious why you refactored all of my individual setup functions into one, though.

ankith26:
> I'm curious why you refactored all of my individual setup functions into one, though.

Those functions had some unnecessary repetition, and this one function conditional approach seems more elegant to me than having functions kept in a dict and each of those functions having repetition of  `global` statement and `_camera` import 😄 

> Maybe colorspace should have a different error message if it isn't able to be imported. Which it could be, if pygame was built without the _camera module. I don't think it's built on Windows 7.

A `NotImplementedError` is the most pythonic error to be using in a case like this, it was made for a case like this.

> I don't want to expand environment variables for vidcapture, because that thing is 100% deprecated. In fact, in my last camera PR, illume told me I could remove it, but I kept it, in a role reversal for the ages.

Ah, ok. Should I revert my change then? Well it does not matter either way IG, this vidcapture changes should remain undocumented anyways

Another thing I wanted to ask, does windows `_camera` work on win8? Because the code does `int(platform.win32_ver()[0]) > 8` check, should it be `>=`?

Starbuck5:
It should work on Windows 8, since every function I used has been around since then, but I’m still not sure. It claims to support it in the docs, so we should try to use it in the code. If it doesn’t work, somebody will open an issue. (Or nobody will care, since not many people use pygame.camera and not many people use Windows 8)

ankith26:
Alrighty :)
I also updated the `colorspace` function to emit better errors when not available, although yeah, it should not get triggered in most cases

src_py/camera.py
```diff
@@ -1,82 +1,132 @@
...
+    if not _is_init:
+        raise error("pygame.camera is not initialized")
 
-    list_cameras = _camera_opencv.list_cameras
-    Camera = _camera_opencv.Camera
-    if _camera:
-        colorspace = _camera.colorspace
+    # camera was init, and yet functions are not monkey patched. This should
+    # not happen
+    raise NotImplementedError()
```
Starbuck5, line 56: is this supposed to be here?

src_py/camera.py
```diff
@@ -1,82 +1,132 @@
...
 
+_is_init = False
 
-def _setup_opencv_mac():
-    global list_cameras, Camera, colorspace
 
-    from pygame import _camera_opencv
+class AbstractCamera(ABC):
+    @abstractmethod
+    def __init__(self, device=0, size=(320, 200), mode="RGB"):
```
Starbuck5, line 14: default size is (640, 480)

src_py/camera.py
ankith26: Yeah, just for extra safety, IG. I can't think of a case where this codepath will actually be executed

src_py/camera.py
ankith26: ooh, nice catch!

Starbuck5 (approved):
Thanks, Ankith!

This looks good to me. You could fix that tiny remaining thing (the default size is wrong on the class), and the it'll be good.

illume (approved):
👍 🎉 thanks

PR #2965, opened 2021-12-30
Fix Pygame imported class typehints

### LINKED ISSUE #2809: Autoimported classes have different types in type hints

MightyJosip:
Fix #2809
Continued from #2811 (that PR had much more work, I will add stuff from there in other PR, this one will just be foused on fixing the reported issue)

### DIFF
--- a/buildconfig/pygame-stubs/__init__.pyi
+++ b/buildconfig/pygame-stubs/__init__.pyi
@@ -38,13 +38,26 @@ from pygame import scrap as scrap
 from ._common import _AnyPath
 
 # These classes are auto imported with pygame, so I put their declaration here
-class Rect(rect.Rect): ...
-class Surface(surface.Surface): ...
-class Color(color.Color): ...
-class PixelArray(pixelarray.PixelArray): ...
-class Vector2(math.Vector2): ...
-class Vector3(math.Vector3): ...
-class Cursor(cursors.Cursor): ...
+from pygame.rect import Rect
+from pygame.surface import Surface
+from pygame.color import Color
+from pygame.pixelarray import PixelArray
+from pygame.math import Vector2, Vector3
+from pygame.cursors import Cursor
+from pygame.bufferproxy import BufferProxy
+from pygame.mask import Mask
+
+__all__ = [
+    "Rect",
+    "Surface",
+    "Color",
+    "PixelArray",
+    "Vector2",
+    "Vector3",
+    "Cursor",
+    "BufferProxy",
+    "Mask",
+]
 
 def init() -> Tuple[int, int]: ...
 def quit() -> None: ...


buildconfig/pygame-stubs/__init__.pyi
```diff
@@ -38,13 +38,26 @@ from pygame import scrap as scrap
...
+    "Rect",
+    "Surface",
+    "Color",
+    "PixelArray",
+    "Vector2",
+    "Vector3",
+    "Cursor",
+    "BufferProxy",
+    "Mask",
+]
```
ankith26, line 60: this should list all other modules and functions defined in this module too

ankith26 (approved):
Nice PR! 👍 🎉 
Like we discussed some time back, there is one more little addition to make here, but it's not super critical, even without it this PR works great

buildconfig/pygame-stubs/__init__.pyi
Zeelot: yeah usually a good pattern to use is to keep `__init__.py` files clear of any functions and simply use them to import and expose things through `__all__`. might be nice to move things around a little in some future PRs if y'all agree. I usually put all my classes/functions in private modules (`_foo.py`) and then import and expose them using relative imports in my `__init__.py` file.

```python
from ._foo import SomeFoo, OtherFoo

__all__ = ["SomeFoo", "OtherFoo"]
```

This just makes it easier to control the public API so that no one ever sees the random libraries you happen to depend on and you can shuffle files around without changing the public API.

ankith26: I don't quite get what you are trying to say, do you mean we should move stuff from init into a separate file and then import stuff from init? I'm not exactly sure of the advantages here, because either way it's almost the same thing, these functions have to be still exported from `__init__.py`. But feel free to open a PR with your suggested changes after this PR is merged in, that also makes it easier for us to see and understand the changes :)

MyreMylar (approved):
LGTM. Hopefully this will please all the linters.

MyreMylar:
This just needs appveyor kickstarting again. The last build failed due to the sdl website being down for a bit, or something like that.

Starbuck5:
Appveyor restarted

PR #2968, opened 2022-01-02
Add optional parameter 'consider_alpha' to the transform.average_colo…

dr0id:
…r() method to remove the black artifacts if alpha is present.

I wanted to blur an image using pygame but didn't find any good way until I saw the average_color method. I know, pygame is not built for such tasks and there are better options out there. Nonetheless I tried. Unfortunately the average_color does not take into account alpha and generates dark artifacts (see image). I found the explanation on [stackoverflow](https://stackoverflow.com/a/35481968) and also coded some methods around the limitation of average_color. But it would be a lot simpler if average_color would consider alpha, hence the pull request.

Here a visual reference:

Explanation

Here the average_color is used like a 'kernel' to do a very simple 'box blur' like blur of the image using following code (or similar):

```
    def box_blur_per_pixel5(image, box):
        bw, bh = box
        pygame_transform_average_color = pygame.transform.average_color
        destination = image.copy()
        with pygame.PixelArray(destination) as w_array:
            for x, col in enumerate(w_array[:]):
                col[:] = [pygame_transform_average_color(image, (x, y, bw, bh), True) for y, c in enumerate(col)]
        return destination

```

Note the 'True' argument in average_color call.


MyreMylar:
So far all my blurring has been done via `transform.scale()` scaling an image down then `transform.smoothscale()` scaling it back up again. I'll be interested to see how this method compares!

dr0id:
@MyreMylar 
Well, it will be much slower than just scaling the image 2 times, but the quality might be better. It is slower because every pixel needs to be calculated. I think there are ways to optimize such a blur function if it is implemented in a own function. But as I already pointed out, there are other libraries that do that much better and faster.

illume (approved):
Looking good to me, thanks :)

Could you please add changes to transform.pyi as well?

dr0id:
@illume thanks for the review and thanks for the hint. Added the change to the pyi file. Didn't know that this file exists and what it is for.

buildconfig/pygame-stubs/transform.pyi
```diff
@@ -28,7 +28,7 @@ def average_surfaces(
     dest_surface: Optional[Surface] = None,
     palette_colors: Union[bool, int] = 1,
 ) -> Surface: ...
-def average_color(surface: Surface, rect: Optional[_RectValue] = None) -> Color: ...
+def average_color(surface: Surface, rect: Optional[_RectValue] = None, consider_alpha: Optional[bool] = False) -> Color: ...
```
ankith26, line 31:
```suggestion
def average_color(surface: Surface, rect: Optional[_RectValue] = None, consider_alpha: bool = False) -> Color: ...
```
According to the [typing docs](https://docs.python.org/3/library/typing.html#typing.Optional) for the `Optional` typehint
> Optional type.
> `Optional[X]` is equivalent to `X | None` (or `Union[X, None]`).
> Note that this is not the same concept as an optional argument, which is one that has a default. An optional argument with a default does not require the `Optional` qualifier on its type annotation just because it is optional.

So here the argument is optional, but cannot take `None`. So `Optional` must not be used here

ankith26 (changes requested):

buildconfig/pygame-stubs/transform.pyi
dr0id: Thanks for the review. Must have overlooked this in the documentation but that is the way to learn things. I haven't written much type hints so any hints are welcome.

PR #2969, opened 2022-01-03
Added Rect typing

novialriptide:
Fix: https://github.com/pygame/pygame/issues/2681


Starbuck5:
Typing might want to be something like this instead: https://stackoverflow.com/questions/49427944/typehints-for-sized-iterable-in-python

ankith26 (commented):
While `Rect` is indeed iterable, the C code does not actually implement the `__iter__` method as quickly demonstrated by this example. 
```
>>> a = pygame.Rect(0, 0, 0, 0)
>>> a.__iter__()
AttributeError: 'pygame.Rect' object has no attribute '__iter__'. Did you mean: '__str__'?
```
However, it does implement a subset of the `Sequence` protocol, including a `__len__` method, which does need typehints here. An `Iterable` would implement an `__iter__` method, but a `Sequence` only needs `__getitem__` and python puts a mixin for `__iter__` that uses this under the hood. Perhaps the `Rect` typestub class should inherit from the `Sequence` ABC?

illume (approved):
🎉 👍

ankith26:
So I did more digging, and posting some notes of my findings here
1) Objects that provide `__getitem__` dunder (examples in pygame include `Rect`, `Color`, `Vector` and possibly more things) are practically `Iterable` but they don't actually confirm to `collections.abc.Iterable` if they don't provide an `__iter__` method as also highlighted by the [docs for the same](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable).
2) This PR does not fix the issue at hand. `mypy` fails to recognise these pygame objects as `Iterable` because they don't confirm to `collections.abc.Iterable` and lack an `__iter__` method. The only proper way to fix this is to implement an `__iter__` function in C code for these objects, and also typehint it. I know some people are against making these objects `Iterable`, but the fact is that these are already mostly iterable. By putting an `__iter__` method we will be explicitly declaring these as proper `Iterable`s. We cannot make it not iterate and break lots of compatibility. Doing something like `list(colour_object)` or `x, y, w, h = rect_obj` is fairly commonplace, and all these would break if we changed this

We can still merge this PR because adding `__len__` typehint to `Rect` is still a good addition, the iter fix can happen in future PR(s) because it involves changes to C code, tests, docs and types for many objects

illume:
I took this "We can still merge this PR" as an approval :)

Thankses.

PR #2973, opened 2022-01-07
Make sure surfaces start as opaque black. (MacOS)

### LINKED ISSUE #1395: Not all surfaces are going to full black
It seems some surfaces are not going to all black.

- When I run all the tests locally, I get some pixel errors in `test_circle_zero_radius`.
- When I run that single test by itself, it passes.
- When I add a `surf.fill((0,0,0))` as below the test passes.

```
        surf = pygame.Surface((200, 200))
        circle_color = pygame.Color('red')
        surf_color = pygame.Color('black')
        surf.fill((0, 0, 0))
```

I'm not sure if it's related to the set_shifts issue mentioned here: https://github.com/pygame/pygame/issues/1383

- Could it be that surfaces are not getting wiped to zero properly now? Maybe some memory is being reused or something? 
- Could it be related to a SDL 2.0.10 change with their new SDL_SIMDAlloc(), and SDL_SIMDFree?

Starbuck5:
Fixes #1395

Thanks to MyreMylar, LordMauve, Illume, and MightyJosip for helping research this, pin it down, and suggest fixes.

This is one take at a fix, it could be definitely be done differently.

**TODO**: get rid of unnecessary fills in test suite put in to avoid this issue.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -711,6 +711,18 @@ surface_init(pgSurfaceObject *self, PyObject *args, PyObject *kwds)
     if (!(flags & PGS_SRCALPHA)) {
         /* We ignore the error if any. */
         SDL_SetSurfaceBlendMode(surface, SDL_BLENDMODE_NONE);
+
+        /* When the display format has a full alpha channel (MacOS right now),
+         * Surfaces may be created with an unreqested alpha channel, which
+         * could cause issues.
+         * pygame Surfaces are supposed to be (0, 0, 0, 255) by default.
+         * This is a simple fix to fill it with (0, 0, 0, 255) if necessary.
+         * See Github issue: https://github.com/pygame/pygame/issues/1395
+         */
+        if (Amask != 0) {
+            SDL_FillRect(surface, NULL,
+                         SDL_MapRGBA(surface->format, 0, 0, 0, 255));
+        }
     }
 
     if (SDL_ISPIXELFORMAT_INDEXED(surface->format->format)) {


Starbuck5:
MyreMylar convinced me to go for another fix, where it instead guarantees not having an alpha channel.

Which would look like this:
```c
        SDL_PixelFormat *pix = NULL;
        if (depth && pgSurface_Check(depth))
            pix = ((pgSurfaceObject *)depth)->surf->format;
        else if (pg_GetDefaultWindowSurface()) {
            pix = pgSurface_AsSurface(pg_GetDefaultWindowSurface())->format;

            /* When the display format has an alpha channel (MacOS right now),
            * Surfaces may be created with an unreqested alpha channel, which
            * could cause issues.
            * pygame Surfaces are supposed to be (0, 0, 0, 255) by default.
            * This tells it to ignore the display format and use the default.
            * See Github issue: https://github.com/pygame/pygame/issues/1395
            */
            if (pix->Amask != 0) {
                pix = NULL;
            }
        }

        if (pix == NULL) {
            pix = &default_format;
            pix->BitsPerPixel = 32;
            pix->Amask = 0;
            pix->Rmask = 0xFF0000;
            pix->Gmask = 0xFF00;
            pix->Bmask = 0xFF;
        }
        bpp = pix->BitsPerPixel;
``` 

But that causes this failure:
```
FAIL: test_solarwolf_rle_usage_2 (pygame.tests.surface_test.SurfaceTypeTest)
Test for RLE status after setting alpha
  File "/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/pygame/tests/surface_test.py", line 362, in test_solarwolf_rle_usage_2
    self.assertTrue(image.get_flags() & pygame.RLEACCEL)
AssertionError: 0 is not true
```

Which I don't even begin to understand

Starbuck5:
Ok, I want to merge this now.

I've tried a couple other solutions and I always get that RLE error, which doesn't make sense.

But the tests are updated.

This isn't 100% ideal but it's better than the situation right now.

Starbuck5:
I force pushed a `setup.py format` into the previous commit.

src_c/surface.c
```diff
@@ -711,6 +711,18 @@ surface_init(pgSurfaceObject *self, PyObject *args, PyObject *kwds)
...
         SDL_SetSurfaceBlendMode(surface, SDL_BLENDMODE_NONE);
+
+        /* When the display format has a full alpha channel (MacOS right now),
+         * Surfaces may be created with an unreqested alpha channel, which
+         * could cause issues.
+         * pygame Surfaces are supposed to be (0, 0, 0, 255) by default.
+         * This is a simple fix to fill it with (0, 0, 0, 255) if necessary.
+         * See Github issue: https://github.com/pygame/pygame/issues/1395
+         */
+        if (Amask != 0) {
```
illume, line 722: I wonder if we can avoid this on other platforms that don't need it?

illume: I expect it's not a performance issue on small surfaces, because the memory is probably already in cache. But for larger ones this could be a significant slow-down.

Starbuck5: It is avoided on platforms that don't need it. The only time it fires is if Amask is some number, but SRCALPHA wasn't requested. That only happens if the preferred display pixel format has alpha, which we've determined only happens on Mac.

illume: Cool. Thanks for the explanation.

illume:
The CI failure is from the new black.

illume (approved):
🎉 👍 thanks


(there's a CI failure because of the new black)

mariospapaz:
This is the closest issue I could find related with MacOS and transparency. 
Before I post an issue I want to be sure if this is related.

I am very confident that MacOS had (and has again) an issue related with Alpha values 
(Apple thought having alpha 0 by default is a good idea) 
 and it was fixed by @Starbuck5 when it was discussed in the community server long ago.

I can't remember which dev version it was fixed but there is a small chance it broke again with the new `Ventura` update.
 
Examples from my game which has lots of transparency work 
Screenshot 2022-11-12 at 12 40 55 AM
Screenshot 2022-11-12 at 9 48 03 AM

PR #2977, opened 2022-01-08
Fix logos page

Grimmys:
This PR should fix issue #2851.

`logos.html` page has been rewrited in RST format with introduction of @novialriptide low res pygame logo.
Reference to `LGPL.txt` file in index is also fixed.

Please tell me if I missed some other unavailable pages, @ankith26 you were the one spotting the issue with LGPL, do you see something else?


ankith26:
Nice PR! I did some local testing and the logos page is fixed now! However, seems like the LGPL.txt file is not copied to the `generated` folder, and another thing, `setup.py docs` gives this warning
` D:\gitstuff\pygame\docs\reST\logos.rst: WARNING: document isn't included in any toctree`

Grimmys:
Yes for the LGPL the URL is an hash... You can get it by clicking on the link from the index page. 
I understand that it could be annoying if you want direct access, we can look for another way to integrate it. 

Thanks for recalling me this warning, I saw it at the beginning but forgot to fix it!

Grimmys:
Okay thanks Ankith for you input, really great remarks.

I indeed forgot to check if `legacy_logos.zip` was correctly included and it was not. I could have done the same approach I did for `LGPL` and reference it with `:download:` role but apparently this wasn't a good idea because you were not able to find the ref to the file this way, so I just **moved both files to the `_static` folder** (RIP `:download:`, thought it was a good idea 😢).

Has you mentioned, we could imagine **better splitting of this folder** since it's holding a lot of different files as of now.
`_static` could become `images` for images and `static` for other static files (archives and txt files for example).

I also fixed the missing reference to the `logos` page by **adding a new link to it from the index page**, tell me if it seems good for you.

illume:
Will this keep the old LGPL.txt file URLs working? `https://www.pygame.org/docs/LGPL.txt`

Grimmys:
No, it would be `https://www.pygame.org/docs/_static/LGPL.txt` I guess, but the old one was already not working (at least it is not working right now).

Some images URL would change too because they have been move to the `_static` folder too to be all at the same place, but they were referenced only on this `logos` page that was also not working before.

ankith26:
> Will this keep the old LGPL.txt file URLs working?

No, but fixing that is out of the scope of this PR, IG. Perhaps on the server side some symlinks could be added to link the needed stuff from `docs/_static` to `docs`?

Starbuck5:
We could put the generated folder itself on Github? Just chuck it in there?

Grimmys:
I'm not fan of versionning a generated folder on git... That's not how things should be done.

Replacing `_static/docs` by `docs` is a solution but we will have to take care about name conflict (it would be ok for the file explorer to have two files with the same name in `_static/docs` and `docs` folder but it would not be ok for the generation). 

I will try to find another solution.

Starbuck5:
> I'm not fan of versionning a generated folder on git... That's not how things should be done.

That's correct. I mean just the folder itself, with the one file that needs to be there.

Grimmys:
Oh this way, sorry, that's slightly better but still not a so good solution... It would be a bit tricky and we may have to put a comment somewhere to clarify this for other contributors.

But still it seems to be the solution if we don't want to have this `_static` directory in the URL.

So please I would just like to have a **better understanding of the context** because it is not clear for me:

What is the exact problem with the current state of my PR? Why having this `_static` reference is a problem?
Is it a problem for the images too or only for `legacy_logos.zip` and `LGPL.txt` files?

(We can have the discussion on discord if it's simpler for you)

illume:
> No, but fixing that is out of the scope of this PR, IG. Perhaps on the server side some symlinks could be added to link the needed stuff from docs/_static to docs?

I'd prefer if the files that were in docs like LGPL.txt were as before. As well as fixing the website, it is nicer for someone to do a `cd docs/` and see files there, rather than in a _static folder.

Grimmys:
So if it's only a matter of seeing the file from the repository, maybe it could be at both places.
What do you think @illume?

dr0id:
> So if it's only a matter of seeing the file from the repository, maybe it could be at both places. What do you think @illume?

I do not think that it is a good idea to copy/paste the same file to two different location at the servers side, even if the deployment is scripted. It requires more space (maybe negligible) but people in the future will wonder why they are copied to two different places.

I wonder if it would be possible to leave the files where they are and reference them there from the docs?

dr0id:
@illume  I wonder where the logos page is currently because
neither of the following urls do resolve (all give a 404):
https://www.pygame.org/docs/logos.html
https://www.pygame.org/logos.html
https://www.pygame.org/docs/generated/logos.html

dr0id:
Just got an idea: leave a docs/logos.html that redirects or has a link/hint to the new logos page.

Grimmys:
> > So if it's only a matter of seeing the file from the repository, maybe it could be at both places. What do you think @illume?
> 
> I do not think that it is a good idea to copy/paste the same file to two different location at the servers side, even if the deployment is scripted. It requires more space (maybe negligible) but people in the future will wonder why they are copied to two different places.
> 
> I wonder if it would be possible to leave the files where they are and reference them there from the docs?

I totally agree with you, it would be best to not have any duplication but I'm afraid that either way we have to make some concessions.
It is not possible to reference them from where they were before, it is a Sphinx restriction: static files must be in the `_static` folder.



> @illume I wonder where the logos page is currently because neither of the following urls do resolve (all give a 404): https://www.pygame.org/docs/logos.html https://www.pygame.org/logos.html https://www.pygame.org/docs/generated/logos.html

That's the whole point of this PR: fix urls that are not accessible online as of today.



> Just got an idea: leave a docs/logos.html that redirects or has a link/hint to the new logos page.

Why not go this way indeed!

dr0id:
> It is not possible to reference them from where they were before, it is a Sphinx restriction: static files must be in the `_static` folder.

hmm, if the deployment script ensures that those files are found with a certain url then my guess would be that it should be possible to reference that url, no? Hmm but then the image is probably not shown...

... maybe this might help:
[display external image in sphinx](https://stackoverflow.com/questions/26396163/how-can-i-display-externally-hosted-images-in-my-sphinx-restructured-text-docu)

dr0id (commented):
Unless there is really the requirement that certain files cannot be moved I would approve this changes.

Grimmys:
> > It is not possible to reference them from where they were before, it is a Sphinx restriction: static files must be in the `_static` folder.
> 
> hmm, if the deployment script ensures that those files are found with a certain url then my guess would be that it should be possible to reference that url, no? Hmm but then the image is probably not shown...
> 
> ... maybe this might help: [display external image in sphinx](https://stackoverflow.com/questions/26396163/how-can-i-display-externally-hosted-images-in-my-sphinx-restructured-text-docu)

That's an alternative solution, change deployment script to not include only generated folder.
But I think it would be cleaner to have the whole "generated docs content" at the same place.

(also displaying an external image is not an issue as your link is proves ; the issue is that this external files have to be hosted somewhere, so it's part of the deployment script solution)

MyreMylar:
Could we use `html_extra_path` to solve this instead of `_static`?

See:
https://stackoverflow.com/questions/49967691/sphinx-link-to-an-internal-file-declared-as-html-extra-path

Grimmys:
Oh right, that's a good idea. Wasn't aware of this option, only about `html_static_path` that is basically doing the same but in worse...

But what do we want to have instead of `_static`? Should we totally remove this `_static` folder or not? 

Can somebody, please, sum up me what should be done in order to be able to merge this, because I'm a bit lost now.

MyreMylar:
I think in summary ideally we want:

-  LGPL.txt to remain at its existing position in the repo so its in an obvious place (docs/)
-  Personally I'm not _that_ fussed about the position of the various logo image files - but if they don't need to move they can probably stay where they are. It's kind of nice to have them in docs I guess but less useful that the license text and readme file?
-  the logos page to work online. I think its fine to get rid of logos.html and have that be generated from an .rst as you are doing now, I don't think anyone browses the docs/ subdirectory to see some raw html to display some images that they could actually see by clicking on them.

Grimmys:
Got it, so I can put back the logos when they were if it is what you prefer.

For LGPL, that a bit more complex.
If I move it back to docs/, it would not be available on web version (current issue).

Grimmys:
Ok so I succeed to have LGPL.txt file at the docs root directory and let Sphinx add it to the generated folder by adding its reference in the `html_extra_path` parameter, thanks @MyreMylar!

@illume is there still something blocking you on this PR?

dr0id:
not sure why this shows up in the file changes: 
image

Other than that it looks good to me.

Grimmys:
Good catch, I moved the file only one folder backward not two...

Now it should be fine.

dr0id (commented):
There is a merge conflict to be resolved. Can't approve until then.

Grimmys:
Merge conflict resolved.

Tbh I don't know what was this conflict, I didn't change anything about `add_stubs` statement but it was telling I added it... So I just deleted it, hope it's what should have be done.

dr0id (approved):
Looks good to me.

ankith26:
I tested this PR locally, and while it's almost good to go, it seems like the `docs/generated` folder is not fully included into pygame wheels. While the generated folder looks correct with `LGPL.txt` at the base and all the logos files in `_static` folder, some files seem to be skipped while building wheels with docs
`py setup.py docs`+`py setup.py install` and then `py -m pygame.docs` launches local doc copy which is stored in the wheels. Here the main page works, but clicking LGPL link does not work because `LGPL.txt` was not included. Logos page almost works, but links to svgs and zips are broken (links to png logos still work!) 

I believe this is something that can be fixed in
https://github.com/pygame/pygame/blob/c9fb2d8dd7cee24d43717e63389fa7d5aae9036c/setup.py#L456

dr0id:
> `py -m pygame.docs` launches local doc copy

right, didn't check that. I checked if it was generated correctly, but then I opened index.html directly in the browser from the generated directory.

Grimmys:
It's weird because on my side I was able to access LGPL.txt file from the browser (after cleaning the previous build). 

I will try again.

Grimmys:
@ankith26 can you try again please?

I was not able to reproduce the issue since `py setup.py install` is not working for me, it is not able to detect the Microsot Build Tools that are installed on my computer (why I even need them to build the docs?).

On my side the links were correctly working but well I changed the line you pointed out.

dr0id (approved):
I just tested a full re-build of this branch and then using `py -m pygame.docs` it all seems to work. I can open all logo images and the gpl license from that browser that opened.

ankith26 (approved):
LGTM! I tested again with `py -m pygame.docs` and now logos and LPGL text links are working again, thanks for the PR 👍 🎈

PR #2978, opened 2022-01-08
Snake casify `_sdl2.video.Image.flipX` and `.flipY`

avaxar:


avaxar:
image
Yes, hello? Anyone got a Python 3.10 installation? I'm on Mint 20.3, so.. I'm pretty much stuck with Python 3.8.

Starbuck5:
@ankith26 Cython generation is dependent on the Cython version, not the Python version. The generated result is compatible with all Cython-supported Python versions.

@avaxar I like this change, but I'd like to batch several changes like this into one PR (or one pygame version), to minimize how much of a pain it is to our early adopters.

For example, if we're going to rename renderer.present() to renderer.flip() / renderer.update(), as I think we should, we could do that at the same time.

avaxar:
Maybe there's a way to make this PR merge onto another branch of the unifying PR?

Starbuck5 (approved):
Looks good to me, thanks!

ankith26 (approved):
LGTM! Thanks 🎉

Starbuck5:
We've got to be careful when merging these that the Cython generation isn't overwriting other changes.

It should be fine for this one. But other Cython PRs will need to be rebased against main after this merges.

Issue #1574, opened 2020-02-09
Using 'SDL_WINDOWID' does not embed pygame display correctly into another application


AleffCorrea:
Windows 10, pygame 2.0.0.dev6 (SDL 2.0.10, python 3.8.1). tkinter 8.6

What previously worked on the latest stable release of pygame (1.9.6) for embedding the pygame wrapped SDL render window into a Tkinter frame doesn't work on the prerelease version I'm using anymore. Instead of properly drawing into the frame with id set into the SDL_WINDOWID environment variable, it simply spawns another window instead.

```   root = tkinter.Tk()
    self.frame = tkinter.Frame(root, width=self.width, height=self.height)
    self.frame.grid(row=0, column=2)
    root.update()
(...)
    os.environ['SDL_WINDOWID'] = str(self.frame.winfo_id())
    if sys.platform == "win32":
      os.environ['SDL_VIDEODRIVER'] = 'windows'
    pygame.display.init()
    self.screen = pygame.display.set_mode(canvas_res)
```
**Related Docs**: https://www.pygame.org/docs/ref/display.html#pygame.display.set_mode

robertpfeiffer:
I'm pretty sure this is a SDL2 thing. SDL_WINDOWID is not part of SDL2 any more. We might try to make this work again, but maybe not with the old API.

robertpfeiffer:
Functionality to do this used to exist in _sdl2, I'll link the relevant stuff for all the non-Discord people:

https://wiki.libsdl.org/SDL_GetWindowID
https://wiki.libsdl.org/SDL_GetWindowFromID
https://github.com/pygame/pygame/commit/45035cb136d85f95eae4fbc16da270aa8c72fd2f#diff-1b1808dc40ab95d48365c05261709336

kntroh:
I'm having trouble migrating my application to pygame 2 because of this problem.
Is it possible to use `SDL_CreateWindowFrom`?
https://wiki.libsdl.org/SDL_CreateWindowFrom

kntroh:
I tried SDL_CreateWindowFrom, and it seems to work well.
But, I don't know the right rule to convert a string to a pointer in pygame.
display.c:
```C
    char *window_handle = SDL_getenv("SDL_WINDOWID");
    (snip)
                /*open window*/
                if (window_handle) {
                    win = SDL_CreateWindowFrom(atoll(window_handle));
                }
                else {
[...]

PR #2981, opened 2022-01-09
"SDL_CreateWindowFrom" in "display" module

Rabbid76:
the environment variable "SDL_WINDOWID" is set, use "SDL_CreateWindowFrom" to create the window in "display.set_mode"

[Using 'SDL_WINDOWID' does not embed pygame display correctly into another application #1574](https://github.com/pygame/pygame/issues/1574)

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1096,7 +1096,10 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
 
             if (!win) {
                 /*open window*/
-                win = SDL_CreateWindow(title, x, y, w_1, h_1, sdl_flags);
+                const char *window_handle = SDL_getenv("SDL_WINDOWID");
+                win = window_handle
+                    ? SDL_CreateWindowFrom((const void*)atoll(window_handle))
+                    : SDL_CreateWindow(title, x, y, w_1, h_1, sdl_flags);
                 if (!win)
                     return RAISE(pgExc_SDLError, SDL_GetError());
                 init_flip = 1;


ankith26 (commented):
Thanks for the PR! 🎉 
I can't really provide an in-depth review because I'm not too experienced around display.c myself, but at a glance I noticed
1) `atoll` can return 0 when no conversion happens. IG in that case, `SDL_CreateWindow` should be called
2) [SDL_CreateWindowFrom](https://wiki.libsdl.org/SDL_CreateWindowFrom) docs mention 
> In some cases (e.g. OpenGL) and on some platforms (e.g. Microsoft Windows) the hint SDL_HINT_VIDEO_WINDOW_SHARE_PIXEL_FORMAT needs to be configured before using SDL_CreateWindowFrom().

So I wonder how we should be handling this

illume (changes requested):
Thank you very much. This is an important backwards compatibility fix.

Ankith already mentioned that atol can return an error.

Additionally, I wanted to ask about 32bit and 64bit windows... Is atol (long long?) the correct call to make on both 32/64 bit architectures?

illume:
I see there is test code here: https://github.com/pygame/pygame/issues/1574#issuecomment-1003910013
It would be good to confirm that the test code works on 32bit python and 64bit python.

Rabbid76:
The return type of [`std::atoll`](https://en.cppreference.com/w/cpp/string/byte/atoi) is `long long`. An alternative solution might be [`std::stoll`](https://en.cppreference.com/w/cpp/string/basic_string/stol).

ankith26:
Revisiting this PR, I believe [`atol`](https://www.tutorialspoint.com/c_standard_library/c_function_atol.htm) is the right function to use here. It returns a `long` not a `long long`, which should be the better datatype to use here IG. Because the size of a void pointer is not fixed at 8 bytes like `long long`, instead it can vary by implementation. I believe the sizeof `long` best matches the size of a void pointer on most platforms and configurations.

As mentioned above, like `atoll`, this function can return `0` on fail too. The error handling bit can look something like
```c
if (!window_handle) {
    /* NULL window_handle can segfault when *window_handle executes */
    window_handle = "";
}

long window_id = atol(window_handle);
if (!window_id && *window_handle != '0') {
    /* window_id is 0 when atol errors. It can also be 0 when the input was 0 */
    return RAISE(pgExc_SDLError, "'SDL_WINDOWID' must be an integer");
}
```

As illume pointed above, it would be good to test this change on 32 bit and 64 bit windows, bonus points if this can be put in a unit test ;)

MyreMylar:
@Rabbid76 are you still interested in finishing this off? It has been a while, so we could find someone else to take this over to get it into 2.1.3.

illume (approved):
👍 Thanks

PR #2982, opened 2022-01-09
class method "from_foreign_window" added to class "Window"

Rabbid76:
[Using 'SDL_WINDOWID' does not embed pygame display correctly into another application #1574](https://github.com/pygame/pygame/issues/1574)


MyreMylar (approved):
This seems to work. I can see the Renderer code embedded into a tkinter window.

Though I believe we also commit the generated cython .c files to the repo so you may also want to add that into your PR. As far as I can tell the existing _sdl2 .c files are all on different subversions of cython so it shouldn't matter too much as long as it is a 0.29.XX variant.

dr0id (approved):
code looks right, trusting the test of the other reviewer

ShadowElf37:
Updates on this?

PR #2939, opened 2021-12-21
SDL 2.0.18, skip some focus related tests

illume:
- buildconfig: Add SDL 2.0.18 to manylinux build
- buildconfig: Add SDL 2.0.18 to windows build

PR #2983, opened 2022-01-09
SDL 2.0.20

illume:


Starbuck5:
We had some event test failing on SDL 2.0.18, which you put in a skip for. I believe that's fixed now, so we can take that out.

illume:
The skips were added in this PR for SDL 2.0.18 https://github.com/pygame/pygame/pull/2939

ankith26:
SDL 2.0.20 has been released, this PR should move to that and not use the pre-release anymore now

illume:
I left tests skipping, as apparently they are still broken on windows.

illume:
Note, the manylinux images have been push with 2.0.20, and the tests here ran to completion with them.

ankith26 (approved):
LGTM! 🎉 Thanks! 👍

Starbuck5:
> I left tests skipping, as apparently they are still broken on windows.

They are, I re checked. But it's bizarre, if I run the module test individually, it passes. But if I run them through the complete test suite they fail.

Starbuck5 (approved):
Thanks!

I've tested locally on Windows.

PR #2987, opened 2022-01-16
Misc fixes: Fix UCS_4 UCS4 typo, incorrect color methods export and display.c error handling bugs

ankith26:
As the title says, another misc fixes:tm: PR.
1. In an older PR I had made a typo in code of naming the `UCS_4` runtime check constant as `UCS4`. Now reverting that will do more harm, so instead I updated docs and tests to check for `UCS4` instead
2. The color module was exporting the methods of `pygame.Color` as functions of the module itself, so those methods were usable from the `pygame.color` namespace, and calls to these functions would segfault. This is a one-liner fix
3. Some functions in display.c were not handling errors correctly, returning `None` instead of the expected `NULL` on error


illume:
- I'm unsure what impact the UCS_4 change will have, or what is the better way to go. It was only released a while ago, so changing it to UCS_4 and fixing the docs could be an option IMHO. But is the UCS_4 name really worth breaking that for? I don't know. Seems people use UCS4, UCS-4, and UCS_4... I feel if we had used "UCS4" in one of the examples this issue would have been caught. How did you spot it?
- pygame.color fix, cool. I guess better naming in the color.c module would have prevented this. How did you spot this error? With the type testing?
- display fix, cool. Would any tooling have caught this issue? Warnings/cppcheck?

illume (approved):
👍 🎉 thanks

ankith26:
Answered on discord, but just reposting here for others to see with some more info
1) IMHO `UCS4` is fine, it's not a typo that needs a minor breaking change to fix. Yeah I spotted this one while using mypy stubtest 
2) Yeah, stubtest caught this bug. Those functions were not typehinted (for obvious reasons, they are not supposed to exist) but the runtime attribute check of mypy stubtest caught these functions
3) Surprisingly any tooling (compiler warnings/static analyser/cppcheck) did not catch this one. I suspect it could be caught by a dedicated Python C extension checker (something like cpychecker but more modern perhaps?)

Starbuck5:
Fun fact about the color module thing, it's been in pygame since at least 1.9.2, the earliest version I could install from pip.

Starbuck5 (approved):
Alright I've looked through the code (tried to break your _render info thing), and the run tests locally, and things check out.

I don't understand the `UCS_4` thing really, but I trust your consideration of the outcomes.

PR #2988, opened 2022-01-16
SIMD version of BLEND_RGBA_MULT for 32bit surfaces

MyreMylar:
This PR explorers adding a SIMD version of the `BLEND_RGBA_MULT` blend mode. The goal is to speed up this blend mode and hopefully provide a template for speeding up the other special blend modes such as `BLEND_RGBA_ADD` and `BLEND_RGBA_MAX` with SIMD techniques.

**Background**

The special blend modes are used by supplying one of them to the `special_flag` parameter of blit, you can see a list of them in the documentation for `blit()` here:
https://www.pygame.org/docs/ref/surface.html#pygame.Surface.blit

They are useful for various effects, but up until now have been implemented entirely using standard C maths, generally doing the operations on a colour channel at a time. This is in contrast to our normal blitters and the `BLEND_PREMULTIPLIED_ALPHA` blitter which all use some for of SIMD acceleration either via SDL, or our own handwritten SIMD instructions.

Generally the existing SIMD blitters in pygame are written in SSE2 compliant intrinsic functions as these have wide support and have existed in most computers since around the year 2000. There is potential for speeding things up further with newer intrinsic standards (though those will have less widespread support), but I think it is best to at least get us into the year 2000 before we start reaching for 2022.

Arm processers have their own, similar intrinsic functions for speeding up these kinds of math operations and at the minute we use a library called `sse2neon` from here:
https://github.com/DLTcollab/sse2neon

To automatically convert our SSE2 intrinsic functions to arm neon versions. I suspect the porting is not as fast as it could be with native arm instructions written but I personally have no experience with arm intrinsic functions or a good setup to test them. In my experience programming with intrinsic functions is already fairly confusing, which is part of why I'm writing a lot in this PR to try and explain what I'm doing.

**SSE2 SIMD intrinsic functions**

The clearest guide to these is from intel here:
https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html

You can search for individual functions with the search box and tick the standards at the side to narrow down what intrinsic standard you wish to support up to. Generally newer standards are inclusive of older ones ( I think always?) so a processor supporting SSE2 also supports SSE and MMX.

The MMX standard supported 64 bit memory registers, while the SSE standard supports 128bit registers. If we were to create AVX SIMD versions we would have 256bit memory registers and the newest AV-512 standard supports, you guessed it 512 bit memory registers.

The point of these large memory registers and SIMD in general, is to run simultaneous functions on multiple parts of the large registers at the same time. For example you might have a function that does an add operation 8 times on each of the 8, 16 bit chunks of a 128bit register in one call. This is perfect for pixel operations as each colour channel is only an 8 bit number between 0 to 255 (or from 0x00 to 0xFF to put it in hexadecimal). Some operations are also saturating so if you add two 8 bit numbers together with an 8 bit adding function it will clamp the result to a max of 255 for you.

Running operations on data in the registers is very fast but loading data from normal memory into the SIMD registers can be quite slow. This is why you need to carefully profile to see if it is actually worth spending a lot of time loading multiple pixels worth of data into a SIMD register before running an operation on it.

Further complicating matters, some scalar math code gets automatically vectorised by compilers so you may be competing against a compiler that has already done some obvious optimisations in basic maths.

**This PR specifically** 

As I said at the beginning I want to create a simple template for SIMD versions of the special blend modes.

The last time I worked on SIMD it was suggested it might be better to move the SIMD functions with their extra imports and compiler demands into a separate file, so I thought I would open with an attempt to do this. This is why the new `blit_blend_rgba_mul_simd` is in its own header file - I'm not sure it is possible to stick it into a .c file and still have it be static and called from `alphablit.c` but having their own header is something at least. If this approach seems worthwhile more of the existing SIMD blitters could move into here instead of in `alphablit.c` . I also moved the blit_info structure into it's own header file which may not be necessary without a new .c file for the SIMD blitters but it tidied up the top of `alphablit.c` a bit so I left it.

Some extra if checks were then added to push SSE2 supporting processors handing a 32 bit to 32 bit surface blit over to the SIMD version of the blending function.

Then we get to the meat of the blending function. To make the blend work in a reasonable SIMD way I tweaked the formula for the blend. Both the existing formula and my new version are approximations of this operation:

```
dst = (dst * src) / 255
```
which is a standard multiply blend. However divide is too expensive an operation to do four times for every colour channel of every pixel so the existing formula for 32 bit surfaces is actually this:

```
tmp = (dst && src) ? (dst * src) >> 8 : 0
dst = (tmp <= 255 ? tmp : 255);
```

which uses a bitshift to approximate divide and branches to try and clamp the result between 0 and 255. I'm fairly sure that neither of these branches is necessary and the bit shift approximation could be better because it is prone to an off by one issue. If you do:

```
(255 * 255) >> 8
```

You will see that the result is 254 not 255 as you would expect from `dst = (dst * src) / 255` likewise:

```
((1 * 255)) >> 8
```
gives a result of 0 rather than 1.

Anyway, its just an approximation so it's not too bothersome but it could be better and the clamps between 0, and 255 are unnecessary as the max & min values you can inject into the bitshift calculation evaluate to 0 and 254. 

So, the first step of my SIMD journey is to fix this calculation up to be more accurate and faster. I propose we use this instead as the underlying operation:

```
dst = ((dst * src) + 255) >> 8
```

Adding 255 to the result of the multiplication fixes the off by one errors of the bitshift:

```
>>>((255 * 255) + 255) >> 8
255
>>>((1 * 255) + 255) >> 8
1
>>>((0 * 255) + 255) >> 8
0
>>>((128 * 200) + 255) >> 8 
100
```
And there is no need for any ternary operator branches because the result is already bound between 0 and 255. Withotu any branches it is also easy to vectorise with SIMD.

So far I've created a one-pixel-at-a-time version that:

 - first copies a single 32 bit pixels worth of data from the normal memory of the src and dst surfaces into two SIMD registers.
 - then spreads each 8 bit colour channel out over 16 bits so there is room for a multiplication (multiplying two 8 bit numbers together requires 16 bits of memory room for the result)  
 - then does the 16bit multiplication of the two SIMD registers, effectively multiplying each colour channel simultaneously.
 - then adds 255 onto the result of each channels multiply using another SIMD register filled with spread out 0xFF (255).
 - then does the bitshift on all the channels completing the blend.
 - finally we repack the "16 bit" channels back down to 8bit (they only have 8bit data at this point after the bit shift, it is just spread out)
 - then we take the new 32bit pixel out of the SIMD register back into normal memory. 

And that's it.

The next step is to see if I can make a version that does two pixels at a time, and if that version is actually any faster or if the time spend loading pixels in actually makes it slower than one pixel at a time. Doing 2 pixels at a time also requires an even width blit to function in the current blitting paradigm which complicates matters further.

**Performance results so far**

The only reason to do this is if it speeds up the blitter so performance data is important. I made a test program which does blits at various sizes with picked blend modes over a number of iterations and times how long it takes. Here is the program:

```
import time
import math
import pygame


def test(target_surf, test_surf, methods, tries, blit_pos):
    for method in methods:
        start = time.time()
        name = method(target_surf, test_surf, tries, blit_pos)
        results = (time.time() - start) * 1000
        name_tab_length = math.ceil(len('tested {}:'.format(name)) / 4)
        tabs = '\t' * (8 - name_tab_length)
        string_to_format = 'tested {}:' + tabs + '{}ms'
        print(string_to_format.format(name, round(results, 3)))


def blend_rgba_mult(target_surf, blit_surf, tries, blit_pos):
    for t in range(tries):
        target_surf.blit(blit_surf, blit_pos, special_flags=pygame.BLEND_RGBA_MULT)
    return 'Blit BLEND_RGBA_MULT'


def run_rgba32_blit_test(target_surf, iterations, width, height, blend_mode_funcs):
    test_name = 'Blend RGBA 32 Test'
    test_surf = pygame.Surface((width, height), flags=pygame.SRCALPHA, depth=32)
    test_surf.fill((100, 100, 100, 100))
    num_tries = iterations
    start_pos = ((screen.get_width() - width), screen.get_height() - height)
    odd_start_pos = ((screen.get_width() - width) + 1, screen.get_height() - height)
    print(f'\nTest type - {test_name}: {num_tries} blits of image size {width}, {height}')
    test(target_surf, test_surf, blend_mode_funcs, num_tries, start_pos)


pygame.init()
screen = pygame.display.set_mode((1280, 1024))

surf_to_blit_onto = pygame.Surface((1280, 1024))
surf_to_blit_onto.fill((50, 50, 50))

test_iterations = 1000
blend_modes_to_test = [blend_rgba_mult]  # (blend_add, blend_rgb_add, blend_mult, blend_rgb_mult)

run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 32, 32, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 64, 64, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 128, 128, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 256, 256, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 512, 512, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 1024, 1024, blend_modes_to_test)

``` 

Which produces the following results on 2.1.2 on my windows PC:

```
Test type - Blend RGBA 32 Test: 1000 blits of image size 32, 32
tested Blit BLEND_RGBA_MULT:	13.012ms

Test type - Blend RGBA 32 Test: 1000 blits of image size 64, 64
tested Blit BLEND_RGBA_MULT:	48.043ms 

Test type - Blend RGBA 32 Test: 1000 blits of image size 128, 128
tested Blit BLEND_RGBA_MULT:	191.175ms

Test type - Blend RGBA 32 Test: 1000 blits of image size 256, 256
tested Blit BLEND_RGBA_MULT:	761.176ms

Test type - Blend RGBA 32 Test: 1000 blits of image size 512, 512
tested Blit BLEND_RGBA_MULT:	3036.053ms

Test type - Blend RGBA 32 Test: 1000 blits of image size 1024, 1024
tested Blit BLEND_RGBA_MULT:	12022.943ms
```

Then after this PR in it's current state we get:

```
Test type - Blend RGBA 32 Test: 1000 blits of image size 32, 32
tested Blit BLEND_RGBA_MULT:	1.002ms = 13x faster

Test type - Blend RGBA 32 Test: 1000 blits of image size 64, 64
tested Blit BLEND_RGBA_MULT:	4.003ms = 12x faster

Test type - Blend RGBA 32 Test: 1000 blits of image size 128, 128
tested Blit BLEND_RGBA_MULT:	12.012ms = 15x faster

Test type - Blend RGBA 32 Test: 1000 blits of image size 256, 256
tested Blit BLEND_RGBA_MULT:	48.043ms = 15x faster

Test type - Blend RGBA 32 Test: 1000 blits of image size 512, 512
tested Blit BLEND_RGBA_MULT:	190.173ms = 15x faster

Test type - Blend RGBA 32 Test: 1000 blits of image size 1024, 1024
tested Blit BLEND_RGBA_MULT:	750.235ms = 16x faster
```

So between 12 and 16 times faster at all the different blit sizes so far. 

I'll update this draft as I try other things with this blend mode until I think I've got a satisfactory template for 32bit blits for the other RGBA_ special blend modes (MAX,MIN,SUB & ADD). To keep it (relatively) simple I'll add changes for those blend modes in other PRs later as I think this topic is complicated enough already.

Let me know if you have any ideas, feedback or thoughts. I'm sure there is lots of stuff I missed as this is a complex topic and I forget half of it each time I go away from it and come back after a break.


MyreMylar:
After some further experimentation I have:

 - **Determined that 'stride switching' works much better than expected.** This is where blits with widths that don't fit neatly into an '8 or 2 pixels at a time' blits do the excess pixels with a 'one-pixel-at-a-time' algorithm before switching to the 8 or 2 pixel version afterwards. Performance was easily comparable with the width switch version I had before in the optimum case and much better for the non-optimum cases. As a result all version of the blitter use this stride switch approach and average blit performance (assuming widths are random) is improved. 
 - **Upgraded the SSE2 version to do 2 pixels at a time when it can.** This unsurprisingly boosts it's performance.
 - **Implemented an AVX2 version of the blitter.** This is about as far into the SIMD tech as we can go at the minute and should get good usage (86.31% in the January Steam hardware survey - compared to 100% for SSE2). AVX-512 is not a usable technology at present ( there are a ton of sub-versions, intel has just dropped it, AMD hasn't added it yet and it only has about 5% uptake)
 - **Created a new script to test 'real-world' usage of the blitter.** I suspected the stats from my speed test were being inflated by the test being a 'perfect' setup for SIMD which was unlikely to occur in a real program. This proved to be true, the new blitters are faster - but not absurdly faster in a more real world scenario.
 - **Expanded the speed test to check very small (sub 4x4) blits in case they were slower**. They are not slower - but they also aren't any faster, unsurprisingly blitting many pixels at a time works better when there are many pixels to blit.

Here is my updated pure speed test script:
```
import time
import math
import pygame


def test(target_surf, test_surf, methods, tries, blit_pos, extra_desc):
    for method in methods:
        start = time.time()
        name = method(target_surf, test_surf, tries, blit_pos)
        results = (time.time() - start) * 1000
        desc_string_start = 'tested {}' + extra_desc + ':'
        name_tab_length = math.ceil((len(desc_string_start.format(name)) + 1) / 4)
        tabs = '\t' * (19 - name_tab_length)
        string_to_format = desc_string_start + tabs + '{}ms'
        print(string_to_format.format(name, round(results, 3)))


def blend_add(target_surf, blit_surf, tries, blit_pos):
    for t in range(tries):
        target_surf.blit(blit_surf, blit_pos, special_flags=pygame.BLEND_ADD)
    return 'Blit BLEND_ADD'


def blend_rgb_add(target_surf, blit_surf, tries, blit_pos):
    for t in range(tries):
        target_surf.blit(blit_surf, blit_pos, special_flags=pygame.BLEND_RGB_ADD)
    return 'Blit BLEND_RGB_ADD'


def blend_mult(target_surf, blit_surf, tries, blit_pos):
    for t in range(tries):
        target_surf.blit(blit_surf, blit_pos, special_flags=pygame.BLEND_MULT)
    return 'Blit BLEND_MULT'


def blend_rgb_mult(target_surf, blit_surf, tries, blit_pos):
    for t in range(tries):
        target_surf.blit(blit_surf, blit_pos, special_flags=pygame.BLEND_RGB_MULT)
    return 'Blit BLEND_RGB_MULT'


def blend_rgba_mult(target_surf, blit_surf, tries, blit_pos):
    for t in range(tries):
        target_surf.blit(blit_surf, blit_pos, special_flags=pygame.BLEND_RGBA_MULT)
    return 'Blit BLEND_RGBA_MULT'


def run_rgba32_blit_test(target_surf, iterations, width, height, blend_mode_funcs):
    test_name = 'Blend RGBA 32 Test'
    test_surf = pygame.Surface((width, height), flags=pygame.SRCALPHA, depth=32)
    test_surf.fill((100, 100, 100, 100))
    num_tries = iterations
    start_pos = ((screen.get_width() - width), screen.get_height() - height)
    odd_start_pos = ((screen.get_width() - width) + 1, screen.get_height() - height)
    # print(f'\nTest type - {test_name}: {num_tries} blits of image size {width}, {height}')
    test(target_surf, test_surf, blend_mode_funcs,
         num_tries, start_pos, f' - {num_tries} tries, even dimensions {width}, {height} ')
    test(target_surf, test_surf, blend_mode_funcs,
         num_tries, odd_start_pos, f' - {num_tries} tries, odd dimensions {width-1}, {height} ')


pygame.init()
screen = pygame.display.set_mode((1280, 1024))

surf_to_blit_onto = pygame.Surface((1280, 1024))
surf_to_blit_onto.fill((50, 50, 50))

test_iterations = 1000
blend_modes_to_test = [blend_rgba_mult]  # (blend_add, blend_rgb_add, blend_mult, blend_rgb_mult)

# standard size ranges
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 32, 32, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 64, 64, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 128, 128, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 256, 256, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 512, 512, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 1024, 1024, blend_modes_to_test)

# very small size ranges
test_iterations = 50000
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 16, 16, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 8, 8, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 4, 4, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 2, 2, blend_modes_to_test)
run_rgba32_blit_test(surf_to_blit_onto, test_iterations, 1, 1, blend_modes_to_test)

```

And here is my 'real-world' test script that switches on a full-screen 'post-process' vignette and noise when you press the return key:

```
from collections import deque
import random
import pygame


def generate_random_white_noise(size):
    noise_surf = pygame.Surface(size, flags=pygame.SRCALPHA)
    noise_surf.fill((255, 255, 255, 255))

    for x in range(0, size[0]):
        for y in range(0, size[1]):
            noise_val = random.gauss(0.975, 0.008)
            rand_lum = min(255, max(0, int(noise_val * 255.0)))
            noise_surf.set_at((x, y), pygame.Color(rand_lum, rand_lum, rand_lum, 255))
    return noise_surf


pygame.init()
window_size = (1920, 1080)
window_surf = pygame.display.set_mode(window_size)
background_surf = pygame.Surface(window_size)
background_surf.fill((0, 0, 0))
game_level_image = pygame.image.load('test_images/game_level_test.jpg').convert()
game_level_image = pygame.transform.smoothscale(game_level_image, window_size)

vignette_small = pygame.Surface((240, 135))
vignette_small.fill((0, 0, 0))
pygame.draw.rect(vignette_small, pygame.Color((32, 32, 32)),
                 pygame.Rect((-20, -20), (280, 175)), border_radius=128)
pygame.draw.rect(vignette_small, pygame.Color((64, 64, 64)),
                 pygame.Rect((-15, -15), (270, 165)), border_radius=128)
pygame.draw.rect(vignette_small, pygame.Color((96, 96, 96)),
                 pygame.Rect((-10, -10), (260, 155)), border_radius=128)
pygame.draw.rect(vignette_small, pygame.Color((128, 128, 128)),
                 pygame.Rect((-5, -5), (250, 145)), border_radius=128)
pygame.draw.rect(vignette_small, pygame.Color((160, 160, 160)),
                 pygame.Rect((0, 0), (240, 135)), border_radius=128)
pygame.draw.rect(vignette_small, pygame.Color((192, 192, 192)),
                 pygame.Rect((5, 5), (230, 125)), border_radius=128)
pygame.draw.rect(vignette_small, pygame.Color((224, 224, 224)),
                 pygame.Rect((10, 10), (220, 115)), border_radius=128)
pygame.draw.rect(vignette_small, pygame.Color((255, 255, 255)),
                 pygame.Rect((15, 15), (210, 105)), border_radius=128)

vignette_small = pygame.transform.smoothscale(vignette_small, (60, 32))
vignette = pygame.transform.smoothscale(vignette_small, window_size)

noise_anim = [pygame.transform.smoothscale(generate_random_white_noise((480, 270)), window_size),
              pygame.transform.smoothscale(generate_random_white_noise((480, 270)), window_size),
              pygame.transform.smoothscale(generate_random_white_noise((480, 270)), window_size),
              pygame.transform.smoothscale(generate_random_white_noise((480, 270)), window_size),
              pygame.transform.smoothscale(generate_random_white_noise((480, 270)), window_size),
              pygame.transform.smoothscale(generate_random_white_noise((480, 270)), window_size),
              pygame.transform.smoothscale(generate_random_white_noise((480, 270)), window_size),
              pygame.transform.smoothscale(generate_random_white_noise((480, 270)), window_size)]
noise_index = 0
noise = noise_anim[noise_index]
noise_acc = 0.0
noise_frametime = 0.05

font = pygame.font.Font(None, 64)

time_delta_stack = deque([])
display_vignette = False
clock = pygame.time.Clock()
running = True
fps_text = "FPS: "
average_time_delta = 0.0
while running:
    time_delta = clock.tick()/1000.0

        if event.type == pygame.QUIT:
            running = False
            if event.key == pygame.K_ESCAPE:
                running = False
            if event.key == pygame.K_RETURN:
                if display_vignette:
                    display_vignette = False
                else:
                    display_vignette = True

    noise_acc += time_delta
    if noise_acc >= noise_frametime:
        noise_acc = 0.0
        noise_index += 1
        if noise_index == len(noise_anim):
            noise_index = 0
        noise = noise_anim[noise_index]

    window_surf.blit(background_surf, (0, 0))
    window_surf.blit(game_level_image, (0, 0))
    if display_vignette:
        window_surf.blit(noise, (0, 0), special_flags=pygame.BLEND_RGBA_MULT)
        window_surf.blit(vignette, (0, 0), special_flags=pygame.BLEND_RGBA_MULT)

    time_delta_stack.append(time_delta)
    if len(time_delta_stack) > 100:
        time_delta_stack.popleft()

    if len(time_delta_stack) == 100:
        average_time_delta = sum(time_delta_stack) / 100.0
        fps_text = f'FPS: {min(999.0, 1.0 / max(average_time_delta, 0.0000001)):.2f}'
        fps_surf = font.render(fps_text, True, pygame.Color('#FFFFFF'))
        window_surf.blit(fps_surf, (1650, 80))

    pygame.display.update()

print(fps_text)
print("time delta: ", average_time_delta)

```

I used this test image:
game_level_test

Which looks like this with the post-process running:
image

Here are the results from the speed tests and real-world tests on my windows PC with an AVX2 capable processor (AMD Ryzen 5 3600X):

**Current (pygame 2.1.2):**

```
OLD
----
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 32, 32 :		13.012ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 31, 32 :		12.011ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 64, 64 :		50.046ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 63, 64 :		48.044ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 128, 128 :		190.172ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 127, 128 :		187.171ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 256, 256 :		747.405ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 255, 256 :		740.675ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 512, 512 :		2986.622ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 511, 512 :		2984.607ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 1024, 1024 :		11956.732ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 1023, 1024 :		11858.296ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 16, 16 :		178.633ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 15, 16 :		167.445ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 8, 8 :		63.527ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 7, 8 :		59.053ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 4, 4 :		36.033ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 3, 4 :		32.03ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 2, 2 :		27.025ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 1, 2 :		26.023ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 1, 1 :		25.023ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 0, 1 :		21.019ms

Real world test:
----------------
FPS: 28.77
time delta:  0.03476000000000002
```

**This branch with SSE 2 only (AVX2 disabled):**

```
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 32, 32 :		1.001ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 31, 32 :		1.001ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 64, 64 :		2.002ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 63, 64 :		2.09ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 128, 128 :		6.917ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 127, 128 :		7.007ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 256, 256 :		26.023ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 255, 256 :		28.026ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 512, 512 :		104.094ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 511, 512 :		107.097ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 1024, 1024 :		408.841ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 1023, 1024 :		425.025ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 16, 16 :		30.982ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 15, 16 :		30.992ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 8, 8 :		26.005ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 7, 8 :		28.026ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 4, 4 :		24.977ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 3, 4 :		25.996ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 2, 2 :		25.005ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 1, 2 :		25.004ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 1, 1 :		23.995ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 0, 1 :		21.001ms

Real world test:
----------------
FPS: 147.06
time delta:  0.006800000000000005
```

**This branch with AVX2:**

```
AVX2
----
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 32, 32 :		1.0ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 31, 32 :		0.0ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 64, 64 :		2.002ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 63, 64 :		1.001ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 128, 128 :		3.003ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 127, 128 :		4.004ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 256, 256 :		13.012ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 255, 256 :		14.013ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 512, 512 :		44.04ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 511, 512 :		48.043ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 1024, 1024 :		172.158ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 1023, 1024 :		176.678ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 16, 16 :		28.999ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 15, 16 :		32.0ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 8, 8 :		26.999ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 7, 8 :		28.007ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 4, 4 :		25.994ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 3, 4 :		25.006ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 2, 2 :		25.988ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 1, 2 :		25.014ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 1, 1 :		25.915ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 0, 1 :		21.082ms

Real world test:
----------------
FPS: 149.93
time delta:  0.006670000000000005
```

As you can see, while AVX2 looks great in the speed benchmark, it is only marginally better than SSE2 in the real world test. Both SIMD blitters, in both tests, seem to be a bit improvement over the current non-SIMD version. 

Assuming most blits are not supper small or super large and that real world blitting is not going to be as good as the speed benchmark - perhaps it comes out to about 5 times faster?

I'd like to see this tested on ARM (needs SIMD enabled) and I guess linux and mac (GCC rather than MSVC compiler) to see if it get similar results. 

I also clearly need to figure out how to install clang to get his passing the CI.

MyreMylar:
I realised there was a bug in my original 'real world' blitter test. I was missing a `pygame.SRCALPHA` flag in the creating of the vignette overlay. Removing this bug the 'real world' performance is a lot better.

This did highlight that the multiply work won't *really* be complete until I also upgrade the RGB_MULT blend mode to SIMD and can pass RGBA blends with no `SRCALPHA` flag over to SIMD versions of that blitter. However I still think that is best done in a second, simpler PR, if this one manages to pass review muster.

dr0id (approved):
I would lie if I claim that I understood all of it, but it looks reasonable what you did. Ans since all tests pass it seems to work. Maybe I will try it later locally.

Here are my results for the 'real world' example:
```
...> py -3 .\realworldtest.py
pygame 2.1.3.dev5 (SDL 2.0.20, Python 3.9.7)
FPS: 220.26
time delta:  0.004540000000000003

...> py -3 .\realworldtest.py
pygame 2.1.3.dev5 (SDL 2.0.20, Python 3.9.7)
FPS: 8.77
time delta:  0.11398000000000005




...> py -3 .\realworldtest.py
pygame 2.1.3.dev1 (SDL 2.0.20, Python 3.9.7)
FPS: 215.05
time delta:  0.004650000000000004

...>py -3 realworldtest.py
pygame 2.1.3.dev1 (SDL 2.0.20, Python 3.9.7)
FPS: 125.16
time delta:  0.007990000000000006
```
And here the performance tests:
```
pygame 2.1.3.dev5 (SDL 2.0.20, Python 3.9.7)
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 32, 32 :              32.002ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 31, 32 :               28.964ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 64, 64 :              107.94ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 63, 64 :               126.928ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 128, 128 :    428.773ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 127, 128 :             371.809ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 256, 256 :    1606.415ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 255, 256 :             1723.095ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 512, 512 :    6217.93ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 511, 512 :             6560.483ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 1024, 1024 :  26056.485ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 1023, 1024 :   25871.302ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 16, 16 :             390.777ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 15, 16 :              368.789ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 8, 8 :               117.93ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 7, 8 :                115.934ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 4, 4 :               64.964ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 3, 4 :                68.96ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 2, 2 :               47.993ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 1, 2 :                44.974ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 1, 1 :               46.973ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 0, 1 :                37.978ms



pygame 2.1.3.dev1 (SDL 2.0.20, Python 3.9.7)
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 32, 32 :              2.997ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 31, 32 :               2.999ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 64, 64 :              4.997ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 63, 64 :               5.996ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 128, 128 :    11.996ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 127, 128 :             15.99ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 256, 256 :    36.977ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 255, 256 :             38.977ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 512, 512 :    156.029ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 511, 512 :             135.041ms
tested Blit BLEND_RGBA_MULT - 1000 tries, even dimensions 1024, 1024 :  778.122ms
tested Blit BLEND_RGBA_MULT - 1000 tries, odd dimensions 1023, 1024 :   866.996ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 16, 16 :             52.967ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 15, 16 :              60.964ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 8, 8 :               49.084ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 7, 8 :                51.97ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 4, 4 :               58.981ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 3, 4 :                45.975ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 2, 2 :               43.126ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 1, 2 :                43.971ms
tested Blit BLEND_RGBA_MULT - 50000 tries, even dimensions 1, 1 :               43.975ms
tested Blit BLEND_RGBA_MULT - 50000 tries, odd dimensions 0, 1 :                39.052ms
```

dr0id:
This should fix #3071 if I understood the comment correctly.

setup.py
```diff
@@ -71,6 +71,53 @@
...
+compiler_options = {
+    'unix': ('-mavx2',),
+    'msvc': ('/arch:AVX2',)
+}
+
+
+def spawn(self, cmd, **kwargs):
+    if (('ARM' not in platform.uname().version) and
+            platform.machine() != 'aarch64' and
+            'arm' not in platform.machine()):
```
ankith26, line 88: I'm not sure how this would affect our mac M1 builds. Currently, we have a bit of a hacky solution where our M1 builds are made on an intel mac, but cross compiled to `arm`. So in `setup.py`, `platform.machine()` and `platform.uname()` would still return intel info on our arm builds.

Since sse2 and avx are intel/AMD specific stuff, this check needs to check for that somehow, because checking for ARM to not match misses other exotic platforms like `ppc` and `s390x`

MyreMylar: I think I figured out a reasonable way to check for M1 mac cross compiling by using the fact that the arm-simd flags are set in the mac .sh file when building for M1 macs. 

There still seems to be ongoing debate over at CPython HQ on a standardised way to communicate a cross compilation scenario. At least I found an unresolved discussion & PEP from 2021:
https://github.com/benfogle/cross-compile-pep-draft/pull/1/files


Do you know of any list of supported pygame platforms? I could exclude them by `platform.machine()`/uname name as I've done for arm and perhaps provide a fall back compiler flag `-disable-avx2` to catch anything else we miss for really obscure platforms. Or I could perhaps reverse the check and just add a list of supported architectures (windows, non-m1 macs, non arm linux), or add an `-enable-avx2` flag that would could put on for our avx2 supporting wheels.

ankith26: hmmm I still don't think checking for `-enable-arm-(neon|simd)` would be accurate on our arm mac builds, because these flags are not passed to `setup.py` itself. One way I can think of check of is checking for `MAC_ARCH` environment variable, which is a custom thing defined by our mac CI, that would be `arm64` on M1 and `x86_64` otherwise

> Do you know of any list of supported pygame platforms?

AFAIK there is no such list for pygame, but I find [this wiki](https://en.wikipedia.org/wiki/Uname) handy that keeps a list of possible `uname` return codes.

I think the test can look something like this, which tests for intel/amd specifically, but allows both 32-bit and 64-bit variants
```py
mach = platform.machine()
if (
    mach.startswith("x86") or mach.startswith("i686") or mach.lower() == "amd64"
) and os.environ.get("MAC_ARCH") != "arm64":
    # do intel specific stuff here
    pass
```

robertpfeiffer:
Is this still faster than compiling the simple code with `gcc -O2 -march=native -ftree-vectorize`? or `gcc -O3` or clang?

MyreMylar:
> Is this still faster than compiling the simple code with `gcc -O2 -march=native -ftree-vectorize`? or `gcc -O3` or clang?

Well, this also changes the formula used and adds CPU dispatch so that AVX2 code containing wheels should run fine on non-AVX2 having computers. I don't have gcc on my windows machine so I couldn't tell you if could optimise it on linux by itself, if anyone else wants to try it out they are welcome.

MyreMylar:
> Is this still faster than compiling the simple code with `gcc -O2 -march=native -ftree-vectorize`? or `gcc -O3` or clang?

Looking at our build_wheels script line 29:

`export CFLAGS="-g0 -O3"`

It looks like we already use `-O3`, which includes auto vectorization up to SSE2. Since the SSE2 versions of these blitting functions are much faster than the scalar version already as far as I can tell, it seems unlikely that there is a combo of gcc flags that would do better.

Secondary evidence is that SDL still uses intrinsic (and assembly) blitter code across multiple platforms rather than just relying on auto-vectorization. I suspect that auto-vectorization hasn't yet eliminated all programmers from the SIMD arena yet. 

Would be happy to be proved wrong though, but don't think we should let this sit around waiting for some compiler genius.

Currently this looks like about a ten times speed up in the common use case (32x32ish pixel blits), and even better for larger blits. That's a lot of speed to leave on the table.

ankith26 (approved):
Approving this, even though I admit I don't 100% understand it. But since the unit tests pass, it should be good to merge. Once merged, it will also be used and tested more widely by users, which is the best way of testing this sort of thing, because this can be hardware dependent, as you mentioned on discord.

Thanks for the PR and testing with the benchmarks, these are super nice speedups! 🎉 👍

PR #2989, opened 2022-01-16
Rect collidelistall unittests added and updated docs to document different argument types

dr0id:
Reading the collidelistall method in c and then the unittests I found that there are some missing to fully test the c code paths. Therefor added more tests with the different types of arguments.

Also updated the docs to document the different argument types.


docs/reST/ref/rect.rst
```diff
@@ -348,6 +348,52 @@
...
+          list_of_lists = [[1,1,1,1], [2,2,2,2]]
+          indices1 = r.collidelistall(list_of_lists)
+    
+          list_of_tuples = [(1,1,1,1), (2,2,2,2)]
+          indices2 = r.collidelistall(list_of_tuples)
+    
+          list_of_double_tuples = [((1,1), (1,1)), ((2,2), (2,2))]
+          indices3 = r.collidelistall(list_of_double_tuples)
+    
+          class ObjectWithRectAttribute(object):
```
novialriptide, line 371: You could remove the (object) here because of Python3. Same for the others

novialriptide (changes requested):

docs/reST/ref/rect.rst
dr0id: sure, old habits, thanks :)

docs/reST/ref/rect.rst
```diff
@@ -348,6 +348,52 @@
...
+                  self._rect = r
+    
+              def rect(self):
+                  return self._rect
+    
+          list_of_object_with_callable_rect = [
+                                              ObjectWithCallableRectAttribute(Rect(1,1,1,1)), 
+                                              ObjectWithCallableRectAttribute(Rect(2,2,2,2))
+          ]
+          indices5 = r.collidelistall(list_of_object_with_callable_rect)
```
ankith26, line 394: minor nitpick, this codeblock could do better with `black` formatting. We use black throughout the codebase, so as to maintain consistency ;)

dr0id: your are right, reformatted example code with black

MyreMylar (approved):
LGTM. Can't hurt to have some more valid tests with lots of Rect changes being worked on.

illume (approved):
👍 🎈 thanks

PR #2931, opened 2021-12-19
added a limit method to Vector2 & Vector3

Emc2356:
this is a shortcut for:
`if vector.magnitude() > 4: vector.scale_to_length(4)`

PR #2990, opened 2022-01-17
Add a .clamp_magnitude() to Vector2 and Vector3

novialriptide:
Continuation from: https://github.com/pygame/pygame/pull/2931


src_c/math.c
```diff
@@ -792,6 +796,75 @@ vector_copy(pgVector *self)
...
+    if (old_length == 0) {
+        PyErr_SetString(PyExc_ValueError,
+                        "Cannot clamp a vector with zero length");
+        return NULL;
+    }
+
+    if (old_length > max_length) {
+        /* Scale to length */
+        old_length =
+            sqrt(_scalar_product(self->coords, self->coords, self->dim));
```
dr0id, line 852: this has already be calculated on line 833?

src_c/math.c
```diff
@@ -792,6 +796,75 @@ vector_copy(pgVector *self)
...
+    else if (old_length < min_length) {
+        /* Scale to length */
+        old_length =
+            sqrt(_scalar_product(self->coords, self->coords, self->dim));
+
+        fraction = min_length / old_length;
+        for (i = 0; i < self->dim; ++i)
+            self->coords[i] *= fraction;
+
+        Py_RETURN_NONE;
```
dr0id, line 870: this looks like the same code as in the other if block... maybe this can be extracted so the code duplication is removed?

test/math_test.py
```diff
@@ -1046,6 +1046,13 @@ def test_project_v2_raises_if_other_is_not_iterable(self):
 
         # act / assert
         self.assertRaises(TypeError, v.project, other)
+        
+    def test_clamp_mag_v2(self):
```
dr0id, line 1144: I miss a test for the clamp_magnitude  the minimum. 
Sometimes it helps to think about the limits and write tests especially for those. So in this case I would expect:

1. a test to clamp to a minimum
2. a test to clamp to the maximum
3. a test where the vector is left as it is (no clamping)

And of course for each implemented method (v2.clamp_magnitude, v2.clamp_magnitude_ip, v3.clamp_magnitude, v3.clamp_magnitude_ip)

src_c/math.c
```diff
@@ -792,6 +796,75 @@ vector_copy(pgVector *self)
...
+    double fraction;
+    static char *keywords[] = {"max_length", "min_length", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "d|d", keywords,
+                                     &max_length, &min_length)) {
+        return NULL;
+    }
+
+    /* Get magnitude of Vector */
+    old_length = sqrt(_scalar_product(self->coords, self->coords, self->dim));
```
dr0id, line 841: Optimization: I'm not sure how fast the sqrt() method is today... but why not avoid it here and only apply it when needed?

E.g. 
```
old_length_sq = sqrt(_scalar_product(self->coords, self->coords, self->dim));
if(old_length_sq == 0)
  ...
...
if(old_length_sq > max_length * max_length)
  fraction = max_length /  sqrt(old_length_sq);
  ...
...
```

dr0id (changes requested):

src_c/math.c
```diff
@@ -800,6 +804,73 @@ vector_copy(pgVector *self)
...
+    if (old_length > max_length) {
+        /* Scale to length */
+        fraction = max_length / old_length;
+        for (i = 0; i < self->dim; ++i)
+            self->coords[i] *= fraction;
+
+        Py_RETURN_NONE;
+    }
+    
+    old_length =
```
dr0id, line 858: isn't old_length already calculated on line 841?

src_c/math.c
novialriptide: There's a difference between the two since in the variable `fraction`, the numerator is different in both blocks of code. Since C doesn't support nested functions, I'm not sure how I'm able to remove this duplicate code.

dr0id (approved):

src_c/math.c
```diff
@@ -800,6 +804,69 @@ vector_copy(pgVector *self)
...
+    pgVector *ret;
+
+    ret = (pgVector *)pgVector_NEW(self->dim);
+    if (ret == NULL)
+        return NULL;
+
+    for (i = 0; i < self->dim; ++i)
+        ret->coords[i] = self->coords[i];
+
+    vector_clamp_magnitude_ip(ret, args, kwargs);
```
ankith26, line 820:
```suggestion
    PyObject *ret_val = vector_clamp_magnitude_ip(ret, args, kwargs);
    if (!ret_val) {
        return NULL;
    }
    Py_DECREF(ret_val);
```
There's a memory leak and an error checking issue here. When `vector_clamp_magnitude_ip` fails it returns `NULL`, so this function should return `NULL` too when that happens. When `vector_clamp_magnitude_ip` returns python object `None` after a successful call, it should be `Py_DECREF`d like any other python object

test/math_test.py
```diff
@@ -1141,6 +1141,20 @@ def test_project_v2_raises_if_other_is_not_iterable(self):
         # act / assert
         self.assertRaises(TypeError, v.project, other)
 
+    def test_clamp_mag_v2_max(self):
+        v1 = Vector2(7, 2)
+        v2 = v1.clamp_magnitude(5)
```
ankith26, line 1146:
```suggestion
        v2 = v1.clamp_magnitude(5)
        v1.clamp_magnitude_ip(5)
        self.assertEqual(v1, v2)
```
Would be nice to include `clamp_magnitude_ip` checks in integration with these tests, and the other `clamp_magnitude` tests

test/math_test.py
```diff
@@ -1141,6 +1141,20 @@ def test_project_v2_raises_if_other_is_not_iterable(self):
...
+        expected_v2 = Vector2(4.80762, 1.37361)
+        self.assertEqual(expected_v2.x, round(v2.x, 5))
+        self.assertEqual(expected_v2.y, round(v2.y, 5))
```
ankith26, line 1149: It's better to use the `assertAlmostEqual` function here, and everywhere else in the remaining `clamp_magnitude` tests

ankith26 (changes requested):
Nice PR, looking almost good to go, but before that, left some reviews :)

src_c/math.c
```diff
@@ -800,6 +804,80 @@ vector_copy(pgVector *self)
...
+
+    /*
+    Notes for other contributors reading this code:
+    The numerator for the fraction is different.
+    */
+    length_greater = old_length_sq > max_length * max_length;
+    if (length_greater) {
+        /* Scale to length */
+        fraction = max_length / sqrt(old_length_sq);
+        Py_RETURN_NONE;
```
dr0id, line 865: the fraction needs to be applied before leaving the method

src_c/math.c
```diff
@@ -800,6 +804,80 @@ vector_copy(pgVector *self)
...
+    }
+
+    length_less = old_length_sq < min_length * min_length;
+    if (length_less) {
+        /* Scale to length */
+        fraction = min_length / sqrt(old_length_sq);
+        Py_RETURN_NONE;
```
dr0id, line 872: the fraction needs to be applied before leaving the method

dr0id (changes requested):

dr0id (approved):

novialriptide:
> there are some build errors:
> 
> ```
>      src_c/math.c: In function ‘vector_clamp_magnitude_ip’:
>     src_c/math.c:874:25: error: ‘fraction’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
>       874 |         self->coords[i] *= fraction;
>           |                         ^~
>     cc1: all warnings being treated as errors
> ```

Thanks! Hopefully, this latest commit has fixed that.

docs/reST/ref/math.rst
```diff
@@ -917,6 +945,34 @@ Multiple coordinates can be set using slices or swizzling
...
+      | :sl:`Clamps the vector's magnitude between max_length and min_length`
+      | :sg:`clamp_magnitude_ip(max_length) -> Vector3`
+      | :sg:`clamp_magnitude_ip(max_length, min_length) -> Vector3`
+
+      Clamps the vector's magnitude between max_length and min_length. If a min_length value is not assigned,
+      then it will automatically be set to 0.
+
+      .. versionadded:: 2.1.3
+
+      .. ## Vector3.clamp_magnitude ##
```
ankith26, line 974: should read `.. ## Vector3.clamp_magnitude_ip ##` although I dunno whether it would break something because it's just a comment in my understanding.

dr0id (approved):

src_c/math.c
```diff
@@ -800,6 +804,80 @@ vector_copy(pgVector *self, PyObject *_null)
     return (PyObject *)ret;
 }
 
+static PyObject *
+vector_clamp_magnitude(pgVector *self, PyObject *args, PyObject *kwargs)
+{
+    Py_ssize_t i;
+    pgVector *ret;
+
+    ret = (pgVector *)pgVector_NEW(self->dim);
```
MyreMylar, line 813: Will this also need to be updated to use the new subtype changes in #3088 ?

MyreMylar (changes requested):
LGTM, apart from the question over the new subtype PR.

src_c/math.c
novialriptide: I plan on updating this in the subtype changes PR itself, unless the subtype PR gets merged first.

MyreMylar (approved):

Issue #2623, opened 2021-06-11
negative decimal treated differently then positive decimals when moving rect


cheeze36:
I noticed a weird bug and I don't know if it in pygame or python. If I try to move rect class by add .1 to the x it doesn't move which is correct cause it expects an integer bit if I try moving it by -.1 it moves -1.

Starbuck5:
I can't replicate this.

pygame 2.0.1 windows.

pradkrish:
I am new to pygame, is it okay if I work on this issue?

CosmicPegasis:
> isn't that because of rounding down, `int(9.9)` becomes 9 and `int(10.1)` becomes 10. Is this really a bug?

I think this is more of a semantics and uniformity thing. Pardon me I'm not that familiar with the source code but if on adding a float > 1 and < 2, if it increments the position by 1, then this behaviour should be allowed, preferably with documentation and vice versa.

snyntakk404:
what happens is that they round it. when the rounded result equals 0 they change it to 1 if arg was a positive decimal or -1 if arg was a negative decimal. this isn't a bug. it is phisically impossible to move an object by 0.1 pixel. pixels are what makes up the screen.

d34d0s:
Note: The FRect PR is linked but this issue remained open.

PR #2995, opened 2022-01-19
Round floating point values to integers in rect property assignment

dr0id:
This should solve issue #2623.

I do not like the solution. Not sure why. If there is something that can be done differently, please let me know.

1. Not sure if working with doubles is acceptable (performance?).
2. Not sure how the performance is.
3. Not sure if the tests will work on all platforms (it assumes that INT_MAX and INT_MIN is the same on every platform).

I had some trouble using floats because it seems that it looses precision. Especially around INT_MAX and INT_MIN.


MyreMylar:
Do we need to solve this? Pygame is using the default casting behaviour from floats to ints. If you want to change that rounding behaviour the correct solution is generally to use round:

```
import pygame

my_rect = pygame.Rect(10, 10, 100, 100)

my_rect.x = 3.99999
print(my_rect.x)

my_rect.x = round(3.99999)
print(my_rect.x)

my_rect.x = round(my_rect.x - 0.1)
print(my_rect.x)

my_rect.x = my_rect.x - 0.1
print(my_rect.x)
``` 

If we change this default behaviour then it would no doubt confuse people who expect it to work the normal way, and also break a bunch of existing code that depends on the standard python way of doing things with float to integer casting. 

I think it is probably better to leave Rect alone in it's integer based world, and have a FRect class for those who want a floating point Rectangle class.

dr0id:
This branch won't be needed anymore if FRect makes it into pygame I think. Close at will or let me know if there other reasons to keep or merge this.

PR #2997, opened 2022-01-20
Added support for complex numbers for Vector2

gresm:
Allows to convert python "complex" object into pygame Vector2 and adds ``vector2.to_complex()`` to convert Vector2 into "complex" object.


MyreMylar:
I did a quick investigation into complex numbers as my videogame math foo is often terrible, and looks like they do have plenty of videogame-y uses for rotation. See:
https://gamedev.stackexchange.com/questions/178736/how-are-imaginary-numbers-useful-in-video-game-creation


Looks like this has PR built up a couple of conflicts. I don't fancy resolving them in the command line as github is currently recommending so hopefully @gresm can fix them again.

gresm:
> Apart from the merge issues this looks good to me

Merge issues have been now fixed.

Starbuck5:
I don't think this should be merged, because it seems to be a very uncommon use and because it's possible to emulate without much python code on the users end if they need it.

Novial is working on vector subclassing, which would allow users to define their complex number support however they want, if they need it. #3088

gresm:
Linking this pull request to the issue:
Fixes #2997

gresm:
Fixes #2997

MyreMylar:
Here is a simple example program that hopefully demonstrates where this might be useful in a game:

```
import math

import pygame
from pygame import Vector2
from math import sin, cos, radians

pygame.init()

window_surface = pygame.display.set_mode((800, 600))

angle = 90
start_vec = Vector2(0, -100)

# Two rotation options here: standard Vector2 rotate(), or complex conversion rotation by multiplication
# rotated_vec = start_vec.rotate(angle)

# first we create the complex number to perform the rotation and convert our start vector to complex
complex_rotation = complex(cos(radians(angle)),
                           sin(radians(angle)))
complex_start_vec = complex(start_vec)

# then to do the actual rotation is only a multiply, this could allow for a lot of cheap rotations
# at run time where the angle to rotate at is known ahead of time
rotated_vec = Vector2(complex_start_vec * complex_rotation)

centre_vec = Vector2(400, 300)

translated_start_vec = start_vec + centre_vec
translated_rotated_vec = rotated_vec + centre_vec

running = True
while running:

        if event.type == pygame.QUIT:
            running = False

    window_surface.fill((0, 0, 0))

    pygame.draw.line(window_surface, (255, 255, 255), centre_vec, translated_start_vec)
    pygame.draw.line(window_surface, (255, 0, 255), centre_vec, translated_rotated_vec)

    pygame.display.flip()



```

Issue #2867, opened 2021-11-26
pygame.event.EventType lacks proper typing definition


ItsDrike:
## Environment

You can get some of this info from the text that pops up in the console when you run a pygame program.

- **Operating system**: Arch Linux (5.15.4-arch1.1)
- **Python version**: 3.9.7 (poetry virtual environment)
- **SDL version**: SDL 2.0.16
- **PyGame version**: 2.1.0

## Current behavior

Importing `pygame.event.EventType` causes issues with static type checkers because pygame/event.pyi doesn't define it
anywhere. There is only the type definition of `_EventTypes`, which on the other hand isn't actually importable and causes
and `ImportError` exception when an import is attempted.

## Expected behavior

I assume that the `_EventTypes` type defined in the pyi file was actually supposed to say `EventType`, and if not, there may
be some bigger issue but there should be a valid type definition for importing `EventType` from `pygame.event` and the invalid
definitions should be removed.

While I'm not aware of anything else having this issue, I wouldn't be surprised if there were more variables/classes/functions that
weren't added into the pyi files and yet they still exist and are importable.

## Screenshots

![eventtype](https://user-images.githubusercontent.com/20902250/143592490-debb6ab1-ec26-4545-9775-29ff7714e07b.png)

## Test code

```py
from pygame.event import EventType  # will run, but causes typing problems since there's no definition of it
from pygame.event import _EventTypes  # won't run, yet there is a typing definition of it
```

## Type checker error reports:

Pyright:
```
 /home/itsdrike/Personal/Programming/Python/Projects/GUI/PyGame/Template/src/util/base_game.py:4:26 - error: "EventType" is unknown import symbol (reportGeneralTypeIssues)
```

MyPy:
```
src/util/base_game.py:4: error: Module "pygame.event" has no attribute "EventType"; maybe "_EventTypes"?
```

ankith26:
Thanks for reporting the issue, nice catch! IMHO I think going forward, `EventType` should be named just `Event`, with the former being a deprecated alias.

RN, `Event` is merely a function that returns an instance of `EventType`, but the typehints indicate `Event` to be a class in itself (AKA `EventType` itself)

This might also apply to a few other classes in pygame codebase

Starbuck5:
Complete (hopefully) list of "function classes"
```
Event
Clock
some _camera.Camera stuff?
Joystick
```

ankith26:
[...]

PR #3001, opened 2022-01-23
Add `__new__` and `__init__` functions for some types

zoldalma999:
As per  #2867, I added `__init__` and/or  `__new__` to the following classes:
 - Event
 - Clock
 - Channel

I also documented that Joysticks do not have these functions and can only be opened trough the `pygame.joystick.Joystick` function.
PixelArray seemed to have already exported it's class properly. I did not want to touch the Camera class, since I do not have a camera to test with.


docs/reST/ref/event.rst
```diff
@@ -431,6 +431,8 @@ Most these window events do not have any attributes, except ``WINDOWMOVED``,
    Creates a new event with the given type and attributes. The attributes can
    come from a dictionary argument with string keys or from keyword arguments.
 
+   .. versionchanged:: 2.1.3 Event works the same way ``EventType`` does.
+
    .. ## pygame.event.Event ##
 
 .. class:: EventType
```
ankith26, line 438: It would be nicer if the docs for `Event` and `EventType` were unified here, and a note added that `EventType` is an alias for `Event` since 2.1.3

src_c/mixer.c
```diff
@@ -1013,10 +1013,9 @@ static PyTypeObject pgSound_Type = {
     0,             /* tp_dictoffset */
     sound_init,    /* tp_init */
     0,             /* tp_alloc */
-    0,             /* tp_new */
+    PyType_GenericNew,    /* tp_new */
```
ankith26, line 1016: It's fine if it's a `0` here, `tp_new` is `PyType_GenericNew` by default anyways

src_c/event.c
```diff
@@ -1488,6 +1488,69 @@ pg_event_richcompare(PyObject *o1, PyObject *o2, int opid)
     return Py_NotImplemented;
 }
 
+static int
+pg_event_populate(pgEventObject *event, int type, PyObject *dict)
```
Starbuck5, line 1492: `_pg_event_populate` if it's an internal function?

Starbuck5:
Why not make Joystick a class like the rest? Seems simpler and prettier than dealing with JoystickTypes all over the place.

Would it be better for Clock to have an `__init__` and a generic `tp_new` rather than a "new" factory function, like `Event` and `Channel`?

To get rid of the failing setup.py lint CI, you can run `setup.py format`, which black and clang-format installed.

zoldalma999:
I got my hands on a linux laptop with a builtin webcam, so I did that one as
well. I was still not able to test it on windows, so it would be nice if
someone could double check if everything is working as it should.

> Why not make Joystick a class like the rest? Seems simpler and prettier than
> dealing with JoystickTypes all over the place.

Because the joystick function does not always always construct a new object, but
might return an already created one. I was not sure how to implement that,
without changing a lot of code, and possibly breaking some things.


> Would it be better for Clock to have an `__init__` and a generic tp_new
> rather than a "new" factory function, like Event and Channel?

You are right, I changed it.


> To get rid of the failing setup.py lint CI, you can run setup.py format,
> which black and clang-format installed.

Ah. Yeah, I forgot to do it again.

ankith26 (approved):
LGTM! Thanks for the PR! :+1: 🎉 
(PS: lint is still failing 😅)

Starbuck5:
> Because the joystick function does not always always construct a new object, but
might return an already created one. I was not sure how to implement that,
without changing a lot of code, and possibly breaking some things.

We could override `__new__` to return the new or old instance? 

I don't want to merge this in with the current JoystickType documentation changes, because it makes it seem more complex and exposes that type as part of our API, when we might not want that.

I'd prefer the joystick status quo.

Everything else looks good, but I'll be sure to test webcam on Windows.

For the failing format status, you can rebase the branch against main. It's because black updated, and so now all the in progress branches will potentially get a few unrelated changes. But novial put through a PR that fixes it in main, if you can rebase.

ankith26:
I don't really prefer status quo, I prefer things documented as accurate to the implementation as possible. We could also investigate implementing `Joystick.__new__`, I believe if the type is immutable, there should be no problem if `__new__` returns a new reference to an existing object

Starbuck5:
@ankith26 I have to confess my objection originally stems from it making the docs ugly.

But we haven't gotten any issues with the status quo, unlike the Event status quo which kicked off this PR.

If we haven't decided to go with JoystickType or turn Joystick into a class of its own, it shouldn't be merged with everything else in this PR. I believe everything else is good.

illume:
Sorry, there's a conflict now with _camera.c because I merged another PR.

Starbuck5 (approved):
Thanks!

This is one of those things that future contributors will never think of but will help them out. Because of course the objects are objects, not functions. Why would they be functions? 😄 

One sorta suspicious thing I found is that PyChannel_New is still around and used, and the new channel_init has copied the same routines. Just in case one of them is changed in the future, that could trip someone up. Maybe add a comment? What do you think?

Is Event() now the preferred way? Should the docs assert that?

I've tested this locally and it works.

Anyways, this PR is good to go code wise, AFAICT, there's a couple things to think about, but it could be merged without that. There's also conflicts now (probably from my PRs sorry)

src_c/mixer.c
```diff
@@ -1918,16 +1926,14 @@ pgSound_New(Mix_Chunk *chunk)
...
-    chanobj = PyObject_New(pgChannelObject, &pgChannel_Type);
-    if (!chanobj)
+    pgChannelObject *chanobj = PyObject_New(pgChannelObject, &pgChannel_Type);
+    if (!chanobj) {
         return NULL;
-
-    chanobj->chan = channelnum;
+    }
+    if (channel_init(chanobj, Py_BuildValue("(i)", channelnum), NULL)) {
+        PyObject_Free(chanobj);
```
ankith26, line 1934: 2 things here, the return of `Py_BuildValue` functioncall is leaked here, because `channel_init` only borrows a reference. Secondly, IG `Py_DECREF` should be used instead of `PyObject_Free` here, because that is the most commonly used API while cleaning a python object

I suggest having a `_channel_init` that takes a C int, so that both `channel_init` and `pgChannel_New` can call it

PR #3006, opened 2022-01-26
Port `SDL_RenderGeometry`; rendering triangles and quads with `_sdl2.Renderer` and drawing from `_sdl2.Texture`

avaxar:
I figure to try actually contributing to pygame and not fix minor things. I ported the relatively new `SDL_RenderGeometry` SDL 2.0.18 feature and implemented it as `pygame._sdl2.Texture.draw_at` and `pygame._sdl2.Renderer.polygon` (The method names might need to be changed). However, `pygame._sdl2.Texture.draw_at` could only render at vertices of quads and `pygame._sdl2.Renderer.polygon` currently cannot be fed with per-vertex color (Uses solid color, set by the renderer draw color). I haven't tested it, as I'm using my Linux Mint which only has SDL 2.0.16, therefore unable to use `SDL_RenderGeometry`. Feel free to test it and find if there are any code mistakes/bugs it may have.

TL;DR Edits:
- No longer does it render polygons, but only triangles and quads.
- `Texture.draw_at` is split into `.draw_triangle` and `.draw_quad`
- `Renderer.polygon` is split into `.fill_triangle` and `.fill_quad`
- Additions: `Renderer.draw_triangle` and `Renderer.draw_quad`
- A minor change of making all `res` variables in `video.pyx` `cdef`'d as `int`s for compilation optimization


avaxar:
Oh right. I forgot to make the RST documentation and there are no unittests for it yet.

Starbuck5:
I haven't tested it yet, but I want you to be able to test it as you develop it.

I personally almost never write code that works the first time. And it'll be more fun for you to see polygons on the screen.

Idk Linux stuff, but you can't update with a package manager? Or you could just build SDL from source with ./configure, make, and make install.

avaxar:
Since @Mega-JC told me on Discord that `SDL_RenderGeometry` doesn't render polygons, rather vertices sets of triangles.. I changed the methods a bit.

- `Texture.draw_at` -> `.draw_triangle` and `.draw_quad` (Only renaming for the latter)
- `Renderer.draw_polygon` -> `.draw_triangle` and `.draw_quad`
- `Renderer.fill_polygon` -> `.fill_triangle` and `.fill_quad`

avaxar:
After multiple force pushings and painful debugging, finally I got it to work. Here are all of the methods in action:
image

The used code:
```py
import pygame
import pygame._sdl2


pygame.init()

window = pygame._sdl2.Window("amogus", (1066, 600))
renderer = pygame._sdl2.Renderer(window, vsync=True)
surface = pygame.image.load("avaxar.png")
texture = pygame._sdl2.Texture.from_surface(renderer, surface)

running = True
while running:
        if event.type == pygame.QUIT:
            running = False
            break

    renderer.draw_color = (0, 0, 0, 255)
    renderer.clear()

    renderer.draw_color = (255, 255, 255, 255)
    renderer.fill_triangle((50, 50), (200, 100), (50, 100))
    renderer.fill_quad((250, 50), (500, 50), (600, 400), (200, 400))

    renderer.draw_color = (255, 0, 0, 255)
    renderer.draw_triangle((50, 150), (200, 200), (50, 200))
    renderer.draw_quad((250, 150), (500, 150), (600, 500), (200, 500))

    texture.draw_triangle((700, 400), (900, 400), (700, 550))
    texture.draw_quad((600, 100), (900, 100), (1000, 300), (700, 300))

    renderer.present()
```

avaxar:
Sorry about the mess in the chronology of commits w/ the comments in this PR :eyes:
I force-pushed to clean the commit history of this PR.

avaxar:
Oh shoot, what have I done to the commit history. I was trying to get rid of the merge conflicts for the Cython generated C files and docs by PR #2978 getting merged to main.

Starbuck5:
> Oh shoot, what have I done to the commit history. I was trying to get rid of the merge conflicts for the Cython generated C files and docs by PR https://github.com/pygame/pygame/pull/2978 getting merged to main.

I've done this several times accidentally. And I've handled it by closing the PR and porting my changes to new a branch, and then opening a new PR. There's probably a smarter way to handle it but this way works.

ankith26:
Okay, I fixed up the commit mess in this PR itself.

docs/reST/ref/sdl2_video.rst
```diff
@@ -183,6 +183,16 @@
       | :sl:`Copy a portion of the texture to the rendering target.`
       | :sg:`draw(srcrect=None, dstrect=None, angle=0, origin=None, flip_x=False, flip_y=False) -> None`
 
+   .. method:: draw_triangle
+      
+      | :sl:`Copy a triangle portion of the texture to the rendering target by vertices coordinates.`
+      | :sg:`draw_triangle(p1_xy, p2_xy, p3_xy, p1_uv=(0.0, 0.0), p2_uv=(1.0, 0.0), p3_uv=(1.0, 1.0), p1_mod=(255, 255, 255, 255), p2_mod=(255, 255, 255, 255), p3_mod=(255, 255, 255, 255)) -> None`
```
dr0id, line 189: Is the order of the p1_uv, p2_uv and p3_uv relevant? Because its not the same as on line 773 in video.pyx

dr0id:
> ```python
>  texture.draw_triangle((700, 400), (900, 400), (700, 550))
> ```

Looking at the image I would guess its the triangle on the bottom right. That looks a bit distorted.

On second look at the method, the default uv points seem to be strange, but I don't know the axis (I would have assumes like normal screen coordinates, but I might be wrong, e.g. (0,0) at topleft, (1,0) topright, (1,1) bottomright, (0,1) bottomleft).

docs/reST/ref/sdl2_video.rst
avaxar: Oh, I didn't realize that. I've made a fix commit.

avaxar:
Yeah, it is distorted. It's an SDL specific thing, I reckon though.

docs/reST/ref/sdl2_video.rst
```diff
@@ -183,6 +183,16 @@
...
 
+   .. method:: draw_triangle
+      
+      | :sl:`Copy a triangle portion of the texture to the rendering target by vertices coordinates.`
+      | :sg:`draw_triangle(p1_xy, p2_xy, p3_xy, p1_uv=(0.0, 0.0), p2_uv=(1.0, 1.0), p3_uv=(0.0, 1.0), p1_mod=(255, 255, 255, 255), p2_mod=(255, 255, 255, 255), p3_mod=(255, 255, 255, 255)) -> None`
+
+   .. method:: draw_quad
+      
+      | :sl:`Copy a quad portion of the texture to the rendering target by vertices coordinates.`
+      | :sg:`draw_quad(p1_xy, p2_xy, p3_xy, p4_xy, p1_uv=(0.0, 0.0), p2_uv=(1.0, 0.0), p3_uv=(1.0, 1.0), p4_uv=(0.0, 1.0), p1_mod=(255, 255, 255, 255), p2_mod=(255, 255, 255, 255), p3_mod=(255, 255, 255, 255), p4_mod=(255, 255, 255, 255)) -> None`
```
ankith26, line 194:
```suggestion
      | :sg:`draw_quad(xy, uv=((0.0, 0.0), (1.0, 0.0), (1.0, 1.0),(0.0, 1.0)), mod=((255, 255, 255, 255),) * 4) -> None`
```
The function signatures here could look something like this. The `draw_triangle` function could also be updated similarly

dr0id:
> Yeah, it is distorted. It's an SDL specific thing, I reckon though.

My question was more like: is the distortion correct? Using the correct UV coordinates?

dr0id:
> > Yeah, it is distorted. It's an SDL specific thing, I reckon though.
> 
> My question was more like: is the distortion correct? Using the correct UV coordinates?

image
Maybe I don't understand it: why is it distorted this way?
In the sample code ` texture.draw_triangle((700, 400), (900, 400), (700, 550))`  you used `texture.draw_triangle(V1, V2, V3)` and this is like a mapping 
```
T0 -> V1 
T3 -> V2 
T2 -> V3
```
Maybe this is arbitrary or I don't understand how it works. I hope my notation is not confusing.

avaxar:
By the way, in your drawing, T2 and T3 are at the wrong vertices; T2's at T3's place, and vice versa. The default args for the UV coordinates in `draw_triangle` are `p1_uv=(0.0, 0.0), p2_uv=(1.0, 1.0), p3_uv=(0.0, 1.0)`, with the code drawing the vertices at `(700, 400), (900, 400), (700, 550)` or can be seen as `(0, 0), (1, 0), (0, 1)`

Thus, the example code draws it as:
UV -> XY
(0, 0) -> (0, 0)
(1, 1) -> (1, 0)
(0, 1) -> (0, 1)

dr0id:
> default args for the UV coordinates in ``

Thanks for the explanation!

MyreMylar:
Looks like this needs typestubs and that conflict fixing.

ankith26:
pushed this PR to 2.1.4 milestone because this needs more work and discussion

docs/reST/ref/sdl2_video.rst
avaxar: Anyway, I'm back after 6 months. I don't think I agree with this change. `pygame.draw.line` takes seperate `start_pos` and `end_pos` arguments; and triangles and quads have exact amounts of points anyway, unlike something like `pygame.draw.polygon`.

PR #3013, opened 2022-02-02
New black formatting

novialriptide:


MyreMylar (approved):
LGTM 👍  I get the same results running black locally with the latest version

Starbuck5 (approved):
*No description provided*

Just joking.

For context this is so that other PRs don't get these unnecessary diffs, since black has updated and changed some behavior.

PR #3014, opened 2022-02-03
C code cleanups + cdrom/overlay removal

Starbuck5:
I was browsing the python mailing list, as you do, and I found [post](https://mail.python.org/archives/list/python-dev@python.org/thread/DN6JAK62ZXZUXQK4MTGYOFEC67XFQYI5/) about the evolution of the C API and the limited ABI.

And Victor Stinner, the author, provided a [tool ](https://github.com/pythoncapi/pythoncapi_compat) that could do some upgrades automatically.

So I ran that, picked out the changes I wanted (I didn't feel like putting the new in 3.10 `Py_Is` functions in our compatibility layer).

The benefits of these changes are that it makes our code a bit more standard and modern.

While I was at it I finally removed the cdrom and overlay code that's been sitting 100% unused since 2.1 removed SDL1 compatibility.


ankith26 (approved):
Good cleanups, thanks! 👍 🥳

src_c/_pygame.h
```diff
@@ -319,7 +315,6 @@ struct pgColorObject {
 #define PYGAMEAPI_PIXELARRAY_NUMSLOTS 2
 #define PYGAMEAPI_COLOR_NUMSLOTS 5
 #define PYGAMEAPI_MATH_NUMSLOTS 2
-#define PYGAMEAPI_CDROM_NUMSLOTS 2
```
illume, line 322: Does this need to reflect the number of slots used somewhere else?

ankith26 (commented):
While on the topic of C-API cleanups, the linked library provides a compatibility header, would it be vendoring this header into pygame? I imagine it might be useful as more python 3.x releases roll in, we can make use of newer API quicker
Header here: https://github.com/pythoncapi/pythoncapi_compat/blob/main/pythoncapi_compat.h

illume:
That compatibility header fatally is flawed. Because reasons... (sorry I didn't have time to make this shorter/clearer).

For compatibility, we had that issue of forwards/backwards compat. If we have something making defines, and compiling with an old version of the library... then when someone uses a new library there is a symbol crash. So it's better to use a new symbol instead.

Example to try and make it clear:

- pygame main compiled with SDL 2.0.2
- user drops in SDL 2.0.20 -> crash! because we defined a symbol which was missing in 2.0.2.

src_c/_pygame.h
Starbuck5: You'd think so, but it must've been removed already. I used `grep` to check the codebase for extraneous CDROM mentions.

illume:
Sorry, I guess the conflicts are in here because I merged that big warnings C PR.

illume (approved):
👍 🎉 thanks heaps

Issue #3007, opened 2022-01-29
Rect, but with python numbers not ints (draft proposal)


illume:
## Description

Users like python numbers. SDL2 now has float using draw calls, and sub-pixels using floats are a thing. Why not float rect? Because python users like python numbers, and python numbers can be ints or floats (and more!).

## Demo of Python number using Rects (PRect)

Here we see a rect being defined as an using python numbers which are ints.

```python
>>> r = Rect(1,2,3,4)
>>> r.x
1
>>> r.x += 0.1
>>> r.x
1.1
>>> r
<Rect(1.1, 2, 3, 4)>
>>> r.x = 1
>>> r
<Rect(1, 2, 3, 4)>
>>> r.y
2
>>> r.y /= 2
>>> r.y
1
>>> r.y = 2
>>> r.y /= 3
>>> r.y
0.6666666666666666
>>> list(map(type, r))
[int, float, int, int]
```

As we can see it follows python number behaviour. When we first assign it as an int, then it's an int. If we divide and it goes to a float, then it's a float. Just like python numbers do.

## To explore

We need to explore a lot of different areas. Because rect is use EVERYWHERE pretty much.

### Gradual adoption, feature flat, and conversion

It's not possible to convert the whole code base to by python-number-rect aware (PRect).
Also things are likely to be buggy to start with. For these reasons we could hide the new implementation behind a feature flag.
Perhaps PYGAME_PYTHON_NUMBER_RECTS environment variable for example.

- PRect implementation
- converting code to be PRect aware
- using both PRect and Rect at the same time

Perhaps PRect being a subclass of Rect would make things easier.

Maybe we discover users want an IntRect (like we have now... Rect is an IntRect) or a FloatRect.

### Backwards compatibility

It may be that "PRect" breaks backwards compatibility in a major way. I expect that it won't however. Because if you assign ints to rect, then it will try and use ints. Additionally, already many of the other pygame functions accept floats and "rect like" lists of floats. It may be that we could keep an integer mode for these cases.

### Speed

It could be that this way is slower. If it is, we could perhaps offset that with speed improvements to the rect class. For example, rects do not use a freelist yet, and it has been proven to be a 

Another aspect could be allowing the use of pure python implementations on things like PyPy.

### Prior art

There are several implementations of python based rect classes. Either so people can use floats, but also so that they are less surprising to newbies (because they act much like other python numbers).

Starbuck5:
[...]

PR #3017, opened 2022-02-04
added a FRect object

Emc2356:
created a new object that is similar to pygame.Rect but instead of using integers pygame.FRect uses floats

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -310,7 +310,7 @@ struct pgColorObject {
  * Remember to keep these constants up to date.
  */
 
-#define PYGAMEAPI_RECT_NUMSLOTS 5
+#define PYGAMEAPI_RECT_NUMSLOTS 10
 #define PYGAMEAPI_JOYSTICK_NUMSLOTS 2
 #define PYGAMEAPI_DISPLAY_NUMSLOTS 2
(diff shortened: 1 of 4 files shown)


Starbuck5:
This is related to #3007, but uses floats instead of python numbers.

It looks like there aren't any tests, so I'm not sure how compliant this is with the Rect api.

Perhaps it would be better to do FRect in a different C file, since there doesn't seem to be much code overlap. (So just save the future readers from ginormous files). Alternatively, is there a way to share more code between implementations? Like a bootleg C++ template system in C with macros or something?

If we're going to do it this way, I'd like it to be based on SDL_FRect. I know there aren't docs, but you can check the SDL source (Rect.h header would have it?). And if version added is a concern, you can use `#ifndef` right?

novialriptide:
Any reason why Pygame.Rect isn't turned into FRect?

Starbuck5:
> Any reason why pygame.Rect isn't turned into FRect?

Backwards compatibility.

mcpalmer1980:
As I suggested earlier, I like this idea of a separate FRect that won't effect any legacy code. Users could either add FRects were necessary (sdl2.video drawing/scaling) and use Rects everywhere else. Or, if they prefer, they can add something like import FRect as Rect to make it global.

PR #3022, opened 2022-02-07
Fix there typo in tests.rst

eoghanriley:
tests.rst says " so frozen tests are killed when there time allotment expired." and I fixed it to say "their" instead.


ankith26 (approved):
LGTM! Thanks for the PR :) 🎉

MightyJosip (approved):
Nice one. Thanks 👍

PR #3026, opened 2022-02-12
Proposal: add Rect.collidelistallobjects that returns the colliding objects 

dr0id:
There are already Rect.collidelist and Rect.collidelistall which return a index or a list of indices. 
The new proposed methods would return the object itself (or None) or the list of objects it collides. 
As a bonus feature there is also a key function that can be passed, allowing to collide against a list of objects that are not a Rect or have a rect attribute. There are cases where an object can have multiple rects (like draw rect, collide rect, etc.) where this can be handy.


buildconfig/pygame-stubs/_common.pyi
```diff
@@ -31,4 +31,7 @@ _CanBeRect = Union[
 class _HasRectAttribute(Protocol):
     rect: _CanBeRect
 
-_RectValue = Union[_CanBeRect, _HasRectAttribute]
+class _HasCallableRectAttribute(Protocol):
```
dr0id, line 34: Is this the right way to express a type hint for all the possible ways the collision methods can accept?

src_c/rect.c
```diff
@@ -695,6 +695,132 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
...
+                     "Key function must be callable with one argument.");
+    }
+
+    ret = PyList_New(0);
+    if (!ret) {
+        return NULL;
+    }
+
+    size = PySequence_Length(list); /*warning, size could be -1?*/
+    for (loop = 0; loop < size; ++loop) {
```
dr0id, line 752: Please keep a good eyes on the ref counting, not sure if I did it correctly everywhere (executing this method in a loop did not reveal any memory increase as a simple test... but that might not be very accurate)

test/rect_test.py
```diff
@@ -1359,16 +1357,6 @@ def test_unionall_ip(self):
         # Bug for an empty list. Would return a Rect instead of None.
         self.assertTrue(r1.unionall_ip([]) is None)
 
-    def test_unionall__invalid_rect_format(self):
```
dr0id, line 1362: test_unionall__invalid_rect_format()  removing duplication of this method

docs/reST/ref/rect.rst
```diff
@@ -357,6 +357,44 @@
...
+      | :sg:`collideobjects(obj_list, key=func) -> object`
+
+      Test wheter the rectangle collides with any object in the sequence.
+      The object of the first collision found is returned. If no collisions are
+      found then ``None`` is returned
+
+      If key is given, then it should be a method taking an object from the list
+      as input and returning a rect like object e.g. ``lambda obj: obj.rectangle``.
+      If an object has multiple attributes of type Rect then key could return one
+      of them.
```
dr0id, line 373: Note to self: add example code (?)

buildconfig/pygame-stubs/_common.pyi
ankith26: I'd suggest leaving out this typing fix in this PR, I've already fixed this in #2999 a few days back :)
your fix will conflict with my fix

dr0id: yes, especially since you have moved the file to a different location if I have interpreted your pr correctly.

ankith26: Yeah. And the fix I make is the same fix made here

dr0id: reverted the changes except the typo... should I revert the changes in rect.pyi too?

illume:
Cool! It makes more sense to return collided objects.

ankith26 (changes requested):
I reviewed on the typehints and I have requested some changes. I believe the `rect.pyi` typing can be simplified to 
```py
    def collideobjectsall(
        self,
        obj_list: Sequence[_T],
        key: Optional[Callable[[_T], RectValue]],
    ) -> List[_T]: ...
    def collideobjects(
        self,
        obj_list: Sequence[_T],
        key: Optional[Callable[[_T], RectValue]],
    ) -> Optional[_T]: ...
```
Without any additional overloads. `_T = TypeVar("_T")` should come somewhere at the top of the codebase

src_c/rect.c
```diff
@@ -695,6 +695,142 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
...
+    Py_ssize_t size;
+    int loop;
+    PyObject *list, *obj;
+    PyObject *keyfunc = NULL;
+    PyObject *ret = NULL;
+    static char *keywords[] = {"list", "key", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|$O:collideobjectsall",
+                                     keywords, &list, &keyfunc)) {
+        return RAISE(PyExc_TypeError, "Could not parse args and kw.");
```
ankith26, line 727: A `return NULL;` is needed here because `PyArg_ParseTupleAndKeywords` already sets a more informative error message

src_c/rect.c
```diff
@@ -695,6 +695,142 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
...
+        return RAISE(PyExc_TypeError,
+                     "Key function must be callable with one argument.");
+    }
+
+    ret = PyList_New(0);
+    if (!ret) {
+        return NULL;
+    }
+
+    size = PySequence_Length(list); /*warning, size could be -1?*/
```
ankith26, line 749:
```suggestion
    size = PySequence_Length(list);
    if (size == -1) {
        return NULL;
    }
```

src_c/rect.c
```diff
@@ -695,6 +695,142 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
...
+    for (loop = 0; loop < size; ++loop) {
+        obj = PySequence_GetItem(list, loop);
+
+        if (!obj) {
+            Py_XDECREF(obj);
+            Py_DECREF(ret);
+            return RAISE(PyExc_TypeError, "Get item from sequence failed.");
```
ankith26, line 756: This should just be
```py
Py_DECREF(ret);
return NULL;
````
Because `obj` is already `NULL` here, and `PySequence_GetItem` already sets python exception

File: src_c/rect.c [hunk 8]
```diff
@@ -695,6 +695,142 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
...
+        }
+
+        if (!(argrect = pgRect_FromObjectAndKeyFunc(obj, keyfunc))) {
+            Py_XDECREF(obj);
+            Py_DECREF(ret);
+            return RAISE(
+                PyExc_TypeError,
+                "Item in sequence or returned by key function must be Rect.");
```
ankith26, line 764: This should be 
```py
Py_DECREF(obj);
Py_DECREF(ret);
return NULL;
```
Because python error is already set

src_c/rect.c
```diff
@@ -695,6 +695,142 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
...
+            }
+        }
+        Py_DECREF(obj);
+    }
+
+    return ret;
+}
+
+static PyObject *
+pg_rect_collideobjects(pgRectObject *self, PyObject *args, PyObject *kwargs)
```
ankith26, line 781: this functions has some mistakes, similar to the above function which I elaborated on in the reviews

src_c/rect.c
ankith26: There don't seem to be memleaks here, but there are error handling bugs, which I have reviewed below

buildconfig/stubs/pygame/__init__.pyi
```diff
@@ -633,4 +632,3 @@ from .constants import (
     WINDOWSIZECHANGED as WINDOWSIZECHANGED,
     WINDOWTAKEFOCUS as WINDOWTAKEFOCUS,
 )
-
```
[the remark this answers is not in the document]
dr0id: > minor nitpick, It'd be better if this file is not modified by hand because it is generated by a script. So the next time someone runs the script, it'd generate diffs again when these linebreaks are re-added

updated generator to generate black conform files

buildconfig/stubs/pygame/_common.pyi
```diff
@@ -34,4 +34,7 @@ class _HasRectAttribute(Protocol):
     # that returns a rect confirms to the rect protocol
     rect: Union[RectValue, Callable[[], RectValue]]
 
-RectValue = Union[_CanBeRect, _HasRectAttribute]
+class _HasCallableRectAttribute(Protocol):
+    rect: Callable[[], RectValue]
+
+RectValue = Union[_CanBeRect, _HasRectAttribute, _HasCallableRectAttribute]
```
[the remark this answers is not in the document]
dr0id: you are right, removed _HasCallableRectAttribute

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -147,6 +147,43 @@ class Rect:
     ) -> bool: ...
     def collidelist(self, rect_list: Sequence[RectValue]) -> int: ...
     def collidelistall(self, rect_list: Sequence[RectValue]) -> List[int]: ...
+
+    class _T(Protocol):
```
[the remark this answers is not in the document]
dr0id: thanks, learning something new

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -147,6 +147,43 @@ class Rect:
...
+    def collideobjectsall(
+        self, rect_list: Sequence[Union[Rect, RectValue]], key: None = ...
+    ) -> List: ...
+    # given a list of rect like objects, still could use a key to extract a different rect
+    @overload
+    def collideobjectsall(
+        self,
+        rect_list: Sequence[Union[Rect, RectValue]],
+        key: Callable[[Union[Rect, RectValue]], Union[Rect, RectValue]],
+    ) -> List: ...
```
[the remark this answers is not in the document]
dr0id: ok, I see, removed

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -147,6 +147,43 @@ class Rect:
...
+    def collideobjects(
+        self, rect_list: Sequence[Union[Rect, RectValue]], key: None = ...
+    ) -> Union[Rect, RectValue]: ...
+    # given a list of rect like objects, still could use a key to extract a different rect
+    @overload
+    def collideobjects(
+        self,
+        rect_list: Sequence[Union[Rect, RectValue]],
+        key: Callable[[Union[Rect, RectValue]], Union[Rect, RectValue]],
+    ) -> Union[Rect, RectValue]: ...
```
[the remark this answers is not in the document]
dr0id: removed

src_c/rect.c
dr0id: alright, changed

src_c/rect.c
dr0id: removed

src_c/rect.c
dr0id: ok, learning more and more, thanks

src_c/rect.c
dr0id: same mistake

src_c/rect.c
dr0id: ok, corrected as hinted in the other method but now I get this when I run the tests:

```
ERROR: test_collideobjects_call_variants (__main__.RectTypeTest)
  File "C:\Data\Programs\python390\lib\site-packages\pygame-2.1.3.dev3-py3.9-win-amd64.egg\pygame\tests\rect_test.py", line 2169, in test_collideobjects_call_variants
    self.assertRaises(TypeError, r.collideobjects, objects)
  File "C:\Data\Programs\python390\lib\unittest\case.py", line 731, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "C:\Data\Programs\python390\lib\unittest\case.py", line 201, in handle
    callable_obj(*args, **kwargs)
SystemError: <built-in method collideobjects of pygame.Rect object at 0x00000220D0F0C480> returned NULL without setting an error

```

src_c/rect.c
dr0id: well, if I return NULL here then I get those tests to fail. Changed it back to RAISE(...). But maybe I'm doing it wrong in the `pgRect_FromObjectAndKeyFunc` method. Please review this method too. Thanks for your time.

dr0id:
@ankith26 thanks for your review. I have updated according to your feedback. Please resolve your comments if you agree with the change(s) or feel free to leave other comments.

buildconfig/stubs/pygame/__init__.pyi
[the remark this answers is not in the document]
ankith26: That's a good improvement!

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -147,6 +158,12 @@ class Rect:
     ) -> bool: ...
     def collidelist(self, rect_list: Sequence[RectValue]) -> int: ...
     def collidelistall(self, rect_list: Sequence[RectValue]) -> List[int]: ...
+    def collideobjectsall(
+        self, objects: Sequence[Any], key: Optional[Callable[[Any], RectValue]] = None
+    ) -> List[Any]: ...
+    def collideobjects(
+        self, objects: Sequence[Any], key: Optional[Callable[[Any], RectValue]] = None
```
ankith26, line 165: I'd still prefer if

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -147,6 +158,12 @@ class Rect:
...
+    ) -> Any: ...
```
ankith26, line 166: I'd still prefer if generics are used here, instead of `Any`. The key benefit of that is that it checks for complete compatibility of the function call, as in something like
`collided = pygame.rect.collideobjects([1, 4, 5], key=lambda x: (x.lower(), x.lower(), 100, 100))`

With the generics, it would know that `x` in the lamba function *has* to be an integer, and it would correctly show a warning for calling `lower()` at analysis. Likewise, with generics, the typechecker also knows the precise type of the return value of the function. IE. it will return the type, which `objects` was a sequence of the same type

Check out the codeblock I gave in my first review :)

src_c/rect.c
ankith26: Yeah, the problem lies in `pgRect_FromObjectAndKeyFunc`. I'll review some changes to be made there, and this bit of code should do a `return NULL;` again

src_c/rect.c
```diff
@@ -695,6 +695,141 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
     return ret;
 }
 
+static SDL_Rect *
+pgRect_FromObjectAndKeyFunc(PyObject *obj, PyObject *keyfunc)
+{
+    SDL_Rect temp;
+    if (keyfunc == NULL) {
+        return pgRect_FromObject(obj, &temp);
```
ankith26, line 703:
```suggestion
        SDL_Rect *ret = pgRect_FromObject(obj, &temp);
        if (!ret) {
            PyErr_SetString(
                PyExc_TypeError, "Sequence must contain rect or rect-like objects");
        }
        return ret;
```
The problem here is that `pgRect_FromObject` does not set a python error when it fails, and returns `NULL`. So we have to set python error here, so that the end result is that `pgRect_FromObjectAndKeyFunc` always sets python error on fail

src_c/rect.c
```diff
@@ -695,6 +695,141 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
...
+    }
+
+    PyObject *obj_with_rect = PyObject_CallFunctionObjArgs(keyfunc, obj, NULL);
+    if (obj_with_rect == NULL) {
+        return NULL;
```
ankith26, line 708: This branch is good, `PyObject_CallFunctionObjArgs` would have set python error already

src_c/rect.c
```diff
@@ -695,6 +695,141 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
...
+    }
+
+    return pgRect_FromObject(obj_with_rect, &temp);
```
ankith26, line 711:
```suggestion
    SDL_Rect *ret = pgRect_FromObject(obj_with_rect, &temp);
    Py_DECREF(obj_with_rect);
    if (!ret) {
        PyErr_SetString(
            PyExc_TypeError, "Key function must return rect or rect-like objects");
    }
    return ret;
```
Similar thing as above review, but with another small fix suggestion, `obj_with_rect` is a "new reference" returned by `PyObject_CallFunctionObjArgs`, so after it's used in the function, it must be `Py_DECREF`d

src_c/rect.c
dr0id: changed

src_c/rect.c
dr0id: changed

src_c/rect.c
dr0id: changed `pgRect_FromObjectAndKeyFunc`

src_c/rect.c
dr0id: this is resolved now

src_c/rect.c
```diff
@@ -695,6 +695,147 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
...
+        return RAISE(PyExc_TypeError,
+                     "Key function must be callable with one argument.");
+    }
+
+    size = PySequence_Length(list);
+    for (loop = 0; loop < size; ++loop) {
+        obj = PySequence_GetItem(list, loop);
+
+        if (!obj) {
+            Py_XDECREF(obj);
```
ankith26, line 821: This line is pretty redundant here because `obj` is already `NULL`

src_c/rect.c
ankith26: I don't think this has been fixed fully yet, if `size == -1` it is an indication that `PySequence_Length` failed, and the function must exit and forward error

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -147,6 +158,13 @@ class Rect:
     ) -> bool: ...
     def collidelist(self, rect_list: Sequence[RectValue]) -> int: ...
     def collidelistall(self, rect_list: Sequence[RectValue]) -> List[int]: ...
+    T = TypeVar("T")
```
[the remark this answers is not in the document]
dr0id: ok ,moved to the other type vars

src_c/rect.c
```diff
@@ -695,6 +695,147 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *args)
...
+        }
+        return ret;
+    }
+
+    PyObject *obj_with_rect = PyObject_CallFunctionObjArgs(keyfunc, obj, NULL);
+    if (obj_with_rect == NULL) {
+        return NULL;
+    }
+
+    SDL_Rect *ret = pgRect_FromObject(obj_with_rect, &temp);
```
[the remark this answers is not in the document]
dr0id: ah, right, thanks

src_c/rect.c
dr0id: ok, I see, removed

src_c/rect.c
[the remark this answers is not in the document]
dr0id: So in general I can assume that when I get a pointer then the reference count has been increased?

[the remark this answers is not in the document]
ankith26: In general, functions that return python objects return "strong references", which means that it is up to the caller to `DECREF` a reference when the caller does not need it anymore. 

Functions that accept a python object as an argument generally "borrow references". This means that the caller function "holds a reference to the object", therefore the object should remain valid and accessible for the entire duration the subfunction is being called, which does not need to `DECREF` the borrowed reference because it does not "own that reference". In contrast to this, there can be some functions that "steal references" (usually not a recommended thing to do, but some python API does this). These are the kind of functions where a subfunction would `DECREF` its python argument, or put it to use elsewhere without `INCREF`ing it. This means that the caller essentially has "lost a reference" and now it's not guaranteed the object will remain a valid object after the function call. This can be easy worked around, the caller can `INCREF` an object before passing it to a function stealing a reference, so that the caller too holds a reference which is valid after the function call, if the caller needs to do anything with it later

When I refer to a "caller" anywhere above, I mean a function that is calling another function

src_c/rect.c
ankith26: Cool 👍

src_c/rect.c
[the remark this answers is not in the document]
ankith26: also about this line `Py_DECREF(obj_with_rect);`, it should be called regardless the error state, so this line needs moving to ouside the `if (!ret)` statement :)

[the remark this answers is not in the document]
dr0id: Thanks for the explanation. I know the concept of referencing, But to me it was not clear, where or which object is in control of the reference. I think you summed it up pretty well I guess it just takes a bit of practice or reading the documentation/source to understand who owns which reference.

src_c/rect.c
dr0id: I actually copied this part from another already existing method. This would mean this isn't  well implementer there either.

dr0id: updated as suggested

src_c/rect.c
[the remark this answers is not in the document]
dr0id: updated as suggested

dr0id:
> I reviewed on the typehints and I have requested some changes. I believe the `rect.pyi` typing can be simplified to
> 
> ```python
>     def collideobjectsall(
>         self,
>         obj_list: Sequence[_T],
>         key: Optional[Callable[[_T], RectValue]],
>     ) -> List[_T]: ...
>     def collideobjects(
>         self,
>         obj_list: Sequence[_T],
>         key: Optional[Callable[[_T], RectValue]],
>     ) -> Optional[_T]: ...
> ```
> 
> Without any additional overloads. `_T = TypeVar("_T")` should come somewhere at the top of the codebase

changed as suggested

dr0id:
@ankith26 Thanks for the hint about the rect on the stack. I think I have fixed it, but some build still fail (no clue why..) but the 32bit failures are gone.

dr0id:
> I reviewed on the typehints and I have requested some changes. I believe the `rect.pyi` typing can be simplified to
> 
> ```python
>     def collideobjectsall(
>         self,
>         obj_list: Sequence[_T],
>         key: Optional[Callable[[_T], RectValue]],
>     ) -> List[_T]: ...
>     def collideobjects(
>         self,
>         obj_list: Sequence[_T],
>         key: Optional[Callable[[_T], RectValue]],
>     ) -> Optional[_T]: ...
> ```
> 
> Without any additional overloads. `_T = TypeVar("_T")` should come somewhere at the top of the codebase

changed as suggested

dr0id:
@ankith26 ready to be reviewed again (not sure how to get rid of your requested changes otherwise). Thanks for your support.

ankith26 (approved):
Yeah this PR looks ready to me! Thanks for the PR 👍 🎉 
This is a useful feature indeed, the examples and tests are explanative and nice too

test/rect_test.py
```diff
@@ -2155,6 +2149,300 @@ def test_collidelistall_list_of_object_with_rect_property(self):
...
+        # act / verify
+        print("rects")
+        r.collideobjects(rects)
+        print("rects, key=None")
+        r.collideobjects(rects, key=None)
+        print("objects, key=lambda")
+        r.collideobjects(objects, key=lambda o: o.rect1)
+        print("assertRaises: objects")
+        self.assertRaises(TypeError, r.collideobjects, objects)
+
```
MyreMylar, line 2174: There are some debug print statements left in the code here.

src_c/doc/rect_doc.h
```diff
@@ -21,6 +21,8 @@
 #define DOC_RECTCOLLIDERECT "colliderect(Rect) -> bool\ntest if two rectangles overlap"
 #define DOC_RECTCOLLIDELIST "collidelist(list) -> index\ntest if one rectangle in a list intersects"
 #define DOC_RECTCOLLIDELISTALL "collidelistall(list) -> indices\ntest if all rectangles in a list intersect"
+#define DOC_RECTCOLLIDEOBJECTS "collideobjects(rect_list) -> object\ncollideobjects(obj_list, key=func) -> object\ntest if an objects in a list intersects"
```
MyreMylar, line 24: > test if an objects in a list intersects

Should this be 'object' here, or 'any' ? I think 'any' actually.

docs/reST/ref/rect.rst
```diff
@@ -401,6 +401,124 @@
 
       .. ## Rect.collidelistall ##
 
+   .. method:: collideobjects
+
+      | :sl:`test if an objects in a list intersects`
+      | :sg:`collideobjects(rect_list) -> object`
+      | :sg:`collideobjects(obj_list, key=func) -> object`
+
+      Test wheter the rectangle collides with any object in the sequence.
```
MyreMylar, line 410: > Test wheter the rectangle collides with any object in the sequence.

Should be 'Test **whether** the rectangle...'

MyreMylar (changes requested):
Looking pretty good. Have tested it locally and it all works, just a few minor spelling errors and some rogue debug prints.

test/rect_test.py
dr0id: removed, thanks for spotting

src_c/doc/rect_doc.h
dr0id: not sure, since I'm not a native English speaker I need some help: the method does the same as collidelist with the difference, that the list may contain objects and the object itself is returned instead of an index. Its like a FirstOrDefault filter. Could be worded like:

1. test if one object in a list intersects
2. test if any object in a list intersects

Both sound good to me, but may not to a native speaker.

dr0id: changing as suggested

docs/reST/ref/rect.rst
dr0id: thanks for spotting

MyreMylar (approved):
LGTM 👍

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -158,6 +171,12 @@ class Rect(Collection[int]):
     ) -> bool: ...
     def collidelist(self, rect_list: Sequence[RectValue]) -> int: ...
     def collidelistall(self, rect_list: Sequence[RectValue]) -> List[int]: ...
+    def collideobjectsall(
+        self, objects: Sequence[_T], key: Optional[Callable[[_T], RectValue]]
+    ) -> List[_T]: ...
+    def collideobjects(
+        self, objects: Sequence[_T], key: Optional[Callable[[_T], RectValue]]
```
ankith26, line 178: Ah this should take `key: Optional[Callable[[_T], RectValue]] = None` for the argument because `None` is the default value. Same with the other function

ankith26 (commented):
Was just about to merge this PR when I noticed another typehint issue

buildconfig/stubs/pygame/rect.pyi
dr0id: @ankith26 sorry for the late reply, mail about this notification must have gone unread with all the other changes that where made in the last 3 weeks (200+ notifications, you all were busy). Updated as suggested.

illume (commented):
🎉 

(
One thing I'm not sure of is if these support objects with a rect attribute without specifying a key function? Guess I could have just tried it... but it would probably would be good to know looking at the docs/tests. 

Probably this could be used within the sprite module to great effect. I think.
)

PR #3028, opened 2022-02-12
Save Surfaces with palettes as indexed 8P PNG

### LINKED ISSUE #646: Save Surfaces with palettes as indexed 8P PNG
Steps to fix this issue.

- [ ] first we need to write a unit test in [test/image_test.py](https://github.com/pygame/pygame/blob/master/test/image_test.py) (**good first issue**)
- [ ] fix the issue in [src_c/imageext.c](https://github.com/pygame/pygame/blob/master/src_c/imageext.c#L254) SavePNG function. (harder)
- [ ] try SDL 2.0.2 using the SDL_image save function rather than our own. (harder)

SDL_image 2.0.2 has this line:
` * Added libpng save support for much smaller 8-bit images`
http://hg.libsdl.org/SDL_image/log/acc936055688/IMG_png.c

There's a page on the topic here with some notes.
https://stackoverflow.com/questions/718933/how-to-save-an-sdl-surface-to-a-png-and-get-the-colors-right

**Related Docs**: https://www.pygame.org/docs/ref/image.html#pygame.image.save

ThanhChinhBK:
Fix #646
- [x] first we need to write a unit test in [test/image_test.py](https://github.com/pygame/pygame/blob/master/test/image_test.py) (good first issue)
- [x] fix the issue in [src_c/imageext.c](https://github.com/pygame/pygame/blob/master/src_c/imageext.c#L254) SavePNG function. (harder)
- [X] ~~try SDL 2.0.2 using the SDL_image save function rather than our own. (harder)~~ - The SDL function doesn't work.

### DIFF
--- a/src_c/imageext.c
+++ b/src_c/imageext.c
@@ -221,6 +221,7 @@ write_png(const char *file_name, SDL_RWops *rw, png_bytep *rows,
             color_ptr[i].blue = palette->colors[i].b;
         }
         png_set_PLTE(png_ptr, info_ptr, color_ptr, ncolors);
+        free(color_ptr);
     }
 
     /* doing = "write info"; */
--- a/src_c/imageext.c
+++ b/src_c/imageext.c
@@ -221,7 +221,7 @@ write_png(const char *file_name, SDL_RWops *rw, png_bytep *rows,
             color_ptr[i].blue = palette->colors[i].b;
         }
         png_set_PLTE(png_ptr, info_ptr, color_ptr, ncolors);
-        free(color_ptr);
+        SDL_free(color_ptr);
     }
 
     /* doing = "write info"; */
--- a/src_c/imageext.c
+++ b/src_c/imageext.c
@@ -208,12 +208,13 @@ write_png(const char *file_name, SDL_RWops *rw, png_bytep *rows,
                  PNG_INTERLACE_NONE, PNG_COMPRESSION_TYPE_BASE,
                  PNG_FILTER_TYPE_BASE);
 
-
-    if (palette){
+    if (palette) {
+        doing = "set pallete";
         const int ncolors = palette->ncolors;
         int i;
-        if (!(color_ptr = (png_colorp)SDL_malloc(sizeof(png_colorp) * ncolors)))
-            goto fail;      
+        if (!(color_ptr =
+                  (png_colorp)SDL_malloc(sizeof(png_colorp) * ncolors)))
+            goto fail;
         for (i = 0; i < ncolors; i++) {
             color_ptr[i].red = palette->colors[i].r;
             color_ptr[i].green = palette->colors[i].g;
@@ -221,7 +222,6 @@ write_png(const char *file_name, SDL_RWops *rw, png_bytep *rows,
         }
         png_set_PLTE(png_ptr, info_ptr, color_ptr, ncolors);
     }
-
 
     /* doing = "write info"; */
     png_write_info(png_ptr, info_ptr);
@@ -265,7 +265,7 @@ SavePNG(SDL_Surface *surface, const char *file, SDL_RWops *rw)
     SDL_Rect ss_rect;
     int r, i;
     int alpha = 0;
-    SDL_Palette* palette;    
+    SDL_Palette *palette;
     Uint8 surf_alpha = 255;
     Uint32 surf_colorkey;
     int has_colorkey = 0;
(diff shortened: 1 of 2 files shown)


ThanhChinhBK:
I temporarily replaced our own savePNG function with IMG_SavePNG from SDL image and it seems something wrong when saving 24bit indexed image:
```C
else if (!strcasecmp(ext, "png")) {
    if (rw != NULL) {
        result = IMG_SavePNG_RW(surf, rw, 0);
    }
    else {
        result = IMG_SavePNG(surf, name);
     }
}
```
Error when run testSavePNG24:
`
ERROR: testSavePNG24 (__main__.ImageModuleTest)
see if we can save a png with color values in the proper channels.
  File "..../pygame-2.1.3.dev1-py3.8-linux-x86_64.egg/pygame/tests/image_test.py", line 210, in testSavePNG24
    width, height, pixels, metadata = reader.asRGB()
  File "..../pygame/tests/test_utils/png.py", line 2178, in asRGB
    raise Error("will not convert image with alpha channel to RGB")
pygame.tests.test_utils.png.Error: Error: will not convert image with alpha channel to RGB

`

ankith26 (commented):
Thanks for the PR! 👍 
I cannot give an in-depth review because I am not too familiar with the image saving internals, but from a quick skim through the code it looks good except for a minor review I left

Additionally, you must run `python3 setup.py format` to format your code with `clang-format` so that our lint CI passes :)

src_c/imageext.c
```diff
@@ -207,6 +208,21 @@ write_png(const char *file_name, SDL_RWops *rw, png_bytep *rows, int w, int h,
                  PNG_INTERLACE_NONE, PNG_COMPRESSION_TYPE_BASE,
                  PNG_FILTER_TYPE_BASE);
 
+
```
ankith26, line 211:
```suggestion
`doing = "set pallete";`
```
This should be set here so that when `goto error;` happens, a more informative error is printed

src_c/imageext.c
```diff
@@ -207,6 +208,21 @@ write_png(const char *file_name, SDL_RWops *rw, png_bytep *rows, int w, int h,
                  PNG_INTERLACE_NONE, PNG_COMPRESSION_TYPE_BASE,
                  PNG_FILTER_TYPE_BASE);
 
+    if (palette) {
+        doing = "set pallete";
+        const int ncolors = palette->ncolors;
+        int i;
+        if (!(color_ptr =
+                  (png_colorp)SDL_malloc(sizeof(png_colorp) * ncolors)))
```
illume, line 216: Does this need to be freed anywhere?

test/image_test.py
```diff
@@ -222,6 +222,43 @@ def testSavePNG24(self):
             del reader
             os.remove(f_path)
 
+    def testSavePaletteAsPNG8(self):
+        """see if we can save a png with color values in the proper channels."""
+        # Create a PNG file with known colors
+        pygame.display.init()
+
+        reddish_pixel = (215, 0, 0)
```
illume, line 230: Nice test :)

src_c/imageext.c
ThanhChinhBK: thank you for your comment, I add a commit to free it after set PLTE

src_c/imageext.c
```diff
@@ -207,6 +208,22 @@ write_png(const char *file_name, SDL_RWops *rw, png_bytep *rows, int w, int h,
...
+            goto fail;
+        for (i = 0; i < ncolors; i++) {
+            color_ptr[i].red = palette->colors[i].r;
+            color_ptr[i].green = palette->colors[i].g;
+            color_ptr[i].blue = palette->colors[i].b;
+        }
+        png_set_PLTE(png_ptr, info_ptr, color_ptr, ncolors);
+        free(color_ptr);
```
illume, line 224: This should use SDL_free because it was malloc'd with SDL_malloc. It could be that a malloc implementation is not exactly compatible on all platforms/configurations. So we try to match up python/SDL/libc uses when we can.

illume:
Hrmm. I couldn't figure out why it is crashing on appveyor windows.

ankith26:
I believe the crash is related to the fact that `free` was called on memory allocated with `SDL_malloc`. IG SDL uses `VirtualAlloc` under hood on windows, so the memory allocator systems are probably not compatible here. As the review made by illume above, `SDL_free` must be used

illume:
Looks like that fixed the windows appveyor errors :)

illume (approved):
👍 🎈 thanks

ankith26 (approved):
LGTM! Thanks again :) 👍

PR #3032, opened 2022-02-15
Regenerate cython files

### LINKED ISSUE #3031: doesnt build with python 3.11
https://github.com/NixOS/nixpkgs/issues/160150

ankith26:
PR to fix #3031

Updated all cython files with a forced regeneration on the latest cython version, in particular, the pypm file which was not updated for a while.

#2916 is the way to go for the future, but since that PR is a draft and might take longer time, this is a good stopgap update for now


dr0id (approved):
well, if it is generated code and all tests pass...  can't be so wrong then, can it?
can't really judge all lines, but most make sense

MyreMylar (approved):
LGTM

PR #3037, opened 2022-02-17
Cursors example update

### LINKED ISSUE #3034: Update cursors example
I've noticed that the cursors example is rather ugly and out of date. There's a whole pygame.cursors.Cursor class now, and new types of cursors (system cursors, cursors from surfaces). Updating (and making it prettier while you're at it) that example would be super helpful.

thanice:
Closes #3034


thanice:
Yes that would be better! I will make the changes you mentioned and add them to the pr.

thanice:
I made the changes you noted and also added a button to click in order to change the cursor. 
It is not ready yet but I wanted to know if it is like what you had in mind.

Starbuck5:
I was envisioning a couple circles scattered across the screen area just to give the cursors something to interact with.

I like the click to advance mechanic, I think that should be kept.

It would be cool to have a couple more system cursor examples as well. HAND and IBEAM would be good.

thanice:
So i should add more circles and change their positions? Also should i keep the hover to change color or make them change color when clicked?
And about the text with the info about the cursor type should i keep it as it is or change it to be more descriptive?

examples/cursors.py
```diff
@@ -1,104 +1,221 @@
...
+    circle5 = pg.draw.circle(bg, (255, 255, 255), (pos_x5, pos_y5), radius5)
+    circle6 = pg.draw.circle(bg, (255, 255, 255), (pos_x6, pos_y6), radius6)
+    circle7 = pg.draw.circle(bg, (255, 255, 255), (pos_x7, pos_y7), radius7)
+
+
+    #Initialize button
+    button_text = font1.render("Click here to change cursor", True, (0, 0, 0))
+    button = pg.draw.rect(bg, (180,180,180),(139, 300, button_text.get_width()+5, button_text.get_height() + 50))
+    button_text_rect = button_text.get_rect(center=(button.center))
+    bg.blit(button_text, button_text_rect)
```
MyreMylar, line 137: You can remove the brackets around `button.center` here.

examples/cursors.py
```diff
@@ -1,104 +1,221 @@
...
+        bg.blit(text1, text_rect1)
+        
+        
+        for event in pg.event.get():
+
+            button = pg.draw.rect(bg, (100,149,240),(139, 300, button_text.get_width()+5, button_text.get_height() + 50))
+            bg.blit(button_text, button_text_rect)
+
+            #Check if button was clicked and change cursor
+            if mouse_x > button.x and mouse_x < button.x + button.width and mouse_y > button.y and mouse_y < button.y + button.height:
```
MyreMylar, line 203: we can rewrite this as:

```
if ( button.x < mouse_x < button.x + button.width and button.y < mouse_y < button.y + button.height ):
```

MyreMylar (changes requested):
All seems to work OK, couple of minor tweaks - and it could use a general formatting pass. Download the latest version of `black` and do:

`black examples/cursors.py` in the terminal from the pygame directory.

examples/cursors.py
Starbuck5: Rect.collidepoint would probably be better, if I understand what this code is doing properly.

MyreMylar: Agreed.

examples/cursors.py
```diff
@@ -1,105 +1,257 @@
...
+        bitmap_cursor1,
+        system_cursor2,
+        bitmap_cursor2,
+        image_cursor,
+    ]
+    index = 0
+    pg.mouse.set_cursor(cursors[index])
+
+    while True:
+
```
dr0id, line 159: please limit the fps to not hog the cpu! (Hint: use  `clock.tick(50)`  )

examples/cursors.py
```diff
@@ -1,105 +1,257 @@
...
+            button = pg.draw.rect(
+                bg,
+                (100, 149, 240),
+                (139, 300, button_text.get_width() + 5, button_text.get_height() + 50),
+            )
+            bg.blit(button_text, button_text_rect)
+
+            # Check if button was clicked and change cursor
+            if button.collidepoint(mouse_x, mouse_y):
```
dr0id, line 219: I find it a bit strange to have code in the `for event in pg.event.get()` loop that does not use the event. There is also a funny effect if you press a mouse button and wander over the button: the cursors change wildly! 😀

dr0id:
Keep up the good work.

examples/cursors.py
thanice: Oh yeah i haven't noticed that! I will fix it and add it to the pr.

dr0id: Thanks for the update. It works better. Two things I noticed: when you click, the pressed colors only shows for about 1 frame and the other thing is, if you press the mouse button outside the box and move the mouse into the box area while keeping the mouse button pressed, then it changes too. But I think most wont see this.

examples/cursors.py
```diff
@@ -1,104 +1,260 @@
...
+    pressed = False
+    clock = pg.time.Clock()
+
+    while True:
+
+        clock.tick(50)
+
+        mouse_pos = pg.mouse.get_pos
+        mouse_x = mouse_pos()[0]
+        mouse_y = mouse_pos()[1]
```
dr0id, line 165:
```
        mouse_pos = pg.mouse.get_pos
        mouse_x = mouse_pos()[0]
        mouse_y = mouse_pos()[1]
```
this can be written more elegantly as

```
        mouse_x, mouse_y = pg.mouse.get_pos()
```

dr0id (approved):
Although there is some repetitions in the code I think it is ok for a simple example.

MyreMylar (approved):
LGTM 👍

PR #3038, opened 2022-02-19
Add `pygame.get_preferred_locales`

ankith26:
Add `get_pref_locales` to wrap [`SDL_GetPreferredLocales`](https://wiki.libsdl.org/SDL_GetPreferredLocales)

I believe this is a useful feature to games wanting to customise in-game text to user locale settings.

A companion PR for #2953 which exposes `LOCALECHANGED` event, which is sorta related to this PR.

### Things to discuss
- [X] *Is it fine to put this function in `base.c` (comes in pygame main namespace)? If not, which submodule is more suitable for this kind of function?*
**Resolution:** Make a new module for this function, functions from `SDL_power` and SDL path related functions

### DIFF
--- a/src_c/context.c
+++ b/src_c/context.c
@@ -26,9 +26,81 @@ pg_context_get_pref_path(PyObject *self, PyObject *args, PyObject *kwargs)
     return ret;
 }
 
+static PyObject *
+pg_context_get_pref_locales(PyObject *self, PyObject *_null)
+{
+    PyObject *ret_list = PyList_New(0);
+    if (!ret_list) {
+        return NULL;
+    }
+
+#if SDL_VERSION_ATLEAST(2, 0, 14)
+    PyObject *dict, *val = NULL;
+    SDL_Locale *locales = SDL_GetPreferredLocales();
+    if (!locales) {
+        /* Return an empty list if SDL function does not return any useful
+         * information */
+        return ret_list;
+    }
+
+    SDL_Locale *current_locale = locales;
+    while (current_locale->language) {
+        dict = PyDict_New();
+        if (!dict) {
+            goto error;
+        }
+
+        val = PyUnicode_FromString(current_locale->language);
+        if (!val) {
+            goto error;
+        }
+        if (PyDict_SetItemString(dict, "language", val)) {
+            goto error;
+        }
+        Py_DECREF(val);
+
+        if (current_locale->country) {
+            val = PyUnicode_FromString(current_locale->country);
+            if (!val) {
+                goto error;
+            }
+        }
+        else {
+            Py_INCREF(Py_None);
+            val = Py_None;
+        }
+        if (PyDict_SetItemString(dict, "country", val)) {
+            goto error;
+        }
+        Py_DECREF(val);
+
+        /* reset val to NULL because goto XDECREF's this */
+        val = NULL;
+        if (PyList_Append(ret_list, dict)) {
+            goto error;
+        }
+        Py_DECREF(dict);
+        current_locale++;
+    }
+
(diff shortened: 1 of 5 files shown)


illume:
I wonder if something like this is already exposed in the python standard library?

ankith26:
I played around with the python [`locale`](https://docs.python.org/3/library/locale.html) module, and I found it to be a bit lacking, atleast on my software and setup (windows 11). First of all, it does not provide a preference list like the SDL function does, which I think is especially handy for games that don't support the main locale of a system, and would like to fallback to another locale setting the user prefers. I also tried to write a unit test that compares the output of the pygame function with `getlocale` of the locale module, but for me on windows, locale module seems to return non-standard strings for my locale (`en_IN`, and locale returned something like `English_India`) so parsing and handling all cases like this would be a pain for users, while SDL function conveniently returns `{'language': 'en', 'country': 'US'}` (dunno why it's returning US for me, could be due to the fact that some of my configs have `US`)

Starbuck5:
I've been thinking about adding a `pygame.context` module for things like this.

Specifically I wanted to port [SDL_power ](https://wiki.libsdl.org/CategoryPower) functionality and [SDL_GetPrefPath](https://wiki.libsdl.org/SDL_GetPrefPath). That would logically shake out into a `pygame.power` module and a `pygame.path` module, but both modules would be tiny, with just a couple of functions each. So `context` sorta makes sense as a broader category to me.

ankith26:
I like the `pygame.context` idea too, except maybe it should have a different name maybe? `pygame.utils`? maybe some other name like this?

Starbuck5:
The reason I was thinking “context” is because these functions give you additional context for the system, but info that doesn’t fit in any module so far. Like system locale, or system power status, it’s a different bit of context.

“Utils” seems less descriptive.

MyreMylar:
> system ...  system locale ... system power status

Maybe it should be `pygame.system` :)

ankith26:
Ah.. I like `pygame.system` better than `pygame.context` too

Starbuck5:
I was also thinking about `pygame.platform`

ankith26:
`pygame.system.get_preferred_locales` vs `pygame.platform.get_preferred_locales`
`pygame.system.get_power_info` vs `pygame.platform.get_power_info`
`pygame.system.get_pref_path` vs `pygame.platform.get_pref_path`
My vote is for `pygame.system` because all these functions give informations regarding the system, and not the platform

MyreMylar (approved):
Reviewed and tested locally and all seems to work. Should we just merge this into context and then review it all again from there?

ankith26:
I'll base this PR on `main` once the base context PR is merged, so this PR can be merged in independently

src_c/context.c
```diff
@@ -26,9 +26,81 @@ pg_context_get_pref_path(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    PyObject *dict, *val = NULL;
+    SDL_Locale *locales = SDL_GetPreferredLocales();
+    if (!locales) {
+        /* Return an empty list if SDL function does not return any useful
+         * information */
+        return ret_list;
+    }
+
+    SDL_Locale *current_locale = locales;
+    while (current_locale->language) {
```
Starbuck5, line 47: Maybe add a comment explaining the list is terminated by a locale with a NULL language field. Just since this is kind of a strange pattern without `context`.

Starbuck5 (approved):
Looks good!

Thanks Ankith!

src_c/context.c
ankith26: yeah good idea

PR #3041, opened 2022-02-20
Get rid of sprintf calls in the codebase

ankith26:
sprintf is considered an 'unsafe' function, sometimes compilers warn about its usage, so use PyOS_snprintf/PyUnicode_FromFormat/PyErr_Format in its place

### DIFF
--- a/src_c/_camera.c
+++ b/src_c/_camera.c
@@ -160,41 +160,46 @@ list_cameras(PyObject *self, PyObject *arg)
 #else
     WCHAR **devices;
 #endif
-    int num_devices, i;
+    int j, i = 0, num_devices = 0;
 
-    num_devices = 0;
-    ret_list = NULL;
-    ret_list = PyList_New(0);
-    if (!ret_list)
-        return NULL;
-
+    /* TODO for future PRs: errors in these functions are being ignored as
+     * of now, and an empty list is being returned by this function */
 #if defined(__unix__)
     devices = v4l2_list_cameras(&num_devices);
 #elif defined(PYGAME_WINDOWS_CAMERA)
     devices = windows_list_cameras(&num_devices);
 #endif
+
+    ret_list = PyList_New(num_devices);
+    if (!ret_list) {
+        goto error;
+    }
+
     for (i = 0; i < num_devices; i++) {
 #if defined(PYGAME_WINDOWS_CAMERA)
         string = PyUnicode_FromWideChar(devices[i], -1);
 #else
         string = PyUnicode_FromString(devices[i]);
 #endif
-        if (0 != PyList_Append(ret_list, string)) {
-            /* Append failed; clean up and return */
-            Py_DECREF(ret_list);
-            Py_DECREF(string);
-            for (; i < num_devices; i++) {
-                free(devices[i]);
-            }
-            free(devices);
-            return NULL; /* Exception already set. */
+        if (!string) {
+            goto error;
         }
-        Py_DECREF(string);
+        /* steals reference to 'string' */
+        PyList_SET_ITEM(ret_list, i, string);
         free(devices[i]);
     }
     free(devices);
 
     return ret_list;
+error:
+    /* free the remaining devices */
+    for (j = i; j < num_devices; j++) {
+        free(devices[j]);
+    }
+    free(devices);
+    Py_XDECREF(ret_list);
+    return NULL;
+
 #else
     Py_RETURN_NONE;
 #endif
--- a/src_c/camera_v4l2.c
+++ b/src_c/camera_v4l2.c
@@ -35,8 +35,15 @@ v4l2_list_cameras(int *num_devices)
     num = *num_devices;
 
     devices = (char **)malloc(sizeof(char *) * 65);
+    if (!devices) {
+        return NULL;
+    }
 
     device = (char *)malloc(sizeof(char) * 13);
+    if (!device) {
+        return NULL;
+    }
+
     strcpy(device, "/dev/video");
     fd = open(device, O_RDONLY);
     if (fd != -1) {
@@ -47,7 +54,11 @@ v4l2_list_cameras(int *num_devices)
     close(fd);
     /* v4l2 cameras can be /dev/video and /dev/video0 to /dev/video63 */
     for (i = 0; i < 64; i++) {
-        sprintf(device, "/dev/video%d", i);
+        int ret = PyOS_snprintf(device, 13, "/dev/video%d", i);
+        if (ret < 0 || ret >= 13) {
+            return NULL;
+        }
+
         fd = open(device, O_RDONLY);
         if (fd != -1) {
             devices[num] = device;
--- a/src_c/camera_v4l2.c
+++ b/src_c/camera_v4l2.c
@@ -29,53 +29,64 @@ char **
 v4l2_list_cameras(int *num_devices)
 {
     char **devices;
-    char *device;
+    char *device = NULL;
     int num, i, fd;
 
-    num = *num_devices;
+    num = *num_devices = 0;
 
     devices = (char **)malloc(sizeof(char *) * 65);
     if (!devices) {
         return NULL;
     }
 
-    device = (char *)malloc(sizeof(char) * 13);
-    if (!device) {
-        return NULL;
-    }
-
-    strcpy(device, "/dev/video");
-    fd = open(device, O_RDONLY);
-    if (fd != -1) {
-        devices[num] = device;
-        num++;
-        device = (char *)malloc(sizeof(char) * 13);
-    }
-    close(fd);
     /* v4l2 cameras can be /dev/video and /dev/video0 to /dev/video63 */
-    for (i = 0; i < 64; i++) {
-        int ret = PyOS_snprintf(device, 13, "/dev/video%d", i);
-        if (ret < 0 || ret >= 13) {
-            return NULL;
+    for (i = -1; i < 64; i++) {
+        device = (char *)malloc(sizeof(char) * 13);
+        if (!device) {
+            goto error;
+        }
+
+        if (i == -1) {
+            strcpy(device, "/dev/video");
+        }
+        else {
+            int ret = PyOS_snprintf(device, 13, "/dev/video%d", i);
+            if (ret < 0 || ret >= 13) {
+                goto error;
+            }
         }
 
         fd = open(device, O_RDONLY);
-        if (fd != -1) {
+        if (fd == -1) {
+            free(device);
+        }
+        else {
+            /* 'device' is in array now, don't free it here */
             devices[num] = device;
             num++;
-            device = (char *)malloc(sizeof(char) * 13);
         }
-        close(fd);
+        device = NULL;
+
+        if (close(fd) == -1) {
+            /* Error while closing file */
+            goto error;
+        }
     }
 
-    if (num == *num_devices) {
-        free(device);
-    }
-    else {
-        *num_devices = num;
-    }
-
+    *num_devices = num;
     return devices;
+
+error:
+    /* During goto here, 'device' must be either NULL or free-able allocated
+     * memory */
+    free(device);
+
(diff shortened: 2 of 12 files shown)


src_c/display.c
```diff
@@ -345,26 +344,25 @@ pg_vidinfo_str(PyObject *self)
...
-            info->vfmt->Gshift, info->vfmt->Bshift, info->vfmt->Ashift,
-            info->vfmt->Rloss, info->vfmt->Gloss, info->vfmt->Bloss,
-            info->vfmt->Aloss, current_w, current_h, pixel_format_name);
-    return PyUnicode_FromString(str);
+    return PyUnicode_FromFormat(
+        "<VideoInfo(hw = %u, wm = %u,video_mem = %u\n"
+        "         blit_hw = %u, blit_hw_CC = %u, blit_hw_A = %u,\n"
+        "         blit_sw = %u, blit_sw_CC = %u, blit_sw_A = %u,\n"
+        "         bitsize  = %u, bytesize = %u,\n"
+        "         masks =  (%u, %u, %u, %u),\n"
```
MyreMylar, line 352: for this masks line can we change it to:
```
masks =  (%08x, %08x, %08x, %08x),\n"
```
Masks make a lot more sense as a hex value rather than large ints

MyreMylar (approved):
LGTM 👍 

I had one non-essential suggestion for changing the way masks are displayed. Probably we could do that in a separate PR and update all the places masks are displayed.

src_c/display.c
ankith26: good idea!

ankith26: Fixed it in this PR itself because it was related, and also I did not find any other instances of masks being repr'd as strings, in the quick skim I did through the code

src_c/camera_v4l2.c
```diff
@@ -35,8 +35,15 @@ v4l2_list_cameras(int *num_devices)
     num = *num_devices;
 
     devices = (char **)malloc(sizeof(char *) * 65);
+    if (!devices) {
+        return NULL;
+    }
 
     device = (char *)malloc(sizeof(char) * 13);
+    if (!device) {
+        return NULL;
```
illume, line 44: Looks like there's a leak here at least with devices needing to be freed.

ankith26: Good catch

ankith26: Ah looks like there are more leaks in this function, gonna fix all of em

src_c/camera_v4l2.c
```diff
@@ -57,13 +75,9 @@ v4l2_list_cameras(int *num_devices)
         close(fd);
```
dr0id, line 75: Suggestion: this code block
```

        fd = open(device, O_RDONLY);
        if (fd != -1) {
            devices[num] = device;
            num++;
            device = (char *)malloc(sizeof(char) * 13);
        }
        close(fd);
```
is on line 49 already, maybe it could be extracted to its own method?

dr0id (approved):
Left a suggestion, implement if you agree, ignore it if you disagree.

src_c/camera_v4l2.c
ankith26: Thanks for the review! 
Hmmm however I don't think this would work well as a seperate function, because there are many local variables it needs to make use of, like `num`, `device` and `devices` and passing all of them by ref as args seems excessive for a 7 line code block to me. It would work better as a macro, but again, macros must be avoided when they can.

ankith26: Now that you reviewed on this bit of code, I see the malloc return value is not been error-checked here, which needs to be done

src_c/freetype/ft_render.c
```diff
@@ -655,9 +655,12 @@ _PGFT_Render_Array(FreeTypeInstance *ft, pgFontObject *fontobj,
...
     if (_validate_view_format(view_p->format)) {
         static const char fmt[] = "Unsupported array item format '%.*s'";
-        char msg[100 + sizeof(fmt)];
 
-        sprintf(msg, fmt, (int)(sizeof(msg) - sizeof(fmt)), view_p->format);
+        /* set a default error message incase PyOS_snprintf fails */
+        char msg[100 + sizeof(fmt)] = "Unsupported array item format";
+
+        PyOS_snprintf(msg, sizeof(msg), fmt, (int)(sizeof(msg) - sizeof(fmt)),
+                      view_p->format);
```
Starbuck5, line 663: I believe this one could use PyUnicode_FromFormat instead, which seems preferable. `view_p` is a Py_buffer, the `format` member is a null terminated char array. https://docs.python.org/3/c-api/buffer.html#c.Py_buffer.format

ankith26: Here `msg` is a C string, that function is for python unicode objects. Yeah that API can be used, and then another function to convert the python string to a C string would need calling, which is more code, and not as performant as `PyOS_snprintf`

Starbuck5: Well can't PyUnicode_Format or PyErr_Format take c strings?

src_c/camera_v4l2.c
```diff
@@ -29,42 +29,64 @@ char **
...
-    }
-
+    /* Free device because it is always freshly allocated at this stage */
+    free(device);
+    *num_devices = num;
     return devices;
+
+error:
+    free(device);
+    /* free individial device already in devices array */
```
Starbuck5, line 84: Should be `individual`

src_c/freetype/ft_render.c
ankith26: `PyErr_Format` seems to do the trick here, so I've implemented that, thanks!

src_c/camera_v4l2.c
```diff
@@ -29,42 +29,64 @@ char **
...
+    }
+
     strcpy(device, "/dev/video");
     fd = open(device, O_RDONLY);
     if (fd != -1) {
         devices[num] = device;
         num++;
         device = (char *)malloc(sizeof(char) * 13);
+        if (!device) {
+            goto error;
```
dr0id, line 54: should there be a `   close(fd);` before the `goto error;`?

src_c/camera_v4l2.c
```diff
@@ -29,42 +29,64 @@ char **
...
+        }
+
         fd = open(device, O_RDONLY);
         if (fd != -1) {
             devices[num] = device;
             num++;
             device = (char *)malloc(sizeof(char) * 13);
+            if (!device) {
+                goto error;
```
dr0id, line 71: `close(fd)` here too before `goto error;`

src_c/camera_v4l2.c
ankith26: Good catch! I pushed another commit where I reworked the code, so it should be a bit simpler now, and this has been fixed

src_c/camera_v4l2.c
```diff
@@ -29,42 +29,64 @@ char **
 v4l2_list_cameras(int *num_devices)
 {
     char **devices;
-    char *device;
+    char *device = NULL;
     int num, i, fd;
 
     num = *num_devices;
```
dr0id, line 35: why is num not initialized to 0 ? I mean later it is used as an index into the  devices array and incremented, therefore if should at most have the value of 64 I guess.

ankith26: `num` is assigned to `*num_devices`, and the caller has already set `*num_devices` to 0 :)

dr0id: correction: at most 65

ankith26: Well I see your point, the current code is fine but it relies on the caller `0` initialising the argument passed by reference which is not ideal, more explicit initialisation to 0 is better here

dr0id: good to go I think now

src_c/_camera.c
```diff
@@ -160,41 +160,46 @@ list_cameras(PyObject *self, PyObject *arg)
...
-                free(devices[i]);
-            }
-            free(devices);
-            return NULL; /* Exception already set. */
+        if (!string) {
+            goto error;
         }
-        Py_DECREF(string);
+        /* steals reference to 'string' */
+        PyList_SET_ITEM(ret_list, i, string);
```
Starbuck5, line 188: Should we use the function form instead to stay closer to the stable ABI?

src_c/display.c
```diff
@@ -345,26 +344,25 @@ pg_vidinfo_str(PyObject *self)
...
+        "<VideoInfo(hw = %u, wm = %u,video_mem = %u\n"
+        "         blit_hw = %u, blit_hw_CC = %u, blit_hw_A = %u,\n"
+        "         blit_sw = %u, blit_sw_CC = %u, blit_sw_A = %u,\n"
+        "         bitsize  = %u, bytesize = %u,\n"
+        "         masks =  (0x%08x, 0x%08x, 0x%08x, 0x%08x),\n"
+        "         shifts = (%u, %u, %u, %u),\n"
+        "         losses =  (%u, %u, %u, %u),\n"
+        "         current_w = %d, current_h = %d\n"
+        "         pixel_format = %s)\n"
+        ">\n",
```
Starbuck5, line 357: It doesn't look like it needs the trailing newline.

src_c/camera_v4l2.c
```diff
@@ -29,42 +29,64 @@ char **
...
-        sprintf(device, "/dev/video%d", i);
+    for (i = -1; i < 64; i++) {
+        device = (char *)malloc(sizeof(char) * 13);
+        if (!device) {
+            goto error;
+        }
+
+        if (i == -1) {
+            strcpy(device, "/dev/video");
+        }
```
Starbuck5, line 51: Why do you do this? "This" being the i == -1 iteration.

src_c/_camera.c
ankith26: Well... A lot of pygame code uses non-stable ABI code ATM... porting to stable ABI will be hard anyways, because even basic and important structures like `PyTypeObject` are not a part of the limited API. So if we ever need to support it, it would basically mean rewriting most of pygame, either implementing all pygame classes in python with calls to internal pygame functions written in C, or maybe dynamic classes in C code, which I am not really sure, will definitely be harder than a python rewrite. This could also be handled if we ever do a `HPy` port.

TL;DR stable ABI is still very far from now, so better to not worry about it for now

src_c/display.c
ankith26: good catch

src_c/camera_v4l2.c
ankith26: This is a special case. `/dev/video` is a thing, and `/dev/video0` to `/dev/video63`. So in total we need 65 iterations to handle all cases.

The previous code handled `/dev/video` outside the loop, but with this, it is handled in the loop for better consistency and easier error handling

Starbuck5: Oh, I missed that. Thanks.

src_c/_camera.c
Starbuck5: PyTypeObject is part of the stable ABI.

I'm not pushing the stable ABI hard because we already have good infrastructure to handle wheel generation, and the stable ABI can be a bit slower in places.

Up to you to resolve.

ankith26: `PyTypeObject` is definitely not a part of the stable ABI. This is something I got stuck with when I was trying to port a small C extension to the stable ABI.
According to the [docs](https://docs.python.org/3/c-api/typeobj.html#static-types)
> Also, since [PyTypeObject](https://docs.python.org/3/c-api/type.html#c.PyTypeObject) is not part of the [stable ABI](https://docs.python.org/3/c-api/stable.html#stable), any extension modules using static types must be compiled for a specific Python minor version.

So for the same reason, I will leave this function in macro form for now, because it can be a tad bit faster and the code is also simpler without unnecessary error checking

Starbuck5 (approved):
Looks good to me!

Lots of these look a lot cleaner and safer than the sprintfs.

PR #3042, opened 2022-02-21
Update some deps in manylinux scripts

ankith26:
Updates some deps to the respective latest version in the manylinux scripts. Opened PR because we need a docker image rebuild to fix an issue caused by an older version of auditwheel, so while we are at that we can also update some deps ;)

### Potential TODOs for this PR
- [x] Use pypy from manylinux base image, which means it will be kept updated without much changes on our side. ~~Also generate pypy 3.8 and 3.9 manylinux wheels~~ this will be done in a future PR
- [x] Update CMake to the latest version
- [x] Update PortMidi to the latest version that's on github (needs above step to be completed first)
- [x] ~~Update pulseaudio (might potentially need cmake update step to be done first)~~ Updated it from 14.0 to 14.2. Further updating it to 15.0 requires meson build system, so this update can happen in the future.
- [ ] ~~Update glib maybe? (mac only)~~ skipped due to build issues for now
- [x] Update Alsa (linux only)


Starbuck5:
+ SDL_ttf 2.0.18

dr0id:
The builds seem to fail because of 
`/io/buildconfig/manylinux-build/build-wheels.sh: line 40: /opt/python/pp38-pypy38_pp75/bin/pypy: No such file or directory`
Not sure what the problem is. Either pypy is not installed correctly or this path is wrong.

ankith26:
That's something that should be fixed when the docker images are repushed with the latest manylinux base images

ankith26:
Okay, this PR should be done from my side, I have tested this locally (Ubuntu 22.04 WSL, manylinux2014 x86_64 image built using docker) by running all unittests and examples and this PR does not seem to break anything. On the bright side, audio was not working for me on WSL (it gave a load of ALSA errors in shell) and now it seems to work after I updated ALSA in this PR.

I tried updating SDL from 2.0.20 to 2.0.22, but this gave a serious regression - the games I tested played at a much lower FPS, around 5. With some profiling, I found that `pygame.display.update()` takes over 140 ms per call, it normally executes within a sub-millisecond duration. Still don't know what exactly went wrong here, for now it's better to stick to 2.0.20 and investigate this issue further in the future (I don't know whether it's an issue on my end, or whether it could be an SDL issue, or an issue with other lower-level dependencies)

Previously this PR added more pypy builds and updated manylinux CI to use cibuildwheel, but I moved that to a separate branch and will open a PR for it in the future.

Starbuck5 (approved):
Looks good to me.

A little strange that it doesn't do SDL 2.0.22, but we've talked about that being weird.

I'm happy to get SDL_ttf 2.0.18 and it's good to keep all the subdeps updated too.

MightyJosip (approved):
Nice modernization. Thanks.

ankith26:
Thanks for the reviews!

I wasn't planning to do more changes on this branch, but there were weird appveyor issues with appeyor refusing to run on this PR, so I pushed another (very minor) commit that updates some more stuff.

This latest commit is untested, but I plan on testing it soon

illume:
Can you please rebase this?

What's with the circle CI error? Is it expected?

ankith26:
The CircleCI error is popping up because this PR is too old ;)

It's been a while since I updated this PR, in the mean time SDL and its satellite libraries have had huge updates.

The SDL satellite libraries vendor in their dependencies too (and in the recent updates, these have a new set of dependencies), and if we decide to use that in the future, we will also be reducing a lot of config on our end (so far this extra config on our end was a good idea because the SDL satellite libs were updated less frequently, but this seems to be changing now)

However I don't think those updates should happen in 2.1.3 at all, because that may delay the release further (hopefully we can add this sort of thing early in a 2.1.4 dev release, so that it's tested better)

About this PR... Even though I think it's pretty safe in the current state, it would still need more testing on a larger variety of platforms though (and *maybe* some more minor updates)... I don't think this PR should block 2.1.3 either

ankith26:
I've rebased this PR on `main`

Everything up till the latest commit has been tested. The latest commit just does a bit more catching up with the sub-dep versions (which is still untested as of now, but I'll get to it in a day or two after I get back home to my main laptop)

@illume thanks for giving me access to docker, I think going forward we can share the work on this front if you are okay with this (I have a raspberry pi 3 running 64-bit PiOS at hand, I can use it to make and test our aarch64 images, and I believe this will be faster than QEMU on a amd64 machine)

Since you once mentioned about dropping 32-bit manylinux support, I now think that's a good idea, but I guess we can do it in smaller steps? (instead of getting rid of all three 32-bit manylinux wheels together, we can get rid of the older two of those, and we can keep the other one in for a bit, and see if we get any complaints. We can always redirect people to update their `pip` in this case)

Like I said above, this PR is not a 2.1.3 blocker or super urgent, it's only a nice to have.

illume:
> The SDL satellite libraries vendor in their dependencies too (and in the recent updates, these have a new set of dependencies), and if we decide to use that in the future, we will also be reducing a lot of config on our end (so far this extra config on our end was a good idea because the SDL satellite libs were updated less frequently, but this seems to be changing now)

How about we see in a few releases if they keep updating frequently? Then if they do remove our config. Sound ok to you?

There's something to be said about using the same versions they are... but in general latest versions of these packages are better with important fixes usually.

> However I don't think those updates should happen in 2.1.3 at all, because that may delay the release further (hopefully we can add this sort of thing early in a 2.1.4 dev release, so that it's tested better)

Hrmm.

In the past we've tried to release all the platforms with the same SDL libs all together. Because that way testing is a lot easier for:
- pygame developers, because bug reports with SDL versions are easier and we have to support less SDL version issues at once.
- for pygame app developers, because they can often link their SDL version issues to pygame version issues. Less of a test matrix for pygame app developers too.

I think it's important we keep the SDL versions in sync for the release.

Options?
- revert windows change
- update these manylinux scripts

Reverting windows changes is not so good, because we have some important-to-us-bug-fixes in there.

> Like I said above, this PR is not a 2.1.3 blocker or super urgent, it's only a nice to have.

To me this looks good, it can also be a step towards updating to latest SDL* libs.

illume:
> In theory... docker image generation can also be automated with CI, maybe this could be a good middle term goal

Yeah. I can spin up a big VM to do the builds pretty quickly. The CI builds are kind of slow in comparison. Hopefully when ARM vms are added to CI this will be fine. Fingers crossed it doesn't take too long.

> Since you once mentioned about dropping 32-bit manylinux support, I now think that's a good idea, but I guess we can do it in smaller steps? (instead of getting rid of all three 32-bit manylinux wheels together, we can get rid of the older two of those, and we can keep the other one in for a bit, and see if we get any complaints. We can always redirect people to update their pip in this case)

Sounds like a good plan.

illume (approved):
🎉

ankith26:
> How about we see in a few releases if they keep updating frequently? Then if they do remove our config. Sound ok to you?

Definitely! It's a long term thing that we can only keep an eye for (we don't have any issues/bugs with the existing config so there's no solid reason to be removing it at all)

PR #3043, opened 2022-02-22
Fix APPMOUSEFOCUS naming, update const values

ankith26:
pygame1 compatibility fix PR. Events of type `ACTIVEEVENT` must have a `state` attribute which is one of the possible 3 states. These three states had different constant values in SDL1 and pygame2, and there was a misnamed constant. This PR is a quickfix for both these issues


dr0id (approved):
LGTM (hopefully nobody is already using the wrong written APPFOCUSMOUSE constant!)

ankith26:
This event is not really documented that well, so I doubt anyone noticed this issue before. Since this is a "legacy" event, I don't think this needs documentation, people should switch to using the new `WINDOWEVENT` API

Starbuck5:
How was there a misnamed constant? The compiler would've failed if there wasn't anything named `SDL_APPFOCUSMOUSE` right?

Starbuck5 (approved):
Thanks for this!

This seems very harmless, just patching up a potential hole for old code. Neither the incorrect or the correct name was documented, just that ACTIVEEVENT has a state attribute. And that's okay.

PR #3044, opened 2022-02-22
Fix really tiny mistake

gresm:
It seems like this mistake is a bit popular...


ankith26 (approved):
LGTM! Thanks for the PR 👍 🎉

dr0id (approved):

Issue #3040, opened 2022-02-20
Improve example `textinput`


blankRiot96:
**Description**
I was looking through some examples and came across <a href="https://github.com/pygame/pygame/blob/main/examples/textinput.py">textinput.py</a>. I found the code to be a bit unconventional, and also found the usage of global variables.
Additionally, I would also like to take an object oriented approach to this example, I'm not sure if this is okay since I couldn't find much OOP being used in other examples, but even if not I feel the example can be improved by a bit - Adding cursor blink, keeping the input centered in the window and also for some reason the text moves a bit downward/upward based on the characters that have been given? I would like to fix that.

ankith26:
OOP is definitely a good idea, and some of our examples do have OOP :)

robertpfeiffer:
Wait, when did we add this? Looks like I can take my textinput off the to-do list

ankith26:
The linked PR is merged, closing issue

PR #3050, opened 2022-02-23
Improve example `textinput`

blankRiot96:
Fixes issue <a href="https://github.com/pygame/pygame/issues/3040">#3040</a>. 
I was unable to add the cursor blink due to personal time constraints, but I've refactored the code and re structured it to use OOP.

### DIFF
--- a/examples/textinput.py
+++ b/examples/textinput.py
@@ -184,19 +184,6 @@ class Game:
     SCREEN_WIDTH, SCREEN_HEIGHT = 640, 480
     BG_COLOR = "black"
 
-    # Add font name for each language,
-    # otherwise some text can't be correctly displayed.
-    FONT_NAMES = [
-        "notosanscjktcregular",
-        "notosansmonocjktcregular",
-        "notosansregular,",
-        "microsoftjhengheimicrosoftjhengheiuilight",
-        "microsoftyaheimicrosoftyaheiuilight",
-        "msgothicmsuigothicmspgothic",
-        "msmincho",
-        "Arial",
-    ]
-
     def __init__(self, caption: str) -> None:
         # Initalize
         pg.init()
--- a/examples/textinput.py
+++ b/examples/textinput.py
@@ -43,7 +43,7 @@ class TextInput:
         self.promt = promt
         self.print_event = print_event
         # position of chatlist and chatbox
-        self.CHAT_LIST_POS = pg.Rect(pos, (screen_dimensions[0], 400))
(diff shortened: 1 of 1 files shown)


dr0id:
I don't think that it is intended that the first written text gets behind the input box?
image

examples/textinput.py
```diff
@@ -6,169 +6,235 @@
...
+    def __init__(
+        self, promt: str,
+        pos, screen_dimensions,
+        print_event: bool,
+        text_color="white"
+    ) -> None:
+        self.promt = promt
+        self.print_event = print_event
+        # position of chatlist and chatbox
+        self.CHAT_LIST_POS = pg.Rect(pos, (screen_dimensions[0], 400))
```
dr0id, line 46: Use a slightly different position for the CHAT_LIST_POS (+50 in y direction would do the trick I guess)

blankRiot96:
Weird it works fine for me 
image

examples/textinput.py
blankRiot96: Got it, thanks for the review!

dr0id:
> Weird it works fine for me image

Well, yes I can input 'normal text' fine but I thought I try some special utf characters.
Thanks for the position fix.

dr0id (approved):

blankRiot96:
Thanks for the approval! My first time here so I wasn't too sure if I was doing it right 😅

dr0id:
@robertpfeiffer did you say in discord that you had it working with UTF symbols too? I wonder what is needed to get it working with UTF symbols. Can you elaborate? I do not mind if it is not in this PR.

robertpfeiffer:
> UTF

I had written a textinput example app in 2020. We discussed it on Discord. I did not add it to the pygame examples because it doesn't work in any *meaningful* way with UTF-8 without something like GNU Unifont (that covers the whole Unicode basic multilingual plane). It also doesn't work properly with ibus/mozc or fcitx, because of the way manylinux wheels are linked. So I had written that textinput example that works really well, but more or less only on my machine, because of packaging issues.

The same issues are also affecting this textinput example. It's not as nice looking and feature-rich as the thing I posted on the Discord, but that's not really the problem. The problem is it's not properly working when you install pygame from the official wheels.

I guess that means I am giving my :+1: for this PR. The issue is packaging though.

dr0id:
> The issue is packaging though.

Thanks for the explanation. In this case I would say this PR is good to be done.

MyreMylar (approved):
LGTM 👍

PR #3060, opened 2022-02-27
2.1.3.dev4 and 2.1.3dev5

illume:


dr0id (approved):
Only odd numbers? I recall something about even are stable, public and odd are dev versions. Or similar. Or any other reason?

Starbuck5:
> Only odd numbers? I recall something about even are stable, public and odd are dev versions. Or similar. Or any other reason?

It's to distinguish release builds from locally built builds. I think we were inspired by SDL to do it. SDL 2.0.20 released after SDL 2.0.18, for example, and SDL 2.0.19 referred to the builds in the development cycle.

Starbuck5 (approved):
Let's merge this one in.

PR #3061, opened 2022-02-28
Use SDL_WaitEventTimeout, update windowevent handling (Minor event refactor part 1)

### LINKED ISSUE #3036: `pygame.event.wait` always polls
I studied the C source for `pygame.event.wait`, and it always polls for events, waking up every millisecond (or as frequently as the OS will allow) between polls. SDL can sometimes use the appropriate platform function to actually wait for an event. This has two somewhat paradoxical benefits:

1. If the timeout isn't specified, or if it's long, the OS can actually put the process to sleep for long periods of time, saving power.
2. On the other hand, if events are arriving in quick succession, but not all at once, the fixed delay between polls won't be a bottleneck. This will be important for any future implementation of accessibility (e.g. for screen readers) on Windows and Mac, since those platforms' accessibility APIs are based on request/response IPC where the requests must be handled by the OS event pumping mechanism, and a screen reader may have to send several requests in quick succession. IPC overhead is bad enough without adding a 1ms+ delay between round trips.

I don't know how to use the actual `SDL_WaitEventTimeout` function without breaking pygame's custom event filtering, so I'm just pointing out the issue here.

ankith26:
PR to fix #3036

I'm in the process of a minor event refactor. Since my previous larger event refactor I did back in late 2020, I've noticed some things that could have been done better, and now I'll be doing a series of more manageable and easy to review PRs (learning from last time, not going to make another super large try-to-fix-everything PR ;) )
While the goals of the previous refactor was fixing bugs, this one will focus on bugs *and* code quality, I will try to comment and document things better in this PR series.

Coming to this PR, this PR is pretty simple, use `SDL_WaitEventTimeout` and `SDL_PollEvent` over our custom implementation. Doing this has more advantages in the newer SDL version (explained by OP of the linked issue), where `SDL_WaitEventTimeout` saw many good improvements. Our custom implementation only existed to handle `WINDOWEVENT`s, and this PR makes some changes to handle things a bit differently.

This PR should remain compatible with older versions, one minor downside to this PR is that this does not provide any safeguards against users blocking `WINDOWEVENT`, which potentially breaks some internal stuff that rely on checking for `WINDOWEVENT` with an "event watch function". The previous implementation never directly blocked `WINDOWEVENT`s, only filtered *after* pumps, so that event watch functions could use it. This implementation filters blocked events *during* the pump. And there's nothing much we can do about this on our side, except documenting this.

### DIFF
--- a/docs/reST/ref/event.rst
+++ b/docs/reST/ref/event.rst
@@ -373,6 +373,11 @@ On Android, the following events can be generated
    If ``None`` is passed as the argument, ALL of the event types are blocked
    from being placed on the queue.
 
+   Note: Blocking some kinds of events (like windowevents) could potentially
+   break internal pygame machinery that relies on these events being unblocked,
+   so use this function with caution. When unsure, it is better to simply
+   ignore an event during event handling, rather than trying to block the event.
+
    .. ## pygame.event.set_blocked ##
 
 .. function:: set_allowed
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -1619,9 +1619,16 @@ pg_event_poll(PyObject *self, PyObject *_null)
     SDL_Event event;
     VIDEO_INIT_CHECK();
 
-    if (!SDL_PollEvent(&event))
-        return pgEvent_New(NULL);
-    return pgEvent_New(&event);
+    /* SDL_PollEvent seems to be flaky with some unit tests, use peep event */
+    SDL_PumpEvents();
+    switch (PG_PEEP_EVENT_ALL(&event, 1, SDL_GETEVENT)) {
+        case 0:
+            return pgEvent_New(NULL);
+        case 1:
+            return pgEvent_New(&event);
+        default:
+            return RAISE(pgExc_SDLError, SDL_GetError());
+    }
 }
 
 static PyObject *
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -53,6 +53,9 @@
 
 #define PG_GET_LIST_LEN 128
 
+/* Get event type of WINDOWEVENT from event.window.event */
+#define PG_TO_WINDOWEVENT(e) (PGE_WINDOWEVENTSTART + (e).window.event)
+
 // Map joystick instance IDs to device ids for partial backwards compatibility
 static PyObject *joy_instance_map = NULL;
 
@@ -419,16 +422,6 @@ _pg_pgevent_deproxify(Uint32 type)
     return _pg_pgevent_proxify_helper(type, 0);
 }
 
-static int
-_pg_translate_windowevent(void *_, SDL_Event *event)
-{
-    if (event->type == SDL_WINDOWEVENT) {
-        event->type = PGE_WINDOWSHOWN + event->window.event - 1;
-        return SDL_EventState(_pg_pgevent_proxify(event->type), SDL_QUERY);
-    }
-    return 1;
-}
-
 static int SDLCALL
 _pg_remove_pending_VIDEORESIZE(void *userdata, SDL_Event *event)
 {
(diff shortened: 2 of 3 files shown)


robertpfeiffer:
Shouldn't this release the GIL when it does not busy-wait?

ankith26:
It does release GIL in `pygame.event.wait()`. Releasing GIL in `pygame.event.poll()` is kinda pointless because the function is supposed to return almost immediately

docs/reST/ref/event.rst
```diff
@@ -369,6 +369,11 @@ On Android, the following events can be generated
    If ``None`` is passed as the argument, ALL of the event types are blocked
    from being placed on the queue.
 
+   Note: Blocking some kinds of events (like windowevents) could potentially
+   break internal pygame machinery that relies on these events being unblocked,
+   so use this function with caution. When unsure, it is better to simply
+   ignore an event during event handling, rather than trying to block the event.
```
dr0id, line 375: This is diffuse and confusing. Why can't this be dealt with internally?

dr0id (commented):
I do not know the SDL api well enough. Maybe I will learn it some day.

docs/reST/ref/event.rst
robertpfeiffer: Because we don't know if blocking the internal processing of events is intended by the end user or not!

dr0id: How much control should the end user have over the *internal processing* of events? 

Not sure if the end user is aware and expects that some internal pygame machinery breaks or behaves differently when blocking some events (which ones?). I would expect that the only thing that happens when blocking events is that my python code does not receive those events but pygame still behaves 'as usual'.

I might be wrong thought. Since there is a simple workaround (just don't block nor process those events) this fine with me.

ankith26: A quick local test shows that blocking `pygame.KEYDOWN` has no effect on `pygame.key.get_pressed()`. But blocking `WINDOWEVENT` can break some things [here in SDL](https://github.com/libsdl-org/SDL/blob/4b38d4c96b03b6e1e845ee2a0338d78aca652f5a/src/video/cocoa/SDL_cocoametalview.m#L38) as well as in pygame `display.c`.
Problem is that we cannot accurately predict nor document the events whole blocking can affect internal state because it is fairly platform and SDL-version dependent

dr0id: Hm, ok, quick thought: never block WINDOWEVENT  in SDL but if the user blocks it then just filter it out of the events that are returned? I guess the effort to implement this is not worth it.

ankith26: That is pretty much the existing implementation but it means that a custom implementation of `SDL_WaitEventTimeout` has to be used

src_c/event.c
```diff
@@ -1586,53 +1593,36 @@ get_grab(PyObject *self)
...
 _pg_event_pump(int dopump)
 {
     if (dopump) {
         SDL_PumpEvents();
     }
-    /* We need to translate WINDOWEVENTS. But if we do that from the
-     * from event filter, internal SDL stuff that rely on WINDOWEVENT
-     * might break. So after every event pump, we translate events from
-     * here */
     SDL_FilterEvents(_pg_translate_windowevent, NULL);
```
robertpfeiffer, line 1617: why isn't the filtering done in get then?

dr0id: ...and in the `pg_event_wait` method probably too?

ankith26: This filter `_pg_translate_windowevent` called from `_pg_event_pump` runs in `clear`, `get` and `peek`; the kind of functions that deal with the entire event queue and not one at a time. Functions like `wait` and `poll` return atmost one event and don't filter for a particular type of event, so that is handled directly in `pgEvent_New`

MyreMylar (approved):
Only arguable slight problem I noticed was no space between 'window' and 'event' in the event.rst change. All the test pass locally and events still run fine in my GUI applications.

I suppose it'd be good to see some feedback from somebody who uses `event.poll()` and `event.wait()` more rigorously, but from me this gets a 👍

ankith26:
Force pushed without any changes to rebase on main, to get the latest tests from main onto this branch

PR #3062, opened 2022-02-28
Add pg_post_event and dictproxy C API to release GIL, fix timer issues and speedups (Minor event refactor part 2)

ankith26:
Part 2 of event refactor started in #3061

This PR makes a few changes to pygame C API involving event posting from different threads, and after this we should see performance improvements in `mixer.music` and `time.set_timer` callback functions, because GIL has been eliminated in many places (posted benchmarks of this in this PR convo thread).

First step here was to eliminate GIL when the `dict` associated with an event was empty, because it can be special-cased to `NULL`. Then, this PR adds a `DictProxy` API to encapsulate event dicts in a custom struct which has a per-dict spinlock and maintains a count of the number of such events on queue. With this API, GIL is held much less often, only during creation and cleanup of the timer. Cleanups are handled properly with a flag in the dictproxy struct that controls whether the freeing should happen in event handling functions if there are references there, or in the timer callback function.

Did some other code cleanups, one of them being batching the `pg_timer_mutex` holds to top-level functions, which also release GIL. Documented the code better and also added a test to catch more timer bugs.

This system is still not fully robust and needs more testing, there are still a few corner cases, and error-handling from callbacks need to be improved (which is out of scope of this PR) and a few more future improvements to be done, which I'll be doing in the next few PRs. This one should be good to go when there are no critical bugs with this (will report back with some more testing)


MyreMylar (approved):
LGTM 👍  

The tests pass locally on windows and I've given it a good test in an app that uses a lot of custom events. Code looks solid too.

ankith26:
Thanks for the review! Looking back at this PR, I noticed a few more improvements in code I can make, so I'll push a commit for that and also comment stuff better especially around `set_timer` bits. Will try to benchmark the effects of this PR too

ankith26:
Pushed a new commit that documents code better, makes mutex locking system simpler (with locks held directly from top level functions) and other minor misc improvements.

To test the effects of this PR, I benchmarked putting `USEREVENT` on timer, and measured the time taken for each event to arrive.

```py
import json
import time

from rich.progress import track

import pygame

LOOPS = 100
EXPECTED_TIME = 5

pygame.init()
win = pygame.display.set_mode((100, 100))


def test_time():
    """
    Do a busy loop to time the arrival of events as accurately as possible.
    This also acts like a load on the app
    """
    old_time = time.perf_counter() * 1000
    while True:
        if pygame.event.poll().type == pygame.USEREVENT:
            return time.perf_counter() * 1000 - old_time - EXPECTED_TIME


for load_level in (0, 3, 30, 100, 320):
    # load: heavy timers in the background
    for _ in range(load_level):
        pygame.time.set_timer(pygame.event.custom_type(), 1)

    pygame.time.set_timer(pygame.USEREVENT, EXPECTED_TIME, LOOPS)

    times = [test_time() for _ in track(range(LOOPS), f"Load Level {load_level}")]
    mean = sum(abs(i) for i in times) / len(times)
    variance = sum((x - mean) ** 2 for x in times) / len(times)
    std_dev = variance**0.5

    print(f"Mean: {mean} | Standard deviation about mean value: {std_dev}\n")

    with open(f"stats-pygame-{pygame.version.ver}-load-{load_level}.json", "w+") as f:
        json.dump(times, f)
```

In the above code `test_time()` waits for events on a busy loop (this busy loop is a load on the code by itself, and I think this is good because it would simulate average event handling overheads) and returns the deviation from the expected time. 
Then, the test code simulates a "load", which in this case, is heavy timers in the background. The "load_level" is the number of such timer loads running, which is a parameter that will be varied across arbitrary values for testing purposes. 

The timer loops 100 times, mean and standard deviation is calculated and the data is flushed to a json file, which I later used to generate matplotlib graphs. Added `rich.progress.track` for cosmetic reasons and debugging reasons, because the 2.1.2 code runs super slow under heavy load, so it's handy to know the time remaining hehe.

The code gives the below results on my windows 11 laptop.
pygame-2 1 2-run
pygame-2 1 3-run

Mean of deviated time is a measure of the accuracy of the timer, the larger the value, the less *accurate* the timing behaviour.
Standard deviation is a measure of how much spread the data is, the larger this value, the less *precise* the timing behavior.

As the numbers show, 2.1.2 timers seems to be bad to begin with even under 0 external timer load. The mean is almost 6x worse and the standard deviation is almost 9x worse. And increasing the load exponentially widens this gap, the 2.1.3 data seems to be consistent regardless external load (this is mainly because GIL overhead has been eliminated here). In the extreme case of "load" 320 I measured here, 2.1.2 mean is about 1043x worse and standard deviation is about 300x worse. Further increasing the loads only makes things worse, at this point the 2.1.2 code appears to hang. While this optimisation may seem more impressive, the optimisation I care about is the base-case low load optimisation, because that would affect more real code than the extremes of this benchmark, which I doubt real code ever goes to.

graphs
(omitted load 100 graph, the graphs of that were similar to the 320 load one)

Finally, here is a graph of the individual loads directly compared against 2.1.2 and this PR. From this we can say that under both low load and heavy load, 2.1.2 timers are inconsistent in both their precision *and* accuracy when compared to this PR.

I did not benchmark timer posting with a `dict`, I imagine that is sped up by this PR due to other locking improvements I made, but probably not as much as the case without the dict, where the GIL overhead has been elimininated.

ankith26:
Benchmarking timer with a dict attribute, performance seems to have degraded. Found a fix for it, will push another commit for it after more extensive testing and such

ankith26:
Latest commits improved performance of event-with-dict posting too, and now it is much closer to the 2.1.3 benchmarks posted above for non-dict events. Added a unit test for more extensive testing, which helped catch a few bugs in the code.
Code-wise, this PR is almost done now, after a bit more testing, I might add another unittest or two if I find something.

ankith26:
I did some testing with a few apps that used some timers, and this PR should be done from my end

There are still a few minor issue fixes/changes I'd like to make, but like I said earlier I'd like to make them in future PRs because this PR already ended up a bit bigger than I initially expected it to be

ankith26:
Force pushed with a minor change, and update from `main`

ankith26:
The latest force pushes are for fixing merge conflicts, and minor emscripten compile issues

MyreMylar (approved):
Still looking good 👍

src_c/event.c
```diff
@@ -928,14 +971,37 @@ dict_from_event(SDL_Event *event)
...
         return NULL;
 
+    /* check if a proxy event or userevent was posted */
+    if (event->type >= PGPOST_EVENTBEGIN) {
+        int is_freed;
+        pgEventDictProxy *dict_proxy = (pgEventDictProxy *)event->user.data1;
+        if (!dict_proxy) {
+            /* the field being NULL implies empty dict */
+            return dict;
+        }
```
nthykier, line 985: Nit: As a slight optimization, we could move the `dict = PyDict_New();` down below the `if (event->type >= PGPOST_EVENTBEGIN) {` and then replace `return dict;` with `return PyDict_New();`).

It would avoid an unnecessary allocation when we are processing a dict proxy and would have the same speed otherwise (technically, it would be one `if`-less, but I am not sure we are in that hot of a code path).

src_c/_pygame.h
```diff
@@ -42,6 +42,19 @@
...
+ * state of number of events on queue and whether the owner of this struct
+ * wants this dict freed. This DictProxy is only to be freed when there are no
+ * more instances of this DictProxy on the event queue. Access to this is
+ * safeguarded with a per-proxy spinlock, which is more optiminal than having
+ * to hold GIL in case of event timers */
+typedef struct _pgEventDictProxy {
+    PyObject *dict;
+    SDL_SpinLock lock;
+    int num_on_queue;
+    Uint8 is_freed;
```
nthykier, line 55: Looking at the code, I feel that the name `is_freed` is not helping me understand the purpose.

As far as I can gather, it is basically `free_when_queue_counter_is_zero`... but as far as names go, it has its own problem.  Maybe a `proxy_source_is_closed` if the argument is we should free it because the proxy will never be emitted in an event again.

nthykier: Maybe, I am also missing a comment describing when `num_on_queue` would be `>0` but `is_freed` would be true.  I can see there is code for it but mentally wrapping my head around when this can happen in practice is a different matter.

ankith26:
@nthykier Thanks for taking the time and effort to review this PR. Unfortunately, I don't see this PR going ahead anytime soon, so I'm closing it for now. However I do have this branch saved locally, and if/when I get back to it I will surely consider your reviews, and also comment my code better :)

PR #3063, opened 2022-03-01
Event tests and types improvements (Minor event refactor part 3)

ankith26:
Part 3 of event refactor started in https://github.com/pygame/pygame/pull/3061

This PR aims to improve the tests of the event module, and fix any incorrect typestubs

The tests added are intended to work on both older pygame versions and after my minor event refactor. So this PR should be merged first in the series.


test/event_test.py
```diff
@@ -834,6 +870,57 @@ def test_poll(self):
...
+            (65, pygame.USEREVENT, 50),
+            (20, pygame.NOEVENT, 20),
+            (45, pygame.USEREVENT, 30),
+            (70, pygame.NOEVENT, 70),
+        ):
+            start_time = time.perf_counter()
+            self.assertEqual(pygame.event.wait(wait_time).type, expected_type)
+            self.assertAlmostEqual(
+                time.perf_counter() - start_time, expected_time / 1000, delta=0.01
+            )
```
Starbuck5, line 910: I'd rather have the test have wait_time == expected_time, with a greater delta.

I don't like having a test that is like "oh, lets wait 60 ms for this, but we expect it to happen in 50ms, and if it took 61ms the test will fail."

buildconfig/stubs/pygame/event.pyi
```diff
@@ -17,10 +17,9 @@ class Event:
     dict: Dict[str, Any]
     __dict__: Dict[str, Any]
     __hash__: None  # type: ignore
-    @overload
-    def __init__(self, type: int, dict: Dict[str, Any]) -> None: ...
-    @overload
-    def __init__(self, type: int, **attributes: Any) -> None: ...
+    def __init__(
+        self, type: int, dict: Dict[str, Any] = ..., **kwargs: Any
```
Starbuck5, line 21: What does having some default to ellipses mean in type annotation?

test/event_test.py
ankith26: Well, we can't pass the exact expected time to `pygame.event.wait`, we have to give some delta for that function itself

buildconfig/stubs/pygame/event.pyi
ankith26: It indicates that the argument is a "default argument" but does not specify the default type

test/event_test.py
Starbuck5: Why not? Just use the delta time in assertAlmostEqual.

Function calling overhead should be insignificant in the order of milliseconds, if that's what you mean. Also the wait vs expected gaps you have so far don't seem to be a consistent pattern.

ankith26: what time is to be passed to the `pygame.event.wait` then? If the exact duration is passed, the event can be missed due to timing inaccuracies. A delta time must be passed to both `event.wait` and the corresponding `assertAlmostEqual` time check.

I wrote this block to try to test a wider range of stuff, including cases like waiting for a bit longer, waiting a bit less and missing an event which is then caught in the next wait loop. That's why there's no consistent pattern here

ankith26: I understand this test is magic-number-ey, but I feel this is a good way to test timing related stuff

Starbuck5: Alrighty, discord discussion straightened me out. I thought you were posting them and checking to see that you received them in a good time, but it makes sense now I realize you're grabbing events posted by a timer.

Starbuck5 (approved):
I couldn't poke any holes in it, great!

test/event_test.py
```diff
@@ -834,6 +870,57 @@ def test_poll(self):
         self.assertEqual(pygame.event.poll().type, pygame.NOEVENT)
 
 
+class EventModuleTestsWithTiming(unittest.TestCase):
+    __tags__ = ["timing"]
+
+    def test_event_wait(self):
```
MyreMylar, line 876: This test fails when run locally on windows:

```
ERROR: test_event_wait (__main__.EventModuleTestsWithTiming)
Ensure wait() waits for an event on the queue.
  File "C:\Users\dan\Programming\pygame\test\event_test.py", line 880, in test_event_wait
    pygame.event.post(event)

```

I assume it needs a `pygame.display.init()` in the setup somewhere.

MyreMylar (changes requested):
LGTM, just noticed one apparently failing test when run locally on windows.

test/event_test.py
ankith26: Oh yeah, good catch! I had tested this function before I moved it into its own class and it worked then, but I missed `setUp` and `tearDown` functions, which I've added now!

MyreMylar (approved):
Works now 👍  All looks good to merge

ankith26:
Thanks for the reviews! Will merge this one in now first because it's tests

PR #3064, opened 2022-03-01
Add docstrings to modules in typestubs

ankith26:
Pretty self explanatory PR title, adds docstings to typestub pyi files, so that editors can display useful information about modules on mouse hover for user convenience. These docstrings are picked up from `__doc__` attribute of the modules.

~~PR also fixes a minor issue https://github.com/pygame/pygame/pull/2999 introduced, when hovering over `pygame` base module, editor displays irrelevant docstrings about the typestub file being a generated file (the "generated file" comment has been moved into a separate comment in this PR)~~ This has been fixed in another one of my PRs

### Things to discuss

- [ ] Should all the typestub classes and functions get docstrings in typestubs too? While this would mean that text editors can display useful information on hover, it also means more maintenance effort because there would be 2 places to update doc strings on future changes. This sort of thing could be hard to automate with a script, but it is theoretically possible.


buildconfig/stubs/pygame/fastevent.pyi
```diff
@@ -1,3 +1,10 @@
+"""
+A compatibility shim for pygame.fastevent based on pygame.event.
+This module was deprecated in pygame 2.2, and is scheduled for removal in a
+future pygame version. If you are using pygame.fastevent, please migrate to
+using regular pygame.event module
```
novialriptide, line 5: Shouldn't this be "please migrate to using the regular pygame.event module"?

buildconfig/stubs/pygame/joystick.pyi
```diff
@@ -1,3 +1,7 @@
+"""
+Pygame module for interacting with joysticks, gamepads, and trackballs.
```
novialriptide, line 2: There shouldn't be a capital in "Pygame"

novialriptide (changes requested):
I added some docstring suggestions.

buildconfig/stubs/pygame/fastevent.pyi
dr0id: probably yes, but this is generated so the 'source' of this text is somewhere else.

ankith26:
Good review suggestions, but in this PR I merely copied `__doc__` of module objects into the typestubs, so `__doc__` would need fixing first. Also I'm still not sure whether we want to go this way, because this way can add a bit more maintenance, so we need to discuss this PR and any other possible alternatives first

buildconfig/stubs/pygame/fastevent.pyi
MyreMylar: Agreed, there is a missing 'the'

MyreMylar (approved):
LGTM 👍  couple of tiny tweaks that were pointed out already.

MyreMylar:
PyCharm already picks up the existing documentation on hover btw.

image

So I'm not sure which code editor these changes are helping, but I wouldn't spend ages doubling up documentation if it is only helping a small fraction of users.

ankith26:
Well this PR was intended to fix the doc strings that you get when you hover over modules, on VS Code.

I'm not really sure how to proceed on this PR, because at the moment it just adds doc repetition which is not ideal from a maintenance POV. Maybe this is something to be fixed upstream in one of the editor tools. 

Till then let this PR be a draft

illume (approved):
Nice improvement.

Yeah, I think someway to keep the docstrings in sync would be nice. But also we don't have that sync tool now... so it would be better to merge without it.

ankith26:
Thanks for the approval

Hmmm... might try to come up with a little script to automate this

PR #3066, opened 2022-03-01
Rect and Color support collections.abc.Container, add tests and stubs

### LINKED ISSUE #2681: Rect is iterable but not typed or documented

ankith26:
PR to fix #2681

After this PR, there should be minimal-to-none changes in observed behaviour. Existing code will behave the same, but `__iter__` method will be present on `Rect` and `Color`, and `__contains__` will be present on `Color`, `Vector2` and `Vector3`. 

In the absence of `__iter__` or `__contains__`, python falls back to using `__getitem__` for iteration and membership tests. Since explicit is better than implicit, this PR adds the methods to be defined explicitly and this way, they can also be typehinted to benefit typecheckers.
For now, a compatibility `__contains__` implementation has been added to vectors and `Color` that behave the same way as they did before, but there is room for future changes if we want to give special behavior for `in` checks on these objects, so this (`__contains__` behaviour) can remain undocumented for now

Along with that, `Rect`, `Color` and vectors now support the `collections.abc.Collection` (this is distinct from `Sequence`, `Sequence` is narrower base class than `Collection`) because they logically make sense as a collection, and this should help with any runtime checks users would want to implement 

After this PR, mypy will stop complaining about code expecting these objects to iterate.


dr0id (approved):
LGTM

dr0id:
not sure why some of the builds failed

Starbuck5:
I believe objects that implement getitem and len implement "iter" automatically. So I'm not sure why we should change our source code for this.

ankith26:
Without `__iter__`, the objects can iterate but they won't confirm to the iterable protocol, and runtime checks for iterable objects would fail. This PR just explicitly declares these objects to iterate, and adding typestubs for `__iter__` means that mypy will stop complaining about anything iteration related (like `x, y, w, h = rect_obj`)

An alternative approach was to simply add `__iter__` in typestubs but not in C code, but this would be wrong, typestubs are meant to be as close to the implementation as possible. Also it's worse for mypy to allow incorrect code to pass, than error at correct code

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -1,11 +1,14 @@
...
 
 from ._common import Coordinate, Literal, RectValue
 
 _K = TypeVar("_K")
 _V = TypeVar("_V")
 
-class Rect:
+# Rect confirms to the Collection ABC, since it also confirms to
+# Sized, Iterable and Container ABCs
+class Rect(Collection):
```
zoldalma999, line 11: Same question here as with color

src_c/rect.c
```diff
@@ -980,14 +980,18 @@ pg_rect_contains_seq(pgRectObject *self, PyObject *arg)
...
+
         return coord == self->r.x || coord == self->r.y ||
                coord == self->r.w || coord == self->r.h;
     }
     int ret = _pg_rect_contains(self, arg);
     if (ret < 0) {
-        PyErr_SetString(PyExc_TypeError,
-                        "'in <pygame.Rect>' requires rect style object"
-                        " or int as left operand");
+        /* happens for a wrong type, return 0 */
```
zoldalma999, line 993: Are we sure we want to remove the error here? Not sure what's the convention for a collection with specific types, but for example strings do raise an error in such cases.

zoldalma999 (commented):
Left a few questions

buildconfig/stubs/pygame/color.pyi
```diff
@@ -1,10 +1,13 @@
...
+from typing import Any, Dict, Iterator, Tuple, overload
 
 from ._common import ColorValue
 
 THECOLORS: Dict[str, Tuple[int, int, int, int]]
 
-class Color:
+# Color confirms to the Collection ABC, since it also confirms to
+# Sized, Iterable and Container ABCs
+class Color(Collection):
```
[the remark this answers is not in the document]
ankith26: Yup, the idea behind this PR is to make some runtime checking better along with typechecking

PR #3067, opened 2022-03-01
Use static dot declarations for type objects

### LINKED ISSUE #2176: PyTypeObject static object dot declarations
**The declarations for PyTypeObjects should use dot notations**

This increases readability, cuts down potential bugs when migrating to new python versions, i.e. Python 3.9.  specifying 0's in these static objects serves no good purpose and is bug-prone.  I have patches for all of the 1.9.7 src_c files (done while patching 1.9.7 for Python 3.9) and could easily update them for the 2.x branch.

`+    TYPE_HEAD(0, 0)
+	.tp_name = PROXY_TYPE_FULLNAME,
+    .tp_basicsize = sizeof(pgBufproxyObject),
+    .tp_dealloc = (destructor)proxy_dealloc,
+    .tp_repr = (reprfunc)proxy_repr,
+    .tp_as_buffer = PROXY_BUFFERPROCS,
+    .tp_flags = PROXY_TPFLAGS,
+    .tp_doc = DOC_PYGAMEBUFFERPROXY,
+    .tp_traverse = (traverseproc)proxy_traverse,
+    .tp_weaklistoffset = offsetof(pgBufproxyObject, weakrefs),
+    .tp_methods = proxy_methods,
+    .tp_getset = proxy_getsets,
+    .tp_dictoffset = offsetof(pgBufproxyObject, dict),
+    .tp_alloc = PyType_GenericAlloc,
+    .tp_new = proxy_new,
+    .tp_free = PyObject_GC_Del,
`

ankith26:
PR to fix #2176

### DIFF
--- a/src_c/color.c
+++ b/src_c/color.c
@@ -220,61 +220,32 @@ static PyGetSetDef _color_getsets[] = {
     {NULL, NULL, NULL, NULL, NULL}};
 
 static PyNumberMethods _color_as_number = {
-    (binaryfunc)_color_add,  /* nb_add */
-    (binaryfunc)_color_sub,  /* nb_subtract */
-    (binaryfunc)_color_mul,  /* nb_multiply */
-    (binaryfunc)_color_mod,  /* nb_remainder */
-    NULL,                    /* nb_divmod */
-    NULL,                    /* nb_power */
-    NULL,                    /* nb_negative */
-    NULL,                    /* nb_positive */
-    NULL,                    /* nb_absolute */
-    NULL,                    /* nb_nonzero / nb_bool*/
-    (unaryfunc)_color_inv,   /* nb_invert */
-    NULL,                    /* nb_lshift */
-    NULL,                    /* nb_rshift */
-    NULL,                    /* nb_and */
-    NULL,                    /* nb_xor */
-    NULL,                    /* nb_or */
-    (unaryfunc)_color_int,   /* nb_int */
-    NULL,                    /* nb_reserved */
-    (unaryfunc)_color_float, /* nb_float */
-    NULL,                    /* nb_inplace_add */
-    NULL,                    /* nb_inplace_subtract */
-    NULL,                    /* nb_inplace_multiply */
-    NULL,                    /* nb_inplace_remainder */
-    NULL,                    /* nb_inplace_power */
-    NULL,                    /* nb_inplace_lshift */
-    NULL,                    /* nb_inplace_rshift */
-    NULL,                    /* nb_inplace_and */
-    NULL,                    /* nb_inplace_xor */
-    NULL,                    /* nb_inplace_or */
-    (binaryfunc)_color_div,  /* nb_floor_divide */
-    NULL,                    /* nb_true_divide */
-    NULL,                    /* nb_inplace_floor_divide */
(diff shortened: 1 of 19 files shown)


dr0id:
Never seen this before on a build agent:
```
 /io/buildconfig/manylinux-build/build-wheels.sh: line 60:  6067 Segmentation fault      (core dumped) ${PYTHON} -m pygame.tests -vv --exclude opengl,music,timing
Makefile:13: recipe for target 'wheels-manylinux2010-x86' failed
make: *** [wheels-manylinux2010-x86] Error 1
Error: Process completed with exit code 2.
```

Starbuck5 (approved):
This looks good to me. I ran tests locally.

The python type objects docs use this style for their examples, so it's good to emulate. And it makes the code nicer too.

One thing I noticed is that some types have explicit tp_flags to `Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE`. But the python type object docs suggest that is just the default value. Some other things do weird things with tp_flags, like `color` uses a value set somewhere else-- I didn't check the rest of the code but it looks weird.

Additionally, I'm not sure this should be merged in 2.1.3, since we've done two test releases already. But it is an unknown situation about when that will be released, as I'm sure you know.

dr0id (approved):
It just a different syntax for the same thing. Not sure I like it. On one hand now the assignments are clearer due to the variable name, on the other hand one has to look up the possible things (where it was documented in the old way). But since this is just preference I approve it too.

ankith26:
Yeah I believe `Py_TPFLAGS_CHECKTYPES` can be safely removed now. Because it is defined to `0` in a compat header. Also I'll make a commit simplifying all `tp_flags` everywhere

dr0id (approved):
LGTM

ankith26:
The latest commit ads static dot declarations for more structs, mainly for consistency reasons.
The rationale behind this PR is to make these structs easier to maintain in the long run, and make it less error prone, because the current system heavily relies on maintaining the right order of stuff. I believe these new struct declarations will also make it easier to new contributors who are more familiar with the same. It also makes the code more concise and easier to understand IMHO

But we still need checking for whether this can break on any old compiler on any exotic configuration, which we support (I don't think this is the case because this PR seems to pass on CI, the fail ATM is unrelated)

src_c/color.c
```diff
@@ -220,119 +220,54 @@ static PyGetSetDef _color_getsets[] = {
...
-    _color_getsets,        /* tp_getset */
-    NULL,                  /* tp_base */
-    NULL,                  /* tp_dict */
-    NULL,                  /* tp_descr_get */
-    NULL,                  /* tp_descr_set */
-    0,                     /* tp_dictoffset */
-    (initproc)_color_init, /* tp_init */
-    NULL,                  /* tp_alloc */
-    _color_new,            /* tp_new */
-#ifndef __SYMBIAN32__
```
dr0id, line 326: is id different for symbian? Or id this not used anymore?

ankith26: We probably don't actively support symbian ATM, atleast to my knowledge, so I think this change should be fine

dr0id (approved):
I start to like the new notation, your argument about the position is the reason for it. Also it is much more compact.

PR #3074, opened 2022-03-12
Improve draw doc accuracy and example style

### LINKED ISSUE #3048: Fix up draw.rect docs
The draw.rect documentation needs to be fixed to reflect https://github.com/pygame/pygame/pull/2927

Specifically this bit:

> Note When using width values > 1, the edge lines will grow outside the original boundary of the rect. For more details on how the thickness for edge lines grow, refer to the width notes of the [pygame.draw.line()draw a straight line](https://www.pygame.org/docs/ref/draw.html#pygame.draw.line) function.

Which is no longer true.

> Note The [pygame.Surface.fill()fill Surface with a solid color](https://github.com/pygame/pygame/issues/surface.html#pygame.Surface.fill) method works just as well for drawing filled rectangles and can be hardware accelerated on some platforms with both software and hardware display modes.

^ This should also be revised to drop hardware acceleration mention, just say it works well for filling a whole surface and may be faster.

Starbuck5:
Closes #3048 

- Gets rid of mention of the old draw.rect behavior, before it was changed to be an actual rectangle
- Gets rid of a couple "hardware mode" mentions that are definitely not relevant to current pygame
- Moves away from colors as constants model in the example code, instead using more modern pygame strategies.
- `black`s draw example code (This is not really an important thing, readability is more important for newbies than pep-8 correctedness)


docs/reST/ref/code_examples/draw_module_example.py
```diff
@@ -1,94 +1,92 @@
-# Import a library of functions called 'pygame'
 import pygame
```
dr0id, line 1: if you want more readability and a bit more clean code then there are too many 'magic constants'. Code should be self explanatory without having to have to look up the function signatures. But is fine with me except for the arc call, see comments there.

Resolve at will.

docs/reST/ref/code_examples/draw_module_example.py
```diff
@@ -1,94 +1,92 @@
...
+    pygame.draw.polygon(screen, "black", [[100, 100], [0, 200], [200, 200]], 5)
+
+    # Draw an arc as part of an ellipse.
     # Use radians to determine what angle to draw.
-    pygame.draw.arc(screen, BLACK,[210, 75, 150, 125], 0, pi/2, 2)
-    pygame.draw.arc(screen, GREEN,[210, 75, 150, 125], pi/2, pi, 2)
-    pygame.draw.arc(screen, BLUE, [210, 75, 150, 125], pi,3*pi/2, 2)
-    pygame.draw.arc(screen, RED,  [210, 75, 150, 125], 3*pi/2, 2*pi, 2)
-    
+    pygame.draw.arc(screen, "black", [210, 75, 150, 125], 0, pi / 2, 2)
```
dr0id, line 73: For beginners radians may be difficult to understand. I would propose to use `math.radians()` here instead. They can look up radians later,

dr0id (approved):
Approved but left some comments. Resolve them if you wish of leave it as it is or drop a comment if you disagree.

ankith26 (commented):
I'm not really sure about this. Can change in this behaviour have broken existing code, which expected drawn rects to grow inwards? Since I imagine rect drawing is a widely used function, and it was documented that drawing rects with widths would expand inwards, this seems like an API detail that should not change like this.

Starbuck5:
> I'm not really sure about this. Can change in this behaviour have broken existing code, which expected drawn rects to grow inwards? Since I imagine rect drawing is a widely used function, and it was documented that drawing rects with widths would expand inwards, this seems like an API detail that should not change like this.

We already changed it though, and I don't think we're going to change it back, so no reason to not update the docs. And I do think the rect draw change is good for the long run. They're actual rectangles now!

docs/reST/ref/code_examples/draw_module_example.py
Starbuck5: I was thinking about that but I didn't want confusing line breaks, as the line would like be too long then.

dr0id: ok, this sounds reasonable and one way or another they need to learn about radians too :)

ankith26 (approved):
LGTM! Thanks 🎉 
While I was initially reluctant to accept this change because it seemed like an API change (because of the visual change in drawing), now I'm approving this PR after some discussion, on the basis that the new behavior seems more consistent and logical.

docs/reST/ref/draw.rst
```diff
@@ -64,12 +63,9 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
          | if ``width > 0``, used for line thickness
          | if ``width < 0``, nothing will be drawn
          |
+      
+      .. versionchanged:: 2.1.1 Drawing rects with width now draws the width correctly inside the rect's area, rather than using an internal call to draw.lines(), which had half the width spill outside the rect area.
```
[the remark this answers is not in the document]
Starbuck5: That's what I was trying to talk to you about on discord! I don't know how / if there is a way to put one of these tags on multiple lines!

Starbuck5:
Thank you for the approval

> it seemed like an API change

All I'm doing is documenting the older PR, as I'm sure you know. The decision was made then.

docs/reST/ref/draw.rst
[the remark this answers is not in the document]
ankith26:
```suggestion
      .. versionchanged:: 2.1.1 
          Drawing rects with width now draws the width correctly inside the 
          rect's area, rather than using an internal call to draw.lines(), 
          which had half the width spill outside the rect area.
```
Should do?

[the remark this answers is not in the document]
Starbuck5: Ugh, idk why I was certain that wouldn't work.

I bet sphinx 1.8 or something didn't work with it, I swear I've tried and failed with this before.

PR #3075, opened 2022-03-12
Fix more typestub warnings with dunders

ankith26:
Since the release of the latest mypy version `0.940`, mypy stubtest has started to test dunders correctly. A follow up PR on #2999 which fixed most stubtest warnings, this one fixes some incorrect dunders.

Also marks unsubclassable pygame classes with `@final`, mypy has started to correctly warn about this

Code changes are minimal, `__PYGAMEinit__` and `__PYGAMEquit__` internal functions have been renamed to start with a single underscore, which is the recommended way of defining a private function.

Without this PR, our ubuntu sdist builds will keep failing going forward so this PR is semi-urgent ;)


dr0id (approved):

Starbuck5 (approved):
Cool stuff, looks good.

"semi urgent" is my middle name

PR #3076, opened 2022-03-12
I tried cleaning Scale2X but I have a few questions

mariospapaz:
function got dropped to 47 lines from 146, the logic should be the same with a small difference.

Since I removed the switch that contained all scenarios, I had no choice but to pick Uint32 as the main holder of `E0, E1, E2, E3, B, D, E, F, H;`
if the bytes per pixel of the sdl surface is 1 or  2, the remaining bits of Uint32 should be filled with zeroes. So I believe I have not increased memory.

I need further details regarding `READINT24` and `WRITEINT24` macros and why in case 3 you used `int` but in `default:` you used `Uint32`

### DIFF
--- a/src_c/scale2x.c
+++ b/src_c/scale2x.c
@@ -48,10 +48,8 @@
   blindly assume you didn't flounder.
 */
 
-void
-scale2x(SDL_Surface *src, SDL_Surface *dst)
+void scale2x(SDL_Surface* src, SDL_Surface* dst)
 {
-    int looph, loopw;
 
     Uint8 *srcpix = (Uint8 *)src->pixels;
     Uint8 *dstpix = (Uint8 *)dst->pixels;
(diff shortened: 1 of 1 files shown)


src_c/scale2x.c
```diff
@@ -61,139 +59,36 @@ scale2x(SDL_Surface *src, SDL_Surface *dst)
...
-                                    (2 * MAX(0, loopw - 1)));
-                    E = *(Uint16 *)(srcpix + (looph * srcpitch) + (2 * loopw));
-                    F = *(Uint16 *)(srcpix + (looph * srcpitch) +
-                                    (2 * MIN(width - 1, loopw + 1)));
-                    H = *(Uint16 *)(srcpix +
-                                    (MIN(height - 1, looph + 1) * srcpitch) +
-                                    (2 * loopw));
+    int bytes_ppixel = src->format->BytesPerPixel; 
+    Uint32 E0, E1, E2, E3, B, D, E, F, H;
+    // I had to sacrifice Uint16/8. the compiler shall fill the remaining area with zeroes
```
dr0id, line 64: `// I had to sacrifice Uint16/8. the compiler shall fill the remaining area with zeroes`
I'm not sure about the compiler filling the zeros. IIRC you must initialize if manually to the desired bits and bytes.

Also: when it comes to bytes the question comes up: does it still work with little- and/or big-endian platforms?

mariospapaz: hopefully it should work!

mariospapaz:
This is the test sample of my refactored scale2x algorithm.
What's changed is that it's 2% faster and more readable.

or to be more precise. 1.9% 
image

**Visual presentation:**

My version:
image

Image from 2.0.1 stable:
image

Continuing the conversation from social media. Both of these algorithms share
an issue, the yellow line is wrong.
We found it is not an issue of code but rather the algorithm itself.  🙂 


Since the parent code has an issue, I will assume mine cannot continue its path.
Feel free to close it since the problem is unrelated with this PH. Inform me if I have to post an issue.
In my opinion I think scale2xraw should replace scale2x instead of maintaining 2 portions, since hq2x algorithm is much better~ and faster.

dr0id:
> In my opinion I think scale2xraw should replace scale2x instead of maintaining 2 portions, since hq2x algorithm is much better~ and faster.

In my understanding: scale2xraw is not hq2x. hq2x is an alternative to [scale2x](http://www.scale2x.it/) (from advanceMame project). They are aimed to scale pixel art in a special way. Also it is claimed that scale2x is faster than hq2x (not sure how important this is nowadays). 

I would find it cool if pygame would support more pixel art special scaling alorithms (like incorporating hq2x, hq3x, hq4x, [see hqx source](https://github.com/grom358/hqx) and maybe others too) but not sure about the effort to implement and maintain this. Also not sure if they are really that useful in the days where hardware acceleration plays a bigger role. On the other hand pygame is still about surfaces with pixels on it...

PR #3077, opened 2022-03-12
Features/rect/scaleby

dr0id:
Take over branch for PR #2409.

Implementation followed the `scale_by` from PR #2723.


dr0id:
Please credit [mcpalmer1980](https://github.com/mcpalmer1980) for this work. 

I do not know how to add participants.

dr0id:
In PR #2409 which this is the follow up it is pointed to the other PR #2723 for implementation details.

I'm not sure if one parameter (potentially a tuple) is the right way to go because the `inflate` method also takes two parameters. The other solution would be to allow the same signature in `inflate` too. Not sure what would be more 'consistent' overall.

Please advise.

src_c/rect.c
```diff
@@ -421,6 +421,38 @@ pg_rect_inflate_ip(pgRectObject *self, PyObject *args)
...
+        (int)(self->r.y + (self->r.h / 2) - (self->r.h * factor_y / 2));
+    self->r.w = (int)(self->r.w * factor_x);
+    self->r.h = (int)(self->r.h * factor_y);
+    Py_RETURN_NONE;
+}
+
+static PyObject *
+pg_rect_scale_by(pgRectObject *self, PyObject *args)
+{
+    pgRectObject *returnRect = (pgRectObject *)_pg_rect_subtype_new4(
```
dr0id, line 450: `returnRect` <- is there a naming convention for variables? 
It looks like all lowercase and without underline thought.

I know, I'm inconsistent right now, line 427, should probably be `float, factorx, factory = 0;` and `returnrect` here.

Starbuck5:
I've been a bit leery on this feature addition proposal for a while now.

It's not much code to emulate this without the feature, it's not fully clear what the use case is, and it's not fully clear what the arguments should be (does it need an origin to scale around)

mcpalmer's use case has been the _sdl2.video API, but that could change to not necessitate so much rect scaling.

dr0id:
> I've been a bit leery on this feature addition proposal for a while now.
> 
> It's not much code to emulate this without the feature, it's not fully clear what the use case is, and it's not fully clear what the arguments should be (does it need an origin to scale around)
> 
> mcpalmer's use case has been the _sdl2.video API, but that could change to not necessitate so much rect scaling.

I see. The origin idea is nice but complicates things a bit but would be cool. I'm more worried about consistency: looking at Rect.Inflate this should also need to parameters (always). As Inflate uses the rect center so this should too by default. I look at it this way: its like inflate but just by a factor instead of a given offset.

On the other hand the discussion about the parameters already happened in the other PR about transform.scale_by.

After taking another look at the methods of the two modules:
 
  - in `transform` all the comparable methods have a size parameter (single parameter, two values)
  - in `rect` all the comparable methods have two parameters

So this should probably work with two parameters.

PR #3080, opened 2022-03-15
fix/norm function pointer calls for METH_NOARGS

pmp-p:
- FIXES ( ie when *_null arg is missing in C)  are related to Web Assembly port ( most of it currently working ).
some background here https://blog.pyodide.org/posts/function-pointer-cast-handling/

- the rest is a normalization of METH_NOARGS conventions, *noarg was used only once here https://github.com/pmp-p/pygame-wasm/blob/37cc2e82ebba825dc90aedbd7ee620b93931a8ee/src_c/music.c#L447

once done i will be able to add fixes to support better Limited API and full wasm support which must be reviewed more carefully than this one.

As the current changes proposed where made guided by a checker (https://gist.github.com/pmp-p/5641b7edfffba457d3eed875454ab5ca)

### DIFF
--- a/src_c/_camera.c
+++ b/src_c/_camera.c
@@ -150,7 +150,7 @@ surf_colorspace(PyObject *self, PyObject *arg)
 
 /* list_cameras() - lists cameras available on the computer */
 PyObject *
-list_cameras(PyObject *self, PyObject *arg)
+list_cameras(PyObject *self, PyObject *_null)
 {
 #if defined(__unix__) || defined(PYGAME_WINDOWS_CAMERA)
     PyObject *ret_list;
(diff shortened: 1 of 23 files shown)


dr0id:
Thanks for the PR and trying to make pygame better.

The only long-term solution I see is to use some checker script in the CI builds. Otherwise those errors will creep in again.

Starbuck5:
> The only long-term solution I see is to use some checker script in the CI builds. Otherwise those errors will creep in again.

I'm 100% spitballing here, but what if one of pygame's base header files undefined or invalidated the actual things you would use to cast functions, forcing people to write the functions without casts.

pmp-p:
> This PR doesn't remove function pointer casting fully, just makes it an easier cast for Webassembly

indeed, it just try to fix what is completely illegal in web assembly, trying to leave existing code as possible untouched.

Starbuck5 (approved):
Let's do it!

Thanks for working on this pmp-p!

The funny thing is that this actually might be a code quality enhancement, since it gives a consistent way to express the PyObject in the slot isn't and shouldn't actually be used.

ankith26 (approved):
LGTM! Thanks for the PR 👍 😄 

As mentioned above, it might be a good idea to automate testing this on CI somehow, but apart from that, this PR is nice and it makes things more consistent!

PS: The Ubuntu fail on this PR is unrelated to this issue, it should already be fixed on `main` branch

Issue #3023, opened 2022-02-10
Update to SDL_ttf 2.0.18


Starbuck5:
Porting the multiline rendering stuff is another issue, but first we need to update to SDL_ttf 2.0.18 in the first place.

SDL_ttf distribution is now compiled with freetype and harfbuzz statically.

This is a problem for windows since we get our modern freetype from the SDL_ttf package. Also, if we wanted to minimize file size we would want pygame.freetype and pygame.font -> SDL_ttf to share the underlying freetype library.

So this might be a good time to replace the old "pygame 1.9.2" prebuilt downloads that still gets factored in by Windows buildconfig, especially now that SDL1 support has been dropped, which should simplify things.

Tasks:
- [ ] Linux buildconfig
- [ ] Mac buildconfig
- [x] Windows buildconfig

ankith26:
Your linked PR handled windows buildconfig, so my PR that updates mac/linux buildconfig should close this issue ;)

Issue #2574, opened 2021-04-30
Multiline support missing


Loremoth:
It seems very strange to me that a famous videogame library like this misses simple support for writing multiline text on screen. I hope it will be fixed soon...

Starbuck5:
It's not like pygame has just forgotten to do this, it's more that "GUI" stuff has been seen as "out of scope" for pygame itself.

However, this could probably be doable. There could be a `wrap_pixel` keyword argument on font.render and freetypefont.render. But would it also have to handle `\n` and `\t` characters?

I wonder what people who have written GUI libraries for pygame think. Could this simplify / optimize your workflow? Should pygame add this? What sort of interface should it use? @ppizarror @MyreMylar

ppizarror:
I think there's plenty of libraries that wrap text, for example ``textwrap`` (used by pygame-menu); having a simple core API is better because this kind of feature would not be used for many people, as in general a GUI library is preferred.

For example, if that ``wrap_pixel``is implemented, what is the distance between lines? (so, line height should also be provided). And about new lines (\n, \t), or language support (for word-breaking). And what about RTL-languages, and the text remaining these lines, they must be centered, justified, left, right? There's a lot of things to wonder. Generally this high-level features introduce lots of problems.

BastiHz:
For anyone who wants to implement this for themselves there is a helpful example in the docs: https://www.pygame.org/docs/ref/freetype.html#pygame.freetype.Font.render_to

robertpfeiffer:
In the past this would have been closed as wontfix, because it can easily be done in pure Python, and typesetting whole paragraphs is definitely out of scope.

ankith26:
I agree, this kind of feature is probably best implemented in user code, or a 3rd party pygame library. pygame has always been a lower level tool which you can use to build your engine and game on, and this feature would need a higher level API, with the existing pygame API it would not be elegant to integrate this feature. There are many good libs that use pygame and provide this feature.

Starbuck5:
I disagree, I think this functionality can and should be added into base pygame.

It's a big edge for users to go from rendering single lines of text to full GUI libraries. And it seems awfully inconvenient for users to have to reimplement the same functionality over and over, if they aren't aware of GUI libraries or don't wish to use them.

The fact that we have reference code on how to do this supports that it is a common operation that we can abstract.

[...]

Issue #464, opened 2018-06-29
Indic Scripts Including Hindi, Telugu, Tamil and many more not rendering as they supposed to be.


lifeisshubh:
When I am trying to render Hindi text using to_render(). It is rendering inappropriately. for eg. 
![image](https://user-images.githubusercontent.com/8062446/42087683-87adb820-7bb4-11e8-9a11-ceb747f75673.png) which should be written like 
![image](https://user-images.githubusercontent.com/8062446/42087836-1195aaf2-7bb5-11e8-9b1c-b47fc5db05f4.png) . Also like 
![image](https://user-images.githubusercontent.com/8062446/42087893-3a149204-7bb5-11e8-8cca-309ce4742b96.png) which should be displayed like 
![image](https://user-images.githubusercontent.com/8062446/42087926-5356cb6a-7bb5-11e8-9242-21b39411b6cc.png). 
Also there is a stackoverflow question that raise the similar issue. [https://stackoverflow.com/questions/44254171/devanagari-text-rendering-improperly-in-pygame](url)
Moreover there is a bugzilla post that discuss similar [issue.](https://bugzilla.libsdl.org/show_bug.cgi?id=3046). 
Can someone fix this bug ??

rohitsaluja22:
Hi, Every font file comes with a gsub file which is a set of rules that define substituting consegutive glyphs with a joint glyph for example. Pygame ignores such gsub files while rendering font. Can anyone tell how to force pygame to use gsub rules?

When gsub rules are not used, in Devanagari "म्ह" looks like "म् ह"(ignore space) as shown in the image:-

![pygame2](https://user-images.githubusercontent.com/16031087/42432117-5ed7f37c-838c-11e8-8612-e3273f03cb7e.png)

IngleJaya95:
PyGame is a waste of time if you are looking for the solution in the context of Indic Language. If you are looking for solution purely based on text rendering then you can take a look at this https://github.com/ldo/qahirah

illume:
Ah ok. So it seems the patch on the libsdl bugzilla for SDL_ttf (posted in 2015) is for using HarfBuzz. I haven't looked at the patch, but it seems getting someone to work on it in libsdl would be one way to get support there.

bedapudi6788:
Has anyone been able to get this working?

peanutbutterandcrackers:
[...]

PR #3086, opened 2022-03-20
Windows prebuilts updates with freetype and SDL_ttf 2.0.18

### LINKED ISSUE #1113: Update freetype version to 2.9.1+
I noticed that pygame is about 6 versions (and at least 5 years) behind the latest version of freetype. It looks like there have been some potentially nice changes over the last five years that may improve font rendering for pygame - mainly the addition of ClearType hinting.

I suppose there is a chance it might just be an easy copy and paste job :)

**To Do**

- [ ] Test what freetype version reported on linux with `freetype.get_version()`.
- [x] Test what freetype version reported on mac with `freetype.get_version()`.
- [x] Test what freetype version reported on windows dev10 wheel with `freetype.get_version()`.
- [ ] Resolve issue with out of date freetype library used in windows wheels.

**Related Docs:** https://www.pygame.org/docs/ref/freetype.html#pygame.freetype.get_version

Starbuck5:
For #3023
Also fixes #1113 
Also makes progress on #464 and #2574

This PR updates our SDL_ttf version to 2.0.18 (from 2.0.15) on Windows. This is a little more complex than usual, since SDL_ttf now is statically links to freetype. We were relying on their freetype DLL for pygame.freetype. 

So I thought this was a good opportunity to go through and modernize the old "pygame-1.9.2-prebuilt" downloads. I removed a bunch of superfluous stuff, bringing the download size down a lot, and then I added in a modern freetype DLL and headers provided by https://github.com/ubawurinna/freetype-windows-binaries

SDL_ttf 2.0.18 is exciting because it has a multiline text rendering API, it has support for color emojis, and has harfbuzz support (harfbuzz is also statically linked into the SDL_ttf lib so we don't need to worry about that).

Emojis through pygame.font:
Capture

### DIFF
--- a/buildconfig/config_win.py
+++ b/buildconfig/config_win.py
@@ -389,8 +389,8 @@ def setup():
     DEPS.add('IMAGE', 'SDL2_image', ['SDL2_image-[1-9].*'], r'(lib){0,1}SDL2_image\.dll$',
              ['SDL', 'jpeg', 'png', 'tiff'], 0)
     DEPS.add('FONT', 'SDL2_ttf', ['SDL2_ttf-[2-9].*'], r'(lib){0,1}SDL2_ttf\.dll$', ['SDL', 'z', 'freetype'])
-    DEPS.add('FREETYPE', 'freetype', ['freetype-[1-9].*'], r'(lib){0,1}freetype[-0-9]*\.dll$',
-             find_header=r'ft2build\.h', find_lib=r'(lib)?freetype[-0-9]*\.lib')
+    DEPS.add('FREETYPE', 'freetype', ['freetype'], r'freetype[-0-9]*\.dll$',
+             find_header=r'ft2build\.h', find_lib=r'freetype[-0-9]*\.lib')
     DEPS.configure()
     _add_sdl2_dll_deps(DEPS)
     for d in get_definitions():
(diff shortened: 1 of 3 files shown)


buildconfig/config_win.py
```diff
@@ -421,15 +421,8 @@ def setup_prebuilt_sdl2(prebuilt_dir):
...
-    ftDep = DEPS.add('FREETYPE', 'freetype', ['SDL2_ttf-[2-9].*', 'freetype-[1-9].*'], r'(lib)?freetype[-0-9]*\.dll$',
-                     find_header=r'ft2build\.h', find_lib=r'libfreetype[-0-9]*\.lib')
-    ftDep.path = fontDep.path
-    ftDep.inc_dir = [
-        os.path.join(prebuilt_dir, 'include').replace('\\', '/')
-    ]
-    ftDep.inc_dir.append('%s/freetype2' % ftDep.inc_dir[0])
-    ftDep.found = True
+    DEPS.add('FREETYPE', 'freetype', ['freetype'], r'freetype[-0-9]*\.dll$',
+                     find_header=r'ft2build\.h', find_lib=r'freetype[-0-9]*\.lib')
```
ankith26, line 425: Why is this line needed here? I believe the same thing is done a few lines above too

ankith26 (approved):
Overall, really nice PR! 👍 💯 

I like how the prebuilts are lighter now, and only contain the things needed. The previous prebuilts had SDL1 which gave autocomplete issues while working on pygame, so this change should fix this.

As discussed before, for the middle term it's still worth getting a new dependency system for windows, but since that can take a while to get correct, this PR improves things in the mean time

Left a review, which is more of a question :)

dr0id (approved):
resolve open comment, approved anyway

buildconfig/config_win.py
Starbuck5: I'm not confident enough in the control flow of config_win to say definitely without further testing. The two methods appear to be run in different "modes," but I'm only familiar with the one mode for windows buildconfig. Regardless, I think it's safe to just change them both to point to the correct freetype.

Starbuck5:
> As discussed before, for the middle term it's still worth getting a new dependency system for windows, but since that can take a while to get correct, this PR improves things in the mean time

Never!
I mean, I don't see the need.

All of the SDL_ deps come nicely from their website prebuilt with all of their deps
The only other things we need are freetype and portmidi

We can get freetype programmatically from https://github.com/ubawurinna/freetype-windows-binaries.
And we could do the same with portmidi if someone 😉 went to the portmidi github and implemented windows prebuilt uploading for them.

Issue #2996, opened 2022-01-20
Vector subclassing


Starbuck5:
When you subclass a pygame Vector object, operations that return an instance of the class return a pygame Vector object, rather than the subclass. The Rect class behaves correctly in this.

See this REPL to understand what I mean:
```py
>>> class MyRect(pygame.Rect):
...     pass
... 
>>> a = MyRect(1,2,3,4)
>>> b = a.copy()
>>> type(b)
<class '__main__.MyRect'>
>>> class MyVector(pygame.Vector2):
...     pass
... 
>>> c = MyVector(5,10)
>>> d = c.copy()
>>> type(d)
<class 'pygame.math.Vector2'>
```

This was pointed out to me by Cozyfractal on discord.

PR #3088, opened 2022-03-21
Fix Vector subclass methods to return the correct subtype instance

novialriptide:
Fix: https://github.com/pygame/pygame/issues/2996

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -60,6 +60,10 @@ Multiple coordinates can be set using slices or swizzling
 
    Some general information about the ``Vector2`` class.
 
+   .. versionchanged:: 2.1.3 
+      Inherited methods of vector subclasses now correctly return an instance of the 
+      subclass instead of the superclass
+
    .. method:: dot
 
       | :sl:`calculates the dot- or scalar-product with the other vector`
@@ -434,6 +438,10 @@ Multiple coordinates can be set using slices or swizzling
    | :sg:`Vector3((x, y, z)) -> Vector3`
 
    Some general information about the Vector3 class.
+
+   .. versionchanged:: 2.1.3 
+      Inherited methods of vector subclasses now correctly return an instance of the 
+      subclass instead of the superclass
 
    .. method:: dot
 
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -4025,7 +4025,8 @@ vector_elementwise(pgVector *vec, PyObject *_null)
     if (proxy == NULL)
         return NULL;
     Py_INCREF(vec);
-    proxy->vec = (pgVector *)vec;
+    proxy->vec = (pgVector *)_vector_subtype_new(Py_TYPE(vec), &pgVectorElementwiseProxy_Type);
+    // proxy->vec = (pgVector *)vec;
     return (PyObject *)proxy;
 }
 
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -135,10 +135,8 @@ _vector_move_towards_helper(Py_ssize_t dim, double *origin_coords,
 /* generic vector functions */
 static PyObject *
 _vector_subtype_new(PyTypeObject *type, Py_ssize_t dim);
-/* 
 static PyObject *
 pgVector_NEW(Py_ssize_t dim);
-*/
 static void
 vector_dealloc(pgVector *self);
 static PyObject *
@@ -574,21 +572,40 @@ _vector_subtype_new(PyTypeObject *type, Py_ssize_t dim)
     return (PyObject *)vec;
 }
 
-/*
 static PyObject *
 pgVector_NEW(Py_ssize_t dim)
 {
+    pgVector *vec;
     switch (dim) {
         case 2:
-            return _vector_subtype_new(&pgVector2_Type, dim);
+            vec = PyObject_New(pgVector, &pgVector2_Type);
+            break;
         case 3:
-            return _vector_subtype_new(&pgVector3_Type, dim);
+            vec = PyObject_New(pgVector, &pgVector3_Type);
+            break;
+            /*
+                case 4:
+                    vec = PyObject_New(pgVector, &pgVector4_Type);
+                    break;
+            */
         default:
             PyErr_SetString(PyExc_SystemError,
                             "Wrong internal call to pgVector_NEW.\n");
             return NULL;
     }
-}*/
+
+    if (vec != NULL) {
+        vec->dim = dim;
+        vec->epsilon = VECTOR_EPSILON;
+        vec->coords = PyMem_New(double, dim);
+        if (vec->coords == NULL) {
+            Py_DECREF(vec);
+            return PyErr_NoMemory();
+        }
+    }
+
+    return (PyObject *)vec;
+}
 
 static void
 vector_dealloc(pgVector *self)
@@ -1805,7 +1822,7 @@ vector_getAttr_swizzle(pgVector *self, PyObject *attr_name)
     }
 
     if (len == 2 || len == 3) {
-        res = (PyObject *)_vector_subtype_new(Py_TYPE(self), (int)len);
+        res = (PyObject *)pgVector_NEW((int)len);
     }
     else {
         /* More than 3, we return a tuple. */
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -134,6 +134,8 @@ _vector_move_towards_helper(Py_ssize_t dim, double *origin_coords,
 
 /* generic vector functions */
 static PyObject *
+_vector_subtype_new(PyTypeObject *type, Py_ssize_t dim);
+static PyObject *
 pgVector_NEW(Py_ssize_t dim);
 static void
 vector_dealloc(pgVector *self);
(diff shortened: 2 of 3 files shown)


src_c/math.c
```diff
@@ -547,40 +549,47 @@ static PyMemberDef vector_members[] = {
 };
 
 static PyObject *
-pgVector_NEW(Py_ssize_t dim)
+_vector_subtype_new(PyTypeObject *type, Py_ssize_t dim)
 {
     pgVector *vec;
+    Py_ssize_t i;
```
ankith26, line 555: what is this for here? Can be removed

src_c/math.c
```diff
@@ -547,40 +549,47 @@ static PyMemberDef vector_members[] = {
...
 
-    if (vec != NULL) {
+    if (vec) {
         vec->dim = dim;
-        vec->epsilon = VECTOR_EPSILON;
         vec->coords = PyMem_New(double, dim);
-        if (vec->coords == NULL) {
-            Py_DECREF(vec);
-            return PyErr_NoMemory();
-        }
```
ankith26, line 578: why was this check removed? It is very much needed, so it needs to be added back

src_c/math.c
```diff
@@ -756,7 +765,10 @@ vector_inplace_floor_div(pgVector *o1, PyObject *o2)
 static PyObject *
 vector_neg(pgVector *self)
 {
-    pgVector *ret = (pgVector *)pgVector_NEW(self->dim);
+    pgVector *ret;
+
+    ret = (pgVector *)_vector_subtype_new(Py_TYPE(self), self->dim);
```
ankith26, line 770: minor code appearance nitpick, this bit of code will look neater if it's expressed in one line, like 
` pgVector *ret  = (pgVector *)_vector_subtype_new(Py_TYPE(self), self->dim);`
Here, and also in all functions below

src_c/math.c
```diff
@@ -792,8 +807,11 @@ vector_nonzero(pgVector *self)
 static PyObject *
 vector_copy(pgVector *self)
 {
-    pgVector *ret = (pgVector *)pgVector_NEW(self->dim);
+    pgVector *ret;
     Py_ssize_t i;
+
+    ret = (pgVector *)_vector_subtype_new(Py_TYPE(self), self->dim);
+
```
ankith26, line 814:
```suggestion
    if (!ret) {
        return NULL;
    }
```
Not really a bug introduced by this PR but can be fixed here, this is missing error forwarding return

src_c/math.c
```diff
@@ -3426,33 +3447,33 @@ static PyMethodDef vectoriter_methods[] = {
...
+    0,                                        /* tp_as_number */
+    0,                                        /* tp_as_sequence */
+    0,                                        /* tp_as_mapping */
+    0,                                        /* tp_hash */
+    0,                                        /* tp_call */
+    0,                                        /* tp_str */
+    PyObject_GenericGetAttr,                  /* tp_getattro */
+    0,                                        /* tp_setattro */
+    0,                                        /* tp_as_buffer */
+    Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE, /* tp_flags */
```
ankith26, line 3467: You put this flag for the `pgVectorIter_Type` which is not needed IMHO. I don't think users want to subclass something that is not even documented pygame API, probably out of scope for this PR

ankith26 (changes requested):
Needs some unit tests too

Starbuck5:
This is starting to look really clean! As Ankith mentioned, it needs tests. It could also probably use a note in the docs.

docs/reST/ref/math.rst
```diff
@@ -60,6 +60,10 @@ Multiple coordinates can be set using slices or swizzling
 
    Some general information about the ``Vector2`` class.
 
+   .. versionadded:: 2.1.3
+      The Vector2 class can be subclassed. Methods such as ``copy()`` and ``move_towards()``
+      will recognize this and return instances of the subclass.
```
ankith26, line 65: vectors were always subclassable. What this PR changes is the return types of inherited methods to match the subclass

src_c/math.c
```diff
@@ -547,40 +549,40 @@ static PyMemberDef vector_members[] = {
...
-        default:
-            PyErr_SetString(PyExc_SystemError,
-                            "Wrong internal call to pgVector_NEW.\n");
-            return NULL;
-    }
 
-    if (vec != NULL) {
+    vec = (pgVector *)(type->tp_new(type, NULL, NULL));
+
+    if (vec) {
```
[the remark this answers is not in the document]
ankith26: This makes sense, pretty sure `tp_new` sets a python error and returns `NULL`, afterall it's just another python function.

ankith26 (approved):
LGTM! Thanks for the PR 👍 🎉

Starbuck5:
You got the explicit methods, but it still fails on math.

```
>>> class VEC(pygame.Vector2):
...     pass
...
>>> a = VEC(5,6)
>>> b = VEC(7,8)
>>> a + b
<Vector2(12, 14)>
>>> type(a + b)
<class 'pygame.math.Vector2'>
```

It passes `-a` though. You should also add a test for that.

src_c/math.c
```diff
@@ -756,7 +758,8 @@ vector_inplace_floor_div(pgVector *o1, PyObject *o2)
 static PyObject *
 vector_neg(pgVector *self)
 {
-    pgVector *ret = (pgVector *)pgVector_NEW(self->dim);
+    pgVector *ret = (pgVector *)_vector_subtype_new(Py_TYPE(self), self->dim);
+
     if (ret != NULL) {
         Py_ssize_t i;
```
Starbuck5, line 764: This isn't your fault, but I noticed almost all the rest check if the value is NULL and return NULL there. You could change this method to be consistent with that.

MyreMylar (changes requested):
This resolves the original issue mentioned however I noticed some inconsistency in whether we now return a VectorN or whether we return a subtype.

For example:

```
c = MyVector(5,10)
d = c.copy()
type(d)
<class '__main__.MyVector'>
e = c + d
type(e)
<class 'pygame.math.Vector2'>
```

Why does adding two MyVector objects produce a Vector2? Is this what we want to happen or should it also return an instance of the subtype? There are still a handful of places (including these basic math ops) where we are returning an instance of the parent class rather than the subclass and I think we should definitely have a solid think about each one.

Apologies if this has already been done somewhere, but I couldn't see it in the PR comment thread.

novialriptide:
> Apologies if this has already been done somewhere, but I couldn't see it in the PR comment thread.

Yup, Starbuck5 has mentioned this issue. I'm currently working on a solution to that. 

> is this what we want to happen or should it also return an instance of the subtype?

Nope, this isn't what is supposed to happen.

ankith26 (commented):
In addition to the generic math function not being fixed that is already mentioned before, I noticed that vector swizzling also returns a pygame vector instance... which should probably be fixed too...
```
>>> class MyVec(pygame.Vector2): pass
... 
>>> MyVec(2, 3).xx 
<Vector2(2, 2)>
```
And there are some elementwiseproxy functions too that seem to deal with vectors directly

novialriptide:
There are a few more that needs to support subtype, but I'm pushing my latest changes to show some progress.

novialriptide:
There's a little bit of an oopsie with the subclassing PR, so for swizzling, let's say you have this piece of code
```python
class MyVector2(pygame.Vector2):
  pass

print(MyVector2(3, 5).xy)
```
This piece of code will work just fine, assuming it'll return a `MyVector2` type instead of a `pygame.Vector2`, but what if we add this at the end?
```python
print(MyVector2(3, 5).xyx
```
This will return a `pygame.Vector3` type instead of a `MyVector2` since it has 3 dimensions instead of 2.

Any ideas on this problem? Should we disable swizzling if the Vector class has been subclassed, only return `pygame.VectorN` for sizzling, or keep it how I described it, return `pygame.VectorN` if it can't return itself.

ankith26 (commented):
Since I see this function call pattern `(pgVector *)_vector_subtype_new(Py_TYPE(var), var->dim);` is almost universal in this file, I'd suggest implementing the function signature like 
`pgVector *_vector_subtype_new(pgVector *base);` 
so that the call simplifies to `_vector_subtype_new(var);`

src_c/math.c
```diff
@@ -550,6 +552,28 @@ static PyMemberDef vector_members[] = {
     {NULL} /* Sentinel */
 };
 
+static PyObject *
```
ankith26, line 555: Return type here could be `pgVector *` because almost all function calls to this cast this to `pgVector *`

test/math_test.py
```diff
@@ -1157,6 +1157,27 @@ def test_clamp_mag_v2_min(self):
...
+        self.assertNotEqual(type(v), type(other))
+        self.assertNotEqual(type(v), type(other.copy()))
+        self.assertEqual(type(other), type(other.reflect(v)))
+        self.assertEqual(type(other), type(other.lerp(v, 1)))
+        self.assertEqual(type(other), type(other.slerp(v, 1)))
+        self.assertEqual(type(other), type(other.rotate(5)))
+        self.assertEqual(type(other), type(other.rotate_rad(5)))
+        self.assertEqual(type(other), type(other.project(v)))
+        self.assertEqual(type(other), type(other.move_towards(v, 5)))
+
```
ankith26, line 1180: extend this with tests for subclass math operations and elementwise tests too

test/math_test.py
```diff
@@ -2454,6 +2475,27 @@ def test_clamp_mag_v3_min(self):
...
+        self.assertNotEqual(type(v), type(other))
+        self.assertNotEqual(type(v), type(other.copy()))
+        self.assertEqual(type(other), type(other.reflect(v)))
+        self.assertEqual(type(other), type(other.lerp(v, 1)))
+        self.assertEqual(type(other), type(other.slerp(v, 1)))
+        self.assertEqual(type(other), type(other.rotate(5, v)))
+        self.assertEqual(type(other), type(other.rotate_rad(5, v)))
+        self.assertEqual(type(other), type(other.project(v)))
+        self.assertEqual(type(other), type(other.move_towards(v, 5)))
+
```
ankith26, line 2498: same as above review

src_c/math.c
```diff
@@ -3858,6 +3891,7 @@ vector_elementwiseproxy_pow(PyObject *baseObj, PyObject *expoObj,
     if (vector_elementwiseproxy_Check(baseObj)) {
         dim = ((vector_elementwiseproxy *)baseObj)->vec->dim;
         tmp = ((vector_elementwiseproxy *)baseObj)->vec->coords;
+        vec = ((vector_elementwiseproxy *)baseObj)->vec;
```
ankith26, line 3894:
```suggestion
        vec = ((vector_elementwiseproxy *)baseObj)->vec;
        dim = vec->dim;
        tmp = vec->coords;
```
For better code clarity

src_c/math.c
```diff
@@ -3885,6 +3919,7 @@ vector_elementwiseproxy_pow(PyObject *baseObj, PyObject *expoObj,
     else {
         dim = ((vector_elementwiseproxy *)expoObj)->vec->dim;
         tmp = ((vector_elementwiseproxy *)expoObj)->vec->coords;
+        vec = ((vector_elementwiseproxy *)expoObj)->vec;
```
ankith26, line 3922:
```suggestion
        vec = ((vector_elementwiseproxy *)expoObj)->vec;
        dim = vec->dim;
        tmp = vec->coords;
```

ankith26 (changes requested):
Also needs typehint changes in addition to some extra reviews I left

Starbuck5 (approved):
Thanks for sticking through this Novial! This is cool. 🎉

The type test might want a test that the swizzle type is unrelated to the subtype, since we've decided to do it that way.

Ankith says something about needing type hints, I'm not sure how that would work at all. Typing makes my brain hurt.

src_c/math.c
```diff
@@ -550,6 +552,28 @@ static PyMemberDef vector_members[] = {
...
+    vec = (pgVector *)(type->tp_new(type, NULL, NULL));
+
+    if (vec) {
+        vec->dim = dim;
+        vec->epsilon = VECTOR_EPSILON;
+        vec->coords = PyMem_New(double, dim);
+
+        if (vec->coords == NULL) {
+            Py_DECREF(vec);
+            return PyErr_NoMemory();
```
ankith26, line 571:
```suggestion
            return (pgVector *)PyErr_NoMemory();
```

ankith26 (approved):
LGTM! Thanks for the PR 👍 🎉 

(I mentioned that typestubs need fixes after this PR, but as discussed on discord I will be opening a new PR for that)

ankith26:
For those OOTL, we discussed swizzling return types on discord, and the current decision is to keep old behaviour for swizzling, returning `VectorN` instances even when vector subclasses are swizzled.

This PR is bit of an API change, but it's a step in the right direction. User code that subclasses vectors should not break if those subclasses were designed with Liskov substitution principle in mind

PR #3092, opened 2022-03-23
Adjusted examples and rewrote documentation accordingly

dbrown3562:
Made a few changes in regards to the following examples:

arraydemo - no longer scrolls via any mouse input, requires the main mouse button be clicked. Also changed it such that a click only counts when the mouse button is unclicked.
camera - changed the format for the camera list so it might be more readable to the user (although still console-based). Added the name of the camera to the display of the window so it says <camera name> (fps) instead of simply CAMERA! (fps).
font_viewer - made the font color of the directions yellow so they stand out better.
scroll - improved the window scrolling so that it can accept multiple keyboard inputs at once.

moveit - there was an error where the delay() function wasn't being called, causing it to move unreasonably fast. However, I thought it better to use clock.tick() to handle this. Additionally, I expanded the tutorial to allow the user to control a character on screen, rather than simply watching objects move. This included changing the object class to allow for multidirectional movement, and I essentially used what I did for scroll to allow the user control over a unique object.
moveit.rst - Documentation updated accordingly. The tutorial has been expanded to include a few "advanced" sections where I describe in thorough detail the changes I made to the base code and the additions to user input.


examples/scroll.py
```diff
@@ -151,29 +150,62 @@ def main(image_file=None):
...
                     going = False
+
+                # handle mouse button presses on arrows.
                 elif e.type == pg.MOUSEBUTTONDOWN:
                     direction = regions.get_at(e.pos)[0]
                 elif e.type == pg.MOUSEBUTTONUP:
                     direction = None
+            
+            # given a scroll event, preform that action. This enables concurrent presses and multidirectional scrolling.
+            if(scroll_up):
```
novialriptide, line 201: For these if statements, it would be better to get rid of the parathesis to stay consistent with the other if statements in the code.

examples/camera.py
```diff
@@ -87,11 +103,11 @@ def main(self):
                     going = False
                 if e.type == pg.KEYDOWN:
                     if e.key in range(pg.K_0, pg.K_0 + 10):
-                        self.init_cams(e.key - pg.K_0)
+                        camera = self.init_cams(e.key - pg.K_0)
 
             self.get_and_flip()
             self.clock.tick()
-            pygame.display.set_caption(f"CAMERA! ({self.clock.get_fps():.2f} FPS)")
+            pygame.display.set_caption(camera + f" ({self.clock.get_fps():.2f} FPS)")
```
novialriptide, line 110: You could do `f"{camera} ({self.clock.get_fps():.2f} FPS)"` instead.

novialriptide (changes requested):

Starbuck5:
Thanks for the changes so far @dbrown3562!

You might want to split off the moveit changes into a separate pull request. It's much easier to review the small things than that one. And so we could probably merge them quicker separately.

`arraydemo` - "main button" is unclear to me, I think "left click" is more authoritative
`camera` - simple change makes it look a lot better already
`font_viewer`- Simple change, looks nicer! Since `dparams` is never a thing, you could remove that and simplify the code. What do you think about the wording of "Foreign fonts render incorrectly"? It seems like it shouldn't use the word foreign, since people from countries with scripts that might render improperly, like Hindi and Arabic, can look at the example too.
`scroll` - the functionality is better now, but there is a simpler way to do the code. Rather than using keyups and keydowns, you could use pygame.key.get_pressed(), which is meant for continuous input like this. Also perform is misspelled.

`scroll` and `font_viewer` need titles with display.set_caption

`camera` and `arraydemo` could be revamped to fully use the GUI, but let's say that's out of scope for this PR.

We use the tool `black` for code formatting. You can `pip install black`, then `black examples` (from the root of the pygame folder) to make sure all the code is up to the style standard.

For `moveit`, the tutorial now has a few issues. The docs generate incorrectly, something is wrong with your code blocks. Btw, you can generate the docs by running `setup.py docs` and then visit them in the `docs/generated` folder. It's also strange to move from a conceptual basis to treating it as a file, since all the code blocks have each line with the `>>>` of a REPL. Also snippet is spelled incorrectly.

examples/arraydemo.py
```diff
@@ -31,8 +30,10 @@ def surfdemo_show(array_img, name):
     pg.display.set_caption(name)
     while True:
         e = pg.event.wait()
-        if e.type == pg.MOUSEBUTTONDOWN:
-            break
+        #Force application to only advance when main button is released
+        if e.type == pg.MOUSEBUTTONUP:
+            if(e.button == 1):
```
MyreMylar, line 35: probably better to use the pygame constant here:
```
if e.type == pg.MOUSEBUTTONUP and e.button == pg.BUTTON_LEFT:
    break;
```

examples/arraydemo.py
```diff
@@ -57,7 +58,7 @@ def main():
 
     pg.init()
     print("Using %s" % surfarray.get_arraytype().capitalize())
-    print("Press the mouse button to advance image.")
+    print("Press the main mouse button to advance image.")
```
MyreMylar, line 61: Agreed that his should be `"Press the left mouse button to advance image."`

examples/scroll.py
```diff
@@ -151,29 +150,62 @@ def main(image_file=None):
         clock.tick()
 
         going = True
+
+        # scroll events
+        scroll_down = 0
+        scroll_up = 0
+        scroll_left = 0
+        scroll_right = 0
```
MyreMylar, line 158: Why are these integers instead of booleans?

MyreMylar (changes requested):
Added a few comments from a first look through 👍

dbrown3562:
Hey all! Just want to let everyone know I'm still here, been getting hit hard by school. I will have these changes fixed up very soon, and the pull request split into two (for the moveit and other examples). Thanks all.

dbrown3562:
Hey all! Just checking up to see if you'd like any more changes.

illume:
Hello. Sorry no one got back to you. It seems now some other people have made some changes causing your changes to conflict. 

I put a 'needs-review' label on here. Hopefully the previous reviewers can check if they are happy now you've done some more changes.

If you don't have time/or interest to fix the conflicts anymore, let us know and someone else can probably fix them for you :)

examples/arraydemo.py
```diff
@@ -31,7 +30,8 @@ def surfdemo_show(array_img, name):
     pg.display.set_caption(name)
     while True:
         e = pg.event.wait()
-        if e.type == pg.MOUSEBUTTONDOWN:
+        # Force application to only advance when main button is released
```
illume, line 33: 👍

examples/moveit.py
```diff
@@ -65,6 +65,7 @@ def main():
             screen.blit(o.image, o.pos)
 
         pg.display.update()
```
illume, line 67: ps. Since update() is called with no arguments, it is equivalent to flip. So probably this should have display.flip here.

(I'm not saying anyone should necessarily make this fix, just something I noticed.)

examples/moveit.py
```diff
@@ -65,6 +65,7 @@ def main():
...
+        pg.time.delay(100)
```
illume, line 68: I guess probably a clock would be better than the delay. However, time.delay here is ok too.

examples/scroll.py
```diff
@@ -104,6 +104,7 @@ def main(image_file=None):
     background_color = pg.Color("beige")
 
     pg.init()
+    pg.display.set_caption("Scroll Example")
```
illume, line 107: 👍

examples/scroll.py
```diff
@@ -151,25 +152,29 @@ def main(image_file=None):
         clock.tick()
 
         going = True
+
         while going:
             # wait for events before doing anything.
             # events = [pg.event.wait()] + pg.event.get()
             events = pg.event.get()
 
+            # During the loop, if a key is held, scroll the view.
```
illume, line 161: I often mostly tell people to avoid using key.get_pressed and just use the event loop. But I wonder what other people think? I'm also not sure about just using if, and not elif.

MyreMylar (approved):

examples/scroll.py
```diff
@@ -151,25 +152,29 @@ def main(image_file=None):
...
+            keys = pg.key.get_pressed()
+            if keys[pg.K_UP]:
+                scroll_view(screen, image, DIR_UP, view_rect)
+            if keys[pg.K_DOWN]:
+                scroll_view(screen, image, DIR_DOWN, view_rect)
+            if keys[pg.K_LEFT]:
+                scroll_view(screen, image, DIR_LEFT, view_rect)
+            if keys[pg.K_RIGHT]:
+                scroll_view(screen, image, DIR_RIGHT, view_rect)
```
ankith26, line 170: This change in code here, means that this example does not need to set keyrepeat anymore (because those are not handled)

ankith26 (approved):
Thanks for contributing to pygame! 🥳 🎉 

I left a minor review, but I don't think that should be blocking a merge, considering the example still plays well

(noticed this PR has a bunch of reviews that are nice-to-address but not critical, but I'm inclined to merging this one in, these can be handled later)

examples/scroll.py
MyreMylar: I think we should merge this and then perhaps make a few mental notes (or issues) to fiddle with later. Original PR submitter has waited long enough, and we could probably fiddle with the examples forever (and will).

MyreMylar:
Merging for now as original submission is several months old and has likely ran out of steam for minor tweaks that could be sorted later.

PR #3094, opened 2022-03-27
Small fixes to the documentation header

hnrkcode:
- Replaced the html table with css flex box to make the header in the documentation stretch out all the way on the right side.
- Also fixed the black borders in the header to be drawn the same on all screen sizes.

### DIFF
--- a/docs/reST/themes/classic/elements.html
+++ b/docs/reST/themes/classic/elements.html
@@ -8,16 +8,15 @@
  #}
 {%- macro header() %}
   <div class="header">
-    <table>
-      <tr>
-	<td class="logo">
+	<div class="flex-container">
+	<div class="logo">
 	  <a href="{{ theme_home_uri }}">
 {%-         set logo_path = '_static/' + logo %}
 	    <img src="{{ pathto(logo_path, 1) }}"/>
 	  </a>
 	  <h5>pygame documentation</h5>
-	</td>
-	<td class="pagelinks">
+	</div>
+	<div class="pagelinks">
 	  <div class="top">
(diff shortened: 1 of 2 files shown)


ankith26:
Hi, thanks for the PR! 🥳 

I cannot review your changes in detail since I'm not fluent with CSS, but I did test your PR locally and I'll post some screenshots here so that others can see the improvements made by your changes.

### On a small screen
image

The right side of the box is drawn correctly now, and also there's a line separating the different sections of the header, which is nice.

### On a zoomed-out browser window
image
image

So yeah, it looks neater now that the header fills the entire space now. Minor nitpick, could the header be centered somehow?

Another thing, which could potentially be out of scope for this PR; but could the docs text fill the entire width of the screen too, just like the header? I believe that the docs text width is currently hardcoded, but it could css flex to fill?

Grimmys:
About the content width, I was into the process to change it to something like 90/100% as you suggested Ankith but some people (I think @Starbuck5 ) explained me that the limited width is intended for better readibility (less horizontal movement for eyes). 
But maybe at least it could be something adaptative like 50/60% instead of a fixed value. 

Good PR beside that, it would be great to use more modern CSS techniques. Flex > table. 
I take a look at the CSS code and it looks correct.

MyreMylar:
Yeah I think the upgrade to a more modern web standard that can cater to the enormous variety of screens we have now seems like a good change. Setting the header to 60% or something width, if UX designers agree that is the way could be easily added later.

MyreMylar (approved):
LGTM 👍

ankith26 (approved):
Thanks for the PR! 👍 🎉

Issue #3087, opened 2022-03-20
Unify version getters


Starbuck5:
Pygame depends on several C libraries, and occasionally people might want to know the versions of said libraries, so pygame provides version getter functions.

SDL: `pygame.get_sdl_version()`
SDL_ttf: missing
Freetype: `pygame.freetype.get_version()`
SDL_image: `pygame.image.get_sdl_image_version()`
SDL_mixer: `pygame.mixer.get_sdl_mixer_version()`

These all have slightly different semantics, it would be a cool cleanup effort to get them going the same way, and to add an SDL_ttf version function.

pygame.get_sdl_version returns what SDL was compiled against
pygame.freetype.get_version returns what freetype was compiled against
pygame.mixer.get_sdl_mixer_version returns the runtime version by default, but can also return the compiled version with a parameter
pygame.image.get_sdl_image_version  returns what SDL_image was compiled against, or None

My feeling is that they should all work like `get_sdl_mixer_version`. They should all default to the runtime check, with the compile time check as a backup. This is a behavior change but I don't think this would introduce backwards compatibility problems.

Tasks:
- [ ] convert existing version functions to work like `get_sdl_mixer_version`
- [x] add a new version function for SDL_ttf

Starbuck5:
I just noticed this code in `get_sdl_mixer_version` that may need to be fixed up

```c
    if (NULL != linkedobj) {
        linked = PyObject_IsTrue(linkedobj);

        if (-1 == linked) {
            return RAISE(PyExc_TypeError, "linked argument must be a boolean");
        }
    }
```

It looks like the author intended this to be a true/false check, with a fail if not a boolean. But it's really a truthiness check. You could pass in the string "pygame" and the function would interpret that as `linked = True`.

ankith26:
+1 for the general idea, I would love to see more consistency here, and keyword argument for users to get access to both compiled and linked versions sounds useful for reporting bugs related to the respective submodules.

However I believe that the compiled version is often the more useful number for users experimenting with things, so that should be returned by default, this is also inline with most of the existing functions behaviour.

About `get_sdl_mixer_version`, we can either
1. keep things as is, return linked version by default
2. break compatibility a bit, return compiled version by default, make `linked=False` the default

I don't have a particular preference in the above 2 options, anything should be fine really

Starbuck5:
> However I believe that the compiled version is often the more useful number for users experimenting with things, so that should be returned by default, this is also inline with most of the existing functions behaviour.

[...]

PR #3095, opened 2022-03-27
pygame.font.get_ttf_version() request from issue 3087

mariospapaz:
issue link: https://github.com/pygame/pygame/issues/3087

Now the user can get the SDL_ttf version in both linked and compiled ways.

Documentation:
Chapter 3.1.1
https://www.libsdl.org/projects/docs/SDL_ttf/SDL_ttf.html#SEC7

Screen Shot 2022-03-27 at 9 46 50 PM


novialriptide:
Looks good, all you need to do now is add this to the docs.

src_c/font.c
```diff
@@ -788,14 +789,53 @@ get_default_font(PyObject *self, PyObject *_null)
...
+}
+
+
 static PyMethodDef _font_methods[] = {
     {"init", (PyCFunction)fontmodule_init, METH_NOARGS, DOC_PYGAMEFONTINIT},
     {"quit", (PyCFunction)fontmodule_quit, METH_NOARGS, DOC_PYGAMEFONTQUIT},
     {"get_init", (PyCFunction)get_init, METH_NOARGS, DOC_PYGAMEFONTGETINIT},
     {"get_default_font", (PyCFunction)get_default_font, METH_NOARGS,
      DOC_PYGAMEFONTGETDEFAULTFONT},
+    {"get_ttf_version", (PyCFunction) get_ttf_version, METH_VARARGS | METH_KEYWORDS, DOC_PYGAMEFONTGETINIT},
```
novialriptide, line 834: It would be better to rename this function to `get_version()`, since `freetype` uses `get_version()` as well.

novialriptide (changes requested):

src_c/font.c
Starbuck5: Freetype is the odd one out, all the rest use this format.

test/font_test.py
```diff
@@ -35,6 +35,9 @@ def setUp(self):
 
     def tearDown(self):
         pygame_font.quit()
+    
+    def get_ttf_version(self):
+        pygame_font.get_ttf_version()
```
Starbuck5, line 40: I know you're not done with this yet, but the reason this doesn't run is that test methods must start with `test_`

mariospapaz: Issue resolved.

Which are the other steps to get it working with the docs ?

mariospapaz:
if there are no further requests regarding this function, the feature is finally finished.

test/font_test.py
```diff
@@ -36,6 +36,9 @@ def setUp(self):
     def tearDown(self):
         pygame_font.quit()
 
+    def test_get_sdl_ttf_version(self):
+        pygame_font.get_sdl_ttf_version()
```
novialriptide, line 40: Must add another test for unlinked as well

novialriptide, line 40: Might as well add a test if this function actually returns a string or tuple with 3 ints.

novialriptide (changes requested):
Good job on the docs. Just a few more changes.

MyreMylar (changes requested):
Given this a good local test and code wise it all seems to work 👍 

This is how the docs look right now:

image


I'm thinking we should add an extra sentence explaining what SDL_ttf is to the average pygame user, and a full stop after the first sentence.

Something like:

> Returns a tuple of integers that identify SDL_ttf's version. SDL_ttf is the underlying font rendering library, written in C, on which pygame's font module depends.

```

**Comment by MyreMylar** [comment]
Also, this test will always fail for ftfont because that doesn't use SDL_ttf.

So perhaps the test should be:

```
def test_get_sdl_ttf_version(self):
    if pygame_font.__name__ != "pygame.ftfont":
        pygame_font.get_sdl_ttf_version()
```

**Comment by mariospapaz** [comment]
Dear @MyreMylar, I've added the things you requested

An overview of the first request:
Screen Shot 2022-04-03 at 12 33 58 PM

**Review by MyreMylar — APPROVED** [review]

File: buildconfig/stubs/pygame/font.pyi [hunk 4]
```diff
@@ -11,6 +11,7 @@ UCS4: Literal[1]
 def init() -> None: ...
 def quit() -> None: ...
 def get_init() -> bool: ...
+def get_sdl_ttf_version() -> Tuple[int, int, int]: ...
```
ankith26, line 14:
```suggestion
def get_sdl_ttf_version(linked: bool = True) -> Tuple[int, int, int]: ...
```

docs/reST/ref/font.rst
```diff
@@ -79,6 +79,19 @@ solves no longer exists, it will likely be removed in the future.
 
    .. ## pygame.font.get_default_font ##
 
+.. function:: get_sdl_ttf_version
+
+   | :sl:`gets SDL_ttf version`
+   | :sg:`get_sdl_ttf_version() -> (major, minor, patch)`
```
ankith26, line 85:
```suggestion
   | :sg:`get_sdl_ttf_version(linked=True) -> (major, minor, patch)`
```

docs/reST/ref/font.rst
```diff
@@ -79,6 +79,19 @@ solves no longer exists, it will likely be removed in the future.
...
+
+   Returns a tuple of integers that identify SDL_ttf's version.
+   SDL_ttf is the underlying font rendering library, written in C,
+   on which pygame's font module depends.
+
```
ankith26, line 90: Needs a mention of the `linked` argument here

ankith26 (commented):
Left some additional reviews in addition to the tests review I already gave on discord

MyreMylar (approved):
Looking good 👍

MyreMylar:
lint test is failing on the `font_test.py`. Just needs a quick:

```
black test/display_test.py
```

to pass the test. Think it is just spacing around the new function in the test.

Everything looks good though

MyreMylar (commented):
Couple of fixes for missing docs regeneration.

Apparently I can apply them myself, so I did.

ankith26 (approved):
LGTM! Thanks for the PR 🎉

illume:
get_ttf_version was a bit confusingly named to me. The function is for getting SDL_ttf version, so probably a better name would be get_sdl_ttf_version()? We have pygame.mixer.get_sdl_mixer_version() and pygame.get_sdl_version().

Starbuck5:
> get_ttf_version was a bit confusingly named to me. The function is for getting SDL_ttf version, so probably a better name would be get_sdl_ttf_version()? We have pygame.mixer.get_sdl_mixer_version() and pygame.get_sdl_version().

get_sdl_ttf_version is what was actually implemented, the PR was misnamed. I've fixed that.

illume:
Ah, cool. Thanks for updating the title.

Issue #718, opened 2019-01-06
emscripten pyodide pygame for web browser


illume:
pyodide https://github.com/iodide-project/pyodide is a emscripten compiled python scientific stack.

Could be interesting to port pygame to this.

dugo3number2:
This would be very helpful! I don't use Pygame very much, but when I think of anything related to WASM / programs usually on PC in the web browser, I think of Pygame. All 4 of the other engines/libraries I have used already have it either through emscripten or their own compilation (Raylib, SDL2, libGDX, Unity).

Olney1:
This would be extremely beneficial for Play to Earn cryptocurrency game development.

robertpfeiffer:
https://github.com/ethanhs/python-wasm is especially promising, as it is just a bunch of shell scripts that compile regular CPython without any patches!

Zireael07:
@robertpfeiffer: In my experience, combining any of those "compile X to WASM+Emscripten" projects are basically unusable for combining with any project (they're just REPLs and can't really be expanded upon to provide graphics etc.)

robertpfeiffer:
> In my experience, combining any of those "compile X to WASM+Emscripten" projects are basically unusable for combining with any project (they're just REPLs and can't really be expanded upon to provide graphics etc. - I tried to add some graphics on top of such a Lua project)

AFAIK getting CPython to compile with emscripten was the hard part all along. SDL2 already supports emscripten, so we have graphics, we just need to put it together.

pmp-p:
> they're just REPLs and can't really be expanded upon to provide graphics etc. 

no there are projects already very advanced.
like that one  - still experimental regarding Panda3D quality standards - like https://rdb.name/panda3d-webgl/editor.html  ( btw Panda3D is built automatically a as part of pydk mentionned in first post )


>  SDL2 already supports emscripten, so we have graphics, we just need to put it together.

indeed, but it won't support threading/socket/remote IO unless all your python code is fully async and handle all those if you stay on browser thread.

wasm WILL get better with cpython 3.11. as it builds fine and can run in a web worker But there are still things to tinker to get all SDL2 running from **inside** the worker 
 ( related https://github.com/ethanhs/python-wasm/issues/32 )

i managed to init SDL2 along with pymain 3.11 yesterday in a worker so i'll try to build pygame soon. sadly pygame does not access SDL2 via ffi/ctypes, that complicates things **a lot** and makes everything huge.

sleepntsheep:
[...]

PR #3104, opened 2022-03-31
missing pointer cast fix

pmp-p:
my dumb checker missed one because of the newline at 80cols.
following https://github.com/pygame/pygame/pull/3080#issuecomment-1068441255 for https://github.com/pygame/pygame/issues/718

@dr0id  indeed, the real C parser is the way !

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -236,7 +236,7 @@ image_get_extended(PyObject *self, PyObject *_null)
 }
 
 static PyObject *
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
Thanks 👍

dr0id (approved):
yes, the real parser is the way. Maybe I'll try to implement the script in the next month, not sure though.
Thanks.

PR #3105, opened 2022-03-31
pygame.h is in _sdl2 parent directory

pmp-p:
this should be harmless but helps fixing static build for https://github.com/pygame/pygame/issues/718.

### DIFF
--- a/src_c/cython/pygame/_sdl2/controller.pyx
+++ b/src_c/cython/pygame/_sdl2/controller.pyx
@@ -1,6 +1,6 @@
 from . import error
 
-cdef extern from "pygame.h" nogil:
+cdef extern from "../pygame.h" nogil:
     int pgJoystick_Check(object joy)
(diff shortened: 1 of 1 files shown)


ankith26 (changes requested):
Since this is a cython change, a cython files regen is needed in this PR. After installing cython with pip, run `python3 setup.py cython`

pmp-p:
@ankith26 then it's probably better to postpone that one until cpython 3.11 + cython 3 is out 
as it only has sense for static build with webassembly.

MyreMylar (approved):
LGTM 👍 Just got to wait for 3.11

Starbuck5:
I'd still rather regenerate the files, or just controller.c, just to keep the cython code and the generated code on the same page. It's no problem to have a huge diff because of it.

illume (approved):
👍 thanks

PR #3106, opened 2022-03-31
fix name/definition conflict for pgSurface_New

pmp-p:
There's a name conflict between https://github.com/pygame/pygame/blob/a1725cf7375f50a1719c09c758096617b2f4dc07/src_c/include/_pygame.h#L295 and  https://github.com/pygame/pygame/blob/a1725cf7375f50a1719c09c758096617b2f4dc07/src_c/surface.c#L107   

i suggest `pgSurface_New2` for the latter instead  because `pgSurface_New` already mentionned in C-API docs with a single argument.

that conflict appears when turning off C-API and trying to build static for https://github.com/pygame/pygame/issues/718

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -104,7 +104,7 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
 
 /* statics */
 static PyObject *
-pgSurface_New(SDL_Surface *info, int owner);
+pgSurface_New2(SDL_Surface *info, int owner);
 static PyObject *
 surf_subtype_new(PyTypeObject *type, SDL_Surface *s, int owner);
 static PyObject *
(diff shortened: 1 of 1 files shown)


novialriptide:
> @novialriptide pgSurface_New2() is the historic _pygame.h helper https://github.com/pygame/pygame/blob/a1725cf7375f50a1719c09c758096617b2f4dc07/src_c/include/_pygame.h#L275
> 
> Using the same name just make the symbol reachable directly without pointer cast - and any code addition - when C-api is turned off.

Ah okay, my bad.

MyreMylar (approved):
This seems to compile fine and the surfaces tests and a relatively complex application all run. I sort of hate the name `pgSurface_New2` but that was in there already.

ankith26 (changes requested):
I think there are many more places where such kind of conflict can arise, it's a common theme in the pygame C API. Why is only this one being fixed? Does this fix help with WASM? Because I think this conflict shouldn't be a problem otherwise

ankith26 (approved):
Thanks for the PR 👍 
I think in the long term this might need some more improvements, because ATM it's easy to confuse this as I got confusing while reviewing, then it took me more digging around the code to figure out what was going on

PR #3108, opened 2022-03-31
static build support for WebAssembly

pmp-p:
It implements https://github.com/pygame/pygame/issues/718 and it runs better than i expected, enough for making some noise and finding bugs :)

could also implements https://github.com/pyodide/pyodide/issues/289 but working with upstream is a safer bet imho.

Those code changes only sort name conflicts in global namespace of libpygame.a
and add a config_emsdk to handle Emscripten platform : no impact on native builds and no change to C-API.
all potentially harmfull name changes ( like PyInit_* or module names )  are protected with the condition BUILD_STATIC

~~to be fully useable rely on https://github.com/pygame/pygame/pull/3104 + https://github.com/pygame/pygame/pull/3106 + https://github.com/pygame/pygame/pull/3105 which itself rely on cython 3.0.0 being out along CPython 3.11~~.

But a builder script is already available here https://github.com/pmp-p/python-wasm-plus that will use git versions.

it's not just show-off i'm ready to support that platform in the long run and later provide a bridge for loading and running direcly android apk on the web.

tests (older last):
https://pmp-p.github.io/pygame-wasm/

initial test:
 ( debug version , full stdlib **oversized** ) https://pmp-p.github.io/pygame-wasm/pong-1/


MyreMylar:
This seems interesting. 

Looking at the example so far I have some questions about frame rates - pong seemed pretty sluggish on my Chrome browser - and load times. However it's cool to see some progress on pygame in a browser! 👍

pmp-p:
>  I have some questions about frame rates

framerate is fixed by the browser and the vendor of the device, usually on desktop it should match VSYNC but on more mobile platform it could be lowered by the vendor.

> pong seemed pretty sluggish

this one looks more vivid https://pmp-p.github.io/pygame-wasm/pygame.html?org.pygame.asteroids 
but it sometimes has a race condition for the filesystem and may need a reload.

Load time is always the big web problem. But don't worry there are plenty of things to remove to make the initial download time shorter.

MyreMylar:
Cool, the asteroids game does seem a lot snappier to run.

MyreMylar:
Aeroblaster also runs very well 👍 

I'm getting pretty confident this will be a good way to run pygame games over the web.

MyreMylar (approved):
OK, I gave this a good once over with my eyes as well as checking the hosted games, the current local tests and some general tinkering with a built version of this PR in my demo projects that exercise a good chunk of pygame.

My conclusion is that everything seems to run the same as normal on the non-web side and some of the web stuff is looking very promising. I think we should approve this PR and hope for more developments along this path in the future.

buildconfig/config.py
```diff
@@ -210,6 +217,11 @@ def main(auto=False):
         additional_platform_setup = open(
             os.path.join(BASE_PATH, 'buildconfig', "Setup_Darwin.in"), "r"
         ).readlines()
+    elif sysconfig.get_config_var('MACHDEP') == 'emscripten':
+        pass
```
ankith26, line 221: minor nitpick, unnecessary pass here

buildconfig/config_emsdk.py
```diff
@@ -0,0 +1,239 @@
+"""Config on Unix"""
```
ankith26, line 1: probably better to expand on this docstring with better related documentation

src_c/base.c
```diff
@@ -2094,11 +2095,21 @@ static PyMethodDef _base_methods[] = {
      "return an array struct interface as an interface dictionary"},
     {NULL, NULL, 0, NULL}};
 
+#if defined(BUILD_STATIC) && defined(NO_PYGAME_C_API)
+// in case of wasm+dynamic loading it could be a trampoline in the globals
+// generated at runtime.
+// when building static make global accessible symbol directly.
+static PyObject *pgExc_SDLError;
```
ankith26, line 2102: I think it's fine if this pointer would be unconditionally global.. would make the code simpler+any bugs related to this could be caught even on regular cpython

src_c/font.c
```diff
@@ -60,11 +60,15 @@ static PyObject *
 PyFont_New(TTF_Font *);
 #define PyFont_Check(x) ((x)->ob_type == &PyFont_Type)
 
-static int font_initialized = 0;
 static unsigned int current_ttf_generation = 0;
-static const char font_defaultname[] = "freesansbold.ttf";
+#if defined(BUILD_STATIC)
+static int font_initialized = 1;
```
ankith26, line 65: why is this change being made here? Why is this initialised at the beginning, even before the init function is called?

src_c/rwobject.c
```diff
@@ -709,11 +709,9 @@ _rwops_from_pystr(PyObject *obj)
             strcpy(extension, ext);
         }
         rw->hidden.unknown.data1 = (void *)extension;
-        Py_DECREF(oencoded);
         return rw;
     }
 
-    Py_DECREF(oencoded);
```
ankith26, line 716: the changes in this file should be reverted. This was a recent fix I made to fix an issue with DECREFs. Basically, `oencoded` should not be freed while `encoded` is still being used

ankith26 (commented):
Left a few reviews.

Another question: why is `void.c` (which is an empty file) being added?

src_c/base.c
pmp-p: i agree, but i'm not sure of the shields to put there because when C-API is active the global is a MACRO pointing to the slot.
unconditionally global would conflict with the _pygame.h macro https://github.com/pygame/pygame/blob/d72f4be879e47f426b565f56a5b5b89edc8aa12b/src_c/include/_pygame.h#L95

ankith26: Well I think there are some header guards that prevent this confict, IIRC

src_c/rwobject.c
pmp-p: right i missed the strrchr(encoded, '.');

src_c/font.c
pmp-p: yeah far from obvious that because SDL_Init + TTF_Init()  are made in main before CPython process the module inittab. so the emscripten handler knows it will use SDL2 next cycle.
i'll put a comment

buildconfig/config.py
pmp-p: indeed

buildconfig/config_emsdk.py
pmp-p: -on +like

pmp-p:
> Another question: why is void.c (which is an empty file) being added?

@ankith26 when the BUILD_STATIC flag is some each invidual module cannot be anymore compiled as shared objects. It's the simplest way i've found to express that they are no more compiled as single element. in the final version i expressed it and sorted them ( but the file is not matched with that PR ) https://github.com/pmp-p/pygame-wasm/blob/5c1d0757e4c81ee52839a80192fdf1082e4b0b52/buildconfig/Setup.Emscripten.SDL2.in#L33

src_c/mixer.c
```diff
@@ -100,8 +100,8 @@ struct ChannelData {
 static struct ChannelData *channeldata = NULL;
 static int numchanneldata = 0;
 
-Mix_Music **current_music;
-Mix_Music **queue_music;
+Mix_Music **mx_current_music;
```
illume, line 103: ps. We were trying to move everything to a pg_ prefix, but hadn't converted all the files yet.

illume (approved):
🎉 🎈 wonderful

illume:
Looks like the CI wasn't running in this PR.
When merged to main if failed here: https://github.com/pygame/pygame/runs/6480832728?check_suite_focus=true

pmp-p:
> It looks like the CI wasn't running in this PR for build-emsdk.

CI won't be able to fully does the job atm for multiple reasons :
 - There are some problems to fix in sdl2 first. The CI is using ubuntu-latest whose libtool is not able to link correctly SDL2_image. i'll try to fix that asap.
 - cython git generated code will break CI on   pypm.c and some C switches fallthrough.
it's also very slow to rebuild cpython it should use a cpython prebuilt beta/release but not git.

Starbuck5:
> cython git generated code will break CI on pypm.c and some C switches fallthrough even on native.

I'd be fine if we committed Cython 3 alpha10 generated code, that will work with the latest 3.11 versions I've tested.

Presumably it won't actually touch pypm.c, since the emscripten SETUP file doesn't try to build with it?

Still a problem with all the other cython modules though, since they're generated with cython 0.29.x, which I'm not sure will work with 3.11.

PR #3111, opened 2022-04-01
examples/mask.py change default image to one with alpha channel

brccabral:
chimp image has a red background. When calculating collisions, the red background is determining the collision, removing the red background from the chimp image the collisions happen only between chimps, not between backgrounds.

### DIFF
--- a/examples/mask.py
+++ b/examples/mask.py
@@ -138,9 +138,7 @@ def main(*args):
     images = []
     masks = []
     for impath in args:
-        im = pg.image.load(impath).convert_alpha()
-        im.set_colorkey('red') # remove red background
-        images.append(im)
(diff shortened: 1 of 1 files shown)


MyreMylar (changes requested):
this will only work for the current default chimp.png, not any other images supplied by a user to the example.

I think a better way to fix it right now is to swap the default image to one of the other example images that has an alpha channel rather than a colour key.

You can do this on line 207, changing it to:

```
imagename = os.path.join(main_dir, "data", "alien1.png")
```

brccabral:
Done, changed to alien1.png

MyreMylar (approved):

ankith26 (approved):
Thanks for the PR! 👍 🎉

PR #3112, opened 2022-04-01
examples/moveit.py slow down

brccabral:
Without clock.tick() it was running too fast to see any difference


MyreMylar:
There is an open PR #3092 regarding the moveit example which will conflict with this one. I think we should resolve that PR first before returning to see if there is anything else to do here.

brccabral:
PR 3092 applies the same change here, if that PR is merged then there is nothing to add from this one. Close it if you want to.

PR #3113, opened 2022-04-01
added SCALED flag

brccabral:
When using FULLSCREEN flag, if there is no SCALED with it my system throws X11 errors.

```
X Error of failed request:  BadRRCrtc (invalid Crtc parameter)
  Major opcode of failed request:  140 (RANDR)
  Minor opcode of failed request:  20 (RRGetCrtcInfo)
  Crtc id in failed request: 0x0
  Serial number of failed request:  234
  Current serial number in output stream:  234
```

SCALED is also needed if using `pygame.display.toggle_fullscreen()`
`pygame.error: X11_XRRSetCrtcConfig failed`

Simple tests can be run
```
import pygame

pygame.init()
# pygame.display.set_mode((640, 480), pygame.SCALED)
pygame.display.set_mode((640, 480), pygame.FULLSCREEN | pygame.SCALED)

while True:
        if event.type == pygame.QUIT:
            pygame.quit()
            if event.key == pygame.K_f:
                pygame.display.toggle_fullscreen()
    pygame.display.flip()

```

### DIFF
--- a/examples/scaletest.py
+++ b/examples/scaletest.py
@@ -33,7 +33,7 @@ def main(imagefile, convert_alpha=False, run_speed_test=False):
         return
 
     # start fullscreen mode
-    screen = pg.display.set_mode((1024, 768), pg.FULLSCREEN)
+    screen = pg.display.set_mode((1024, 768), pg.FULLSCREEN | pg.SCALED)
     if convert_alpha:
         background = background.convert_alpha()
 
--- a/examples/testsprite.py
+++ b/examples/testsprite.py
@@ -54,7 +54,7 @@ if "-flip" in sys.argv:
     flags ^= pg.DOUBLEBUF
 
 if "-fullscreen" in sys.argv:
-    flags ^= pg.FULLSCREEN
+    flags ^= pg.FULLSCREEN | pg.SCALED
 
 if "-sw" in sys.argv:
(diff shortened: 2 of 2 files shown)


MyreMylar:
Pretty sure this isn't a universal issue, so more likely something particular to your setup. For example on my Windows machine testsprite.py runs fine without in full screen without any changes to the example.

brccabral:
Hi @MyreMylar ,
I am running Ubuntu 20.04 on Wayland. I tested in Ubuntu 20.04 on Xorg and it works OK indeed, but on Wayland I get the error.
And that error above was when I had three screens, when using just my laptop screen the error is slightly different.


```
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  151 (XFree86-VidModeExtension)
  Minor opcode of failed request:  1 (XF86VidModeGetModeLine)
  Value in failed request:  0x21
  Serial number of failed request:  63
  Current serial number in output stream:  63
```

MyreMylar:
Sounds like a Wayland specific full screen issue then.

MyreMylar:
I opened an issue #3114 to look into this, feel free to add any other relevant details about your system (e.g. python, pygame and SDL versions).

PR #3116, opened 2022-04-04
gl_set_attribute() test: Take over of original PR by @anesantos

### LINKED ISSUE #1748: Add unit test: display.gl_set_attribute()

**Test stub:** https://github.com/pygame/pygame/blob/main/test/display_test.py#L255-L274
**Docs**: https://www.pygame.org/docs/ref/display.html#pygame.display.gl_set_attribute
**Code to be tested**: https://github.com/pygame/pygame/blob/main/src_c/display.c#L774-L787

MyreMylar:
Original PR #2556

Closes #1748

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -334,7 +334,14 @@ class DisplayModuleTest(unittest.TestCase):
         with self.assertRaises(TypeError):
             pygame.display.gl_get_attribute("DUMMY")
 
-    def todo_test_gl_set_attribute(self):
+    @unittest.skipIf(
+        (
+            "skipping for all because some failures on rasppi and maybe other platforms"
+            or os.environ.get("SDL_VIDEODRIVER") == "dummy"
+        ),
+        'OpenGL requires a non-"dummy" SDL_VIDEODRIVER',
+    )
+    def test_gl_set_attribute(self):
 
         # __doc__ (as of 2008-08-02) for pygame.display.gl_set_attribute:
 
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
LGTM on a skim through the code! Thanks for the PR! 👍 🎉

test/display_test.py
```diff
@@ -334,7 +334,14 @@ def test_gl_get_attribute(self):
...
             pygame.display.gl_get_attribute("DUMMY")
 
-    def todo_test_gl_set_attribute(self):
+    @unittest.skipIf(
+        (
+            "skipping for all because some failures on rasppi and maybe other platforms"
+            or os.environ.get("SDL_VIDEODRIVER") == "dummy"
+        ),
+        'OpenGL requires a non-"dummy" SDL_VIDEODRIVER',
+    )
```
Starbuck5, line 343: I don't get it. What's the point of a test that's always skipped?

illume (changes requested):
I know the other PR author is credited in the title already, but...

Could you please add the Co-authored-by to the end of your commit message? Then github links them to this commit. https://github.blog/2018-01-29-commit-together-with-co-authors/

`Co-authored-by: FirstName LastName <email@example.com>`

test/display_test.py
illume: Yeah, it's kind of difficult. Because OpenGL is broken in quite a few places I guess.
But, also maybe if we enable the test then people will fix OpenGL issues more.

Further work would be to slowly find, and special-case-skip all known broken places. But that could take a while getting tests from lots of different system config failures, and be risky/time consuming.

Hrmm. For me, merging the PR as is would be fine.

illume: We also have some opengl tag that's used to disable other opengl tests.

MyreMylar:
This article was very helpful in updating the commit by the way in case someone else runs into the issue of crediting someone with a hidden email:

https://www.nymeria.io/blog/how-to-manually-find-email-addresses-for-github-users

ankith26:
Hmm... This PR seems to have duplicate commits (the older one without co-author is still around).

Maybe that should be `drop`ped in a rebase?

PR #3122, opened 2022-04-06
Math's Vectors have not documented methods

### LINKED ISSUE #3121: Math's Vectors have not documented methods
In https://www.pygame.org/docs/ref/math.html there are no such methods as "copy".

This mistake is caused by this line:
https://github.com/pygame/pygame/blame/9ce94b0d0c48d088e896a10d924f6f3d812ff982/docs/reST/ref/math.rst#L368

Currently working on the fix.

gresm:
Fix  #3121

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -365,7 +365,7 @@ Multiple coordinates can be set using slices or swizzling
       .. ## Vector2.project ##
 
    
-   .. method :: copy
+   .. method:: copy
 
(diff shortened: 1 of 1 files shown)


MyreMylar:
The docs actually seem to build fine locally without this PR - I think there might just be a mismatch with what is on the website vs GitHub.

Regardless I agree that these methods spacing is inconsistent with the rest of the doc file so I'll approve this anyway.

ankith26 (approved):
Thanks for the PR! 👍 🎉

PR #3126, opened 2022-04-08
Cleanup old compatibility code in windows sysfont

Starbuck5:
- got rid of a Windows 95/98/ME compatibility check
- changed EnvironmentError to OSError, because:
> Changed in version 3.3: [EnvironmentError](https://docs.python.org/3/library/exceptions.html#EnvironmentError), [IOError](https://docs.python.org/3/library/exceptions.html#IOError), [WindowsError](https://docs.python.org/3/library/exceptions.html#WindowsError), [socket.error](https://docs.python.org/3/library/socket.html#socket.error), [select.error](https://docs.python.org/3/library/select.html#select.error) and mmap.error have been merged into [OSError](https://docs.python.org/3/library/exceptions.html#OSError), and the constructor may return a subclass.

( https://docs.python.org/3/library/exceptions.html )

- got rid of different checks around encoding the font name into a string. I believe these checks are not necessary in python 3, and on my system every font name was already given by winreg as a string.

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -65,42 +65,16 @@ def initsysfonts_win32():
     fonts = {}
 
     # add fonts entered in the registry
-
-    # find valid registry keys containing font information.
-    # http://docs.python.org/lib/module-sys.html
-    # 0 (VER_PLATFORM_WIN32s)          Win32s on Windows 3.1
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
LGTM! 👍

MyreMylar (approved):
LGTM 👍 

Tested locally too (on windows 10) and sysfont seems to work OK still

PR #3132, opened 2022-04-10
Change renderer names to strings in `get_drivers`

### LINKED ISSUE #3005: Remove bytes strings from _sdl2 module
I recently removed one from _sdl2.audio. I just saw that the Renderer names in the get render drivers function are byte strings. Perhaps there are more.

They should be rooted out and replaced with normal python3 strings before _sdl2 stuff is released.

zoldalma999:
Fix #3005

I went trough every _sdl2 module (including touch), to check if there were any bytes strings left, but could not find any apart from the one mentioned by Starbuck (in `get_drivers`).


ankith26 (approved):
LGTM! Thanks for the PR 👍 🎉

MyreMylar (approved):
👍

PR #3134, opened 2022-04-10
added FRect object

Emc2356:
a FRect object is similar to Rect but it is able to handle floats
currently it doesnt have proper docs nor tests
it also doesnt have any docs on the templated api that both Rect and FRect share


novialriptide (commented):
Is there any reason why FRect isnt in its own file?

Issue #3071, opened 2022-03-06
White surfaces blit with pygame.BLEND_MULT darken the target slightly


jeremycryan:
**Environment:**

You can get some of this info from the text that pops up in the console when you run a pygame program.

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Windows 10 Home v10.0.19044 Build 19044
- **Python version** (e.g. 3.7.9, 3.8.5) : 3.9.2
- **SDL version** (e.g. SDL 2.0.12): 2.0.18
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.1.2
- **Relevant hardware** (e.g. if reporting a bug about a controller, tell us the brand & name of it): N/A

**Current behavior:**

Using a blend mode of pygame.BLEND_MULT will darken pixels on the target surface if the value of the blit surface is pure white. If a pure white (255, 255, 255) surface is blit onto a pure white surface (255, 255, 255), the resulting surface will have color (254, 254, 254).

This results in unintended behavior and forces weird workarounds when the user wants to apply multiple multiply effects to a single surface (darkening the surface with each blit) or uses a colorkey for transparency (since colorkeys rely on exact pixel values, darkening them by one will "miss" the colorkey and draw the slightly-off color).

**Expected behavior:**

The behavior of multiply blend modes in image editing software (Photoshop, Paint.NET, etc.) is universally that solid white (255, 255, 255) will have no affect on the target, while solid black (0, 0, 0) will completely overwrite it. This is opposite for a blend mode of add, which pygame seems to respect (blitting solid black does not change the target).

**Screenshots**

Here's an image of three surfaces I've processed with pygame.BLEND_MULT blits:

![image](https://user-images.githubusercontent.com/22649301/156938110-0d738acc-5109-4230-af5a-54a01c67d03b.png)

- **Surface 1** (left) is a white (255, 255, 255) circle with a red (255, 0, 0) background and a colorkey of red. Drawing it on the black background correctly treats the red pixels as transparent and only draws the white circle.
- **Surface 2** (center) is identical to surface 1, except I have blit a new surface onto it with a multiply blend mode. The new surface has light blue (200, 220, 255) pixels in the circle region and white pixels outside that. As you can see, the white pixels slightly darken the target surface, resulting in a slightly-not-red color (254, 0, 0) that doesn't apply the colorkey because it's background color no longer matches the colorkey exactly.
[...]

PR #3136, opened 2022-04-14
Add SIMD versions of RGB_MUL blitter

MyreMylar:
Following the pattern from the RGBA_MUL pull request, this one adds accelerated versions for the RGB only multiply blend mode. 

The blend is slightly more complicated than RGBA, but the actual PR is shorter because the RGBA one already did most of the setup work.


ankith26:
Can't review this in detail, but I compiled and tested it locally and it seems to work fine... I can also confirm this PR fixes #3071 for me...
Although I have a question, does this PR fix the blendmode formula (for the linked issue) for the non SIMD codepath too?

MyreMylar:
> Can't review this in detail, but I compiled and tested it locally and it seems to work fine... I can also confirm this PR fixes #3071 for me... Although I have a question, does this PR fix the blendmode formula (for the linked issue) for the non SIMD codepath too?

The non-SIMD formula for RGB_MULT was changed at the same time as the formula was changed for RGBA_MULT, in PR #2988 .

Those changes needed to be intertwined to make the tests work, as some RGBA_MULT paths head down RGB_MULT eventually so the formulas have to match. This PR just means that all the remaining MULT paths (whether they start as MULT, RGB_MULT or RGBA_MULT) should now use SIMD accelerated code paths (where CPU hardware is available).

The next two special blend mode PRs are a lot simpler than these first two as there are no formula changes and the underlying formula is simpler anyway :)

itzpr3d4t0r:
I don't know if it's right to comment on this one, but i'm using pygame since 2017 and been building a big game since then. It has now come to reading documentation and discussions here on github for better performance in game so i'm really looking forward to any optimization to"basic" functionalities like this one. It's a nice coincidence that i read this commit as i'm now working with masked images and have to use pygame.BLEND_RGB_MULT for that so PLEASE, continue this commit, it would be really helpful! Oh and any suggestions on possible optimizations i can apply even without this commit to get a surface that's masked by another BW surface faster? thank you and continue the great job! love you guys!

Starbuck5:
@itzpr3d4t0r

> for better performance in game

When people are looking for optimizations I always recommend to profile your code with cProfile and snakeviz, so you can actually see where runtime is going. Here's a tutorial but it overcomplicates it a bit: https://audhiaprilliant.medium.com/python-runtime-profiling-using-snakeviz-how-to-inspect-the-codes-performances-b9183ced32ed.

All you need to do is
`python -m cProfile -o out.prof YOURFILE.py`
`pip install snakeviz`
`snakeviz out.prof`

itzpr3d4t0r:
> @Starbuck5

Yes man thank you, i've been using cProfile and snakeviz for about two years as well as matplotlib and timeit with custom functions for plotting and evaluating functions over different work loads/types, so i'm not really new to this world. I wrote the previous comment as i've noticed that a function shouldn't really take that long to run, and about 80% of the time it took was because of a blit that used pygame.BLEND_RGB_MULT as a flag. It basically creates a 24bit blank surface of about 330x330, then takes a bigger image(of about 1457x961 px) and blits a portion of that surface onto my smaller surface. then I take a black and white "mask" that's basically a vignette and blits it onto the 330x330 img. I've noticed that the logic doesn't take much, but out of the 2 blits, the one that's taking incredibly long is the pygame.BLEND_RGB_MULT one.
code is something like this:

> draw_img = pygame.surface.Surface((r.width, r.height))
> draw_img.blit(boxes_grid_color, (0, 0), r)
> draw_img.blit(grid_mask_img, (0, 0), None, 3)  <- this here takes quite a while

where:
- draw_img  is a 330x330px surface
- boxes_grid_color is a 24 bit 1457x961px
- grid_mask_img is a 24 bit 330x330px surface

And of course, having 2 blits will take more than just one and yes, blitting with RGB_MULT is "i think" slower than with a simple blit. I've tried everything, and i think i've dound the faster way i can. At this point i can only hope for these kinds of foundational optimizations only you guys can do. Maybe there is a way to get this to run faster with surfarray and some math+numpy? maybe that way is just slower tho...
I can't include real number because if there is a thing i hate about snakeviz is that it doesn't show correct blit time taken for each function and instead it shows the total blit time of every blit function. I was just cheering the work you've planned and are doing as well as noticing some small, yet meaningful bottlenecks.

illume (approved):
🎉 🎈

PR #3138, opened 2022-04-15
Sprite typestub updates, use generics and protocols

### LINKED ISSUE #2722: Group should be typed as Generic
**Description**

Pygame `Group` objects should probably be typed as a `Generic`. This would allow creating a `Group` of a particular subclass of `Sprite` without using `assert` everywhere to check the types.

In order to keep Mypy happy in the common case in which you create a `Group` without specifying a type (meaning generic `Sprite`), something like https://github.com/python/mypy/issues/4236#issuecomment-344660299 can be used.

### LINKED ISSUE #2726: Type hinting for Sprite image and rect should not be optional
I am not an expert on Pygame, but I think that when a `Sprite` object has an `image` or `rect` attribute, they must be a `Surface` and a `Rect` respectively. Currently they are being typed as `Optional`, which means that they can exist and be `None`.

Instead, the `Sprite` class should define them with their actual types. Mypy won't warn about them maybe not existing, but that also happens now.

ankith26:
PR to fix #2722 and fix #2726 

Updates to sprite typestubs now bring them more inline with code.
This PR makes use of `typing.Protocol` to mark arguments that need `.rect`, `.image` or `.mask` attributes on sprite subclasses, that are not present in the `Sprite` baseclass. Also makes `Group` and its subclasses a `Generic` that accepts a sprite subclass instance that supports `Group` (must have `rect` and `image` attributes in addition to being a subclass of `Sprite`)

Apart from that there are also a few more minor misc fixes in the same file


illume:
There's a conflict in here now.

ps. Reading the types I wouldn't know how what a _Group or AbstractGroup is. I wouldn't know where to import those, or find docs for those.

ankith26:
That is a valid point, and to my knowledge there is no other way of properly bringing types inline with the code in this case, that don't need source changes.

Kinda unrelated to this PR but it is related to the point you bring up, we have stuff like `RectValue`, `ColorValue` etc that are used a lot in typing but not documented. These currently live in `_common.pyi` that cannot be imported or used from python code.

I've come across some people asking questions about this, and also some people asking that this be exported properly in pygame so that they can use these as typehints in their own pygame code.

Would it be worth implementing all these in a pure python pygame submodule? Something like `pygame.types`?

MightyJosip (approved):
Nice changes thanks.

It doesn't brake anything and fixes some issues

MyreMylar (approved):
LGTM 👍

illume:
Very nice. I added your demo code to the PR description for passers by.

PR #3140, opened 2022-04-15
Improve clang format source file detection

zoldalma999:
Previously, on windows, `setup.py format` would format the doc headers, and did not format touch.c (Due to it being in `/_sdl2/`). This pr fixes both of these issues (and formats touch.c)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -793,6 +793,7 @@ class LintFormatCommand(Command):
         import subprocess
         import sys
         import warnings
+        import pathlib
 
         def check_linter_exists(linter):
             if shutil.which(linter) is None:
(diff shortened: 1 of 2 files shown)


ankith26 (approved):
LGTM, thanks for the PR 👍 🎉

setup.py
```diff
@@ -793,16 +793,42 @@ def run(self):
...
+                for disallowed in disallowed_files:
+                    if file.match(str(path_obj / disallowed)):
+                        break
+                else:  # no-break
+                    files.append(str(file))
+
+                for allowed in allowed_files:
+                    if file.match(str(path_obj / allowed)):
+                        files.append(str(file))
+                        break
```
MyreMylar, line 816: Since the allowed files must first be disallowed for this function to work, I think it should be possible to combine the if statements and reduce the code to one inner for loop:

```
for disallowed in disallowed_files:
    if file.match(str(path_obj / disallowed)) and not file.match(str(path_obj / allowed)):
        break
else:  # no-break
    files.append(str(file))
```

Or something like that. With the current design you could theoretically have a file that was in allowed, but not in disallowed added to the returned files list twice.

MyreMylar (changes requested):
LGTM, apart from I feel the filter function could be tightened up a little to avoid any future mess ups.

setup.py
ankith26: This has been fixed now by the looks of things, an allowed file won't be added twice if it's not disallowed. However the loop simplification you suggested may not work here because there are two different lists (allowed/disallowed)

illume (approved):
👍 🎈 thanks

Issue #3019, opened 2022-02-06
Performance of draw.polygon vs gfxdraw.filled_polygon


arganoid:
On my machine I find that drawing a 960x480 polygon with draw.polygon takes about 6ms, while drawing it with gfxdraw.filled_polygon takes a tenth of the time. Does anyone know why there is such a difference - and is there some reason why draw.polygon could not be brought up to the same speed?

Looking at the code I see that draw.polygon draws line by line, while gfxdraw.filled_polygon calls filledPolygonRGBA from the SDL_GFX library.

Starbuck5:
Hey @arganoid this was probably just fixed in #3141 

If you're still interested in testing, you can download a wheel from our continuous integration to test this locally.

It's a bit hard to describe how that would work.

If you're on windows: Go here, https://ci.appveyor.com/project/pygame/pygame/builds/43246331, pick out whichever python / (32/64 bit) you use, click on the run, go to the "artifact" tag, download and pip install [wheel filename].

If you're on Mac: Go here, https://github.com/pygame/pygame/actions/runs/2171769274, download the pygame wheels file at the bottom, pick out whichever wheel is right for you, pip install [wheel filename].

If you're on Linux: Go here, https://github.com/pygame/pygame/actions/runs/2171769272, download the pygame wheels monstrosity of a file, pick out whichever wheel is right for you, pip install [wheel filename].

I can't test your use because you didn't provide any testing code. But I'm pretty close to just assuming that this issue is fixed, maybe I'll leave this open a week or two just in case and close then.

PR #3141, opened 2022-04-15
Use more optimal line functions in polygon

### LINKED ISSUE #3135: draw.polygon slower in 2.1.2 than in 1.9.6
Recently I've noticed that an algo I made a while ago had become noticeably slow, so I plugged a profiler and found out that calls made to `pygame.draw.polygon` were causing my issue. Since I did not notice performance problems back then, I compared its speed with an older version of pygame.

```
$python
Python 3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:59:51) [MSC v.1914 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>>import pygame
pygame 1.9.6
>>>import timeit
>>>setup = """from random import randint
... import pygame
... import pygame.gfxdraw
... surf = pygame.Surface((120, 64))
... def half_triangle_poly(surface: pygame.Surface, multiplier: int):
...     w, h = surface.get_size()
...     top_left = (w*(1 - multiplier), h*(1 - multiplier))
...     top_right = (w*multiplier, top_left[1])
...     bottom_left = (top_left[0], h*multiplier)
...     return [top_left, top_right, bottom_left]
... """
>>>for scale in range(1, 9):
...    print(timeit.timeit(stmt=f'pygame.draw.polygon(surf,(128, 0, 255), half_triangle_poly(surf, {scale}))',
                           setup=setup, number=10000))
0.13838968499999993
0.21100721700000236
0.3219181770000006
0.4155462180000029
0.5251170810000048
0.6205457520000053
0.730503971999994
0.8374572599999937
```

```
$python
Python 3.7.0 (v3.7.0:1bf9cc5093, Jun 27 2018, 04:59:51) [MSC v.1914 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import pygame
pygame 2.1.2 (SDL 2.0.18, Python 3.7.0)
# same setup
>>>for scale in range(1, 9):
...    print(timeit.timeit(stmt=f'pygame.draw.polygon(surf,(128, 0, 255), half_triangle_poly(surf, {scale}))',
                           setup=setup, number=10000))
0.5657153599999987
2.387819206000003
5.963148969999992
11.077614591
17.871329568999997
26.282841297000004
36.43204714100003
48.14373332600002
```

Note that performance degradation seems worst when drawing a lot of stuff outside the surface's clip.

As mentioned in #3019 I've switched to `gfxdraw` and the performance is back to something similar to what I had in 1.9.6

Starbuck5:
Patch by MightyJosip

Fixes #3135 

Here's a more realistic test case:
```py
import time
import pygame
surf = pygame.Surface((1280, 720))
s = time.time()

for scale in range(1, 9):
    for _ in range(1000):
        pygame.draw.polygon(surf, "red", [[0,0], [0,100], [100,100]])
        pygame.draw.polygon(surf, "blue", [[150,100], [400,200], [300,500], [700,200]])

print(time.time() - s)
```

Numbers:
dev4 = 2.299182653427124 seconds
this patch = 0.255979061126709 seconds

Output is visually the same, I haven't checked pixel by pixel

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -2257,7 +2257,7 @@ draw_fillpoly(SDL_Surface *surf, int *point_x, int *point_y,
             minx = MIN(minx, point_x[i]);
             maxx = MAX(maxx, point_x[i]);
         }
-        draw_line(surf, minx, miny, maxx, miny, color, drawn_area);
+        drawhorzlineclipbounding(surf, color, minx, miny, maxx, drawn_area);
         PyMem_Free(x_intersect);
         return;
     }
@@ -2303,8 +2303,8 @@ draw_fillpoly(SDL_Surface *surf, int *point_x, int *point_y,
         qsort(x_intersect, n_intersections, sizeof(int), compare_int);
 
         for (i = 0; (i < n_intersections); i += 2) {
-            draw_line(surf, x_intersect[i], y, x_intersect[i + 1], y, color,
-                      drawn_area);
+            drawhorzlineclipbounding(surf, color, x_intersect[i], y,
+                                     x_intersect[i + 1], drawn_area);
         }
     }
 
@@ -2322,8 +2322,8 @@ draw_fillpoly(SDL_Surface *surf, int *point_x, int *point_y,
         y = point_y[i];
 
         if ((miny < y) && (point_y[i_previous] == y) && (y < maxy)) {
-            draw_line(surf, point_x[i], y, point_x[i_previous], y, color,
-                      drawn_area);
+            drawhorzlineclipbounding(surf, color, point_x[i], y,
+                                     point_x[i_previous], drawn_area);
         }
     }
     PyMem_Free(x_intersect);


Starbuck5:
As related to #3019, gfxdraw still seems to be much faster

```py
import time

import pygame
import pygame.gfxdraw

surf = pygame.Surface((1280, 720))

s = time.time()

for scale in range(1, 9):
    for _ in range(1000):
        pygame.draw.polygon(surf, "red", [[0,0], [0,100], [100,100]])
        pygame.draw.polygon(surf, "blue", [[150,100], [400,200], [300,500], [700,200]])

print(time.time() - s)

s = time.time()

for scale in range(1, 9):
    for _ in range(1000):
        pygame.gfxdraw.polygon(surf, [[0,0], [0,100], [100,100]], pygame.Color("red"))
        pygame.gfxdraw.polygon(surf, [[150,100], [400,200], [300,500], [700,200]], pygame.Color("blue"))

print(time.time() - s)
```

This yields:
0.25918126106262207
0.029999494552612305

So gfxdraw is still 8x faster in this specific case.

ankith26 (approved):
The code is LGTM, thanks for the PR 👍 🎉 

One minor issue I found with the benchmarking examples is that the `scale` variabled is varied from 1 to 9 but it's not actually used? Maybe we should benchmark the effects of that by thiss PR

MyreMylar (approved):
LGTM, simple change 👍

Starbuck5:
> One minor issue I found with the benchmarking examples is that the scale variable is varied from 1 to 8 but it's not actually used? Maybe we should benchmark the effects of that by this PR

It's just running the things 8000 times, and the scale thingy is inherited from the old test. That test is also much faster, matching 1.9.6. I just don't think it's a realistic test, since most people aren't going to have polygons 8x larger than the surface they're working with.

PR #3142, opened 2022-04-15
Remove Solaris compatibility code to re-enable SIGALARM

### LINKED ISSUE #2017: pygame import interferes with signal.alarm
Here's something interesting. Under some circumstances (second start after a crash, say), pygame.display.set_mode will hang indefinitely. So I have some wrapper code that uses `signal.alarm` to catch this and send a `KeyboardInterrupt` which kicks it off again. [lots more about this problem here.](https://stackoverflow.com/questions/17035699/pygame-requires-keyboard-interrupt-to-init-display) This issue is about something related but different:

At some point, something inside pygame seems to have started interfering with `signal.alarm`. If you run this sample code:

```
import signal, time
signal.alarm(10)
i = 0
while True:
    print(i, "seconds passed")
    i += 1
    time.sleep(1)

```

all is well and it quits after 10 seconds on the following environments:

```
Linux hostname 4.15.0-109-generic #110-Ubuntu SMP Tue Jun 23 02:39:32 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
Python 3.6.9 (default, Apr 18 2020, 01:56:04) 
>>> pygame.__version__
'1.9.3'
-
Linux hostname 4.19.97+ #1294 Thu Jan 30 13:10:54 GMT 2020 armv6l GNU/Linux
Python 3.7.3 (default, Dec 20 2019, 18:57:59) 
pygame 1.9.4.post1
```

But if you `import pygame` before requesting `signal.alarm`, it never fires and continues marching on eternally:

```
import signal, time
import pygame
signal.alarm(10)
i = 0
while True:
    print(i, "seconds passed")
    i += 1
    time.sleep(1)

```

```
>>> while True:
...     print(i, "seconds passed")
...     i += 1
...     time.sleep(1)
... 
0 seconds passed
1 seconds passed
2 seconds passed
3 seconds passed
4 seconds passed
5 seconds passed
6 seconds passed
7 seconds passed
8 seconds passed
9 seconds passed
10 seconds passed
11 seconds passed
12 seconds passed

```

Doesn't seem to matter which order the import occurs.

I also see the problem in Cythonized .pyx modules. python2 does it too.

Doesn't need to have `pygame.display.init()` or anything called in order to trigger this, the import itself suffices.

I don't know exactly when this began to occur, but it was sometime after April 2018.

MyreMylar:
Reasoning:

- Solaris seems long dead now as a desktop OS and is nearly dead on servers too: https://www.osnews.com/story/29994/oracle-kills-solaris/
- this code might not even be necessary there anyway as the bug was likely related to interactions between SDL 1 and Solaris, impossible to know conclusively without finding a Solaris desktop user
- Should fix #2017

Perhaps @combs can try it and confirm or anyone else running on linux.

This is the test case provided in the original issue:

```
import signal, time
import pygame
signal.alarm(10)
i = 0
while True:
    print(i, "seconds passed")
    i += 1
    time.sleep(1)
``` 

If all is well it should stop printing after around 10 seconds.

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -2040,20 +2040,6 @@ pg_install_parachute(void)
         }
     }
 
-#if defined(SIGALRM) && defined(HAVE_SIGACTION)
-    { /* Set SIGALRM to be ignored -- necessary on Solaris */
-        struct sigaction action, oaction;
-
-        /* Set SIG_IGN action */
-        memset(&action, 0, (sizeof action));
-        action.sa_handler = SIG_IGN;
-        sigaction(SIGALRM, &action, &oaction);
-        /* Reset original action if it was already being handled */
-        if (oaction.sa_handler != SIG_DFL) {
-            sigaction(SIGALRM, &oaction, NULL);
-        }
-    }
-#endif
 #endif
     return;
 }


ankith26 (approved):
LGTM! Thanks for the PR 👍 🎉 
Can confirm this fix works on Ubuntu 21.10 running under WSL+WSLg - windows 11. I could repro this issue on 2.1.2 
```
ankith@Ankiths-HP-Laptop:~$ python3 test.py
pygame 2.1.2 (SDL 2.0.16, Python 3.9.7)
0 seconds passed
1 seconds passed
2 seconds passed
3 seconds passed
4 seconds passed
5 seconds passed
6 seconds passed
7 seconds passed
8 seconds passed
9 seconds passed
10 seconds passed
11 seconds passed
12 seconds passed
13 seconds passed
14 seconds passed
^CTraceback (most recent call last):
  File "/home/ankith/test.py", line 8, in <module>
    time.sleep(1)
KeyboardInterrupt

ankith@Ankiths-HP-Laptop:~$ python3 test.py
pygame 2.1.3.dev5 (SDL 2.0.14, Python 3.9.7)
0 seconds passed
1 seconds passed
2 seconds passed
3 seconds passed
4 seconds passed
5 seconds passed
6 seconds passed
7 seconds passed
8 seconds passed
9 seconds passed
Alarm clock
```

I think at this point we can safely drop solaris support, unless we get an active contributor who is a solaris user (very unlikely given solaris is so dead these days)

Starbuck5:
Some brief research on this gets me the function SDL_InstallParachute from SDL 1.2.
https://www.libsdl.org/release/SDL-1.2.15/src/SDL_fatal.c

So it looks like it's something to do with installing a parachute, rather than an interaction between SDL1 and Solaris per se.

Starbuck5 (approved):
I haven't tested because I mainly use Windows, but this seems fine to remove.

illume:
> Solaris seems long dead now as a desktop OS

hehe. This OS is still a thing. It's called... illumos (openindiana/SmartOS) these days.

(Not saying this patch needs to be reverted.)

PR #3146, opened 2022-04-20
Updates to moveit.py example

dbrown3562:
This adds some extra functionality to moveit.py. This is sort of a companion pull request to PR [#3092](https://github.com/pygame/pygame/pull/3092) . MoveIt now explains how to handle some simple user input, as well as move an object in more directions than just right. The documentation has been updated accordingly.

Changes made:
The moveit example has been slowed down using clock.tick(), which is now reflected in the documentation (originally used time.clock.delay()).
Improved user input to allow for the user to move their object. Useful for learning more advanced input.
Objects can move in all 4 main directions (up, down, left, and right), to ensure the user understands objects can move in ways other than just right.
Unique player sprite.

Documentation now reflects these changes.

### DIFF
--- a/docs/reST/tut/MoveIt.rst
+++ b/docs/reST/tut/MoveIt.rst
@@ -469,5 +469,3 @@ issues. So when you're ready to keep learning, keep on reading. :-)
 Lastly, you can feel free to come to the pygame mailing list or chatroom
 with any questions on this stuff. There's always folks on hand who can help
 you out with this sort of business.
-
-Lastly, have fun, that's what games are for!
--- a/docs/reST/tut/MoveIt.rst
+++ b/docs/reST/tut/MoveIt.rst
@@ -378,7 +378,7 @@ user asks us to stop. ::
   >>> while True:
   ...     for event in pygame.event.get():
   ...         if event.type in (QUIT, KEYDOWN):
-  ...             sys.exit()
+  ...             return
   ...     move_and_draw_all_game_objects()
 
 
@@ -434,7 +434,7 @@ is what the main function for our program will look like. ::
   >>> while True:
   ...     for event in pygame.event.get():
   ...         if event.type in (QUIT, KEYDOWN):
-  ...             sys.exit()
+  ...             return
   ...     for o in objects:
   ...         screen.blit(background, o.pos, o.pos)
   ...     for o in objects:
@@ -561,7 +561,7 @@ sure we understand everything. ::
   ...         p.move(right=True)
   ...     for event in pygame.event.get():
   ...         if event.type in (QUIT, KEYDOWN):
-  ...             sys.exit()
+  ...             return
   ...     for o in objects:
   ...         screen.blit(background, o.pos, o.pos)
   ...     for o in objects:
--- a/docs/reST/tut/MoveIt.rst
+++ b/docs/reST/tut/MoveIt.rst
@@ -473,14 +473,14 @@ our move function under our GameObject class. ::
   ...       self.pos.top += self.speed
   ...   if up:
   ...       self.pos.top -= self.speed   
-  ...   if self.pos.right > 640:
+  ...   if self.pos.right > WIDTH:
   ...       self.pos.left = 0
-  ...   if self.pos.top > 420:
+  ...   if self.pos.top > HEIGHT-SPRITE_HEIGHT:
   ...       self.pos.top = 0
-  ...   if self.pos.right < 79:
-  ...       self.pos.right = 640
+  ...   if self.pos.right < SPRITE_WIDTH:
+  ...       self.pos.right = WIDTH
   ...   if self.pos.top < 0:
-  ...       self.pos.top = 420
+  ...       self.pos.top = HEIGHT-SPRITE_HEIGHT
 
 There's certainly a lot more going on here, so let's take it one step at a time.
 First, we've added some default values into the move function, declared as up,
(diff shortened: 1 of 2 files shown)


docs/reST/tut/MoveIt.rst
```diff
@@ -449,6 +452,127 @@ here it may not matter, but when objects are overlapping, using two loops
...
+  ...   if up:
+  ...       self.pos.top -= self.speed   
+  ...   if self.pos.right > 640:
+  ...       self.pos.left = 0
+  ...   if self.pos.top > 420:
+  ...       self.pos.top = 0
+  ...   if self.pos.right < 79:
+  ...       self.pos.right = 640
+  ...   if self.pos.top < 0:
+  ...       self.pos.top = 420
```
ankith26, line 483: Why are there hardcoded magic integers here? It would be better to use constants, since this is a newbie guide we need to introduce readers to some good programming practices here ;)

examples/moveit.py
```diff
@@ -49,23 +69,45 @@ def main():
...
+        if keys[pg.K_RIGHT]:
+            p.move(right=True)
+
+        #Draw the background
+        screen.blit(background, (0, 0))
+        for e in pg.event.get():
+
+            # quit upon screen exit
+            if e.type == pg.QUIT:
+                exit()
```
ankith26, line 99: It'd still be better to `return` here IMHO.... `exit` is not the best of functions to use to exit applications

ankith26 (changes requested):
Thanks for the PR! 👍 🎉 
Left some reviews, additionally the file probably needs a round of code formatting with `black`

examples/moveit.py
```diff
@@ -49,21 +76,42 @@ def main():
...
+        if keys[pg.K_RIGHT]:
+            p.move(right=True)
+
+        # Draw the background
+        screen.blit(background, (0, 0))
+        for e in pg.event.get():
+
+            # quit upon screen exit
+            if e.type == pg.QUIT:
+                return ()
```
ankith26, line 106:
```suggestion
                return
```
A plain return would do here. `return ()` returns an empty tuple

dbrown3562:
Hey all! Just checking up to see if you'd like any more changes.

ankith26 (commented):
I took a look at the doc tutorial and I found places where 
```py
if event.type in (QUIT, KEYDOWN):
    sys.exit()
````
Was done, so this should be updated too

File: docs/reST/tut/MoveIt.rst [hunk 4]
```diff
@@ -376,7 +378,7 @@ user asks us to stop. ::
   >>> while True:
   ...     for event in pygame.event.get():
   ...         if event.type in (QUIT, KEYDOWN):
-  ...             sys.exit()
+  ...             return
```
ankith26, line 381: Uhm I did not mean that this change should be a `return`. Here it should probably be `sys.exit()` itself because this is running in interactive prompt. 
What I meant was to make `if event.type in (QUIT, KEYDOWN):` to `if event.type == KEYDOWN:`

docs/reST/tut/MoveIt.rst
```diff
@@ -431,14 +434,14 @@ is what the main function for our program will look like. ::
...
```
ankith26, line 437: same as the above review

docs/reST/tut/MoveIt.rst
```diff
@@ -449,6 +452,132 @@ here it may not matter, but when objects are overlapping, using two loops
...
+  ...         p.move(up=True)
+  ...     if keys[pygame.K_DOWN]:
+  ...         p.move(down=True)
+  ...     if keys[pygame.K_LEFT]:
+  ...         p.move(left=True)
+  ...     if keys[pygame.K_RIGHT]:
+  ...         p.move(right=True)
+  ...     for event in pygame.event.get():
+  ...         if event.type in (QUIT, KEYDOWN):
+  ...             return
```
ankith26, line 564: needs changes here too

examples/moveit.py
```diff
@@ -14,17 +16,40 @@
...
 class GameObject:
     def __init__(self, image, height, speed):
         self.speed = speed
         self.image = image
         self.pos = image.get_rect().move(0, height)
 
-    def move(self):
-        self.pos = self.pos.move(self.speed, 0)
-        if self.pos.right > 600:
+    # move the object. Defaults to moving right.
```
ankith26, line 33: this comment is wrong, because the default here is to not move at all

examples/moveit.py
```diff
@@ -49,21 +76,42 @@ def main():
...
+            p.move(down=True)
+        if keys[pg.K_LEFT]:
+            p.move(left=True)
+        if keys[pg.K_RIGHT]:
+            p.move(right=True)
+
+        # Draw the background
+        screen.blit(background, (0, 0))
+        for e in pg.event.get():
+
```
ankith26, line 103: very minor nitpick, this spacing seems redundant

ankith26 (commented):
I tested it locally and it works well! 🥳 
Thanks again for the PR, I'm happy after a few rounds of reviewing and I'll go ahead and approve this 👍 

Left a few minor reviews to resolve

examples/moveit.py
dbrown3562: Could you clarify what you mean? Like extra blank lines?

ankith26: Yeah. The blank line below `for e in pg.event.get()` seems out of place to me, but yeah this is a minor aesthetic thing, feel free to ignore if you wish

ankith26 (approved):

docs/reST/tut/MoveIt.rst
```diff
@@ -449,6 +451,132 @@ here it may not matter, but when objects are overlapping, using two loops
...
+  ...     for o in objects:
+  ...         screen.blit(background, o.pos, o.pos)
+  ...     for o in objects:
+  ...         o.move()
+  ...         screen.blit(o.image, o.pos)
+  ...     pygame.display.update()
+  ...     clock.tick(60)
+
+A few things not mentioned earlier: we load in a second image and call it entity,
+and we user that for all objects that aren't the player, which uses the player
```
MyreMylar, line 573:
```suggestion
and we use that for all objects that aren't the player, which uses the player
```

MyreMylar (changes requested):
Looks good, just noticed one incorrect looking word.

MyreMylar (approved):
LGTM

PR #3148, opened 2022-04-22
Raise only pygame.error when mixer loading fails

### LINKED ISSUE #3139: Unhelpful error message suppresses true error
https://github.com/pygame/pygame/blob/main/src_c/mixer.c#L1744-L1759

It tries to load the sound
If it fails it tries to encode the filename as a string
If that fails it gives the ACTUAL error message from SDL
If it can encode the filename it suppresses the real error and just says "unable to open"

It seems like it would be more helpful to always just give the error from SDL. The FileNotFound case is already handled by the `pgRWops_FromObject`.

ankith26:
Simple PR to fix #3139

### DIFF
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -1742,20 +1742,7 @@ sound_init(PyObject *self, PyObject *arg, PyObject *kwarg)
         chunk = Mix_LoadWAV_RW(rw, 1);
         Py_END_ALLOW_THREADS;
         if (chunk == NULL) {
-            if (obj) {
-                PyErr_SetString(pgExc_SDLError, SDL_GetError());
-                return -1;
-            }
-
-            obj = pg_EncodeString(file, NULL, NULL, NULL);
-            if (obj == Py_None) {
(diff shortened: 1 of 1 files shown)


Starbuck5 (approved):
Awesome!

To test this, I just made a copy of setup.py and named it setup.mp3. If you try to load that as a sound, it used to say:
```py
>>> pygame.mixer.Sound("setup.mp3")
pygame.error: Unable to open file 'setup.mp3'
```

Now it says:
```py
>>> pygame.mixer.Sound("setup.mp3")
pygame.error: Unrecognized audio format
```
Which seems more helpful. 

Although now I'm realizing it's nice to have the filename in there as well, for when you're loading a list of things, for example. You think something like this would be a good idea?
```py
pygame.error: 'setup.mp3': Unrecognized audio format
```

I'm approving because this does the issue 100%, but I wonder what you think about my new idea.

ankith26:
I do like the idea you put up, and I think it can be extended to basically every pygame function that deals with resource loading so it's better if it got a dedicated PR probably

MyreMylar (approved):
LGTM 👍

PR #3150, opened 2022-04-23
Add _GenericVector to math typestubs, fixes

ankith26:
A followup PR for #3088, after this PR typecheckers should handle return types of some methods of vector subclasses correctly.

While working on this I realised there was a lot of code repetition in the `VectorN` stubs, so I added a `_GenericVector` private superclass for `VectorN` that implements the common methods of both vector objects.

Another idea of mine was to add `GenericVector` to C code itself, user code should continue to use `VectorN` but this `GenericVector` could exist only to make the C implementation easier and exist as an ABC for vectors maybe? This would be new API and needs more discussion so is something for the future, this PR only affects the typestubs for now


buildconfig/stubs/pygame/math.pyi
```diff
@@ -18,91 +18,195 @@ if sys.version_info >= (3, 9):
...
+        self: _TVec,
+        target: Union[Sequence[float], _TVec],
+        max_distance: float,
+    ) -> None: ...
+    @overload
+    def clamp_magnitude(self: _TVec, max_length: float) -> _TVec: ...
+    @overload
+    def clamp_magnitude(self: _TVec, min_length: float, max_length: float) -> _TVec: ...
+    @overload
+    def clamp_magnitude_ip(self, min_length: float) -> None: ...
```
novialriptide, line 122: This is supposed to be `clamp_magnitude_ip(self, max_length)`, it seems like there was a mistake with typing from the clamp_magnitude PR. Might as well fix it here.

novialriptide (changes requested):

buildconfig/stubs/pygame/math.pyi
ankith26: good catch!

buildconfig/stubs/pygame/math.pyi
```diff
@@ -18,91 +18,195 @@ if sys.version_info >= (3, 9):
...
     from typing import Collection
 
-_VectorTypeVar = TypeVar("_VectorTypeVar", Vector2, Vector3)
+_TVec = TypeVar("_TVec", bound=_GenericVector)
+
+# not implemented in code, only implemented here for ease of implementing
+# typestubs. Contains attributes/methods common to Vector2 and Vector3
+# Also used with _TVec generics
+class _GenericVector(Collection[float]):
+    epsilon: float
```
Starbuck5, line 27: Good old epsilon, typed, implemented, but not documented.

ankith26: does anyone know what it's for, lol? Might be worth document it if it has some significance

MyreMylar: https://en.wikipedia.org/wiki/Machine_epsilon

MyreMylar (approved):
LGTM 👍  Less duplication

buildconfig/stubs/pygame/math.pyi
ankith26: This already got documented in #3373

Starbuck5 (approved):
Typing is far too 🧠 for me right now, but this looks good as far as I can tell.

Thanks for working on this stuff Ankith!

ankith26:
Fixed merge conflicts in here.

Was giving a final scroll and I noticed a few more typing issues (not something this PR introduced but I will fix those here)

illume (commented):
> Another idea of mine was to add GenericVector to C code itself

Yes, this would be nice.

There is also the Vec4 and Matrix to consider, and a VectorN would be useful (maybe just called Vector). What I'm getting at is perhaps we should plan where these are going. Personally I have need for matrix and larger vectors. Also there is the immutable discussion.

With regard to type changes... I think we still have some work to do to make them more usable. Especially with testing changes are usable by humans. How do I type my pygame using function to take a ColorLike for example? How do we know type changes aren't breaking user programs? Can we automate more testing of types?

Issue #3143, opened 2022-04-17
Circle is drawing a white bar


novialriptide:
This issue has been brought up by imallet from the pygame community discord server, who has asked that someone else create the issue.

**Environment:**
- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Windows 10
- **Python version** (e.g. 3.7.9, 3.8.5) : 3.8.9
- **SDL version** (e.g. SDL 2.0.12): 2.0.20
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.1.3.dev4

**Current behavior:**
This should draw a black screen since the white circle is well off the screen. However, it produces a white bar instead because some internal variable is overflowing inside the rasterizer.

**Screenshots**
![image](https://user-images.githubusercontent.com/35881688/163695317-eb402fee-90d5-4f32-903a-2d492d08290e.png)

**Test code**
```
import pygame

display = pygame.display.set_mode((800,600))

running = True
while running:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            running = False

    display.fill((0,0,0))
    pygame.draw.circle(display, (255,255,255), (-1e30,300), 10 )
    pygame.display.flip()

pygame.quit()
```

PR #3151, opened 2022-04-24
Fix drawing of circles there intersect -+INT_MAX

jhilmer:
A fix for https://github.com/pygame/pygame/issues/3143

`draw_circle_filled` calls `drawhorzlineclipbounding` the arguments
was not protected for signed interger overflow - which produced a
write bar when e.g. a circle like:
`pygame.draw.circle(display, (255,255,255), (-1e30,300), 10 )` was
drawn.

This commit limit the `x` corrdinate to `drawhorzlineclipbounding`
so signed interger overflow is avoided.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1952,7 +1952,8 @@ draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
 
     if (x0 < 0) {
         xmin = x0 + INT_MAX + 1;
-    } else {
+    }
+    else {
         xmax = INT_MAX - x0;
     }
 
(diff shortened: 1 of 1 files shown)


robertpfeiffer:
Are you imallet's friend, or did you find this independently? I have a different fix for this, but I did not want to make a PR before I looked at this issue. I think we should merge both in any case, and this one before mine, for the sake of correctness.

jhilmer:
I just stumbled over this open bug, and thought I could fix it.

illume (changes requested):
@jhilmer could you please add a unit test for this?

illume:
Hey hey. Let me know if you want someone else to take over writing the test for this?

illume (approved):
👍 thanks.

It fixes the issue linked.

PR #3152, opened 2022-04-26
cull circles outside the clipping area

robertpfeiffer:
Drawing a circle previously meant running Bresenham's circle drawing
algorithm point by point, and then only setting pixels within the clip
rect. If the whole circle is outside of the clip rect, we can just stop
early and not run Bresenhams's algorithm at all.

Also fixes #3151, by not entering the faulty code path at all, but for the sake of correctness, we should do both.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -717,8 +717,10 @@ circle(PyObject *self, PyObject *args, PyObject *kwargs)
 
     SDL_GetClipRect(surf, &cliprect);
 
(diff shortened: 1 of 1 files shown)


ankith26 (commented):
This needs tests as already highlighted in the other PR...

How do we proceed with this, can we just merge this PR onto that? (it's easy to change the target branch on github UI)

MyreMylar (approved):
LGTM 👍

PR #3157, opened 2022-04-28
pass "-auto" flag into buildconfig

robertpfeiffer:


ankith26 (approved):
LGTM! 👍 🎉 
Thanks for the PR!

ankith26:
I also noticed `auto` is only a thing on the unix configuration, maybe a good future enhancement would be to extend it to other platforms? Although it could also happen on this PR if you wished 😉

MyreMylar (approved):
LGTM 👍

PR #3158, opened 2022-04-29
Fixed windows version parsing in Camera that errored on versions like 8.1

Matiiss:
...that have a "." in the name (e.g. Windows 8.1), now it converts only the integer before the first ".".


Starbuck5 (approved):
Looks good to me. Thanks Matiis!

Issue #3083, opened 2022-03-17
Use SDL_image image saving routines


Starbuck5:
In [pygame.imageext](https://github.com/pygame/pygame/blob/main/src_c/imageext.c), we have custom PNG and JPEG saving routines.     

SDL_image has routines for this, and it would be advantageous to use them. It would reduce our code footprint, maintenance burden, and mean we don't have to provide png and jpeg lib headers in our build system.

According to their changelog IMG_SavePNG is new in SDL_image 2.0.0 and IMG_SaveJPG is new in 2.0.2. I'd be okay raising our minimum supported SDL_image to 2.0.2 (as stated in the README) to support that all the time.

And imageext already uses IMG_SaveJPG if it can. It's also worth noting that @ThanhChinhBK just made our custom PNG saver support 8 bit surfaces in https://github.com/pygame/pygame/pull/3028. I believe the IMG_SavePNG function will support that out of the box but the contributed test will be handy to verify that.

Tasks:
- [x] Investigate removing custom jpeg saver and using IMG_SaveJPG all the time
- [ ] Investigate using IMG_SavePNG

FierceJDM:
@Starbuck5 Shall i open a pull request and add the changes to it ? Not very sure on how i should proceed

PR #3166, opened 2022-05-05
Remove old thread compat code

Starbuck5:
I noticed while looking into #3083 that some older files had these `WITH_THREAD` `#ifdef`s spread liberally around, cluttering the code up a bit and hurting readability.

I researched this and learned that Python used to be able to be built without threading support, which may have been advantageous when threads were exotic, in the python 2 days. The threads documentation states `Changed in version 3.7: This module used to be optional, it is now always available.`

So the only people potentially impacted by this are users on custom builds of Python 3.6, who also update pygame when new releases come out. I'd be surprised if this described any people. Plus we should drop Python 3.6 support soonTM anyways.

### DIFF
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -760,9 +760,7 @@ _ftfont_init(pgFontObject *self, PyObject *args, PyObject *kwds)
     else {
         PyObject *str = 0;
         PyObject *path = 0;
-#ifndef WITH_THREAD
-        goto end;
-#endif
+
         source = pgRWops_FromFileObject(original_file);
         if (!source) {
             goto end;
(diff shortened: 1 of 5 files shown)


ankith26 (approved):
LGTM! 🥳 

As discussed on discord, we should wait on this PR till we drop 3.6 support, which we should be doing soon

ankith26:
On a re-visit... I wonder how this PR effects @ pmp-p s current WASM port efforts, since I believe threads are not well supported with WASM?

illume (changes requested):
Seems threadless builds are still possible with patched python versions being used. We don't have a test for this or (static builds) so I expect there's a high chance of it being broken.

Starbuck5:
@ankith26 and @illume I have high confidence this will work in exotic environments.

1. `WITH_THREAD` is truly obscure. A search of the python docs, 2.7 or 3.x, yields 0 results for this.
2. All the macros guard is code talking about the GIL. A threadless python should have these functions, even if they do nothing.

Starbuck5:
I'm giving up on this.

I think there are simplifications we can make to the code here in the future, but it seems like this PR is never going to be merged, so I'll go ahead and close it.

PR #3167, opened 2022-05-05
Drop python 3.6

Starbuck5:
I almost definitely missed stuff, but I think I got the important bits.

Python 3.6 has been out of security fixes since 2021.

It would be good to consolidate our version support, reduce the number of wheels where possible, and get rid of old APIs.

Ankith I discussed CI needs on discord, it will still need work after this PR
I got the mac ones I think (If you think the distribution of versions is strange, I'm factoring in slots for 3.11 and more pypy versions)
I'm a bit scared to tackle pypy appveyor, so I left it, but that definitely needs to be updated (in another PR)

CI plan for Mac that Ankith likes (for the future):
Capture

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,13 +83,6 @@ jobs:
         # builds on macOS
         include:
           - { 
-            name: "MacOS x86_64 (cp36-cp37)",
-            os: macos-10.15,
-            macarch: x86_64,
-            pyversions: cp36-* cp37-*
-          }
-
-          - { 
             name: "MacOS x86_64 (cp37-cp38)",
             os: macos-10.15,
             macarch: x86_64,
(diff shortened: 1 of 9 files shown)


illume (changes requested):
We are not dropping 3.6 support yet.

Starbuck5:
@illume Please enlighten me why?

illume:
I don't see an urgent maintenance need to drop this. Seems like the benefit is mainly just dropping one python versions worth of wheels. As far as I'm aware there's no significant difficulty to continue to support it at the moment.

So I think we can look back next year to see how 3.6 is going.

Starbuck5:
We've started giving things deprecation warnings, but those are only emitted by default in Python 3.8 and above. So I would like to wait until Python 3.8 has been our minimum version for a while before removing any of those functions.

That's the main reason I want to start raising our minimum version.

The wheel filesize and CI build times are just ancillary benefits.

And just because we drop a version doesn't mean people can't use pygame on that version anymore, just that we've stopped providing updates. And in my experience, the majority of pygame users never even try to update.

Anyways, I'll drop it for now.

PR #3170, opened 2022-05-08
SIMD versions of RGB_ADD, RGBA_ADD, RGB_MUL & RGBA_MUL

MyreMylar:
Using the same blueprint as in the RGBA_MUL/RGB_MUL pull requests, this one adds, the much simpler to create, SIMD versions of the ADD and MUL special blend modes for blitting in SSE2 and AVX2 flavours.

If you found the SIMD code tricky to understand in the other two PRs this one is a lot more approachable - especially for the RGBA flavours as the 'active' code that actually blends the pixels from one surface with another is just one instruction, helpfully named exactly the same thing as the blend mode. See:

```
mm_dst = _mm_adds_epu8(mm_dst, mm_src);
```

which as intel describes it does:

> Add packed unsigned 8-bit integers in a and b using saturation, and store the results in dst.

Saturation meaning the values for each pixel are clamped between 0 and 255. In SSE2 we have 128 bit registers - so with four channel values between 0 and 255 in each pixel (red, green, blue and alpha) we can see that 128 divided by 4, then divided by 8 bits per channel, we get four pixels worth of data in each register and can thus speed through our adding operation pretty swiftly.

Everything else is just setup to get us to that single line that does the blend.

AVX2 version works the same except we can do 8 pixels at a time due to the larger registers.

There is one more of these PRs, after this one, for the remaining special effect blend modes MIN & MAX.

N.B.
I expect there is some future improvement work to be done on the standard alpha blend and the pre-multiplied blend, but I'd like to see how these special effect modes work out in user land first - plus I would need some time to work on those improvements when work is less busy :)

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -263,10 +263,88 @@ SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
                     break;
                 }
                 case PYGAME_BLEND_ADD: {
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+                    if (src->format->BytesPerPixel == 4 &&
+                        dst->format->BytesPerPixel == 4 &&
+                        src->format->Rmask == dst->format->Rmask &&
+                        src->format->Gmask == dst->format->Gmask &&
+                        src->format->Bmask == dst->format->Bmask &&
+                        !(src->format->Amask != 0 && dst->format->Amask != 0 &&
+                          src->format->Amask != dst->format->Amask) &&
+                        SDL_HasAVX2() && (src != dst)) {
+                        blit_blend_rgb_add_avx2(&info);
+                        break;
+                    }
+#if defined(__SSE2__)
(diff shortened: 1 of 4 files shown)


itzpr3d4t0r:
super cool! is the AVX2 version available to all or is it platform/hardware dependant?

MyreMylar:
> super cool! is the AVX2 version available to all or is it platform/hardware dependant?

AVX2 is available on x86/x64 which is basically all windows, pre-Arm Macs and linux flavours on that hardware. Arm platforms should still get a ported SSE2 version via SSE2neon.

evenetually we may try an upgrade from SSE2neon to SIMDe (https://github.com/simd-everywhere/simde) which pots AVX2 instructions to arm as well, but I wouldn't want to attempt such a switchover without an Arm development platform, something I don't have, but perhaps some future contributor will have as Arm Macs become more common.

illume (approved):
wonderful

illume:
Based on how long the last PR stayed up here... I'm not sure how many people want to read through this low level SIMD code. So I'll just merge this one now rather than wait for another reviewer.

Issue #1363, opened 2019-10-03
toggle_fullscreen tasks


illume:
- [ ] test on windows
- [ ] do not raise exceptions on unsupported platforms, simply return -1.
- [ ] write docs about toggle_fullscreen being platform-dependent? (for now)
- [ ] have a 'call toggle_fullscreen` type test case. So at least the function is run on all the different CI. The tests should not make it fullscreen by default. There should be a flag to run them on CI, and they should be an interactive test for testing locally.
- [ ] How should people toggle fullscreen? We should probably have an example which shows how to do it. For the longest time, the safest way is to basically reinit your app?
- [ ] How do we query if we are in fullscreen? I've kept that state in the app before. But really, it should query that from the system. Because on some systems you can send it to fullscreen from the OS. (like on Mac).

PurityLake:
So, should there be both an interactive test and a regular automated test for this?

PR #3171, opened 2022-05-09
Add test for fullscreen toggle

PurityLake:
So far I have a visual test for this functionality, making a PR to confirm with people that the test works on other machines.

Should there be an automated test also, or is a visual one ok for this kind of functionality?

Adding this regarding: #1363


test/display_test.py
```diff
@@ -782,6 +783,75 @@ def test_set_gamma_ramp(self):
...
+            self.screen.fill((255, 255, 255))
+            pygame.display.flip()
+            self.font = pygame.font.Font(None, 32)
+
+    def quit(self):
+        if self.screen is not None:
+            if self.isfullscreen:
+                pygame.display.toggle_fullscreen()
+            pygame.quit()
+        self.aborted = True
```
ankith26, line 816: Maybe this method should be named `tearDown`? So that it does cleanup of calling `pygame.quit()` even after a successful run. Also storing `self.aborted` state is kinda redundant IMO, can probably be removed.

ankith26 (commented):
Thanks for the first time PR! 👍 🎉 

As discussed on discord, this file also needs a run with the `black` code formatter to fix the failing lint checks

test/display_test.py
PurityLake: I've taken on board what you said and have make changes accordingly

test/display_test.py
```diff
@@ -844,7 +844,6 @@ def visual_test(self, fullscreen=False):
                     if event.key == pygame.K_n:
                         return False
                 if event.type == pygame.QUIT:
-                    self.quit()
```
ankith26: There is still a `self.quit()` a few lines above that you missed to remove

test/display_test.py
```diff
@@ -783,8 +783,9 @@ def test_set_gamma_ramp(self):
 
         pygame.display.quit()
 
+
 class FullscreenToggleTests(unittest.TestCase):
-    __tags__ = [ "interactive" ]
+    __tags__ = ["interactive"]
 
     screen = None
     font = None
```
ankith26, line 791: `aborted = False` defined here as a classvar is redundant

PurityLake:
@ankith26 made the changes you mentioned

test/display_test.py
```diff
@@ -783,6 +784,73 @@ def test_set_gamma_ramp(self):
...
+
+    screen = None
+    font = None
+    isfullscreen = False
+
+    WIDTH = 800
+    HEIGHT = 600
+
+    def setUp(self):
+        if self.screen is None:
```
ankith26, line 798: I think init should run regardless of the current `self.screen` state (because this function will overwrite it anyways)... also because `tearDown` does not set this to `None`, the `setUp` function will run only once, instead of before every test like we want it to

PurityLake: Fair point, made those changes for you now

ankith26 (approved):
Looks aight now, thanks! 😄 🥳

illume (approved):
👍 🎉 thanks

Issue #3131, opened 2022-04-09
Can't hold the windows in the center


simpia:
in 1.9.6,this setting will hold  the windows in the center even if the windows size changed

os.environ['SDL_VIDEO_CENTERED'] = '1'
pygame.display.set_mode((200, 200))
pygame.display.set_mode((500, 300))

in pygame2,the start position of the windows can't change

ankith26:
Thanks for reporting this issue!
Did a quick test locally on windows 11 and can confirm this issue reproduces.
To be noted that this issue only seems to happen when `set_mode` is called twice with different dimensions

PR #3172, opened 2022-05-10
Fixed window not staying centered on second call of `display.set_mode`

PurityLake:
References #3131

Fixed an issue where calling `display.set_mode` twice with the environment variable `SDL_VIDEO_CENTERED`.

The problem was the when `pg_set_mode` is called more than once, it finds the previous window that was created. If it finds this it overrides the x and y value of the centered position and causes the window to not center. I added a condition in the branch where a window is found, that was determined earlier in the function i.e. `enter_window`, now the window remains centered.

Here is some sample code for testing this issue:

```python
import pygame
import os

pygame.init()
os.environ['SDL_VIDEO_CENTERED'] = '1'
#print(os.environ['SDL_WINDOW_POS'])
screen = pygame.display.set_mode((100, 100))

play = True

while play:
        if event.type == pygame.QUIT:
            play = False
            break
            if event.key == pygame.K_SPACE:
                screen = pygame.display.set_mode((200, 200))
```

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1011,7 +1011,7 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                         if (y == (int)SDL_WINDOWPOS_UNDEFINED_DISPLAY(display))
                             y = SDL_WINDOWPOS_CENTERED_DISPLAY(display);
                     }
-                    else {
+                    else if (!center_window) {
                         SDL_GetWindowPosition(win, &x, &y);
                     }
                 }


ankith26:
Hi, thanks for the PR! 👍 🎉 

Quick local testing shows that this PR indeed fixes the linked issue, but I see that there is another corner case and the behaviour after this PR does not exactly match 1.9.6 behaviour.

I took your script and made a few tweaks to it to test resizing to different resolutions
```py
import os

import pygame

pygame.init()
os.environ["SDL_VIDEO_CENTERED"] = "1"
screen = pygame.display.set_mode((200, 200))

play = True
while play:
        if event.type == pygame.QUIT:
            play = False
            break

            if event.key == pygame.K_1:
                screen = pygame.display.set_mode((200, 300))

            if event.key == pygame.K_2:
                screen = pygame.display.set_mode((300, 200))

            if event.key == pygame.K_3:
                screen = pygame.display.set_mode((300, 500))
```

After this PR, the window recentres itself everytime `set_mode` is called, whereas pygame 2.1.2 never recentres after the first recenter.
Pygame 1.9.6 only seems to recentre if the resolution changed, otherwise it seems to remember the last screen coordinates, and this is what the issue poster of the linked issue wants too, IG.

I don't know whether this change in behaviour is trivial enough to be ignored, it would surely be nice to stick to old behaviour if possible (if it's too hard, then I'm not sure if it's worth workarounding but we can wait for others opinions on this) but apart from that, this PR LGTM! 🥳

PurityLake:
Ok, thanks for pointing that out, I'll be back at my dev computer in a while and I'll straight away take a look and see what I can do

PurityLake:
Is this PR going to be added?

ankith26:
Yes 😄 
Once a few more people review it and help in deciding which way we should go

ankith26:
Oh, no don't worry it was not rude at all 😃

src_c/display.c
```diff
@@ -1011,7 +1011,7 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                         if (y == (int)SDL_WINDOWPOS_UNDEFINED_DISPLAY(display))
                             y = SDL_WINDOWPOS_CENTERED_DISPLAY(display);
                     }
-                    else {
+                    else if (!center_window) {
                         SDL_GetWindowPosition(win, &x, &y);
```
ankith26, line 1015:
```suggestion
                    else {
                        int old_w, old_h;
                        SDL_GetWindowSize(win, &old_w, &old_h);

                        /* Emulate SDL1 behaviour: When the window is to be
                         * centred, the window shifts to the new centred
                         * location only when resolution changes and previous
                         * position is retained when the dimensions don't
                         * change.
                         * When the window is not to be centred, previous
                         * position is retained unconditionally */
                        if (!center_window || (w == old_w && h == old_h)) {
                            SDL_GetWindowPosition(win, &x, &y);
                        }
```
I think this should solve the issue I mentioned in my first review.

I tested with the modified version of your example I posted above and this patch does the trick and now the behaviour seems to match 1.9.6 behaviour accurately.

PurityLake: Ok now I feel like an idiot, when you said resolution, in my head I stupidly associated it with aspect ratio... That certainly recreates the behaviour

ankith26 (commented):
Again, thanks for contributing to pygame! 😎 🍰 

I'm not so good around `display.c` myself, but in this particular case I think the code is safe (it can't regress anything unrelated hopefully) so I am approving this (but we should wait for a another review or two before this is merged)

ankith26 (approved):

MyreMylar (approved):
LGTM 👍 

Confirmed it works locally.

PR #3173, opened 2022-05-10
Move ball movement to class

csagataj2:
The provided ball movement example does not work:

```
...\pygame\examples\go_over_there.py", line 73, in <module>
    o.position.move_towards_ip(target_position, o.speed * delta_time)
AttributeError: 'pygame.math.Vector2' object has no attribute 'move_towards_ip'
```

I fixed it by moving the `move_towards_ip` to the `Ball` class.

I also added some type hints-

### DIFF
--- a/examples/go_over_there.py
+++ b/examples/go_over_there.py
@@ -27,9 +27,18 @@ balls = []
 
 
 class Ball:
-    def __init__(self, position, speed):
+    def __init__(self, position: pg.Vector2, speed: float):
         self.position = position
         self.speed = speed
+
+    def move_towards_ip(self, target_position: pg.Vector2, amount: float):
+        distance_vec = target_position - self.position
+        length = distance_vec.length()
+        if length < amount:
+            self.position = target_position
+        else:
+            self.position += distance_vec.normalize() * amount
+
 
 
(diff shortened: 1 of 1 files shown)


ankith26 (changes requested):
Hi!
This example is not broken: `move_towards_ip` is a new pygame vector mector method added recently, and this example was added at the same time to showcase the usage of this function.

Thanks for taking the time and efforts to contribute, nonetheless

csagataj2:
Thanks for the quick reply. (I think) I have the latest version of the package but it still produces the same result. In which version will the `move_towards_ip` be included?
```
pip list
Package    Version
---------- -------
numpy      1.22.3
pip        22.0.4
pygame     2.1.2
PyOpenGL   3.1.6
setuptools 58.1.0
```

ankith26:
You have pygame `2.1.2` which is the latest stable version. `move_towards_ip` was added in [`2.1.3.dev4`](https://github.com/pygame/pygame/releases/tag/2.1.3.dev4) and will make it to the upcoming `2.1.3` stable release

csagataj2:
Thanks for the information, I will close this

PR #3176, opened 2022-05-11
fix typo in korean document

EunhoKang:


dr0id (approved):
I don't know that language... so I just trust your corrections.

Starbuck5 (approved):
Same as Dr0id, I don't know the language. I put it into Google translate and everything seemed fine though.

Thanks!

PR #3177, opened 2022-05-11
Add evfilter guard mutex, fix pymalloc without GIL (Minor event refactor part 2.1)

ankith26:
Initially this fix was part of #3062 but since that PR can take a while to review, I separated this smaller one out so these PRs can be merged independently (with this one going first and before 2.1.3 release, since this is more critical)

This bug seems to trigger when `pygame.event.wait()` is used (which makes sense because this function releases GIL).

To reproduce, run an example that uses the wait function python "dev mode" (this mode enables python runtime checks which catch and report this issue)
`py -X dev -m pygame.examples.mask` gives output (windows)
```
pygame 2.1.0 (SDL 2.0.16, Python 3.10.1)
...
Fatal Python error: _PyMem_DebugMalloc: Python memory allocator called without holding the GIL

Thread 0x000054a0 (most recent call first):
  <no Python frame>

Current thread 0x000026bc (most recent call first):
  File "...\Python310\lib\site-packages\pygame\examples\mask.py", line 175 in main
  File "...\Python310\lib\site-packages\pygame\examples\mask.py", line 208 in <module>
  File "...\Python310\lib\runpy.py", line 86 in _run_code
  File "...\Python310\lib\runpy.py", line 196 in _run_module_as_main
```

Also adds some extra mutex safeguards (in order to not hurt performance, these are only held in the branches where they are needed)

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -96,21 +96,27 @@ static SDL_Event _pg_repeat_event;
 static SDL_Event _pg_last_keydown_event = {0};
 
 #define PG_LOCK_EVFILTER_MUTEX                                             \
-    if (SDL_LockMutex(pg_evfilter_mutex) < 0) {                            \
-        /* TODO: better error handling with future error-event API */      \
-        /* since this error is very rare, we can completely give up if     \
-         * this happens for now */                                         \
-        printf("Fatal pygame error in SDL_LockMutex: %s", SDL_GetError()); \
-        PG_EXIT(1);                                                        \
+    if (pg_evfilter_mutex) {                                               \
+        if (SDL_LockMutex(pg_evfilter_mutex) < 0) {                        \
+            /* TODO: better error handling with future error-event API */  \
+            /* since this error is very rare, we can completely give up if \
+             * this happens for now */                                     \
+            printf("Fatal pygame error in SDL_LockMutex: %s",              \
+                   SDL_GetError());                                        \
+            PG_EXIT(1);                                                    \
+        }                                                                  \
     }
 
-#define PG_UNLOCK_EVFILTER_MUTEX                                             \
-    if (SDL_UnlockMutex(pg_evfilter_mutex) < 0) {                            \
-        /* TODO: handle errors with future error-event API */                \
-        /* since this error is very rare, we can completely give up if       \
-         * this happens for now */                                           \
-        printf("Fatal pygame error in SDL_UnlockMutex: %s", SDL_GetError()); \
-        PG_EXIT(1);                                                          \
+#define PG_UNLOCK_EVFILTER_MUTEX                                           \
+    if (pg_evfilter_mutex) {                                               \
+        if (SDL_UnlockMutex(pg_evfilter_mutex) < 0) {                      \
+            /* TODO: handle errors with future error-event API */          \
+            /* since this error is very rare, we can completely give up if \
+             * this happens for now */                                     \
+            printf("Fatal pygame error in SDL_UnlockMutex: %s",            \
+                   SDL_GetError());                                        \
+            PG_EXIT(1);                                                    \
+        }                                                                  \
     }
 
 static Uint32
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -620,10 +620,12 @@ static PyObject *
 pgEvent_AutoQuit(PyObject *self, PyObject *_null)
 {
     if (_pg_event_is_init) {
+        PG_LOCK_EVFILTER_MUTEX
         if (_pg_repeat_timer) {
             SDL_RemoveTimer(_pg_repeat_timer);
             _pg_repeat_timer = 0;
         }
+        PG_UNLOCK_EVFILTER_MUTEX
         /* The main reason for _custom_event to be reset here is so we
          * can have a unit test that checks if pygame.event.custom_type()
          * stops returning new types when they are finished, without that
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -76,6 +76,13 @@ static int _pg_event_is_init = 0;
  * arbitrary number you like ;) */
 #define MAX_SCAN_UNICODE 15
 
+/* SDL mutex to be held in the event filter when global state is modified.
+ * This mutex is intentionally immortalised (never freed during the entire
+ * duration of the program) because its cleanup can be messy with multiple
+ * threads trying to use it. Since it's a singleton we don't need to worry
+ * about memory leaks */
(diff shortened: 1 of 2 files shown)


src_c/event.c
```diff
@@ -418,7 +442,10 @@ _pg_remove_pending_VIDEOEXPOSE(void *userdata, SDL_Event *event)
...
 
 /* SDL 2 to SDL 1.2 event mapping and SDL 1.2 key repeat emulation,
- * this can alter events in-place */
+ * this can alter events in-place.
+ * This function can be called from multiple threads, so a mutex must be held
+ * when this function tries to modify any global state (the mutex is not needed
+ * on all branches of this function) */
 static int SDLCALL
 pg_event_filter(void *_, SDL_Event *event)
 {
```
dr0id, line 451: Why do `    int x, y, i;` not need a lock?

src_c/event.c
```diff
@@ -590,6 +620,13 @@ pgEvent_AutoInit(PyObject *self, PyObject *_null)
...
+        if (!pg_evfilter_mutex) {
+            /* Create mutex only if it has not been created already */
+            pg_evfilter_mutex = SDL_CreateMutex();
+            if (!pg_evfilter_mutex)
+                return RAISE(pgExc_SDLError, SDL_GetError());
+        }
+
         SDL_SetEventFilter(pg_event_filter, NULL);
     }
     _pg_event_is_init = 1;
```
dr0id, line 632: not sure, what about the `_joy_map_*` methods?

src_c/event.c
```diff
@@ -590,6 +620,13 @@ pgEvent_AutoInit(PyObject *self, PyObject *_null)
     if (!_pg_event_is_init) {
```
dr0id, line 620: What about `pg_EnableKeyRepeat` and `pg_GetKeyRepeat`? Can they be called from another thread than the one that is currently running code from `pg_event_filter`?

src_c/event.c
ankith26: Those functions have nothing to do with the event filter, so they don't need mutex safeguards and such

src_c/event.c
ankith26: Yes, those can be called from a different thread (the main thread).. They modify state that the event filter reads (but does not write)... So I thought a mutex is not needed in this case... but maybe for extra safety it can be added I agree

src_c/event.c
```diff
@@ -86,6 +93,25 @@ static int pg_key_repeat_interval = 0;
...
+    if (SDL_UnlockMutex(pg_evfilter_mutex) < 0) {                            \
+        /* TODO: handle errors with future error-event API */                \
+        /* since this error is very rare, we can completely give up if       \
+         * this happens for now */                                           \
+        printf("Fatal pygame error in SDL_UnlockMutex: %s", SDL_GetError()); \
+        PG_EXIT(1);                                                          \
+    }
 
 static Uint32
 _pg_repeat_callback(Uint32 interval, void *param)
```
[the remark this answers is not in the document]
ankith26: Yes, `_pg_repeat_callback` gets called in a new thread... Good catch, this needs some mutex safety too

dr0id (approved):

Starbuck5:
So I've been reading up on this just a little. Wouldn't it also work to replace the PyMem stuff around `_pg_last_keydown_event` with malloc and free?

ankith26:
That would work, too (infact was my original fix). But I think using a global (static variable) is more elegant here, because `malloc` can error.

Starbuck5:
Had a conversation with Ankith on discord that I should put up here for visibility.

I was questioning the use of the mutex, since getting rid of the memory allocation through Python seems like it's good enough to fix this.

It seems to me that a mutex on our end is unnecessary, since the event filter callback is guarded by a mutex on SDL's end. https://github.com/libsdl-org/SDL/blob/main/src/events/SDL_events.c#L1121

ankith26:
I see your point, and now that you showed me the relevant SDL code I'm convinced that `event_filter`, even though it can be called from multiple threads, can only have one thread running at a time. However I'd still like point out that this is an undocumented SDL implementation detail.

Another reason why the mutexes are still needed, because we can still have something like user code calling `pygame.key.set_repeat()` on a python thread, the event filter running on the main thread after a `pygame.event.wait()` and a key repeat timer thread running on another background C thread and all these access common state and can execute concurrently.

Even though the chances of something like this happening are really tiny and the code to trigger this must be really obscure, better safe than sorry since this is still a possibility! I personally don't think having this extra safety hurts readability or performance.

Starbuck5:
> However I'd still like point out that this is an undocumented SDL implementation detail so it could be subject to change in the future. In general we shouldn't rely on implementation details, they might break in a future version.

Ankith! This is not an implementation detail, it's thread safety. If SDL wasn't internally thread safe, that would be a bug in SDL.

>  I personally don't think having this extra safety hurts readability or performance.

I disagree. Imagine you're trying to maintain some 20 year old C code (sound familiar?) - one has a mutex and one doesn't. The one with a mutex is scarier and harder to refactor for anybody.

> Another reason why the mutexes are still needed, because we can still have something like user code calling pygame.key.set_repeat() on a python thread, the event filter running on the main thread after a pygame.event.wait() and a key-repeat timer thread running on another background C thread and all these access common state and can execute concurrently.

Ugh, fine. :smile:

This is still two separate issues, two separate PRs?
One "eliminate use of Python APIs in code where GIL is not held" ?
One "add mutex around shared event system state for uber safety" ?

Can the mutex be narrower than it is now with what we've discovered about SDL?

Starbuck5 (approved):
Alrighty, I'm convinced! Thanks for working to make the Event system more robust Ankith! 🚀 

Pointed out one location missing a lock (in autoquit), you can make a change and then merge as you see fit.

ankith26:
> This is not an implementation detail, it's thread safety. If SDL wasn't internally thread safe, that would be a bug in SDL.

Well... hypothetically SDL code could still do something like (psuedo-code)

```
acquire_internal_mutex()
local_callback = global_event_filter_callback
release_internal_mutex()
local_callback()
```

And this would be threadsafe on SDL end and `local_callback` could still execute concurrently which is why I said it's an implementation detail

> This is still two separate issues, two separate PRs?
One "eliminate use of Python APIs in code where GIL is not held" ?
One "add mutex around shared event system state for uber safety" ?

I agree, but I've split this PR many times already and I'm tired 😉 

Thanks for the approval! I will be adding the lock on the quit too.

illume (commented):
:tada:

PR #3178, opened 2022-05-15
Added quotation marks to prevent name error

sulphatet:
in line 57, entering QUIT without the quotation raises a name error on my version on PyGame, adding quotations removed the error

### DIFF
--- a/docs/reST/tut/tom_games2.rst
+++ b/docs/reST/tut/tom_games2.rst
@@ -54,7 +54,7 @@ The full code for this example looks like this::
       # Event loop
       while True:
           for event in pygame.event.get():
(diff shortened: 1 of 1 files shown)


ankith26 (changes requested):
Hi, thanks for taking the time to contribute to pygame! 🚀 

This fix does not work, because marking `QUIT` in quotes makes it a string, and a string never compares equal to `event.type` which is an integer constant.

`QUIT` is a name supposed to be imported when `from pygame.locals import *` runs, and that is being done in this example, this code should have worked.

Alternatively, replacing `QUIT` with `pygame.QUIT` should also work because this name is available in the main pygame namespace too

PR #3180, opened 2022-05-16
Remove custom JPEG saver

Starbuck5:
Part 1 of #3083

Pygame.imageext had implemented it's own support for saving jpeg images, through jpeglib.

This is cool and all, but SDL_image has the functionality. And so this removes a bit of hard to maintain code, reducing the footprint of imageext. It also means we don't need to distribute jpeg headers to pygame build environments. Also, since the SDL_image backend was preferred anyway, this won't even change the code path that runs to save a jpeg on normal installations of pygame.

This raises our minimum version of SDL_image to 2.0.2, as that is when the SDL_image jpeg saving functionality was added. SDL_image 2.0.2 was released in 2017.


ankith26 (approved):
Nice cleanups, thankies! 🥳 👍 

I was about to ask you to stall this PR for a couple of months because debian stretch still uses SDL_image 2.0.1 and its in LTS support phase (EOL in a few months). But the same debian stretch also uses python 3.5 which we've already dropped support for a few months back and did not get any complaints, so maybe most stretch users already migrated early to buster/bullseye or are sticking with older pygame versions.

On Ubuntu, 18.04 is the oldest LTS (this ships with python 3.6) and this has SDL_image 2.0.3 so we are safe on this.
Surprisingly... Ubuntu 14.04 and 16.04 still get security patches under ubuntus ESM scheme, again, these use python 3.4 and 3.5 respectively so we have already dropped support for this a long time back (I think ESM applies to servers anyways)

illume (approved):
👍 🎈 thanks

PR #3181, opened 2022-05-16
Add SIMD versions of RGB & RGBA BLEND_MAX, BLEND_MIN blit modes 

MyreMylar:
Following in the footsteps of the other special blend mode SIMD version PRs, this one adds the final set of currently existing special effect blend modes (i.e. we will have covered all the blit modes excluding normal non-alpha blitting, normal alpha blending and pre-multiplied alpha blending).

Don't think there is much else to say that hasn't been gone over in the last three PRs. 👍


src_c/simd_blitters_avx2.c
```diff
@@ -582,3 +582,321 @@ blit_blend_rgb_sub_avx2(SDL_BlitInfo *info)
...
+                n, post_8_width);
+        }
+        srcp += srcskip;
+        dstp += dstskip;
+    }
+}
+#else
+void
+blit_blend_rgba_max_avx2(SDL_BlitInfo *info)
+{
```
ankith26, line 656: Just thinking... should places like this raise an exception? I believe this function is defined as an empty function when avx is not available at compile time, but this gets called when avx is available at runtime? In this case, no blitting would happen at runtime and there'd be no warning/error too?

Another alternative is to use the non-SIMD path in this case, and warn something like this at runtime `"your CPU supports AVX2 but pygame was not compiled to use it. Falling back to slower non-SIMD codepath."`

ankith26: All wheels pygame ships should have avx2 support... but maybe some linux distro packagers would not enable it at compile time?

MyreMylar: Hmm. I'll have a quick think about it. 

I do want to be careful not to accidentally spread avx2 optimizations beyond this .c file, as I know that is an issue that even google managed to mess up, but I think it should be OK in this case as we are in normal C land to call another function from the simd_blitters.h file. Right now it is not possible to compile pygame without SSE2 support so we don't really need to worry further than that...

Yeah I think it should work.

ankith26: Nice, this seems like it should work now! 
But maybe this should raise a warning too before falling back to sse2?

Also, theoretically there's a possibility that the sse2 fallback is called on a CPU that does not support it, but given that this block is only executed after checking for avx2 ability, this should not cause any problems practically because I imagine no CPU that supports avx2 would unsupport sse2

illume:
The emscripten stuff was merged, and it's not configured to do SIMD yet. 
Would you mind re-basing against main and adding the guard to this find blitter as well?

illume (changes requested):
needs emscripten simd guards

src_c/simd_blitters_avx2.c
```diff
@@ -4,6 +4,14 @@
 #include <immintrin.h>
 #endif /* defined(HAVE_IMMINTRIN_H) && !defined(SDL_DISABLE_IMMINTRIN_H) */
 
+#define RAISE_AVX2_RUNTIME_SSE2_COMPILED_WARNING()     \
```
illume, line 7: ps. I'm not sure the consequences of this warning. I think warnings only get printed once in modern pythons (but I can't remember if it's 3.6+). We can remove the warning later if it turns out to spam the console.

illume (approved):
🎉🎈

(thanks for adding the emscripten guard too!)

PR #3183, opened 2022-05-18
Regenerate cython files

Starbuck5:
Why?

I believe pmp-p's automated emscripten workflow needs this. This checks in Cython files that can be built with Python 3.11 beta1.


ankith26:
This does not fix the failing emsdk workflow because that has another issue, but regenerating cython files is a good update nonetheless

Starbuck5:
I knew it wouldn't fix the first issue, but this might be the second hidden issue 😄 

Also this will just smooth things out. Like people will be able to build from source for 3.11 more easily.

novialriptide:
LGTM
image

novialriptide (approved):
LGTM
image

vnmabus:
May I ask why are generated files in version control to begin with?

pmp-p:
@vnmabus
main reasons that come to my mind are:
 - providing those files allow to pin a specific version of cython for bug tracking ( reproducibility ). 
 - faster CI builds.
 - because a big portion of C code of pygame module is cython generated C code that usually don't need to change for a whole CPython release cycle so no need force users to install cython and regen them for pygame stable.

vnmabus:
>     * providing those files allow to pin a specific version of cython for bug tracking ( reproducibility ).

Is it not possible to generate them as part of the release process?
 
>     * faster CI builds.

Do we have evidence that the generation of Cython files causes an important slowdown here?

>     * because a big portion of C code of pygame module is cython generated C code that usually don't need to change for a whole CPython release cycle so no need force users to install cython and regen them for pygame stable.

Again, is it not possible to generate them as part of the release process?

pmp-p:
> Do we have evidence that the generation of Cython files causes an important slowdown here?

for wasm CI build i can testify for yes. Anything not prebuilt or pre-generated is a slowdown

Starbuck5:
@vnmabus I feel like you're judging us for some reason. We are indeed aware that committing generated files is not conventional.

I've personally tried to get them generated at build time here: #2916

But honestly, just committing them has benefits too, and it's not the only generated file committed (src_c/doc/*.h), so I don't think anyone cares enough to change it at this point.

vnmabus:
@Starbuck5 Yes, I was only worried because I have seen a lot of people (myself included) committing generated files to version control and every time it turned out to be a BAD IDEA™.

zoldalma999 (approved):
Thanks!

PR #3186, opened 2022-05-20
Refactor font.render

Starbuck5:
Why?
Well, I kinda just started to do it... spent some time on it... then finished doing it.

It gets rid of some bit rot, simplifies the code a bit, adds a couple new comments. I hope this makes the font.render function a better place for future expansions with recent SDL_ttf features.

### DIFF
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -389,7 +389,7 @@ font_render(PyObject *self, PyObject *args)
     }
 
     SDL_Color foreg = {rgba[0], rgba[1], rgba[2], SDL_ALPHA_OPAQUE};
-    /* might be overriden right below, with an explicit background color */
+    /* might be overridden right below, with an explicit background color */
     SDL_Color backg = {0, 0, 0, SDL_ALPHA_OPAQUE};
 
     if (bg_rgba_obj != Py_None) {
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -416,7 +416,7 @@ font_render(PyObject *self, PyObject *args)
         }
     }
 
-    if (PyBytes_Check(text)) {
+    else if (PyBytes_Check(text)) {
         /* Bytes_AsStringAndSize with NULL arg for length emits
            ValueError if internal NULL bytes are present */
         if (PyBytes_AsStringAndSize(text, (char **)&astring, NULL) == -1) {
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -371,15 +371,14 @@ static PyObject *
 font_render(PyObject *self, PyObject *args)
 {
     TTF_Font *font = PyFont_AsFont(self);
-    int aa;
+    int antialias;
     PyObject *text, *final;
-    PyObject *fg_rgba_obj, *bg_rgba_obj = NULL;
+    PyObject *fg_rgba_obj, *bg_rgba_obj = Py_None;
     Uint8 rgba[] = {0, 0, 0, 0};
     SDL_Surface *surf;
-    SDL_Color foreg, backg;
-    int just_return;
+    const char *astring = "";
 
-    if (!PyArg_ParseTuple(args, "OpO|O", &text, &aa, &fg_rgba_obj,
+    if (!PyArg_ParseTuple(args, "OpO|O", &text, &antialias, &fg_rgba_obj,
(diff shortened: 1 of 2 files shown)


Starbuck5:
This was intended as purely a code quality thing, but benchmarking shows this happens to give a 4% speedup too.

src_c/font.c
```diff
@@ -371,15 +371,14 @@ static PyObject *
...
-    int aa;
+    int antialias;
     PyObject *text, *final;
-    PyObject *fg_rgba_obj, *bg_rgba_obj = NULL;
+    PyObject *fg_rgba_obj, *bg_rgba_obj = Py_None;
     Uint8 rgba[] = {0, 0, 0, 0};
     SDL_Surface *surf;
-    SDL_Color foreg, backg;
-    int just_return;
+    const char *astring = "\x00";
```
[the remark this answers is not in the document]
Starbuck5: You're right. I thought it wouldn't work, as I've had weird experiences with 0 length arrays in C. But I guess this ends up as a 1 length array.

Starbuck5:
I made that change you suggested Ankith, then I squashed and rebased to make things tidy.

src_c/font.c
```diff
@@ -388,121 +387,83 @@ font_render(PyObject *self, PyObject *args)
...
-        astring = PyBytes_AsString(bytes);
-        if (strlen(astring) != (size_t)PyBytes_GET_SIZE(bytes)) {
-            Py_DECREF(bytes);
+        if (strlen(astring) != (size_t)_size) {
             return RAISE(PyExc_ValueError,
                          "A null character was found in the text");
         }
+    }
+
+    if (PyBytes_Check(text)) {
```
ankith26, line 419:
```suggestion
    else if (PyBytes_Check(text)) {
```
We don't need to bother with a bytes check if `text` was already unicode

ankith26 (approved):
Nice cleanups, the code is sure simpler now! 🥳 🎈 

Left a minor nitpick review but approved nonetheless, resolve at will

src_c/font.c
```diff
@@ -371,15 +371,14 @@ static PyObject *
...
-    PyObject *fg_rgba_obj, *bg_rgba_obj = NULL;
+    PyObject *fg_rgba_obj, *bg_rgba_obj = Py_None;
     Uint8 rgba[] = {0, 0, 0, 0};
     SDL_Surface *surf;
-    SDL_Color foreg, backg;
-    int just_return;
+    const char *astring = "";
 
-    if (!PyArg_ParseTuple(args, "OpO|O", &text, &aa, &fg_rgba_obj,
+    if (!PyArg_ParseTuple(args, "OpO|O", &text, &antialias, &fg_rgba_obj,
```
MyreMylar, line 381: Could take this opportunity to add keyword arguments if you fancy?

Or just leave it for another PR.

MyreMylar (approved):
LGTM 👍  tested it locally with local test run and rendering text with a variety of options in a test program.

src_c/font.c
```diff
@@ -388,121 +387,83 @@ font_render(PyObject *self, PyObject *args)
...
-    foreg.b = rgba[2];
-    foreg.a = SDL_ALPHA_OPAQUE;
-    if (bg_rgba_obj == Py_None) {
-        /* Explicit None is the same as not passing a color for us */
-        bg_rgba_obj = NULL;
-    }
-    if (bg_rgba_obj != NULL) {
+
+    SDL_Color foreg = {rgba[0], rgba[1], rgba[2], SDL_ALPHA_OPAQUE};
+    /* might be overriden right below, with an explicit background color */
```
novialriptide, line 392: "overriden" should be "overridden"

novialriptide (changes requested):
I haven't tested it but I just found a spelling mistake.

src_c/font.c
Starbuck5: Thanks.

src_c/font.c
Starbuck5: I'll leave it for another PR. I did an experiment with METH_FASTCALL on this, and got an additional 4% speedup for rendering text. But I'm not trying to merge that because FASTCALL is 3.7 and up, and there isn't a good way to parse keywords with it.

PR #3187, opened 2022-05-21
Make latest pypy wheels on mac/linux, cibuildwheel on manylinux and other minor updates

ankith26:
This PR updates our manylinux build to use cibuildwheel ~~(and also splits our manylinux arm builds across multiple parallel workflows because it's really slow, I'm trying to look into other CI alternatives that provide native linux arm)~~ and adds CircleCI to run our manylinux aarch64 builds.

this PR also makes our ci config generate pypy 3.8 and pypy 3.9 wheels in addition to pypy 3.7 wheels that were previously generated.

Updates some actions to their respective latest versions, and some other cleanups/comment updates


Starbuck5 (approved):
It all looks good to me.

I wonder if there is a speedup to be observed with adding the -j3.

Thanks for working to improve the CI Ankith!

ankith26:
> I wonder if there is a speedup to be observed with adding the -j3.

There is, on the mac builds. For instance our arm64 build went from taking anywhere between 15-20 minutes, to now 12 minutes on this PR. Our linux builds were already using `-j4` before this PR, and using `-j3` on this PR does not create much of a difference since linux runners in github actions have 2 cores (github actions macs have 3)

illume:
circleci is running in here now

illume:
I mentioned in the other PR that Azure arm64 is available as preview now for a couple of months, and github actions has been waiting for them to release.

ankith26:
Oh, nice!
Once that is done we could revert back to gh actions, but there seems to be no ETA on it so IG it might take a while.

PR #3188, opened 2022-05-21
Use CircleCI for manylinux aarch64 builds

ankith26:
aarch64 builds on github actions are really slow. Out of curiousity of wanting to know how much time the builds would take if it ran on native aarch64 CI hardware, I setup CircleCI on my fork. I did a bit of reading on it, and it was not too hard to setup.

ATM... our aarch64 builds take about 1.5 hours for doing only cpython versions which is not exactly ideal. In #3187 (the PR that this PR is based on on) I'm trying to add pypy builds and it quickly became clear that adding this would significantly worsen the time taken by the aarch64 builds... So in that PR, I took the approach of further splitting aarch64 builds into 4 parallel builds (this too is not ideal because this means aarch64 docker base images will be pulled 4 times) and now each of those takes anywhere between 35 to 55 minutes, if they ran sequentially it would take around 3 hours for each aarch64 build.

CircleCI takes 20 minutes when both cpython and pypy builds are run sequentially, putting this at par with our manylinux2014 x86_64 github actions CI that takes around the same time, which is a large speedup.

CircleCI however, does not have unlimited builds under its free plan; it has 30,000 "credits" per month. Since I intend to only use one CircleCI workflow (for aarch64 linux), we will get 3,000 build minutes per month (10 "credits" are used per minute spend building). This translates to about 150 builds per month which is well enough for now.

This is intentionally a separate PR because it can be dropped if anyone has serious objections to using CircleCI, but IMHO it's worth adding.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,42 @@
+# Use the latest 2.1 version of CircleCI pipeline process engine.
+# See: https://circleci.com/docs/2.0/configuration-reference
+version: 2.1
+
+# Define a job to be invoked later in a workflow.
+# See: https://circleci.com/docs/2.0/configuration-reference/#jobs
+jobs:
+  linux-arm-wheels:
+    working_directory: ~/linux-wheels
+    machine:
+      image: ubuntu-2004:current
+    resource_class: arm.medium
+
+    environment:
+      # these environment variables will be passed to the docker container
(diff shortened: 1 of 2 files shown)


illume:
Thanks.

I wouldn't mind having the circleci folder in there until github actions support arm64. Not sure how long it will take for github to use arm64, but... Azure now has arm in public preview at least and Visual Studio arm64 was just announced. https://azure.microsoft.com/en-us/updates/public-preview-arm64based-azure-vms-can-deliver-up-to-50-better-priceperformance/

ps. I don't think it is possible to move the circleci folder out of the root.

illume (approved):
👍 🎈

Issue #3175, opened 2022-05-11
More global key information


Starbuck5:
**Description**

I'd like to add the functions `pygame.key.get_just_pressed()` and `pygame.key.get_just_released()` to make it possible for entities to get more information about keyboard state without being tied to the event loop.

When I say entities, I'm assuming there's a list of entities that make up the sprites and "things" of the game. This is a very common pattern.

Imagine you're making an Asteroids game. You want "press held" information for turning and for firing the thruster. However, you want "discrete keypress" for the shooting behavior. You can get the "press held" info "globally" using `pygame.key.get_pressed()`. For "discrete keypress" you can of course turn to KEYDOWN events. 

But how do you pipe your KEYDOWNs to your ship entity? Do you pipe every event to every entity in the event loop, even though the asteroids will never need any events? Do you special case the ship and lose a bit of elegance? This is by no means unsolvable, you could even do falling edge detection on the data from get_pressed. This API just aims to simplify it, give a convenient way to work with key input while in a situation where it's fine / preferable to be decoupled from the event loop.

**Implementation**
Ankith was grilling me about this on discord.

Every time the event system is pumped, it can clear out some global state "dict". And then when it sees KEYUPs or KEYDOWNs it adds them to the relevant "dict." When the user calls `get_just_pressed()` or `get_just_released()` it shows them the data gathered in the last "sample" taken, during the last call to pygame.event.get().

*not a dict, but whatever

Starbuck5:
Tiny example to show how I imagine the API looking.

```py
if pygame.key.get_just_pressed()[pygame.K_a]:
  print("A key was just pressed")
```

python-ast-person:
My  proposal(#3479) would allow you to handle this by defining `Sprite.handle_event()`:
```
class PlayerShip(pygame.sprite.Sprite):
    ...
    listening_events = {pygame.locals.KEYDOWN,}
    def handle_event(self,event):
        if event.type == pygame.locals.KEYDOWN:
            #insert what ever you want to do
        else:
            super().handle_event(event)
    ...
```

FeldrinH:
What became of this feature request? It has been closed, but I don't see `get_just_pressed` or `get_just_released` in the latest Pygame.

PR #3190, opened 2022-05-22
Added get_just_pressed() and get_just_released() functions to satisfy global key information

ScriptLineStudios:
As suggested in https://github.com/pygame/pygame/issues/3175 it would be useful to have access to some global information about the event loop, specifically key pressed and releases. ```get_just_pressed()``` when the event loop gets pumped I check if an ```SDL_KEYDOWN``` or ```SDL_KEYUP``` event type occurs. And if some store the ```eventbuf[loop].key.keysym.sym``` to be returned by the respective function. Small example of the functions in action: 

```python
import pygame
pygame.init()

display = pygame.display.set_mode((600, 600))

while True:
	display.fill((0,0,0))
	for event in pygame.event.get(): #Pump The event loop
		pass

	if pygame.event.get_just_pressed() == pygame.K_a: #Check if a key was just pressed, returns -1 if no key was pressed
		print("A key was just pressed!")
	
	if pygame.event.get_just_released() == pygame.K_a: #Check if a key was just released, returns -1 if no key was released
		print("A key was just released!")
			
	pygame.display.update()

```

### DIFF
--- a/src_c/doc/event_doc.h
+++ b/src_c/doc/event_doc.h
@@ -17,6 +17,9 @@
 #define DOC_PYGAMEEVENTEVENT "Event(type, dict) -> Event\nEvent(type, **attributes) -> Event\npygame object for representing events"
 #define DOC_EVENTTYPE "type -> int\nevent type identifier."
 #define DOC_EVENTDICT "__dict__ -> dict\nevent attribute dictionary"
+#define DOC_PYGAMEEVENTGETJUSTPRESSED "get_just_pressed() -> int\n get the id of the key that was just pressed"
+#define DOC_PYGAMEEVENTGETJUSTRELEASED "get_just_released() -> int\n get the id of the key that was just released"
+
 
 
 /* Docs in a comment... slightly easier to read. */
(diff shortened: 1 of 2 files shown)


MyreMylar:
looks like this PR needs some `python setup.py format` action. It's easy to setup on linux and mac but a little trickier on windows.

Starbuck5 (changes requested):
Thank you for working on this. I didn't think anyone was going to.

However, I don't think this functionality will be very useful if it only stores the very last key pressed in the last run through the event loop. That's why in my example code in the issue I had it return (like pygame.key.get_pressed) a "ScancodeWrapper" object containing just the keys from KEYDOWNs in the last run through the event loop, or just the KEYUPs in the last run through the event loop.

Scancode wrapper stuff: https://github.com/pygame/pygame/blob/main/src_c/key.c#L62-L111
Seems to be a subclass of tuple.

My intended implementation would necessitate communicating between event.c and key.c to make this work. You can do this by adding to the "C api" exported by the event module. See https://www.pygame.org/docs/c_api/event.html.

ScriptLineStudios:
@Starbuck5 
Ah I understand now, sorry for the misunderstanding, I will look into this soon. Also, could you provide some guidance on creating documentation for these functions? I don't know if i'm missing something but i'm not sure how to go about setting up a ```DOC_PYGAMEEVENTGETJUSTRELEASED``` for example. 

Thanks!

Starbuck5:
@ScriptLineStudios All of event_doc.h is autogenerated, not just the DOC_ constants. They are autogenerated when you add the functions to the documentation and then run `setup.py docs`

The documentation files are here: https://github.com/pygame/pygame/tree/main/docs/reST/ref

Issue #3096, opened 2022-03-27
C code handling of input booleans


Starbuck5:
I've just been looking the source code at how we handle receiving booleans from python
- Some code parses them as ints, using i in the PyArg_Parse
- Some code parses them as "boolean predicates", using p
- Some code parses them as Python objects and then uses PyObject_IsTrue

The problem with PyObject_IsTrue is that it's a truthiness check, not a True check. These pygame functions accept non empty strings as True, or arbitrary objects without a `__bool__` method.
The problem with parsing as ints is that the error message would be weird if someone put in the wrong object, theoretically.
I think they should all be using p, but I haven't tested the above claims much yet

ankith26:
From the [docs](https://docs.python.org/3/c-api/arg.html)
> p ([bool](https://docs.python.org/3/library/functions.html#bool)) [int]
Tests the value passed in for truth (a boolean predicate) and converts the result to its equivalent C true/false integer value. Sets the int to 1 if the expression was true and 0 if it was false. This accepts any valid Python value. See [Truth Value Testing](https://docs.python.org/3/library/stdtypes.html#truth) for more information about how Python tests values for truth.
New in version 3.3.

So even `p` is a truthiness check, so code using `p` is pretty much the same as using `PyObject_IsTrue`, except the latter can be a bit more verbose

Starbuck5:
> So even p is a truthiness check, so code using p is pretty much the same as using PyObject_IsTrue, except the latter can be a bit more verbose

I see. It's strange to me that there isn't just a "bool" setting. We could check against `Py_True` and `Py_False` themselves, but then `0` and `1` wouldn't work.

I wrote a new strategy for handling those pesky PyObjects:
```c
    PyObject* bool_obj;

    if (!PyArg_ParseTuple(args, "O", &bool_obj))
        return NULL;

    if (!PyLong_Check(bool_obj)) {
        //TODO: error handling
        PyObject* class = PyObject_GetAttrString(bool_obj, "__class__");
        PyObject* classname = PyObject_GetAttrString(class, "__name__");
        
        PyErr_Format(PyExc_TypeError, "A bool is required (got type %S)", classname);
        return NULL;
    }

    int val = PyLong_AsLong(bool_obj);
```

This does what I expect in parsing a boolean value.

Do you think it would be worth doing something like this? We could put that up as a macro in pygame.h or somewhere and get all the functions using this.

robertpfeiffer:
[...]

PR #3192, opened 2022-05-22
Simplify some boolean handling

Starbuck5:
This doesn't change the status quo with how booleans are handled at all, it just simplifies the code, using the "p" flag of PyArg_Parse, rather than PyObject_IsTrue.

This is for #3096. Depends how people want to proceed whether this closes it.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -247,7 +247,7 @@ static PyObject *
 aalines(PyObject *self, PyObject *arg, PyObject *kwargs)
 {
     pgSurfaceObject *surfobj;
-    PyObject *colorobj, *closedobj;
+    PyObject *colorobj;
     PyObject *points, *item = NULL;
     SDL_Surface *surf = NULL;
     Uint32 color;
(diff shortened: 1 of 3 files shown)


ankith26 (approved):
LGTM! thanks 😎 

I'm in favour of using `p` formatting for accepting `bool`s everywhere for consistency, even though its just a truthiness check.

In python code, the most common way to handle bools is usually `if my_bool_flag: ...` and `if not my_bool_flag: ...`, and you would almost never see a `if my_bool_flag is True: ...`, `if my_bool_flag == False` or something of that sort (infact these are considered bad practices). So common python code is actually doing the equivalent of `p` in a sense, truthiness checks are more pythonic than hard true/false checks.

Starbuck5:
Thanks for the approval!

I was hoping to keep the debate on that on the other issue (since my PR isn't shaking that up at all).

One thing though is with this treatment of booleans, the type hinting will either be unhelpful or wrong. Unhelpful if it's typed as Any, since you should use a bool. And wrong if it's not typed as Any.

zoldalma999 (approved):
Looks good to me! 🎉

Issue #3144, opened 2022-04-17
font.render segfault in pygame 2.1.3dev running under windows with a specific font and size


ankith26:
This issue was originally identified by OP of #3128 who reported this segfault when debugging with a 2.1.3dev5 wheel from a development branch, and I tracked down this issue to happen on `main` too. This is a regression from 2.1.2, for which the test code given below works fine. This issue only seems to happen on Windows and not Ubuntu, which also leads me to believe that this is a result of the SDL_ttf 2.0.18 upgrade we did recently with windows buildconfig.

```py
import pygame

pygame.init()

for size in (60, 40, 10, 20, 70, 45, 50, 10):
    font = pygame.font.Font("PlayfairDisplaySemibold.ttf", size)
    font.render("WHERE", True, "black")
    print(f"rendered with size {size}!")

pygame.quit()
```

Gives output
```
pygame 2.1.3.dev5 (SDL 2.0.20, Python 3.10.1)
Hello from the pygame community. https://www.pygame.org/contribute.html
rendered with size 60!
rendered with size 40!
rendered with size 10!
rendered with size 20!
Windows fatal exception: code 0xc0000028

Current thread 0x00000148 (most recent call first):
  File "E:\tests\test.py", line 7 in <module>
```
Trying some other input combinations shows that some font sizes segfault consistently while others don't.

Attaching font file as a zip below because the issue seems to be specific to the file (file provided by OP of issue linked above)
[PlayfairDisplaySemibold.zip](https://github.com/pygame/pygame/files/8501976/PlayfairDisplaySemibold.zip)

Zireael07:
For those curious, the font seems to be indeed free and has been available since 2012 or thereabouts (so unlikely to be something bad in the font itself, but rather in SDL or some other font display code)

Ref: https://www.whatfontis.com/Playfair-Display-SemiBold.font
https://www.fontspace.com/claus-eggers-sorensen

Starbuck5:
I can reproduce on Windows, but I can't on Mac no matter what I try.

I compiled SDL_ttf at the 2.0.18 release tag, built pygame with that, and it worked fine.

I wrote a c test program, it works fine on Mac
```c
#include "SDL2/SDL.h"
#include "SDL2/SDL_ttf.h"
#include <stdio.h>

// gcc font_segfault.c -lsdl2 -lsdl2_ttf

int main(int argc, char *argv[])
{
    SDL_SetMainReady();
[...]

PR #3193, opened 2022-05-23
Use custom SDL2_ttf on Windows

Starbuck5:
Potential solution for #3144 

I have built my own DLLs of SDL_ttf and uploaded them to Github for easy download. This eliminates the segfaulting seen in #3144.

Obviously we don't usually use versions of our dependencies between releases, but I say we should go for it this time. SDL_ttf 2.0.15 is very dated at this point, and it would be a pain to roll back the buildconfig changes I needed to make to upgrade to SDL_ttf 2.0.18.


ankith26 (approved):
Generally I'd be against using dev releases of our deps in a release, but I think it's fine for this time. Considering the bug this is fixing has a pretty large impact, this PR should definitely go by this release, or we should revert to 2.0.15 if any other reviewer does not like this PR (which might be tedious but oh well)

ankith26:
On the topic of this PR... might be worth making this a unit test to test the fix for now, and also make sure it does not regress in the upcoming SDL stable versions

illume:
I wonder how simple it is to build SDL_ttf on CI?

I'm a little uneasy releasing with new custom built libraries, and am thinking we maybe rollback until we automate the building of the custom SDL_ttf etc.

Starbuck5:
I understand the concern about custom libraries.

It took me a long time to do this manually in Visual Studio, and there was no command line involved, so I'm not sure where to even start trying to automate this.

I don't want to rollback because of all the cool features I'd like to use from SDL_ttf 2.0.18 (harfbuzz font support for much better support for non-latin languages, support for setting font direction, a multiline text rendering API). Also much better emoji support, including full color emoji.

The SDL_ttf team fixed the issue on their end, it was a compiler bug in an older version of Mingw. (Yes, even their VS DLLs are made with Mingw apparently).

Anyways, I'd like to use this for the next pygame dev release, and we can see whether any problems pop up for our beta testers. I haven't had any problems locally with this, but it's always get to testing on lots of systems.

illume:
Ah ok. I was afraid it would be complex.

Seems SDL folks are working on a new SDL_image release, and so I guess also a SDL_ttf will come in time too. Not sure if we should wait for those before a real release though(could be months). Yeah, ok... I guess a dev release can happen with this one.

illume:
Would it be possible to add the test case, so we can make sure we're not regressing when the new SDL_ttf comes out?

Starbuck5:
Done. Had to force push because I forgot to format.

Starbuck5:
Ok, not the formatting one is failing on a piece of code I didn't touch at all. Did clang format change?

illume (approved):
Thanks for supplying the build, and also for adding the unit test too.

🎉🎈

Issue #2647, opened 2021-07-21
failed loading libmpg123-0.dll also after reinstall


raghav3148:
pycham error : failed loading libmpg123-0.dll

MightyJosip:
Not enough information, closing it now

jacqobi:
Hi @MightyJosip,
I am able to reproduce the issue without using PyCharm as well, found the root cause, possible solution and interim workaround for @raghav3148 and other affected users until fix is released.

The full error message, that i believe is the same also when using PyCharm, is:
> pygame.error: Failed loading libmpg123-0.dll: The specified module could not be found.
### Cause
Related to [bpo-36085](https://bugs.python.org/issue36085) on how the DLL loading is handled by Python interpreter on Windows.
As noted in the [change logs](https://docs.python.org/3/whatsnew/changelog.html#id213) but better described in [Python's 3.8 What's New](https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew)

> DLL dependencies for extension modules and DLLs loaded with ctypes on Windows are now resolved more securely. Only the system paths, the directory containing the DLL or PYD file, and directories added with add_dll_directory() are searched for load-time dependencies.
Specifically, PATH and the current working directory are no longer used, and modifications to these will no longer have any effect on normal DLL resolution.
**_If your application relies on these mechanisms, you should check for add_dll_directory() and if it exists, use it to add your DLLs directory while loading your library._**
Note that Windows 7 users will need to ensure that Windows Update KB2533623 has been installed (this is also verified by the installer). (Contributed by Steve Dower in bpo-36085.)

### Proposed Solution
Adding the dll directory path to the __init__.py of pygame using the [os.add_dll_directory()](https://docs.python.org/3/library/os.html#os.add_dll_directory) under the Windows specific clause as documented below should do the trick.
```
...
...
# Choose Windows display driver
if os.name == 'nt':
    #pypy does not find the dlls, so we add package folder to PATH.
    pygame_dir = os.path.split(__file__)[0]
    os.environ['PATH'] = os.environ['PATH'] + ';' + pygame_dir
    #Fix for the bpo-36085 change in Python3.8 on Windows
    os.add_dll_directory(pygame_dir)
...
...
```

### Workarounds
1. add the lines below to your code when using pygame.
```
import os
os.add_dll_directory(<full path to pygame directory>)
```
[...]

PR #3196, opened 2022-05-25
Fix DLL loading problems on Windows Store Python installations

Starbuck5:
We've all been there as pygame contributors. You look at reddit or discord or stackoverflow, and somebody comes in and says "pygame doesn't work: `pygame.error: DLL load failed libmpg123`"

Where do you even start with that? It works on my machine^TM. It seems to work on the vast majority of machines, but it just doesn't work for some people.

Thanks to @EcoFreshKase for helping track this down on discord. The difference is that some people are using Windows Store Python. We've known Windows Store Python is flaky for a while, but I've never put two and two together with this issue.

EcoFreshKase confirms that this fix solves the DLL loading on their machine, with Windows Store Python. They were having the issue before.

Also thanks to @jacqobi for proposing a version of this fix a while ago: https://github.com/pygame/pygame/issues/2647#issuecomment-957864812

### DIFF
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -28,8 +28,14 @@ import os
 
 # Choose Windows display driver
 if os.name == "nt":
+    pygame_dir = os.path.split(__file__)[0]
+
     # pypy does not find the dlls, so we add package folder to PATH.
-    os.environ["PATH"] = os.environ["PATH"] + ";" + os.path.split(__file__)[0]
+    os.environ["PATH"] = os.environ["PATH"] + ";" + pygame_dir
+
+    # windows store python does not find the dlls, so we run this
+    if sys.version_info > (3, 8):
+        os.add_dll_directory(pygame_dir)  # only available in 3.8+
 
 # when running under X11, always set the SDL window WM_CLASS to make the
 #   window managers correctly match the pygame window.


src_py/__init__.py
```diff
@@ -28,8 +28,14 @@
...
 # Choose Windows display driver
 if os.name == "nt":
+    pygame_dir = os.path.split(__file__)[0]
+
     # pypy does not find the dlls, so we add package folder to PATH.
-    os.environ["PATH"] = os.environ["PATH"] + ";" + os.path.split(__file__)[0]
+    os.environ["PATH"] = os.environ["PATH"] + ";" + pygame_dir
+
+    # windows store python does not find the dlls, so we run this
+    if sys.version_info > (3, 8):
```
ankith26, line 37: this should be `>=`

Starbuck5: It will never be equal. version_info is a 5 element named tuple.

```py
>>> a = (3, 8, 0, 'final', 0)
>>> a > (3, 8)
True
```

ankith26: Oh, yeah that makes sense. For some reason I thought that `version_info` would be a custom tuple subclass that would handle comparisons differently, but that's not the case

ankith26 (approved):
Good to see this long term sporadic issue getting an actual fix, thanks! 🍰 🥳

illume (approved):
👍 🎈 thanks

PR #3185, opened 2022-05-18
Update build-emsdk.yml

pmp-p:
typo missing $ !

PR #3197, opened 2022-05-26
Wrong path when creating symlink in the workflow file for pygame-wasm

tusqasi:
There seems to be a missing `$` when creating the symlink


ankith26:
Hello!
First of all, thank you for taking time and considering contributing to pygame 🎉

This PR is doing the same fix as #3185 which was opened over a week ago, so this PR has to be closed unfortunately.

PR #3207, opened 2022-06-04
Remove more code for legacy Python

cclauss:
Use `flake8 --select=F821` to find and fix all `undefined names` in the codebase.
The majority of these are about builtins that were removed or moved in the transition from Python 2 to Python 3.

The failing test seems unrelated.

### DIFF
--- a/buildconfig/config_emsdk.py
+++ b/buildconfig/config_emsdk.py
@@ -5,8 +5,6 @@ import os
 import sys
 from glob import glob
 
-from distutils.sysconfig import get_python_inc
-
 from .config_unix import get_porttime_dep
 
 configcommand = os.environ.get('SDL_CONFIG', 'sdl-config',)
@@ -19,11 +17,9 @@ else:
 
 EMSDK = os.environ.get('EMSDK', None)
 is_wasm = EMSDK is not None
-
 if not is_wasm:
     print("EMSDK not found")
     raise SystemExit(1)
-
 
 # EMCC_CFLAGS="-s USE_SDL=2 is required to prevent '-iwithsysroot/include/SDL'
 # which is SDL1 from ./emscripten/tools/ports/__init__.py
@@ -34,7 +30,6 @@ os.environ["EMCC_CFLAGS"]=EMCC_CFLAGS.strip()
 
 CC=os.environ.get("CC","emcc")
 os.environ["CC"] = CC.strip()
-
 
 
 class DependencyProg:
@@ -79,19 +74,18 @@ class DependencyProg:
                 inc = f"-I{EMSDK}/upstream/emscripten/cache/sysroot/include/freetype2/freetype "
                 self.cflags = inc + ' ' + self.cflags
 
-
         except (ValueError, TypeError):
-            print ('WARNING: "%s" failed!' % command)
+            print('WARNING: "%s" failed!' % command)
             self.found = 0
             self.ver = '0'
             self.libs = defaultlibs
 
     def configure(self, incdirs, libdir):
         if self.found:
-            print (self.name + '        '[len(self.name):] + ': found ' + self.ver)
+            print(self.name + '        '[len(self.name):] + ': found ' + self.ver)
             self.found = 1
         else:
-            print (self.name + '        '[len(self.name):] + ': not found')
+            print(self.name + '        '[len(self.name):] + ': not found')
 
 class Dependency:
     def __init__(self, name, checkhead, checklib, libs):
@@ -121,15 +115,15 @@ class Dependency:
                     self.lib_dir = dir
 
         if (incname and self.lib_dir and self.inc_dir) or (not incname and self.lib_dir):
-            print (self.name + '        '[len(self.name):] + ': found')
+            print(self.name + '        '[len(self.name):] + ': found')
             self.found = 1
         else:
 
             if self.name in ["FONT","IMAGE","MIXER","PNG","JPEG","FREETYPE"]:
                 self.found = 1
-                print (self.name + '        '[len(self.name):] + ': FORCED (via emsdk builtins)')
+                print(self.name + '        '[len(self.name):] + ': FORCED (via emsdk builtins)')
                 return
-            print (self.name + '        '[len(self.name):] + ': not found')
+            print(self.name + '        '[len(self.name):] + ': not found')
             print(self.name, self.checkhead, self.checklib, incdirs, libdirs)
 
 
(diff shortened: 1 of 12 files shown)


buildconfig/config_emsdk.py
```diff
@@ -1,18 +1,18 @@
...
-from glob import glob
-import platform
 import logging
+import os
+import sys
+from glob import glob
+
 from distutils.sysconfig import get_python_inc
 
+from .config_unix import get_porttime_dep
```
ankith26, line 10: why is this extra import here?

buildconfig/config_msys2.py
```diff
@@ -16,6 +16,11 @@
 from glob import glob
 from distutils.sysconfig import get_python_inc
 
+try:
+    BuildError
+except NameError:
+    BuildError = TypeError
```
ankith26, line 22: what is this for? Why is this here and in some other files?

buildconfig/stubs/pygame/.flake8
```diff
@@ -16,10 +16,11 @@
 #       W504 line break after binary operator
 
 [flake8]
-ignore = F401, F403, F405, F811, E301, E302, E305, E501, E701, E704, E741, B303, W504
+ignore = B303, E301, E302, E305, E501, E701, E704, E741, F401, F403, F405, F811, W504
 # We are checking with Python 3 but many of the stubs are Python 2 stubs.
 # A nice future improvement would be to provide separate .flake8
 # configurations for Python 2 and Python 3 files.
```
ankith26, line 22: these comments are rather outdated, this PR would be a good opportunity to clean up these :smile:

ankith26 (commented):
Nice cleanups, thanks a ton!

Left a few reviews to address before approval

buildconfig/config_emsdk.py
```diff
@@ -1,18 +1,18 @@
...
+
 configcommand = os.environ.get('SDL_CONFIG', 'sdl-config',)
 configcommand = configcommand + ' --version --cflags --libs'
 localbase = os.environ.get('LOCALBASE', '')
-if os.environ.get('PYGAME_EXTRA_BASE', ''):
-    extrabases = os.environ['PYGAME_EXTRA_BASE'].split(':')
-else:
-    extrabases = []
+extrabases = os.environ.get('PYGAME_EXTRA_BASE', '').split(':')
```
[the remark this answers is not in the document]
cclauss:
```suggestion
if os.environ.get('PYGAME_EXTRA_BASE'):
    extrabases = os.environ['PYGAME_EXTRA_BASE'].split(':')
else:
    extrabases = []
```

buildconfig/config_msys2.py
cclauss: It is an _undefined name_ which has the potential to raise `NameError` at runtime so let's avoid that.

buildconfig/stubs/pygame/.flake8
cclauss, line 22:
```suggestion
```

buildconfig/stubs/pygame/.flake8
```diff
@@ -16,10 +16,11 @@
...
-builtins = StandardError,apply,basestring,buffer,cmp,coerce,execfile,file,intern,long,raw_input,reduce,reload,unichr,unicode,xrange
-exclude = .venv*,@*,.git
+# builtins = StandardError,apply,basestring,buffer,cmp,coerce,execfile,file,intern,long,raw_input,reduce,reload,unichr,unicode,xrange
```
cclauss, line 23:
```suggestion
```

buildconfig/config_emsdk.py
cclauss: It is an _undefined name_ that can potentially raise `NameError` at runtime so let's avoid that.

ankith26: That condition that calls `get_porttime_dep` in this file does not actually execute at runtime... I think that codepath should be removed instead of adding this import

buildconfig/config_msys2.py
ankith26: Again... the codepath that actually calls `BuildError` is super unlikely to run, and even if it does, I think it's better if it errored with `RuntimeError` directly in code instead of dealing with `BuildError` aliases which is not even present elsewhere in the code

cclauss:
https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead, these tests are focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python, they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python, a __NameError__ is raised which will halt/crash the script on the user.

buildconfig/config_emsdk.py
cclauss: I feel unqualified to make this change.

buildconfig/config_msys2.py
ankith26: Nice, I see you fixed this. But you are raising `TypeError` there which seems to be the wrong error to raise here, ideally it should be a `ValueError` (because the value should have been 32 or 64) or a `RuntimeError`

buildconfig/config_emsdk.py
ankith26: There is a `if not is_wasm:` block guarding the `get_porttime_dep` function call... but at the top of the file there is also a systemexit happening for the same condition, so this can be removed safely.

If you wish, I could make a commit for it in this branch, or you could do it yourself too :)

ankith26: You removed that unused block correctly, but the import is still here so this needs to be removed too

buildconfig/config_emsdk.py
```diff
@@ -1,27 +1,26 @@
...
+import sys
+from glob import glob
+
+from .config_unix import get_porttime_dep
 
 configcommand = os.environ.get('SDL_CONFIG', 'sdl-config',)
 configcommand = configcommand + ' --version --cflags --libs'
 localbase = os.environ.get('LOCALBASE', '')
-if os.environ.get('PYGAME_EXTRA_BASE', ''):
+if os.environ.get('PYGAME_EXTRA_BASE'):
```
ankith26, line 13: minor nitpick
```suggestion
if "PYGAME_EXTRA_BASE" in os.environ:
```

ankith26 (approved):
LGTM! 🎉 
Thanks again for the PR

dr0id (approved):

illume (approved):
👍 🎉 thanks

PR #3210, opened 2022-06-05
Prettify joystick documentation + add Nintendo Switch Controller docs

novialriptide:
This PR does a couple things.
 - Changes the format of the joystick documentation of the PlayStation 4 and Xbox 360.
 - Adds Nintendo Switch Pro Controller documentation
 - Adds Nintendo Switch Left Joy-Con documentation
 - Adds Nintendo Switch Right Joy-Con documentation
 - Replace the old joystick example screenshot with a new one (the original was captured on a Windows 7 or 8 window)
 - Merges the 2 joystick code examples into 1


ankith26 (approved):
I can't exactly verify the info here since I don't have any joystick hardware on me ATM, but I will approve nonetheless.

Looking at the new screenshot you posted, it's interesting to see that some bools went to being ints. Is this an intended change or is this a regression? (in the latter case it might need followup in a future PR)
Can someone who has better joystick experience than me clear things up here?

Thanks for the PR novial! 🎉 🍰

docs/reST/ref/joystick.rst
```diff
@@ -364,20 +364,65 @@ So call one of pygame.event.get, pygame.event.wait, or pygame.event.pump regular
...
+    Up -> Down      - Axis 1
+
+* **Right Stick**::
+
+    Left -> Right   - Axis 2
+    Up -> Down      - Axis 3
+
+* **Left Trigger**::
+
+    Out -> In       - Axis 2
```
ankith26, line 395: is this supposed to be 4?

docs/reST/ref/joystick.rst
```diff
@@ -500,9 +545,11 @@ In pygame 1 the X360 controller mapping has 5 Axes, 10 buttons and 1 hat.
...
 
-**Playstation 4 Controller (name: "Wireless Controller")**
+Playstation 4 Controller (pygame 1.x)
+*************************************
 
-In pygame 1 the PS4 controller mapping has 6 Axes and 14 buttons and 1 hat.
+The PlayStation 4 controller mapping has 6 axes, 14 buttons, and 1 hat.
+name: "Wireless Controller"
 
 * **Left Stick**::
```
ankith26, line 554: I don't think pygame 1.x info is particularly relevant here, this might need updating (perhaps in a future PR)

docs/reST/ref/joystick.rst
novialriptide: Whoops, I just fixed it.

docs/reST/ref/joystick.rst
illume: It's relevant to users of the library who want to support pygame 1.9.x and 2.x. No?

illume:
Probably the image could be cropped like the old one to remove all that empty space.
Is there a way to increase the font size to match the old image? The new image font size is a isn't readable.

novialriptide:
> Probably the image could be cropped like the old one to remove all that empty space. Is there a way to increase the font size to match the old image? The new image font size is a isn't readable.

Yea, I'll increase the font size and crop the image.

novialriptide:
I'm not sure why appveyor is failing..

docs/reST/ref/joystick.rst
ankith26: yeah, true.

Looks like I was having a brain fart when I reviewed this 😅

Starbuck5:
The joystick example code is in "examples/" and "docs/reST/ref/code_examples" and they are starting to diverge. It would be great if you could get the docs to reference the one in examples.

novialriptide:
> The joystick example code is in "examples/" and "docs/reST/ref/code_examples" and they are starting to diverge. It would be great if you could get the docs to reference the one in examples.

On it.

>Mega on discord verified your axis / button readouts?

He has provided button mappings for the Nintendo Switch Joy Cons, not the Pro Controller. I'll be adding them to this PR shortly

MyreMylar (changes requested):
Looks like the changes to variable names to camel case don't meet python's PEP8 so I'm not sure why some of those changed.

I know we don't lint the examples yet but I expect that camel case variable names won't pass muster with one of the tools. 

Otherwise it seems fine.

Starbuck5 (changes requested):
> I don't think it's possible to do this, as it wouldn't be possible to reference examples/joystick.py from docs/reST/ref. I pushed a new commit updating examples/joystick.py.

examples/joystick.py had meaningful changes in it as well, I don't want to erase those. They need to be "merged" together manually.

novialriptide:
> > I don't think it's possible to do this, as it wouldn't be possible to reference examples/joystick.py from docs/reST/ref. I pushed a new commit updating examples/joystick.py.
> 
> examples/joystick.py had meaningful changes in it as well, I don't want to erase those. They need to be "merged" together manually.

I see, I'll get to work on it maybe rewriting the example all together as well.

ankith26 (changes requested):
This PR certainly diverged a lot since my first approval😄 

The current `examples/joystick.py` code was added by zoldalma in #2741 (as an update on the older joystick example code). This also added stuff like rumble and hotplugging support, which we definitely need to retain.

I would prefer if the example was still in `examples` directory along with the rest. The docs can just have a simple link to the example's github page.

There may not be a need to completely rewrite it, IMHO. Also it may be out of scope of this PR so you can do any polish to it if you want to, in another PR.

novialriptide:
> I don't think it's possible to do this, as it wouldn't be possible to reference `examples/joystick.py` from `docs/reST/ref`. I pushed a new commit updating `examples/joystick.py`.

I am wrong! I got this working thanks to @Starbuck5. I will be adding joystick documentation for the Nintendo Switch Joy-Cons by @Mega-JC shortly.

MyreMylar (approved):

examples/joystick.py
```diff
@@ -32,11 +36,16 @@ def main():
...
+    # Loop until the user clicks the close button.
+    # Used to manage how fast the screen updates.
     clock = pygame.time.Clock()
 
     # Get ready to print.
     text_print = TextPrint()
 
+    # Initialize the joysticks.
+    pygame.joystick.init()
+
```
Starbuck5, line 48: pygame.init() is called above.

examples/joystick.py
```diff
@@ -76,24 +85,39 @@ def main():
...
+        # Get count of joysticks.
+        joystick_count = pygame.joystick.get_count()
+
+        text_print.tprint(screen, f"Number of joysticks: {joystick_count}")
         text_print.indent()
 
         # For each joystick:
         for joystick in joysticks.values():
             jid = joystick.get_instance_id()
+            joystick.init()
```
Starbuck5, line 100: docs mark this as deprecated

examples/joystick.py
```diff
@@ -76,24 +85,39 @@ def main():
...
+
+            try:
+                jid = joystick.get_instance_id()
+            except AttributeError:
+                # get_instance_id() is an SDL2 method
+                jid = joystick.get_id()
```
Starbuck5, line 106: This can be written to only work on pygame 2. No need for the try except then?

examples/joystick.py
```diff
@@ -76,24 +85,39 @@ def main():
...
             name = joystick.get_name()
             text_print.tprint(screen, f"Joystick name: {name}")
 
-            guid = joystick.get_guid()
-            text_print.tprint(screen, f"GUID: {guid}")
+            try:
+                guid = joystick.get_guid()
+            except AttributeError:
+                # get_guid() is an SDL2 method
+                pass
```
Starbuck5, line 118: Same as above. We don't support SDL1 anymore, we write the examples for the current version. It should be fine without a try except?

novialriptide:
Pushed a new commit, everything looks fine after the changes
image

novialriptide:
Some changes were made. I rebased this branch to the main branch which removed the unneeded "merge" commits and squashed some commits from 16 to 2.

PR #3212, opened 2022-06-05
More Python f-strings with flynt

cclauss:
This PR was created with the command: [`flynt .`](https://pypi.org/project/flynt)
```
Running flynt v.0.76

Flynt run has finished. Stats:

Execution time:                            1.944s
Files checked:                             204
Files modified:                            39
Character count reduction:                 592 (0.02%)

Per expression type:
Old style (`%`) expressions attempted:     120/241 (49.8%)
`.format(...)` calls attempted:            19/19 (100.0%)
No concatenations attempted.
F-string expressions created:              119
```

PR #3213, opened 2022-06-05
Fix print () --> print()

cclauss:
[`psf/black`](https://github.com/psf/black#readme) would probably do most of this but just to make the code look modern.

As suggested by @novialriptide at https://github.com/pygame/pygame/pull/3212#pullrequestreview-995979633

### DIFF
--- a/buildconfig/config.py
+++ b/buildconfig/config.py
@@ -127,7 +127,7 @@ def writesetupfile(deps, basepath, additional_lines):
                 aparts = al.split()
                 if aparts and parts:
                     if aparts[0] == parts[0]:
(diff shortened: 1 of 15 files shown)


novialriptide:
That was quick.

Starbuck5:
> [psf/black](https://github.com/psf/black#readme) would probably do most of this but just to make the code look modern.

We do use black, but not on the buildconfig folder I guess 🙂
All the ones you've corrected outside the buildconfig folder were inside comments, which makes sense.

dr0id (approved):

MyreMylar (approved):
LGTM 👍

PR #3214, opened 2022-06-05
CircleCI: Let's get started...

cclauss:
We can do much more with CircleCI in the future but for now, let's at least make things green...
* https://app.circleci.com/pipelines/github/pygame/pygame

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,21 @@
+# https://circleci.com/docs/2.0/language-python
+# https://circleci.com/developer/orbs/orb/circleci/python
+
+version: 2.1
+orbs:
+  python: circleci/python@2.0.3
+jobs:
+  build:
+    executor: python/default
+    steps:
+      - checkout
+      - run: sudo apt-get update
+      - run: sudo apt-get install --yes libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev libsdl2-ttf-dev libfreetype6-dev \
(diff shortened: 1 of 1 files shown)


ankith26 (commented):
At the current state, this PR is doing something similar to what our ubuntu-sdist builds are doing on github actions, so this is redundant.

CircleCI will be added as part of #3187 and this PR will heavily conflict with that.
Till that PR is merged I don't think there's any problem if CircleCI keeps complaining, since that does not block merges.

A better solution would be to somehow turn of circleci for now, I will try to look into that in the circleci dashboard and see if I have access to do the same

Thanks for taking the time and effort to contribute nonetheless, you are already doing amazing work with so many other PRs ;)

cclauss:
How does a single file PR heavily conflict with anything?  Green is better than red.  #3187 has been sitting open for a week and CircleCI has been failing every job for a week.

ankith26:
When I said heavily I meant that this entire PR would essentially be reverted, as you would understand too.

That being said, I see there's no easy way for me to just stop CircleCI from running ATM.

Sorry if the PR closing was rather hasty, I'm still up for this PR if it does something more minimal for now (like just passing directly) so that it does not waste our build minutes (yeah circleci does not have unlimited "credits" on free plans unlike github actions, which we use for most of our needs)

cclauss:
Let’s see if this passes as is and then before we merge, I can make one more commit that comments out lines 11_17 and adds `- run: echo "Hello"` so we minimize runtime.

PR #3215, opened 2022-06-05
AppVeyor: Use `py` instead of %PYTHON% and %PYTHON_VERSION% ?

cclauss:
Would use of [`py`](https://docs.python.org/3/using/windows.html#python-launcher-for-windows) allow us to simplify `appveyor.yml` by dropping variables `%PYTHON%` and `%PYTHON_VERSION%`?

### DIFF
--- a/buildconfig/appveyor.yml
+++ b/buildconfig/appveyor.yml
@@ -84,6 +84,8 @@ environment:
 
 init:
   - "ECHO %PYTHON% %PYTHON_VERSION% %PYTHON_ARCH%"
+  - py --version
+  - py -VV
 
 install:
 


ankith26:
Not sure how this would work out on appveyor... because appveyor has python installs on custom paths

cclauss:
Seems to work for them… https://www.appveyor.com/docs/lang/python/#quick-start

cclauss:
AppVeyor images already have py3.{7,8,9,10} pre installed on their images and `py` works as expected with them but if they are replace the `py` is broken.
https://www.appveyor.com/docs/windows-images-software/#python

PR #3216, opened 2022-06-05
Typos: get em out of here!

Starbuck5:
Used `codespell` as suggested by cclauss, going through manually and applying the ones that weren't false positives.

**the diff size is because of cython, don't worry**

This does change a couple error messages from starting with the word "unsupport" to "unsupported." I think this is alright, since I could barely find evidence of the error messages' existence on the internet, so it seems they aren't being run into much, don't need to maintain searchability.

I also added the new threshold fixed name method to the cython sprite module, since the spellchecker picked up the "treshold". Not sure if _sprite is out of date in more places, but that's not the purpose of this PR.

### DIFF
--- a/examples/music_drop_fade.py
+++ b/examples/music_drop_fade.py
@@ -111,7 +111,7 @@ def play_next():
 def draw_text_line(text, y=0):
     """
     Draws a line of text onto the display surface
-    The text will be centered horizontally at the given y postition
+    The text will be centered horizontally at the given y position
     The text's height is added to y and returned to the caller
     """
     screen = pg.display.get_surface()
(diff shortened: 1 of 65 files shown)


examples/music_drop_fade.py
```diff
@@ -122,12 +122,12 @@ def draw_text_line(text, y=0):
     return y
 
 
-def change_music_postion(amount):
+def change_music_position(amount):
```
ankith26, line 125: This function name fix does not have the needed corresponding fixes in function calls in this file

ankith26 (approved):
Overall, nice PR! 🎉 

I wonder whether it'd be useful getting this to run on our CI with a way to suppress false positives

cclauss:
Sweet!

examples/music_drop_fade.py
Starbuck5: Ah, thank you!

illume (approved):
Begone typos! 🪄

cclauss (approved):

novialriptide (approved):

PR #3217, opened 2022-06-05
Deprecate event.joy and joystick.get_id

zoldalma999:
`event.joy` and `joystick.get_id` is deprecated in the docs (and has been for a long time now), and should be removed soon. PR #2834 does that already, but since there were no prior warnings (as in python warnings), removing it now would break a fair amount of existing code. Hopefully by waiting a release or two, with warnings on these, there won't be much issues.

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -1284,14 +1284,21 @@ pg_event_dealloc(PyObject *self)
 }
 
 #ifdef PYPY_VERSION
-/* Because pypy does not work with the __dict__ tp_dictoffset. */
+/* Because in pypy PyObject_GenericGetAttr and PyObject_GenericSetAttr does
+ * not work with the __dict__ tp_dictoffset. */
 PyObject *
 pg_EventGetAttr(PyObject *o, PyObject *attr_name)
 {
-    /* Try e->dict first, if not try the generic attribute. */
-    PyObject *result = PyDict_GetItem(((pgEventObject *)o)->dict, attr_name);
+    /* check attribute first and if it cannot find it, check event.dict */
+    PyObject *result = PyObject_GenericGetAttr(o, attr_name);
+
     if (!result) {
-        return PyObject_GenericGetAttr(o, attr_name);
+        result = PyDict_GetItem(((pgEventObject *)o)->dict, attr_name);
+
+        if (!result) {
+            return NULL; /* error is set by GenericGetAttr already */
+        }
+        PyErr_Clear();
     }
     return result;
 }
@@ -1365,9 +1372,6 @@ pg_event_joy_get(pgEventObject *self, void *closure)
 {
     PyObject *joy = PyDict_GetItemString(self->dict, "joy");
     if (!joy) {
-        if (PyErr_Occurred()) {
-            return NULL;
-        }
         return RAISE(PyExc_AttributeError,
                      "'Event' object has no attribute 'joy'");
     }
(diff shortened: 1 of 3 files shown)


src_c/event.c
```diff
@@ -1365,6 +1365,54 @@ static PyMemberDef pg_event_members[] = {
...
+    if (self->type < PGE_USEREVENT) {
+        if (PyErr_WarnEx(PyExc_DeprecationWarning,
+                         "the joy attribute is deprecated in favour of "
+                         "instance_ids, and "
+                         "will be removed in the future",
+                         1) != 0) {
+            return NULL;
+        }
+    }
+    Py_INCREF(joy);
```
Starbuck5, line 1391: I'm not sure this is necessary. PyDict_GetItemString borrows the reference, and then if it gets put into a variable on the python side that would do the incref?

Starbuck5:
This looks great! Even a test for a deprecation warning, amazing.

Not sure why the CI is having such trouble.

illume:
I'm not sure if joy really should be removed.

src_c/event.c
ankith26: This is necessary. On the python side, it is expected that a strong reference is always returned (so the incref does not happen automatically, the C extension programmer must take care of it)

Removing this will definitely cause problems and lead to segfaults in random places

Starbuck5:
Could you rebase this against main and let the tests run again?

src_c/event.c
```diff
@@ -1284,14 +1284,21 @@ pg_event_dealloc(PyObject *self)
...
 pg_EventGetAttr(PyObject *o, PyObject *attr_name)
 {
-    /* Try e->dict first, if not try the generic attribute. */
-    PyObject *result = PyDict_GetItem(((pgEventObject *)o)->dict, attr_name);
+    /* check attribute first and if it cannot find it, check event.dict */
+    PyObject *result = PyObject_GenericGetAttr(o, attr_name);
+
     if (!result) {
-        return PyObject_GenericGetAttr(o, attr_name);
+        result = PyDict_GetItem(((pgEventObject *)o)->dict, attr_name);
```
ankith26, line 1296: Since you changed the order of `PyObject_GenericGetAttr` and `PyDict_GetItem`, there is a new potential bug. 

While `PyDict_GetItem` is documented to not setting new errors, it is also documented that this can silencing errors in some cases. In this case, the error set by `PyObject_GenericGetAttr` is discarded, and this function will end up returning `NULL` without an error set.

Just to be safe, we should be using [`PyErr_Fetch`](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Fetch)/[`PyErr_Restore`](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Restore) pair around these bits

zoldalma999: Good catch!

illume:
If it's used in old programs I don't think it should be removed. Why break them?

I'm fine to deprecate it, but I don't think we want to remove it.

zoldalma999:
> I'm fine to deprecate it, but I don't think we want to remove it.

I would not remove them yet either, since probably almost all pygame games that support joysticks use it. Personally I think we should wait at least a year or two before considering removing it (or during a release which break backcompat if that ever happens)

MyreMylar:
Toggle this PR to see if the test still pass.

MyreMylar (approved):
This seems fine to me.

PR #3208, opened 2022-06-04
pyupgrade --py36-plus **/*.py

cclauss:
https://github.com/asottile/pyupgrade/blob/main/README.md

Python f-strings are smaller, faster, and easier to read.  https://www.scivision.dev/python-f-string-speed

PR #3218, opened 2022-06-05
Add pre-commit to run setup.py format and lint

cclauss:
Add [`pre-commit`](https://pre-commit.com) to run `setup.py format` and `setup.py lint` on the changed files on every commit.  Avoids creating pull requests that will fail the basic continuous integration tests. Avoids having to run these format and lint manually.  Makes it easy to add other tools like bandit, codespell, isort, mypy, etc.

To enable these pre-commit hooks run:
`python3 -m pip install pre-commit` or `brew install pre-commit`
Then in the project root directory run `pre-commit install`
Learn more about this config here: https://pre-commit.com

As discussed at https://github.com/pygame/pygame/pull/3208#issuecomment-1146624904

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,14 +11,14 @@ repos:
     hooks:
       - id: format
         name: format
-        entry: python3 setup.py format
+        entry: bash -c "python3 setup.py format"
         language: python
         additional_dependencies:
           - black
           - clang-format
       - id: lint
         name: lint
-        entry: python3 setup.py lint
+        entry: bash -c "python3 setup.py lint"
         language: python
         additional_dependencies:
           - black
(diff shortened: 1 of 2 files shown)


illume:
Is there a way to move this file out of the root these days (perhaps in the setup.cfg)?

When I asked for this feature with pre-commit in the past, I got an... interesting response. So I don't know if they want to add support for that.

If there's no way still to move the pre-commit config file out of root, I'd be strongly against adding it. Based on the interactions I had with a maintainer of pre-commit I'd be mildly against using it unless there's no other alternative and lots of other pygame devs are strongly for it.

cclauss:
I feel your pain on the _interesting responses_...  I have gotten several of those myself. :-)

This is a .file so it will not appear in directory listings but it would appear https://github.com/pygame/pygame  I do not see a way to put that file elsewhere.

Formatting changes can be a pain because we want them to be automatic, not manual.  Getting all collaborators to configure autoformatting in their IDEs is a hit-or-miss proposition.  This project has some very long CI jobs and it would be good to not run all those jobs if basic format and lint fail.  https://pre-commit.ci does a good job of making automatic commits on formatting changes even for contributors that do not have pre-commit installed locally.

illume:
Yeah, I'm all for pre commit hooks... just not for extra files in the root which "pre-commit" requires (I think requires anyway). This was the older issue for `setup.py lint` and `setup.py format` and where pre-commit was discussed a little bit: https://github.com/pygame/pygame/issues/1896

We have gone to extremes to avoid the "config file soup" common in many projects. In order to make the repo a little bit more approachable. Less files there means it confuses people less.

If there's a way to use "pre-commit" (or another tool) so it can use a config file in a different place ("buildconfig/" probably but even in src_c/src_py) I'd be happy with that. Then people who want to use a commit hook can run their command and use it, without it affecting newbies. Perhaps pre-commit can be extended, or a wrapper written so the config file doesn't need to be there? Alternatively, perhaps another tool for pre-commit hooks can work ok.

---

- ps. With at least vs-code there is a config file which can recommend extensions to use. Maybe that's something to consider (and for other editors). I think they can be in any subfolder. Of course this is only for one editor.
- pps. Tox, and docker are other tools which might be considered to solve `make check` local checking.

cclauss:
Understood.  Let's close this for now and I will give it some more thought.  Thx.

PR #3220, opened 2022-06-06
setup.cfg: Sort pylint disables to simplify future edits

cclauss:
Sorting long lists makes it easier to spot missing entries and more difficult to create duplicate entries.

In VSCode, select lines, Shift-Command-P and Sort Lines Ascending.

Removed `fixme` from the enable list because it is also on the disable list and it must be disabled.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,14 +1,14 @@
 [tox:tox]
 envlist = py{36,37,38,39,310}
-skipsdist = True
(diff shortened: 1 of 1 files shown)


illume (approved):
👍 nice. Thanks.

ankith26 (approved):
LGTM! Thanks again! 😎

PR #3221, opened 2022-06-06
setup.py lint: Add codespell to find typos

cclauss:
https://pypi.org/project/codespell -- Follow-on to #3216

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -25,7 +25,7 @@ pyinstaller40 =
 #   ignored by codespell.  Typos should be fixed but for instance Mechanim
 #   is the name of a software product so we want codespell to ignore that.
 ignore-words-list = ans,appen,datas,fo,mechanim,nd,nin,nnumber,paeth,pevents,statics,testin
-skip = *.json
+skip = ./src_c/*,*.json
 
 [isort]
 include_trailing_comma = True
--- a/setup.cfg
+++ b/setup.cfg
@@ -20,6 +20,13 @@ commands =
 pyinstaller40 =
 	hook-dirs = pygame.__pyinstaller:get_hook_dirs
 
+[codespell]
+# ignore-words-list is a comma-separated list of lowercase words to be
+#   ignored by codespell.  Typos should be fixed but for instance Mechanim
+#   is the name of a software product so we want codespell to ignore that.
+ignore-words-list = ans,appen,datas,fo,mechanim,nd,nin,nnumber,paeth,pevents,statics,testin
+skip = *.json
+
 [isort]
 include_trailing_comma = True
 known_first_party = pygame
--- a/setup.py
+++ b/setup.py
@@ -140,10 +140,8 @@ def compilation_help():
     if the_system == 'Linux':
         if hasattr(platform, 'linux_distribution'):
             distro = platform.linux_distribution()
-            if distro[0].lower() == 'ubuntu':
-                the_system = 'Ubuntu'
-            elif distro[0].lower() == 'debian':
-                the_system = 'Debian'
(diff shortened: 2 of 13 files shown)


setup.cfg
```diff
@@ -20,6 +20,10 @@ commands =
 pyinstaller40 =
 	hook-dirs = pygame.__pyinstaller:get_hook_dirs
 
+[codespell]
+ignore-words-list = ans,appen,datas,fo,nd,mechanim,paeth,testin
```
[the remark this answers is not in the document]
illume: Cool. Sorry, I missed that. Thanks for the explanation.

setup.cfg
```diff
@@ -20,6 +20,10 @@ commands =
...
+skip = ./src_c/*,*.json
```
ankith26, line 25: I'm confused what this line is meant to skip. Are all files in `src_c` being skipped here?

ankith26: Since this PR is not very large, I wouldn't mind if that happened in this PR. I think https://github.com/pygame/pygame/pull/3216 fixed most of the issues anyways, right?

ankith26 (commented):
There seem to be some cython (generated files) changes that don't have corresponding changes in the cython source.

These generated files must never be modified by hand or by any other tool

ankith26:
I was referring to cython files in general... 

Currently, cython generated files are

0. All files under `src_c/_sdl2` except `touch.c`
1. `_sprite.c`
2. `pypm.c`

src_c/include/sse2neon.h
```diff
@@ -4569,7 +4569,7 @@ FORCE_INLINE __m128 _mm_cvtpi32_ps(__m128 a, __m64 b)
 
 // Convert packed signed 32-bit integers in a to packed single-precision
 // (32-bit) floating-point elements, store the results in the lower 2 elements
-// of dst, then covert the packed signed 32-bit integers in b to
+// of dst, then convert the packed signed 32-bit integers in b to
 // single-precision (32-bit) floating-point element, and store the results in
 // the upper 2 elements of dst.
 //
```
ankith26, line 4575: This is a vendored file from https://github.com/DLTcollab/sse2neon

I don't think we should be manually editing this one too much

cclauss: The typo is already fixed on that repo.

setup.cfg
ankith26: I would suggest removing this `skip` line here. I have an alternate proposal and I'll mention it in my next review

setup.py
```diff
@@ -904,6 +894,7 @@ def filter_files(path_obj, all_files, allowed_files, disallowed_files):
             commands = {
                 "clang-format": ["--dry-run", "--Werror", "-i"] + c_files,
                 "black": ["--check", "--diff"] + python_directories,
+                "codespell": [],  # Codespell configuration is in setup.cfg
```
ankith26, line 897: I would suggest that the skip files are passed here via the command line.

The advantage of doing this is that we can re-use the `c_file_disallow` already defined here

setup.py
```diff
@@ -904,6 +894,7 @@ def filter_files(path_obj, all_files, allowed_files, disallowed_files):
             commands = {
                 "clang-format": ["--dry-run", "--Werror", "-i"] + c_files,
                 "black": ["--check", "--diff"] + python_directories,
+                "codespell": [],  # Codespell configuration is in setup.cfg 
                 # Test directory has too much pylint warning for now
                 "pylint": ["src_py"],
             }
```
ankith26, line 900: `codespell` should be added in `commands` in the else clause below, too, I think.

ankith26:
Any reason for closing this PR?

PR #3222, opened 2022-06-06
setup.cfg: Add mypy configuration

cclauss:
$ `mypy .`  # https://mypy.readthedocs.io
```
Success: no issues found in 51 source files
```
### setup.cfg
```
[mypy]
exclude = setup.py|(buildconfig*|src_py*|test*)/$
ignore_missing_imports = True
install_types = True
show_error_codes = True
```


examples/aliens.py
```diff
@@ -186,7 +187,7 @@ class Bomb(pg.sprite.Sprite):
     """A bomb the aliens drop."""
 
     speed = 9
-    images = []
+    images: List = []
```
ankith26, line 190: Would be nice to have the type hinting narrowed down further, wherever a plain `List` is used... it should be `List[type]`. 

PS: `Union`s can be used if the lists are heterogenous but you probably know this already

src_py/sysfont.py
```diff
@@ -24,11 +24,12 @@
...
 import warnings
 from os.path import basename, dirname, exists, join, splitext
+from typing import Dict
 
 from pygame.font import Font
 
 OpenType_extensions = frozenset((".ttf", ".ttc", ".otf"))
-Sysfonts = {}
+Sysfonts: Dict[str, str] = {}
 Sysalias = {}
```
ankith26, line 33: why is `Sysalias` not typed here?

examples/aliens.py
cclauss: What data types does this list hold?

src_py/sysfont.py
cclauss: `mypy` does not request that a type be specified (hopefully because it can intuit that from the code).

examples/aliens.py
cclauss:
```suggestion
    images: List[pg.Surface] = []
```

MightyJosip (approved):
Thank you. In my opinion it looks good now, we will wait for ankith opinion and then merge it

examples/midi.py
```diff
@@ -639,7 +640,7 @@ class Keyboard:
     white_key_width, white_key_height = _rects["white none"].size
     black_key_width, black_key_height = _rects["black none"].size
 
-    _updates = set()
+    _updates: Set = set()
```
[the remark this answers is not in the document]
cclauss: What data types does this set hold?

setup.cfg
```diff
@@ -28,6 +28,12 @@ line_length = 88
 multi_line_output = 3
 src_paths = src_py
 
+[mypy]
+exclude = setup.py|(buildconfig*|src_py*|test*)/$
```
ankith26, line 32: Why are these files being excluded here?

PR #3223, opened 2022-06-08
Update README.rst

vnesslange:
Added Community Standard to README for better visibility.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,6 +16,17 @@ these programs a more intuitive task.
 New contributors are welcome.
 
 
+Community Standards
+-------------------
+
+We are respectful, open and considerate to each other. We follow the Python comuunity code of conduct. 
+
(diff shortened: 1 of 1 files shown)


novialriptide:
I don't think we should add this, since it's common sense to be respectful to others. There are also a lot of grammar and spelling mistakes in this PR.

ankith26 (commented):
Hello, thanks for contributing to pygame! 🚀 

Most projects seem to use the file `.github/CODE_OF_CONDUCT.md` for this purpose (this way github picks it up as well) so I would not mind if this went in there, and the readme linked to it.

As pointed above, this has a couple of typos too, that would need fixing

Starbuck5:
I'm going to ahead and close this PR now, for the issues pointed about by the other commenters.

We would need to really discuss what we want our community standards to be before doing this.

PR #3225, opened 2022-06-08
sysfont.py: check for emscripten platform

pmp-p:
emscripten platform does/can not scan for fonts with subprocess

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -20,21 +20,23 @@
 """sysfont, used in the font module to find system fonts"""
 
 import os
-import subprocess
 import sys
 import warnings
 from os.path import basename, dirname, exists, join, splitext
 
 from pygame.font import Font
 
+if sys.platform != "emscripten":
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
LGTM! Thanks 🥳 🚀

Starbuck5:
Note to self: emscripten is considered a unix environment. Makes sense, I suppose.

Starbuck5 (approved):

src_py/sysfont.py
```diff
@@ -196,6 +198,9 @@ def initsysfonts_unix(path="fc-list"):
     """use the fc-list from fontconfig to get a list of fonts"""
     fonts = {}
 
+    if sys.platform == "emscripten":
```
illume, line 201: This makes the code run, which is good. 

I made an issue because this is silently disabled so it's good to track somewhere that we need this fixed/implemented: https://github.com/pygame/pygame/issues/3432

PR #3226, opened 2022-06-08
Add a grid movement example

Masonstiller:
Adds example of basic grid movement


examples/grid.py
```diff
@@ -0,0 +1,62 @@
+import pygame
```
novialriptide, line 1: In other example files, they import `pygame` as `pg`, I don't like it, but it's better to stay consistent.

examples/grid.py
```diff
@@ -0,0 +1,62 @@
...
+    def __init__(self):
+        pygame.init()
+        self.clock = pygame.time.Clock()
+        pygame.display.set_caption(TITLE)
+        self.surface = pygame.display.set_mode((WINDOW_WIDTH, WINDOW_HEIGHT))
+        self.loop = True
+        self.player = Player(self.surface)
+
+    def main(self):
+        while self.loop:
```
novialriptide, line 36: I think it's better to merge the two functions `events()` and `draw()` together.

examples/grid.py
```diff
@@ -0,0 +1,62 @@
...
+    def draw(self):
+        self.surface.fill((0, 0, 0))
+        for row in range(TILES_HORIZONTAL):
+            for col in range(row % 2, TILES_HORIZONTAL, 2):
+                pygame.draw.rect(self.surface, (40, 40, 40), (row * TILE_SIZE, col * TILE_SIZE, TILE_SIZE, TILE_SIZE))
+        self.player.draw()
+        pygame.display.update()
+
+
+if __name__ == "__main__":
```
novialriptide, line 60: A question to future reviewers, should we keep this in? `if __name__ == "__main__":` isn't in other examples.

novialriptide (changes requested):
Great PR, I haven't run the example yet, but I have a few nitpicks here.

examples/grid.py
ankith26: I think this is good practice... so maybe all examples should have it?

novialriptide: Good practice indeed, but `import pygame as pg` is a little wack as well imo.

MyreMylar:
looks like this needs a pass through the formatting linters with a:

```
black examples/grid.py
```

Are you still interested in finishing this off @Masonstiller ?

Masonstiller:
I’ll submit an improved version based off the feedback I’ve gotten in the next couple of days. Sorry for the long pause in development have been crazy busy lately.

examples/grid.py
```diff
@@ -0,0 +1,59 @@
+import pygame as pg
+
+TITLE = "Grid"
+TILES_HORIZONTAL = 10
+TILES_VERTICAL = 10
+TILE_SIZE = 100
+WINDOW_WIDTH = 1000
+WINDOW_HEIGHT = 1000
```
Starbuck5, line 8: This window height will be too big for some people (like if they have display scaling from the OS). Maybe something like 800x800 would be safer.

examples/grid.py
```diff
@@ -0,0 +1,59 @@
...
+                    self.loop = False
+            elif event.type == pg.MOUSEBUTTONUP:
+                pos = pg.mouse.get_pos()
+                self.player.move(pos)
+        pg.display.update()
+
+
+if __name__ == "__main__":
+    mygame = Game()
+    mygame.main()
```
Starbuck5, line 59: Needs a trailing newline.

Starbuck5: Also should have a `pg.quit()` at the end

examples/grid.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self, surface):
+        self.surface = surface
+        self.pos = (50, 50)
+
+    def draw(self):
+        pg.draw.circle(self.surface, (255, 255, 255), self.pos, 50)
+
+    def move(self, target):
+        x = (100 * (target[0] // 100)) + 50
+        y = (100 * (target[1] // 100)) + 50
```
Starbuck5, line 21: Is this the same as a math.floor?

Masonstiller: Yes. Should I use math.floor for syntax reasons?

examples/grid.py
Masonstiller: Good idea.

examples/grid.py
novialriptide: @Masonstiller Yup

MyreMylar:
This needs a:

```
black format examples/grid.py
```

to pass the linter CI up there, otherwise we can't merge it.

MyreMylar (changes requested):
Just needs linting @Masonstiller.

illume (approved):
Thanks 👍

illume:
Merged in https://github.com/pygame/pygame/pull/3754

Issue #2724, opened 2021-09-17
Port off of Distutils


Starbuck5:
This has come up in some discussions, but it would be good to have an issue tracking it.

Distutils has been marked as deprecated in Python 3.10, and will be removed in Python 3.12.

Reference: https://www.python.org/dev/peps/pep-0632/

We should probably move to setuptools, rather than some third party solution. The PEP authors say they expect minimal disruption from the switch, since setuptools has most of the same functionality, the same way. However, I'm not sure it's going to be so easy for us, based on my limited knowledge of setup.py and the config_ scripts.

The aging of distutils is already hurting us though. For example, it can't find vcvarsall.bat (MSVC) when I build locally using VS 2019, so I had to patch my own installation of distutils.

One good first step for this is to simplify the build process itself, independent of distutils or setuptools. For example, I believe there's unused code to make .msi installers, unused since pygame 1.9.1 according to the pygame downloads page. And dropping SDL1 will help, because there's still varied things to handle the various SETUP configs.

illume:
There was another issue discussing this meson/cmake, where at least Ankith and I talked about wanting to wait to see where the pieces land.

Then I noticed this issue here...

> One good first step for this is to simplify the build process itself

I agree! haha.

I started a [buildconfig modernization](https://github.com/pygame/pygame/issues/3415) issue. The idea is to better document/test/type/check with pylint that code. From the conda packaging patches and problems I think we can use that as a way to try and standarise the config APIs for at least things like includes/libraries/etc. 

Modernization of buildconfig will make it easier for us to port to another system (if we go that way).

PR #3228, opened 2022-06-08
setup.py: fix user+deprecation warnings on 3.11+

pmp-p:
should fix both
```
setup.py:71: DeprecationWarning: The distutils package is deprecated and slated for removal in Python 3.12. Use setuptools or check PEP 632 for potential alternatives
  import distutils
```
```
_distutils_hack/__init__.py:17: UserWarning: Distutils was imported before Setuptools, but importing Setuptools also replaces the `distutils` module in `sys.modules`. This may lead to undesirable behaviors or errors. To avoid these issues, avoid using distutils directly, ensure that setuptools is installed in the traditional way (e.g. not an editable install), and/or make sure that setuptools is always imported before distutils.
```


ankith26 (commented):
distutils should be fully removed and we should be migrating to setuptools (related issue: #2724)

cclauss (approved):
> distutils should be fully removed and we should be migrating to setuptools

Let's land this PR as is because it already solves the issues in the PR title.

You can open a separate PR to do the complete migration away from distutils.

ankith26:
I have a WIP distutils port branch that I've been working on already (not much progress and no ETA)
It would be hard to do this considering the amount of distutils specific hacks we have, that (afaik) has no convenient setuptools stable API to accomplish (except from directly using the patched distutils that setuptools uses)

That being said, I did not request any change on this PR, my first comment was more of an FYI
I will approve this once I learn more about the implications (this might not be as straightforward as it looks, because setuptools patches its own distutils and we've been using this patch unknowingly for a while)
CI passes on this PR so maybe this PR is not as risky as I think

MyreMylar (approved):
Tested this a bunch locally and pygame still compiles and runs fine on windows with the order of includes swapped.

nthykier (approved):

PR #3230, opened 2022-06-11
FRect implementation

Emc2356:
in short this is `pygame.Rect` but it has the ability to hold floats instead of just integers
```py
>>> import pygame
pygame 2.1.3.dev5 (SDL 2.0.22, Python 3.9.4)
>>> r = pygame.FRect(1.1, 2.2, 3, 5.6)
>>> r
<frect(1.1, 2.2, 3.0, 5.6)>
>>> r.topleft += pygame.Vector2(2.2, 5.1)
>>> r
<frect(3.3, 7.3, 3.0, 5.6)>
```

**Things to verify once everything else is done: Starbuck**
- [ ] This comes from an older branch, we need to make sure recent Rect.c patches make it in here.
- [ ] Since rect.c build SETUP changes, we need to verify changes to Emscripten and Android setup can support this.


Starbuck5:
High level description of how it works shared on discord:

> basically i have 3 different "APIs"
RectExport_*
RectImport_*
RectOptional_*
RectExport_* is meant to receive a name to define 
RectImport_* is meant to receive something that wont define but will use to implement RectExport_*
RectOptional_* is meant to be the optional features, currently just freelist that is supported only with PyPy

src_c/pgcompat.h
```diff
@@ -1,10 +1,11 @@
 /* Python 2.x/3.x compatibility tools (internal)
  */
+#include "include/pgcompat.h"
+#include <SDL.h>
+
 #ifndef PGCOMPAT_INTERNAL_H
 #define PGCOMPAT_INTERNAL_H
 
-#include "include/pgcompat.h"
-
```
Starbuck5, line 7: I think should go back inside the header guard.

src_c/include/_pygame.h
```diff
@@ -64,7 +64,7 @@
     (PG_VERSIONNUM(PG_MAJOR_VERSION, PG_MINOR_VERSION, PG_PATCH_VERSION) >= \
      PG_VERSIONNUM(MAJOR, MINOR, PATCH))
 
-#include "pgcompat.h"
+#include "../pgcompat.h"
```
Starbuck5, line 67: This scares me to my core
/s
But only kinda

Emc2356: i couldnt find another way to include `src_c/pgcompat.h` 
if you have any suggestions about how i should do that please tell me

fortwoone (approved):
Seems fine to me now

novialriptide:
I need to recap a few things that was from the old pull request.
- Rect and FRect are not in separate files since it would require 3000+ lines to do so.
- Some variables are capitalized since they were copied from SDL.

MyreMylar:
Notes on things for me to check on review:

- [ ] performance of Rect and FRect - both for interest and to see if there has been any change over pre-this PR for Rect.
- [ ] can you use FRects in the common Rect places? blitting, sprites
- [ ] rounding behaviour of floats to pixel positions (does this just solve the 'moving backwards issue' we see reported a bit?)

MyreMylar (approved):
Looking good from my testing.

The only thing that I spotted was FRect not being accepted in typehint for sprite, which I mentioned to you in discord.

I think changing to this:
```
class _HasRect(Protocol):
    rect: Rect, FRect
```
in `sprite.pyi` should fix it but a typehint ninja might want to take a peek.

PR #3231, opened 2022-06-11
Use cibuildwheel on our appveyor builds, latest pypy wheels on windows

ankith26:
Updates our appveyor build to make use of cibuildwheel, reducing a lot of config on our end.

The wheels we make on cpython should not change, but we will now be generating pypy 3.7, 3.8 and 3.9 wheels instead of the pypy 3.6 wheel.

All our builds will happen in one appveyor matrix, this is advantageous because it saves some build time: windows prebuilts are pulled only once and docs are generated only once. But using cibuildwheel is also going to add a little bit of extra overhead on build minutes (one of the reasons being that cibuildwheel installs python from nuget, even when appveyor has all cpython versions installed already, I've opened an issue for this on the cibuildwheel repo)

So overall, we may not see any significant changes in build time either way in this PR. Appveyor takes 21-22 minutes and a build on this PR took 26 minutes, which makes sense given we have 2 extra pypy builds on this PR.

Also updates appveyor image from VS 2015 (default) to VS 2022. As a result of this change, the MSVC static analyzer has started issuing new warnings. ~~so for now this PR does not run the static analyzer (in a future PR it needs fixing)~~
EDIT: These warnings are fixed now, and static analyzer is running

Also takes the opportunity to remove old travis and conan config while the appveyor one was removed.


Starbuck5:
> All our builds will happen in one appveyor matrix, this is advantageous because it saves some build time: windows prebuilts are pulled only once and docs are generated only once.

The disadvantage is that it doesn't show as nicely which builds are failing separately.

> Also updates appveyor image from VS 2015 (default) to VS 2022. As a result of this change, the MSVC static analyzer has started issuing new warnings, so for now this PR does not run the static analyzer (in a future PR it needs fixing)

Separate PR then, do it all at once? I wonder if the update effects optimization or built code size in a noticeable way.

> Also takes the opportunity to remove old travis and conan config while the appveyor one was removed.

This is the easiest thing to approve, ofc.

ankith26:
> The disadvantage is that it doesn't show as nicely which builds are failing separately.

Yes, and this is the case on our manylinux builds already. Github Actions has nice collapse-able categories on the logs which appveyor does not. And now, the entire build is going to fail if one thing fails (like the rest of our CI)

I think this is not that much of an issue, because it's not too hard to find out which build is failing.

I suppose we could split x86 and x64 builds into two separate matrices (this should not affect build time much, these have different prebuilts to pull, but the doc generation is common for both) if that helps for a bit more readability?

> Separate PR then, do it all at once? 

Yeah I did consider that, but that PR can take a while (if I take it on) and I think this update is a bit more urgent. I think our intermittent appveyor fails (which we have a lot of) have something to do with VS 2015 being old (I have stumbled across a compiler bug in an old VS version before, and had to workaround)

MyreMylar (approved):
Seems like a good change overall to me. I've linked a new issue to cover fixing up the static analysis build.

illume:
The conan config is potentially useful in the future, but we can always revive it from git I guess.

This does change the total build time to 31 minutes VS the older 21 minutes. Is there more stuff being built? I would think it might be slightly faster considering there is less downloading of dependencies.

illume (approved):
👍 Left a comment about the longer build time. But moving to ci build wheel is probably a good idea.

Can you please confirm if the build artifacts(wheels) are there for people to download? It seems they changed it to 1 month retention policy and there are no artifacts listed there.

ankith26:
Yup, there are more pypy builds on this PR, and the static analyzer runs under two cpython versions (one 32-bit and one 64-bit) which is why this has longer CI times.

The build artifacts are for people to download for some quick testing, useful in cases where we want to verify with an issue reporter that the bug they report has been fixed, without them having to get the whole compile toolchain and do things manually on their end

Since this has new pypi wheels, I'd like if this PR makes in time for a dev release (if it doesn't, we can also push this PR to 2.1.4)

ankith26:
Fixed a merge conflict in here with a merge commit

PR #3236, opened 2022-06-16
Fix Vector memory leak

### LINKED ISSUE #3234: Memory leak when doing Vector operations
This bug was first reported by @Matiiss.

On current main there is a memory being leaked when doing Vector calculations.

Minimal code example:
```py
import pygame
while True:
    pos = pygame.Vector2(0, 0) * 0.1
```

Few seconds after I start the code above
image

10 seconds later
image

Matiiss also tested this code in somewhat older branch (from February) and it worked well back then.

Playing around this bug a bit I noticed this change: https://github.com/pygame/pygame/blob/main/src_c/math.c#L668. This was added in April this year. And I think it is the function leaking the memory. Proof: Following lines don't leak the memory `pos = pygame.Vector2(0, 0) * pygame.Vector2(0, 0)`, or `pos += pygame.Vector2(0, 0)`. So in place and multiplication with other Vector are the only one working, and they are the only one that don't call line 668. So moving on to `_vector_subtype_new`. I am not sure, but I suspect that the problem is ~in line~ https://github.com/pygame/pygame/blob/bd3ab4ff16edf07c957d694ecdb05d3c6988f5c9/src_c/math.c#L561. ~This pointer type is never cleared from the memory, and in my opinion it is the one causing the memory leak (that is the main difference between that method and `pgVector_NEW` method that doesn't leak.~ CHECK EDIT

Anyways gonna continue my investigation tomorrow, I will try to fix it, but just in case I don't here are the relevant info I found.

I advise this bug to be the top priority and must fix before the new version because basically entire math module (one of the most used ones) is leaking the memory.

EDIT 1: @zoldalma999 
> The original implementation called pgVector_NEW, which in a sense created its own vector (with PyObject_New and then populated the class itself), instead of calling a new/init function of the class. And while the new implementation calls the new and init function of the (sub)class (in this case vector2_new, line 2078) which allocates memory for vec->coords, _vector_subtype_new will also reallocate that same thing on line 569 (while the other one not being freed) losing the memory

MightyJosip:
Fix #3234

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -561,7 +561,7 @@ _vector_subtype_new(pgVector *base)
     PyTypeObject *type = Py_TYPE(base);
     Py_ssize_t dim = base->dim;
 
-    vec = (pgVector *)(type->tp_new(type, NULL, NULL));
+    vec = (pgVector *)type->tp_alloc(type, 0);
 
     if (vec) {
         vec->dim = dim;
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -557,58 +557,26 @@ static PyMemberDef vector_members[] = {
 static pgVector *
 _vector_subtype_new(pgVector *base)
 {
-    pgVector *vec;
-    PyTypeObject *type = Py_TYPE(base);
-    Py_ssize_t dim = base->dim;
-
-    vec = (pgVector *)type->tp_alloc(type, 0);
-
-    if (vec) {
-        vec->dim = dim;
-        vec->epsilon = VECTOR_EPSILON;
-        vec->coords = PyMem_New(double, dim);
-
-        if (vec->coords == NULL) {
-            Py_DECREF(vec);
-            return (pgVector *)PyErr_NoMemory();
-        }
-    }
-    return vec;
+    return (pgVector *)(Py_TYPE(base)->tp_new(Py_TYPE(base), NULL, NULL));
 }
 
 static PyObject *
 pgVector_NEW(Py_ssize_t dim)
 {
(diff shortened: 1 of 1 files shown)


ankith26 (commented):
Hi, thanks for the PR! 🎉 

While this fix should work, I would recommend replacing the entire function with something like
```c
return (pgVector *)(Py_TYPE(base)->tp_new(Py_TYPE(base), NULL, NULL));
```
to reduce code repetition and make it less error prone to breaking when there are changes in the future here.

Also this is unrelated to the issue this PR fixes, but while you are at it you can also put a similar fix for `pgVector_NEW` right below this function

novialriptide:
Thanks for this! 👍

Starbuck5:
Is it ridiculous to think about changing _vector_subtype_new into a macro so it will be inlined? I mean, most compilers would do it already with such a small function.

Starbuck5 (approved):
LGTM. I could reproduce it on main, can't reproduce it here, and the code changes make sense.

I love these collaborative bugfixes. Matt discovered it, Zoldalma found the issue, MightyJosip fixed it. I'm pretty sure Ankith was involved somehow as well :eyes:

Anyways, I left a suggestion that we could change the function into a macro. That's up to you.

MightyJosip:
Ok I will convert it to macro, don't merge it yet

illume:
ps. Macros are usually a bad idea these days. They are harder to read and debug than functions. They are also not a good idea for performance, because the compiler can better decide to inline or not, and memory/cache is the slow part mostly now.

PR #3237, opened 2022-06-16
get_just_pressed() and get_just_released() functions

ScriptLineStudios:
Response to changes requested in #3190 (unfortunately had to open another PR because I now longer have the repo I made those changes in). As requested in #3175 there should be more global access to key information which can be accessed without going via the event loop. Here is an example of the api usage:

```python
if pygame.key.get_just_pressed()[pygame.K_b]:
    print("B Pressed")
if pygame.key.get_just_pressed()[pygame.K_a]: 
    print("A Pressed")

if pygame.key.get_just_released()[pygame.K_b]:
    print("B Released")
if pygame.key.get_just_released()[pygame.K_a]: 
    print("A Released")
 ```

### CODE REVIEW

File: docs/reST/ref/key.rst [hunk 1]
```diff
@@ -427,4 +427,28 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
 
    .. ## pygame.key.set_text_input_rect ##
 
+.. function:: get_just_pressed
```
Starbuck5, line 430: Should these by up by get_pressed?

docs/reST/ref/key.rst
```diff
@@ -427,4 +427,28 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
...
+
+   | :sl:`returns a tuple containing the most recent key presses`
+   | :sg:`get_just_pressed() -> bools`
+
+   Returns a sequence of boolean values representing the most recent key presses
+   which are fetched through the use of pygame.event.get()
```
Starbuck5, line 436: This explanation could use a bit more clarity about how/why it's interacting with pygame.event.get. Also would be great to have a small code sample (just a couple lines) that shows how it works.

Starbuck5:
This is a very impressive pull request for a first time contributor!

Left a couple notes from a quick skim.

It would be good for you to do a `setup.py format` and add type stubs to get a couple more of the test runners passing.
Type stubs are here: https://github.com/pygame/pygame/tree/main/buildconfig/stubs/pygame

src_c/constants.c
```diff
@@ -27,6 +27,7 @@
 #include "scrap.h"
 
 /* macros used to create each constant */
+
```
Starbuck5, line 30: I'd like you to delete this added line so this file doesn't become part of the diff.

docs/reST/ref/key.rst
```diff
@@ -269,6 +269,40 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
 
    .. ## pygame.key.get_pressed ##
 
+.. function:: get_just_pressed
+
+   | :sl:`returns a tuple containing the most recent key presses`
```
Starbuck5, line 274: It's technically not a tuple

docs/reST/ref/key.rst
```diff
@@ -269,6 +269,40 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
...
+   | :sg:`get_just_pressed() -> bools`
+
+   Returns a sequence of boolean values representing the most recent key presses.
+   When the event loop is pumped using pygame.event.get() The most updated state of the keyboard can be retrived using
+   SDL_GetKeyboardState and is used to get the most recent keydowns. Example:
+
+   ::
+      if pygame.key.get_just_pressed()[pygame.K_a]:
+         print("A key pressed")
```
Starbuck5, line 283: Super nitpick, but "A key pressed" is ambiguous without context. Use "b" or something else.

src_c/event.c
```diff
@@ -1901,6 +1919,29 @@ _pg_get_all_events(void)
...
+            if (eventbuf[loop].type == SDL_KEYDOWN) {
+                const Uint8 *keyboard_state = SDL_GetKeyboardState(NULL);
+                int i;
+                for (i = 0; i < NUM_KEYS; i++) {
+                    if (keyboard_state[i] && !keys_being_held_down[i]) {
+                        held_down_keys[i] = 1;
+                        keys_being_held_down[i] = 1;
+                    }
+                }
+            }
```
Starbuck5, line 1932: This is not what I had in mind. Now I understand why you documented it like that. You should use the KEYDOWN or KEYUP event itself, use it's data, not poll the whole keyboard every event.

src_c/event.c
```diff
@@ -2291,13 +2338,15 @@ MODINIT_DEFINE(event)
     }
 
     /* export the c api */
-    assert(PYGAMEAPI_EVENT_NUMSLOTS == 6);
+    // assert(PYGAMEAPI_EVENT_NUMSLOTS == 7);
```
Starbuck5, line 2341: put in assert 8?

src_c/_pygame.h
```diff
@@ -334,6 +334,6 @@ struct pgColorObject {
 #define PYGAMEAPI_COLOR_NUMSLOTS 5
 #define PYGAMEAPI_MATH_NUMSLOTS 2
 #define PYGAMEAPI_BASE_NUMSLOTS 24
-#define PYGAMEAPI_EVENT_NUMSLOTS 6
+#define PYGAMEAPI_EVENT_NUMSLOTS 9
```
Starbuck5, line 337: shouldn't this be 8? It only looks like you added two functions.

src_c/event.c
ScriptLineStudios: Okay thanks, that makes a lot more sense than what i was doing

src_c/event.c
```diff
@@ -1901,6 +1919,29 @@ _pg_get_all_events(void)
...
+            if (eventbuf[loop].type == SDL_KEYDOWN) {
+                int i;
+                for (i = 0; i < NUM_KEYS; i++) {
+                    if (i == (int)eventbuf[loop].key.keysym.scancode &&
+                        !keys_being_held_down[i]) {
+                        held_down_keys[i] = 1;
+                        keys_being_held_down[i] = 1;
+                    }
+                }
+            }
```
Starbuck5, line 1932: If my understanding of how this works is correct, isn't this for loop completely unnecessary?

You can just index using the scancode of the event directly?

Starbuck5: Also isn't `keys_being_held_down` unnecessary information now? If you get a keydown, it means it was just pressed. If you get a keyup, it means it was just released.

src_c/event.c
```diff
@@ -99,6 +99,12 @@ static SDL_TimerID _pg_repeat_timer = 0;
 static SDL_Event _pg_repeat_event;
 static SDL_Event _pg_last_keydown_event = {0};
 
+#define NUM_KEYS 512
+
+int held_down_keys[NUM_KEYS];
+int released_keys[NUM_KEYS];
+int keys_being_held_down[NUM_KEYS];
```
Starbuck5, line 106: This is nitpicky, but since you're only using 1 bit of information, this could use less memory by using a smaller integer type than int. Ints are 4 bytes (or 8 sometimes?), chars are only 1 byte, so you could do this as an array of chars. Maybe with a comment explaining that it isn't text, but an array of "bools."

src_c/event.c
```diff
@@ -1982,6 +2023,12 @@ pg_event_get(PyObject *self, PyObject *args, PyObject *kwargs)
 
     _pg_event_pump(dopump);
 
+    int i;
+    for (i = 0; i < NUM_KEYS; i++) {
+        held_down_keys[i] = 0;
+        released_keys[i] = 0;
+    }
+
```
Starbuck5, line 2031: The "fancy" way of doing this (imo) would be to use memset. (https://www.man7.org/linux/man-pages/man3/memset.3.html)

I believe this would work?
`memset(held_down_keys, 0, sizeof(held_down_keys);`

src_c/event.c
ScriptLineStudios: Thanks for this, yeah you are right. Didn't think of removing the for loop after I stopped using SDL_GetKeyboardState

src_c/event.c
```diff
@@ -1901,6 +1919,14 @@ _pg_get_all_events(void)
         for (loop = 0; loop < len; loop++) {
             if (!_pg_event_append_to_list(list, &eventbuf[loop]))
                 goto error;
+
+            if (eventbuf[loop].type == SDL_KEYDOWN) {
+                pressed_keys[(int)eventbuf[loop].key.keysym.scancode] = 'T';
```
Starbuck5, line 1924: Even though they are chars now you can and should use `0` and `1` still, no need for T and F

docs/reST/ref/key.rst
```diff
@@ -269,6 +269,40 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
...
    .. ## pygame.key.get_pressed ##
 
+.. function:: get_just_pressed
+
+   | :sl:`returns a pygame.key.ScancodeWrapper containing the most recent key presses`
+   | :sg:`get_just_pressed() -> bools`
+
+   Returns a sequence of boolean values representing the most recent key presses.
+   When the event loop is pumped using pygame.event.get() If the event is found to be an
+   SDL_KEYDOWN the SDL_Event.key.keysym.scancode is used to update the most recent key presses.
```
Starbuck5, line 279: I wrote a long review with suggestions about how to reword this and things, but I apparently never posted it?

Things
- It should say that it tracks the KEYDOWNs behind the scenes when the event system is pumped (for example by pygame.event.get())
- Therefore it's similar to checking KEYDOWN in the event loop
- It should say, as a note (.. note::), that if you require getting presses in order, you should still use the event queue KEYDOWN events.
- The fact that it's an SDL_KEYDOWN or that the scancode is what's being used to address into things is not relevant to the reader

src_c/constants.c
```diff
@@ -635,4 +635,4 @@ MODINIT_DEFINE(constants)
     }
 
     return module;
-}
+}
```
Starbuck5, line 638: Keep the trailing newline

src_c/static.c
```diff
@@ -254,6 +254,7 @@ PyGame_static_init()
 #undef pgEvent_FillUserEvent
 #undef pgEvent_Type
 #undef pgEvent_New
+#undef pgEvent_GetMostRecentEvent
```
Starbuck5, line 257: What is this?

src_c/event.c
```diff
@@ -1964,6 +1990,12 @@ pg_event_get(PyObject *self, PyObject *args, PyObject *kwargs)
 
     _pg_event_pump(dopump);
 
+    int i;
+    for (i = 0; i < NUM_KEYS; i++) {
+        memset(pressed_keys, 0, sizeof(pressed_keys));
+        memset(released_keys, 0, sizeof(released_keys));
+    }
```
Starbuck5, line 1997: This for loop zeroes the whole array out 512 times in a loop. The loop is unnecessary.

I believe sizeof(pressed_keys) will return the total bytes of the array because it's defined "statically", it has a set size.

You can memset over the whole thing at once.

src_c/key.c
```diff
@@ -68,6 +68,7 @@ key_get_repeat(PyObject *self, PyObject *_null)
 #define _PG_SCANCODEWRAPPER_TYPE_NAME "ScancodeWrapper"
 #define _PG_SCANCODEWRAPPER_TYPE_FULLNAME \
     "pygame.key." _PG_SCANCODEWRAPPER_TYPE_NAME
+#define NUM_KEYS 512
```
Starbuck5, line 71: Getting this wrong is extremely risky. Imagine an SDL scancode comes in with an id of 550, it tries to put that in the array and the entirety of pygame segfaults.

It looks like you can get this a constant from SDL in case they ever change it: https://github.com/libsdl-org/SDL/blob/main/include/SDL_scancode.h#L427 (SDL_NUM_SCANCODES)

docs/reST/ref/key.rst
```diff
@@ -269,6 +269,46 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
...
+
+   | :sl:`returns a pygame.key.ScancodeWrapper containing the most recent key presses`
+   | :sg:`get_just_pressed() -> bools`
+
+   Returns a sequence of boolean values representing the most recent key presses.
+   done by tracking the KEYDOWN events when the event loop is pumped using pygame.event.get() 
+   and is therefore similar to checking KEYDOWNS in the event loop.
+
+   .. note::
+      If you require getting the key presses in order. Use the event queue KEYDOWN events
```
Starbuck5, line 282: This is grammatically weird, it should be one sentence.

docs/reST/ref/key.rst
```diff
@@ -269,6 +269,46 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
...
+
+   ::
+      if pygame.key.get_just_pressed()[pygame.K_b]:
+         print("B key pressed")
```
Starbuck5, line 286: nitpick, could say "B key just pressed" or "B key pressed in the last game loop"

docs/reST/ref/key.rst
```diff
@@ -269,6 +269,46 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
...
    .. ## pygame.key.get_pressed ##
 
+.. function:: get_just_pressed
+
+   | :sl:`returns a pygame.key.ScancodeWrapper containing the most recent key presses`
+   | :sg:`get_just_pressed() -> bools`
+
+   Returns a sequence of boolean values representing the most recent key presses.
+   done by tracking the KEYDOWN events when the event loop is pumped using pygame.event.get() 
+   and is therefore similar to checking KEYDOWNS in the event loop.
```
Starbuck5, line 279: While it is technically the most recent key presses, I feel that could be easily confused for some sort of queue strategy. This represents the KEYDOWN events as of the last pygame.event.get() call.

ScriptLineStudios:
lol, I accidentally closed this

PR #3238, opened 2022-06-17
Fix controller build on SDL < 2.0.9

Starbuck5:
For some reason this doesn't fail on our "SDL 2.0.8" Ubuntu test machinery, but it fails on my Windows machine on SDL 2.0.8, and it makes sense that it fails.

This needs to be fixed before Emc can get his branch truly building on SDL 2.0.8 in #3230.

It would be great to get this one in quickly to help out the FRect branch.


src_c/pgcompat.c
```diff
@@ -43,3 +43,13 @@ SDL_CreateRGBSurfaceWithFormat(Uint32 flags, int width, int height, int depth,
 }
 
 #endif /*if !(SDL_VERSION_ATLEAST(2, 0, 5))*/
+
+#if !(SDL_VERSION_ATLEAST(2, 0, 9))
+int
+SDL_GameControllerRumble(void *gamecontroller, Uint16 low_frequency_rumble,
+                         Uint16 high_frequency_rumble, Uint32 duration_ms)
+{
+    SDL_SetError("SDL 2.0.9 or greater required");
```
ankith26, line 52:
```suggestion
    SDL_SetError("SDL 2.0.9 or greater required");
    return -1;
```
To set the correct error code that function calls to this will expect

ankith26 (commented):
To workaround this bug, earlier I had added `if SDL_VERSION_ATLEAST(2, 0, 9):` checks in the cython code. And I believe this only worked because of GCC compile time optimisations, that did not bother compiling `if (0) { ... }` cases. That explains why the ubuntu builds pass.

The original fix for this issue was somewhat similar to this PR, but this fix did not exactly work back then (dunno how or why)

Starbuck5:
These intermittent compiler errors are demoralizing.

Maybe it's an instance of cython-c interop being weird. Maybe I could do a pure cython solution without using pgcompat.c.

Starbuck5:
```
C:\projects\pygame\src_c\pgcompat.c : fatal error C1083: Cannot open compiler generated file: 'C:\projects\pygame\build\temp.win32-3.10\Release\src_c\pgcompat.obj': Invalid argument
```

Flaky still

Starbuck5:
The easiest solution for this may be for @zoldalma999 to complete converting this module to C.

Which is ridiculous. But may somehow be the easiest way to do this.

Starbuck5:
> But may somehow be the easiest way to do this.

Or maybe not...

PR #3239, opened 2022-06-17
Raise minimum SDL to 2.0.4, patch so it can build on that

Starbuck5:
I came in with a single question. Does pygame actually build on on SDL 2.0.1 like claimed in the README? No, it does not.

So this raises the minimum version to 2.0.4, as I saw a couple things I was hesitant to remove that were minimum 2.0.4. (Like gnarly display.c internals).

I removed compile time checks for 2.0.4 and lower, and I added a check for something that was unguarded that needed 2.0.6

I was unable to get _sdl2.video building on this in the past 30 minutes, so I'm going to post what I've got. I guess I'll mark it as a draft though. Putting this separate to #3238 for easier reviews of both.

(But this also needs #3238, or it fails on _sdl2.controller on old SDL versions)


src_c/event.c
```diff
@@ -788,14 +782,10 @@ _pg_name_from_eventtype(int type)
         case SDL_AUDIODEVICEREMOVED:
             return "AudioDeviceRemoved";
 #endif /* SDL2_AUDIODEVICE_SUPPORTED */
-#if SDL_VERSION_ATLEAST(2, 0, 2)
         case SDL_RENDER_TARGETS_RESET:
             return "RenderTargetsReset";
-#endif
-#if SDL_VERSION_ATLEAST(2, 0, 4)
```
ankith26, line 795: In this function, line 691, you missed the removal of a `#if SDL_VERSION_ATLEAST(2, 0, 4)` conditional

Starbuck5:
I suppose this is waiting on #3272 to be merged

src_c/include/pgcompat.h
```diff
@@ -102,7 +101,6 @@ typedef uint8_t Uint8;
  * Ref: https://wiki.libsdl.org/SDL_AudioDeviceEvent
  */
 #define SDL2_AUDIODEVICE_SUPPORTED
```
ankith26, line 103: Since this will be always defined now, this macro can be removed, and all guards for this should be allowed

MyreMylar (approved):
LGTM. Less outdated define salad!

src_c/include/pgcompat.h
Starbuck5: Yeah good catch once again.

ankith26 (approved):
LGTM! Thanks for the PR 🎉 🎈

PR #3242, opened 2022-06-18
Use SDL_image PNG save functions

### LINKED ISSUE #3083: Use SDL_image image saving routines

ankith26:
Closes #3083

Removes our custom PNG save code to use the SDL functions for the same, which have been available since SDL_image 2.0.0

Also removed our buildconfig dependency on pnglib and jpeglib, because we don't directly depend on these anymore (they are only sub-dependencies now)

This is a draft PR for now, there is still a failing unit test.

### DIFF
--- a/buildconfig/Setup.Android.SDL2.in
+++ b/buildconfig/Setup.Android.SDL2.in
@@ -4,9 +4,7 @@ SDL = {sdl_includes} -D_REENTRANT -DSDL2 -lSDL2
 FONT = {sdl_ttf_includes} -lSDL2_ttf
 IMAGE = {sdl_image_includes} -lSDL2_image
 MIXER = {sdl_mixer_includes} -lSDL2_mixer
-JPEG = {jpeg_includes} -ljpeg
 SCRAP =
-PNG = {png_includes} -lpng16
 FREETYPE = {freetype_includes} -lfreetype -lharfbuzz
 
 DEBUG =
@@ -15,7 +13,7 @@ DEBUG =
 #everything you can, but you can ignore ones you don't have
 #dependencies for, just comment them out
 
-imageext src_c/imageext.c $(SDL) $(IMAGE) $(PNG) $(JPEG) $(DEBUG)
+imageext src_c/imageext.c $(SDL) $(IMAGE) $(DEBUG)
 font src_c/font.c $(SDL) $(FONT) $(DEBUG)
 mixer src_c/mixer.c $(SDL) $(MIXER) $(DEBUG)
 mixer_music src_c/music.c $(SDL) $(MIXER) $(DEBUG)
--- a/buildconfig/Setup.Emscripten.SDL2.in
+++ b/buildconfig/Setup.Emscripten.SDL2.in
@@ -4,9 +4,7 @@
 #FONT = -lSDL2_ttf
 #IMAGE = -lSDL2_image
 #MIXER = -lSDL2_mixer
-#JPEG = -ljpeg
 #SCRAP =
-#PNG = -lpng
 #FREETYPE = -lfreetype -lharfbuzz
 
 DEBUG =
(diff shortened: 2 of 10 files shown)


Starbuck5:
This is important for the upcoming version of SDL_image, which is now a static build, which won't supply us with a convenient PNG lib for our buildconfig.

Starbuck5:
Sorry to revert a commit on your own branch while you're asleep, but I got the itch to work on this too, because I thought I could fix it all the way.

The PNG and JPEG libs still need to be dependencies for now, so they can be copied into the wheel. What likely happened is that you reinstalled pygame and the old libs were still in your filesystem at the correct place, but if you had uninstalled and reinstalled it wouldn't have worked. As you can see from the CI runs imageext was entirely failing to build.

Anyways, we cannot fix this well on our end, because SDL_Image 2.0.5 doesn't support 24 bit saves at all. SDL_Image 2.5.2 supports them with an oversight that we could fix on our end.

So I wrote this patch but it will not work currently:
```c
        else if (!strcasecmp(ext, "png")) {
            char converted = 0;

            /* SDL image currently saves surfaces as RGB32 if they are not exactly RGB24
               So BGR24 surfaces would be saved as RGB32. This is not a huge deal,
               but it's a simple fix on our end to keep our unit tests happy and keep
               the exact behavior people already have with our code. (We used to have a
               custom PNG saver, but now we want to just rely on SDL_image instead)
               See https://github.com/libsdl-org/SDL_image/issues/282 */
            if (!SDL_ISPIXELFORMAT_INDEXED(surf->format->format) && !SDL_ISPIXELFORMAT_ALPHA(surf->format->format)) {
                surf = SDL_ConvertSurfaceFormat(surf, SDL_PIXELFORMAT_RGB24, 0);
                converted = 1;
            }

            if (rw != NULL) {
                result = IMG_SavePNG_RW(surf, rw, 0);
            }
            else {
                result = IMG_SavePNG(surf, name);
            }

            if (converted) {
                SDL_FreeSurface(surf);
            }
        }
    }
```

In terms of our next course of action, how about we "expected fail" the test for now, and I try to patch SDL_image upstream.

Starbuck5:
Ok, I didn't expect it to fail like it's failing right now. Ugh.

Starbuck5:
Feel free to revert any of my stuff if you disagree with anything.

ankith26:
> Sorry to revert a commit on your own branch while you're asleep

Oh, that's not a problem at all :)

I think I put a 'proper' fix for this now. I reverted your revert (but left in your changes of the test skip and the setup changes which I too added some updates). I made the dependency finder handle jpeg and png dlls like it handles the rest, and now it works for me locally even on a clean build

PS: I force pushed this branch so you might need to hard reset your local branch

ankith26:
Very weird 32 bit windows tests are failing. I got a fresh python 3.9 32 bit install and built pygame with that locally, and the tests pass for me

ankith26:
Found something relevant to the current fail: https://github.com/libsdl-org/SDL_image/issues/131

Could be the same issue. So this PR has to wait

ankith26:
This PR is mostly being superseded by #3449 when that works, so I'm closing

PR #3244, opened 2022-06-18
Make textinput show a candidate list

Starbuck5:
It's what the people want!

And it's very non obvious that it needs an environment variable set. I would love someone to take this over and try to document this behavior. Or maybe we should put a programmatic api on top of this using SDL_SetHint.

Screenshot (5)
Now it can do this ^

### DIFF
--- a/examples/textinput.py
+++ b/examples/textinput.py
@@ -6,14 +6,15 @@ A little "console" where you can write in text.
 Shows how to use the TEXTEDITING and TEXTINPUT events.
 """
 import sys
+import os
 
 import pygame
 import pygame as pg
 import pygame.freetype as freetype
 
-# Version check
-if pg.get_sdl_version() < (2, 0, 0):
-    raise Exception("This example requires pygame 2.")
+# This environment variable is important
+# If not added the candidate list will not show
+os.environ["SDL_IME_SHOW_UI"] = "1"
 
 
 class TextInput:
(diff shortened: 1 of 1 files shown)


examples/textinput.py
```diff
@@ -6,14 +6,15 @@
...
 
 import pygame
 import pygame as pg
 import pygame.freetype as freetype
 
-# Version check
-if pg.get_sdl_version() < (2, 0, 0):
-    raise Exception("This example requires pygame 2.")
+""" This is very important
+    If not added the candidate list will not show """
```
ankith26, line 16: minor nitpick: This can just be regular `#` comments. `"""` are generally reserved for doc strings and multiline strings

Starbuck5: Sure.

ankith26 (approved):
👍

MyreMylar:
How can I make this UI appear while using the example with an English/US keyboard?

Starbuck5:
> How can I make this UI appear while using the example with an English/US keyboard?

That's a great question, and what took the most time on this PR, because I needed to figure that out.

This is not a complete guide, you'll have to play around with it.

Go to language settings and add a new language:
Capture

See language option in the bottom right of the screen now (ENG):
Capture

In settings > language > Japanese > Microsoft IME > Appearance, you may have to switch "IME toolbar" on.

When you switch ENG to J in the bottom right, you should have an option to change "A" to another mode. This seems to be per process, so click the window first that where you want to use it, then change the setting to "Hiragana"
Screenshot (6)

MyreMylar (approved):
Thanks for the write-up. I'd got most of the way there - didn't figure out toggling the 'A' to enable keyboard input.

I can confirm that this works great now. 👍

PR #3247, opened 2022-06-21
wasm: wrong logic masks pgExc_SDLError

pmp-p:
that is still assigning a local instead of the required global pgExc_SDLError
probably fixes all kind of  `SystemError: returned NULL without setting an exception`  because of pgExc_SDLError == NULL

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -2097,7 +2097,7 @@ MODINIT_DEFINE(base)
 {
     PyObject *module, *apiobj, *atexit;
     PyObject *atexit_register;
-#if !defined(BUILD_STATIC) || defined(NO_PYGAME_C_API)
+#if !(defined(BUILD_STATIC) && defined(NO_PYGAME_C_API))
     // only pointer via C-api will be used, no need to keep global.
     PyObject *pgExc_SDLError;
 #endif


src_c/base.c
```diff
@@ -2097,7 +2097,7 @@ MODINIT_DEFINE(base)
 {
     PyObject *module, *apiobj, *atexit;
     PyObject *atexit_register;
-#if !defined(BUILD_STATIC) || defined(NO_PYGAME_C_API)
+#if !(defined(BUILD_STATIC) && defined(NO_PYGAME_C_API))
```
ankith26, line 2100: I understand the check for `BUILD_STATIC`, but I'm not sure why `NO_PYGAME_C_API` is relevant here

pmp-p: i think when pygame c api is active the pgExc_SDLError becomes a macro that point to slot 0 of base.c, which may not work with dynamic linking and wasm ( need a global, you cannot address a pointer from another C module without using a js trampoline ( the one that handle dlopen ) to bridge the wasm memories ).

ankith26: What about the case where `BUILD_STATIC` is defined but `NO_PYGAME_C_API` is not?

pmp-p: that would be possible (far) future dynamic linking ( converting the whole libpygame.a into a pygame.so module ).

ankith26 (approved):
Thanks 🥳

MyreMylar (approved):
LGTM 👍

PR #3248, opened 2022-06-22
Fix rumble returning wrong bool on error

### LINKED ISSUE #3240: _sdl2.controller rumble result seems to be inverted
In controller.pyx:
```py
res = SDL_GameControllerRumble(self._controller, low * 0xFFFF, high * 0xFFFF, duration)
return bool(res)
```

Since Rumble returns 0 on success, a successful run will return False to the Python caller, the opposite of the documented behavior.

ankith26:
Quick PR to fix #3240


Starbuck5 (approved):
Looks good to me. Obviously it would be great to get a Zoldalma review of this.

Also, Zoldalma should add this into his WIP port of controller.pyx to C.

zoldalma999 (approved):
Thanks! 🎉

> Also, Zoldalma should add this into his WIP port of controller.pyx to C.

Yeah, will do!

PR #3253, opened 2022-06-24
Improve name parsing for windows sysfonts

Starbuck5:
By adding another "mod" to be removed from the name.

While testing Novial's #3184, I downloaded "Poppins Regular" from Google Fonts and installed it into a windows user font directory. Novial's code picks it up fine, but match_font returns none when you search for "poppins." It only knows the name "poppinsregular." This fixes that.

This could impact existing behavior. On a quick scan, I don't see any preinstalled windows fonts that use this pattern (ending with " Regular"). So it won't impact much existing stuff. I think the change is worth it, because it makes the results more correct.

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -95,7 +95,7 @@ def _parse_font_entry_win(name, font, fonts):
     :return: Tuple of (bold, italic, name)
     """
     true_type_suffix = "(TrueType)"
-    mods = ("demibold", "narrow", "light", "unicode", "bt", "mt")
+    mods = ("demibold", "narrow", "light", "unicode", "bt", "mt", "regular")
     if name.endswith(true_type_suffix):
         name = name.rstrip(true_type_suffix).rstrip()
     name = name.lower().split()


src_py/sysfont.py
```diff
@@ -95,7 +95,7 @@ def _parse_font_entry_win(name, font, fonts):
     :return: Tuple of (bold, italic, name)
     """
     true_type_suffix = "(TrueType)"
-    mods = ("demibold", "narrow", "light", "unicode", "bt", "mt")
+    mods = ("demibold", "narrow", "light", "unicode", "bt", "mt", "regular")
     if name.endswith(true_type_suffix):
         name = name.rstrip(true_type_suffix).rstrip()
```
ankith26, line 100: Unrelated to this PR but I noticed it here

[`rstrip`](https://docs.python.org/3/library/stdtypes.html#str.rstrip) is not the correct method to remove suffixes. This method will strip *any* character that is passed in the argument string and won't take care of the order.

[`removesuffix`](https://docs.python.org/3/library/stdtypes.html#str.removesuffix) is a newly added python string method for doing exactly this, but since we need to maintain 3.6+ compat I suggest
```suggestion
        name = name[:-len(true_type_suffix)].rstrip()
```

ankith26:
> This could impact existing behavior. On a quick scan, I don't see any preinstalled windows fonts that use this pattern (ending with " Regular"). So it won't impact much existing stuff. I think the change is worth it, because it makes the results more correct.

Well, this PR can still break things, tested locally (pygame main, without this PR)
```py
>>> [i for i in pygame.font.get_fonts() if i.endswith("regular")]
['yugothicuiregular', 'yugothicregular', 'dubairegular', 'lucidafaxregular', 'lucidasansregular', 'lucidasanstypewriterregular', 'hpsimplifiedjpanregular', 'hpsimplifiedhansregular']
>>> pygame.font.match_font("hpsimplifiedhans")  # This is None now, but after this PR this would return the path
>>> pygame.font.match_font("hpsimplifiedhansregular")  # this would break after this PR I think
'C:\\WINDOWS\\Fonts\\hpsimplifiedhans-regular.ttf'
```

Starbuck5:
> Well, this PR can still break things, tested locally (pygame main, without this PR)

Alrighty, your automated scan is far better than my quick visual scan, I see :smile:

The "right" way to do this is a pain, but I guess I'll try and do it that way.

What I can do is add regular as a `mod`, but in the sysfont constructor on windows, add an additional check for if the name minus "regular" is in the sysfont dictionary, and try to use that instead. But if it does do that, emit a deprecation warning.

src_py/sysfont.py
Starbuck5: Yikes, that seems like a really serious bug. I did a blame, it's been like that for at least 6 years.

Starbuck5:
> What I can do is add regular as a mod, but in the sysfont constructor on windows, add an additional check for if the name minus "regular" is in the sysfont dictionary, and try to use that instead. But if it does do that, emit a deprecation warning.

Well this actually won't be perfectly compatible, since the new font name could shadow existing fonts labeled with that name.

For example, see these two "conversion charts." First one is from current pygame, second one is after this change.

```
Yu Gothic UI Semibold -> yugothicuisemibold
Yu Gothic UI Bold (TrueType) -> yugothicui
Yu Gothic Light -> yugothic
Yu Gothic UI Light (TrueType) -> yugothicui
Yu Gothic Medium -> yugothicmedium
Yu Gothic UI Regular (TrueType) -> yugothicuiregular
Yu Gothic Regular -> yugothicregular
```

```
Yu Gothic UI Semibold -> yugothicuisemibold
Yu Gothic UI Bold (TrueType) -> yugothicui
Yu Gothic Light -> yugothic
Yu Gothic UI Light (TrueType) -> yugothicui
Yu Gothic Medium -> yugothicmedium
Yu Gothic UI Regular (TrueType) -> yugothicui
Yu Gothic Regular -> yugothic
```

If you asked for "yugothic" before, you would get "yugothic light."

Maybe the takeaway is that if I break this I should do it well. I could overhaul sysfont with a "weight" system. Rather than just bold or not, the whole scale of font weight.

src_py/sysfont.py
Starbuck5: One way to fix this would be to add "(truetype)" to the mods list. That way is uses the already existing mod removal code, so it makes things a little simpler.

PR #3259, opened 2022-06-26
Improve palette operations

Starbuck5:
I was testing something with palettes out in a REPL without a pygame window, and I was surprised when it wouldn't let me set a palette without pygame.display initialized. That didn't really make sense either, as it's a palette on a Surface in memory, it shouldn't need anything to do with display. So I tested removing the pygame code checks preventing this from happening, and I didn't find any problems.

I also modified the tests now that it doesn't need an initialized display module.

I was curious what would happen with set_at() and a palettized surface, so I documented what it does and made a test to monitor that the behavior stays that way.

### DIFF
--- a/docs/reST/ref/surface.rst
+++ b/docs/reST/ref/surface.rst
@@ -443,6 +443,9 @@
 
       This function will temporarily lock and unlock the Surface as needed.
 
+      .. note:: If the surface is palettized, the pixel color will be set to the
+                most similar color in the palette.
+
       .. ## Surface.set_at ##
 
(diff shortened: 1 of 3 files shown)


MyreMylar (approved):
LGTM tests pass locally and I also can't see any reason the display needs to be initialised to set a palette on a surface in SDL2. It may have been a hang over from something in SDL1 at some point, or perhaps some copy and paste from the similar `display.set_palette()` function 👍

ankith26 (approved):
LGTM, thank you for this PR 👍 🍰

Issue #3258, opened 2022-06-25
New Surface.ublits() function for blitting uniform Surface sequences


itzpr3d4t0r:
Follows #3257 .
# Rationale
The Surface.blits() function is exceptional at providing freedom to the user as to how a sequence of surfaces are blit, letting them decide on a per-surface basis whether to draw just a portion of them, using a different blend-type or both. It's a very good general function, but as being so, it also pays the cost of "checking what we have and choose what to do" on a per surface basis. We need something stricter, yet more performant.

# Idea
What I propose is a narrower blits() function, that I'll call **ublits()**, where the "u" stands for uniform, that will draw every Surface in the sequence fully (meaning area=None) while using the same blend-mode for each one, therefore having a more delimited behaviour, but with more performance overall.

# Implementation
I'll start from the general to then go more in depth:
### From the user perspective
The final user would implement something that follows the "signature":
```python 
ublits(((source, dest), ...), blend_flags=None) -> [Rect, ...]
or
ublits([(source, dest), ...], blend_flags=None) -> [Rect, ...]
```
So we want:
- A sequence of exact pairs: (source, dest), being (Surface, dest) anything different will result in an error.
- To be able to specify the blit-blend-mode flag once using the "blend_flags" function parameter, that will be used for every surface in the sequence
- Implicitly say that the entire surfaces will be drawn

## Current blits() slowdowns
The current blits() implementation : 
https://github.com/pygame/pygame/blob/d31eae2c804592a1b1875122c139e209055fa90e/src_c/surface.c#L1902
**has to do the following for each and  every item from the sequence**:
- `assert(itemlength >= 2);` , as the items from the sequence can't be "less than" a `(Surface, dest)` pair, but can be more (line 1959)
- Check `if (itemlength >= 3)` if so -> `argrect = PySequence_GetItem(item, 2);` (line 1965)
- Check `if (itemlength >= 4)` if so -> `special_flags= PySequence_GetItem(item, 3);` (line 1969)
- Do validity checks on the extracted `argrect`, set it to `NULL` at the start of the while loop, manage it and throw errors otherwise. (line 2004)
- Do validity checks for the `special_flags` variable (line 2022)

So to summarize there's a lot of getting, setting and checking for every surface in the sequence. All this with a specialized version of blits() would not be needed, not for each surface at least.

## Potential ublits() implementation
The proposed implementation is basically a stripped down version of the current blits() function:
[...]

PR #3261, opened 2022-06-26
Surface.fblits() function for blitting uniform Surface sequences

itzpr3d4t0r:
Implements #3258


itzpr3d4t0r:
For now this is just me messing up to see if it works. First time ever working with Python ABI or pygame. Help is very much needed.

itzpr3d4t0r:
I'm clueless as to why it doesn't build. On my machine it builds (Windows 11, ryzen 5 3600x 4.0GHz, 16 GB ddr4 300 MHz) and works with no segfaults/errors, using python 3.10. Am I missing something? Right now i'm doing performance tests that I'll publish shortly.

itzpr3d4t0r:
I ran some performance test with the following code:
```python
pygame.init()
WIDTH, HEIGHT = 400, 400
win = pygame.display.set_mode((WIDTH, HEIGHT))
NUMSURFS = 100
IMAGE = pygame.image.load("cell.png").convert()

blits_sequence_noflags = [(IMAGE, (random() * WIDTH, random() * HEIGHT)) for _
                          in range(NUMSURFS)]

blits_sequence = [(IMAGE, (random() * WIDTH, random() * HEIGHT), None,
                   pygame.BLEND_MULT) for _
                  in range(NUMSURFS)]
ublits_sequence = [(IMAGE, (random() * WIDTH, random() * HEIGHT)) for _ in
                   range(NUMSURFS)]


NUM = 1
TESTS = 500
schedule = range(1, TESTS + 1)
CONST = 0

GLOB = {
    "blits_sequence": blits_sequence,
    "ublits_sequence": ublits_sequence,
    "blits_sequence_noflags": blits_sequence_noflags,
    "win": win,
    "BLEND_ADD": pygame.BLEND_ADD
}

data = (
    [timeit("win.blits(blits_sequence)",
            number=NUM if CONST else i, globals=GLOB) for i in schedule],

    [timeit("win.blits(blits_sequence, doreturn=0)",
            number=NUM if CONST else i, globals=GLOB) for i in schedule],

    [timeit("win.ublits(ublits_sequence, special_flags=3)",
            number=NUM if CONST else i, globals=GLOB) for i in schedule],

    [timeit("win.ublits(ublits_sequence, special_flags=3, doreturn=0)",
            number=NUM if CONST else i, globals=GLOB) for i in schedule],

    [timeit("for elem in blits_sequence:"
            "   win.blit(*elem)",
            number=NUM if CONST else i, globals=GLOB) for i in schedule]
)

labels = "blits", "blits_noret", "ublits", "ublits_noret", "for-blit"

create_plot("Original", "id-num", "time(s)", schedule, labels, data,
            do_show_info=True)
create_plot("Sorted", "id-num", "time(s)", schedule, labels, data,
            sort_data=True)
```

The image being drawn is this one:
cell

if you want i can iclude the plotting.py code that includes the create_plot function and the way i get the outputs. For shortness sake i won't iclude it for now

## About the tests:
I tested in the order
- Linear Evolution: it test the behaviour of the function while it escalates linearly in call number, meaning for the first test-> 1 call, second->2, third->3 and so on.
- Constant evolution: it calls the function for a fixed number of calls many times to estimate a precise mean at that call number level

Test output include:
- Cumulative time: the cumulative sum of all test for the category
- Mean time: mean time of the category

#Tests
## LINEAR: BLEND_MODE=pygame.BLEND_ADD (1)
CUMULATIVE TIME:
BLITS |  6.629137
BLITS_NORET |  6.0083593
UBLITS |  6.4487816999
UBLITS_NORET |  5.8136385999
FOR-BLIT |  7.667642

MEAN:
BLITS |  0.013258274
BLITS_NORET |  0.012016719
UBLITS |  0.012897563
UBLITS_NORET |  0.011627277
FOR-BLIT |  0.015335284
lin_sorted_blend_add

## LINEAR: BLEND_MODE=pygame.BLENDMODE_NONE (0)
ublits() is marginally slower or equal to blits(), still faster than a for loop tho
CUMULATIVE TIME:
BLITS |  12.600961
BLITS_NORET |  11.9953409
UBLITS |  12.868253
UBLITS_NORET |  12.6041397
FOR-BLIT |  14.0967542001

CUMULATIVE TIME:
BLITS |  0.025201922
BLITS_NORET |  0.023990682
UBLITS |  0.025736506
UBLITS_NORET |  0.025208279
FOR-BLIT |  0.028193508
lin_sorted_blend_none
## CONSTANT: BLEND_MODE=pygame.BLENDMODE_NONE (0), CALLS_PER_TEST=1
this would simulate how it should behave in a while game loop, being called once per frame (once per test here test here)
CUMULATIVE TIME:
BLITS |  0.0560123
BLITS_NORET |  0.0534855001
UBLITS |  0.0558555
UBLITS_NORET |  0.0524055
FOR-BLIT |  0.0605789

MEAN:
BLITS |  0.000112025
BLITS_NORET |  0.000106971
UBLITS |  0.000111711
UBLITS_NORET |  0.000104811
FOR-BLIT |  0.000121158
myplot

## CONSTANT: BLEND_MODE=pygame.BLENDMODE_NONE (0), CALLS_PER_TEST=100
a high number of calls per test, this would simulate many ublits calls in a while game loop, each drawing its set of surfaces.
CUMULATIVE TIME:
BLITS |  4.1166529
BLITS_NORET |  3.8067728
UBLITS |  3.9735215
UBLITS_NORET |  3.7963031
FOR-BLIT |  4.5843011

MEAN:
BLITS |  0.008233306
BLITS_NORET |  0.007613546
UBLITS |  0.007947043
UBLITS_NORET |  0.007592606
FOR-BLIT |  0.009168602
myplot

## CONSTANT: BLEND_MODE=pygame.BLENDMODE_ADD (1), CALLS_PER_TEST=100
CUMULATIVE TIME:
BLITS |  2.6502764
BLITS_NORET |  2.3538351
UBLITS |  2.6044833
UBLITS_NORET |  2.3464966
FOR-BLIT |  2.9797620999

MEAN:
BLITS |  0.005300553
BLITS_NORET |  0.00470767
UBLITS |  0.005208967
UBLITS_NORET |  0.004692993
FOR-BLIT |  0.005959524
myplot

itzpr3d4t0r:
Other performance tests after the latest commit:
CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 1, SURFACES TO DRAW = 1000

CUMULATIVE TIME:
BLITS |  0.2810937001
BLITS_NORET |  0.2529978
UBLITS |  0.2454326
UBLITS_NORET |  0.2228538

MEAN:
BLITS |  0.000562187
BLITS_NORET |  0.000505996
UBLITS |  0.000490865
UBLITS_NORET |  0.000445708
Best Performance | UBLITS_NORET | 14.23 % better than the others' mean: 0.000519683
___
CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 1, SURFACES TO DRAW = 100

CUMULATIVE TIME:
BLITS |  0.0295115
BLITS_NORET |  0.0262459
UBLITS |  0.026741
UBLITS_NORET |  0.0240724

MEAN:
BLITS |  5.9023e-05
BLITS_NORET |  5.2492e-05
UBLITS |  5.3482e-05
UBLITS_NORET |  4.8145e-05
Best Performance | UBLITS_NORET | 12.46 % better than the others' mean: 5.4999e-05
___
CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 1, SURFACES TO DRAW = 10

CUMULATIVE TIME:
BLITS |  0.0039487
BLITS_NORET |  0.0034617
UBLITS |  0.0033592
UBLITS_NORET |  0.0032676

MEAN:
BLITS |  7.897e-06
BLITS_NORET |  6.923e-06
UBLITS |  6.718e-06
UBLITS_NORET |  6.535e-06
Best Performance | UBLITS_NORET | 8.97 % better than the others' mean: 7.179e-06
___
CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 0, SURFACES TO DRAW = 1000

CUMULATIVE TIME:
BLITS |  0.583424
BLITS_NORET |  0.5606737
UBLITS |  0.5600451
UBLITS_NORET |  0.5438119

MEAN:
BLITS |  0.001166848
BLITS_NORET |  0.001121347
UBLITS |  0.00112009
UBLITS_NORET |  0.001087624
Best Performance | UBLITS_NORET | 4.27 % better than the others' mean: 0.001136095
___
CONST = True, CALLS_PER_TEST = 10, TESTS = 500, BLEND_MODE = 0, SURFACES TO DRAW = 100

CUMULATIVE TIME:
BLITS |  0.519022
BLITS_NORET |  0.4912638
UBLITS |  0.4894999
UBLITS_NORET |  0.4720876

MEAN:
BLITS |  0.001038044
BLITS_NORET |  0.000982528
UBLITS |  0.000979
UBLITS_NORET |  0.000944175
Best Performance | UBLITS_NORET | 5.57 % better than the others' mean: 0.000999857
___
CONST = False, CALLS_PER_TEST = increases linearly, TESTS = 500, BLEND_MODE = 0, SURFACES TO DRAW = 100

CUMULATIVE TIME:
BLITS |  12.2607434
BLITS_NORET |  12.0529438
UBLITS |  11.8520416001
UBLITS_NORET |  11.275765

MEAN:
BLITS |  0.024521487
BLITS_NORET |  0.024105888
UBLITS |  0.023704083
UBLITS_NORET |  0.02255153
Best Performance | UBLITS_NORET | 6.47 % better than the others' mean: 0.024110486
___
Performance here is worse than blits(), apparently using pygame.BLEND_MULT causes a performance loss
CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 3, SURFACES TO DRAW = 1000

CUMULATIVE TIME:
BLITS |  0.2752464
BLITS_NORET |  0.2447393
UBLITS |  0.3064143
UBLITS_NORET |  0.2827593

MEAN:
BLITS |  0.000550493
BLITS_NORET |  0.000489479
UBLITS |  0.000612829
UBLITS_NORET |  0.000565519
Best Performance | BLITS_NORET | 15.06 % better than the others' mean: 0.00057628
___

itzpr3d4t0r:
Other performance tests after the latest commit:
---
CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 3, SURFACES TO DRAW = 1000

CUMULATIVE TIME:
BLITS |  0.3410083
BLITS_NORET |  0.3062588
UBLITS |  0.3258221
UBLITS_NORET |  0.2959275

MEAN:
BLITS |  0.000682017
BLITS_NORET |  0.000612518
UBLITS |  0.000651644
UBLITS_NORET |  0.000591855

Best Performance | UBLITS_NORET | 8.77 % better than the others' mean: 0.000648726
___
CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 3, SURFACES TO DRAW = 100

CUMULATIVE TIME:
BLITS |  0.0373023
BLITS_NORET |  0.0338321
UBLITS |  0.0342632
UBLITS_NORET |  0.0331091

MEAN:
BLITS |  7.4605e-05
BLITS_NORET |  6.7664e-05
UBLITS |  6.8526e-05
UBLITS_NORET |  6.6218e-05

Best Performance | UBLITS_NORET | 5.76 % better than the others' mean: 7.0265e-05
___
CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 3, SURFACES TO DRAW = 10

CUMULATIVE TIME:
BLITS |  0.0045026
BLITS_NORET |  0.0043775
UBLITS |  0.0044717
UBLITS_NORET |  0.0040913

MEAN:
BLITS |  9.005e-06
BLITS_NORET |  8.755e-06
UBLITS |  8.943e-06
UBLITS_NORET |  8.183e-06

Best Performance | UBLITS_NORET | 8.07 % better than the others' mean: 8.901e-06
___

CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 0, SURFACES TO DRAW = 1000

CUMULATIVE TIME:
BLITS |  0.5302416
BLITS_NORET |  0.5012424999
UBLITS |  0.5009465
UBLITS_NORET |  0.4938086001

MEAN:
BLITS |  0.001060483
BLITS_NORET |  0.001002485
UBLITS |  0.001001893
UBLITS_NORET |  0.000987617

Best Performance | UBLITS_NORET | 3.33 % better than the others' mean: 0.00102162
___
CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 0, SURFACES TO DRAW = 100

CUMULATIVE TIME:
BLITS |  0.0572716001
BLITS_NORET |  0.056282
UBLITS |  0.057264
UBLITS_NORET |  0.0606113

MEAN:
BLITS |  0.000114543
BLITS_NORET |  0.000112564
UBLITS |  0.000114528
UBLITS_NORET |  0.000121223

Best Performance | BLITS_NORET | 3.6 % better than the others' mean: 0.000116765
___
CONST = True, CALLS_PER_TEST = 10, TESTS = 500, BLEND_MODE = 0, SURFACES TO DRAW = 1000

CUMULATIVE TIME:
BLITS |  5.258288
BLITS_NORET |  5.1224673
UBLITS |  5.2613103
UBLITS_NORET |  5.0556682

MEAN:
BLITS |  0.010516576
BLITS_NORET |  0.010244935
UBLITS |  0.010522621
UBLITS_NORET |  0.010111336

Best Performance | UBLITS_NORET | 3.04 % better than the others' mean: 0.010428044
___
# These last two are important because you can clearly see that when the BLEND_MODE = 0, all the performance benefits decrease greatly... I'll include the plots so you can have a clearer view

CONST = False, CALLS_PER_TEST = increases linearly, TESTS = 500, BLEND_MODE = 0, SURFACES TO DRAW = 100
myplot

CUMULATIVE TIME:
BLITS |  13.6892292
BLITS_NORET |  13.0456055
UBLITS |  13.6542669
UBLITS_NORET |  13.2652559

MEAN:
BLITS |  0.027378458
BLITS_NORET |  0.026091211
UBLITS |  0.027308534
UBLITS_NORET |  0.026530512

Best Performance | BLITS_NORET | 3.62 % better than the others' mean: 0.027072501
___
CONST = False, CALLS_PER_TEST = increases linearly, TESTS = 500, BLEND_MODE = 1, SURFACES TO DRAW = 100
myplot

CUMULATIVE TIME:
BLITS |  6.7484185
BLITS_NORET |  6.0577261999
UBLITS |  6.0320673
UBLITS_NORET |  5.4255563

MEAN:
BLITS |  0.013496837
BLITS_NORET |  0.012115452
UBLITS |  0.012064135
UBLITS_NORET |  0.010851113
Best Performance | UBLITS_NORET | 13.6 % better than the others' mean: 0.012558808
___

itzpr3d4t0r:
I'd like to also include a test that draws onto a small surface for an extreme case scenario:
Draws onto a very small surface DEST = pygame.surface.Surface((10, 10)), using the same code as before.

CONST = True, CALLS_PER_TEST = 1, TESTS = 500, BLEND_MODE = 0, SURFACES TO DRAW = 200

CUMULATIVE TIME:
BLITS |  0.0212731
BLITS_NORET |  0.0186363
UBLITS |  0.0209098
UBLITS_NORET |  0.0156267

MEAN:
BLITS |  4.2546e-05
BLITS_NORET |  3.7273e-05
UBLITS |  4.182e-05
UBLITS_NORET |  3.1253e-05
Best Performance | UBLITS_NORET | 22.92 % better than the others' mean: 4.0546e-05

src_c/surface.c
```diff
@@ -2105,6 +2109,193 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+    PyObject *blitsequence;
+    PyObject *iterator = NULL;
+    PyObject *item = NULL;
+    PyObject *special_flags = NULL;
+    PyObject *ret = NULL;
+    PyObject *retrect = NULL;
+    Py_ssize_t itemlength;
+    int doreturn = 1;
+    int bliterrornum = 0;
+    static char *kwids[] = {"blit_sequence", "special_flags", "doreturn",
```
novialriptide, line 2136: I think it's better to have the function return something at all times instead of having a ``doreturn`` kwarg

novialriptide (changes requested):
🥳 Nice PR, found one problem though.

src_c/surface.c
itzpr3d4t0r: Yeah this is something blits() implements and i thought keeping the same use woud be the way. If it's there is because you can use the returned rect list later for other things. We can either remove this functionality entirely or keep it. I'm not sure...

itzpr3d4t0r: > 🥳 Nice PR, found one problem though.

and thank you, first ever and i've never seen anything about the Python ABI before, so this is just messing up with stuff for many hours :)

novialriptide: Oh, I didn't know that. Then I think we should keep it. I'll keep this unresolved incase there's opposition.

itzpr3d4t0r: >I think it's better to have the function return something at all times instead of having a `doreturn` kwarg even if the performance is slightly boosted without the return. If we want to keep it, we should rename it to `return_rects`

Yes but wouldn't that be the same as now? Do you mean that we would need to create something like an empty list on every call and return that? but in that case you'd still need some flag to decide whether to fill up the list or not.

itzpr3d4t0r: The renaming of `doreturn` to `return_rects` alone would be nice for readability reasons tho. The problem is that `doreturn` is used for `blits()` aswell, so you'd have to change that too be level. Thats not for this PR tho.

src_c/surface.c
```diff
@@ -2105,6 +2109,193 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+    }
+
+    if (special_flags) {
+        if (!pg_IntFromObj(special_flags, &the_args)) {
+            bliterrornum = BLITS_ERR_MUST_ASSIGN_NUMERIC;
+            goto bliterror;
+        }
+    }
+
+    iterator = PyObject_GetIter(blitsequence);
```
itzpr3d4t0r, line 2160: An idea. We are accepting everything that is an iterable. I've seen that you can create a `PySequence_Fast`: https://docs.python.org/3/c-api/sequence.html#c.PySequence_Fast
Couldn't we convert the blit_sequence to `PySequence_Fast` (or maybe to a PySequence_Fast_ITEMS) and use PySequence_Fast functions like `PySequence_Fast_GET_ITEM` to get items from the sequence? Would't that be faster? We'd need a `Py_ssize_t counter` and then we'd have a normal for(counter=0; counter < sequence_length; counter++) to access the items?

itzpr3d4t0r: You'd have to assume just a list or a tuple is passed as `blit_sequence` function parameter, and so you'd restrict the function's purpose even more, but you'd get (possibly?) even more performance because at that point you'd just be looping over an array ( if you use `PySequence_Fast_ITEMS` ). This function could very well become a very optimized version of blits() if you're willing to restrict the function's behaviour, but would probably take some time in that case.

Starbuck5: But then you lose the possible speed benefit of using a generator. Since currently you can pass in "a list" without ever having to construct it. You could have it both ways by trying to use PySequence_Fast and then using an iterator strategy? Not sure what would be worth it with performance.

Also if you're not in the pygame community discord you should join, we have a "contributing" channel for pygame development. https://discord.com/invite/ZuB2RySPRJ

itzpr3d4t0r: > 
Yes i'll ty and join the discord.
Yeah something more along the "both ways" would be ideal. I could try it and post some tests... 
and i've ran a quck test and there seems to be no real difference whether you construct a list on the fly or use a generator, there may still be a difference if you don't construct the list at all just like you said:
WITH IMPLICIT GENERATOR
blits: 0.014914000001226668
ublits: 0.012218600000778679

WITH EXPLICIT GENERATOR
blits: 0.012284700000236626
ublits: 0.012045800000123563

WITH LIST:
blits: 0.012100500000087777
ublits: 0.011486600000353064

Code i used:
```python
print("WITH IMPLICIT GENERATOR")
print("blits: ", end="")
print(timeit(f"win.blits((IMAGE, POSITIONS[i]) for i in range(NUMSURFS))", number=100, globals=GLOB))
print("ublits: ", end="")
print(timeit(f"win.ublits((IMAGE, POSITIONS[i]) for i in range(NUMSURFS))", number=100, globals=GLOB))

print("\nWITH EXPLICIT GENERATOR")
print("blits: ", end="")
print(timeit(f"win.blits(((IMAGE, POSITIONS[i]) for i in range(NUMSURFS)))", number=100, globals=GLOB))
print("ublits: ", end="")
print(timeit(f"win.ublits(((IMAGE, POSITIONS[i]) for i in range(NUMSURFS)))", number=100, globals=GLOB))

print("\nWITH LIST:")
print("blits: ", end="")
print(timeit(f"win.blits([(IMAGE, POSITIONS[i]) for i in range(NUMSURFS)])", number=100, globals=GLOB))
print("ublits: ", end="")
print(timeit(f"win.ublits([(IMAGE, POSITIONS[i]) for i in range(NUMSURFS)])", number=100, globals=GLOB))
```

src_c/surface.c
```diff
@@ -342,6 +344,8 @@ static struct PyMethodDef surface_methods[] = {
      DOC_SURFACEBLIT},
     {"blits", (PyCFunction)surf_blits, METH_VARARGS | METH_KEYWORDS,
      DOC_SURFACEBLITS},
+    {"ublits", (_PyCFunctionFast)surf_ublits, METH_FASTCALL,
```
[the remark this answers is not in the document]
Starbuck5: I believe it should be cast to a PyCFunction. Look at `blits` right above. Due to the VARARGS | KEYWORDS, the method is of of type `PyCFunctionWithKeywords`, yet it is cast to a `PyCFunction`.

The docs specify that the blocks in the PyMethodDef must be PyCFunctions. Does this make sense to me: no. But it must work, since that's how everything else does it.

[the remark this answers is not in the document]
itzpr3d4t0r: Updated it, now only build that fails is the first one for AppVeyor with pypy3.6

src_c/surface.c
itzpr3d4t0r: I've done more testing and came to the conclusion that using lists is overall faster than using generators, and that makes sense because when you call `PyIter_Next(iterator)`, you get the next element from the sequence, and in the case of generators there's extra operations to get the next value that can be arbitrarily complex. 

That being said I'd like to iterate over this idea of using PySequence_Fast and discuss why we should move to a simple C array of PyObject*. But first i'd like to respond to what you said:

> ...You could have it both ways by trying to use PySequence_Fast and then using an iterator strategy? Not sure what 
 would be worth it with performance.

- You can't have it both ways because `PySequence_Fast` objects can only be created converting a List or Tuple object. Therefore you cannot convert a generator object.
-  `PySequence_Fast` objects are intended for accessing the sequence's data fields directly, with either a PySequence_Fast_GET_ITEM or a PySequence_ITEM, so an iterator strategy could be overkill in this case.

Let me unpack what we have to do right now  vs. what we could have into two lists here

RIGHT NOW:
1. Have to declare, assign, manage and DECREF a `PyObject *iterator`
2. Check the validity of the blit sequence with `PyIter_Check` and `PySequence_Check`
4. call `PyIter_Next(iterator)` on every loop
5. `Py_DECREF(item);` on every loop because `PyIter_Next` gives a new reference
6. In the end we `Py_DECREF(iterator);`

INSTEAD WE COULD HAVE:
1. Have a `blit_sequence = PySequence_Fast(args[0], errcode)`
2. Convert the `blit_sequence ` into an array of the underlying sequence objects with `PySequence_Fast_ITEMS(blit_sequence)`
3. Loop over the array with `for (seq_counter = 0; seq_counter < sequencelength; seq_counter++)`, so no iterators
4. Have `item = f_blitsequence[seq_counter]` and just use it with no DECREF

## Positives
- We should have less code to manage and an easier to understand function
- We should get a modest performance uplift as we are no longer using iterators, but looping through an array
- If not we have a less noisy (meaning little negative performance spikes), and more reliable function

## Negatives
- Having to use lists or tuples

In my opinion, given that this function already simplifies the `blit_sequence `by only having pairs (Surface, dest), it only makes sense to have only lists or tuples, because even in the worst case scenario the user would have to write something like 
`surf.ublits([(img, pos) for pos in positions], 0, 1)`
instead of 
`surf.ublits(((img, pos) for pos in positions), 0, 1)`
and the same is true if a uses uses a function to return a generator. Simple generators are easy to substitute with list comprehensions and complex ones don't really fit this use case.
So i don't think that swapping two "( )" for a "[ ]" is really all that bad.

Starbuck5: You can have it both ways, you can have a code path for one and a code path for the other.

Generators "overhead" constructing the next element is irrelevant because the list would have to do that work as well.

> So i don't think that swapping two "( )" for a "[ ]" is really all that bad.

A small change in syntax, but a large change in what's actually happening.

Starbuck5: Although a quick personal test doesn't seem very favorable on generators. It beat the list comp, but not by a significant margin.

itzpr3d4t0r: can i get the code snippet so i can replicate this?

itzpr3d4t0r: > You can have it both ways, you can have a code path for one and a code path for the other.

code would start to get very long i'm just saying, because on one you'd have a while(item=PyIter_Next(blit_sequence))... and on the other a for loop over an array.

itzpr3d4t0r: > Although a quick personal test doesn't seem very favorable on generators. It beat the list comp, but not by a significant margin.

Did you use any blend flags over the default one?

itzpr3d4t0r: or are you just talking about difference between gen and list relative perf

test/blit_test.py
```diff
@@ -130,6 +130,87 @@ def blits(blit_list):
...
+
+        # tests for the doreturn parameter
+        self.assertRaises(ValueError, dst.ublits, blit_list, 0, "str")
+        self.assertRaises(ValueError, dst.ublits, blit_list, 0, [])
+        self.assertRaises(ValueError, dst.ublits, blit_list, 0, [1])
+        self.assertRaises(ValueError, dst.ublits, blit_list, 0, ())
+        self.assertRaises(ValueError, dst.ublits, blit_list, 0, (1,))
+
+        # tests for return value
+        self.assertEqual(dst.ublits(blit_list, 0, False), None)
```
itzpr3d4t0r, line 177: This test fails with a `TypeError: Source objects must be a Surface`. Problem is tricky because it really starts at line 159. there's some kind of undefined behaviour that transforms the `blit_list` into a infinite recursive list that's obviously not compatible with ublits. Can't really figure it out.

src_c/surface.c
```diff
@@ -2105,6 +2108,300 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+        goto on_error;
+    }
+
+    if (!dest) {
+        errornum = BLITS_ERR_DISPLAY_SURF_QUIT;
+        goto on_error;
+    }
+
+    /* Generator */
+    if (!PyList_Check(blitsequence) && !PyTuple_Check(blitsequence)) {
```
ankith26, line 2162: Not sure whether this block of code is worth the extra complexity here. 
`PySequence_Fast` should be able to handle most, if not all iterators

src_c/surface.c
```diff
@@ -2105,6 +2108,300 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+        Py_DECREF(iterator);
+        if (PyErr_Occurred()) {
+            Py_XDECREF(ret);
+            return NULL;
+        }
+    }
+    /* List or Tuple */
+    else if (PySequence_Check(blitsequence)) {
+        tmpblitseq =
+            PySequence_Fast(args[0], UBLITS_SERR_LISTORTUPLE_REQUIRED);
```
ankith26, line 2264: This can return `NULL` with error, that should be handled (as I said in my previous review comment that we should be using this block for all iterables, so error handling becomes relevant here)

src_c/surface.c
itzpr3d4t0r: Yes I'm aware that it's not ideal. I tried using the return value from PySequence_Fast but I couldn't get it to correctly distinguish between a generator and a tuple/list. I'll try again because I agree this must change.

itzpr3d4t0r: And i don't get how PySequence_Fast  treats errors because the docs say it 

> raises [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) with m as the message text. Returns NULL on failure.
So what does the function do? shouldn't the program end with a TypeError? i don't get what happens

ankith26: The function basically returns `NULL`, and on our side we don't need to set a new error again. We only need to pass the message string to the function as argument `m`

itzpr3d4t0r: The big problem is that PySequence_Fast works with all iterables so you can't get whether it's a generator or a list/tuple being passed.

src_c/surface.c
itzpr3d4t0r: I've added a check to this conversion that leads to an error if it fails.

test/blit_test.py
itzpr3d4t0r: Ok i found uou that something corrupts only the first surface of the first tuple transforming the sequence to:
`[(<NULL>, (80.36447577626312, 241.40381563470999)), ...]` (i put the dots but everything past this first element is ok)
by doing
```python
try:
    win.ublits(ublits_sequence, -999, 1)
except TypeError:
    print(ublits_sequence)
```
and if i do
```python
try:
    win.ublits(ublits_sequence, 12.3324, 1)
except ValueError:
    print(ublits_sequence)
```
the sequence is ok just like in the tests.

itzpr3d4t0r: I think i fixed it with the latest commit!

ankith26:
Nice, all builds on this PR seem to pass now, except the pypy3.6 windows build.
That means this PR could potentially be blocked on #3231 being merged (another alternative is to put pypy3.6 specific workarounds, that I'm not really in support of)

itzpr3d4t0r:
> Nice, all builds on this PR seem to pass now, except the pypy3.6 windows build. That means this PR could potentially be blocked on #3231 being merged (another alternative is to put pypy3.6 specific workarounds, that I'm not really in support of)

Want me to update the docs and add a note that specifically says that?

MyreMylar (approved):
LGTM, just wanted a little expansion to the documentation for users to better understand while reading the docs why this exists over `.blits()`.

docs/reST/ref/surface.rst
```diff
@@ -154,6 +154,27 @@
...
+      :param blend_flags: the flag(s) representing the blend mode used for each surface
+      :param return_rects: if ``True``, return a list of rects of the areas changed,
+         otherwise return ``None``
+
+      :returns: a list of rects of the areas changed if ``return_rects`` is
+         ``True``, otherwise ``None``
+      :rtype: list or None
+
+      .. ## Surface.ublits ##
+
```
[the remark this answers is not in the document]
itzpr3d4t0r: Yeah sure working on that.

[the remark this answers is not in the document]
itzpr3d4t0r: > Having a stricter use case and treating every surface the same (by drawing the entire surfaces all with the same blend_flag), ublits() can achieve better performance where blits() cannot. Further optimizations are applied if the blit_sequence parameter is a list or a tuple (using one of them is recommended).

is this ok to you?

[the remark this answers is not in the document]
itzpr3d4t0r: Or maybe something more along these lines

> This function only accepts a sequence of (surf, dest) pairs and a single blend_flag.
>                 This knowledge is exploited to achieve faster iteration over the sequence and
>                 therefore better performance whereas blits() cannot. Further optimizations are applied
>                 if the blit_sequence parameter is a list or a tuple (using one of them
>                 is recommended).

Starbuck5:
I think this is failing on pypy3.6 more because it's python 3.6 than because it's pypy, because theoretically this should not work on 3.6 at all (FASTCALL).

I would like it to use preprocessor directives to compile a function that just raises a NotImplementedError on python 3.6 and below. (And of course document that this is 3.7+ only)

You can use `#if PY_VERSION_HEX < 0x03070000`.

itzpr3d4t0r:
> 

could you do id for me? It would take ages if i did that because i don't know how the preprocessor directive interacts with the list of method definitions...

src_c/surface.c
```diff
@@ -2105,6 +2116,313 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+        }
+        if (!(PyIter_Check(blitsequence))) {
+            errornum = BLITS_ERR_SEQUENCE_REQUIRED;
+            goto on_error;
+        }
+        iterator = PyObject_GetIter(blitsequence);
+        if (!iterator) {
+            Py_XDECREF(ret);
+            return NULL;
+        }
```
Starbuck5, line 2184: The type checking -> execution paths in this PR make me nervous.

This pseudocode is how it works now
```
if generator:
    if iterator:  # <- what generator wouldn't be an iterator?
       PyIter strategy
    else:
        fail

else if sequence:
    if sequence_fast_support: # <- this is not the same as if sequence, there are more sequences than lists or tuples
        sequence_fast strategy
    else:
        fail

else:
    fail
```

I think it should be something like
```
if sequence_fast_support:
    sequence_fast strategy
else if itererator:
    iter strategy
else
    fail

Comment (Starbuck5) [reply src_c/surface.c:2184]: And by if sequence_fast_support, I just mean checking whether `PySequence_Fast` returned NULL. (Also you can use PyErr_Clear to clear the resulting set error if it is NULL).

Comment (itzpr3d4t0r) [reply src_c/surface.c:2184]: Yeah i see the problem... but PySequence_Fast somehow corrupts the generator, invalidating the iteration

Comment (itzpr3d4t0r) [reply src_c/surface.c:2184]: if you put it before the generator

Comment (itzpr3d4t0r) [reply src_c/surface.c:2184]: i'll tackle this tomorrow

File: src_c/surface.c [hunk 9]
```diff
@@ -2105,6 +2116,313 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+}
+#else
+
+static PyObject *
+surf_ublits(pgSurfaceObject *self, PyObject *_null)
+{
+    return RAISE(PyExc_NotImplementedError,
+                 "ublits is not compatible with Python 3.6 or older");
+}
+
```
ankith26, line 2423: I'm not a fan of this, I believe it's best if all functions are consistently defined on all versions we support.

I propose that *only* on python 3.6, we have a wrapper `METH_VARARGS` function that calls the `METH_FASTCALL` function (which should be defined regardless of python version, with full implementation), and this wrapper should be like

Here is some (untested) code that demonstrates what I'm thinking

```c
#if PY_VERSION_HEX < 0x03070000
static PyObject *
surf_ublits_wrapper(PyObject *self, PyObject *args)
{
    Py_ssize_t nargs = PyTuple_GET_SIZE(args);
    PyObject *const *vector_args = PyMem_New(PyObject *const *, nargs);
    if (!vector_args) {
        return PyErr_NoMemory();
    }

    PyObject *ret = surf_ublits(self, vector_args, nargs);
    PyMem_Free(vector_args);
    return ret;
}
#endif
```

Starbuck5: I agree with you in theory, but in practice, how many people even use `blits`?

And how many of those people actually update pygame and read the release notes? And how many of those people haven't updated their python installation since 2016/2017? I don't see this affecting [m]any people.

I think it would be fine to do this for a specialized function on an unsupported version of python that we should drop support for soon.

ankith26: I see where you are coming from, but in this case the little patch I suggested above is simple enough (atleast I think so) so for the sake of consistency and cleanliness IMHO there is no harm in using it. Obviously this codepath would miss out on the performance enhancements of 3.7 `METH_FASTCALL`, but I think that's fine for the same reasons you give (very little number of users affected by this)

Better to have a slower specialised function with just some more extra lines of code, than one that won't even work and give that one user a hard time

itzpr3d4t0r: So what i stick with the current version?

itzpr3d4t0r: > I'm not a fan of this, I believe it's best if all functions are consistently defined on all versions we support.
> 
> I propose that _only_ on python 3.6, we have a wrapper `METH_VARARGS` function that calls the `METH_FASTCALL` function (which should be defined regardless of python version, with full implementation), and this wrapper should be like
> 
> Here is some (untested) code that demonstrates what I'm thinking
> 
> ```c
> #if PY_VERSION_HEX < 0x03070000
> static PyObject *
> surf_ublits_wrapper(PyObject *self, PyObject *args)
> {
>     Py_ssize_t nargs = PyTuple_GET_SIZE(args);
>     PyObject *const *vector_args = PyMem_New(PyObject *const *, nargs);
>     if (!vector_args) {
>         return PyErr_NoMemory();
>     }
> 
>     PyObject *ret = surf_ublits(self, vector_args, nargs);
>     PyMem_Free(vector_args);
>     return ret;
> }
> #endif
> ```

this version you propose won't even be possible anymore, as the function now always returns None.

ankith26: I don't see how it's not possible? Afterall python `None` is still `Py_None` (which is a `PyObject *`) at the C level

Starbuck5: > as the function now always returns None.

Did you forget to push a commit or something? The function still has optional returns on this end.

itzpr3d4t0r: It's not finished 100% i didn't push it because i'm having an exam today.

itzpr3d4t0r: i mean it should work in theory, i'll push it and see how it goes

itzpr3d4t0r: i did something strange with i hope it worked

src_c/surface.c
itzpr3d4t0r: if you do this 
```
if PySequence_Fast(blitsequence, errmsg): <- is always true with generators/lists/tuples
    sequence_fast strategy
else if PyIter_Check(blitsequence):
    iter strategy
else
    fail
```

best i can do is 
```
if PySequence_Fast(blitsequence, errmsg) and !PyGen_Check(blitsequence):
    sequence_fast strategy
else if PyIter_Check(blitsequence):
    iter strategy
else
    fail
```

src_c/surface.c
itzpr3d4t0r: Can we consider this resolved?

src_c/surface.c
itzpr3d4t0r: mosty importantly i noticed this function segfaults if strings are passed as blit_sequence

src_c/surface.c
Starbuck5: You didn't resolve the issue though?

src_c/surface.c
Starbuck5: > Would lead to too much nesting because at that point you have to check for each type inside the first if clause and implement each strategy all inside that block. Not to say you'd have to DECREF nmultiple times if the type doesn't exactly match a list or tuple as calling PySequence_Fast includes assigning its return to a variable.

- For strings, check that before and exit on wrong type.
- "implement each strategy in each block" -> no, if it enters the first block, use sequence_fast.
- "you'd have to DECREF nmultiple times" -> no, if it gives you a valid reference, use it, otherwise it will give you NULL.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -70,6 +70,13 @@ class Surface:
         ],
         doreturn: Union[int, bool] = 1,
     ) -> Union[List[Rect], None]: ...
+    def ublits(
+            self,
+            blit_sequence: Union[
+                Sequence[Tuple[Surface, Union[Coordinate, RectValue]]],
+                List[Tuple[Surface, Union[Coordinate, RectValue]]]],
+            blend_flags: int
```
ankith26, line 78:
```suggestion
            blend_flags: int, 
            /,
```
This will mark that the function takes positional-only args

docs/reST/ref/surface.rst
```diff
@@ -154,6 +154,31 @@
...
+      and a blend_flags parameter that applies to every surface being drawn. All surfaces are fully drawn.
+      It needs at minimum a sequence of (source, dest).
+
+      :param blit_sequence: a sequence of (Surface, dest)
+      :param blend_flags: the flag(s) representing the blend mode used for each surface
+
+      :returns: always returns ``None``
+      :rtype: None
+
+      .. note:: ublits is compatible only with Python 3.7 and above
```
ankith26, line 172: This note can be removed now

test/blit_test.py
```diff
@@ -130,6 +130,78 @@ def blits(blit_list):
         if PRINT_TIMING:
             print(f"Surface.blits generator: {t1 - t0}")
 
+    def test_ublits(self):
+        import sys
+
+        if sys.version_info < (3, 7):
+            return
```
ankith26, line 137: This can be hopefully removed now

buildconfig/stubs/pygame/surface.pyi
itzpr3d4t0r: Pushed commit to address the remaining issues

ankith26: This one does not seem to be resolved yet

itzpr3d4t0r: oh yeah i thought i put / but ended up only putting a , lol

Issue #3252, opened 2022-06-24
Unit test for the validity of sysfonts


Starbuck5:
I would like there to be a unit test that verifies pygame's system font code by making sure every font possibility returned by pygame.font.get_fonts returns a valid and absolute path with pygame.font.match_fonts. Maybe use os.path.isabs and os.path.isfile.

Sysfont test file: https://github.com/pygame/pygame/blob/main/test/sysfont_test.py

SSS-Says-Snek:
I'll add the sysfont unittest

PR #3266, opened 2022-06-30
Add SysFont pygame.font.get_fonts() verification test

SSS-Says-Snek:


ankith26 (commented):
Thanks for contributing to pygame, snek! 🥳 🎈 

I just realised, there is a unit test in `font_test.py` that does something similar to this, so issue #3252 is actually redundant sadly

However that does not mean you need to close this PR, you can make updates to the existing function `test_match_font_all_exist`, as it is missing a test for `os.path.isfile`

SSS-Says-Snek:
Ah alright, I'll add that when I get home

test/font_test.py
```diff
@@ -118,7 +118,10 @@ def test_match_font_all_exist(self):
         for font in fonts:
             path = pygame_font.match_font(font)
             self.assertFalse(path is None)
-            self.assertTrue(os.path.isabs(path))
+            self.assertTrue(
+                os.path.isabs(path) and
+                os.path.isfile(path)
+            )
```
ankith26, line 124: Super minor thing, black is going to complain about the formatting here (lint CI will fail if I re-run it).
This needs to be on one line because it is within the line limit

ankith26 (approved):
This looks good to go to me! Thanks for the first time contribution! 🎉

MyreMylar (approved):
LGTM 👍

PR #3268, opened 2022-06-30
Add an AVX2 version of BLEND_PREMULTIPLIED.

MyreMylar:
In some quick testing this seems to be between 2-6x faster than the current SSE2 version depending on the size of the blit. Obviously more and, better testing would be useful. Perhaps @itzpr3d4t0r would like to have a look with those fancy performance graphs they've bene using for ublits.

In my testing larger blits have bigger performance improvements as they can take better advantage of the parallelism.

--- 

This is intended for version 2.1.4 assuming the AVX2 versions of the special blendmodes in 2.1.3 don't have any issues once released into userland.

If I have a chance to fix up the ancient `.convert_alpha()` PR I think a good demo might be showing the performance difference between a bunch of alpha blits of the same image in 2.1.3 using the normal alpha blending, and then the exact same image converted to pre-multiplied format and blitted using this flag for hopefully some decent performance wins.

### DIFF
--- a/src_c/simd_blitters_avx2.c
+++ b/src_c/simd_blitters_avx2.c
@@ -1030,8 +1030,8 @@ blit_blend_premultiplied_avx2(SDL_BlitInfo *info)
                         *dstp = _mm_cvtsi128_si32(mm_dst);
                     }
 
-                    srcp = (Uint32 *)srcp256 + srcskip;
-                    dstp = (Uint32 *)dstp256 + dstskip;
+                    srcp += srcpxskip;
+                    dstp += dstpxskip;
                 },
                 n, pre_8_width);
         }
@@ -1086,8 +1086,8 @@ blit_blend_premultiplied_avx2(SDL_BlitInfo *info)
                 },
                 n, post_8_width);
         }
-        srcp += srcskip;
-        dstp += dstskip;
+        srcp = (Uint32 *)srcp256 + srcskip;
+        dstp = (Uint32 *)dstp256 + dstskip;
     }
 }
 #else
--- a/src_c/simd_blitters_avx2.c
+++ b/src_c/simd_blitters_avx2.c
@@ -941,12 +941,10 @@ blit_blend_premultiplied_avx2(SDL_BlitInfo *info)
     Uint32 *srcp = (Uint32 *)info->s_pixels;
     int srcskip = info->s_skip >> 2;
     int srcpxskip = info->s_pxskip >> 2;
-    int srceightpxskip = 8 * srcpxskip;
 
     Uint32 *dstp = (Uint32 *)info->d_pixels;
     int dstskip = info->d_skip >> 2;
     int dstpxskip = info->d_pxskip >> 2;
-    int dsteightpxskip = 8 * dstpxskip;
 
     int pre_8_width = width % 8;
     int post_8_width = (width - pre_8_width) / 8;
@@ -1032,8 +1030,8 @@ blit_blend_premultiplied_avx2(SDL_BlitInfo *info)
                         *dstp = _mm_cvtsi128_si32(mm_dst);
                     }
 
-                    srcp += srcpxskip;
-                    dstp += dstpxskip;
+                    srcp = (Uint32 *)srcp256 + srcskip;
+                    dstp = (Uint32 *)dstp256 + dstskip;
                 },
                 n, pre_8_width);
         }
@@ -1085,8 +1083,6 @@ blit_blend_premultiplied_avx2(SDL_BlitInfo *info)
 
                     srcp256++;
                     dstp256++;
-                    srcp += srceightpxskip;
-                    dstp += dsteightpxskip;
                 },
                 n, post_8_width);
         }
--- a/src_c/simd_blitters_avx2.c
+++ b/src_c/simd_blitters_avx2.c
@@ -928,3 +928,178 @@ blit_blend_rgb_min_avx2(SDL_BlitInfo *info)
 }
 #endif /* defined(__AVX2__) && defined(HAVE_IMMINTRIN_H) && \
           !defined(SDL_DISABLE_IMMINTRIN_H) */
+
+#if defined(__AVX2__) && defined(HAVE_IMMINTRIN_H) && \
+    !defined(SDL_DISABLE_IMMINTRIN_H)
+void
+blit_blend_premultiplied_avx2(SDL_BlitInfo *info)
+{
+    int n;
+    int width = info->width;
+    int height = info->height;
+
+    Uint32 *srcp = (Uint32 *)info->s_pixels;
+    int srcskip = info->s_skip >> 2;
+    int srcpxskip = info->s_pxskip >> 2;
+    int srceightpxskip = 8 * srcpxskip;
+
+    Uint32 *dstp = (Uint32 *)info->d_pixels;
+    int dstskip = info->d_skip >> 2;
+    int dstpxskip = info->d_pxskip >> 2;
+    int dsteightpxskip = 8 * dstpxskip;
+
+    int pre_8_width = width % 8;
+    int post_8_width = (width - pre_8_width) / 8;
+
+    /* if either surface has a non-zero alpha mask use that as our mask */
+    Uint32 amask = info->src->Amask | info->dst->Amask;
+
+    __m256i *srcp256 = (__m256i *)info->s_pixels;
+    __m256i *dstp256 = (__m256i *)info->d_pixels;
+
+    __m128i mm_src, mm_dst, mm_zero, mm_alpha, mm_sub_dst, mm_ones;
+    __m256i mm256_src, mm256_dst, mm256_shuff_mask_A, mm256_shuff_mask_B,
+        mm256_src_shuff, mm256_dstA, mm256_dstB, mm256_ones, mm256_alpha,
+        mm256_shuff_alpha_mask_A, mm256_shuff_alpha_mask_B;
+
+    mm_zero = _mm_setzero_si128();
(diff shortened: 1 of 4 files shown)


itzpr3d4t0r:
Super cool! Yeah Sure i can take a look with my "fancy graphs" lol. Also i've never used premultiplied as a blend... how does it work? is it some sort of BLEND_MULT?

MyreMylar:
> Super cool! Yeah Sure i can take a look with my "fancy graphs" lol. Also i've never used premultiplied as a blend... how does it work? is it some sort of BLEND_MULT?

It is basically a better version of a standard alpha blend (it produces better results when blending two surfaces that both have an alpha channel) that requires some setup/forethought. You need to start with a surface where all the colour channels have been pre-multiplied by the alpha channel so:

```
#FF000090 becomes #09000009 (or (255,0,0144) becomes (144,0,0144)
```

This is pretty easy to setup when creating surfaces but a little trickier when dealing with images unless you have  a program that will let you save out into a pre-multiplied alpha format. I believe a lot of mobile graphics use pre-multiplied alpha images.

There is a nice little primer on the formulas from Microsoft here:

https://microsoft.github.io/Win2D/WinUI3/html/PremultipliedAlpha.htm


I thin that straight alpha is easier to understand - but gives worse results, and pre-multiplied alpha is usually faster and better, but more complicated to think about and understand.

itzpr3d4t0r:
Oh ok now i'm interested, so for now i can't use premultiplied alpha for images i've loaded with pygame.image.load? how does it interact with .convert_apha()?

MyreMylar:
> Oh ok now i'm interested, so for now i can't use premultiplied alpha for images i've loaded with pygame.image.load?

You can if the image is in premultiplied format - some rare image programs will let you save 'normal' images like this with the pre-multiplication done on export. But yes, it is not particularly easy.

> how does it interact with `.convert_apha()`? Could we implement something like a `.convert_premultiplied()` to do what you said?

I have an old stalled PR still in the list to do exactly this - I hope to get it up and running again over the summer.

itzpr3d4t0r:
I'll post some results here so if people want to examine them they easily can:

This one is a test that examines how it fairs with increasingly bigger images, starts at 5x5 and goes up by an increment of 5 pixels for every test. Here it's 220 tests so 220*5=up to 1100x1100 pixels images.
image
CONST = True, CALLS_PER_TEST = 1, TESTS = 220

Blit premultiplied is: 256.4867% faster than alpha blit

CUMULATIVE TIME:
ALPHA BLIT |  0.1084297
PREMULTIPLIED |  0.0304162

MEAN:
ALPHA BLIT |  0.000492862
PREMULTIPLIED |  0.000138255
___
A constant test on an image that's 1000x1000px:
image

CONST = True, CALLS_PER_TEST = 1, TESTS = 220

Blit premultiplied is: 335.7932% faster than alpha blit

CUMULATIVE TIME:
ALPHA BLIT |  0.2332714
PREMULTIPLIED |  0.053528

MEAN:
ALPHA BLIT |  0.001060325
PREMULTIPLIED |  0.000243309
___
This is a test on a 1920x1080px image
image

CONST = True, CALLS_PER_TEST = 1, TESTS = 220

Blit premultiplied is: 156.0069% faster than alpha blit

CUMULATIVE TIME:
ALPHA BLIT |  0.3647999
PREMULTIPLIED |  0.1424961

MEAN:
ALPHA BLIT |  0.001658181
PREMULTIPLIED |  0.00064771
___
And last on a 20x20 image
image

Blit premultiplied is: 69.9007% faster than alpha blit

CUMULATIVE TIME:
ALPHA BLIT |  0.0004448
PREMULTIPLIED |  0.0002618

MEAN:
ALPHA BLIT |  2.022e-06
PREMULTIPLIED |  1.19e-06
___

MyreMylar:
Thanks for the tests! The graphs seem to show about what I expected - the lines all look like improvements and it gets better with bigger surfaces.

I guess I need to get to work on that `surface.premultiply_alpha()` function. I'm thinking of simplifying it a bit from my original design (which was adding it to `.convert_alpha()` as an argument). As the pre-multiply blend is only really good looking (you need 8bit alpha for a good looking blend which is the appearance advantage of using premultiplied blending) and performant (you need 32 bit to use AVX and SSE2) with 32bit RGBA surfaces anyway - I was thinking of just adding it as a separate method on a surface that errors/warnings out doing nothing if you call it on a surface that is not in the correct format.

That way the implementation should be easier, and we could always expand it to cover more surface formats later if anybody comes up with compelling use cases for an 8/16 bit surface to pre-multiply alpha.

itzpr3d4t0r:
Oh god what did i start here... And it's curious to see that the alpha blit (in the last graph) goes so much up and down... is there a problem with even/uneven surfaces with a .convert_alpha() surface?

MyreMylar:
> is there a problem with even/uneven surfaces with a .convert_alpha() surface? 

Yes. I mean, it's more of a known issue. 

It's fundamentally faster to blit surfaces whose dimensions are a multiple of 8/4/2 (depending on the blit) because we can bunch multiple pixels together when the dimensions are a multiple of whatever the algorithm is able to scale up to in pixels-per-loop. Generally we then have to drop down to a one-pixel-at-time algorithm to finish off uneven blits. Right now the more recent SIMD implementations do this dropping down more efficiently than the older one used for regular alpha blending.

Eventually I want to go back over the regular alpha blend SIMD implementations and improve them where possible (and add AVX versions). They are just more annoyingly complicated than all the other blits because they need to support all the various pygame options (whole surface alpha mixed with per pixel alpha) and be backwards compatible to how blending worked (slightly weirdly) in the days of SDL 1.

Starbuck5 (commented):
If you rebased this branch against main you could get the moving `blit_blend_premultiplied_sse2` diff out of this PR.

src_c/simd_blitters_avx2.c
```diff
@@ -928,3 +928,174 @@ blit_blend_rgb_min_avx2(SDL_BlitInfo *info)
...
+    int srcskip = info->s_skip >> 2;
+    int srcpxskip = info->s_pxskip >> 2;
+
+    Uint32 *dstp = (Uint32 *)info->d_pixels;
+    int dstskip = info->d_skip >> 2;
+    int dstpxskip = info->d_pxskip >> 2;
+
+    int pre_8_width = width % 8;
+    int post_8_width = (width - pre_8_width) / 8;
+
```
Starbuck5, line 951: It's this way in every stride switching blitter, but it's the same to do

```c
int pre_8_width = width % 8;
int post_8_width = width / 8;
```

Because of integer division.

src_c/simd_blitters_avx2.c
```diff
@@ -928,3 +928,174 @@ blit_blend_rgb_min_avx2(SDL_BlitInfo *info)
...
+                    srcp += srcpxskip;
+                    dstp += dstpxskip;
+                },
+                n, pre_8_width);
+        }
+        srcp256 = (__m256i *)srcp;
+        dstp256 = (__m256i *)dstp;
+        if (post_8_width > 0) {
+            /*8 pixels at a time, need to use shuffle to get everything
+                lined up - see mul for an example*/
```
Starbuck5, line 1042: It's crazy that the current SSE version is 1 pixel at a time and the AVX version is 8 pixels at a time, since the AVX registers are only twice as wide.

src_c/simd_blitters_avx2.c
```diff
@@ -928,3 +928,174 @@ blit_blend_rgb_min_avx2(SDL_BlitInfo *info)
...
+            LOOP_UNROLLED4(
+                {
+                    mm256_src = _mm256_loadu_si256(srcp256);
+                    mm256_dst = _mm256_loadu_si256(dstp256);
+
+                    /* insert 8 pixel at a time blend here */
```
Starbuck5, line 1048: Old comment? :smile:

src_c/simd_blitters_avx2.c
```diff
@@ -928,3 +928,174 @@ blit_blend_rgb_min_avx2(SDL_BlitInfo *info)
...
+
+    __m128i mm_src, mm_dst, mm_zero, mm_alpha, mm_sub_dst, mm_ones;
+    __m256i mm256_src, mm256_dst, mm256_shuff_mask_A, mm256_shuff_mask_B,
+        mm256_src_shuff, mm256_dstA, mm256_dstB, mm256_ones, mm256_alpha,
+        mm256_shuff_alpha_mask_A, mm256_shuff_alpha_mask_B;
+
+    mm_zero = _mm_setzero_si128();
+    mm_ones = _mm_set_epi64x(0x0000000000000000, 0x0001000100010001);
+
+    mm256_shuff_mask_A =
```
Starbuck5, line 966: Maybe these mask values would be clearer as binary numbers `0b`, since the shuffle instructions that use them use the bits directly. It would make this section of code quite bigger though...

src_c/simd_blitters_avx2.c
```diff
@@ -928,3 +928,174 @@ blit_blend_rgb_min_avx2(SDL_BlitInfo *info)
...
+                    mm256_src = _mm256_loadu_si256(srcp256);
+                    mm256_dst = _mm256_loadu_si256(dstp256);
+
+                    /* insert 8 pixel at a time blend here */
+
+                    /* do everything A set first */
+                    mm256_dstA =
+                        _mm256_shuffle_epi8(mm256_dst, mm256_shuff_mask_A);
+                    mm256_src_shuff =
+                        _mm256_shuffle_epi8(mm256_src, mm256_shuff_mask_A);
```
Starbuck5, line 1054: I would appreciate some additional comments in the algorithm, especially for what state the shuffles put things into, since they are hard to decode.

MyreMylar:
Moved to a new branch.

PR #3272, opened 2022-07-02
Fix controller build on SDL < 2.0.9

Starbuck5:
Continuation from #3238 

Cliffnotes version:
We (Emc and I) realized pygame won't build on 2.0.8 (except on gcc ubuntu runner, because weird optimizations), because of this controller function.
It took me like 6 different attempts, over 2 weeks, and many hours of CI time, but I finally have something that works on my computer and the test runners.

This will support my other PR to get pygame to better realize its legacy SDL version support, and it will support Emc's FRect PR (by allowing him to easily test on SDL 2.0.8, which is relevant to FRect).

### DIFF
--- a/src_c/controllercompat.c
+++ b/src_c/controllercompat.c
@@ -6,7 +6,7 @@
 #include <SDL.h>
 
 /*
- * Compat thing for controller (see https://github.com/pygame/pygame/pull/3270)
+ * Compat thing for controller (see https://github.com/pygame/pygame/pull/3272)
  */
 int
 PG_GameControllerRumble(SDL_GameController *gamecontroller,
--- a/src_c/controllercompat.c
+++ b/src_c/controllercompat.c
@@ -1,13 +1,11 @@
 /*
- * This file should not exist. If you exist in the future and are wondering if
- * this file is cool, it is not.
+ * Compat file specifically for controller (see https://github.com/pygame/pygame/pull/3272)
+ * we had trouble using pgcompat.c, because it would be included twice, or it would lead to
+ * flakiness on compilers, because of threads...
  */
 
 #include <SDL.h>
 
-/*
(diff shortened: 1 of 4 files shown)


src_c/controllercompat.c
```diff
@@ -0,0 +1,22 @@
+/*
+ * This file should not exist. If you exist in the future and are wondering if
+ * this file is cool, it is not.
+ */
```
ankith26, line 4: Some elaboration would help someone existing in the future to fix this 😉 

I think we need to clarify here why this file is being used, instead of `pgcompat.c`

src_c/controllercompat.c
```diff
@@ -0,0 +1,22 @@
...
+int
+PG_GameControllerRumble(SDL_GameController *gamecontroller,
+                        Uint16 low_frequency_rumble,
+                        Uint16 high_frequency_rumble, Uint32 duration_ms)
+{
+#if SDL_VERSION_ATLEAST(2, 0, 9)
+    return SDL_GameControllerRumble(gamecontroller, low_frequency_rumble,
+                                    high_frequency_rumble, duration_ms);
+#else
+    return -1;
```
ankith26, line 20: Just a nitpick (I think I mentioned this in a review on the older PR)

I know we don't use the SDL error string in the cython call of this function, but for consistency sake both branches of this compat function should be setting it. This will be useful if in the future any other code uses this compat function and expects to handle an SDL error message

MyreMylar (approved):
OK. Seems fine to me once ankith's issues are resolved.

It seems to me like pygame's mixing of C and Cython ends up causing issues (or maybe it is just Cython in general) and we'd probably be better off switching to 100% C one day - but that is a 'nice to have' future goal, while this fix seems to solve the issues today.

ankith26 (approved):
Thanks👍

PR #3274, opened 2022-07-03
Update sprite_texture.py

beefy:
I got this error without this change: `AttributeError: 'pygame._sdl2.video.Image' object has no attribute 'flip_x'`


Starbuck5:
Thanks for the PR.

This is because the attributes have been renamed since the last release of pygame. The example is actually correct.

You can run the example as `python -m pygame.examples.sprite_texture` and it will work, and the corresponding source for that would be with your pygame installation's files.

PR #3276, opened 2022-07-04
Add premul_alpha method to surface

MyreMylar:
This should simplify working with the BLEND_PREMULTIPLIED blend mode.

This is an adaption and simplification of an old PR I had lying around. If I get some time before the 2.1.4 release, I will add an AVX2 version of the pre-multiplication - though it's not critical as this shouldn't normally be a critical path function like blit. 

I also now kind of want to work up a little BEFORE and AFTER demo program showing how to use the same assets in the old school (normal alpha blending, blits()) way versus the new-ish (premul_alpha(), BLEND_PREMULTIPLED, ublits() - maybe threaded blitting?) way.


docs/reST/ref/surface.rst
```diff
@@ -902,4 +902,26 @@
 
       .. versionadded:: 1.9.2
 
+   .. method:: premul_alpha
```
Starbuck5, line 905: Maybe it should be more explicit about what you are and aren't supposed to do with a premultiplied surface.

Thoughts:
- Pixels are actually mutated, using get_at() won't compensate, you shouldn't draw onto these surfaces?
- Can you blit a premultiplied surface onto another premultiplied surface? I don't think that would work.
- When blitting this, you need to use the BLEND_PREMULTIPLIED flag to get correct results.
- This is for high speed of repeated blitting of static (unchanging) surfaces, right?

MyreMylar:
I added a bit more information to the documentation. I have a task to make a larger tutorial on pre-multiplied alpha as well to try and explain how to use it with some worked examples.

Starbuck5:
If you call premul-alpha on a surface twice, will the first second result be different from the first?

MyreMylar:
Yes, the output surface will be different because you will multiply by the
alpha again.

I considered flagging a surface after alpha multiplying, but since we can't
do that for loaded surfaces, or constructed in pygame by the user surfaces
(there is no difference between a surface with the intent that it is alpha
multiplied and an identical one where this isn't intended), it seemed like
it would just make it more confusing.

Alpha premultiplication is essentially all about the users blitting intent
rather than something inherent to the pixels.

itzpr3d4t0r:
> Yes, the output surface will be different because you will multiply by the
alpha again.

Maybe add this to the docs?

test/surface_test.py
```diff
@@ -3766,6 +3766,82 @@ def test_fill_blend_rgba(self):
                 dst.fill(fill_color, special_flags=getattr(pygame, blend_name))
                 self._assert_surface(dst, p, f", {blend_name}")
 
+    def test_surface_premul_alpha(self):
+        """Ensure that .premul_alpha() works correctly"""
+        pygame.display.init()
+        try:
+            pygame.display.set_mode((640, 480))
```
Starbuck5, line 3773: Why set up a display and quit it for this test? I haven't seen anything yet that suggests a display is necessary to do this operation.

docs/reST/ref/surface.rst
```diff
@@ -902,4 +902,41 @@
...
+
+      Surfaces without an alpha channel cannot use this method and will return an error if you use
+      it on them. It is best used on 32 bit surfaces (the default on most platforms) as the blitting
+      on these surfaces can be accelerated by SIMD versions of the pre-multiplied blitter.
+
+      In general pre-multiplied alpha blitting is faster then 'straight alpha' blitting and produces
+      superior results when blitting an alpha surface onto another surface with alpha - assuming both
+      surfaces contain pre-multiplied alpha colours.
+
+      .. versionadded:: 2.1.4
```
Starbuck5, line 936: illume's probably going to want one of these to be added: https://github.com/pygame/pygame/pull/3511

Starbuck5 (approved):
Left a couple minor things to do before merge, but otherwise I couldn't find anything wrong in the PR. Nice job!

docs/reST/ref/surface.rst
MyreMylar: Ah, I missed the resolution of this debate. Thanks, I've added the line.

test/surface_test.py
MyreMylar: I had no idea, when I first looked at it - then I remembered that this functionality was once bolted on to `surface.convert_alpha()` which does look at the display surface and the test was clearly originally cloned from that functions test and not correctly updated once it became its own thing.

I have done so now, thanks!

MightyJosip (approved):

Issue #3267, opened 2022-06-30
Add support for Windows on ARM


RockLakeGrass:
I tried to build it on a Windows on ARM64 device, but it does not show support.
I looked at the SDL dependencies, except for those not supported by mpg123, which all have ARM64 versions and can be installed by vcpkg.
I would like Pygame to add support for windows on arm64.

ankith26:
I'm +1 for the proposal, but the implementation would probably not be as straightforward.
This will involve some rewriting in our buildconfig, because libsdl.org does not provide arm binaries

I don't think any of our regular contributors have an ARM64 Windows machine, but since you have one, we would be happy if you are interested in submitting some patches for this

RockLakeGrass:
> I'm +1 for the proposal, but the implementation would probably not be as straightforward. This will involve some rewriting in our buildconfig, because libsdl.org does not provide arm binaries
> 
> I don't think any of our regular contributors have an ARM64 Windows machine, but since you have one, we would be happy if you are interested in submitting some patches for this

I don't know where Pygame's SDL2 relies on the source code because I try to build SDL2 from the source code and get a different file than Pygame's own

Starbuck5:
I'm not really sure what this issue entails.

> it does not show support.

We don't have prebuilts for ARM, and SDL doesn't have prebuilts for ARM. But it should still work if you manage to navigate through compiling everything. (Although that would be quite a hassle).

It's also interesting that Python doesn't even provide prebuilt Python for Windows ARM yet. They're only starting with 3.11.

RockLakeGrass:
I have successfully built the sdl2 arm64 Version (except mpg123), but the following errors are reported when compiling Pygame:
buildconfig\obj\win64\scale_mmx.obj : Module computer type "x64" conflicts with target computer type "ARM64"
I don't know where the "scale_mmx.obj" file came from so I can't rebuild it for WoA

> They're only starting with 3.11.

You can see my repository in https://github.com/RockLakeGrass/Windows-on-ARM64-Toolchain , which contains MSI construction of Python 3.7~3.10.

Starbuck5:
> except mpg123

Well that's fine. It just won't support mp3. Nothing in pygame directly touches mpg123, but SDL_mixer uses it. SDL_mixer can also use other libraries as their mp3 backend, so you may have better luck porting one of those.

> buildconfig\obj\win64\scale_mmx.obj : Module computer type "x64" conflicts with target computer type "ARM64"
[...]

PR #3278, opened 2022-07-05
Add support for Windows ARM64

RockLakeGrass:
I have a prebuilt file compiled for ARM64 and I don't know how to upload it, so it's temporarily stored in my repository


Starbuck5:
Hey, I'm quite impressed with your effort on this issue.

I'm not sure why you closed #3276 already.

I think the goal to improve this PR is to minimize the code changes needed, and I have a few suggestions.

Firstly, for SCALE_MMX. See this bit of code: https://github.com/pygame/pygame/blob/main/src_c/scale.h#L32-L36. I think changing this so it just isn't defined on Windows ARM64 would be better than dealing with it further down the line throughout the code.

Secondly, I don't want a custom SETUP file. We already have three, and it's a pain when you have to change one and guess if changing the others will work the same way. I think what you should do is set up a custom SDL arm prebuilt repo that emulates the main SDL prebuilt repo. That way you can use the normal SETUP generation process, just switch out the URLs in the buildconfig to download from the new repo on Windows ARM64.

FYI: Since you haven't previously committed to this repository, Github won't run tests without member approval. That's why your tests may take a while to run. Since one of us has to press the button

FYI2: The best resource for pygame contributors is the `contributing` channel on the pygame discord (https://discord.com/invite/ZuB2RySPRJ), that's the best place to get help from other contributors.

RockLakeGrass:
> I'm not sure why you closed https://github.com/pygame/pygame/pull/3276 already.

I think I've finished the transplant. Do I need to reopen it?

> for SCALE_MMX. See this bit of code

I'll try it later, but I think my solution is the fastest way to get pyGame to support win-arm.

> I don't want a custom SETUP file

Secondly, Since the Python official MSVC module does not support win-arm64, I can only customize the setup file. I can support win arm by modifying the code of MSVC module, but I don't think this is a good method.

> the channel on the pygame discord 

At last, I'm sorry I can't open Pygame discord

Starbuck5:
Oh sorry, wrong issue tag. I meant #3267.

That issue is titled "Add support for Windows on ARM," and that's what this PR does. So I don't see why you would close the issue before the PR is merged.

> I'll try it later, but I think my solution is the fastest way to get pyGame to support win-arm.

I care more about maintainability than convenience of implementation, but I'm confident this will be more convenient to implement as well, with less code change.

> Secondly, Since the Python official MSVC module does not support win-arm64, I can only customize the setup file. I can support win arm by modifying the code of MSVC module, but I don't think this is a good method.

I don't see the relevance of MSVC to what you're talking about. It's just two different ways of generating the SETUP file. I don't want to do it the way you have it currently.

> At last, I'm sorry I can't open Pygame discord

"can't" ? If you don't have discord and don't want to get it, no worries, I just wanted to make sure you knew the deal.

Starbuck5:
The other reason to make a generic SDL Windows ARM prebuilt repo is because other people will be able to use it. Maybe other SDL2 developers could use that, or other libraries that depend on SDL2.

RockLakeGrass:
> So I don't see why you would close the issue before the PR is merged.

I've reopened it.

>  don't see the relevance of MSVC to what you're talking about

```Traceback (most recent call last):
  File "C:\Users\m1781\Downloads\pygame-2.1.2\pygame-2.1.2\setup.py", line 359, in <module>
    buildconfig.config.main(AUTO_CONFIG)
  File "C:\Users\m1781\Downloads\pygame-2.1.2\pygame-2.1.2\buildconfig\config.py", line 225, in main
    deps = CFG.main(**kwds)
  File "C:\Users\m1781\Downloads\pygame-2.1.2\pygame-2.1.2\buildconfig\config_win.py", line 511, in main
    return setup_prebuilt_sdl2(prebuilt_dir)
  File "C:\Users\m1781\Downloads\pygame-2.1.2\pygame-2.1.2\buildconfig\config_win.py", line 471, in setup_prebuilt_sdl2
    DEPS.configure()
  File "C:\Users\m1781\Downloads\pygame-2.1.2\pygame-2.1.2\buildconfig\config_win.py", line 336, in configure
    from . import vstools
  File "C:\Users\m1781\Downloads\pygame-2.1.2\pygame-2.1.2\buildconfig\vstools.py", line 11, in <module>
    compiler.initialize()
  File "C:\Complier\Python310-arm64\lib\distutils\msvc9compiler.py", line 347, in initialize
    raise DistutilsPlatformError("--plat-name must be one of %s" %
distutils.errors.DistutilsPlatformError: --plat-name must be one of ('win32', 'win-amd64')
```
During the generation of the Setup file, the msvc9compiler module was called, which does not support win-arm64.

> "can't" ? 

It seems that due to national policies, this website is blocked, and the use of VPN in our country is not allowed.

> The other reason to make a generic SDL Windows ARM prebuilt repo is because other people will be able to use it. Maybe other SDL2 developers could use that, or other libraries that depend on SDL2.

I don't understand what you're saying. Can't I prebuilt the SDL?

RockLakeGrass:
> See this bit of code: https://github.com/pygame/pygame/blob/main/src_c/scale.h#L32-L36.

The "scale.h" has been modified to suit your suggestion.It's really easier in this way.

src_c/alphablit.c
```diff
@@ -265,7 +265,7 @@ SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
                     break;
                 }
                 case PYGAME_BLEND_ADD: {
-#if !defined(__EMSCRIPTEN__)
+#if !defined(__EMSCRIPTEN__) && !defined(_M_ARM64)
 #if SDL_BYTEORDER == SDL_LIL_ENDIAN
```
Starbuck5, line 269: What happens if you don't add this?

Starbuck5:
> I don't understand what you're saying. Can't I prebuilt the SDL?

So many of your SETUP changes are to accommodate the fact that the prebuilts now live in a folder called prebuilt-ARM64. But it doesn't have to be like that. You could package the prebuilts so they look like pygame is expecting, just at a different website. And that way it would also be helpful to other people trying to use Windows ARM. For example, the SDL_image releases site: https://www.libsdl.org/projects/SDL_image/release/.

> During the generation of the Setup file, the msvc9compiler module was called, which does not support win-arm64.

Ugh. I'd prefer not to have another SETUP file in there, there must be a better way to do this.

src_c/alphablit.c
RockLakeGrass: If you don't, he will get a LINK error and will not be able to parse the symbols.
Because these symbols cannot be used on Windows arm.

Starbuck5: I'm curious what symbols?

RockLakeGrass:
I don't know why there was an error on Ubuntu after modifying "scale.h". In that case, I had to change the code back to the previous version, at least that's correct.

> just at a different website

First, I don't have permission to make changes to the SDL's publishing file. Secondly, since I need to upload the SDL of ARM64, why don't I just pack it into one file for Pygame to download? Then, I made some changes to Pygame in the SDL package, such as generating lib files, which may not be allowed officially in SDL

> there must be a better way to do this

See this bit of code https://github.com/pygame/pygame/blob/main/buildconfig/config_win.py#L420. If you don't customize the SETUP file, you will get an error on this line, which I think is unavoidable during the automatic generation of the SETUP file.

src_c/alphablit.c
RockLakeGrass: Some symbols about RGB functions, such as blit_blend_rgb_sub_sse2.

After my analysis, the reasons for incompatibility are as follows: First, MSVC ARM64 does not support SSE, which is a fact. In this file, there is already a way to be compatible with ARM64, even with sse2neon, in the GitHub repository of this header file, I found an (https://github.com/DLTcollab/sse2neon/issues/384), which means that this header file does not support MSVC.

RockLakeGrass:
> Ubuntu sdist / build (ubuntu-22.04) (pull_request) Failing after 29s — build (ubuntu-22.04)

I don't know why it's so amazing that it doesn't go wrong with the same code before

Starbuck5:
That fails every time.

RockLakeGrass:
https://github.com/pygame/pygame/pull/3278/commits/a5b27aa625f2f3cdd5e438593b46c575d00f4aba
This is the right time.

RockLakeGrass:
SDL components in prebuilt files have been downloaded as separate files, and the reason why the latest version of SDL was used is that libmpg123 was discarded in the latest version, which allows SDLmixer to be fully ported to Windows ARM64.

RockLakeGrass:
From https://ci.appveyor.com/project/pygame/pygame/builds/44099669/job/92rn46mco7ow5rxi
```
Installing Python 3.10.0 64 to C:\Python310-x64...
Downloading https://www.python.org/ftp/python/3.10.0/python-3.10.0-amd64.exe...
Exception calling "DownloadFile" with "2" argument(s): "An exception occurred during a WebClient request."
At line:12 char:5
+     (New-Object Net.WebClient).DownloadFile($downloadUrl, $exePath)
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException
 
Installing...
```
How can there be problems when downloading Python from this website

src_c/alphablit.c
```diff
@@ -310,7 +310,7 @@ SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
                     break;
                 }
                 case PYGAME_BLEND_SUB: {
-#if !defined(__EMSCRIPTEN__)
+#if !defined(__EMSCRIPTEN__) && !defined(_M_ARM64)
```
ankith26, line 313: I feel these checks can be removed.

In theory (untested), the neon path (neon being ARM SIMD) should also work on ARM64 windows.

In this same file, we have
```c
#if !defined(PG_ENABLE_ARM_NEON) && defined(__aarch64__)
// arm64 has neon optimisations enabled by default, even when fpu=neon is not
// passed
#define PG_ENABLE_ARM_NEON 1
#endif
```
somewhere at the top. I doubt `__aarch64__` is defined on MSVC, so I propose that this should become
```c
#if !defined(PG_ENABLE_ARM_NEON) && (defined(__aarch64__) || defined(_M_ARM64))
// arm64 has neon optimisations enabled by default, even when fpu=neon is not
// passed
#define PG_ENABLE_ARM_NEON 1
#endif
```

src_c/transform.c
```diff
@@ -1606,7 +1606,13 @@ surf_set_smoothscale_backend(PyObject *self, PyObject *args, PyObject *kwargs)
     if (!PyArg_ParseTupleAndKeywords(args, kwargs, "s", keywords, &type))
         return NULL;
 
-#if defined(SCALE_MMX_SUPPORT)
+#if defined(_M_ARM64)
+    st->filter_type = "GENERIC";
+    st->filter_shrink_X = filter_shrink_X_ONLYC;
+    st->filter_shrink_Y = filter_shrink_Y_ONLYC;
+    st->filter_expand_X = filter_expand_X_ONLYC;
+    st->filter_expand_Y = filter_expand_Y_ONLYC;
```
ankith26, line 1614: I think (but not sure) that this should not need special-casing.

ankith26 (commented):
Gotta say, really impressive for a first time PR, so thanks!

PS: This PR might take a while to review

Starbuck5:
Wow, big changes since I last looked! I like it.

However, I'm surprised you went back from changing the definition of SCALE_MMX_SUPPORT (https://github.com/pygame/pygame/blob/main/src_c/scale.h#L32-L36) like you did before. That seems more elegant?

src_c/alphablit.c
RockLakeGrass: https://github.com/DLTcollab/sse2neon/issues/384
Sse2neon does not support MSVC.

RockLakeGrass:
> However, I'm surprised you went back from changing the definition of SCALE_MMX_SUPPORT (https://github.com/pygame/pygame/blob/main/src_c/scale.h#L32-L36) like you did before. That seems more elegant?

If you write like last time, you cannot pass format-lint-code-check.

I see your submission, which updates the SDL, and I want to wait until SDL releases the latest official version and you update all SDL components before I migrate the latest version of SDL. Because Windows ARM64 is still using the preview version.

Starbuck5:
> it cannot pass format-lint-code-check.

Run it through clang-format and it will format it so it passes.

ankith26 (commented):
The code looks fine on a skim through.

I see you have put up your own SDL ARM64 releases, this is impressive work, I'd prefer to take this a step further and upstream this (ie hopefully get official SDL support), doing this would be helpful for more SDL users, and would also centralise the maintenance efforts needed to continue supporting compiling on ARM Windows.

Thanks again for contributing!

RockLakeGrass:
>  I'd prefer to take this a step further and upstream this (ie hopefully get official SDL support)

I made this [request](https://github.com/libsdl-org/SDL/issues/5882) in the SDL repository, but in my opinion, the official will not support Windows arm, and even if it does, it will only be supported in the new version. After testing, the latest sdl_image cannot be used in Pygame (it can be compiled but cannot pass the test)

illume (approved):
👍 thank you very much.

I'm going to continue this in another branch / PR.

PR #3283, opened 2022-07-07
rblit(), a raw blit only function

itzpr3d4t0r:
Follows #3324.
# Rationale
The blit function is the core function for blitting surfaces and just like blits(), it's a Swiss army knife of a function. It gives you freedom to choose whether you want to draw a portion of the surface, choose a blend flag and get the blit rectangle as return value. We need something more specific and single purpose, that does only one thing, blitting a surface onto another.
# Solution
Having a rblit() function where "r" stands for raw, that draws an entire surface (meaning area=None) and doesn't create and return a new rect.
# Benefits 

- **Higher performance**
You know you'll always draw the entire surface and won't have to create a return rect
- **Ease of use** 

```Python
# instead of writing
surf.blit(img, pos, None, pygame.BLEND_ADD)
# or
surf.blit(img, pos, special_flags=pygame.BLEND_ADD)
# you would have
surf.rblit(img, pos, pygame.BLEND_ADD)
```
- **End user now has choice**
By having a general use function and a more specific one we can better fit the user's needs.
- **Compatibility with blit**
Modifying blit isn't a thing, too many projects would be directly affected. Having a different function will enable current projects to be compatible and recieve the benefits.
# Negatives
I'll be using METH_FASTCALL as a method type. This will enable a faster calling convention but will require positional arguments to be passed.
- **Python 3.6 users will be able to use rblit() but with smaller benefits**

# First performance results
First results show a performance benefit that ranges from 6-10% up to 30-35%. This is assuming the Surface to blit isn't too big. With big images performance lines up with blit().

### DIFF
--- a/docs/reST/ref/surface.rst
+++ b/docs/reST/ref/surface.rst
@@ -130,6 +130,33 @@
 
       .. ## Surface.blit ##
 
+   .. method:: rblit
+
+      | :sl:`draw one full image onto another, with no rect being returned`
+      | :sg:`rblit(source, dest, special_flags=0) -> None`
+
+      Draws a source Surface onto this Surface. The draw can be positioned with
+      the dest argument. The dest argument can either be a pair of coordinates representing the position of
+      the upper left corner of the blit or a Rect, where the upper left corner of the rectangle will be used as the
+      position for the blit. The size of the destination rectangle does not
+      effect the blit.
+
+      Pixel alphas will be ignored when blitting to an 8 bit Surface.
+
+      For a surface with colorkey or blanket alpha, a blit to self may give
+      slightly different colors than a non self-blit.
+
+      :returns: always returns ``None``
+      :rtype: None
+
+      .. note:: Takes positional only arguments, with special_flags being optional.
+                rblit is a raw blit function with the intended use of the "blit" alone.
+                Use this whenever you have to draw the entire surface at a position without
+                needing a return rectangle of the area changed. Grants increased performance
+                compared to blit() depending on the surface size.
+
+      .. ## Surface.rblit ##
+
    .. method:: blits
 
       | :sl:`draw many images onto another`
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -172,6 +172,9 @@ surf_get_clip(PyObject *self, PyObject *args);
 static PyObject *
 surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds);
 static PyObject *
+surf_rblit(pgSurfaceObject *self, PyObject *const *args, Py_ssize_t nargs);
+PG_DECLARE_FASTCALL_FUNC(surf_rblit, pgSurfaceObject);
+static PyObject *
 surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds);
 static PyObject *
 surf_fill(pgSurfaceObject *self, PyObject *args, PyObject *keywds);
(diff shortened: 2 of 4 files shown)


MyreMylar:
I approve of this general idea, as it will give us a little more flexibility/room to add convenience features to `.blit()` (with some suggested recently involving rect positioning changes) with less concern that overall pygame drawing performance is going to come down even if `.blit()` itself gets a teeny bit slower from the specific change.

In reality pygame performance is constantly improving as we figure out way to do things better, but some times good changes to make functions easier to use and increase clarity of expression also slow things down a bit too. With a raw blit function we can branch in two directions with a still-fast, but friendly, regular `.blit()` for most use cases and a stripped down raw blit for certain performance intensive blittings that puts more restrictions on the input. If teaching it, I would have a guideline of 'start with blit() and then later, if you are having performance issues see if a raw blit would help' much like might be suggested with passing a rect list to `display.update()`. 

While it somewhat goes against the 'one obvious way to do things' principle of python, python itself is moving in this direction with CPython APIs like FASTCALL in the recent versions that are trying to improve python performance in critical areas. There are few more performance sensitive methods in pygame than blit (maybe colliderect?) so if we are going to break the one-obvious-way principle chasing performance then this is the place to do it.

Anyway, I wanted to put that out there in support as I have been a bit conflicted on the idea myself.

itzpr3d4t0r:
Yeah for example we also have clock.tick() and clock.tick_busy_loop(), two function that are very similar, yet one is more precise.
I've seen that python itself with its "adaptive intepreter" is moving in the direction of having specialized "functions" that only do 1 thing, and general purpose functions that are slower but more flexible. I think that somethimes less is more, as having just one functionality to expand will eventually make it a christmas tree of stuff, while keeping stuff clean snd separate just like you said will let us have a cleaner developement devoid of unnecesary slowdowns while decreasing concerns when adding new features to existing functions. It will indeed slow things down a bit because of the branch that we're creating but the positives seem to outweigh the negatives.

docs/reST/ref/surface.rst
```diff
@@ -130,6 +130,24 @@
 
       .. ## Surface.blit ##
 
+   .. method:: rblit
```
Starbuck5, line 133: Needs lack of keyword arguments documented.

src_c/surface.c
```diff
@@ -1879,6 +1890,77 @@ surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+    dest_rect.w = src->w;
+    dest_rect.h = src->h;
+
+    result = pgSurface_Blit(self, (pgSurfaceObject *)srcobject, &dest_rect,
+                            NULL, flags_numeric);
+
+    if (result != 0)
+        return RAISE(PyExc_TypeError,
+                     "Blit failed (probably the flag used does not exist)");
+
```
Starbuck5, line 1938: This should work like in surf_blit. If pgSurface_Blit failed, it already set the error message, so you just need to return NULL.

src_c/surface.c
```diff
@@ -1879,6 +1890,77 @@ surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+    PyObject *srcobject;
+    int result, flags_numeric = 0;
+
+    if (nargs == 0 || nargs > 3)
+        return RAISE(
+            PyExc_ValueError,
+            "Function requires at least 2 positional arguments in the order: "
+            "source, dest, (special_flags optional)");
+    else if (nargs == 3)
+        if (!pg_IntFromObj(args[2], &flags_numeric))
```
Starbuck5, line 1909: pg_IntFromObj also takes floats. You want PyLong_AsLong. See the documentation for the error checking requirements for that, it's more specific than most.

src_c/surface.c
```diff
@@ -1879,6 +1890,77 @@ surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+surf_rblit(pgSurfaceObject *self, PyObject *const *args, Py_ssize_t nargs)
+{
+    SDL_Surface *src, *dest = pgSurface_AsSurface(self);
+    SDL_Rect dest_rect, unused_rect;
+    SDL_Rect *src_rect;
+    PyObject *argpos;
+    PyObject *srcobject;
+    int result, flags_numeric = 0;
+
+    if (nargs == 0 || nargs > 3)
```
Starbuck5, line 1903: What if nargs is 1?

docs/reST/ref/surface.rst
itzpr3d4t0r: This i think i resolved by expanding the docs.

PR #3284, opened 2022-07-07
Add srcrect NULL check to pgSurface_Blit

MyreMylar:
For @itzpr3d4t0r to cherry pick.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -3644,7 +3644,7 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
             subsurface cannot be blitted to its owner because the
             owner is locked.
             */
-         dst->pixels == src->pixels &&
+         dst->pixels == src->pixels && srcrect != NULL &&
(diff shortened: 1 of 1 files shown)


ankith26:
What do we do about this PR?
Can we merge this one into main?

MyreMylar:
> What do we do about this PR? Can we merge this one into main?

Sure. I think it's pretty harmless. 

Basically `SDL_BlitSurface()` and `pygame_Blit()` let srcrect be NULL happily but a helper function (surface_do_overlap()) used as an old optimisation in this wrapper around those two functions is assuming `srcrect` is a valid dereferenceable pointer always. With this change if `srcrect` is NULL we don't go down that branch so `pgSurface_Blit()` also becomes safe to call with `srcrect` as NULL.

ankith26 (approved):
I did not test this locally, but the logic and explanation seems clear and simple. Thanks for the PR 🎉

zoldalma999 (approved):
Thanks!

PR #3293, opened 2022-07-11
Accelerate RGBA_SUB AVX

Starbuck5:
After days of fruitlessly looking around alphablit.c coming up with harebrained optimization ideas, I've finally got some good results.

This is a proof of concept run on a random AVX blitter code, but **I think my strategy here will be applicable to plenty of the other blitters.**

In my test code (100k, 300x301 blits) this took the runtime from 0.95 seconds to 0.76 seconds.

My first realization was that a for loop over the pixels of an image doesn't need to care about height vs width, it just has to iterate over pixels, which are laid out linearly in memory. This is great, because it reduces the amount of optimization "misses." The AVX code can run super fast on 8 pixel groups at a time, but at the end of every line it has to go a little slower to accommodate the non-divisible-by-8 pixels. So worst case scenario, the amount of missed optimization pixels is 7*height. But it doesn't have to go line by line, instead it can iterate over all the pixels, since they are continuous. This reduces the worst case scenario number of missed optimization pixels to 7.

This was responsible for the majority of the performance gain I saw, reducing the runtime of my test to 0.82 seconds.

My second realization -- more of a suspicion -- was that doing the "odd man out" pixels first would hurt performance because of memory alignment stuff that I don't really understand. Changing the order so that it did the accelerated pixels first then the non accelerated pixels second yielded the rest of the performance gain.

### DIFF
--- a/src_c/simd_blitters_avx2.c
+++ b/src_c/simd_blitters_avx2.c
@@ -482,8 +482,8 @@ blit_blend_rgba_sub_avx2(SDL_BlitInfo *info)
             n, aligned_8_length);
     }
 
-    srcp = srcp256;
-    dstp = dstp256;
+    srcp = (Uint32 *)srcp256;
+    dstp = (Uint32 *)dstp256;
 
     if (after_8_length > 0) {
         LOOP_UNROLLED4(
--- a/src_c/simd_blitters_avx2.c
+++ b/src_c/simd_blitters_avx2.c
@@ -446,21 +446,19 @@ void
 blit_blend_rgba_sub_avx2(SDL_BlitInfo *info)
 {
     int n;
-    int width = info->width;
-    int height = info->height;
+
+    /* using long rather than int out of abundance of caution */
+    long length = info->width * info->height;
+    int after_8_length = length % 8;
+    int aligned_8_length = length / 8;
 
     Uint32 *srcp = (Uint32 *)info->s_pixels;
-    int srcskip = info->s_skip >> 2;
     int srcpxskip = info->s_pxskip >> 2;
     int srceightpxskip = 8 * srcpxskip;
 
     Uint32 *dstp = (Uint32 *)info->d_pixels;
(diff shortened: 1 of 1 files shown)


Starbuck5:
This would be a great optimization if it worked, and I'm not sure why it's not. My constructed test case worked, but the tests fail.

I'm working on it still.

MyreMylar:
Probably missing out on the 'skip' ? Pixels in a blit are only guaranteed
linear in memory in a single row. The rest depends on blit position on the
destination surface and the source surface clip rect.

This optimization would work blitting an entire surface to another entire
surface at (0,0) when both surfaces are the same size. Otherwise there are
gaps in memory you have to jump over.

Might be misunderstanding, as I'm on phone walking about

Starbuck5:
You're 100% right. I'm continuing to experiment to see if there's anything I can put up for this PR, but I probably will end up closing it.

MyreMylar:
Since you've learned a bunch about the blitting now, if you want to poke
around something blitter related with definite available wins, the regular
alpha blitter needs refactoring to be more like the AVX ones in several
ways.

At the minute there are separate odd and even bits of code for SSE2 instead
of the multi pixel width switching to 1 pixel style used for the special
blend modes. The latter style is just better in my tests so a bunch of code
can probably be deleted and simplified there.

The SSE2 stuff for the normal blitter should really be moved to the SSE2 .c
file now that exists. Then we will have clear dividing lines for which code
units need which level of available hardware (SSE2 is everywhere now, but
this code is also used as NEON code on Arm which is not on early Raspberry
Pi's)

Also if you want to try actual AVX2 bits the shuffle intrinsic makes some
things a lot easier over SSE2.

Then there is switching our vendored Arm SIMD library from sse2neon to
SIMDe ( SIMD everywhere) which is under the MIT license and has a no-tests
GitHub repo somewhere and a little vendoring usage guide that explains a
few gotchas to watch out for. In theory this should port the AVX2 functions
to platforms like Arm Mac.

PR #3294, opened 2022-07-11
More efficiently track pixels in AVX blitters

Starbuck5:
I think I've finally found a win in the blitter code :smile:

In the multi-pixel bits, the pixel location is being tracked by two pointers on each surface. One "big pointer" that increments "one forward", which is many bytes, and then the "small pointer" that increments forward in increments of 8 to keep up with the big pointer.

However, the pointers are pointing at the same location, so we can take advantage of that to use the big pointer to update the small pointer after the loop. This reduces the amount of work in the hot loop.

I measured between 9 and 17% speedups from this, depending on how demanding the blit mode is.

I also saw that this strategy can be used in some SSE blitters, but I wanted to put this in for review first.

### DIFF
--- a/src_c/simd_blitters_avx2.c
+++ b/src_c/simd_blitters_avx2.c
@@ -4,8 +4,6 @@
 #include <immintrin.h>
 #endif /* defined(HAVE_IMMINTRIN_H) && !defined(SDL_DISABLE_IMMINTRIN_H) */
 
-#define __AVX2__
-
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
I'm no expert around these bits, but these changes make sense to me.
Considering this PR already has the approval of the alphablitter wizard :wink: you get a +1 from me as well.
Thankies! 🎈

PR #3295, opened 2022-07-12
Fix new static analyzer (VS 2022) warnings

### LINKED ISSUE #3251: Fix MSVC static analyzer warnings with VS2018 compiler
Related to #3231 . Hopefully once these warnings are fixed we can re-enable this build on the CI to catch any new ones.

To enable the static analyzer on a build you need to set the environment variable: `PG_MSVC_ANALYZE` to "1" and, to solve this issue, to be compiling with VS2018.

To do:

- [ ] Add a paste of the currently generated warnings and a checklist to solve each one in one/many PRs

ankith26:
PR to fix #3251

This commit was initially a part of #3231, where the issue was found, and I tested a fix on the same branch which has VS 2022 running on CI

Since the fix can be merged in independently, I cherry picked the commit into a new branch to create another PR that should be easier to merge in (this just fixes some error handling issues and also adds better error reporting in the case of some integers being out of the expected range)

(yes, this PR did turn out to be much smaller than I initially expected, I was hoping a whole wall of new warnings)

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1862,14 +1862,29 @@ pg_convert_to_uint16(PyObject *python_array, Uint16 *c_uint16_array)
         return 0;
     }
     for (i = 0; i < 256; i++) {
+        long ret;
         item = PySequence_GetItem(python_array, i);
+        if (!item) {
+            return 0;
+        }
         if (!PyLong_Check(item)) {
             PyErr_SetString(PyExc_ValueError,
                             "gamma ramp must contain integer elements");
             return 0;
         }
-        c_uint16_array[i] = (Uint16)PyLong_AsLong(item);
+        ret = PyLong_AsLong(item);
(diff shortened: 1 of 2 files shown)


MyreMylar:
Looks good, hopefully we can get appveyor chugging a long again.

MyreMylar (approved):

Starbuck5 (approved):
Looks fine to me. Thanks for the patch!

I wonder if the MSVC upgrade will do anything for the pygame binaries, like performance wise or size wise.

ankith26:
I don't expect significant changes: Wheel size of a 3.10 x64 wheel went from 10.3 MB (on main) to 10.4 MB (on the appveyor PR)
There is no visual difference in performance, but I've not benchmarked anything yet (although I'm pretty sure it won't make everything like 2x faster or so)

PR #3296, opened 2022-07-12
Move alphabit.c SSE2 blitters to main SSE2 file

Starbuck5:
First and simplest action to take following https://github.com/pygame/pygame/pull/3293#issuecomment-1180046254

I was tempted to do more (hence the branch title), but I want to be deliberate with this stuff, and keep it easy to review.

I moved all the SSE2 blitters from alphablit.c to simd_blitters_sse2.c. (And consolidated some of the #ifdefs action)


ankith26 (approved):
Just moving stuff without adding anything, LGTM! 🥳 🎉 
(this diff looks real scary hehe)

PR #3297, opened 2022-07-12
Implementation of the fblits() function.

itzpr3d4t0r:
Implements #3258. Follows #3324.
Remade the pull request because the other got to 54 commits.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -2145,7 +2145,7 @@ surf_fblits(pgSurfaceObject *self, PyObject *const *args, Py_ssize_t nargs)
     PyObject *srcobject = NULL, *argpos = NULL;
     PyObject *blitsequence;
     PyObject *item = NULL;
-    PyObject **f_blitsequence;
+    PyObject **blit_seq;
     Py_ssize_t itemlength, sequencelength, seq_counter;
     int errornum = 0;
     int result;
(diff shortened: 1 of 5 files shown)


src_c/surface.c
```diff
@@ -2097,6 +2101,215 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+            if (PySequence_Check(item)) {
+                itemlength = PySequence_Size(item);
+                if (itemlength != 2) {
+                    errornum = FBLITS_ERR_TUPLE_REQUIRED;
+                    goto on_error;
+                }
+            }
+            else {
+                errornum = FBLITS_ERR_TUPLE_REQUIRED;
+                goto on_error;
```
nthykier, line 2216: Missing `Py_DECREF(item);`?

src_c/surface.c
```diff
@@ -2097,6 +2101,215 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+        }
+    }
+    /* Generator */
+    else if (PyIter_Check(blitsequence)) {
+        while ((item = PyIter_Next(blitsequence))) {
+            if (PySequence_Check(item)) {
+                itemlength = PySequence_Size(item);
+                if (itemlength != 2) {
+                    errornum = FBLITS_ERR_TUPLE_REQUIRED;
+                    goto on_error;
```
nthykier, line 2211: Missing `Py_DECREF(item);`?

src_c/surface.c
```diff
@@ -2097,6 +2101,215 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+            else {
+                errornum = FBLITS_ERR_TUPLE_REQUIRED;
+                goto on_error;
+            }
+
+            /* (Surface, dest) */
+            srcobject = PySequence_ITEM(item, 0);
+            argpos = PySequence_ITEM(item, 1);
+
+            Py_DECREF(item);
```
nthykier, line 2223: As I understand `PySequence_ITEM`, "we" receive are now responsible for cleanup (`Py_DECREF`) of these parameters.

But even if I am wrong here and we get a borrowed reference, then we have another problem - namely that we are dereferencing the tuple containing these two items, which might invalidate the references we just pulled out of it.

Either way, I suspect something is wrong here.

Issue #3269, opened 2022-06-30
functions for drawing many shapes


itzpr3d4t0r:
I'll try and keep this as short as possible. We already have `lines()` and `aalines()` for drawing many lines on the screen. Why not have functions that draw many circles, rects and polygons on the screen? Just like we have a `blits()` for surfaces, we could have it for shapes too.

Just as an example, in `pygame.draw.polygon()`, in the case of drawing a hollow poly we call the `pygame.draw.lines` internally, which is a `METH_VARARGS | METH_KEYWORDS` (from python callable) function that has quite an overhead. Imagine having to call `pygame.draw.polygon()` N times from a for loop, how much overhead would that be?

So i'm proposing to have these functions:

- [ ] `pygame.draw.rects()`
- [ ] `pygame.draw.circles()` 
- [ ] `pygame.draw.polygons()`
- [ ] `pygame.draw.notcontiguouslines()` or `pygame.draw.nclines()`
- [ ] `pygame.draw.arcs()`
- [ ] `pygame.draw.ellipses()`

All somewhat similar in behaviour to `blits()` for surfaces.

PR #3299, opened 2022-07-13
pygame.draw.circles() function

itzpr3d4t0r:
Implements one of #3269. Follows #3324.
This one implements the `circles()` function for drawing many circles at once. The function's use case should be heterogeneous circle sequences, with both hollow and filled ones of different scales and colors.
Unlike `draw.circle()`, each circle in the sequence is described by either 3 or 4 parameters, indicating color, center_pos, radius, width(optional).

First test show a 20-55% performance benefit over a raw for loop with `draw.circle()`.

### DIFF
--- a/buildconfig/stubs/pygame/draw.pyi
+++ b/buildconfig/stubs/pygame/draw.pyi
@@ -1,4 +1,4 @@
-from typing import Optional, Sequence
+from typing import Optional, Sequence, Tuple
 
 from pygame.rect import Rect
 from pygame.surface import Surface
@@ -33,6 +33,10 @@ def circle(
     draw_bottom_left: bool = False,
     draw_bottom_right: bool = False,
 ) -> Rect: ...
+def circles(
+        surface: Surface,
+        draw_sequence: Sequence[Tuple[ColorValue, Coordinate, int, Optional[int]]], /
+) -> None: ...
 def ellipse(
     surface: Surface, color: ColorValue, rect: RectValue, width: int = 0
 ) -> Rect: ...
--- a/docs/reST/ref/draw.rst
+++ b/docs/reST/ref/draw.rst
@@ -207,6 +207,41 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
 
    .. ## pygame.draw.circle ##
 
+.. function:: circles
+
+   | :sl:`draws circles on a surface`
+   | :sg:`circles(surface, draw_sequence) -> None`
+
+   Draws a sequence of circles on the given surface.
+
+   :param surface: the surface to draw on
+   :param draw_sequence: a sequence composed of (color, center, radius, width) where:
+
+      - color: color to draw with, the alpha value is optional if using a tuple ``(RGB[A])``,
+
+      - center: center point of the circle as a sequence of 2 ints/floats, e.g. ``(x, y)``
+
+      - radius: radius of the circle, measured from the ``center`` parameter, nothing will
+              be drawn if the ``radius`` is less than 1
+
+      - width(optional): indicates the thickness of the circle (default=0)
+                         | if ``width == 0``, (default) fill the circle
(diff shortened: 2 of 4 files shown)


MyreMylar:
Just toggling this one closed and open again to rerun the tests as one run had expired. 

Possible it may need to merge with main again to pass. We shall see

buildconfig/stubs/pygame/draw.pyi
```diff
@@ -33,6 +33,10 @@ def circle(
     draw_bottom_left: bool = False,
     draw_bottom_right: bool = False,
 ) -> Rect: ...
+def circles(
+        surface: Surface,
+        draw_sequence: Sequence[Tuple[ColorValue, Coordinate, int, Optional[int]]], /
```
MyreMylar, line 38: So one slightly annoying thing about this `Optional[int]` type hint is that it means "An integer value or `None`, is expected here". If you put only three parameters into the tuple you get a typing error, but it will still work.

Perhaps this hint should be:

`Sequence[Union[Tuple[ColorValue, Coordinate, int, int], Tuple[ColorValue, Coordinate, int]]]`

MyreMylar (changes requested):
I used this test program:

```
import pygame
import random

pygame.init()

display_surface = pygame.display.set_mode((800, 600))
background = pygame.Surface((800, 600))
background.fill((158, 158, 170))
running = True

font = pygame.font.Font(None, 16)
mouse_pos_render = font.render("Mouse Pos: " + str(pygame.mouse.get_pos()), True, pygame.Color('#FFFFFF'))

clock = pygame.time.Clock()

circles_to_draw = [(pygame.Color(random.randint(50, 220),
                                 random.randint(50, 220),
                                 random.randint(50, 220)),
                   (random.randint(50, 750), random.randint(50, 550)),
                   random.randint(5, 50), 0) for _ in range(0, 4000)]

delta_times = []

while running:
    time_delta = max(clock.tick()/1000.0, 0.000000001)

    if len(delta_times) > 100:
        delta_times.pop(0)
    delta_times.append(time_delta)

        if event.type == pygame.QUIT:
            running = False

    display_surface.blit(background, (0, 0))

    for circle in circles_to_draw:
        pygame.draw.circle(display_surface, circle[0], circle[1], circle[2])

    #pygame.draw.circles(display_surface, circles_to_draw)

    mouse_pos_render = font.render("Mouse Pos: " + str(pygame.mouse.get_pos()), True, pygame.Color('#FFFFFF'))
    display_surface.blit(mouse_pos_render, (10, 10))

    mouse_pos_render = font.render("FPS: " + f"{(1.0 / (sum(delta_times)/len(delta_times))):.02f}", True, pygame.Color('#FFFFFF'))
    display_surface.blit(mouse_pos_render, (700, 10))

    pygame.display.flip()

```

Uncommenting to switch between the two methods of drawing 4,000 circles. 

In my testing using `circles()` was about 15% to 20% faster than using circle to do the same thing.

I think the API seems reasonable and fits with the pattern established by `draw.line()` & `draw.lines()`. I will suggest that users are less likely to draw a lot of circles all at once than they are to draw a lot of lines at once - but on balance I think I approve of this new API.

I would like the type hint fixed though.

buildconfig/stubs/pygame/draw.pyi
itzpr3d4t0r: About the perf tests, i think we have the same issue here that we have with fblits or blits. Looping will be more efficient with smaller circles or hollow ones so performance gains can vary significantly.

itzpr3d4t0r: btw i still have to add many tests to this PR and all my other PRs in #3269. Might discover more issues when i do so. Just waiting for reviews like this to make sure i don't spend time where it's not wanted/needed.

PR #3300, opened 2022-07-13
removed pts array in aaline()

itzpr3d4t0r:
A simple change that removed the unnecessary use of an array only to assign it and access it right after.
Small performance benefit aswell


Starbuck5 (approved):
Looks good to me. Thanks!

I'm surprised the compiler wouldn't iron this out, but I guess it didn't, if you measured a performance difference.

ankith26 (approved):
LGTM! 👍 
Not sure that this would have much difference on performance, but this definitely helps with code clarity so thanks for the PR 😎 🚀

PR #3304, opened 2022-07-14
wasm: add pygame.gfxdraw to static preloader

pmp-p:
as it was missing and does not rely on numpy there's no reason not to.


Starbuck5 (approved):
We've got to get this to tank-king stat :smile:

ankith26 (approved):
Thanks for the PR 🥳

PR #3306, opened 2022-07-15
Use PyConfig API instead of older API deprecated in 3.12

pmp-p:
that will remove a deprecation warning that CI may turn in an error.


ankith26 (commented):
This is segfaulting because you are replacing a custom function `pg_EncodeString` with a C API one... the custom function forwards errors when it gets `NULL` and special-cases pathlib objects (for pathlib support)

I'm not sure why this change is needed. If there's an issue in `pg_EncodeString`, then it must be fixed in there, instead of replacing the function

pmp-p:
it was a bit late and iirc Py_FileSystemDefaultEncoding was raising a deprecation warning crashing CI.

pg_EncodeFilePath looks a lot like https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_EncodeFSDefault so tbh i made that PR to see if tests would pass ... and they don't.

So i have no clear idea what needs to be changed. i'll add the trace as soon as i get it back.

pmp-p:
here's the trace ```
   ^
In file included from src_c/static.c:346:
src_c/rwobject.c:247:45: error: 'Py_FileSystemDefaultEncoding' is deprecated [-Werror,-Wdeprecated-declarations]
    PyObject *result = pg_EncodeString(obj, Py_FileSystemDefaultEncoding,
                                            ^
/data/git/python-wasm-plus/devices/emsdk/usr/include/python3.12/fileobject.h:22:1: note: 'Py_FileSystemDefaultEncoding' has been explicitly marked deprecated here
Py_DEPRECATED(3.12) PyAPI_DATA(const char *) Py_FileSystemDefaultEncoding;
^
/data/git/python-wasm-plus/devices/emsdk/usr/include/python3.12/pyport.h:314:54: note: expanded from macro 'Py_DEPRECATED'
#define Py_DEPRECATED(VERSION_UNUSED) __attribute__((__deprecated__))
                                                     ^
In file included from src_c/static.c:365:
src_c/key.c:657:9: warning: unknown pragma ignored [-Wunknown-pragmas]
#pragma PG_WARN(Add missing keycode names here ?)
        ^
9 warnings and 1 error generated.

```

**Comment by ankith26** [comment]
Instead of replacing the function, we should be using [`PyConfig.filesystem_encoding`](https://docs.python.org/3.12/c-api/init_config.html#c.PyConfig.filesystem_encoding) instead of the deprecated `Py_FileSystemDefaultEncoding`

File: src_c/rwobject.c [hunk 1]
```diff
@@ -243,8 +243,18 @@ pg_EncodeString(PyObject *obj, const char *encoding, const char *errors,
 static PyObject *
 pg_EncodeFilePath(PyObject *obj, PyObject *eclass)
 {
+#if PY_VERSION_HEX >= 0x030c0000
+    PyConfig config;
+    _PyInterpreterState_GetConfigCopy(&config);
```
ankith26, line 248: Should this be `PyConfig_InitPythonConfig` or something? I don't think it's a good idea to use private python C API

ankith26 (commented):
If the new config API works on python 3.6+, then I think it should be used on all python versions not just 3.12+

src_c/rwobject.c
pmp-p: There's probably another way to do it, but @vstinner is on holiday and he's my go-to info source on IRC for c-api :)
iirc his py init rework was after 3.7b.

vstinner: Are you looking for PyUnicode_EncodeFSDefault()? Calling _PyInterpreterState_GetConfigCopy() here sounds overkill. Casting config.filesystem_encoding wide-string (``wchar_t*`` type) to a byte string (``char*``) sounds plain wrong.

pmp-p: indeed PyUnicode_EncodeFSDefault was my first pick, but i obviously don't know how to use it correctly as i made it segfault.

MyreMylar: Is this PR still good, or is it going to go back to `PyUnicode_EncodeFSDefault`?

vstinner: For me, PyUnicode_EncodeFSDefault() should be used. No idea why you got a crash, you didn't provide any detail (OS, Python version, error message, traceback, etc.).

pmp-p: thanks for the advice, let's try again.

pmp-p: it crashed , i'll add fencing for >=3.7 because it seems to be the last change here https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_EncodeLocale

pmp-p: crashes with  cp37-manylinux_x86_64

ankith26: I've mentioned the reason for this segfault (that I think) in my first comment, `PyUnicode_EncodeFSDefault` is simply not a good full replacement for our custom function. Callers expect `pg_EncodeString` to forward errors when it gets `NULL`, but `PyUnicode_EncodeFSDefault` is segfaulting trying to deref a null pointer (which makes sense because forwarding errors is not a documented API detail for this function, and we can't rely on this)

Fencing version-wise won't fix this segfault, we need to either fix the usage of this function in all instances, or add a null check at the start and forward it

pmp-p: 3.8 so i'm really using it wrong and i can't see where

vstinner: PyUnicode_EncodeFSDefault() must not be called on NULL. Why not adding a simple check to see if obj is NULL?

pmp-p: good idea

vstinner:
>  ManyLinux / build (manylinux1, i686, cp36-* cp37-* cp38-* cp39-*) (pull_request) Failing after 8m — build (manylinux1, i686, cp36-* cp37-* cp38-* cp39-*) 

It failed with:

```
  ERROR: test_path_with_null_bytes (pygame.tests.rwobject_test.RWopsEncodeFilePathTest)
    File "/tmp/tmp.RxyyHU5944/venv/lib/python3.9/site-packages/pygame/tests/rwobject_test.py", line 123, in test_path_with_null_bytes
      encoded_file_path = encode_file_path(b)
  TypeError: bad argument type for built-in operation
```

pmp-p:
so i guess problem is a null byte in the object buffer not a null object.

vstinner:
The new CI failure seems to be unrelated to the PR:

```
   ERROR: test_load_default_font_filename (pygame.tests.ftfont_test.FtFontTypeTest)
    File "/tmp/tmp.YEBvgX5990/venv/lib/python3.9/site-packages/pygame/tests/font_test.py", line 537, in test_load_default_font_filename
      f = pygame_font.Font(pygame_font.get_default_font(), 20)
    File "/tmp/tmp.YEBvgX5990/venv/lib/python3.9/site-packages/pygame/ftfont.py", line 54, in __init__
      super().__init__(file, size=size, resolution=resolution)
  FileNotFoundError: No such file or directory: 'freesansbold.ttf'.
(...)
   Ran 1674 tests in 13.549s
  
  FAILED (failures=2, errors=1)
```

pmp-p:
301

illume:
403

PR #3310, opened 2022-07-15
pygame.draw.rects() function

itzpr3d4t0r:
Implements one of #3269.
This one implements the rects() function for drawing many rectangles at once. The function's use case should be heterogeneous rectangle sequences, of different size, width, radius and colors.
First performance results show a bigger benefit for width=1 rectangles and other improvements vary depending on scale and width. Lower benefits are for big fully drawn rects, at around 5-8%.
This one is probably the function i'm the least confident with both in terms of implementation and documentation, so if you see something wrong please tell me.


docs/reST/ref/draw.rst
```diff
@@ -100,6 +100,45 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
...
+   Draws a sequence of rectangles on the given surface.
+
+   :param surface: the surface to draw the rectangles on
+   :param draw_sequence: a sequence composed of (color, rect, width, radius) where:
+
+      - color: color to draw with, the alpha value is optional if using a tuple ``(RGB[A])``
+
+      - rect: the rectangle to draw, with its position and dimensions
+
+      - width (optional): used for line thickness or to indicate that the rectangle is to be
```
novialriptide, line 117: missing periods for color, rect, and width

novialriptide (changes requested):

docs/reST/ref/draw.rst
itzpr3d4t0r: Thanks

PR #3313, opened 2022-07-16
some wrong relative imports in cython code

pmp-p:
```
>>> pygame._sdl2.sdl2.error is pygame._sdl2.controller.error
True
```

but pygame._sdl2.error is an alias ( from sdl2 import * ) in  pygame._sdl2 package
```
>>> pygame._sdl2.error
<class 'pygame._sdl2.sdl2.error'>
```

### DIFF
--- a/src_c/_sdl2/audio.c
+++ b/src_c/_sdl2/audio.c
@@ -6,7 +6,7 @@
         "depends": [],
         "name": "pygame._sdl2.audio",
         "sources": [
-            "src_c\\cython\\pygame\\_sdl2\\audio.pyx"
+            "src_c/cython/pygame/_sdl2/audio.pyx"
         ]
     },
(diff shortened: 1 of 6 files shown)


ankith26 (commented):
When making cython changes, you also need to push the generated c code you get after doing a `python3 setup.py cython`

pmp-p:
thanks, yes i will but i think there are other circular imports around

pmp-p:
@ankith26 should be ok now, in wasm I was able to do C multiphase init from the statically linked libraries. Maybe WebAssembly dynamic loading will be possible soon.

ankith26 (approved):
I've not tested this PR, but the code looks fine.

Thanks for the PR 🚀

MyreMylar (approved):
LGTM 👍

PR #3314, opened 2022-07-16
conservative blits() improvements

itzpr3d4t0r:
I've made 2 changes to the surface.blits() function:
1. Efficient item extraction using PySequence_ITEM() instead of PySequence_GetItem() for each item in blitsequence.
2. Efficient appending to the return rect list in the case of blitsequence being a countable sequence (list or tuple etc.).

Performance benefits vary heavily depending on the blit time, but i measured a mean  5-15% improvement.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1907,23 +1907,41 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
     PyObject *special_flags = NULL;
     PyObject *ret = NULL;
     PyObject *retrect = NULL;
-    Py_ssize_t itemlength;
+    Py_ssize_t itemlength, sequencelength, curriter = 0;
     int doreturn = 1;
     int bliterrornum = 0;
+    int issequence = 0;
     static char *kwids[] = {"blit_sequence", "doreturn", NULL};
     if (!PyArg_ParseTupleAndKeywords(args, keywds, "O|i", kwids, &blitsequence,
                                      &doreturn))
         return NULL;
 
-    if (doreturn) {
-        ret = PyList_New(0);
-        if (!ret)
-            return NULL;
-    }
-    if (!PyIter_Check(blitsequence) && !PySequence_Check(blitsequence)) {
+    if (!PyIter_Check(blitsequence) &&
+        !(issequence = PySequence_Check(blitsequence))) {
         bliterrornum = BLITS_ERR_SEQUENCE_REQUIRED;
         goto bliterror;
     }
+
+    if (doreturn) {
+        /* If the sequence is countable, meaning not a generator, we can get
+         * faster rect appending to the list by pre allocating it
+         * to later call the more efficient SET_ITEM*/
+        if (issequence) {
+            sequencelength = PySequence_Size(blitsequence);
+            if (sequencelength == -1) {
+                bliterrornum = BLITS_ERR_PY_EXCEPTION_RAISED;
+                goto bliterror;
+            }
+
+            ret = PyList_New(sequencelength);
+        }
+        else {
+            ret = PyList_New(0);
+        }
+        if (!ret)
+            return NULL;
+    }
+
     iterator = PyObject_GetIter(blitsequence);
     if (!iterator) {
         Py_XDECREF(ret);
@@ -1951,16 +1969,16 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
         assert(itemlength >= 2);
 
         /* (Surface, dest) */
-        srcobject = PySequence_GetItem(item, 0);
-        argpos = PySequence_GetItem(item, 1);
+        srcobject = PySequence_ITEM(item, 0);
+        argpos = PySequence_ITEM(item, 1);
 
         if (itemlength >= 3) {
             /* (Surface, dest, area) */
-            argrect = PySequence_GetItem(item, 2);
+            argrect = PySequence_ITEM(item, 2);
         }
         if (itemlength == 4) {
             /* (Surface, dest, area, special_flags) */
-            special_flags = PySequence_GetItem(item, 3);
+            special_flags = PySequence_ITEM(item, 3);
         }
         Py_DECREF(item);
         /* Clear item to avoid double deref on errors */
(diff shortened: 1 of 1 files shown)


src_c/surface.c
```diff
@@ -1907,23 +1907,34 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
-    }
-    if (!PyIter_Check(blitsequence) && !PySequence_Check(blitsequence)) {
+    if (!PyIter_Check(blitsequence) &&
+        !(issequence = PySequence_Check(blitsequence))) {
         bliterrornum = BLITS_ERR_SEQUENCE_REQUIRED;
         goto bliterror;
     }
+
+    if (doreturn) {
+        if (issequence && (sequencelength = PySequence_Size(blitsequence))) {
```
ankith26, line 1926: `PySequence_Size` returns -1 on a failure, and this should be handled here. The code would be more clearer if the `sequencelength = PySequence_Size(blitsequence)` part was inside the conditional block

itzpr3d4t0r: resolved it

src_c/surface.c
```diff
@@ -1907,23 +1907,37 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+        bliterrornum = BLITS_ERR_SEQUENCE_REQUIRED;
+        goto bliterror;
+    }
+
     if (doreturn) {
-        ret = PyList_New(0);
+        if (issequence) {
+            sequencelength = PySequence_Size(blitsequence);
+            if (sequencelength == -1) {
+                bliterrornum = BLITS_ERR_SEQUENCE_REQUIRED;
```
ankith26, line 1929:
```suggestion
                bliterrornum = BLITS_ERR_PY_EXCEPTION_RAISED;
```
Because `PySequence_Size` would have already set something

Alternatively, there can be a simple `return NULL;` too here to exit early (because no other variables that need cleanup are initialised at this point)

itzpr3d4t0r: I'll go with the return NULL way

ankith26 (approved):
LGTM! Thanks for the PR 🎉 😎

Starbuck5 (commented):
I want some comments!

src_c/surface.c
```diff
@@ -1907,23 +1907,35 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
+        bliterrornum = BLITS_ERR_SEQUENCE_REQUIRED;
+        goto bliterror;
+    }
+
     if (doreturn) {
-        ret = PyList_New(0);
+        if (issequence) {
+            sequencelength = PySequence_Size(blitsequence);
+            if (sequencelength == -1)
+                return NULL;
```
Starbuck5, line 1929: Should this be goto blitterror?

src_c/surface.c
```diff
@@ -2029,13 +2041,19 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
-            if (PyList_Append(ret, retrect) != 0) {
-                bliterrornum = BLITS_ERR_PY_EXCEPTION_RAISED;
-                goto bliterror;
+            if (issequence) {
+                PyList_SET_ITEM(ret, curriter, retrect);
+            }
+            else {
+                if (PyList_Append(ret, retrect) != 0) {
+                    bliterrornum = BLITS_ERR_PY_EXCEPTION_RAISED;
+                    goto bliterror;
```
Starbuck5, line 2050: This could be collapsed into a single `else if`, not sure if that would be better. What do you think?

src_c/surface.c
```diff
@@ -2029,13 +2041,19 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
         if (doreturn) {
             retrect = NULL;
             retrect = pgRect_New(&dest_rect);
-            if (PyList_Append(ret, retrect) != 0) {
-                bliterrornum = BLITS_ERR_PY_EXCEPTION_RAISED;
-                goto bliterror;
+            if (issequence) {
+                PyList_SET_ITEM(ret, curriter, retrect);
+            }
```
Starbuck5, line 2046: I think this would work if you put the `curriter++` inside this block here. `PyList_SET_ITEM(ret, curriter++, retrect)`

This keeps the curriter incrementing closer to its area of responsibility, and doesn't spend time doing that on iterators of undefined length.

MyreMylar:
Using 64x64 alpha images into bltis the performance seems pretty much the same. I got:

27 seconds in 2.1.2

and 26.9 seconds with this PR.

Using this test code:

```
from timeit import timeit
setup_code = """
import pygame
import random
pygame.init()
display = pygame.display.set_mode((1280, 720))
background = pygame.Surface((1280, 720))
background.fill((0, 0, 0))
glow_surf = pygame.image.load(\'images/glow.png\').convert_alpha()
blits_to_make = []
for i in range(0, 8000):
    pos_rect = pygame.Rect(0, 0, 64.0, 64.0)
    pos_rect.center = random.uniform(0.0, 1280.0), random.uniform(0.0, 720.0)
    blits_to_make.append((glow_surf, pos_rect))

"""
print(timeit("background.blits(blits_to_make)", setup_code, number=1000))
```

This is on windows. Does it do better with smaller blits or larger blits? Otherwise the tests all pass locally and it visually still seems to do the blitting ;)

itzpr3d4t0r:
basically as it's always the case with blits, smaller and 24 bit blits are faster and therefore get better performance

Starbuck5:
I found an 18% performance improvement on a test blitting (doreturn=True) thousands of tiny surfaces. That should be best case for this PR.

MyreMylar:
I'm happy to see this merge once Starbuck's changes above get made.

src_c/surface.c
itzpr3d4t0r: i just did what Ankith told me here https://github.com/pygame/pygame/pull/3314#discussion_r922873553

src_c/surface.c
itzpr3d4t0r: This is probably the cleanes way, don't even need an else. Alternative would be this
```C
else if (PyList_Append(ret, retrect)) {
      Py_DECREF(retrect);
  }
else {
    Py_DECREF(retrect);
    retrect = NULL;
    bliterrornum = BLITS_ERR_PY_EXCEPTION_RAISED;
    goto bliterror;
}
retrect = NULL; /* Clear to avoid double deref on errors *
}
```

itzpr3d4t0r: which is even more convoluted

src_c/surface.c
itzpr3d4t0r: yep agree

src_c/surface.c
itzpr3d4t0r: I had it that way before

src_c/surface.c
itzpr3d4t0r: In the end i made the change anyway.

itzpr3d4t0r:
@MyreMylar, @Starbuck5 i made the changes, but one is a conflict between you and Ankith vision of the problem

Starbuck5 (approved):
Just the goto blitterror / return NULL thing to resolve, so I'll go ahead and approve.

Resolve that and somebody can merge it.

PR #3317, opened 2022-07-17
Make sure set_mode always returns a cleared window

### LINKED ISSUE #3265: Different `set_mode` behavior between versions
As I have discovered with the help of @oddbookworm then **in 2.1.2 every call to `set_mode` clears/fills the display** while **in 2.1.3.dev5 starting from the 3rd call it doesn't clear/fill the surface** anymore, it leaves as it was.

Is this intended behavior or is it a bug?

_Python version:_ Python 3.10.4
_OS:_ Windows 8.1

MRE:
```py
import pygame


pygame.display.set_mode((100, 50))  # 1st call

screen = pygame.display.set_mode((100, 50))  # 2nd call
pygame.draw.rect(screen, "red", (0, 0, 50, 50))

screen = pygame.display.set_mode((100, 50))  # 3rd call
pygame.draw.rect(screen, "green", (50, 0, 50, 50))

pygame.display.flip()


pygame.image.save(screen, f"{pygame.__version__}_out.png")
```

Results:
2.1.2_out.png:
image

2.1.3.dev5_out.png:
image

ankith26:
fixes #3265

From the code, it seemed like retaining older window surface on subsequent `set_mode` calls is intentional, but no pygame version from 1.9.6 to 2.1.2 actually does it (in my testing, on windows)

This PR enforces the display clear on our end unconditionally when `set_mode` is called

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -205,6 +205,11 @@ required).
 
    .. versionchanged:: 1.9.5 ``display`` argument added
 
+   .. versionchanged:: 2.1.3
+      pygame now ensures that subsequent calls to this function clears the
+      window to black. On older versions, this was an implementation detail
+      on the major platforms this function was tested with.
+
    .. ## pygame.display.set_mode ##
 
 .. function:: get_surface
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -843,7 +843,6 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
        screen as the old one */
     int display = _get_display(win);
     char *title = state->title;
-    int init_flip = 0;
     char *scale_env;
 
     char *keywords[] = {"size", "flags", "depth", "display", "vsync", NULL};
@@ -1101,7 +1100,6 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                 win = SDL_CreateWindow(title, x, y, w_1, h_1, sdl_flags);
                 if (!win)
                     return RAISE(pgExc_SDLError, SDL_GetError());
-                init_flip = 1;
             }
             else {
                 /* set min size to (1,1) to erase any previously set min size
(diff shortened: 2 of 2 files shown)


Starbuck5:
I have a relevant piece of information from my experience with set_mode.

Whether a window sticks around through set_mode or not is consistent and platform dependent. IIRC on MacOS a new window is created on a resize, for example, but not on Windows.

MyreMylar (approved):
LGTM 👍 

Should at least be consistent now.

Starbuck5:
Ok, I haven't really fleshed out this idea yet, but

set_mode is also the window resizing API. It shouldn't lose the contents of the original window surface.

ankith26:
The idea behind this PR is to add consistency. It should be a documented API detail that `set_mode` clears a window and users are supposed to redraw. 

As mentioned above, the behaviour from pygame 1.9.6 to pygame 2.1.2 on windows has been clearing the window. This is only breaking now as an affect of the SDL 2.0.22/2.0.20 update (on windows only)

I also just tested on Ubuntu 22.04 (WSL) and the behaviour is the same as windows on pygame 2.1.2 (SDL 2.0.16), and I also had a SDL 2.0.22 wheel lying around from my manylinux PR (the one that is slow) and this too works the same the rest.

I don't know the status of this on macs.

MyreMylar:
I can't imagine many situations where pygame could successfully manage a window resizing event _without_ the user redrawing.

Where do we put the pixels? Do we run a smoothscale on the display surface? Fill the new space with black pixels in the case of an enlargement? Cut off the surface when shrinking?

I think always clearing to black is the easiest behaviour at pygame's level - because we can't know exactly what the application developer wants to happen when their window is resized.

ankith26:
I just noticed this change is missing some docs, will add that rq

illume:
It doesn't seem there is consensus yet on this one. 

> set_mode is also the window resizing API. It shouldn't lose the contents of the original window surface.

Does resizing the window leave the old surface just on windows? (or other platforms too?)
Seems like that might be useful behaviour?

To me, it's 99% clear that set_mode should definitely clear the surface the first time set_mode is called.

illume:
> Yeah, I also agree with clearing the background on every set_mode() call (so that users can redraw their stuff). I also tested 1.9.6 behaviour and it is inline with 2.1.2 and the screen is always black after a set_mode call (so this issue is definitely a 2.1.3 break)

Ok, since this is old behaviour; I agree that this patch should be merged. The new behaviour could be useful though, but perhaps that can be added with a flag.

If there's no objections from @Starbuck5 I'd like to merge this for 2.1.3.

Issue #3315, opened 2022-07-16
Investigate speeding up `pygame.transform.scale_by`


Starbuck5:
When I wrote `scale_by` and `smoothscale_by`, I made them call the corresponding `scale` and `smoothscale` by creating a python arguments tuple in the c code, and then calling the other function in C code.

It would likely be faster to not have to convert from Python args to C variables back to Python args back to C variables, rather parsing into C once and staying at that level.

Relevant code: https://github.com/pygame/pygame/blob/main/src_c/transform.c#L524-L625, https://github.com/pygame/pygame/blob/main/src_c/transform.c#L1471-L1585

Basically the idea is to make a C function common backend (takes two surfaces and a size probably?) that both `scale_by` and `scale` can send C arguments too, which should speed up `scale_by`. (Same with smoothscale)

Starbuck5:
I'd like to assign @itzpr3d4t0r to this.

itzpr3d4t0r:
I'll have the honor!

PR #3319, opened 2022-07-17
scale(), scale_by(), smoothscale(), smoothscale_by() reorganization and speed up

itzpr3d4t0r:
Implements #3315.
Results are a bit strange, but the functions do work. I used this code:
```Python
from timeit import timeit

import pygame
from pygame.transform import scale, scale_by, smoothscale, smoothscale_by

pygame.init()
WIDTH, HEIGHT = 500, 500
win = pygame.display.set_mode((WIDTH, HEIGHT))

IMAGE = pygame.image.load("images/cell.png").convert()
GLOB = {
    "scale": scale,
    "IMAGE": IMAGE,
    "scale_by": scale_by,
    "smoothscale": smoothscale,
    "smoothscale_by": smoothscale_by
}


def test(test_str: str):
    print(timeit(test_str, globals=GLOB))

print("SCALE: ", end="")
test("scale(IMAGE, (100, 100))")

print("SCALE_BY: ", end="")
test("scale_by(IMAGE, 2.45)")

print("SMOOTHSCALE: ", end="")
test("smoothscale(IMAGE, (100, 100))")

print("SMOOTHSCALE_BY: ", end="")
test("smoothscale_by(IMAGE, 2.45)")
```
and got these results:
OLD
```
SCALE: 10.217117799999869
SCALE_BY: 13.330813199999284
SMOOTHSCALE: 17.059421400000247
SMOOTHSCALE_BY: 22.5625871000002
```
NEW
```
SCALE: 7.131523500000185
SCALE_BY: 10.155133100000057
SMOOTHSCALE: 16.981359200000043
SMOOTHSCALE_BY: 22.357186100000035
```

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -521,13 +521,66 @@ stretch(SDL_Surface *src, SDL_Surface *dst)
     }
 }
 
+static SDL_Surface *
+scale_to(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int width,
+         int height)
+{
+    SDL_Surface *src = NULL;
+    SDL_Surface *retsurf = NULL;
+
+    if (width < 0 || height < 0)
+        return RAISE(PyExc_ValueError, "Cannot scale to negative size");
+
+    src = pgSurface_AsSurface(srcobj);
+
+    if (!dstobj) {
+        retsurf = newsurf_fromsurf(src, width, height);
+        if (!retsurf)
+            return NULL;
+    }
+    else {
+        retsurf = pgSurface_AsSurface(dstobj);
+    }
+
+    if (retsurf->w != width || retsurf->h != height) {
+        return (SDL_Surface *)(RAISE(
+            PyExc_ValueError,
+            "Destination surface not the given width or height."));
+    }
+
+    if (src->format->BytesPerPixel != retsurf->format->BytesPerPixel) {
+        return (SDL_Surface *)(RAISE(
+            PyExc_ValueError,
+            "Source and destination surfaces need the same format."));
+    }
+
+    if ((width && height) && (src->w && src->h)) {
+        SDL_LockSurface(retsurf);
+        pgSurface_Lock(srcobj);
+
+        Py_BEGIN_ALLOW_THREADS;
+        if (width == 2 * src->w && height == 2 * src->h) {
+            scale2xraw(src, retsurf);
+        }
+        else {
+            stretch(src, retsurf);
+        }
+        Py_END_ALLOW_THREADS;
+
+        pgSurface_Unlock(srcobj);
+        SDL_UnlockSurface(retsurf);
+    }
+
+    return retsurf;
+}
+
 static PyObject *
 surf_scale(PyObject *self, PyObject *args, PyObject *kwargs)
 {
     pgSurfaceObject *surfobj;
     PyObject *surfobj2 = NULL;
     PyObject *size;
-    SDL_Surface *surf, *newsurf;
+    SDL_Surface *newsurf;
     int width, height;
     static char *keywords[] = {"surface", "size", "dest_surface", NULL};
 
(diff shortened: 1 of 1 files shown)


Starbuck5:
I didn't measure much of a performance difference at all, I guess this is more of a code quality PR 🙂

Maybe 1.5% better performance in some cases, but the benchmarks are shaky.

src_c/transform.c
```diff
@@ -521,66 +521,84 @@ stretch(SDL_Surface *src, SDL_Surface *dst)
...
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O!O|O!", keywords,
+                                     &pgSurface_Type, &surfobj, &size,
+                                     &pgSurface_Type, &surfobj2))
+        return NULL;
+
+    if (!pg_TwoIntsFromObj(size, &width, &height))
+        return RAISE(PyExc_TypeError, "size must be two numbers");
+
+    if (!(newsurf = scale_to(surfobj, surfobj2, width, height))) {
```
Starbuck5, line 595: I'd rather you assign and then check, rather than doing an assignment inside a comparison. It will compile to the same thing, but the former is more readable.

src_c/transform.c
```diff
@@ -589,39 +607,34 @@ surf_scale(PyObject *self, PyObject *args, PyObject *kwargs)
...
-    if (width < 0 || height < 0)
-        return RAISE(PyExc_ValueError, "Cannot scale to negative size");
-
-    if (surfobj2)
-        new_args = Py_BuildValue("O(ii)O", surfobj, width, height, surfobj2);
-    else
-        new_args = Py_BuildValue("O(ii)", surfobj, width, height);
-    if (new_args == NULL)
+    if (!(newsurf = scale_to(surfobj, surfobj2, (int)(surf->w * scalex),
+                             (int)(surf->h * scaley))))
```
Starbuck5, line 629: Same here about the inline assignment, and I'd also like curly braces around this condition.

src_c/transform.c
```diff
@@ -1568,20 +1596,17 @@ surf_scalesmooth_by(PyObject *self, PyObject *args, PyObject *kwargs)
...
-        return RAISE(PyExc_ValueError, "Cannot scale to negative size");
-
-    if (surfobj2)
-        new_args = Py_BuildValue("O(ii)O", surfobj, width, height, surfobj2);
-    else
-        new_args = Py_BuildValue("O(ii)", surfobj, width, height);
-    if (new_args == NULL)
+    if (!(newsurf =
+              smoothscale_to(self, surfobj, surfobj2, (int)(surf->w * scale),
+                             (int)(surf->h * scaley))))
```
Starbuck5, line 1601: Same here with inline assignment and curly braces.

src_c/transform.c
```diff
@@ -1468,79 +1481,95 @@ scalesmooth(SDL_Surface *src, SDL_Surface *dst, struct _module_state *st)
...
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O!O|O!", keywords,
+                                     &pgSurface_Type, &surfobj, &size,
+                                     &pgSurface_Type, &surfobj2))
+        return NULL;
+
+    if (!pg_TwoIntsFromObj(size, &width, &height))
+        return RAISE(PyExc_TypeError, "size must be two numbers");
+
+    if (!(newsurf = smoothscale_to(self, surfobj, surfobj2, width, height)))
```
Starbuck5, line 1567: here as well

Starbuck5 (commented):
Besides my formatting nitpicks, this PR looks good to me. It's hard to review though, since the diff is so gnarly. (Not your fault)

MyreMylar (approved):
Had a quick look over the code and it looks reasonable with a bit less duplication.

The tests pass and as far as I can tell it has made little to no difference to performance.

Starbuck5 (approved):
Thanks!

As a final check to make sure outputs hadn't changed, I scaled and smoothscaled a surface into a png file before and after, and verified the file hash hadn't changed.

illume (commented):
Very nice. 👍 
It's common to see different speeds depending on image size and CPU cache size/architecture.

ps, as an aside; have you see https://halide-lang.org/ and https://github.com/pydata/numexpr ? The videos that halide provides are especially interesting.

Issue #3316, opened 2022-07-16
`tobytes` and `frombytes` shim for pygame.image


Starbuck5:
The pygame.image module has functions `tostring` and `fromstring`. But here in Python 3, they actually work on bytes.

I figure we could at least put in shim functions that are properly named.

Other libraries (at least PIL) already went through deprecation cycles about this.

But it would be nice to have the properly named functions around, and marked "preferred" perhaps, just so it matches nicely with other libraries, and with people's expectations.

Emc2356:
i could tackle this issue later, i'd be happy to do it

PR #3321, opened 2022-07-18
added aliases for pygame.image.(to|from)string

Emc2356:
#3316

### DIFF
--- a/docs/reST/ref/image.rst
+++ b/docs/reST/ref/image.rst
@@ -189,7 +189,8 @@ following formats.
    | :sl:`transfer image to string buffer`
    | :sg:`tobytes(Surface, format, flipped=False) -> bytes`
 
-   since version 2.1.3 it is recommended that `tobytes` is used instead of `tostring` 
+   .. versionadded:: 2.1.3 added :func:`tobytes` as an alias for :func:`tostring`,
+                     the use of :func:`tobytes` is recommended over :func:`tostring`.
 
    .. ## pygame.image.tobytes ##
 
@@ -223,7 +224,8 @@ following formats.
    | :sl:`create new Surface from a string buffer`
    | :sg:`frombytes(bytes, size, format, flipped=False) -> Surface`
 
-   since version 2.1.3 it is recommended that `frombytes` is used instead of `fromstring` 
+   .. versionadded:: 2.1.3 added :func:`frombytes` as an alias for :func:`fromstring`,
+                     the use of :func:`frombytes` is recommended over :func:`fromstring`.
 
    .. ## pygame.image.frombytes ##
 
--- a/docs/reST/ref/image.rst
+++ b/docs/reST/ref/image.rst
@@ -228,6 +228,7 @@ following formats.
 
    added :func:`formbytes` as an alias for :func:`formstring`,
    the use of :func:`formbytes` is recommended over :func:`formstring`.
+   
    .. versionadded:: 2.1.3 
 
    .. ## pygame.image.frombytes ##
--- a/docs/reST/ref/image.rst
+++ b/docs/reST/ref/image.rst
@@ -184,6 +184,16 @@ following formats.
 
    .. ## pygame.image.tostring ##
 
+.. function:: tobytes
+
+   | :sl:`transfer image to string buffer`
+   | :sg:`tobytes(Surface, format, flipped=False) -> bytes`
+
+   since version 2.1.3 it is recommended that `tobytes` is used instead of `tostring` 
+
+   .. ## pygame.image.tobytes ##
+
+
 .. function:: fromstring
 
    | :sl:`create new Surface from a string buffer`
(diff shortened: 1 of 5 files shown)


docs/reST/ref/image.rst
```diff
@@ -202,10 +212,21 @@ following formats.
...
+             as of pygame 2.1.3.
 
    .. ## pygame.image.fromstring ##
 
+.. function:: frombytes
+
+   | :sl:`create new Surface from a string buffer`
+   | :sg:`frombytes(bytes, size, format, flipped=False) -> Surface`
+
+   since version 2.1.3 it is recommended that `frombytes` is used instead of `fromstring`
```
ankith26, line 226: It would be better to mention with more clarity that this is an alias of `fromstring`.

Also needs the `.. versionadded::` tag

docs/reST/ref/image.rst
```diff
@@ -184,6 +184,16 @@ following formats.
 
    .. ## pygame.image.tostring ##
 
+.. function:: tobytes
+
+   | :sl:`transfer image to string buffer`
+   | :sg:`tobytes(Surface, format, flipped=False) -> bytes`
+
+   since version 2.1.3 it is recommended that `tobytes` is used instead of `tostring`
```
ankith26, line 192: Needs to mention that this is an alias of `tostring`.
Also needs the `.. versionadded::` tag

ankith26 (approved):
LGTM! Thanks for the PR 🎈:balloon:

ankith26 (commented):
A minor nitpick, there should be tests for this. They don't need to be extensive, they only need to check that the alias exists.

docs/reST/ref/image.rst
```diff
@@ -184,6 +184,17 @@ following formats.
 
    .. ## pygame.image.tostring ##
 
+.. function:: tobytes
+
+   | :sl:`transfer image to string buffer`
+   | :sg:`tobytes(Surface, format, flipped=False) -> bytes`
+
+   .. versionadded:: 2.1.3 added :func:`tobytes` as an alias for :func:`tostring`,
```
zoldalma999, line 192: Formatting here is a bit off from other places. I would change it so that the text after `versionadded` is the body/description of the function (and rewording it a little), and keeping just a `.. versionadded:: 2.1.3` at the end

docs/reST/ref/image.rst
```diff
@@ -202,10 +213,22 @@ following formats.
...
+             as of pygame 2.1.3.
 
    .. ## pygame.image.fromstring ##
 
+.. function:: frombytes
+
+   | :sl:`create new Surface from a string buffer`
+   | :sg:`frombytes(bytes, size, format, flipped=False) -> Surface`
+
+   .. versionadded:: 2.1.3 added :func:`frombytes` as an alias for :func:`fromstring`,
```
zoldalma999, line 227: Same as with `tobytes` docs

zoldalma999 (commented):
Left a few things I noticed

Also, are we going to deprecate the original functions?

src_c/image.c
```diff
@@ -1471,7 +1471,9 @@ static PyMethodDef _image_methods[] = {
      METH_NOARGS, DOC_PYGAMEIMAGEGETSDLIMAGEVERSION},
 
     {"tostring", image_tostring, METH_VARARGS, DOC_PYGAMEIMAGETOSTRING},
+    {"tobytes", image_tostring, METH_VARARGS, DOC_PYGAMEIMAGETOBYTES},
```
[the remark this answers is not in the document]
Starbuck5: Good thought, but PIL/pillow calls it `tobytes` and `frombytes`, so I'd like to stick with that.

Starbuck5:
> Also, are we going to deprecate the original functions?

Eventually sure, but I don't feel any rush to do it yet. The big thing for me is just getting the aliases out.

docs/reST/ref/image.rst
```diff
@@ -184,6 +184,17 @@ following formats.
 
    .. ## pygame.image.tostring ##
 
+.. function:: tobytes
+
+   | :sl:`transfer image to string buffer`
```
Starbuck5, line 189: I just noticed all these still say string buffers, that should be changed.

Starbuck5 (approved):
Left a couple small reviews, still a couple small things from zoldalma to resolve.

Thanks for tackling this!

I'll approve it now, even though there are a few small things left to do.

src_c/image.c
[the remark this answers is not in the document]
Emc2356: also other libraries like numpy use `joincase` so it would be preferred to stick with `joincase`

Emc2356:
> A minor nitpick, there should be tests for this. They don't need to be extensive, they only need to check that the alias exists.

yea, this is a good idea
should i just do something like this:
```py
with self.assertRaises(AttributeError):
    pygame.image.tobytes
with self.assertRaises(AttributeError):
    pygame.image.frombytes
```

ankith26:
The exact opposite of that, accessing those two methods should *not* raise `AttributeError`s

docs/reST/ref/image.rst
```diff
@@ -202,10 +215,23 @@ following formats.
...
    .. ## pygame.image.fromstring ##
 
+.. function:: frombytes
+
+   | :sl:`create new Surface from a byte buffer`
+   | :sg:`frombytes(bytes, size, format, flipped=False) -> Surface`
+
+   added :func:`formbytes` as an alias for :func:`formstring`,
+   the use of :func:`formbytes` is recommended over :func:`formstring`.
+   .. versionadded:: 2.1.3
```
Starbuck5, line 231: There needs to be a newline between this sentence and the version added.

The generated checks don't come out properly otherwise, I checked.

docs/reST/ref/image.rst
```diff
@@ -184,9 +184,22 @@ following formats.
 
    .. ## pygame.image.tostring ##
 
+.. function:: tobytes
+
+   | :sl:`transfer image to byte buffer`
+   | :sg:`tobytes(Surface, format, flipped=False) -> bytes`
+
+   added :func:`tobytes` as an alias for :func:`tostring`,
+   the use of :func:`tobytes` is recommended over :func:`tostring`.
```
Starbuck5, line 193: I'd like this documentation to be a little more helpful, same with the other one 🙂.

What does it do? Why was it added? What should people know?

You've got that it was added, and that it is recommended (what people should know).

illume (commented):
👍

Issue #3324, opened 2022-07-20
Tracking FASTCALL/VECTORCALL improvements


ankith26:
Recently, a new contributor @itzpr3d4t0r has expressed a lot of interest in using `FASTCALL` to speedup some pygame functions (thanks a lot for doing this!). I thought it would be a good idea to open an issue to be tracking progress on this, and also put out some info and have discussions about this.

To explain what `FASTCALL` is in short, it is basically a new and faster python calling convention. Usually in the older `VARARGS` calling convention, C API functions expect to handle a tuple `args`. On the CPython side, this adds *small* calling overhead on each function, because it has to make a temporary tuple object from the arguments users pass. With `FASTCALL`, the arguments directly come in as a C array of python object pointers.

There are also a couple of issues here. One of the key issues being, there is no clean C API to be dealing with this. Handling positional args becomes a bit more tedious, and handling keyword args becomes a huge pain. Handling keyword args also adds a bit of its own overhead here (if we do get to the stage where we have custom API to handle this)
A more minor and workaroundable issue is that `FASTCALL` is python 3.7+ only, and at the time of writing of this comment we still support python 3.6. We can put 3.6-only wrappers to our `FASTCALL` functions, and export them as regular `VARARGS` function. Obviously, this would mean that the optimisations won't actually work on python 3.6 (and in a few cases, performance may also regress a tiny bit on 3.6), but we don't have to worry about this as long as the functions keep working on python 3.6 while we support it.
When we do finally drop 3.6 support at some point, the wrapper macros should be easy enough to remove.

Recently pygame has been trying to port more and more of its function to support keyword args, but by applying `FASTCALL` we might be going against it. So I propose that we should not try to make everything `FASTCALL`, or everything support kwargs, but instead strike a balance between both. There are many places where kwargs support is a no-brainer. But there are places where `FASTCALL` has some good benefits.

Here are a few things of the top of my head that should help decide where we can use this, and where we shouldn't
- In general, the `FASTCALL` optimisation best benefits short running pygame functions that are called many times in your average game loop. If a function is long running, or is only expected to be called at the start of programs (like resource loading functions), applying the `FASTCALL` optimisation tricks does not have much of an impact.
[...]

PR #3325, opened 2022-07-20
Add FASTCALL compat macros, update Color.update to use the same

ankith26:
PR to setup some groundwork on #3324

This PR basically adds some compat macros to make defining fastcall functions easier while we still retain python 3.6 support.

To illustrate this, I picked `Color.update` which seemed like a good example because
1) This kind of function does not benefit much from having keyword arguments (and it did not in the past either, most importantly)
2) Using `FASTCALL` here also happens to simplify the implementation of the actual function. I personally like this aspect of the PR more than its performance boost
3) And of course, there is a small speedboost. On a benchmark I measured a 20% improvement in the mean, and the standard deviation about the mean reduced a bit too. However, it is to be noted that even before using `FASTCALL`, `Color.update` was a relatively cheap function anyways. It only takes of the order of a few 100 nanoseconds.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -281,6 +281,41 @@ supported Python version. #endif */
 
 #define PyType_Init(x) (((x).ob_type) = &PyType_Type)
 
+/* CPython 3.6 had initial and undocumented FASTCALL support, but we play it
+ * safe by not relying on implementation details */
+#if PY_VERSION_HEX < 0x03070000
+
+/* Macro for naming a pygame fastcall wrapper function */
+#define PG_FASTCALL_NAME(func) _##func##_fastcall_wrap
+
+/* used to forward declare compat functions */
+#define PG_DECLARE_FASTCALL_FUNC(func, self_type) \
+    static PyObject *PG_FASTCALL_NAME(func)(self_type * self, PyObject * args)
+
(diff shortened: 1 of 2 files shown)


MyreMylar:
image

This build seems to be failing due to it searching for stubs where there will almost certainly not be any stubs + one genuine seeming error. I'm guessing mypy got updated recently.

I've sketched out a PR to try and exclude the troublesome directories. Over here: https://github.com/pygame/pygame/pull/3328

MyreMylar (approved):
Gave this a decent test locally and can confirm the speed up on my Windows PC  - and that it still does what it used to. The color tests also all pass locally so that is good too.

👍

Starbuck5 (approved):
Looks cool!

src_c/_pygame.h
```diff
@@ -281,6 +281,41 @@ supported Python version. #endif */
 
 #define PyType_Init(x) (((x).ob_type) = &PyType_Type)
 
+/* CPython 3.6 had initial and undocumented FASTCALL support, but we play it
+ * safe by not relying on implementation details */
+#if PY_VERSION_HEX < 0x03070000
```
illume, line 286: I think because 3.6 won't change we could use it in 3.6.
Opened an issue for discussion here: https://github.com/pygame/pygame/issues/3527

PR #3326, opened 2022-07-20
Add clamp(val, min, max) function to pygame.math

JamesC01:
I know that it's simple to implement your own clamp function, but clamp seems to be used in games quite a bit. If you're on the fence for whether or not this should be included with pygame, rather than just implemented in python, here are a few points:

- Most other frameworks implement their own clamp function (MonoGame, Unity, Godot, even the C++ standard library)
- There aren't any (that I know of) popular python libraries that have simple functions like clamp.
- Obviously, since pygame.math.clamp is implemented in C, it should be faster than a python implementation. I timed it, and pygame.math.clamp is around 75% faster than an equivalent python implementation, while (I believe) providing all the same behaviour.

The function can take arguments of any type, which may seem strange, but it is equivalent to how it would work if it was implemented in python, _and_ it's much faster than enforcing a specific type, which requires converting the python type to C type, which slows it down a bit.

I'm new to contributing, so I'm sorry if I did something wrong. I couldn't find the test stub generator, so I manually wrote it.


itzpr3d4t0r:
you are missing the math.pyi function declaration + this could be a nice function do make with FASTCALL: info #3324.

JamesC01:
> 

Is That as simple as changing from METH_VARARGS to METH_FASTCALL? Also, how do I generate the pyi?

JamesC01:
> > Is that as simple as changing from METH_VARARGS to METH_FASTCALL? Also, how do I generate the pyi?
> 
> No you have to do a couple of things.
> 
> 1. Yes you have to change `METH_VARARGS` to `METH_FASTCALL`
> 2. You have to change the function signature to `static PyObject *math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)` so now you have `args` which will be a C array of `PyObject*` that are your 3 values. `nargs` represents the number of arguments passed to the function so you use that to validate the correct arguments number.
> 3. Now instead of using `PyObject *val = PyTuple_GET_ITEM(arg, 0);` for example, you write `PyObject *val = args[0];` etc.
> 
> You don't generate a pyi. you have to write it yourself. find the math.pyi file and add your function. if you want to generate docs(like the ones in the .h file) what i do is just write the docs in the .rst file and then run this command: `python setup.py docs --fullgeneration`

Thanks, I'll get to it! What about generating test stubs? I couldn't find the file mentioned in the README.

JamesC01:
> I mean i just add the test to the correct test file. In this case shouldn't it be math_test.py?

Yeah, I added it, but in test/README.md it says to use test/utils/gen_stubs.py to generate test stubs, but the file doesn't exist, so I had to write the test myself. It seems to work, though.

itzpr3d4t0r:
Don't quote me on this but that may be outdated. Nobody went against me on tests so far.

ankith26:
Hello!
First of all thanks for taking the time and effort for getting started with contributing to pygame.

A clamp function was already added as a vector method in #2990

You probably missed it because it's not on the pygame docs as there has been no pygame release since the time the clamp function was added

JamesC01:
> Hello! First of all thanks for taking the time and effort for getting started with contributing to pygame.
> 
> A clamp function was already added as a vector method in #2990
> 
> You probably missed it because it's not on the pygame docs as there has been no pygame release since the time the clamp function was added

Isn't that for clamping vectors? The clamp I added is for clamping floats.

itzpr3d4t0r:
Oh damn i forgot to tell you that FASTCALL is only supported from python 3.7 onwards, so you'll have to do two more things.
Add this under your function definition:
```C
#if PY_VERSION_HEX < 0x03070000
static PyObject *
math_clamp_wrapper(PyObject *self, PyObject *args) {
    return math_clamp(self, (PyObject *const *)PySequence_Fast_ITEMS(args), PyTuple_GET_SIZE(args));
}
#endif
```

src_c/math.c
```diff
@@ -4074,6 +4093,7 @@ math_disable_swizzling(pgVector *self, PyObject *_null)
 }
 
 static PyMethodDef _math_methods[] = {
+    {"clamp", math_clamp, METH_FASTCALL, DOC_PYGAMEMATHCLAMP},
```
itzpr3d4t0r, line 4096: You'll have to also change this
```suggestion
#if PY_VERSION_HEX < 0x03070000
    {"clamp", math_clamp_wrapper, METH_VARARGS, DOC_PYGAMEMATHCLAMP},
#else 
    {"clamp", math_clamp, METH_FASTCALL, DOC_PYGAMEMATHCLAMP},
#endif    
```

ankith26 (approved):
Oh, sorry I misinterpreted the title. This PR is definitely not a duplicate.
Left a few minor reviews

ankith26 (changes requested):
Whoopsies, intended to put my above reviews under "request changes"

src_c/math.c
```diff
@@ -4047,6 +4047,25 @@ vector_elementwise(pgVector *vec, PyObject *_null)
...
+    if (nargs != 3)
+        return RAISE(PyExc_ValueError, "clamp requires 3 float arguments");
+
+    double value = PyFloat_AsDouble(args[0]);
+    double min = PyFloat_AsDouble(args[1]);
+    double max = PyFloat_AsDouble(args[2]);
+
+    if (PyErr_Occurred()) {
+        return RAISE(PyExc_TypeError, "clamp requires 3 float arguments");
+    }
```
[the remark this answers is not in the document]
itzpr3d4t0r: if you do so for 3 times isn't that a performance waste? I've done this too if you consider this a problem. In this way you only do 1 check and save performance if all is right. Ups this was meant to respond to ankith, not start a review

docs/reST/ref/math.rst
```diff
@@ -48,6 +48,13 @@ Multiple coordinates can be set using slices or swizzling
 .. versionchanged:: 1.9.4 Allow scalar construction like GLSL Vector2(2) == Vector2(2.0, 2.0)
 .. versionchanged:: 1.9.4 :mod:`pygame.math` required import. More convenient ``pygame.Vector2`` and ``pygame.Vector3``.
 
+.. function:: clamp
+
+   | :sl:`returns value clamped to min and max.`
+   | :sg:`clamp(value, min, max) -> float`
+
```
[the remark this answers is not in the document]
JamesC01: I did it, can you check to make sure it's right?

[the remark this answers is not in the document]
ankith26: Yes, this is correct

src_c/math.c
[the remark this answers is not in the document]
itzpr3d4t0r: I measured a performance difference that's significant. And here error severity wouldn't be as bad as a memory error + we are already checking for the error right after the function call anyway so there's no extra operation in between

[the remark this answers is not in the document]
itzpr3d4t0r: in one of my functions with 4 checks:
```
4 args:  0.7337168000012753
4 args:  0.6936256000008143
4 args:  0.6863355000023148
```
with 1 check:
```
4 args:  0.7097950999996101
4 args:  0.645052100000612
4 args:  0.6597265000018524
```

src_c/math.c
```diff
@@ -4074,6 +4103,11 @@ math_disable_swizzling(pgVector *self, PyObject *_null)
 }
 
 static PyMethodDef _math_methods[] = {
+#if PY_VERSION_HEX < 0x03070000
+    {"clamp", math_clamp_wrapper, METH_VARARGS, DOC_PYGAMEMATHCLAMP},
+#else
+    {"clamp", math_clamp, METH_FASTCALL, DOC_PYGAMEMATHCLAMP},
```
itzpr3d4t0r, line 4109: You may have to cast to PyCFunction here:
```suggestion
    {"clamp", (PyCFunction)math_clamp_wrapper, METH_VARARGS, DOC_PYGAMEMATHCLAMP},
#else
    {"clamp", (PyCFunction)math_clamp, METH_FASTCALL, DOC_PYGAMEMATHCLAMP},
```

JamesC01 (commented):
Made all the requested changes.

src_c/math.c
itzpr3d4t0r: This of casting to PyCFunction is a compatibility thing. I've made many FASTCALL functions and all had to be cast to PyCFunction, no matter their type.

itzpr3d4t0r:
Also, if you are not in already you should join our Discord: https://discord.gg/JSCwW4fW

JamesC01 (commented):
Implemented requested changes.

src_c/math.c
```diff
@@ -4047,6 +4047,40 @@ vector_elementwise(pgVector *vec, PyObject *_null)
...
+math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
+{
+    const char *err_msg = "clamp requires 3 float arguments";
+
+    if (nargs != 3)
+        return RAISE(PyExc_ValueError, err_msg);
+
+    double value = PyFloat_AsDouble(args[0]);
+    if (PyErr_Occurred())
+        return RAISE(PyExc_TypeError, err_msg);
```
ankith26, line 4060: By setting a new error here, the error already set by `PyFloat_AsDouble` will be discarded. So ideally this should be a `return NULL;` here (which essentially unconditionally forwards the error message)

Even in the case where it is desirable to set a custom error, you must check the older error and forward non-`TypeError`s

JamesC01: I see what you mean. If PyErr_Occured, python has already set an error. Sorry, I didn't actually know anything about the python C api before contributing this, I just figured it out by reading the pygame c files. I'll spend some time to learn it properly if I ever make more complicated contributions. Anyway, I made the change, can you check to make sure I got it right?

ankith26: Yes, you got it right now!

JamesC01: Great! By the way, it still says "changes requested", but I'm pretty sure I've done them all.

itzpr3d4t0r:
Now that #3325 has been merged it is recommended that you use that way to implement a FASTCALL function. Sorry to ask more work from you, but thiss will simplify the code, making it more readable and easier to delete the 3.6 part when we drop compat somewhere in the future.

ankith26 (approved):
I'm happy with this PR now, thanks again for contributing to pygame 🎉 🥳 

The `FASTCALL` macro changes can happen on another PR (we already discussed this on discord)

All we need to do here, is wait for a few more peoples approvals (since this is new API)

itzpr3d4t0r:
Is this any faster?

itzpr3d4t0r:
Actually i found out that you can make it even faster, i'll give you the code i used to test and some results:

### Proposed change
my proposed change/reorganisation:
static PyObject *
```C
math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
{
    double value, min, max;
    if (nargs != 3)
        return RAISE(PyExc_ValueError, "clamp requires 3 float arguments");

    value = PyFloat_Check(args[0]) ? PyFloat_AS_DOUBLE(args[0])
                                   : (double)PyLong_AsLong(args[0]);
    if (PyErr_Occurred())
        return NULL;
    min = PyFloat_Check(args[1]) ? PyFloat_AS_DOUBLE(args[1])
                                 : (double)PyLong_AsLong(args[1]);
    if (PyErr_Occurred())
        return NULL;
    max = PyFloat_Check(args[2]) ? PyFloat_AS_DOUBLE(args[2])
                                 : (double)PyLong_AsLong(args[2]);
    if (PyErr_Occurred())
        return NULL;

    return PyFloat_FromDouble(MIN(max, MAX(min, value)));
}
```

___
The results:
**CURRENT**
```
====||MATH.CLAMP||====
-) 1: 0.68651996
-) 2: 0.7033555
-) 3: 0.69078892
______________
-) Mean time: 0.69355479

====||PYTHON  CLAMP||====
-) 1: 2.17853738
-) 2: 2.17602866
-) 3: 2.10800866
______________
-) Mean time: 2.15419157
```
**WITH MY PROPOSED CHANGE**
```
====||MATH.CLAMP||====
-) 1: 0.55778852
-) 2: 0.56320468
-) 3: 0.56477658
______________
-) Mean time: 0.56192326

====||PYTHON CLAMP||====
-) 1: 2.18498714
-) 2: 2.22100118
-) 3: 2.13322032
______________
-) Mean time: 2.17973621
```
The code:

```Python
import timeit
from statistics import fmean

from pygame.math import clamp


def myclamp(x, minv, maxv):
    return min(maxv, max(minv, x))


NUMBER = 10_000_000
tottime = 0
GLOB = {"clamp": clamp, "myclamp": myclamp}

def test(test_name: str, func: str) -> float:
    value = fmean(timeit.repeat(func, globals=GLOB, number=NUMBER))
    print("-) " + test_name + f": {round(value, 8)}")
    return value


def test_group(
        group_name: str,
        sequence: list[str],
        tests_names: tuple[str, ...] = None,
        include_tot: bool = False,
        include_mean: bool = True
) -> float:
    print("\n====||" + group_name.upper() + "||====")
    
    test_data = [test(tests_names[i] if tests_names else str(i + 1), test_obj)
                 for
                 i, test_obj in
                 enumerate(sequence)]
    
    print("______________")
    if include_tot:
        print(f"-) Total time: {round(sum(test_data), 8)}")
    
    if include_mean:
        print(f"-) Mean time: {round(fmean(test_data), 8)}")
    
    return fmean(test_data)


test_group(
    "math.clamp",
    ["clamp(1, 15, 17)",
     "clamp(22, 15, 17)",
     "clamp(8, 4, 17)",
     ])

test_group(
    "python clamp",
    ["myclamp(1, 15, 17)",
     "myclamp(22, 15, 17)",
     "myclamp(8, 4, 17)"]
)
```

JamesC01:
@itzpr3d4t0r Thanks!

MyreMylar (approved):
LGTM. Gave it a good round of testing locally and ran the unit tests provided.

Approved! 👍 💯 🥂 🎉

JamesC01:
> LGTM. Gave it a good round of testing locally and ran the unit tests provided.
> 
> Approved! 👍 💯 🥂 🎉

Thanks! Did you take a look at the unit tests? I'm very knew to this, so I'm not sure if they're very good. They seem to work. I just check different cases, making sure they give the correct value, but since I'm new, I could be missing something.

MyreMylar:
> > LGTM. Gave it a good round of testing locally and ran the unit tests provided.
> > Approved! 👍 💯 🥂 🎉
> 
> Thanks! Did you take a look at the unit tests? I'm very knew to this, so I'm not sure if they're very good. They seem to work. I just check different cases, making sure they give the correct value, but since I'm new, I could be missing something.

I did take a look, they seemed fine. Could maybe have tested the invalid class/object case, but otherwise good 👍

PR #3328, opened 2022-07-22
Extend the set of directories not to look for stubs in

MyreMylar:
Usually these files are not part of the public interface to pygame and thus don't need type stubs.

Hopefully will get us closer to the stub test CI passing.


MyreMylar:
Well this didn't work.

PR #3329, opened 2022-07-22
Restore correct checksum

itzpr3d4t0r:
Fixes the current AppVeyor build problems we're having. Cause was the just updated SDL ttf prebuilt.


Starbuck5 (approved):
Yes!

Starbuck5:
This is why the checksum changed btw: https://github.com/libsdl-org/SDL_ttf/issues/233

ankith26 (approved):
Thankies! 😎 🎉

PR #3330, opened 2022-07-23
Port new SDL_ttf text shaping capability

Starbuck5:
At least on supported platform (Windows) (since we haven't updated to a new enough SDL_ttf on other platforms-- also we'll need to talk care to built SDL_ttf with harfbuzz now).


MyreMylar:
On my machine trying to render: 
# प्रफुल्ल

Output _without_ `set_script()`:

image

Output with `set_script("Deva")`:

image

Seems like a big improvement!

MyreMylar (approved):
LGTM 👍

src_c/font.c
```diff
@@ -664,11 +696,10 @@ static PyMethodDef font_methods[] = {
      DOC_FONTGETSTRIKETHROUGH},
     {"set_strikethrough", font_set_strikethrough, METH_O,
      DOC_FONTSETSTRIKETHROUGH},
-
```
zoldalma999, line 667: really minor nitpick, but not sure why these whitespaces got removed

docs/reST/ref/font.rst
```diff
@@ -472,6 +472,23 @@ solves no longer exists, it will likely be removed in the future.
...
 
+   .. method:: set_script
+
+      | :sl:`set the script code for text shaping`
+      | :sg:`set_script(str) -> None`
+
+      **Experimental:** feature still in development available for testing and feedback. It may change.
+      `Please leave feedback with authors <https://github.com/pygame/pygame/pull/3330>`_
+
+      Sets the script used by harfbuzz text shaping, taking a 4 character
```
zoldalma999, line 483: Docs should probably mention that this function is only supported on SDL_ttf 2.20 or higher

zoldalma999 (commented):
Looks good to me apart from these two (minor) nitpicks. Ran the test script locally and got the same result as MyreMylar.

src_c/font.c
Starbuck5: Because they shouldn't have been there in the first place? They seem fully extraneous to me.

docs/reST/ref/font.rst
Starbuck5: Done

src_c/font.c
zoldalma999: I think they made sense, but it does not matter that much

zoldalma999 (approved):
Looks good to me!

PR #3332, opened 2022-07-23
Fix latest stubtest issues

ankith26:
This PR picks up where #3328 left

After another mypy update, there are new issues that keep our Ubuntu 22.04 CI failing

stubtest started complaining about missing stubs on internal modules, those are added to the mypy allowlist.
I also updated `gen_stubs` to make stubs for `pygame.locals` (previously mypy picked these from python code, but the new version seems to be pedantic about stubs being in there too). 
Apart from that, I had a few other minor stub improvements to make, and those are in here


src_py/__init__.py
```diff
@@ -28,14 +28,14 @@
 
 # Choose Windows display driver
 if os.name == "nt":
-    pygame_dir = os.path.split(__file__)[0]
+    _pygame_dir = os.path.split(__file__)[0]
```
zoldalma999, line 31: Another solution might be to just `del` this variable once we are done using it. But this works too

MyreMylar (approved):
Hooray! It works.

Thanks for fixing this when I failed :)

zoldalma999 (approved):
Left a little suggestion, but apart from that, looks good to me.

PR #3333, opened 2022-07-23
math.distance(), math.distance_squared() functions

itzpr3d4t0r:
## Info
Follows #3324. This one implements the distance and distance_squared functions for calculating distances between two non vector points. An example could be
```Python
from pygame.math import distance

print(distance((10, 10), (20, 20))) 
```
I made this PR because it's strange that if you want a distance function in pygame you either write it yourself in python or you must use two Vector objects. With this you just import distance and you're set.
## Performance 
From what i've tested it appears we would be a bit better than what a Cython implementation would do, while being massively better than what a Python implementation could do. With the code below i got these results
```
pygame vector:  0.9923575600005279
pygame vector (squared):  0.9553661199992348
pygame 2:  0.7641494199997396
pygame 4:  0.7759733399987454
pygame 2 (squared):  0.7611866199993529
pygame 4 (squared):  0.9043506800000614
sqrt 2:  2.477797339999961
sqrt 4:  1.9382891799999924
cython 2:  0.8770705599992652
cython 4:  0.8885841600000276
cython 2 raw:  0.893820740000956
cython 4 raw:  0.8123775599997316
```
**The code** 

```Python
import timeit
from math import sqrt
from statistics import fmean

import pygame
from pygame.math import distance, distance_squared

from test_cy import distance as cdistance, distance4 as cdistance4, distanceraw, \
    distanceraw4


def sqrtdistance(p1, p2):
    dx = p2[0] - p1[0]
    dy = p2[1] - p1[1]
    return sqrt(dx * dx + dy * dy)


def sqrtdistance4(x1, y1, x2, y2):
    dx = x2 - x1
    dy = y2 - y1
    return sqrt(dx * dx + dy * dy)


NUM = 10000000
v1 = pygame.math.Vector2(10, 10)
v2 = pygame.math.Vector2(20, 20)

GLOB2 = {"distance": distance, "cdistance": cdistance,
         "cdistance4": cdistance4, "sqrtdistance": sqrtdistance,
         "sqrtdistance4": sqrtdistance4, "v1": v1, "v2": v2,
         "distance_squared": distance_squared, "distanceraw": distanceraw,
         "distanceraw4": distanceraw4}

print("pygame vector: ",
      fmean(timeit.repeat("v1.distance_to(v2)", globals=GLOB2,
                          number=NUM)))
print("pygame vector (squared): ",
      fmean(timeit.repeat("v1.distance_squared_to(v2)", globals=GLOB2,
                          number=NUM)))

print("pygame 2: ",
      fmean(timeit.repeat("distance((10, 10), (20, 20))", globals=GLOB2,
                          number=NUM)))
print("pygame 4: ",
      fmean(
          timeit.repeat("distance(10, 10, 20, 20)", globals=GLOB2,
                        number=NUM)))

print("pygame 2 (squared): ",
      fmean(timeit.repeat("distance_squared((10, 10), (20, 20))",
                          globals=GLOB2,
                          number=NUM)))
print("pygame 4 (squared): ",
      fmean(
          timeit.repeat("distance_squared(10, 10, 20, 20)", globals=GLOB2,
                        number=NUM)))

print("sqrt 2: ",
      fmean(timeit.repeat("sqrtdistance((10, 10), (20, 20))", globals=GLOB2,
                          number=NUM)))
print("sqrt 4: ",
      fmean(timeit.repeat("sqrtdistance4(10, 10, 20, 20)", globals=GLOB2,
                          number=NUM)))

print("cython 2: ",
      fmean(timeit.repeat("cdistance((10, 10), (20, 20))", globals=GLOB2,
                          number=NUM)))
print("cython 4: ",
      fmean(timeit.repeat("cdistance4(10, 10, 20, 20)", globals=GLOB2,
                          number=NUM)))

print("cython 2 raw: ",
      fmean(timeit.repeat("distanceraw((10, 10), (20, 20))", globals=GLOB2,
                          number=NUM)))
print("cython 4 raw: ",
      fmean(timeit.repeat("distanceraw4(10, 10, 20, 20)", globals=GLOB2,
                          number=NUM)))
```
And these are the Cython implementations:
```Python
from libc.math cimport hypot, sqrt

cpdef double distance4(double x1, double y1, double x2, double y2):
    return hypot(x2-x1, y2-y1)

cpdef double distance(p1, p2):
    return hypot(p2[0] - p1[0], p2[1] - p1[1])

cpdef double distanceraw4(double x1, double y1, double x2, double y2):
    cdef double dx =  x2 - x1, dy = y2 - y1
    return sqrt(dx*dx + dy*dy)

cpdef double distanceraw(p1, p2):
    cdef double dx =  p2[0] - p1[0], dy = p2[1] - p1[1]
    return sqrt(dx*dx + dy*dy)
```


Starbuck5 (changes requested):
I would vote against adding this API, for several reasons

Most importantly, python already has a built in `math.dist`

Secondly, if we add one vector helper to the top level of pygame.math, why not add all of them? That's a lot of API to add for not much benefit, in my opinion.

Thirdly, just use pygame vectors!

I feel that points 2 and 3 are arguable, but point number 1 is the real kicker.

ankith26:
I think #3335 supersedes this PR by taking the other route of optimising the existing vector method itself

PR #3335, opened 2022-07-23
Optimized _vector_distance_helper

itzpr3d4t0r:
One of #3343 
Follows a discussion we had on Discord. This PR optimizes the _vector_distance_helper() function, hence optimizing every other function that uses this one. 
Optimizations are mainly dedicated to vector-vector distance calculations, improving performance by 31%, but slightly improve(9%) vector-sequence distance calculation as well.
Some performance tests:
**NEW**
```
pygame vector:  0.7792468000001463
pygame vector (squared):  0.7629703400000836
pygame vector tup:  0.9840590200000406
pygame vector (squared) tup:  0.9708823000000848
```
**OLD**
```
pygame vector:  1.0366728799997873
pygame vector (squared):  1.0292829800000618
pygame vector tup:  1.0955151999999544
pygame vector (squared) tup:  1.0389712800000779
```

**Code**
```Python
import timeit
from statistics import fmean
import pygame


NUM = 10000000
v1 = pygame.math.Vector2(10, 10)
v2 = pygame.math.Vector2(20, 20)

GLOB2 = {"v1": v1, "v2": v2}

print("pygame vector: ", fmean(timeit.repeat("v1.distance_to(v2)", globals=GLOB2, number=NUM)))

print("pygame vector (squared): ", fmean(timeit.repeat("v1.distance_squared_to(v2)", globals=GLOB2, number=NUM)))

print("pygame vector tup: ", fmean(timeit.repeat("v1.distance_to((20, 20))", globals=GLOB2, number=NUM)))

print("pygame vector (squared) tup: ", fmean(timeit.repeat("v1.distance_squared_to((20, 20))", globals=GLOB2, number=NUM)))
```

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -1633,18 +1633,52 @@ vector_reflect_ip(pgVector *self, PyObject *normal)
 static double
 _vector_distance_helper(pgVector *self, PyObject *other)
 {
-    Py_ssize_t i;
-    double distance_squared, tmp;
+    Py_ssize_t i, dim = self->dim;
+    double distance_squared = 0;
 
-    distance_squared = 0;
-    for (i = 0; i < self->dim; ++i) {
-        tmp = PySequence_GetItem_AsDouble(other, i) - self->coords[i];
-        distance_squared += tmp * tmp;
+    /* Specialised fastpath for Vector-Vector distance calculation*/
+    if (pgVector_Check(other)) {
+        pgVector *otherv = (pgVector *)other;
+        double dx, dy;
+
+        if (dim != otherv->dim) {
+            PyErr_SetString(PyExc_TypeError, "Vectors must be the same size");
+            return -1;
+        }
+
+        dx = otherv->coords[0] - self->coords[0];
+        dy = otherv->coords[1] - self->coords[1];
+
+        distance_squared = dx * dx + dy * dy;
+
+        if (dim == 3) {
+            double dz;
+            dz = otherv->coords[2] - self->coords[2];
+            distance_squared += dz * dz;
+        }
     }
-    /* PySequence_GetItem_AsDouble can fail in which case it will set an Err */
-    if (PyErr_Occurred())
-        return -1;
+    /* Vector-Sequence distance calculation*/
+    else {
+        double tmp;
+        PyObject *fast_seq = PySequence_Fast(other, "A sequence was expected");
+        if (!fast_seq) {
+            return -1;
+        }
 
+        if (PySequence_Fast_GET_SIZE(fast_seq) != dim) {
+            PyErr_SetString(PyExc_TypeError,
+                            "Vector and sequence must be the same size");
+            return -1;
+        }
+
+        for (i = 0; i < dim; ++i) {
+            tmp = PyFloat_AsDouble(PySequence_Fast_GET_ITEM(fast_seq, i)) -
+                  self->coords[i];
+            if (PyErr_Occurred())
+                return -1;
+            distance_squared += tmp * tmp;
+        }
+    }
     return distance_squared;
 }
 
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -1642,7 +1642,7 @@ _vector_distance_helper(pgVector *self, PyObject *other)
         double dx, dy;
 
         if (dim != otherv->dim) {
-            PyErr_SetString(PyExc_TypeError, "Vectors must be the same size");
+            PyErr_SetString(PyExc_ValueError, "Vectors must be the same size");
             return -1;
         }
 
@@ -1666,7 +1666,7 @@ _vector_distance_helper(pgVector *self, PyObject *other)
         }
 
         if (PySequence_Fast_GET_SIZE(fast_seq) != dim) {
-            PyErr_SetString(PyExc_TypeError,
+            PyErr_SetString(PyExc_ValueError,
                             "Vector and sequence must be the same size");
             return -1;
         }
--- a/test/math_test.py
+++ b/test/math_test.py
@@ -562,21 +562,49 @@ class Vector2TypeTest(unittest.TestCase):
     def test_distance_to(self):
         diff = self.v1 - self.v2
         self.assertEqual(self.e1.distance_to(self.e2), math.sqrt(2))
+        self.assertEqual(self.e1.distance_to((0, 1)), math.sqrt(2))
+        self.assertEqual(self.e1.distance_to([0, 1]), math.sqrt(2))
         self.assertAlmostEqual(
             self.v1.distance_to(self.v2), math.sqrt(diff.x * diff.x + diff.y * diff.y)
         )
+        self.assertAlmostEqual(
+            self.v1.distance_to(self.t2), math.sqrt(diff.x * diff.x + diff.y * diff.y)
+        )
+        self.assertAlmostEqual(
+            self.v1.distance_to(self.l2), math.sqrt(diff.x * diff.x + diff.y * diff.y)
+        )
         self.assertEqual(self.v1.distance_to(self.v1), 0)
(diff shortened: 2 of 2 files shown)


src_c/math.c
```diff
@@ -1633,18 +1633,49 @@ vector_reflect_ip(pgVector *self, PyObject *normal)
...
     }
-    /* PySequence_GetItem_AsDouble can fail in which case it will set an Err */
-    if (PyErr_Occurred())
-        return -1;
+    else {
+        double tmp;
+        if (!(PyTuple_Check(other) || PyList_Check(other))) {
+            PyErr_SetString(PyExc_TypeError, "A sequence was expected");
+            return -1;
+        }
```
ankith26, line 1664:
```suggestion
        PyObject *fast_seq = PySequence_Fast(other, "A sequence was expected");
        if (!fast_seq) {
            return -1;
        }
```

I think we should be supporting generic sequences here, not just tuples and lists... I wouldn't mind if doing so reduces performance for this case by a small bit.
The code I suggested should work for all sequences (not sure about other things like generators and non-sequence iterables, but the old code did not support em anyways) and I'm pretty sure this also has an internal cpython fastpath for tuples and lists

src_c/math.c
```diff
@@ -1633,18 +1633,49 @@ vector_reflect_ip(pgVector *self, PyObject *normal)
...
+        if (PySequence_Fast_GET_SIZE(other) != dim) {
+            PyErr_SetString(PyExc_TypeError,
+                            "Vector and sequence must be the same size");
+            return -1;
+        }
 
+        for (i = 0; i < self->dim; ++i) {
```
ankith26, line 1671: Just a minor nitpick, this can be just `dim` so that it is consistent with other code in this function

src_c/math.c
```diff
@@ -1633,18 +1633,49 @@ vector_reflect_ip(pgVector *self, PyObject *normal)
 static double
 _vector_distance_helper(pgVector *self, PyObject *other)
 {
-    Py_ssize_t i;
-    double distance_squared, tmp;
+    Py_ssize_t i, dim = self->dim;
+    double distance_squared = 0;
+
+    if (pgVector_Check(other)) {
```
ankith26, line 1639: A comment that explains that this is a fastpath for performance reasons would help here, so that people looking at this code a few years down the line don't think that this is redundant and remove it

src_c/math.c
itzpr3d4t0r: I said it once and i'll say it again. PySequence_Fast also converts strings into "PySequence_Fast objects", and implementing what you asked me( i did it and tested it) only gives you a segfault. At that point to resolve it you have to add a string check and exit with error if that's the case.
But at that point you are doing the same as before (basically a two checks + one conversion) so you are back where you started. I don't agree with this and just like you said:

> not sure about other things like generators and non-sequence iterables, but the old code did not support em anyways

yes you are right here that we did not support them before so the current way is clearly the best one, just 2 fast checks amd everything stays the same as before.

itzpr3d4t0r: actually i think i finally understood how this works xD

src_c/math.c
itzpr3d4t0r: ups missed this one

src_c/math.c
ankith26: Strings are sequences, `PySequence_Fast` is doing the right thing. But ofc, `PyFloat_AsDouble` should correctly set an error if it gets characters.

itzpr3d4t0r: yeah now it works fine, just needed to use the converted string sequence

ankith26 (approved):
LGTM! Not only does this PR make speedups, I also like how it improves error handling by properly raising errors on mismatching dimensions
Thanks 🎉 😎

src_c/math.c
```diff
@@ -1633,18 +1633,52 @@ vector_reflect_ip(pgVector *self, PyObject *normal)
...
+    Py_ssize_t i, dim = self->dim;
+    double distance_squared = 0;
+
+    /* Specialised fastpath for Vector-Vector distance calculation*/
+    if (pgVector_Check(other)) {
+        pgVector *otherv = (pgVector *)other;
+        double dx, dy;
+
+        if (dim != otherv->dim) {
+            PyErr_SetString(PyExc_TypeError, "Vectors must be the same size");
```
Starbuck5, line 1645: Should this be a ValueError instead? (same with the other one)

Starbuck5 (commented):
I've just realized that there should be some test updates with this PR.

Testing the exceptions raised
Testing tuples vs vectors in these methods, since the code path is now different

Starbuck5:
```py
>>> a =pygame.Vector2(10,10)
>>> a.distance_to((20,20))
14.142135623730951
>>> a.distance_to((20,20,30))
14.142135623730951
```

So previously this "worked." I'm not classifying this as back compat, I'd call that a bug. But there should definitely be a test to prevent future regressions.

Also, I took my own go at optimizing this further, and failed. I tried using SIMD to do both of the float multiplications at once, which did not help at all. I found stuff online talking about how modern CPUs usually have two FPUs they can use at once (floating point units), so that's probably happening already. I also tried using the `inline` keyword to compile the function into its callers, but that also yielded 0 change (I think the compiler is probably doing it already).

Starbuck5 (commented):
I believe there still is no functional test for vectors vs tuples for the various distance_to scenarios.

itzpr3d4t0r:
> I believe there still is no functional test for vectors vs tuples for the various distance_to scenarios.

Can you be clearer? what do you want?

Starbuck5 (approved):
Thanks! Looks good to me.

src_c/math.c
```diff
@@ -1633,18 +1633,52 @@ vector_reflect_ip(pgVector *self, PyObject *normal)
...
-    distance_squared = 0;
-    for (i = 0; i < self->dim; ++i) {
-        tmp = PySequence_GetItem_AsDouble(other, i) - self->coords[i];
-        distance_squared += tmp * tmp;
+        dx = otherv->coords[0] - self->coords[0];
+        dy = otherv->coords[1] - self->coords[1];
+
+        distance_squared = dx * dx + dy * dy;
+
+        if (dim == 3) {
```
illume, line 1654: In the future, probably we should use dim loops for all code if it's doable. Because of a future Vector4/VectorN.

src_c/math.c
```diff
@@ -1633,18 +1633,52 @@ vector_reflect_ip(pgVector *self, PyObject *normal)
...
-    double distance_squared, tmp;
+    Py_ssize_t i, dim = self->dim;
+    double distance_squared = 0;
+
+    /* Specialised fastpath for Vector-Vector distance calculation*/
+    if (pgVector_Check(other)) {
+        pgVector *otherv = (pgVector *)other;
+        double dx, dy;
+
+        if (dim != otherv->dim) {
```
illume, line 1644: This should probably have used: `pgVectorCompatible_Check(otherv, dim)` ?

itzpr3d4t0r: I didn't use it because of 2 reasons:
- Using it would have forced me to do duplicate checks and possibly duplicate sequence obj extractions as that function doesn't tell you if the object is a vector or sequence
- It would defeat the purpose of having a fast path for vectors and one for sequences

Plus can you explain why we need to check for RealNumberCheck for each element in the sequence? PyFloat_AsDouble errors out if a complex number is passed as it can't convert to double(tested it) and so it should be enough to either check that at vector creation/assignment or not check it at all since it doesn't appear to be an issue.

src_c/math.c
itzpr3d4t0r: > Or error out correctly if the code doesn't handle other dims.

Doesn't it already?

> In the future, probably we should use dim loops for all code if it's doable. Because of a future Vector4/VectorN.

Yeah this would have been better indeed

Issue #2972, opened 2022-01-05
Allow image_frombuffer to understand cairos premultiplied ARGB32 


stuaxo:
It would be good if image_frombuffer could understand pycairos premultiplied ARGB32 format, currently it seems impossible to use, as pygame is doing some byteswapping and it's not obvious how to enable alpha premultiplication.

In detail:

@hanysz found an interesting issue using pycairo to create surfaces for pygame https://github.com/pygobject/pycairo/issues/247

Cairos ARGB32 format stores ARGB starting in the high byte, with alpha premultiplied.

The [docs](https://www.pygame.org/docs/ref/image.html?highlight=frombuffer#pygame.image.frombuffer) make it look like image_frombuffer should be able to handle this format, but unfortunately there is some 
undesired byteswapping happening.

The [implementation](https://github.com/pygame/pygame/blob/main/src_c/image.c#L1092) calls SDL_CreateRGBSurfaceFrom, which looks like it could be given the right parameters to understand the format.

@hanysz test prog creates an image in cairo, rendering text in red, green and blue and renders in pygame - but only the "red" text is output, in blue:

![image](https://user-images.githubusercontent.com/179677/148146875-213cdb39-a2ad-4df8-9fc0-c97ef8a5bb7b.png)

The feature request:

Possibly this needs a new format: native endian RGBA32 with alpha premultiplication, or maybe it's a matter of providing a method of disabling the byteswapping.

Starbuck5:
Hello, I've started to do some research on this issue.

I found a couple people with various integration solutions: (They all seem complicated 😅)
https://stackoverflow.com/questions/11640161/byte-order-when-using-cairo-in-pygame-sdl
https://www.pygame.org/wiki/CairoPygame

I also found a Cairo pygame on the matter:
https://www.cairographics.org/SDL/
They point to using another C library to handle the interop, which was last updated in 2009. I find it kind of funny that we are both here as Python libraries trying to do better integration than our respective C libraries.

In terms of byte swapping, I devised a test program and it seems like Cairo is doing the swapping.
Test program:
```py
import pygame
import cairo

surf2 = cairo.ImageSurface(cairo.FORMAT_ARGB32, 2, 1)
ctx = cairo.Context(surf2)
ctx.rectangle(0, 0, 2, 1)
ctx.set_source_rgb(0, 0.5, 0.5)
ctx.fill_preserve()
print(bytes(surf2.get_data()))
print([int(i) for i in bytes(surf2.get_data())])

[...]

PR #3342, opened 2022-07-25
BGRA format added to fromstring & tostring

rlatowicz:


Starbuck5:
Hey, sorry no one has had the time to review this thoroughly yet. Just wanted to give you a reply so you know somebody is tracking 😅.

Everything I've looked at has looked good so far. Testing changes seem elegant, C code changes seem straightforward.

rlatowicz:
@Starbuck5 

Thanks.
I appreciate the update.

rlatowicz:
Hi,

What is the best way to resolve the conflict?

Since I made the PR, aliases have been added to **fromstring (frombytes)** & **tostring (tobytes)**

The **alias** addition has changed all the files I had edited,

**image.c
image.pyi
image.rst
image_test.py**

Probably best to cancel the PR and do a redo?

test/image_test.py
```diff
@@ -561,37 +560,35 @@ def RotateARGBtoRGBA(str_buf):
...
+            )
 
-        self._assertSurfaceEqual(
-            test_surface, test_from_argb_string, '"RGBA" rotated to "ARGB" failed'
-        )
+            self._assertSurfaceEqual(
+                test_surface,
+                test_rotate_functions,
+                "converting {} to {} and back is not symmetric".format(
+                    pair[0], pair[1]
```
Starbuck5, line 574: We support python 3.6 and above, so f strings are preferred to use.

test/image_test.py
```diff
@@ -561,37 +560,35 @@ def RotateARGBtoRGBA(str_buf):
...
+                ),
+            )
 
-        argb_buf = pygame.image.tostring(test_surface, "ARGB")
-        rgba_buf = RotateARGBtoRGBA(argb_buf)
-        test_to_argb_string = pygame.image.fromstring(
-            rgba_buf, test_surface.get_size(), "RGBA"
-        )
+        for pair in fmt_permutations:
```
Starbuck5, line 578: I'm confused about the difference between this test loop and the test loop above, which is for format combinations. I see they use different lists, and the top one has an additional convert step.

Inside the loops they have variable names like `rgba_buf`, which seems misleading, because I'm pretty sure they won't always be RGBA.

test/image_test.py
```diff
@@ -523,28 +523,27 @@ def _assertSurfaceEqual(self, surf_a, surf_b, msg=None):
...
-                byte_buf[i * 4 + 1] = byte_buf[i * 4 + 0]
-                byte_buf[i * 4 + 0] = alpha
-            return tostring(byte_buf)
+        import itertools
 
-        ####################################################################
-        def RotateARGBtoRGBA(str_buf):
+        fmts = ("RGBA", "ARGB", "BGRA")
+        fmt_permutations = list(itertools.permutations(fmts, 2))
+        fmt_combinations = list(itertools.combinations(fmts, 2))
```
Starbuck5, line 530: Nitpicky thing: I believe these can be left as generators, rather than converted to lists.

src_c/image.c
```diff
@@ -753,6 +753,87 @@ image_tostring(PyObject *self, PyObject *arg)
         }
         pgSurface_Unlock(surfobj);
     }
+    else if (!strcmp(format, "BGRA")) {
+        hascolorkey = 0;
```
Starbuck5, line 757: Such a suspicious line of code, but you're doing it right, you're copying the behavior of the other ones. This is not a request for a code change or anything, this is ~~a cry for help~~ just a comment.

src_c/image.c
```diff
@@ -753,6 +753,87 @@ image_tostring(PyObject *self, PyObject *arg)
...
+                        color = ptr[2] + (ptr[1] << 8) + (ptr[0] << 16);
+#endif
+                        ptr += 3;
+                        data[2] = (char)(((color & Rmask) >> Rshift) << Rloss);
+                        data[1] = (char)(((color & Gmask) >> Gshift) << Gloss);
+                        data[0] = (char)(((color & Bmask) >> Bshift) << Bloss);
+                        data[3] = (char)(Amask ? (((color & Amask) >> Ashift)
+                                                  << Aloss)
+                                               : 255);
+                        data += 4;
```
Starbuck5, line 814: You're doing this right by copying the other ones, but I've just realized that it might be more performant to do the data stores and increments in one operation, that might generate faster code.

```
*data++ = (char)(((color & Bmask) >> Bshift) << Bloss);
*data++ = (char)(((color & Gmask) >> Gshift) << Gloss);
*data++ = (char)(((color & Rmask) >> Rshift) << Rloss);
*data++ = (char)(Amask ? (((color & Amask) >> Ashift) << Aloss) : 255);
```

Maybe not something for this PR, but it would interesting to test this.

test/image_test.py
rlatowicz: Yes.
I started with f strings and then noticed a `format()` in existing nearby code so swapped them out :)

2022-07-31 14_10_13-D__Aa2_temp6_pygame_test_image_test py - SciTE

test/image_test.py
rlatowicz: Apologies.
Yes, the naming is now confusing. 
Those were the previous names that made sense when there was only 2 formats.
I was focusing on the code working and not breaking anything.
Forgot to change the names to reflect the multiple formats.

In terms of perms vs combs:
In the combinations loop there is a 2 way conversion. Permutations here would be redundant.
The permutations loop is only performing a 1 way conversion.

test/image_test.py
rlatowicz: I listified them for debugging purposes (easier to print).
Forgot to remove that :)

src_c/image.c
rlatowicz: I mention here [later PR](https://github.com/pygame/pygame/issues/2972#issuecomment-1193689297) that I couldn't directly use `tostring_surf_32bpp()`, which can be faster. 
It can be adapted to the case of BGRA.

rlatowicz:
@Starbuck5 

I hope to make the changes to those areas, you have referred to, later tonight or tomorrow.

Thanks.

src_c/image.c
Starbuck5: I did see that, but that seems complicated. Doing the data incrementing like this could potentially be a speed up for other bit depths though, besides 32bpp.

rlatowicz:
@Starbuck5 

Thanks again.

I also noticed that `test_frombuffer_BGRA()` needed some name changes.

Done!

src_c/image.c
rlatowicz: I can benchmark it here and see what difference it makes, at least, to my local setup.
Definitely worth exploring.
Of course, we have all seen optimisations that inexplicably make things slower on another platform/s.

Like you say, the indexing approach is used in many critical loops in `image.c` .

test/image_test.py
Starbuck5: The new variables names help the understandability a lot, nice!

Starbuck5 (approved):
Looks good to me! Thanks for doing the work on this!

I'm interested to see if that performance idea does anything, it could probably be applied widely if it pans out.

src_c/image.c
Starbuck5: Just to be clear, this hunch of mine is not based on the access cost, I think the incrementing may be faster this way.

When I took a class learning 8 bit assembly there were specific instructions for loading and storing that would also increment the location pointer by 1 at the same time, there may be some similar effect through pipelining/x86 instructions/compiler magic.

MyreMylar (approved):
LGTM 👍

src_c/image.c
rlatowicz: @Starbuck5 

I'm not sure if this is the best place to continue the discussion on optimisation or create a new thread in the forum?
___________________

**Auto Increment addressing mode** - one op does access and increment (or decrement) at once,
[ref](http://www.cs.iit.edu/~cs561/cs350/addressing/addressing.html)

But not on Intel CPUs,
[stackoverflow](https://stackoverflow.com/questions/49034893/post-increment-and-pre-decrement-addressing-modes-on-intel-80386)
___________________

I put together a quick benchmark to test attempts at speedup,
As an example, I tested the current indexing method vs pointer incrementing,

```
#include <sys/time.h>
#include <stdio.h>
#include <stdlib.h>

typedef unsigned char Uint8;

typedef struct RGBA {
  Uint8 r;
  Uint8 g;
  Uint8 b;
  Uint8 a;
} RGBA_t;

void init_buffer(RGBA_t *b, size_t width, size_t height) {
  size_t i = 0;
  for (size_t h = 0; h < height; ++h) {
    for (size_t w = 0; w < width; ++w) {
      b[i].r = i % 256;
      b[i].g = (i+1) % 256;
      b[i].b = (i+2) % 256;
      b[i].a = (i+3) % 256;
      i++;
    }
  }
}

void print_buffer(RGBA_t *b, size_t width, size_t height) {
  size_t i = 0;
  for (size_t h = 0; h < height; ++h) {
    for (size_t w = 0; w < width; ++w) {
      printf("(%d,%d,%d,%d) ", b[i].r, b[i].g, b[i].b, b[i].a);
      i++;
    }
  printf("\n");
  }
}

void process_1(RGBA_t *b, size_t width, size_t height) {
  RGBA_t *ptr = b;
  Uint8 *data = (Uint8 *)b;
  for (size_t h = 0; h < height; ++h) {
    for (size_t w = 0; w < width; ++w) {
      RGBA_t color = *ptr++;
      data[0] = color.b;
      data[1] = color.g;
      data[2] = color.r;
      data[3] = color.a;
      data += 4;
    }
  }
}

void process_2(RGBA_t *b, size_t width, size_t height) {
  RGBA_t *ptr = b;
  Uint8 *data = (Uint8 *)b;
  for (size_t h = 0; h < height; ++h) {
    for (size_t w = 0; w < width; ++w) {
      RGBA_t color = *ptr++;
      *data++ = color.a;
      *data++ = color.r;
      *data++ = color.g;
      *data++ = color.b;
    }
  }
}

int main() {
  size_t width = 10000;
  size_t height = 10000;
  
  long seconds, microseconds, elapsed;
  struct timeval begin, end;
  
  RGBA_t *buf = calloc(width*height, sizeof(RGBA_t));
  
  // initialize RGBA buffer
  init_buffer(buf, width, height);
  // print_buffer(buf, width, height);
  
  // times process_1
  gettimeofday(&begin, 0);
  process_1(buf, width, height);
  gettimeofday(&end, 0);
  
  seconds = end.tv_sec - begin.tv_sec;
  microseconds = end.tv_usec - begin.tv_usec;
  elapsed = seconds*1e6 + microseconds;
  printf("%.3f ms | process_1\n", elapsed/1000.0);
  // print_buffer(buf, width, height);
  
  // time process_2
  gettimeofday(&begin, 0);
  process_2(buf, width, height);
  gettimeofday(&end, 0);
  
  seconds = end.tv_sec - begin.tv_sec;
  microseconds = end.tv_usec - begin.tv_usec;
  elapsed = seconds*1e6 + microseconds;
  printf("%.3f ms | process_2\n", elapsed/1000.0);
  // print_buffer(buf, width, height);
  
  
  free(buf);
  return 0;
}
```

output is typically,
```
59.981 ms | process_1
91.146 ms | process_2
```
using gcc (Rev3, Built by MSYS2 project) 12.1.0
cflags  = -std=c11 -Wall -O2 -Wextra

Indicating current indexing method is faster?
However,  it doesn't mean much as it is dependent on so many variables

e.g. I changed the optimisation flag from O2 to O3 and the result is typically,

```
104.040 ms | process_1
42.692 ms | process_2
```
Indicating pointer increment method can be faster :)
_______________

What about MSVC?
Microsoft doesn't have gcc's gettimeofday.
Using code from here - [stackoverflow - Equivalent of gettimeday() for Windows](https://stackoverflow.com/a/59359900/243662)

typically,
```
111.197 ms | process_1
112.680 ms | process_2

```

No difference!
using,
cl.exe - Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30146 for x64
optimisation flag **/O2** (maximum speed)

_______________

The point I'm making is optimisation is a tricky business :)
_______________

An area worth exploring is only assigning those bytes that are different, between formats,

e.g. from inner loop for 4 byte color, we have,
```
      data[0] = color.b;
      data[1] = color.g;
      data[2] = color.r;
      data[3] = color.a;
      data += 4;
```

If the original data is RGBA and we are going to BGRA, we can do a datarow memory copy in the outer loop and then only swap R & B in the inner loop.

I will explore this.

rlatowicz: @Starbuck5 

I'm not sure if this is the best place to continue the discussion on optimisation or create a new thread in the forum?
___________________

**Auto Increment addressing mode** - one op does access and increment (or decrement) at once,
[ref](http://www.cs.iit.edu/~cs561/cs350/addressing/addressing.html)

But not on Intel CPUs,
[stackoverflow](https://stackoverflow.com/questions/49034893/post-increment-and-pre-decrement-addressing-modes-on-intel-80386)
___________________

I put together a quick benchmark to test attempts at speedup,
As an example, I tested the current indexing method vs pointer incrementing,

```
#include <sys/time.h>
#include <stdio.h>
#include <stdlib.h>

typedef unsigned char Uint8;

typedef struct RGBA {
  Uint8 r;
  Uint8 g;
  Uint8 b;
  Uint8 a;
} RGBA_t;

void init_buffer(RGBA_t *b, size_t width, size_t height) {
  size_t i = 0;
  for (size_t h = 0; h < height; ++h) {
    for (size_t w = 0; w < width; ++w) {
      b[i].r = i % 256;
      b[i].g = (i+1) % 256;
      b[i].b = (i+2) % 256;
      b[i].a = (i+3) % 256;
      i++;
    }
  }
}

void print_buffer(RGBA_t *b, size_t width, size_t height) {
  size_t i = 0;
  for (size_t h = 0; h < height; ++h) {
    for (size_t w = 0; w < width; ++w) {
      printf("(%d,%d,%d,%d) ", b[i].r, b[i].g, b[i].b, b[i].a);
      i++;
    }
  printf("\n");
  }
}

void process_1(RGBA_t *b, size_t width, size_t height) {
  RGBA_t *ptr = b;
  Uint8 *data = (Uint8 *)b;
  for (size_t h = 0; h < height; ++h) {
    for (size_t w = 0; w < width; ++w) {
      RGBA_t color = *ptr++;
      data[0] = color.b;
      data[1] = color.g;
      data[2] = color.r;
      data[3] = color.a;
      data += 4;
    }
  }
}

void process_2(RGBA_t *b, size_t width, size_t height) {
  RGBA_t *ptr = b;
  Uint8 *data = (Uint8 *)b;
  for (size_t h = 0; h < height; ++h) {
    for (size_t w = 0; w < width; ++w) {
      RGBA_t color = *ptr++;
      *data++ = color.a;
      *data++ = color.r;
      *data++ = color.g;
      *data++ = color.b;
    }
  }
}

int main() {
  size_t width = 10000;
  size_t height = 10000;
  
  long seconds, microseconds, elapsed;
  struct timeval begin, end;
  
  RGBA_t *buf = calloc(width*height, sizeof(RGBA_t));
  
  // initialize RGBA buffer
  init_buffer(buf, width, height);
  // print_buffer(buf, width, height);
  
  // times process_1
  gettimeofday(&begin, 0);
  process_1(buf, width, height);
  gettimeofday(&end, 0);
  
  seconds = end.tv_sec - begin.tv_sec;
  microseconds = end.tv_usec - begin.tv_usec;
  elapsed = seconds*1e6 + microseconds;
  printf("%.3f ms | process_1\n", elapsed/1000.0);
  // print_buffer(buf, width, height);
  
  // time process_2
  gettimeofday(&begin, 0);
  process_2(buf, width, height);
  gettimeofday(&end, 0);
  
  seconds = end.tv_sec - begin.tv_sec;
  microseconds = end.tv_usec - begin.tv_usec;
  elapsed = seconds*1e6 + microseconds;
  printf("%.3f ms | process_2\n", elapsed/1000.0);
  // print_buffer(buf, width, height);
  
  
  free(buf);
  return 0;
}
```

output is typically,
```
59.981 ms | process_1
91.146 ms | process_2
```
using gcc (Rev3, Built by MSYS2 project) 12.1.0
cflags  = -std=c11 -Wall -O2 -Wextra

Indicating current indexing method is faster?
However,  it doesn't mean much as it is dependent on so many variables

e.g. I changed the optimisation flag from O2 to O3 and the result is typically,

```
104.040 ms | process_1
42.692 ms | process_2
```
Indicating pointer increment method can be faster :)
_______________

What about MSVC?
Microsoft doesn't have gcc's gettimeofday.
Using code from here - [stackoverflow - Equivalent of gettimeday() for Windows](https://stackoverflow.com/a/59359900/243662)

typically,
```
111.197 ms | process_1
112.680 ms | process_2

```

No difference!
using,
cl.exe - Microsoft (R) C/C++ Optimizing Compiler Version 19.29.30146 for x64
optimisation flag **/O2** (maximum speed)

_______________

The point I'm making is optimisation is a tricky business :)

Starbuck5: > The point I'm making is optimisation is a tricky business :)

That's the truth.

I was just wondering if there was an easy win there, and I guess not.

PR #3344, opened 2022-07-27
Remove constants repetition and duplicate define

Starbuck5:
I noticed that constants compiled into one of the largest of the pygame modules, so I was looking around constants. I figured that we could use the `__all__` list generated by constants.c instead of redefining the whole thing in locals.py. While validating that my changes didn't affect any output, I discovered that `GL_ACCELERATED_VISUAL` is added twice in constants.c by mistake (the same way each time), so I took out a duplicate definition.


ankith26 (approved):
Nice simplifications! Thanks 👍 🎉

zoldalma999 (approved):
Looks good to me! 🎉

src_py/locals.py
```diff
@@ -19,573 +19,12 @@
 #    pete@shinners.org
 
 
-"""Set of functions from PyGame that are handy to have in
+"""Set of functions from pygame that are handy to have in
```
illume, line 22: lowercase pygame makes me feel happy

illume (commented):
Indeed, nice deletion! :)

PR #3348, opened 2022-07-28
Fixed issue with Freetype not being clipped

### LINKED ISSUE #3346: Freetype.Font.render_to rendering outside surface clipping area
**Environment:**

- **Operating system** : Windows
- **Python version** : 3.9.5
- **SDL version** : 2.0.18
- **PyGame version**: 2.1.2

When using the Freetype.Font.render_to method on a surface with a clip set, the text can still be rendered outside the clipping area.

**Test code**

```
import pygame as pg

pg.init()
pg.freetype.init()

font = pg.freetype.Font(None)
window = pg.display.set_mode((1024, 576))
clip = pg.Rect(0, 0, 100, 20)

while True:
    
    if any({e.type == pg.QUIT for e in pg.event.get()}):
        break


    window.fill((0, 0, 0))
    window.set_clip(clip)
    
    pg.draw.rect(window, (255, 0, 0), clip)
    font.render_to(
        window,
        (0, 0),
        'this text should be clipped inside the red box',
        size=20,
        fgcolor=(255, 255, 255)
    )
    
    pg.display.flip()
    window.set_clip(None)
    
pg.quit()
```

PurityLake:
Fixes #3346 

Simply the clip set for a Surface object was not being copied into a FontSurface object, just its width, and height.

My fix works by setting its width or height to its clip equivalent clip attribute if it is smaller than its counterpart. I hope this is an ok fix, I don't think any tests failed from it.

### DIFF
--- a/test/freetype_test.py
+++ b/test/freetype_test.py
@@ -1794,6 +1794,75 @@ class FreeTypeTest(unittest.TestCase):
 
         self.assertIsNone(error_msg)
 
+class VisualTests(unittest.TestCase):
+
+    __tags__ = ["interactive"]
+
+    screen = None
+    aborted = False
+
+    def setUp(self):
+        if self.screen is None:
+            pygame.init()
+            self.screen = pygame.display.set_mode((600, 200))
+            self.screen.fill((255, 255, 255))
+            pygame.display.flip()
+            self.f = pygame.freetype.Font(None)
+
+    def tearDown(self):
+        pygame.quit()
+    
+    def query(self, clip_rect, is_clipped=False):
+        self.screen.fill((255, 255, 255))
+        question = ""
+        if is_clipped:
+            question = "Is the text clipped? y/n"
+        else:
+            question = "Can the full text be seen? y/n"
+
+        self.f.render_to(
+            self.screen,
+            (100, 100),
+            question,
+            size=20,
+            fgcolor=(0,0,0)
+        )
+
+        if is_clipped:
+            self.screen.set_clip(clip_rect)
+        else:
+            self.screen.set_clip(None)
+    
+        pygame.draw.rect(self.screen, (255, 0, 0), clip_rect)
+        self.f.render_to(
+            self.screen,
+            (0, 0),
+            'this text should be clipped inside the red box',
+            size=20,
+            fgcolor=(255, 255, 255)
+        )
+
+        pygame.display.flip()
+
+        while True:
+            for evt in pygame.event.get():
+                if evt.type == pygame.KEYDOWN:
+                    if evt.key == pygame.K_ESCAPE:
+                        self.abort()
+                        return False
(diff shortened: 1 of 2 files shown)


ankith26 (commented):
Hello, thanks for the contribution!

Can you add a unit test that checks for this case (so that this fix does not regress in the future)?

PurityLake:
I'll do that sometime later today.

PurityLake:
Added some tests for these changes

MyreMylar:
Not tested it myself yet, but these tests (and the changes) don't seem to take into account a clip rect not positioned at (0,0).

I'd like this tested with a clip rect of something like `pygame.Rect(10, 5, 100, 20)`

PurityLake:
I will do so immediately

test/freetype_test.py
```diff
@@ -1795,5 +1795,70 @@ def test_get_error(self):
...
+            self.screen = pygame.display.set_mode((600, 200))
+            self.screen.fill((255, 255, 255))
+            pygame.display.flip()
+            self.f = pygame.freetype.Font(None)
+
+    def tearDown(self):
+        pygame.quit()
+
+    def query(self, clip_rect, is_clipped=False):
+        self.screen.fill((255, 255, 255))
```
MyreMylar, line 1817: the screen background needs to be a different colour to the text that is supposed to be clipped
```suggestion
        self.screen.fill((128, 128, 128))
```

test/freetype_test.py
```diff
@@ -1795,5 +1795,70 @@ def test_get_error(self):
...
+                    if evt.key == pygame.K_y:
+                        return True
+                    if evt.key == pygame.K_n:
+                        return False
+                if evt.type == pygame.QUIT:
+                    self.abort()
+                    return False
+
+    def test_clipped(self):
+        self.assertTrue(self.query(pygame.Rect(0, 0, 100, 20), True))
```
MyreMylar, line 1857: you can test non (0,0) positioned rects here with a small change:

```suggestion
        self.assertTrue(self.query(pygame.Rect(10, 4, 100, 20), True))
```

MyreMylar (changes requested):
OK, I tested it and this doesn't seem to work at all. Once I adjusted the visual tests to be usable the text was clearly not clipped inside the red box.

image

test/freetype_test.py
PurityLake: Added some more tests now the issue is the start of the clip isn't respected and I am unsure how to fix this.

Screenshot from 2022-07-31 18-47-33

PurityLake: I believe it is an issue with the render functions

Starbuck5:
I don't believe this one is ready, I know you were talking on discord about changing this to just document the bug, rather than fixing it, which I would be fine with.

I'm going to go ahead and convert this to a draft in the meantime.

PR #3349, opened 2022-07-29
SDL_UCS4ToUTF8 may exist from SDL_keyboard.c

pmp-p:
test CI for removing function

### DIFF
--- a/src_c/key.c
+++ b/src_c/key.c
@@ -439,6 +439,7 @@ static const char *SDL1_scancode_names[SDL_NUM_SCANCODES] = {
     "AudioFastForward",
 };
 
+#if 0
 /* Taken from SDL_iconv() */
 char *
 SDL_UCS4ToUTF8(Uint32 ch, char *dst)
@@ -468,6 +469,11 @@ SDL_UCS4ToUTF8(Uint32 ch, char *dst)
     }
     return dst;
 }
+#else
+// from SDL_keyboard.c
+extern char *
+SDL_UCS4ToUTF8(Uint32 ch, char *dst);
+#endif
 
 /* Patch in pygame 1 compat names in our key name compat table */
 static void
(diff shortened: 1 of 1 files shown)


ankith26 (commented):
Another alternative is to just use the `pg_` prefix for this, and not rely on SDL at all on WASM (since this is an internal SDL util function, might not be wise to do)

pmp-p:
i think using the internal may gain some space ( .wasm are not compressed unlike .data) it's worth the risk and easy to fix later since function code will stay just above.

Starbuck5:
What about using `#ifndef` to define the function on our end only if it hasn't already been defined?

ankith26:
That would not work because `SDL_UCS4ToUTF8` is a function name not a macro

ankith26 (approved):
LGTM! Thanks 😎 

This has no affect on our non-static builds, but if this makes things better for WASM lets merge this in

MyreMylar (approved):
LGTM 👍

src_c/key.c
```diff
@@ -439,6 +439,11 @@ static const char *SDL1_scancode_names[SDL_NUM_SCANCODES] = {
...
 };
 
+#if defined(BUILD_STATIC)
+// from SDL_keyboard.c
+extern char *
+SDL_UCS4ToUTF8(Uint32 ch, char *dst);
+#else
 /* Taken from SDL_iconv() */
 char *
 SDL_UCS4ToUTF8(Uint32 ch, char *dst)
```
illume, line 449: ps. We should never define our own "SDL_" functions. This has bitten us in the past. Instead we should create pg_XXX functions which inside may call the SDL_XXX function, or include the function body. This way if someone links something with the SDL_XXX function in the future we have no problems.

illume (commented):
👍

PR #3352, opened 2022-07-30
Deprecate `Color.set_length`

### LINKED ISSUE #3341: Deprecate pygame.Color.set_length?
> Set the number of elements in the Color to 1,2,3, or 4.
set_length(len) -> None
The default Color length is 4. Colors can have lengths 1,2,3 or 4. This is useful if you want to unpack to r,g,b and not r,g,b,a. If you want to get the length of a Color do len(acolor).

Documentation: https://www.pygame.org/docs/ref/color.html#pygame.Color.set_length

This does not seem very useful to me, and it complicates the code of pygame.Color internally.

The unpacking functionality can be done like
`r, g, b, _ = pygame.Color(200,100,42)`

Any opinions?

blankRiot96:
Resolves #3341 

Does:
  - Trigger `DeprecationWarning` on calling `pygame.Color.set_length`
  - Warns deprecation in documentation
  - Checks for `DeprecationWarning` in <a href="https://github.com/pygame/pygame/blob/main/test/color_test.py">color_test.py</a>

Screenshots:

image

### DIFF
--- a/docs/reST/ref/color.rst
+++ b/docs/reST/ref/color.rst
@@ -214,6 +214,15 @@
       is useful if you want to unpack to r,g,b and not r,g,b,a. If you want to
       get the length of a Color do ``len(acolor)``.
 
+      .. note:: 
+         DEPRECATED: `set_length` will be deprecated from version 2.1.3
+         You may unpack the values you need like so, 
+         ``r, g, b, _ = pygame.Color(100, 100, 100)``
+         If you only want r, g and b
+         Or 
+         ``r, g, *_ = pygame.Color(100, 100, 100)`` 
+         if you only want r and g
+
       .. versionadded:: 1.9.0
 
       .. ## Color.set_length ##
--- a/docs/reST/ref/color.rst
+++ b/docs/reST/ref/color.rst
@@ -214,15 +214,7 @@
       is useful if you want to unpack to r,g,b and not r,g,b,a. If you want to
       get the length of a Color do ``len(acolor)``.
 
-      .. note:: 
-         DEPRECATED: `set_length` will be deprecated from version 2.1.3
-         You may unpack the values you need like so, 
-         ``r, g, b, _ = pygame.Color(100, 100, 100)``
-         If you only want r, g and b
-         Or 
-         ``r, g, *_ = pygame.Color(100, 100, 100)`` 
-         if you only want r and g
-
+      .. deprecated:: 2.1.3
       .. versionadded:: 1.9.0
 
       .. ## Color.set_length ##
--- a/docs/reST/ref/color.rst
+++ b/docs/reST/ref/color.rst
@@ -210,6 +210,14 @@
       | :sl:`Set the number of elements in the Color to 1,2,3, or 4.`
       | :sg:`set_length(len) -> None`
 
+      DEPRECATED: `set_length` will be deprecated from version 2.1.3
+      You may unpack the values you need like so, 
+      ``r, g, b, _ = pygame.Color(100, 100, 100)``
+      If you only want r, g and b
+      Or 
+      ``r, g, *_ = pygame.Color(100, 100, 100)`` 
+      if you only want r and g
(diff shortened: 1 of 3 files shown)


docs/reST/ref/color.rst
```diff
@@ -214,6 +214,15 @@
...
+      .. note:: 
+         DEPRECATED: `set_length` will be deprecated from version 2.1.3
+         You may unpack the values you need like so, 
+         ``r, g, b, _ = pygame.Color(100, 100, 100)``
+         If you only want r, g and b
+         Or 
+         ``r, g, *_ = pygame.Color(100, 100, 100)`` 
+         if you only want r and g
+
       .. versionadded:: 1.9.0
```
novialriptide, line 226: `.. deprecated:: 2.1.3` tag is needed

docs/reST/ref/color.rst
```diff
@@ -214,6 +214,15 @@
       is useful if you want to unpack to r,g,b and not r,g,b,a. If you want to
       get the length of a Color do ``len(acolor)``.
 
+      .. note::
```
Starbuck5, line 217: You can check other deprecated functions to see how these are usually documented.

they use a tag like `.. deprecated:: 2.1.1` at the end. Should probably say DEPRECATED up top, rather than in a note.

novialriptide (changes requested):
Looks good, just one change needed from me.

EDIT: Deleted duplicate review.

src_c/color.c
```diff
@@ -1679,6 +1679,15 @@ _color_set_length(pgColorObject *color, PyObject *args)
...
     int clength;
 
+    if (PyErr_WarnEx(
+            PyExc_DeprecationWarning,
+            "You can selectively unpack the rgba values you want, without "
+            "having to set length.\nExample: r, g, b, _ = pygame.Color(200, "
+            "100, 42)\nSo, set_length() will be deprecated in pygame 2.1.3",
+            1) == -1) {
+        return NULL;
+    }
```
Starbuck5, line 1689: Maybe it would be best to not try to explain why in the warning.

I also dislike the wording of "will be deprecated in pygame 2.1.3" in several places. It's not something that's happening in the future, if someone is seeing it it's already happened.

docs/reST/ref/color.rst
```diff
@@ -214,6 +214,7 @@
       is useful if you want to unpack to r,g,b and not r,g,b,a. If you want to
```
Starbuck5, line 214: There should be a "DEPRECATED" bit in this documentation, shortly explaining why. 

For example:
Capture

Although unlike this one, Color.set_length is not completely self explanatory, so you should keep the current documentation content as well.

Starbuck5 (approved):
Looks good to me! Thanks.

ankith26 (approved):
The code LGTM! Thanks for contributing! 🎉 🥳 

I don't think anyone would be against this deprecation... but I'd like to give this a few more days before a merge incase anyone opposes the idea

MyreMylar:
This seems a perfectly fine deprecation to me, the newer API makes a lot more sense than setting the length of a colour.

MyreMylar (approved):

src_c/color.c
```diff
@@ -1679,6 +1679,12 @@ _color_set_length(pgColorObject *color, PyObject *args)
 {
     int clength;
 
+    if (PyErr_WarnEx(PyExc_DeprecationWarning,
+                     "pygame.Color.set_length deprecated since 2.1.3",
```
zoldalma999, line 1683: I feel like this is missing an "is" (before "deprecated")

novialriptide (approved):

src_c/color.c
MyreMylar: I think you don't need the 'is' here.

illume (commented):
Cool. I couldn't find a single user of this doing a quick code search.

PR #3354, opened 2022-07-30
Add pg_IntFromObjEx(), pg_TwoIntsFromObjEx() and pg_IntFromSeqIndexEx()

itzpr3d4t0r:
# Info
The current C api functions we use are showing their age, and many of them need updating. Modifying existing api functions could led to undesired or wrong behaviour pretty quickly, being widespread functions, even the strangest of tests could lead to a fail.
What I'm trying to do here is different as i'm trying to add three brand new functions, that leverage a faster implementation and can easily be substitued to the current functions while also streamlining our C code. In this way we won't have to bother with the whole pygame project utilizing those old functions but rather implement the new functions only where needed while slowly substituting the old ones. 

This PR adds three new C API functions:
## pg_IntFromObjEx( )
Faster alternative to ``pg_IntFromObj`` as it uses faster numeric conversion functions(in the case of doubles), uses simpler checking and does not clear errors.

   Converts number like object *obj* to C int and places it in argument *val*.

   If an error is set and *msg* is not ``NULL``, uses *msg* as error message.
   Passing ``NULL`` as *msg* will preserve the original error messages. This can be helpful, but needs to be determined for each use case.

   - Returns ``1`` on success
   - Sets ``TypeError`` and returns ``0`` if the value is not numeric

## pg_IntFromSeqIndexEx( ) PRIVATE API ONLY
Faster alternative to ``pg_IntFromObjIndex`` as it assumes obj is a sequence, uses ``pg_IntFromObjEx`` for faster item->int conversion and does not clear errors.

   Convert number like object at position *index* in sequence *obj* to C int and place in argument *val*.

   If an error is set and *msg* is not ``NULL``, uses *msg* as error message.
   Passing ``NULL`` as *msg* will preserve the original error messages. This can be helpful, but needs to be determined for each use case.

   - Returns ``1`` on success
   - Sets ``TypeError`` and returns ``0`` if the value is not numeric
   
## pg_TwoIntsFromObjEx( )
Faster alternative to ``pg_TwoIntsFromObj`` as it uses ``PySequence_Fast`` for faster item extraction, ``pg_IntFromObjEx`` for faster item->int conversion and does not clear errors.

Converts two number like objects in sequence *obj* (must be of length 2) to C int and places them in arguments *val1* and *val2* respectively.

   If an error is set and *msg* is not ``NULL``, uses *msg* as error message. Passing ``NULL`` as *msg* will preserve the original error messages. This can be helpful, but needs to be determined for each use case.

   - Returns ``1`` on success
   - Sets ``TypeError`` and returns ``0`` if the values inside the sequence are not
     numeric or could not convert *obj* to a sequence
   - Sets ``ValueError`` and returns ``0`` if the sequence is not of size 2


src_c/base.c
```diff
@@ -533,6 +533,54 @@ pg_TwoIntsFromObj(PyObject *obj, int *val1, int *val2)
...
+    if (length == 2) {
+        if (!pg_F_IntFromObj(PySequence_Fast_GET_ITEM(fseqence, 0), val1) ||
+            !pg_F_IntFromObj(PySequence_Fast_GET_ITEM(fseqence, 1), val2)) {
+            Py_DECREF(fseqence);
+            return 0;
+        }
+    }
+    else if (length == 1) {
+        /* Handle case of ((x, y), ) 'sequence inside sequence' */
+        return pg_F_TwoIntsFromObj(PySequence_Fast_GET_ITEM(fseqence, 0), val1,
```
ankith26, line 570: There's a memleak on this codepath, with `fseqence` not being decref'd. Since `fseqence` has to be freed in every condition, I suggest that it be done at the end once, and other paths should `goto` here

ankith26 (commented):
Just a preference... I'd prefer if these new methods are named `pg_IntFromObjEx` and `pg_TwoIntsFromObjEx` respectively. This is the common convention in the Python C API world to use the `Ex` suffix when there is an extension on the older function, and this happens to be used a lot when the new API is error preserving

ankith26 (commented):
Digging deeper through the usage of the older `pg_IntFromObj` and `pg_TwoIntsFromObj` I realise that these functions originally suppressed all errors so that neater custom errors could be set... So while we are at making a newer extension, I'd suggest that these new functions take a `char *msg` argument, and if this function encounters a `TypeError`, it should use this message. All other errors should be forwarded as-is

src_c/base.c
itzpr3d4t0r: Thanks i fixed it and changed the names.

src_c/base.c
```diff
@@ -533,6 +533,62 @@ pg_TwoIntsFromObj(PyObject *obj, int *val1, int *val2)
...
+pg_IntFromObjEx(PyObject *obj, int *val, const char *msg)
+{
+    if (PyFloat_Check(obj)) {
+        *val = (int)PyFloat_AS_DOUBLE(obj);
+        return 1;
+    }
+
+    *val = PyLong_AsLong(obj);
+    if (PyErr_Occurred()) {
+        if (msg) {
```
ankith26, line 546:
```suggestion
        if (msg && PyErr_ExceptionMatches(PyExc_TypeError)) {
```
We should not overwrite other errors with `TypeError`. We should only reword the message in `TypeError`s if the caller of this function needs to (that is when msg is non-`NULL`

src_c/base.c
```diff
@@ -533,6 +533,62 @@ pg_TwoIntsFromObj(PyObject *obj, int *val1, int *val2)
...
+    return 1;
+}
+
+static int
+pg_TwoIntsFromObjEx(PyObject *obj, int *val1, int *val2, const char *msg)
+{
+    PyObject *fseqence = NULL;
+    Py_ssize_t length = 0;
+
+    if (!(fseqence = PySequence_Fast(obj, "A sequence was expected")))
```
ankith26, line 561:
```suggestion
    if (!(fseqence = PySequence_Fast(obj, msg)))
```

src_c/base.c
```diff
@@ -533,6 +533,62 @@ pg_TwoIntsFromObj(PyObject *obj, int *val1, int *val2)
...
+    else if (length == 1) {
+        /* Handle case of ((x, y), ) 'sequence inside sequence' */
+        PyObject *tmp = PySequence_Fast_GET_ITEM(fseqence, 0);
+        Py_DECREF(fseqence);
+        return pg_TwoIntsFromObjEx(tmp, val1, val2, msg);
+    }
+    else {
+        Py_DECREF(fseqence);
+        PyErr_SetString(PyExc_ValueError,
+                        "Invalid sequence size, must be either 1 or 2");
```
ankith26, line 584:
```suggestion
        PyErr_SetString(PyExc_ValueError, msg);
```
This should be a `ValueError` aight, but I think we should still use the more descriptive caller-passed message string here

src_c/base.c
itzpr3d4t0r: this does no longer exist

src_c/base.c
```diff
@@ -533,6 +533,79 @@ pg_TwoIntsFromObj(PyObject *obj, int *val1, int *val2)
...
+        }
+    }
+    /*This mainly accounts for vectors, as using PySequence_Fast on a vector is
+     * error-prone*/
+    else if (PySequence_Check(obj)) {
+        if (!pg_IntFromSeqIndexEx(obj, 0, val1, msg))
+            return 0;
+        if (!pg_IntFromSeqIndexEx(obj, 1, val2, msg))
+            return 0;
+    }
```
ankith26, line 599: I see some inconsistencies between the tuple/list codepath and the sequence codepath, and ideally behaviour should be same regardless of the type of the sequence.

- The first codepath handles the nested sequence case (which we keep for back-compat) while the second does not. I think both codepaths should be handling this case
- The second codepath does not give clean errors if the size of the sequence is unexpected (but the first does)

itzpr3d4t0r: Thanks! i think i resolved these.

src_c/base.c
```diff
@@ -533,6 +533,103 @@ pg_TwoIntsFromObj(PyObject *obj, int *val1, int *val2)
...
+    return 1;
+}
+
+static int
+pg_IntFromSeqIndexEx(PyObject *obj, int index, int *val, const char *msg)
+
+{
+    int result = 0;
+    PyObject *item = PySequence_ITEM(obj, index);
```
Starbuck5, line 559: Won't this segfault if you pass it something that isn't a sequence? Since it's a macro.

src_c/base.c
```diff
@@ -533,6 +533,103 @@ pg_TwoIntsFromObj(PyObject *obj, int *val1, int *val2)
...
+    return result;
+}
+
+static int
+pg_TwoIntsFromObjEx(PyObject *obj, int *val1, int *val2, const char *msg)
+{
+    Py_ssize_t length;
+    /*Faster path for tuples and lists*/
+    if (PyTuple_Check(obj) || PyList_Check(obj)) {
+        length = PySequence_Fast_GET_SIZE(obj);
```
Starbuck5, line 574: You're taking advantage of the fact that Tuples and Lists are returned verbatim from PySequence_Fast, but that is merely an implementation detail. It would be best to write our code to work up to spec, to support alternative versions or future versions of Python.

src_c/base.c
itzpr3d4t0r: In theory yes, but as i've stated in the function docs(And i've been very clear), this function assumes the obj to be a sequence and it's even in the function name! no it's only natural for it to behave wrong if the developer uses this function the wrong way. They have been warned.

Starbuck5: Good work on the docs for these by the way, they look nice.

It seems like you've made this fast path function for the purposes of `pg_TwoIntsFromObjEx`, since it cuts out a redundant sequence check.

I'd rather not have a "public" C api function that can segfault on such an easy mistake to make, what do you think about an internal version of `pg_IntFromSeqIndexEx` just for `pg_TwoIntsFromObjEx`, as well as a public version that includes a sequence check?

src_c/base.c
itzpr3d4t0r: Actually, i'm taking advantage of the fact that lists/tuples are compatible with sequencefast and i can therefore directly call fast functions.
From what i see this code is already up to spec, as it covers every possibility. With that being said we can't support something that doesn't exist yet, and so you'll always have to open a new PR to modify API to be up to spec. After all this is a start, if something other that tuples or lists will recieve the sequencefast implementation one will add nother check on top and you're done.
Also this implementation after a month+ of experimentation i've found it's the cleanest and most importantly the most  stable and performant one you can get. Goint into the PySequenceFast() rabbit hole conversion is a pain as it's not clear how it handles errors and you can get into trouble with strings as they are convertible sequences. With this you don't have to check for strings as a plus.

src_c/base.c
itzpr3d4t0r: Thaks! yeah i think that could also work

itzpr3d4t0r: But again, many times one checks for `PySequence_Check()` anyway, so having a function like this check each time you wanna extract an element can be redundant.
For example it can happen you wanna extract 3 or 4 elements from a sequence. In that case if you add an extra check in `pg_IntFromSeqIndexEx` you'll end up adding 3-4 checks to your extractions, all redundant.

src_c/base.c
```diff
@@ -533,6 +533,124 @@ pg_TwoIntsFromObj(PyObject *obj, int *val1, int *val2)
...
+        PyObject **f_arr = PySequence_Fast_ITEMS(obj);
```
MyreMylar, line 596: According to this draft PEP:

https://peps.python.org/pep-0620/

This function is up for deprecation soon so we may want to either find a different solution now, or bear it in mind for the future.

Is this of any use:

https://peps.python.org/pep-0590/

From skimming some in depth discussion the general vibe I get is that the Vectorcall API is going to be preferred for this sort of thing by the Python devs into the future, but it is only added in 3.8 (with underscores of uncertainty) and 3.9 (for reals). There was some discussion about scrapping FASTCALL for a while - but now it is in the Limited API so who knows.

Issue #3353, opened 2022-07-30
Alias for classes such as `pygame.Joystick` and `pygame.Sound`


novialriptide:
Classes like `Joystick`, `Event`, and `Sound` cannot be called like `Rect` and `Surface`. Currently, we have to do something like `pygame.joystick.Joystick` instead of `pygame.Joystick`, while `pygame.rect.Rect` can be called as `pygame.Rect`

It'll be nice if they have an alias consistent with `pygame.Rect` and `pygame.Surface`. The following classes need correction:
```
Joystick
Font
Event
Sound
Channel
Clock
```

PurityLake:
I'll do this a little later on today

PR #3356, opened 2022-07-30
Add aliases for most common objects

PurityLake:
Refers to #3353

Added aliases to `pygame` module:
* Joystick
* Font
* Event
* Channel
* Clock


Starbuck5 (commented):
It can't be implemented *quite* this simply, because pygame is built to be modular, and to be okay with certain modules not being compiled with it.

Rect is required, but font is not.

These should go in the try excepts for their respective modules.

Starbuck5 (commented):
Some of these will take a while to get used to... But it'll be a cool change!

I've just realized that these should probably get tiny unit tests, to prevent this new from behavior from unintentional regressions.

It could be as simple as asserting pygame.joystick.Joystick == pygame.Joystick, but for each one.

Oh shoot, how could I forget! This also will need documentation and type hinting support.

PurityLake:
I'll hop to that then

Starbuck5:
This still needs tests and type hints.

PurityLake:
I believe that should be it for this PR unless I have overlooked something.

buildconfig/stubs/pygame/__init__.pyi
```diff
@@ -52,6 +52,10 @@ from .math import Vector2 as Vector2, Vector3 as Vector3
 from .cursors import Cursor as Cursor
 from .bufferproxy import BufferProxy as BufferProxy
 from .mask import Mask as Mask
+from .event import Event as Event
+from .font import Font as Font
+from .mixer import Channel as Channel
+from .time import Clock as Clock
```
MyreMylar, line 58: `Joystick` is missing here.

MyreMylar (changes requested):
Seems fine, except `Joystick` is missing from the typestubs.

buildconfig/stubs/pygame/__init__.pyi
PurityLake: That should be fixed now

MyreMylar (approved):
LGTM 👍

Starbuck5 (commented):
The `versionchanged` labels should point to 2.1.4 now

PurityLake:
@Starbuck5 I have made that change to the docs

buildconfig/stubs/pygame/__init__.pyi
```diff
@@ -51,6 +51,11 @@ from .math import Vector2 as Vector2, Vector3 as Vector3
 from .cursors import Cursor as Cursor
 from .bufferproxy import BufferProxy as BufferProxy
 from .mask import Mask as Mask
+from .event import Event as Event
```
zoldalma999, line 54: As mentioned in the comment at the top of the file, this is an autogenerated file. Please edit `buildconfig/stubs/gen_stubs.py` instead

src_py/__init__.py
```diff
@@ -252,6 +258,7 @@ def Overlay(format, size):  # pylint: disable=unused-argument
 
 try:
     import pygame.mixer
+    from pygame.mixer import Channel
```
zoldalma999, line 261: `Sound` is missing from here. It needs a test, stubs, and mention in the docs as well (unless it is intentional?).

src_py/__init__.py
```diff
@@ -129,6 +130,7 @@ def warn(self):
 
 try:
     import pygame.event
+    from pygame.event import Event
```
zoldalma999, line 133: If the event module is not available, the Event class won't exist. Other modules where the import fails, they define a dummy function (see Cursor for example).

PR #3359, opened 2022-07-31
edited docs for pygame.Vector2.angle_to

### LINKED ISSUE #2729: Is there a problem / misunderstanding with `Vector.angle_to()`
According to the docs [angle_to()](https://www.pygame.org/docs/ref/math.html#pygame.math.Vector2.angle_to) returns the angle between self and the given vector. But the underlying C code ([link](https://github.com/pygame/pygame/blob/6eb04ef9fcb6dd37aaaa303c2c16daac70ddec8b/src_c/math.c#L2168)) suggests that the angle _between_ two vectors is actually calculated:

```c
angle = atan2(other_coords[1], other_coords[0]) - atan2(self->coords[1], self->coords[0]);
```

Because the function's name suggests calculating the angle _to_ not _between_ two vectors, it may be worth renaming the function, e.g. into a `angle_between()` function, and changing the code of `angle_to()`, e.g. akin to the following calculation:

```c
angle = atan2(other_coords[1] - self->coords[1], other_coords[0] - self->coords[0]);
```

I took the liberty of setting up a pull request here: #2728; but am unsure if the indexing is correct (i.e. if `atan2()` takes (x, y) or (y, x) as indicated by the array access), and if the intention / naming of `angle_to()` was chosen to mean something different in the first place. That said, maybe I am just not used to the convention; in which case, just ignore this issue and the pull request.

oddbookworm:
PR to fix #2729 

I edited the docs for `pygame.Vector2.angle_to` so that it better describes what the underlying code actually does.


novialriptide (approved):
LGTM
image

Here's what it should look like when you build the docs on my end.

MyreMylar (approved):
LGTM 👍

Starbuck5 (approved):
Looks good, thanks!

docs/reST/ref/math.rst
```diff
@@ -332,7 +332,14 @@ Multiple coordinates can be set using slices or swizzling
       | :sl:`calculates the angle to a given vector in degrees.`
       | :sg:`angle_to(Vector2) -> float`
 
-      Returns the angle between self and the given vector.
+      Returns the angle from self to the passed ``Vector2`` that would rotate self
+      to be aligned with the passed ``Vector2`` without crossing over the negative 
+      x-axis.
+
+      .. figure:: code_examples/angle_to.png
+         :alt: angle_to image
```
illume, line 340: This alt text could have been a bit better:
> Shows the angle_to going anti-clockwise from the first "self" vector to the second "other" vector.

tips for good alt text: https://supercooldesign.co.uk/blog/how-to-write-good-alt-text

illume (commented):
👍 much easier to understand

docs/reST/ref/math.rst
oddbookworm: The problem with that characterization is that it is false. It very well could return a clockwise angle as long as that wouldn’t cross the negative x-axis

PR #3361, opened 2022-07-31
Make pygame.math.clamp only take numbers

JamesC01:
This removes the support for different types, but makes the function clearer for beginners, and gives more comprehensible errors. This also will make implementing other math functions nicer, since we won't have to worry about inconsistencies with some math functions taking any object that defines certain methods, and others taking only numbers.

The reason I'm not simply using PyFloat_AsDouble is because converting to C types is slower than dealing with python objects directly.


docs/reST/ref/math.rst
```diff
@@ -51,11 +51,10 @@ Multiple coordinates can be set using slices or swizzling
 .. function:: clamp
 
    | :sl:`returns value clamped to min and max.`
-   | :sg:`clamp(value, min, max) -> clamped_value`
+   | :sg:`clamp(value, min, max) -> float`
 
-   Takes ``value`` of any type defining ``__lt__`` and ``__gt__``, and clamps it
-   so it is no lower than ``min``, and no higher than ``max``. If it isn't lower
-   than ``min`` or higher than ``max``, ``value`` is returned un-clamped.
+   Clamps ``value`` so that it's no lower than ``min``, and no higher
```
ankith26, line 56: Should mention here that ``value`` is numeric (even though it is fairly obvious, does not hurt to have more docs 😉 )

test/math_test.py
```diff
@@ -49,42 +49,11 @@ def test_clamp(self):
...
-        self.assertTrue(result == Point(40, 3))
-        # Test equal to min
-        result = pygame.math.clamp(Point(2, 2), Point(2, 2), Point(40, 3))
-        self.assertTrue(result == Point(2, 2))
-        # Test between min and max
-        result = pygame.math.clamp(Point(10, 10), Point(2, 2), Point(40, 3))
-        self.assertTrue(result == Point(10, 10))
+        # Error tests
+        # Not enough args
+        self.assertRaises(ValueError, pygame.math.clamp, 10)
```
ankith26, line 54: A `TypeError` must be raised when the argument count does not match. Can you please update this test and also the code to raise `TypeError`?

ankith26 (approved):
I'm fine with this, thanks for contributing 🥳 🎉

src_c/math.c
```diff
@@ -4088,12 +4088,17 @@ static PyObject *
...
     PyObject *min = args[1];
     PyObject *max = args[2];
 
+    if (PyNumber_Check(args[0]) != 1 || PyNumber_Check(args[1]) != 1 ||
+        PyNumber_Check(args[2]) != 1) {
+        return RAISE(PyExc_TypeError, "clamp requires 3 numeric arguments");
+    }
+
     // if value < min: return min
     int result = PyObject_RichCompareBool(value, min, Py_LT);
```
Starbuck5, line 4103: I know this isn't the focus of this PR, but it would be cool to put in a comment that the reason the function is implemented without full conversion into C types is for performance reasons, and this was tested to be faster.

Starbuck5 (approved):
Looks good to me, thanks! I left a small review for your consideration.

illume:
This breaks the existing math.clamp API right? From the removed tests this seems the case to me. There's not a discussion I can see why this was done. 

Can someone please explain why clamp on objects is not useful anymore? Let me know, otherwise I guess this will be reverted.

ankith26:
`pygame.math.clamp` was recently added (in #3326), this change definitely does not break any user code (using a stable version) thankfully.

As to why this change is being made, the contributor of the original function intended to only support `float`, the generic object clamping only came up as an optimisation. The push for this PR was also made by the author.

I think the most common usecase for clamping is real numbers, I've never had the need to clamp anything else (except vectors, but they have their own method for it. And pygame vectors don't support comparisons either).... you can't really clamp many types of objects. Clamping a `Point` between 2 points in 2D space (based on the distance from origin) is not very clear to me and seems kind of arbitrary.

But I'm not an expert by any means, and I think there may be cases (although probably rare) where non-numbers need clamping. I can imagine that clamping would make sense with python sequences (because they have a well defined comparison method)

From my side, I'm fine either way, but if this PR does end up being reverted, one part of it needs to be re-added (the part where `ValueError` is replaced with `TypeError` for mismatching argument sizes)

JamesC01:
> This breaks the existing math.clamp API right? From the removed tests this seems the case to me. There's not a discussion I can see why this was done.
> 
> Can someone please explain why clamp on objects is not useful anymore? Let me know, otherwise I guess this will be reverted.

Ankith said it well. I only recently made the clamp function, and allowed it to work with any type only because of performance reasons (converting to c types was quite a bit slower). There aren't many cases I can think of where clamp is useful, outside of regular numbers. On the surface it seemed like maybe it could be used for things like colours, to make sure they don't go outside of 0-255, but that doesn't work.

And with lerp, the performance was worse than python when letting it support any object, so I stuck with converting to c types (like most of the other math vector functions) so to keep clamp consistent with lerp, and make it a little easier to understand for beginners, I took away the ability to use any object. I think it's more consistent, since it's a math function, and there aren't many use cases for it outside of numbers. As for using it on sequences, numpy already has that functionality, and it's a popular well known library, so I'm not sure if there's a reason to implement that.

illume:
Thanks for the explanations.

PR #3362, opened 2022-07-31
Implement pygame.math.lerp

JamesC01:
This implements pygame.math.lerp. The speed is around 50% faster than if you wrote it in python. I still need some feedback on the documentation, and also whether or not it should raise an error if percentage is outside the range of [0, 1]. Most engines allow for values outside that range, and the ones that don't, just clamp the value to that range. What do you all think?


buildconfig/stubs/pygame/math.pyi
```diff
@@ -18,8 +18,10 @@ if sys.version_info >= (3, 9):
 else:
     from typing import Collection
 
+# Math functions
 _T = TypeVar('_T')
 def clamp(value: _T, min: _T, max: _T, /) -> _T: ...
+def lerp(a: float, b: float, percent: float) -> float: ...
```
ankith26, line 24: Just like the function above, this takes positional-only args so it must have a trailing `/`

src_c/math.c
```diff
@@ -4120,6 +4120,33 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
 
 PG_WRAP_FASTCALL_FUNC(math_clamp, PyObject);
 
+static PyObject *
+math_lerp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
+{
+    if (nargs != 3)
+        return RAISE(PyExc_ValueError, "lerp requires 3 arguments");
```
ankith26, line 4127: Should be a `PyExc_TypeError`

src_c/math.c
```diff
@@ -4120,6 +4120,33 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+        return NULL;
+    double percent = PyFloat_AsDouble(args[2]);
+    if (PyErr_Occurred())
+        return NULL;
+
+    // Not sure about this. Most other frameworks/engines either automatically
+    // clamp it, or allow you to use values outside this range.
+    // if (percent < 0.0 || percent > 1.0)
+    //    return RAISE(PyExc_ValueError, "Percentage value must be between 0.0
+    //    and 1.0");
```
ankith26, line 4143: Personally I'm in favour of raising this error, but before you uncomment this, lets wait to see what others have to say

MyreMylar: Here is some discussion on the topic: https://answers.unity.com/questions/953384/why-is-the-t-value-clamped-to-0-1-in-lerp-function.html

Which sort of makes the point that maybe we should use the clamp function here to clamp the values for percent instead of just erroring if we are going down the "values should only be between 0.0 and 1.0" route. I've personally not intentionally used an unclamped lerp.

Unity seems to provide both options, with the standard lerp being clamped. See:

https://docs.unity3d.com/ScriptReference/Mathf.LerpUnclamped.html

There doesn't seem to be a firm consensus on whether lerp should be clamped or unclamped, I personally don't have a strong preference - I just haven't ever tried to use lerp for extrapolation. Though in the cases I've seen that seems to be mainly used for geometry i.e. lerps on Vectors and maybe colours. What do our current lerps do? Consistency is also a good goal.

Whatever options is chosen (only one function is likely simpler, two is more feature complete - could add another in a separate PR) documentation over what the function does (is it clamped or not) would help.

JamesC01: The Vector lerp raises a ValueError if it's outside the range [0, 1]. I guess for now I'll just be consistent and raise it, but I'm still not sure if that's the best idea. It doesn't really line up with most other frameworks.

JamesC01:
@MyreMylar @ankith26 Do any of you have any suggestions for what I should write in the docs? So far I'm thinking of this:

"Linearly interpolates between ``a`` and ``b`` by ``percent``. If ``percent`` is 0, argument ``a`` will be returned, if ``percent`` is 1, argument ``b`` will be returned. Raises a ValueError if ``percent`` is outside the range of ``[0, 1]``.

I want to explain what happens if percent is between 0 and 1, but I'm not sure of the best way to word it.

oddbookworm:
@JamesC01 one use-case I can think of is if you have an object at position 1 and you want to smoothly move from position 1 to position 2 without risking over/under shooting position 2. Or maybe if you implement a slider in a game, `pygame.math.lerp` could be used to more easily describe what it's actually set to. As for documenting values between 0 and 1, you could probably say something akin to "if `percent` is between 0 and 1, then the position that is `percent * 100`% of the way from a to b on a straight line path."

docs/reST/ref/math.rst
```diff
@@ -61,6 +61,38 @@ Multiple coordinates can be set using slices or swizzling
...
+
+   Linearly interpolates between ``a`` and ``b`` by ``percent`` using the formula ``a + (b-a) * percent``.
+   
+   If ``percent`` is ``0.5``, ``lerp`` will return the value half-way between ``a``
+   and ``b``. When ``a = 10`` and ``b = 20``, ``lerp(a, b, 0.5)`` will return ``15``.
+
+   ``lerp`` can be used for many things. You could rotate a sprite by a percentage with
+   ``angle = lerp(0, 360, percent)``. You could even scale an enemy's attack value
+   based on the level you're playing:
+
```
ankith26, line 77: I like the first part of this doc, I think mentioning the formula it uses clears up a lot of things...

The latter part of this docs is a bit specific and subjective, docs generally don't need to mention too many usage details (although in some cases they make sense and explain stuff better)

I also think that the doc keyword name "percent" could be renamed, because "percent" as a name gives the impression that this has to be between 0 and 100

JamesC01: @MyreMylar said this on my issue: "Though more attention might need to be put on things like documentation and examples to introduce the concept of lerping in games to an audience who might never have heard of it. 'Linear interpolation' sounds like a scary thing that is actually simple and very useful (health bars spring to mind)." 

What do you think a good name would be for percent? Godot using lerp(from, to, weight).

ankith26: I like "weight" as a name

Apart from that naming, the rest of the docs looks good to go to me

docs/reST/ref/math.rst
```diff
@@ -61,6 +61,38 @@ Multiple coordinates can be set using slices or swizzling
...
+      current_level = 2
+
+      attack = lerp(10, 50, current_level/MAX_LEVEL) # 18
+
+   If you're on level 0, ``attack`` will be ``10``, if you're on level 10,
+   ``attack`` will be ``50``. If you're on level 5, the
+   percentage calculated with ``current_level/MAX_LEVEL`` will be ``0.5``
+   which is 50%, therefore ``attack`` will be ``30``, which is half-way between ``10`` and ``50``.
+
+   Raises a ValueError if ``percent`` is outside the range of ``[0, 1]``.
```
Starbuck5, line 90: So I believe this is still under discussion. I don't have a strong opinion. The other lerp functions in pygame (color and vectors) both raise ValueErrors if the input is outside the range as well, so the consistency is good.

One option that I haven't seen yet is an optional error raise, like `zip`. This could allow people to unlock a "crazy range" lerp if they want, while keeping the error by default, which may help people catch mistaken inputs to the function.

ankith26:
We have a feature for 2.1.3, so it's in the 2.1.4 milestone for now

JamesC01:
> We have a feature for 2.1.3, so it's in the 2.1.4 milestone for now

What do you mean?

PR #3364, opened 2022-08-01
Use memcpy and memset when appropriate

Starbuck5:
I was looking around and found a couple places where we should be using memcpy or memset. I was pursuing performance with this project, but there aren't good speedups in this, as far as I can tell, unfortunately.

I did get one performance "win" - for transform.flip(surf, False, False). This is a "flip" that doesn't flip x or y, so it's just a copy. Previously this copy was pixel line by pixel line, but I've changed it to copy the whole buffer in one memcpy call. This is now 20% faster.

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -277,9 +277,7 @@ rotate90(SDL_Surface *src, int angle)
                 dstpix = dstrow;
                 srcpix = srcrow;
                 for (loopx = 0; loopx < dstwidth; ++loopx) {
-                    dstpix[0] = srcpix[0];
-                    dstpix[1] = srcpix[1];
-                    dstpix[2] = srcpix[2];
+                    memcpy(dstpix, srcpix, 3);
                     srcpix += srcstepx;
                     dstpix += dststepx;
                 }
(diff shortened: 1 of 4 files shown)


ankith26 (approved):
Thanks 🥳 😎

MyreMylar (approved):
LGTM 👍 

Straightforward change.

src_c/transform.c
```diff
@@ -812,9 +808,8 @@ surf_flip(PyObject *self, PyObject *args, PyObject *kwargs)
 
     if (!xaxis) {
         if (!yaxis) {
-            for (loopy = 0; loopy < surf->h; ++loopy)
-                memcpy(dstpix + loopy * dstpitch, srcpix + loopy * srcpitch,
-                       surf->w * surf->format->BytesPerPixel);
+            assert(srcpitch == dstpitch);
```
illume, line 811: This isn't always true though. srcpitch can != dstpitch

Starbuck5: I'm not sure how they would be different. The two surfaces have the same format, and presumably SDL would allocate pitches the same way on two same sized surfaces with the same format each time.

But in the interest of safety, and to not prod with things that don't need prodding: #3416

PR #3366, opened 2022-08-01
Fix some METH_NOARGS usage, use METH_O where possible

ankith26:
This PR is primarily intended for better code quality and better error reporting, but this also helps a (really tiny) bit with performance.

Some functions that were supposed to be `METH_NOARGS` were `METH_VARAGS` (which means that you could pass any arbitrary arguments to these functions and python would not complain)

I also tried to apply `METH_O` in some places where they make the code clearer

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -146,7 +146,7 @@ surf_get_palette(PyObject *self, PyObject *args);
 static PyObject *
 surf_get_palette_at(PyObject *self, PyObject *args);
 static PyObject *
-surf_set_palette(PyObject *self, PyObject *args);
+surf_set_palette(PyObject *self, PyObject *list);
 static PyObject *
 surf_set_palette_at(PyObject *self, PyObject *args);
 static PyObject *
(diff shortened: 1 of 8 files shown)


src_c/surface.c
```diff
@@ -1092,7 +1092,7 @@ surf_get_palette_at(PyObject *self, PyObject *args)
 }
 
 static PyObject *
-surf_set_palette(PyObject *self, PyObject *args)
+surf_set_palette(PyObject *self, PyObject *list)
```
Starbuck5, line 1095: Maybe rename to sequence, just because that what it actually needs to be?

src_c/time.c
```diff
@@ -263,13 +257,7 @@ static PyObject *
...
     int ticks, start;
-    PyObject *arg0;
-
-    /*for some reason PyArg_ParseTuple is puking on -1's! BLARG!*/
-    if (PyTuple_Size(arg) != 1)
-        return RAISE(PyExc_ValueError, "delay requires one integer argument");
-    arg0 = PyTuple_GET_ITEM(arg, 0);
-    if (!PyLong_Check(arg0))
+    if (!PyLong_Check(arg))
         return RAISE(PyExc_TypeError, "delay requires one integer argument");
```
Starbuck5, line 261: Not your fault, but it looks like someone made a copy paste error here. Shouldn't it be "wait requires one integer argument" ?

Starbuck5 (approved):
This is quite an impressive roundup! I read through everything and ran the tests locally, it looks good to me.

Left a couple minor things you might want to change before merging.

src_c/surface.c
ankith26: Yup! I only chose `list` because the old code used that name. But I'll update this too in this PR

src_c/time.c
ankith26: good catch

MyreMylar (approved):
LGTM 👍 

Tests all pass locally on my windows machine.

PR #3371, opened 2022-08-03
wasm: multiphase init for cython modules instead of builtins

pmp-p:
one more step toward wasm dynamic linking, also allow to unload/reload modules properly
small changes to `pygame/__init__.py` to allow wasm loading without additionnal helper : now helper is in C and integrated to the static build instead of being buried into pygbag loader.

note that pygame.wasm_patches is missing i'm still working on fixing audio.


MyreMylar (approved):
LGTM 👍

pmp-p:
that PR is ok enough to get dynamic loading working on cpython-wasm and pyodide

https://pmp-p.ddns.net/paste/screen/shot-2022-10-14_1665723412.png
( on cpython-wasm but need https://github.com/emscripten-core/emscripten/pull/17956 applied )


pyodide/pyscript ( when linked to SDL2 : there are forks that do that like the one from pyxel ) should work via micropip.install when on firefox (  v8 imposes streaming compilation for a lib that big ).

pmp-p:
@MyreMylar it should build when adding `#pgrama once` or other fencing here https://github.com/pygame/pygame/blob/main/src_c/_blit_info.h

MyreMylar:
> @MyreMylar it should build when adding `#pragma once` or other fencing here https://github.com/pygame/pygame/blob/main/src_c/_blit_info.h

Do you want to add this:

```
#ifndef _BLIT_INFO_H
#define _BLIT_INFO_H
#define NO_PYGAME_C_API
#include "_surface.h"

/* The structure passed to the low level blit functions */
typedef struct {
    int width;
    int height;
    Uint8 *s_pixels;
    int s_pxskip;
    int s_skip;
    Uint8 *d_pixels;
    int d_pxskip;
    int d_skip;
    SDL_PixelFormat *src;
    SDL_PixelFormat *dst;
    Uint8 src_blanket_alpha;
    int src_has_colorkey;
    Uint32 src_colorkey;
    SDL_BlendMode src_blend;
    SDL_BlendMode dst_blend;
} SDL_BlitInfo;
#endif //BLIT_INFO_H

```

 to `_blit_info.h` as part of this PR then?

pmp-p:
not sure about #pragma once so using your way  thanks !

MyreMylar:
Oops, needed some extra spaces on the comment there to make clang-format happy. I updated the code block above.

pmp-p:
301

PR #3373, opened 2022-08-04
Docs/vector epsilon

### LINKED ISSUE #3368: Document `pygame.Vector2.epsilon` and `pygame.Vector3.epsilon`
In discord, @Starbuck5 pointed out to me that there is no documentation for `pygame.Vector2.epsilon` and `pygame.Vector3.epsilon`, and it was discussed that it would probably be a good idea to document this feature. My thinking is that it might be useful to someone if they have a unique use-case where the default tolerance of `1e-6` might be too small or too large. I think I'd be up for documenting this feature

oddbookworm:
I documented `Vector2.epsilon` and `Vector3.epsilon`. Merging this PR would close #3368

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -445,6 +445,40 @@ Multiple coordinates can be set using slices or swizzling
 
       .. ## Vector2.update ##
 
+   
+   .. attribute:: epsilon
+      
+      | :sl:`Determines the tolerance of vector calculations.`
+      
+      Both Vector classes have a value named ``epsilon`` that defaults to ``1e-6``.
+      This value acts as a numerical margin in various methods to account for floating point
+      arithmetic errors. Specifically, ``epsilon`` is used in the following places:
+
+         * comparing Vectors (``==`` and ``!=``)
+         * the ``is_normalized`` method (if the square of the length is within ``epsilon`` of 1, it's normalized)
+         * slerping (a Vector with a length of ``<epsilon`` is considered a zero vector, and can't slerp with that)
+         * reflection (can't reflect over the zero vector)
+         * projection (can't project onto the zero vector)
+         * rotation (only used when rotating by a multiple of 90 degrees)
+
+      While it's possible to change ``epsilon`` for a specific instance of a Vector, all the other Vectors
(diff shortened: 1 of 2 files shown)


ankith26 (approved):
This is fairly comprehensive and it clears up a lot of things about the secretive and mystic epsilon 😄 
Very nice docs addition, thanks! 🎉 🥳

docs/reST/ref/math.rst
```diff
@@ -1017,6 +1053,10 @@ Multiple coordinates can be set using slices or swizzling
 
       .. ## Vector3.update ##
 
+   .. attribute:: epsilon
+
+      | :sl:`With lengths within this number, vectors are considered equal. For more information see` Vector2_:
```
Starbuck5, line 1058: This does generate a link to Vector.epsilon, but how?

It seems like it would be preferable to have an explicit reference to Vector2.epsilon.

docs/reST/ref/math.rst
```diff
@@ -445,6 +445,42 @@ Multiple coordinates can be set using slices or swizzling
...
+      Both Vector classes have a value named ``epsilon`` that defaults to ``1e-6``.
+      This value acts as a numerical margin in various methods to account for floating point
+      arithmetic errors. Specifically, ``epsilon`` is used in the following places:
+
+      * comparing Vectors (``==`` and ``!=``)
+      * the ``is_normalized`` method (if the square of the length is within ``epsilon`` of 1, it's normalized)
+      * slerping (a Vector with a length of ``<epsilon`` is considered a zero vector, and can't slerp with that)
+      * reflection (can't reflect over the zero vector)
+      * projection (can't project onto the zero vector)
+      * rotation (only used when rotating by a multiple of 90 degrees)
```
Starbuck5, line 464: When I generate this locally the bulleted list doesn't appear indented at all. Maybe giving it some more spaces on the left would fix this?

Starbuck5 (approved):
Looks good to me, thanks! 🎉 

Only thing to resolve before merge is commit history, since it's a bit messy. You could squash some of them together, or we could do a squash-and-merge.

Issue #2540, opened 2021-03-29
Enhancement: Add display server/windowing system to support prompt


robertpfeiffer:
If pygame is running under X11, wayland, or XWayland, if we have any way to detect this, maybe we should consider printing this in the support prompt. There are many display bugs that are impossible to diagnose without this info.

We should probably *also* print the processor architecture if it's anything other than x86_64.

robertpfeiffer:
This is impossible to google. I have no idea if there is any way to query the X Server for information about concrete implementation details. It might be possible by bypassing SDL, but X or X server is just too short. I get information about Apple XServe hardware.

MyreMylar:
Is this what we need:
https://wiki.libsdl.org/SDL_GetCurrentVideoDriver

PR #3374, opened 2022-08-05
debug method

oddbookworm:
I saw #2540 and figured I'd make an attempt to implement it. I tested this on Manjaro and it worked properly for me, but I don't have a pygame dev environment set up in more than one place atm, so I can't test it in other places yet.


oddbookworm:
In accordance with a conversation I had with @ankith26 in discord, I modified the PR to instead display the result of `platform.platform()` instead of the driver reported by sdl2

Starbuck5:
Found a problem.

```py
>>> import os
>>> os.environ["PYGAME_HIDE_SUPPORT_PROMPT"] = "1"
>>> import pygame
  File "C:\Users\charl\AppData\Local\Programs\Python\Python39\lib\site-packages\pygame\__init__.py", line 409, in <module>
    platform,
NameError: name 'platform' is not defined
>>>
```

oddbookworm:
That's an easy fix

oddbookworm:
@Starbuck5 should be fixed now

src_py/__init__.py
```diff
@@ -392,12 +392,23 @@ def __color_reduce(c):
 
 # Thanks for supporting pygame. Without support now, there won't be pygame later.
 if "PYGAME_HIDE_SUPPORT_PROMPT" not in os.environ:
+    import platform
+
     print(
-        "pygame {} (SDL {}.{}.{}, Python {}.{}.{})".format(  # pylint: disable=consider-using-f-string
-            ver, *get_sdl_version() + sys.version_info[0:3]
+        "pygame {} (SDL {}.{}.{}, Python {}.{}.{}, Architecture {})".format(  # pylint: disable=consider-using-f-string
```
ankith26, line 398: This is not the architecture... it's a string with system OS info

ankith26 (commented):
Thanks for the PR! 👍 

I support this change since it helps both issue reporters and issue triagers... but there is a minor downside that it makes the support prompt more verbose

oddbookworm:
This PR is not dead, but is awaiting a pretty significant change according to a conversation in discord

MyreMylar:
OK, I will hold off on my review until the significant change has appeared (cryptic :D)

oddbookworm:
Waiting for #3379 to resolve. The future of this PR is something to the effect of:
A pygame.debug() method that gets everything relevant to the current setup. If display is initialized, it grabs the driver. If mixer is initialized, it grabs the SDL mixer info, etc.. and either prints to console or writes to a log file called pygame_debug.txt in the cwd (or the user can provide a filepath to save to instead?)

MyreMylar:
@oddbookworm looks like this could be whipped into shape now that #3379 is merged.

MyreMylar (changes requested):
Changes requested status: 

Update this to use #3379

oddbookworm:
I'll work on this in the next few days or so

oddbookworm:
Didn't mean to close the PR

oddbookworm:
New summary:
A `pygame.debug` method that can optionally write to a file, but by default prints to the console. Retrieves all the important information about a particular user's setup

oddbookworm:
If #3567 gets merged before this PR, then I don't need the special freetype logic and it'll behave like all of the others

MyreMylar:
Example test program:

```
import pygame

pygame.init()
pygame.debug()
```

Example output:

```
Platform:		Windows-10-10.0.22000-SP0
System:			Windows
Processor:		AMD64 Family 23 Model 113 Stepping 0, AuthenticAMD
Architecture:		Bits: 64bit	Linkage: WindowsPE

pygame version:		2.1.4.dev1
python version:		3.9.0

SDL versions:		Linked: 2.24.0	Compiled: 2.24.0
SDL Mixer versions:	Linked: 2.6.2	Compiled: 2.6.2
SDL Font versions:	Linked: 2.20.1	Compiled: 2.20.1
SDL Image versions:	Linked: 2.0.5	Compiled: 2.0.5
Freetype versions:	Linked: 2.11.1	Compiled: 2.11.1
```

I like the output, should provide a useful block to copy and paste into issue reports. 

Should the function be called `debug_info()`, `print_debug_info()` or `print_system_info()` though instead? We have a few 'info' gathering functions (`display.Info()`, `display.get_wm_info()`, midi.get_device_info()) already that this would then fit among - though none of those also print straight to the console. 

I'm concerned that `debug()` is too general a name for what this function does which is essentially gather and print a bunch of hardware and software type & version information.

Getting names right is hard.

MyreMylar (approved):

ankith26 (commented):
Probably also worth including `platform.python_implementation()` and probably `platform.version()` too in here

ankith26:
> If https://github.com/pygame/pygame/pull/3567 gets merged before this PR, then I don't need the special freetype logic and it'll behave like all of the others

Cool, I think we should wait for that then before merging this one

Starbuck5:
This code needs to be resilient against crashing in different possible pygame environments, including when different modules are unavailable or fail to import.

So I think this needs to handle the case of the sub modules being MissingModule, as well as printing the reason a module is missing (usually a trace back)

docs/reST/ref/pygame.rst
```diff
@@ -192,6 +192,25 @@ object instead of the module, which can be used to test for availability.
 
    .. ## pygame.encode_file_path ##
 
+.. function:: debug
```
MyreMylar, line 195: does this documentation need updating to match the renamed function?

src_c/doc/pygame_doc.h
```diff
@@ -11,6 +11,7 @@
 #define DOC_PYGAMEREGISTERQUIT "register_quit(callable) -> None\nregister a function to be called when pygame quits"
 #define DOC_PYGAMEENCODESTRING "encode_string([obj [, encoding [, errors [, etype]]]]) -> bytes or None\nEncode a Unicode or bytes object"
 #define DOC_PYGAMEENCODEFILEPATH "encode_file_path([obj [, etype]]) -> bytes or None\nEncode a Unicode or bytes object as a file system path"
+#define DOC_PYGAMEDEBUG "debug(filename=None) -> None\nretrieves useful information for debugging and issue-reporting purposes"
```
MyreMylar, line 14: again the name of the function looks to be the old name here.

docs/reST/ref/pygame.rst
oddbookworm: Just pushed a commit to fix this

src_c/doc/pygame_doc.h
oddbookworm: Just pushed a commit to fix this

MyreMylar (approved):
Current output looks like this on my machine:

```
Platform:		Windows-10-10.0.22000-SP0
System:			Windows
System Version:		10.0.22000
Processor:		AMD64 Family 23 Model 113 Stepping 0, AuthenticAMD
Architecture:		Bits: 64bit	Linkage: WindowsPE

Python:			CPython
pygame version:		2.1.4.dev1
python version:		3.11.1

SDL versions:		Linked: 2.24.0	Compiled: 2.24.0
SDL Mixer versions:	Linked: 2.6.2	Compiled: 2.6.2
SDL Font versions:	Linked: 2.20.1	Compiled: 2.20.1
SDL Image versions:	Linked: 2.0.5	Compiled: 2.0.5
Freetype versions:	Linked: 2.11.1	Compiled: 2.11.1
```

oddbookworm:
> EDIT: I am on windows 11, confusingly the version number for windows 11 is still 10.0

I think this is a problem with windows 11 itself. The vast majority of programs report the OS as windows 10, so it has to be something in the OS

MyreMylar:
Yes, sorry my comment was a bit confusing. The debug printout is correct,
it is just Microsoft being confusing with their versioning.

MyreMylar (approved):

PR #3370, opened 2022-08-02
Tracking SSE2 Optimisations

PurityLake:
Refers to #3358 

- [x] Copy Starbuck's optimization into simd_blitters_sse2.c
- [x] Find a way to use SSE2 Blitters explicitly
- [x] Benchmark SS2E code

This PR will be to track progress in optimising SSE2 Blitters based on Starbuck's AVX optimisations.

Currently have implemented the optimisations for SSE2.

PR #3375, opened 2022-08-05
Merge 32/64bit SSE blitter codepaths

Starbuck5:
I sat down yesterday with the intention of updating the older SSE blitters to use the "mixed width" strategy, speeding up odd blits by allowing them to take advantage of the parallelization, but the presence of separate 32 and 64 bit code paths made my task much harder to comprehend.

I did some research, finding https://gcc.gnu.org/bugzilla/show_bug.cgi?id=32869, looking through https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/emmintrin.h (hint: search for `#ifdef __x86_64__`), and of course looking through the friendly neighborhood intel intrinsics guide.

I found that the difference between the 64 and 32 bit code was the presence of cvtsi128_si64 and ctvsi64_si128 in the 64 bit code, and mitigations to their absence in the 32 bit code. I also found that there were two other intrinsics that are 100% equivalent and available on 32 bit.

Why not just use the 32bit-available intrinsics everywhere? According to the intrinsics guide, they have a higher latency.

So this PR just moves all uses of the relevant intrinsics into a macro, defines a valid 32 bit path for the macro, and eliminates all the now redundant 32 bit code paths.

This will probably conflict with #3370, but not too badly.

### DIFF
--- a/src_c/simd_blitters_sse2.c
+++ b/src_c/simd_blitters_sse2.c
@@ -21,6 +21,22 @@
 
 #if (defined(__SSE2__) || defined(PG_ENABLE_ARM_NEON))
 
+/* See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=32869
+ * These are both the "movq" instruction, but apparently we need to use the
+ * load low one on 32 bit. See:
+ * https://github.com/gcc-mirror/gcc/blob/master/gcc/config/i386/emmintrin.h
+ * According to the intel intrinsics guide the instructions this uses on 32
+ * bit are slightly slower
+ * ARGS: reg is a pointer to an m128, num is a pointer to a 64 bit integer */
+#if defined(ENV64BIT)
+#define LOAD_64_INTO_M128(num, reg) *reg = _mm_cvtsi64_si128(*num)
+#define STORE_M128_INTO_64(reg, num) *num = _mm_cvtsi128_si64(*reg)
+#else
+#define LOAD_64_INTO_M128(num, reg) \
+    *reg = _mm_loadl_epi64((const __m128i *)num)
+#define STORE_M128_INTO_64(reg, num) _mm_storel_epi64((__m128i *)num, *reg)
+#endif
+
 void
 alphablit_alpha_sse2_argb_surf_alpha(SDL_BlitInfo *info)
 {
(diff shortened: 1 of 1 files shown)


src_c/simd_blitters_sse2.c
```diff
@@ -788,14 +612,15 @@ alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
...
 
-#if defined(ENV64BIT)
 void
 blit_blend_rgba_mul_sse2(SDL_BlitInfo *info)
 {
     int n;
     int width = info->width;
     int height = info->height;
 
+    printf("inside rgba mul sse\n");
```
MyreMylar, line 622: I guess this still needs to get removed.

Starbuck5: Yes, thank you. 😄

MyreMylar (approved):
Gave this a second look this evening and gave it a solid test. 

I can confirm there is no performance change on 64 bit on my machine (as would be expected as it should be the same code). The tests all run locally too and on a careful eyeballin' it all looks good as well.

I think this is good to go, and should make the next phase  - porting the three normal alpha blitter paths to use the 'stride switch' method used by the all the other blitters - much easier.

👍 🍰 🎉

src_c/simd_blitters_sse2.c
```diff
@@ -424,10 +328,7 @@ alphablit_alpha_sse2_argb_no_surf_alpha(SDL_BlitInfo *info)
                     temp = _mm_and_si128(temp, rgb_mask_128);
                     /* add alpha to RGB */
                     temp = _mm_add_epi16(mm_dst_alpha, temp);
-                    *dstp64 =
-                        (((Uint64)_mm_cvtsi128_si32(_mm_srli_si128(temp, 4)))
-                             << 32 |
-                         (Uint32)_mm_cvtsi128_si32(temp));
+                    STORE_M128_INTO_64(&temp, dstp64);
```
ankith26, line 331: I'm not clear about this diff... I'm in no way experienced around this bits of code, but the code before applying the macro looks complex and this macro clearly does not expand to that

ankith26: In your opening comment you mentioned
> I also found that there were two other intrinsics that are 100% equivalent and available on 32 bit.

I assume that refers to this (please correct me if it's not the case)

MyreMylar: they actually are equivalent, it is just messy. Original is  64 bit SIMD -> to 32 bit non-SIMD via two 32 bit stores OR'd together, with one bitshifted 32 bits one way to fit into 32 bits on the way out of the SIMD registers, then shifted back the other way to fit into a 64 bit int. They were basically in 32 bit int format just for the store instruction. 

For whatever reason, while you can have 64 bit ints on a 32 bit system, but not all the 64 bit SIMD instructions work on a 32 bit architecture. Thus this awkward dance I wrote originally because I thought there were no 64 bit store operations on 32 bit systems. But apparently I missed something as there actually _are_ some 64 bit load and store operations that work on 32 bit (hooray) - they just are a bit slower than the ones I was using for the 64 bit path. 

Thus this new macro that lets us just use the 64 bit code for everything rather than a terrible hacked up version of it on 32 bit.

ankith26: Oh, this makes sense now, thanks for explaining!

ankith26 (approved):
LGTM! This reduces a lot of code duplication and makes complex things a tad bit easier, thankies!

Starbuck5:
Force push was just to get rid of that debug printf removal commit. I'll merge this in a bit.

illume (commented):
👍 wow, very nice cleanup.

PR #3376, opened 2022-08-05
Vendor in SDL PNG save function, fix new 2.1.3 regression

### LINKED ISSUE #3369: Surfaces with palettes are saved incorrectly in 2.1.3.dev
```py
import pygame
pygame.init()

a = pygame.Surface((50,50), depth=8)
a.fill("green")
pygame.draw.rect(a, "red", [20,10,10,40])

pygame.image.save(a, "8bit.png")
pygame.image.save(a, "8bit.jpg")
```

In pygame 2.1.2, these two images have the same content, not so in 2.1.3.dev.

Jpg:
8bit

PNG:
8bit

The natural suspect for this regression is https://github.com/pygame/pygame/pull/3028.

ankith26:
PR to fix #3369

Since this is a regression from 2.1.2, it needs to be in before 2.1.3

I tried a fair bit to not vendor in the SDL function and instead patch the existing code, but no matter what I tried I could not fix the bug. I wrote a unit test to catch the bug which is also a part of this PR (for this to work, I had to make a small bugfix patch in the png util library used for testing pngs)

In the middle term, the goal is to use the SDL function directly without vendoring it (there is #3242 which is a draft ATM). So this PR should help with easing the transition to that

### DIFF
--- a/src_c/imageext.c
+++ b/src_c/imageext.c
@@ -145,7 +145,17 @@ image_load_ext(PyObject *self, PyObject *arg)
     return final;
 }
 
+/* This entire png saving code is directly copied from the SDL_image source
+ * (with minor changes)
+ * Eventually this should be removed, and we should start using the SDL_image
+ * functions directly */
 #ifdef PNG_H
+
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+static const Uint32 png_format = SDL_PIXELFORMAT_ABGR8888;
+#else
+static const Uint32 png_format = SDL_PIXELFORMAT_RGBA8888;
+#endif
 
 static void
 png_write_fn(png_structp png_ptr, png_bytep data, png_size_t length)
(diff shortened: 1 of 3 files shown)


Starbuck5 (approved):
It's a big diff, but vendoring SDL_image here is actually not a bad plan. If we upgrade to SDL_image 2.6 on some platforms and not on others (temporarily), this will help consistency between versions.

Anyways, this fixes my bug, and the code looks good to me. Thanks for adding a test, too! Looks good to me.

I also think it's funny that a pygame person (me) wrote a patch for SDL_image's PNG saving, and now we're copying that whole code into pygame as a patch for our own bug.

MyreMylar (approved):
LGTM 👍 

I can confirm that the tests pass locally and the bug is fixed from the issue test program on my windows PC.

illume (commented):
Cool. 8bit images are important that we get right.

I wonder if we can do grayscale saving somehow.

PR #3378, opened 2022-08-06
SSE alpha blitter optimization

Starbuck5:
Specifically in `alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst`

I found an opportunity to reduce the amount of CPU clocks / instructions needed to move the alpha component from the src pixels into an interleaved 16 bit format.

In my testing, this got 10k 512x512 alpha blits (of this format) from around 2 seconds to around 1.6 seconds, a 15-20% improvement.

```py
import pygame
import time

pygame.init()

screen = pygame.Surface((1920,1080), depth=32)
screen.fill((255,0,23))

surf = pygame.Surface((512,512), pygame.SRCALPHA)
surf.fill((22,156,77,192))

print(screen, surf)

start = time.time()
for _ in range(10000):
    screen.blit(surf, (51, 67))
print(time.time() - start)
```
This is my test program. Note that the "screen" surface has a bit depth of 32 but does not have per pixel alpha.

### DIFF
--- a/src_c/simd_blitters_sse2.c
+++ b/src_c/simd_blitters_sse2.c
@@ -449,25 +449,16 @@ alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
     int srcskip = info->s_skip >> 2;
     int dstskip = info->d_skip >> 2;
 
-    // SDL_PixelFormat *srcfmt = info->src;
-    // SDL_PixelFormat *dstfmt = info->dst;
-
     Uint64 *srcp64 = (Uint64 *)info->s_pixels;
     Uint64 *dstp64 = (Uint64 *)info->d_pixels;
 
-    // Uint64 src_amask64 = ((Uint64)srcfmt->Amask << 32) | srcfmt->Amask;
-
     Uint64 rgb_mask64 = 0x00FFFFFF00FFFFFF;
+    Uint32 rgb_mask32 = 0x00FFFFFF;
 
     Uint32 *srcp32 = (Uint32 *)info->s_pixels;
     Uint32 *dstp32 = (Uint32 *)info->d_pixels;
 
-    // Uint32 src_amask32 = srcfmt->Amask;
-
-    Uint32 rgb_mask32 = 0x00FFFFFF;
-
-    __m128i src1, dst1, sub_dst, mm_src_alpha, mm_zero;
-    __m128i mm_alpha_mask_1, mm_alpha_mask_2, mm_rgb_mask;
+    __m128i src1, dst1, sub_dst, mm_src_alpha, mm_zero, mm_rgb_mask;
 
     /* There are two paths through this blitter:
            1. Two pixels at once.
@@ -479,8 +470,6 @@ alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
         dstskip = dstskip / 2;
 
         mm_zero = _mm_setzero_si128();
-        mm_alpha_mask_1 = _mm_cvtsi32_si128(0x000000FF);
-        mm_alpha_mask_2 = _mm_cvtsi32_si128(0x00FF0000);
 
         /* two pixels at a time */
         LOAD_64_INTO_M128(&rgb_mask64, &mm_rgb_mask);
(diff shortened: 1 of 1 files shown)


Starbuck5:
This depends on #3375, which should be merged first.

src_c/simd_blitters_sse2.c
```diff
@@ -490,21 +479,25 @@ alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
...
-                    /* created squashed alpha -> mm_src_alpha
-                     * (0000000000000A0A) */
+                    /* isolate alpha channels
+                     * 00000000A1000A2000 -> mm_src_alpha */
+                    mm_src_alpha = _mm_andnot_si128(mm_rgb_mask, src1);
+
+                    /* shift right to position alpha channels for manipulation
+                     * 000000000A1000A200 -> mm_src_alpha*/
+                    mm_src_alpha = _mm_srli_si128(mm_src_alpha, 1);
+
```
MyreMylar, line 489: Might be helpful to have a comment indicating where the logic switched from 8 bit integers over to 16 bit
```suggestion
                    /* contents considered as 8 16bit integers :
                     * [00][00][00][00][0A1][00][0A2][00]
                     * [7 ][6 ][5 ][4 ][ 3 ][2 ][ 1 ][0 ]
```

Was something I made while following through the logic of the shuffle.

src_c/simd_blitters_sse2.c
```diff
@@ -490,21 +479,25 @@ alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
...
-                        _mm_add_epi16(_mm_and_si128(_mm_srli_si128(src1, 3),
-                                                    mm_alpha_mask_1),
-                                      _mm_and_si128(_mm_srli_si128(src1, 5),
-                                                    mm_alpha_mask_2));
+                        _mm_shufflelo_epi16(mm_src_alpha, 0b11110101);
 
-                    /* Then Calc RGB */
-                    /* 000000000A10A10A20A2 -> rgb_src_alpha */
+                    /* finally move into final config
+                     * spread out so they can be multipled in 16 bit math
```
MyreMylar, line 496: small spelling error here on multiplied.

src_c/simd_blitters_sse2.c
```diff
@@ -490,21 +479,25 @@ alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
...
+                    /* isolate alpha channels
+                     * 00000000A1000A2000 -> mm_src_alpha */
+                    mm_src_alpha = _mm_andnot_si128(mm_rgb_mask, src1);
+
+                    /* shift right to position alpha channels for manipulation
+                     * 000000000A1000A200 -> mm_src_alpha*/
+                    mm_src_alpha = _mm_srli_si128(mm_src_alpha, 1);
+
+                    /* shuffle alpha channels to duplicate 16 byte pairs
+                     * 000000000A10A10A20A2 -> mm_src_alpha */
```
MyreMylar, line 491:
```suggestion
                   /* shuffle alpha channels to duplicate 16 byte pairs
                     * shuffle operation on lower half = (3,3,1,1)
                     * 000000000A10A10A20A2 -> mm_src_alpha */
```

Another suggestion to go with the suggestion above, for those of us whose brains can't digest binary easily.

MyreMylar (approved):
LGTM 👍 

All seems to work in testing and I was able to follow the logic of the changes. I added a couple of minor suggestions but they don't change the actual code just comments around it that might be helpful for future intrinsic dabblers.

🍰 🎉

Starbuck5:
Thanks for the suggestions and reviews @MyreMylar.

There is now a much more in depth comment. I also found a mistaken "16 byte" reference that should have been "16 bit"

MyreMylar (approved):

src_c/simd_blitters_sse2.c
```diff
@@ -490,21 +479,30 @@ alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
...
+                     * [7 ][6 ][5 ][4 ][ 3 ][2 ][ 1 ][0 ]
+                     * Therefore the previous contents of 16 bit number #1
+                     * Goes into 16 bit number #1 and #2, and the previous
+                     * content of 16 bit number #3 goes into #2 and #3 */
                     mm_src_alpha =
-                        _mm_add_epi16(_mm_and_si128(_mm_srli_si128(src1, 3),
-                                                    mm_alpha_mask_1),
-                                      _mm_and_si128(_mm_srli_si128(src1, 5),
-                                                    mm_alpha_mask_2));
+                        _mm_shufflelo_epi16(mm_src_alpha, 0b11110101);
```
illume, line 498: These are very helpful comments. Thanks.

I wonder if either of you have good docs in your code editor for these intrinsics?

It's possible to look them up, but would be quicker if it's shown in the editor. https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm_shufflelo_epi16&ig_expand=6448

illume (approved):
🎉🎈thanks

src_c/simd_blitters_sse2.c
Starbuck5: Occasionally I see short text descriptions of the intrinsics in VS code (pulled from the header I believe), but mainly I just rely on the intel intrinsics guide for these things.

PR #3379, opened 2022-08-07
Tweaking version getters for sdl modules and freetype

### LINKED ISSUE #3087: Unify version getters

oddbookworm:
I've worked on syncing up the version getters `pygame.get_sdl_version`, `pygame.mixer.get_sdl_mixer_version`, `pygame.font.get_sdl_ttf_version`, `pygame.freetype.get_version`, and `pygame.image.get_sdl_image_version`. Now they all accept a keyword arg to change whether they return the linked version of the indicated module or the compiled version.

By default, `pygame.get_sdl_version`, `pygame.mixer.get_sdl_mixer_version`, and `pygame.font.get_sdl_ttf_version` all return the linked version and use the `linked` boolean keyword to toggle which version number is grabbed. `linked=False` will grab the compiled version.

By default, `pygame.freetype.get_version`, and `pygame.image.get_sdl_image_version` all return the compiled version and use the `compiled` boolean keyword to toggle which version number is grabbed. `compiled=False` will grab the linked version.

This is behavior that already existed for some of these functions, and for those that I had to introduce the keyword toggle, I kept the old behavior as the default. Ideally, all 5 of them would use the same keyword and default behavior for simplicity, but that would come at the cost of changing the default behavior for a few of the methods listed here.

Question: should they be left as-is to avoid the (slight) break of changing the default behavior for some methods or should they all be made consistent across the board for simplicity?

This is intended to close #3087


src_c/imageext.c
```diff
@@ -414,10 +414,27 @@ image_save_ext(PyObject *self, PyObject *arg)
...
+    int compiled = 1;
+
+    static char *keywords[] = {"compiled", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|p", keywords,
+                                     &compiled)) {
+        return NULL;
+    }
+
+    if (!compiled) {
```
novialriptide, line 429: I think it'll be neater to switch these around. Not a big issue though. Like this:
```
if (compiled) {}
else {}
```

oddbookworm: The reason I left it like this for now is that if it's decided to give everything the same default behavior and the same keyword, I only have to change some stuff in 2 or 3 of them, and since `linked` is kinda the opposite of `compiled`, leaving practically the same structure eliminates some intermediate confusion at the moment

ankith26 (commented):
For some initial consistency, I believe all functions should have the same keyword, regardless of whether that defaults to `True` or `False`. Now changing the default value needs more discussion

src_c/imageext.c
oddbookworm: I believe the commit I just pushed resolves this

Starbuck5:
I think the linked version is what should be reported.

If pygame was built with SDL 2.0.14 but it's running on a Linux system and sees a newer libsdl, say SDL 2.0.20, it will use that version instead. This could be significant for joystick support, for instance, where we've frequently seen SDL versions be the entire difference "100% working" and "0% working."

For the runtime behavior, the linked version is more important. This is why the existing functions that have this argument use linked=True.

This is a change in behavior, but it should have very minimal backwards compatibility implications. 
- These functions are not very commonly used (exception being in the pygame display prompt, but that behavior could be preserved)
- Most pygame installs use the SDL they come with anyways, so linked version == compiled version. (Every windows install, most Mac and Linux installs?)
- I can't see any games ending with an exception due to this.

MyreMylar:
I agree, return the linked version by default for everything to synch them up. It is a v. minor change that I doubt will affect anyone and is easily resolved if it does.

dr0id:
I think "returning the linked version by default for everything" is reasonable, increases consistency and won't hurt. Of course it should be documented accordingly and be mentioned in the change log I guess.

MyreMylar (approved):

MyreMylar:
Looks good to me 👍

illume:
Thanks-very-cool. Note pygame.font.get_sdl_ttf_version isn't in this PR but is mentioned in the description. There's going to be a change shortly for 2.1.3 to rename it from pygame.font.get_ttf_version, so maybe wait for that to be merged first.

Starbuck5:
> Thanks-very-cool. Note pygame.font.get_sdl_ttf_version isn't in this PR but is mentioned in the description. There's going to be a change shortly for 2.1.3 to rename it from pygame.font.get_ttf_version, so maybe wait for that to be merged first.

Sorry for the confusion because of the misnamed PR, get_sdl_ttf_version is currently implemented.

oddbookworm:
>Note pygame.font.get_sdl_ttf_version isn't in this PR but is mentioned in the description.

This is because it's already correct and there's nothing to do in this PR. I was just addressing that they are all of the same form now.

nthykier (approved):

ankith26:
This PR is missing unit tests though, maybe a follow up PR to add some basic unit tests would be nice?

PR #3381, opened 2022-08-07
fix CPython was built with clang instead of gcc

pmp-p:
as clang does not support unused-but-set-variable

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@
 
 import io
 import platform
+import sysconfig
 
 with open('README.rst', encoding='utf-8') as readme:
     LONG_DESCRIPTION = readme.read()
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -443,7 +444,8 @@ def run_install_headers(self):
 
     if "freetype" in e.name and sys.platform not in ("darwin", "win32"):
         # TODO: fix freetype issues here
-        e.extra_compile_args.append("-Wno-error=unused-but-set-variable")
+        if sysconfig.get_config_var("MAINCC") != "clang":
```
ankith26, line 447: I'm not sure how robust this check is...

I had added the `sys.platform not in ("darwin", "win32")` check above just to block this arg being set on clang and MSVC respectively (this flag should only be set on gcc) but this is not a robust check either

The best solution to this that I can find involves more distutils hacks (https://stackoverflow.com/a/1671060) but it should give the accurate compiler name that is being used to compile the code

illume (changes requested):
As Ankith says, the check should be for "is gcc". 

Sounds like this bug should block 2.1.3 release? (maybe the dev release too?)

illume (approved):
On second thoughts, I think this check improves things. It also doesn't break things, because msvc is mostly on windows.

It does need to be fixed better... but if there's not a better patch I'll merge this in for 2.1.3.dev6.

MyreMylar (approved):
This seems reasonable without a good gcc check.

PR #3382, opened 2022-08-09
fixed run_tests.py | Quiet bug catch.

Vazno:
If you somehow run this file without knowing which file exactly you are running, you will get no information, just "This module is for import only", So you will need to check every single file that was executed.
Fix:
Replaced it with raise ValueError() because it says what happened, and where exactly.

### DIFF
--- a/test/test_utils/run_tests.py
+++ b/test/test_utils/run_tests.py
@@ -1,7 +1,7 @@
 import sys
 
 if __name__ == "__main__":
-    raise ValueError("This module is for import only")
+    raise RuntimeError("This module is for import only")
 test_pkg_name = ".".join(__name__.split(".")[0:-2])
 is_pygame_pkg = test_pkg_name == "pygame.tests"
 test_runner_mod = test_pkg_name + ".test_utils.test_runner"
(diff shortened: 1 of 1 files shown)


test/test_utils/run_tests.py
```diff
@@ -1,8 +1,7 @@
 import sys
 
 if __name__ == "__main__":
-    sys.exit("This module is for import only")
-
+    raise ValueError("This module is for import only")
```
ankith26, line 4: Just a minor thing: This should probably be a `RuntimeError`. `ValueError`s are usually used to signal incorrect kinds of values that functions receive via their arguments

ankith26 (approved):
Thanks for the first time contribution! 🎉 🥳 
Left a minor review for you to address before this is merged, but the idea behind this PR sounds good to me

test/test_utils/run_tests.py
Vazno: Yeah, you are right, changed it with ``RuntimeError()``, thanks for review.

Starbuck5 (approved):
Sure!

PR #3383, opened 2022-08-09
Changing the syntax of some print statements

Oussama-jabrane:
I made some changes to the 'setup.py' file, by changing some syntax of the print statement to be compatible with all Python versions from the oldest to the newest. Because only python3 accepts  the syntax of "print(f'Hi {name}')", for example. So I made some improvements so that all the versions of Python can run the 'setup.py' file

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,10 +6,27 @@
 # To configure, compile, install, just run this script.
 #     python setup.py install
 
+import shutil
+import stat
+import glob
+import os.path
+import distutils.command.install_headers
+from distutils.command.sdist import sdist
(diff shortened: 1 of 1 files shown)


MyreMylar:
Python 3.6 is the lowest supported version of pygame. We aren't going backwards.

Oussama-jabrane:
> Python 3.6 is the lowest supported version of pygame. We aren't going backwards.

Okay, I am new here so I didn't know a lot about this. Thanks! :)

MyreMylar:
f strings are available in our lowest supported version of python, python
3.6.

PR #3384, opened 2022-08-10
Fix tests that had overshadowed names

ankith26:
I noticed that the math module had tests with the same name and one of those was being overshadowed. On further inspection I realised that these tests were almost exactly the same, so I removed these.

I also went through all other unit tests and found a few more cases of this overshadowing, and this time they were different tests but with the same name (which means that all these time, these tests were not actually being run). Fixed this by giving a new name to these

### DIFF
--- a/test/math_test.py
+++ b/test/math_test.py
@@ -734,242 +734,6 @@ class Vector2TypeTest(unittest.TestCase):
         self.assertEqual(type(self.v1.xyxyx), tuple)
 
     def test_elementwise(self):
-        # behaviour for "elementwise op scalar"
-        self.assertEqual(
-            self.v1.elementwise() + self.s1, (self.v1.x + self.s1, self.v1.y + self.s1)
-        )
-        self.assertEqual(
-            self.v1.elementwise() - self.s1, (self.v1.x - self.s1, self.v1.y - self.s1)
-        )
-        self.assertEqual(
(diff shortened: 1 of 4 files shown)


MyreMylar (approved):
LGTM 👍

test/math_test.py
```diff
@@ -1377,15 +1132,15 @@ def testConstructionKeywords(self):
         self.assertEqual(v.z, 3.0)
 
     def testConstructionMissing(self):
```
Starbuck5, line 1134: Is there a reason this test doesn't just call those Vector3 creation expressions inline with the asserts?

Starbuck5 (approved):
Looks good to me, thanks! Left a minor question.

test/math_test.py
ankith26: Well... we can use a lambda to make it inline, but either way we need to pass a function-like object to `assertRaises`

Starbuck5: A function like object like the Vector3 constructor? :eyes:

ankith26: Ah... `assertRaises` would forward args and kwargs to the constructor

illume (commented):
🎉 

Nice. I guess pylint would have found this?

ankith26:
Yup! pylint/mypy in general has a lot of complaints in the tests directory (but many of those are false positives because the unit tests also test for errors and such). This was a good catch by pylint, and this should definitely be on a pylint allowlist when/if we make the tests pylint-able in the future.

PR #3388, opened 2022-08-12
setup.py | io module no longer used.

Vazno:
setup.py no longer use this module in any way.
Previously it was here for with io.open() function, but then was deleted in this merge: https://github.com/pygame/pygame/commit/e815c442e8bd82a68d1c228e96b65a56eab00084

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,6 @@
 # To configure, compile, install, just run this script.
 #     python setup.py install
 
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
Thanks for contributing! 👍

MyreMylar (approved):
LGTM 👍

PR #3390, opened 2022-08-13
made some minor changes in sysfont.py

lebedevhhh:
Feeling like sysfont.py is very old

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -13,7 +13,7 @@
 #
 # You should have received a copy of the GNU Library General Public
 # License along with this library; if not, write to the Free
-# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston and MA  02111-1307 USA
 #
 # Pete Shinners
 # pete@shinners.org
@@ -23,6 +23,7 @@ import os
 import sys
 import warnings
 from os.path import basename, dirname, exists, join, splitext
+from dotenv import load_dotenv
 
 from pygame.font import Font
 
@@ -54,10 +55,13 @@ def _addfont(name, bold, italic, font, fontdict):
 
 def initsysfonts_win32():
     """initialize fonts dictionary on Windows"""
-
-    fontdir = join(os.environ.get("WINDIR", "C:\\Windows"), "Fonts")
-    fonts = {}
-
+	#faster and quicker
+	try:
+		fontdir = join(os.getenv("WINDIR", "C:\\Windows"), "Fonts")
+		fonts = {}
+	except ImportError as e:
+		print(e)
+			
     # add fonts entered in the registry
     microsoft_font_dirs = [
         "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Fonts",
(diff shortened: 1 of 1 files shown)


src_py/sysfont.py
```diff
@@ -13,7 +13,7 @@
 #
 # You should have received a copy of the GNU Library General Public
 # License along with this library; if not, write to the Free
-# Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
+# Foundation, Inc., 59 Temple Place, Suite 330, Boston and MA  02111-1307 USA
```
MyreMylar, line 16: This change is incorrect. MA is a state in the USA (Massachusetts) and Boston is a city in that state.

src_py/sysfont.py
```diff
@@ -23,6 +23,7 @@
 import sys
 import warnings
 from os.path import basename, dirname, exists, join, splitext
+from dotenv import load_dotenv
```
MyreMylar, line 26: This import is unused.

src_py/sysfont.py
```diff
@@ -54,10 +55,13 @@ def _addfont(name, bold, italic, font, fontdict):
...
-    fontdir = join(os.environ.get("WINDIR", "C:\\Windows"), "Fonts")
-    fonts = {}
-
+	#faster and quicker
+	try:
+		fontdir = join(os.getenv("WINDIR", "C:\\Windows"), "Fonts")
+		fonts = {}
+	except ImportError as e:
+		print(e)
+
```
MyreMylar, line 64: This change is incorrect. Wrapping a call to `getenv()` in a try except block does not make this code faster or quicker as the comment implies.

`getenv()` just returns None if it the environment variable cannot be found, it doesn't raise an `ImportError`. I suggest reading the python documentation for the os module:

https://docs.python.org/3/library/os.html

src_py/sysfont.py
```diff
@@ -156,18 +160,19 @@ def _font_finder_darwin():
         "/Library/Fonts",
         "/Network/Library/Fonts",
         "/System/Library/Fonts",
-        "/System/Library/Fonts/Supplemental",
+        "/System/Library/Fonts/Supplemental"
```
MyreMylar, line 163: This change (and the others like it) is unnecessary and will probably fail our auto formatter `black`.

src_py/sysfont.py
```diff
@@ -156,18 +160,19 @@ def _font_finder_darwin():
...
-        locations.append("/Users/" + username + "/Library/Fonts")
+		#using formatting 
+        locations.append(f"/Users/{username}/Library/Fonts")
 
     strange_root = "/System/Library/Assets/com_apple_MobileAsset_Font3"
     if exists(strange_root):
         strange_locations = os.listdir(strange_root)
         for loc in strange_locations:
-            locations.append(strange_root + "/" + loc + "/AssetData")
+            locations.append(f"{strange_root}/{loc}/AssetData")
```
MyreMylar, line 175: These f string changes are fine.

src_py/sysfont.py
```diff
@@ -185,7 +190,7 @@ def _font_finder_darwin():
 
 
 def initsysfonts_darwin():
-    """Read the fonts on MacOS, and OS X."""
+    """Read the fonts on MacO  and OS X."""
```
MyreMylar, line 193: This is a good typo fix.

MyreMylar (changes requested):
Bit of a mixed bag here. 

This looks like your first PR so I tried to give you some useful feedback. If you could cut this back to just the two f string changes and the one typo fix it might pass review.

src_py/sysfont.py
MyreMylar, line 193: typo added here. You've removed the S and a comma.

src_py/sysfont.py
lebedevhhh: got it !

MyreMylar:
Closing this because another PR has been opened with the only valid changes in it.

PR #3391, opened 2022-08-13
changed the f strings (sysfont.py)

lebedevhhh:
As I was told, I limted my changes up to f strings (doc : sysfont.py) (sry bad english)

PR #3392, opened 2022-08-14
Replaced more expressions to f strings

Vazno:
I looked at this PR #3391 and checked our repository with this regex pattern ``".*[+].*"``
I found more strings that can be converted to f strings for better readability

### DIFF
--- a/src_py/cursors.py
+++ b/src_py/cursors.py
@@ -119,9 +119,9 @@ class Cursor:
     def __repr__(self):
         if self.type == "system":
             id_string = _cursor_id_table.get(self.data[0], "constant lookup error")
-            return f'<Cursor(type: system, constant: {id_string})>'
+            return f"<Cursor(type: system, constant: {id_string})>"
         if self.type == "bitmap":
-            size = f"size: {self.data[0]!s}" # !s for calling str(), or it will call format()
+            size = f"size: {self.data[0]!s}"  # !s for calling str(), or it will call format()
             hotspot = f"hotspot: {self.data[1]!s}"
             return f"<Cursor(type: bitmap, {size}, {hotspot})>"
         if self.type == "color":
--- a/src_py/cursors.py
+++ b/src_py/cursors.py
@@ -121,11 +121,11 @@ class Cursor:
             id_string = _cursor_id_table.get(self.data[0], "constant lookup error")
             return f"<Cursor(type: system, constant: {id_string})>"
         if self.type == "bitmap":
-            size = f"size: {self.data[0]!s}"  # !s for calling str(), or it will call format()
-            hotspot = f"hotspot: {self.data[1]!s}"
+            size = f"size: {self.data[0]}"
+            hotspot = f"hotspot: {self.data[1]}"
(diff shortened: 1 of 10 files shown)


MyreMylar (approved):
A slightly confusing diff because the config files have never been autoformatted before, but I don't think there is any harm in it. The f string changes look fine.

👍

Starbuck5:
@Vazno This is a nice little cleanup, but I don't want the huge diff from the auto formatting.

Use `setup.py format` instead of calling black on all of pygame, if that's what you're doing?

Vazno:
That's how it works, I got it.

src_py/cursors.py
```diff
@@ -119,15 +119,15 @@ def __hash__(self):
     def __repr__(self):
         if self.type == "system":
             id_string = _cursor_id_table.get(self.data[0], "constant lookup error")
-            return "<Cursor(type: system, constant: " + id_string + ")>"
+            return f"<Cursor(type: system, constant: {id_string})>"
         if self.type == "bitmap":
-            size = "size: " + str(self.data[0])
-            hotspot = "hotspot: " + str(self.data[1])
-            return "<Cursor(type: bitmap, " + size + ", " + hotspot + ")>"
+            size = f"size: {self.data[0]!s}"  # !s for calling str(), or it will call format()
```
Starbuck5, line 124: format? I think you mean str() vs repr().

I've never heard of this "!s" syntax, and it was difficult to Google. I finally found a reference in https://peps.python.org/pep-0498/#s-r-and-a-are-redundant, which calls them redundant. I would suggest using `str()` explicitly instead, that way is more easily understood, and doesn't need the comment at least.

What goes wrong with repr vs str here anyways, I'm having a hard time thinking it through

Vazno: I just changed the look of syntax, nothing changed, if I called f"{self.data}" without !s or str() it would call format() method instead of str() which was originally here. I can change it to str() method.

Starbuck5: >  if I called f"{self.data}" without !s or str() it would call format() method instead of str() which was originally here.

It would not call `format()`

Vazno: https://stackoverflow.com/a/72449614/18343885

Starbuck5: TIL there's a `__format__` dunder method.

Still, does it lead to different results?

Vazno: It probably won't because this variable contains only tuple[int] tuple tuple[, ...]

Starbuck5: That's why I was confused over your changes. I assumed you had tested this and it needed to be this way.

Vazno: I didn't want to change the behavior of function, but it seems there is no point in using that !s method

MyreMylar (approved):
Great! The diff is clearer now.

ankith26 (approved):
LGTM, thanks! 🍰 🥳

ankith26:
Do you mind if this PR is squashed and merged? This is because there are huge diffs in between commits that ultimately are not a part of the PR

Vazno:
Ok, squash it.

PR #2844, opened 2021-11-14
py2 code cleanups + doc updates

ankith26:
Continued from #2833 adding some more fixes

PR #3395, opened 2022-08-16
Fix scrap typing (use only `bytes` and not `AnyStr`)

ankith26:
Simple PR, fixes an issue with `scrap` typestubs which was not properly ported to python 3.

Previously, this used `AnyStr` because of python 2 support, but in python 3 scrap functions never handle/expect data in `str`

I went through the C code and there's no way any `get` can return an `str`, or `put` can take an `str` for the second argument.


buildconfig/stubs/pygame/scrap.pyi
```diff
@@ -1,10 +1,11 @@
...
+from typing import List, Optional
+from collections.abc import ByteString
 
 def init() -> None: ...
 def get_init() -> bool: ...
-def get(data_type: str) -> AnyStr: ...
+def get(data_type: str) -> Optional[bytes]: ...
 def get_types() -> List[str]: ...
-def put(data_type: str, data: AnyStr) -> None: ...
+def put(data_type: str, data: ByteString) -> None: ...
```
Starbuck5, line 8: Why is this ByteString and not just bytes? What is ByteString?

Starbuck5 (commented):
If you're wading into the scrap bytes mess,

There are a couple references to "python 2/3" in the scrap documentation that should be streamlined to use the modern terminology and context and everywhere.

ankith26:
And I've already fixed that in the past :) (in #2844)

Just missed the typehints back then

Starbuck5:
> And I've already fixed that in the past :) 

Ah! I was looking at the online docs, makes sense.

buildconfig/stubs/pygame/scrap.pyi
ankith26: `ByteString` is a generic bytes-like object (which in addition to `bytes`, also accepts few more things like `bytearray` and the like)

I'm using this because I observed the C code for `put` use `y#` format which should handle the same

Starbuck5: Linking you to a discussion I found: https://github.com/python/cpython/issues/91896
I'll support proceeding the current way or another way if that changes your mind.

MyreMylar: sounds like we are still waiting for Buffer to make it from PEP into CPython. I guess we should roll with this as it is for now and revisit when Buffer turns up in our lowest version in a few years?

MyreMylar (approved):
LGTM 👍 the `ByteString` versus `Buffer` debate is out of hands and won't be resolved into a format we can use for a good while.

PR #3396, opened 2022-08-16
Vector clamp shouldn't allow kwargs, improve error reporting and unit tests

ankith26:
The recently added Vector clamp methods accept keyword arguments, but these are named `arg0` and `arg1` in the implementation, but are called `min_value` and `max_value` in the docs and typestubs.

This is because clamp functions are supposed to work like pythons `range` constructor: one value means `max` and two values is a `min-max` pair.
So having kwargs like `arg0` and `arg1` is not very useful, infact it goes against what is documented. So I propose that this should be removed (pythons `range` constructor does not handle kwargs either). This is not a breaking change because the clamp functions are not released in a stable version yet (but this also means that we need to come to a decision on this before a release to keep things smoother)

I updated the code to use `FASTCALL` which also helps with code clarity, and also gives a 65%-ish speed boost in my benchmarking, which is quite decent.

I also improved argument validation in the implementation, and improved upon the tests for these.


illume:
I'm generally for keyword arguments everywhere. (not arg0, arg1 though)

However in this case; can we know what the arguments are reading the code without keyword args? I think it means clamp it to and between the two values given. I'm not sure keyword arguments aids readability at all.

Q: Would be interested in knowing what others think on the keyword/no-keyword issue? Is it really worth not supporting them for a bit of extra speed?

ps. Pretty sick of losing readability and hacks just because cpython is slow ;)

test/math_test.py
```diff
@@ -2703,18 +2759,74 @@ def test_collection_abc(self):
...
         expected_v2 = Vector3(4.008918628686366, 1.3363062095621219, 2.6726124191242437)
-        self.assertAlmostEqual(expected_v2.x, v2.x)
-        self.assertAlmostEqual(expected_v2.y, v2.y)
-        self.assertAlmostEqual(expected_v2.z, v2.z)
+        self.assertEqual(expected_v2, v1)
+        self.assertEqual(expected_v2, v2)
+
+    def test_clamp_mag_v3_no_change(self):
+        v1 = Vector3(1, 2, 3)
+        for args in (
```
illume, line 2784: these could use sub tests

itzpr3d4t0r:
> I'm generally for keyword arguments everywhere. (not arg0, arg1 though)
> 
> However in this case; can we know what the arguments are reading the code without keyword args? I think it means clamp it to and between the two values given. I'm not sure keyword arguments aids readability at all.
> 
> Q: Would be interested in knowing what others think on the keyword/no-keyword issue? Is it really worth not supporting them for a bit of extra speed?
> 
> ps. Pretty sick of losing readability and hacks just because cpython is slow ;)

About this we opened #3324

test/math_test.py
ankith26: Good call, implemented

illume:
As well as readability, I'd like to know if usability is affected? Using editors, are people nudged to use keyword arguments? We have a keywords all the things issue open here: https://github.com/pygame/pygame/issues/808 which includes this gem:
> In the end my main interest was getting better auto-complete and linting support, which seems to have been largely solved by the .pyi files.

But... does including .pyi files with kwargs in there solve our usability issues or does it make it worse?

illume:
I'd like to see what sort of speedups are gained at the macro level (as well as micro benchmarks). Would a game using these be any faster or are the benefits only in micro benchmarks? Often calling code in a tight loop will behave very differently compared to when it's called in amongst a whole bunch of other code. Speed benefits can very quickly disappear when the code is outside of cache, and the data is also un-cached.

illume:
Moved my questions into "[Tracking FASTCALL/VECTORCALL improvements](https://github.com/pygame/pygame/issues/3324)" issue.  Maybe someone knows the answers already, or we need to do some more research.

docs/reST/ref/math.rst
```diff
@@ -406,8 +406,12 @@ Multiple coordinates can be set using slices or swizzling
...
       | :sg:`clamp_magnitude(min_length, max_length) -> Vector2`
 
-      Returns a new copy of a vector with the magnitude clamped between max_length and min_length. If
-      a min_length value is not assigned, then it will automatically be set to 0.
+      Returns a new copy of a vector with the magnitude clamped between 
+      ``max_length`` and ``min_length``. If only one argument is passed, it is 
+      taken to be the ``max_length``
+
+      This function raises ``ValueError`` if ``min_length`` is greater than
+      ``max_length``, or if either of these values are negative.
```
illume, line 415: I think we should mark new APIs Experimental in general. Then after some releases(and after more feedback) the experimental marking can be removed from the docs. What do you think?

zoldalma999: While it does sound nice to mark them as experimental in the docs (and we should do it), I don't think a lot of people would just find it in there. They might see it in a forum or something similar, go "oh cool, I could use that", and then the next time they upgrade, their code is going to break.

Just putting out ideas here, but maybe we could do a system similar to pythons future module, where one could enable specific features one by one? So something like this: 
```py
import pygame
from pygame.future import vector_clamp  # or some other module name

vector_clamp.enable()  # commenting this line will cause the code below to raise AttributeError

vec = pygame.Vector2(10, 20)
vec.clamp_magnitude(5)
```

However this might be too much effort to implement, might not be elegant (in pygames code base) and could also make new contributors a bit confused

MyreMylar:
My general feeling on keyword arguments is that they can aid readability once you move past two or three arguments - before that I would generally prefer to pass by value and have the function name be descriptive enough that I can make a fairly firm guess on what the two parameters are and their order. 

For example, a `clamp()` with two parameters, I would assume that the first argument was the min clamp value and the second the max clamp value. If there was a third argument I would be uncertain and need to look at documentation and in that case a keyword argument in the call would be a great reminder next time I looked at the call. 

I guess as well as number of arguments there is the 'unexpected' or 'optional' factors to consider, even a single parameter might be useful with a keyword argument if it is an entirely optional and occasionally used option.

For reference there are python standard library functions that do not accept keyword arguments. As you might expect they are functions like `pow()` where performance is a concern and the arguments are fairly self evident.

See this stack overflow answer on pow:

https://stackoverflow.com/posts/55970419/revisions

Though that brings up a point - we should make sure our documentation includes the trailing slash syntax that indicates positional only.

docs/reST/ref/math.rst
illume: Good points/ideas.

Another problem with experimental stuff is people abandon it and don't finish the work. Leaving their mess to other people to clean up. I've a feeling this would help some people push things over the line to be finished.

Probably the feature flag/future idea should move to a separate issue.

illume:
> My general feeling on keyword arguments is that they can aid readability once you move past two or three arguments

Yes, but also... depends on the reader. Some readers require more help, and some functions are more complex(so any readability help is important).

The most confusing thing is... which functions are weird position only and which are functions? Adding / to the end of a function definition is not going to help most people be less confused IMHO.

> I guess as well as number of arguments there is the 'unexpected' or 'optional' factors to consider, even a single parameter might be useful with a keyword argument if it is an entirely optional and occasionally used option.

Yeah. To illustrate your point; In your examples probably only `int()` could not be improved with keyword arguments when using a single argument.

But when using optional arguments?

```python
int('1', 4)
```

What does that do? Only if you knew the function would you know what that code does.

```python
int('1', base=4)
```

> For reference there are python standard library functions that do not accept keyword arguments. As you might expect they are functions like pow() or int() where performance is a concern and the arguments are fairly self evident.

I feel a lot of bad design in python comes up because CPython is poorly made for speed. Avoid functions because they are slow. Avoid extra classes, because they are slow. Have weird functions because they are slow to do. Alias variables to locals because it's slow not to. All these hacks would be more elegant if CPython was faster.

Many are confused by the third argument of `range(1, 10, 2)` for example. Even though they've learnt what order the range arguments are in. However, even the first two arguments of range are confusing to newbies. There's no denying that `range(start=1, end=10, step=2)` is more readable.

illume (approved):
:tada: This is a very nice improvement.

I think we should put the keyword discussion aside (to be continued elsewhere). 

All the new APIs need to be at least marked experimental. (perhaps in the future we can have a feature flags/future idea) Can you please mark this one experimental in the docs? Otherwise I'll get to it in a pass over all new APIs before the release.

docs/reST/ref/math.rst
ankith26: Done, these functions are now marked experimental in the docs

illume:
Thanks @ankith26

PR #3399, opened 2022-08-18
Color set length deprecation warnings

### LINKED ISSUE #3393: Remove unneeded calls to color.set_length() from tests & examples
This function has been deprecated and now produces deprecation warnings when running some of the tests (e.g. color_test.py and pixel_copy_test.py). It's fair to have a test for the deprecation warning while the function still exists, and perhaps some tests of it's functionality ([perhaps filtered for warnings](https://stackoverflow.com/a/71498196)) - but we should probably try to excise it from unrelated tests at least.

`set_length()` didn't actually do much, it didn't even clear the data from channels, just blocked access when using index based access so probably it can just be deleted in many places, but some care and attention will be needed.

gresm:
This pull request:
 - Fixes #3393
 - ~~Adds ``.rgb`` property to ``pygame.color.Color`` class~~
 - ~~Performed black/clang-format refactor~~

### DIFF
--- a/test/color_test.py
+++ b/test/color_test.py
@@ -70,6 +70,7 @@ def _suppress_warnings(c):
                 return f(*args, **kwargs)
 
         return new_f
+
     return decorator
 
 
--- a/test/color_test.py
+++ b/test/color_test.py
@@ -2,6 +2,7 @@ import math
 import operator
 import platform
 import unittest
+import warnings
 from collections.abc import Collection, Sequence
 
 import pygame
@@ -61,12 +62,24 @@ def _assign_item(x, p, y):
     x[p] = y
 
 
+def _suppress_warnings(c):
+    def decorator(f):
+        def new_f(*args, **kwargs):
+            with warnings.catch_warnings():
+                warnings.simplefilter("ignore", category=c)
+                return f(*args, **kwargs)
+
+        return new_f
+    return decorator
+
+
 class ColorTypeTest(unittest.TestCase):
     def test_new(self):
         c = pygame.Color.__new__(pygame.Color)
         self.assertEqual(c, pygame.Color(0, 0, 0, 255))
         self.assertEqual(len(c), 4)
 
+    @_suppress_warnings(DeprecationWarning)
     def test_init(self):
         c = pygame.Color(10, 20, 30, 200)
         self.assertEqual(c, (10, 20, 30, 200))
@@ -257,6 +270,7 @@ class ColorTypeTest(unittest.TestCase):
 
         # NOTE: assigning to a slice is currently unsupported.
 
+    @_suppress_warnings(DeprecationWarning)
     def test_unpack(self):
         # should be able to unpack to r,g,b,a and r,g,b
         c = pygame.Color(1, 2, 3, 4)
@@ -268,12 +282,7 @@ class ColorTypeTest(unittest.TestCase):
         r, g, b = c
         self.assertEqual((1, 2, 3), (r, g, b))
 
-        # Checking if DeprecationWarning is triggered
-        # when function is called
-        for i in range(1, 5):
-            with self.assertWarns(DeprecationWarning):
-                c.set_length(i)
-
+    @_suppress_warnings(DeprecationWarning)
     def test_length(self):
         # should be able to unpack to r,g,b,a and r,g,b
         c = pygame.Color(1, 2, 3, 4)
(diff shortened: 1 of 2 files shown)


ankith26 (changes requested):
The diff of this PR is too complex because you formatted previously unformatted files.

The easiest thing is to create a new PR but with only the changes you want to keep.
Alternatively if you are good with git, you can rebase this branch too to remove uneeded changes

Starbuck5:
You run the formatting with `setup.py format`, and it only formats what it is supposed to.

gresm:
@ankith26 @Starbuck5 Fixed issues.

test/pixelcopy_test.py
```diff
@@ -185,6 +198,7 @@ def iter_surface_to_array_3d(self, rgba_masks):
                 )
             view = None
 
+    @_suppress_warnings(DeprecationWarning)
     def test_map_array(self):
```
Starbuck5, line 202: Could the function be removed from this unrelated test instead of the deprecation warning being suppressed here?

gresm: I don't really see any way how it would be possible. Just simply removing ``color.set_length(3)`` breaks the test completely and I don't know how to get change the code that use it, so it will work. Also I don't think removing that part of the check would be a good idea. But, if you have any suggestions, I'm open.

Starbuck5: color[:3] ?

test/color_test.py
```diff
@@ -990,6 +1006,7 @@ def test_pickle(self):
     # only available if ctypes module is also available
 
     @unittest.skipIf(IS_PYPY, "PyPy has no ctypes")
+    @_suppress_warnings(DeprecationWarning)
     def test_arraystruct(self):
```
Starbuck5, line 1010: Same here. ArrayInterface(color[:i]) will work perhaps. But this is not a test of set_length functionality, it's an incidental use, so it should be removed.

test/pixelcopy_test.py
gresm: @Starbuck5 "ValueError: tuple object does not export an array buffer", I tried it already...

test/color_test.py
gresm: Apparently ``np.array(color[:3])`` also works...

test/pixelcopy_test.py
gresm: ``np.array(color[:3])`` resolves that issue.

test/color_test.py
gresm: @Starbuck5 I couldn't make it work (your proposition doesn't work, only after some modifications I managed to make it working), but after that I discovered that a lot of sub-checks in that part failed. I started fixing by making modifications for that part of code, but I think there is no need to change anything, as the ``test_arraystruct`` performs (I think) some checks with that ``color.set_length``. Probably there is something that I can do there, but I don't want to touch that part of the code, as I don't fully understand it...

illume:
I wonder if this turns up in vscode/etc as deprecated? I know in typescript/js you can use the "`@deprecated`" line in the functions documentation for that to happen. Not sure if it's possible in python.

ankith26:
The best we can do that I know of is to put a doc string in the typestubs of the function/class being deprecated that clearly mentions the same. Most IDEs should show it on mouse hover, but this won't trigger a warning at code analysis time

ankith26 (approved):
LGTM after a skim through the code, thanks for contributing! 🍪

test/color_test.py
```diff
@@ -268,12 +283,7 @@ def test_unpack(self):
         r, g, b = c
         self.assertEqual((1, 2, 3), (r, g, b))
 
-        # Checking if DeprecationWarning is triggered
-        # when function is called
-        for i in range(1, 5):
-            with self.assertWarns(DeprecationWarning):
-                c.set_length(i)
-
+    @_suppress_warnings(DeprecationWarning)
```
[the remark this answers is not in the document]
MyreMylar: Maybe move the deprecation warning test to a new separate function?

MyreMylar (changes requested):
Added a few nitpicks.

test/color_test.py
gresm: This is how I understand this test:
It checks what structure format Color object has with different color lengths, so changing the Color object to tuple and test it, is pointless because: tuple has different arrangement and checking the structure of tuple provides not a very valuable information. So I think this test is solely for testing ``set_length`` functionality and should be removed in the future.

PR #3400, opened 2022-08-18
Fix edge case bug in Vector move towards, add more tests

ankith26:
Found a bug in move towards while messing around in vectors code
```py
>>> import pygame
pygame 2.1.3.dev5 (SDL 2.0.22, Python 3.10.5)
>>> a = pygame.Vector2(10, 10)
>>> a.move_towards(a, -1)
<Vector2(-nan(ind), -nan(ind))>
```

When a vector move towards itself, it should return itself in all cases. But right now, it returns a vector with `NaN` coords if `max_distance` also happens to be negative.

I added tests to catch this bug, and while I was at it I also noticed that move towards tests were missing for `Vector3`, so I added those in too (based on the existing `Vector2` tests)

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -1382,9 +1382,7 @@ _vector_move_towards_helper(Py_ssize_t dim, double *origin_coords,
                             double *target_coords, double max_distance)
 {
     Py_ssize_t i;
-    double delta[VECTOR_MAX_SIZE];
-    double dist;
-
+    double frac, dist, delta[VECTOR_MAX_SIZE];
     if (max_distance == 0)
         return;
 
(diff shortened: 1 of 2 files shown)


Starbuck5 (approved):
This all looks good to me. I'd like @novialriptide's review on this.

src_c/math.c
```diff
@@ -1401,10 +1404,9 @@ _vector_move_towards_helper(Py_ssize_t dim, double *origin_coords,
         return;
     }
 
+    frac = max_distance / dist;
     for (i = 0; i < dim; ++i)
-        origin_coords[i] = origin_coords[i] + delta[i] / dist * max_distance;
-
-    return;
```
novialriptide, line 1407: I don't even know why I put `return;` here.. lmao

novialriptide (approved):
LGTM! Thanks for this.

zoldalma999 (approved):
Looks good to me!

illume:
cool

ps. I'd like to freeze merging until the release is done. If there's any release blockers, please note them in this issue: https://github.com/pygame/pygame/issues/3408

illume (commented):
👍

PR #3401, opened 2022-08-18
Optimized rect getters

itzpr3d4t0r:
# Info
A big part of rect getters returns a tuple of two int values by using `Py_BuildValue`. Apparently this function is slower than creating a tuple, setting its values and returning it, so i changed those getters to do just that.
# Performance
This change brings an approx. 40% performance improvement on average, some results:
OLD
```
====||RECT||====
-) size: 0.73509932 s
-) center: 0.66942516 s
-) topright: 0.61553792 s
-) topleft: 0.6227682 s
-) bottomright: 0.61191862 s
-) bottomleft: 0.6156618 s
-) midtop: 0.61765648 s
-) midbottom: 0.63002794 s
-) midleft: 0.61581018 s
-) midright: 0.60969368 s
______________
-) Mean time: 0.63435993
```
NEW
```
====||RECT||====
-) size: 0.49888236 s
-) center: 0.54001986 s
-) topright: 0.50331364 s
-) topleft: 0.45303602 s
-) bottomright: 0.41383036 s
-) bottomleft: 0.43431768 s
-) midtop: 0.41643356 s
-) midbottom: 0.41696528 s
-) midleft: 0.42823242 s
-) midright: 0.41350376 s
______________
-) Mean time: 0.45185349
```

Code i used:
```Python
import timeit
from statistics import fmean

import pygame

pygame.init()


def test(test_name: str, func: str, globs: dict, num: int,
         mode: str = "mean") -> float:
    value = 0
    if mode == "mean":
        value = fmean(timeit.repeat(func, globals=globs, number=num))
    elif mode == "cumulative":
        value = timeit.timeit(func, globals=globs, number=num)
    else:
        raise NotImplementedError("Incorrect test mode")
    print("-) " + test_name + f": {round(value, 8)} s")
    return value


def test_group(group_name: str, sequence: list[str],
               globs: dict,
               tests_names: tuple[str, ...] = None, num: int = 10_000_000,
               include_tot: bool = False, include_mean: bool = True,
               test_mode: str = "mean") -> float:
    print("\n====||" + group_name.upper() + "||====")

    test_data = [
        test(tests_names[i] if tests_names else str(i + 1), test_obj, globs,
             num, test_mode)
        for
        i, test_obj in
        enumerate(sequence)]

    print("______________")
    if include_tot:
        print(f"-) Total time: {round(sum(test_data), 8)}")

    if include_mean:
        print(f"-) Mean time: {round(fmean(test_data), 8)}")

    return fmean(test_data)


r1 = pygame.rect.Rect(0, 0, 10, 10)

NUMBER = 10000000
GLOB = {"r1": r1}

test_names = (
    "size", "center", "topright", "topleft", "bottomright", "bottomleft", "midtop",
    "midbottom", "midleft", "midright")
test_group(
    "Rect", [
        "r1.size",
        "r1.center",
        "r1.topright",
        "r1.topleft",
        "r1.bottomright",
        "r1.bottomleft",
        "r1.midtop",
        "r1.midbottom",
        "r1.midleft",
        "r1.midright",
    ],
    GLOB,
    test_names
)
```

### DIFF
--- a/src_c/rect.c
+++ b/src_c/rect.c
@@ -37,6 +37,27 @@
 static PyTypeObject pgRect_Type;
 #define pgRect_Check(x) ((x)->ob_type == &pgRect_Type)
 
+#define PG_RETURN_TUPLE_FROMVALUES_INT(val1, val2) \
+    do {                                           \
+        PyObject *tup = PyTuple_New(2);            \
+        if (!tup) {                                \
+            return NULL;                           \
+        }                                          \
+        PyObject *tmp = PyLong_FromLong(val1);     \
+        if (!tmp) {                                \
+            Py_DECREF(tup);                        \
+            return NULL;                           \
+        }                                          \
+        PyTuple_SET_ITEM(tup, 0, tmp);             \
+        tmp = PyLong_FromLong(val2);               \
+        if (!tmp) {                                \
+            Py_DECREF(tup);                        \
+            return NULL;                           \
+        }                                          \
+        PyTuple_SET_ITEM(tup, 1, tmp);             \
+        return tup;                                \
+    } while (0)
+
 static int
 pg_rect_init(pgRectObject *, PyObject *, PyObject *);
 
@@ -1835,13 +1856,7 @@ pg_rect_setcentery(pgRectObject *self, PyObject *value, void *closure)
 static PyObject *
 pg_rect_gettopleft(pgRectObject *self, void *closure)
 {
-    PyObject *tuple = PyTuple_New(2);
-    if (!tuple) {
-        return NULL;
-    }
-    PyTuple_SET_ITEM(tuple, 0, PyLong_FromLong(self->r.x));
-    PyTuple_SET_ITEM(tuple, 1, PyLong_FromLong(self->r.y));
-    return tuple;
+    PG_RETURN_TUPLE_FROMVALUES_INT(self->r.x, self->r.y);
 }
 
 static int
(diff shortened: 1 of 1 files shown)


ankith26 (changes requested):
`PyLong_FromLong` can fail with returning `NULL`, and this will later cause problems once a tuple is returned and it has `NULL` element(s)

I'd also suggest making this a macro to reduce some code repetition

itzpr3d4t0r:
Yep sure, was thinking of adding a macro already

src_c/rect.c
```diff
@@ -37,6 +37,27 @@
 static PyTypeObject pgRect_Type;
 #define pgRect_Check(x) ((x)->ob_type == &pgRect_Type)
 
+#define PG_RETURN_TUPLE_FROMVALUES_INT(val1, val2) \
```
illume, line 40: I wonder if there is a speed difference if this is a function?

How's the generated code size?

itzpr3d4t0r: > I wonder if there is a speed difference if this is a function?

Apparently not, just tested.

> How's the generated code size?

This I didn't test

illume: > Apparently not, just tested.

Cool

illume (approved):
Impressive results!

(nit: I'd prefer a function instead of a macro)

ankith26 (approved):
LGTM now, thanks! 😎

PR #3404, opened 2022-08-19
Add rgb value to color

gresm:
Cleaner way of ``color[:3]``: ``color.rgb``.

### DIFF
--- a/src_c/color.c
+++ b/src_c/color.c
@@ -130,6 +130,10 @@ _color_get_cmy(pgColorObject *, void *);
 static int
 _color_set_cmy(pgColorObject *, PyObject *, void *);
 static PyObject *
+_color_get_rgb(pgColorObject *, void *);
+static int
+_color_set_rgb(pgColorObject *, PyObject *, void *);
+static PyObject *
 _color_get_arraystruct(pgColorObject *, void *);
 
 /* Number protocol methods */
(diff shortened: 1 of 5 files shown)


src_c/color.c
```diff
@@ -1423,6 +1441,46 @@ _color_set_i1i2i3(pgColorObject *color, PyObject *value, void *closure)
     return 0;
 }
 
+static PyObject *
+_color_get_rgb(pgColorObject *color, void *closure)
+{
+    return Py_BuildValue("(iii)", color->data[0], color->data[1],
```
itzpr3d4t0r, line 1447: Just a question, with my recent PR #3401 i discovered that creating a tuple, setting its arguments and returning it is about 40% faster than using `Py_BuildValue`, so can we implement this here already insted of opening another PR later? Do we need/want to do this?

src_c/color.c
```diff
@@ -298,6 +304,18 @@ _get_double(PyObject *obj, double *val)
...
 
+static int
+_get_int(PyObject *obj, int *val)
+{
+    PyObject *intobj;
+    if (!(intobj = PyNumber_Long(obj))) {
+        return 0;
+    }
+    *val = PyLong_AsLong(intobj);
+    Py_DECREF(intobj);
```
itzpr3d4t0r, line 315: Is there a reason for doing this over just a `*val = PyLong_AsLong(obj)`?
You could just do 
```C
static int
_get_int(PyObject *obj, int *val) {
    *val = PyLong_AsLong(obj);
    if (PyErr_Occurred())
        return 0;
    return 1;
}
```
Or possibly better 
```C
static int
_get_int(PyObject *obj, int *val) {
    if (PyFloat_Check(obj)) {
        *val = (int)PyFloat_AS_DOUBLE(obj);
        return 1;
    }
    *val = PyLong_AsLong(obj);
    if (PyErr_Occurred())
        return 0;
    return 1;
}
```

src_c/color.c
gresm: @MyreMylar Thank you, so I'm waiting for a merge of that.

PR #3416, opened 2022-08-20
Go back to old memcpy strategy for non-flip flip

Starbuck5:


illume (approved):
👍 thanks!

src_c/transform.c
```diff
@@ -823,8 +823,9 @@ surf_flip(PyObject *self, PyObject *args, PyObject *kwargs)
 
     if (!xaxis) {
         if (!yaxis) {
-            assert(srcpitch == dstpitch);
-            memcpy(dstpix, srcpix, srcpitch * surf->h);
+            for (loopy = 0; loopy < surf->h; ++loopy)
+                memcpy(dstpix + loopy * dstpitch, srcpix + loopy * srcpitch,
+                       surf->w * surf->format->BytesPerPixel);
```
ankith26, line 828: How about this.... we do a conditional check?
`if (srcpitch == dstpitch)` we take the faster path, otherwise fallback to this path?

I think in this way we both get the speed boost, and safety so sounds like a win-win to me unless I'm missing something

Starbuck5: It's not much of a speed boost, and it's for a trivial case anyways, so I'm not too concerned

(the case where someone passes in a surface and says not to flip either X or Y direction)

ankith26 (approved):
LGTM! 🎉

Issue #2577, opened 2021-05-04
pygame.mixer.Sound.copy


fortwoone:
I noticed there were copy protocols for some important Pygame objects, such as Surfaces and Rects. I wondered if it would have some interest to add one to Sound objects too, because it would help in some situations, for example if you don't want to use imported sounds directly (also to avoid direct instance editing as we may encounter with Surfaces if we don't copy them before editing)...

fortwoone:
And so I found a technique to copy Sound objects without using a copy method : you only need to make an array from the source Sound and then make another Sound object from the array

illume:
Would be useful to me too.

fortwoone:
> 
> 
> Would be useful to me too.

Do you want me to give you my piece of code ?

novialriptide:
Can I work on this issue? Or is there existing code that can be merged

PR #3418, opened 2022-08-21
Add pygame.mixer.Sound.copy()

novialriptide:
Fixes: https://github.com/pygame/pygame/issues/2577


docs/reST/ref/mixer.rst
```diff
@@ -401,6 +401,14 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
...
 
       .. ## Sound.get_raw ##
+   
+   .. method:: copy
+
+      | :sl:`copy the Sound`
+      | :sg:`copy() -> Sound`
+      
+      Returns a new Sound object with the same data and volume as the original.
+   .. ## pygame.mixer.Sound.copy ##
```
illume, line 411: Needs a `      .. versionadded:: 2.1.x`

test/mixer_test.py
```diff
@@ -1183,6 +1183,24 @@ def test_get_raw(self):
             with self.assertRaisesRegex(pygame.error, "mixer not initialized"):
                 snd.get_raw()
 
+    def test_copy(self):
+        """Test that the copy method would return correct values and etc"""
+        samples = b"abcdefgh"  # keep byte size a multiple of 4
```
illume, line 1188: I have a feeling that buffer requires a different amount of samples depending on what the mixer initializes as. So, probably this test could also test if this copies correctly under different mixer initializations.

test/mixer_test.py
```diff
@@ -1183,6 +1183,24 @@ def test_get_raw(self):
...
+        a = mixer.Sound(buffer=samples)
+        a.set_volume(0.7)
+        b = a.copy()
+        self.assertEqual(a.get_raw(), b.get_raw())
+        self.assertEqual(a.get_volume(), b.get_volume())
+
+        class Amogus(mixer.Sound):
```
illume, line 1195: I think checking if sub classes work should be in a separate test method.

ps. What is amo gus?

test/mixer_test.py
```diff
@@ -1183,6 +1183,24 @@ def test_get_raw(self):
...
+            pass
+
+        c = Amogus(buffer=samples)
+        d = c.copy()
+        self.assertEqual(type(d), Amogus)
+        self.assertEqual(d.get_raw(), c.get_raw())
+        self.assertEqual(d.get_volume(), c.get_volume())
```
illume, line 1202: I wonder if channels work properly?

Another test which plays a longer sound, does a copy, and then compares the number of channels would be good.

I'd expect the channels to not be copied? But I'm unsure... what do you think should happen to channels when the sound is copied?

Asserting the .get_num_channels() is zero in the other tests is probably a good idea.

docs/reST/ref/mixer.rst
```diff
@@ -401,6 +401,14 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
       .. versionadded:: 1.9.2
 
       .. ## Sound.get_raw ##
+   
+   .. method:: copy
+
+      | :sl:`copy the Sound`
+      | :sg:`copy() -> Sound`
+      
+      Returns a new Sound object with the same data and volume as the original.
```
illume, line 410: Would be nice to specify what happens to channels when copying.

illume (commented):
Nice addition.

I'd probably use this method in the parrotjoy app.

src_c/mixer.c
```diff
@@ -760,6 +760,23 @@ snd_get_raw(PyObject *self, PyObject *_null)
...
+    Mix_Chunk *old_chunk = pgSound_AsChunk(self);
+    Mix_Chunk *new_chunk = Mix_QuickLoad_RAW(old_chunk->abuf, old_chunk->alen);
+    new_chunk->volume = old_chunk->volume;
+    pgSoundObject *sound =
+        (pgSoundObject *)pgSound_Type.tp_new(Py_TYPE(self), NULL, NULL);
+    if (sound) {
+        sound->mem = NULL;
+        sound->chunk = new_chunk;
+    }
+    sound->chunk = new_chunk;
```
novialriptide, line 775: Seems to be duplicate code here... I'll fix this in a future commit.

test/mixer_test.py
novialriptide: I think the channels should be left alone when a Sound is copied. I'm not sure if that's desirable though.

src_c/mixer.c
```diff
@@ -760,6 +760,22 @@ snd_get_raw(PyObject *self, PyObject *_null)
...
+{
+    Mix_Chunk *old_chunk = pgSound_AsChunk(self);
+    Mix_Chunk *new_chunk = Mix_QuickLoad_RAW(old_chunk->abuf, old_chunk->alen);
+    new_chunk->volume = old_chunk->volume;
+    pgSoundObject *sound =
+        (pgSoundObject *)pgSound_Type.tp_new(Py_TYPE(self), NULL, NULL);
+    if (sound) {
+        sound->mem = NULL;
+        sound->chunk = new_chunk;
+    }
```
Starbuck5, line 774: I suppose you're saying the tp_new would return NULL and raise a python exception if it fails?

Is that right? That's clever.

It could be good to have a comment there explaining the if block is to try to set attributes only if not NULL, and if it's NULL the exception has already been raised.

src_c/mixer.c
```diff
@@ -760,6 +760,22 @@ snd_get_raw(PyObject *self, PyObject *_null)
...
 }
 
+static PyObject *
+snd_copy(PyObject *self, PyObject *_null)
+{
+    Mix_Chunk *old_chunk = pgSound_AsChunk(self);
+    Mix_Chunk *new_chunk = Mix_QuickLoad_RAW(old_chunk->abuf, old_chunk->alen);
+    new_chunk->volume = old_chunk->volume;
+    pgSoundObject *sound =
+        (pgSoundObject *)pgSound_Type.tp_new(Py_TYPE(self), NULL, NULL);
```
ankith26, line 770: I'd change this to something like `Py_TYPE(self)->tp_new(Py_TYPE(self), NULL, NULL);` so that the `__new__` of the subclass is used (if overridden)

Additionally, I'd like if the `Mix_QuickLoad_RAW` came *after* a successful `tp_new` call, and complete with error handling for when it fails

src_c/mixer.c
novialriptide: I actually have the if statement checker there because I kept seeing that pattern around the source code.

src_c/mixer.c
```diff
@@ -790,6 +790,35 @@ snd_get_raw(PyObject *self, PyObject *_null)
...
+        (pgSoundObject *)Py_TYPE(self)->tp_new(Py_TYPE(self), NULL, NULL);
+
+    if (sound) {
+        new_chunk = Mix_QuickLoad_RAW(old_chunk->abuf, old_chunk->alen);
+        if (new_chunk) {
+            new_chunk->volume = old_chunk->volume;
+        }
+        else {
+            PyErr_Format(PyExc_SystemError,
+                         "Internal pygame error, chunk could not be created.");
```
MyreMylar, line 808: Here, and below we are creating an error, but not immediately exiting the C code via a return, which I think is the normal practice in the pygame codebase.

I assume it won't cause too many problems here, as no doubt these functions return NULL when erroring already, so we'll just end up with returning NULL in the end. However it would probably be a good idea to be explicit that we return NULL/0 here on error by just sticking a return in front of `PyErr_Format`.

MyreMylar (changes requested):
Looks alright to me, I will try and give it a better testing this weekend.

novialriptide:
Test failure seems irrelevant

MyreMylar:
> Test failure seems irrelevant

Think that failure is related to the sphinx 6 upgrade issue (now fixed in main) from a few weeks back, I tried to rebuild but it probably just needs this PR to merge main again or something.

novialriptide:
It seems like I got lucky with the merge from main, no merge conflicts.

MyreMylar (approved):
This looks good to me. It does what I expect when copied for the way sounds work in pygame.

Sounds do not actually carry much state with them and thus the copy also does not carry anything with it. A sound that has been `.play()`'d is not also playing when copied to a new sound, because play 'sends' a sound to a channel to play on rather than doing anything within the sound object itself.

novialriptide:
The person who'll merge this pull request should also squash all these commits into one cuz too many commits for a simple implementation lol

PR #3425, opened 2022-08-23
Add an AVX2 version of alphablit to opaque

Starbuck5:
Ok @MyreMylar I know I told you explicitly I was not going to do this, but I got interested and I did it anyways...

Points of interest:
- No stride switching, using mask load / store with a dynamic mask! Could my mask strategy be an efficiency / code simplicity boon for the other AVX2 blitters?
- I didn't mask out the alpha from the result before writing to dst, which the SSE blitter does.
    - dst is XRGB, I believe.
    - it seems to work fine, is this okay being undefined behavior?

Test program:
```py
import pygame
import time

pygame.init()

screen = pygame.Surface((800,100), depth=32)
screen.fill((60, 70, 80))

surf = pygame.Surface((800,100), pygame.SRCALPHA)
surf.fill((5,216,77,120))

print(screen, surf)

start = time.time()
for _ in range(100000):
    screen.blit(surf, (0, 0))    
print(time.time() - start)
```

Results:
```
# 100k 800x100 alpha opaque dst blits
# existing main: 5.84 seconds
# SSE2 enhanced: 4.84 seconds (my other SIMD PR)
# AVX2: 1.62 seconds
```

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -219,21 +219,10 @@ SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
                             src->format->Rmask == dst->format->Rmask &&
                             src->format->Gmask == dst->format->Gmask &&
                             src->format->Bmask == dst->format->Bmask) {
-                            /* If our source and destination are the same ARGB
-                               32bit format we can use SSE2 to speed up the
-                               blend */
-                            if ((SDL_HasAVX2() == SDL_TRUE) && (src != dst)) {
-                                if (info.src_blanket_alpha == 255 &&
-                                    !(SDL_ISPIXELFORMAT_ALPHA(
-                                          dst->format->format) &&
-                                      info.dst_blend != SDL_BLENDMODE_NONE)) {
-                                    alphablit_alpha_avx2_argb_no_surf_alpha_opaque_dst(
-                                        &info);
-                                    break;
-                                }
-                            }
-#if PG_ENABLE_ARM_NEON
-                            if ((SDL_HasNEON() == SDL_TRUE) && (src != dst)) {
+/* If our source and destination are the same ARGB 32bit
+   format we can use SSE2/NEON/AVX2 to speed up the blend */
+#if PG_ENABLE_SSE_NEON
+                            if ((pg_HasSSE_NEON()) && (src != dst)) {
                                 if (info.src_blanket_alpha != 255) {
                                     alphablit_alpha_sse2_argb_surf_alpha(
                                         &info);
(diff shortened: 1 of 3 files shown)


MyreMylar:
I knew the siren song of AVX2 optimising would suck you in :D

I shall have a look at this.

src_c/alphablit.c
```diff
@@ -192,8 +192,19 @@ SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
                             src->format->Rmask == dst->format->Rmask &&
                             src->format->Gmask == dst->format->Gmask &&
                             src->format->Bmask == dst->format->Bmask) {
-/* If our source and destination are the same ARGB 32bit
-   format we can use SSE2 to speed up the blend */
+                            /* If our source and destination are the same ARGB
+                               32bit format we can use SSE2 to speed up the
```
MyreMylar, line 196: I'm guessing this SSE2 comment should now refer to AVX2 or SIMD generically.

src_c/simd_blitters_avx2.c
```diff
@@ -12,6 +12,206 @@
...
+            temp = _mm256_sub_epi16(mm256_srcB, mm256_dstB);
+            temp = _mm256_mullo_epi16(temp, mm256_srcAlpha);
+
+            mm256_dstB = _mm256_slli_epi16(mm256_dstB, 8);
+            mm256_dstB = _mm256_add_epi16(mm256_dstB, temp);
+            mm256_dstB = _mm256_add_epi16(mm256_dstB, mm256_srcB);
+            mm256_dstB = _mm256_srli_epi16(mm256_dstB, 8);
+
+            /* ==== recombine A and B pixels and store ==== */
+            mm256_dst = _mm256_packus_epi16(mm256_dstA, mm256_dstB);
```
MyreMylar, line 141: Is this chunk now identical to the middle of the 8 pixel width chunk below?

Perhaps we can reduce the duplication somewhat with an inline function call, or a macro - assuming it doesn't affect performance.

Would make it clearer that only the loading and storing changes between the two widths of blend.

The masked load and store seem super useful!

Starbuck5: I'm open to doing that.

I was thinking about macros with these functions-- there could be a possibility where you have something like

Not even close to the right syntax, but an illustration:
```
MACRO build_avx2_blitter (name)
{arbitrary setup code}
{stride switch code}
{full force code}
```

Where the boilerplate of parsing the blitinfo and iterating is handled by the macro to reduce code duplication.

Starbuck5: I have high hopes for macros, but for this PR I'd like to just move forward as is. Macro+masked-load and store could be brought to the blitters in another PR.

PR #3312, opened 2022-07-16
Add use_compat to key.name, fix key_code compat issues

ankith26:
PR to fix #3285 

Basically, I noticed that `name` uses a custom compat table to be as close to pygame 1 as possible, but `key_code` does not use this, and causes a discrepancy in some names.
While I was fixing this, I also added a new optional argument `use_compat` (I'm open to discussing alternate kwarg names here) which basically allows users to access the newer SDL2 names if they want to (by default, the old names are used)

I updated the unittests to test each key, and documented the usage of `use_compat`

It is to be noted that constants that were different in pygame 1 and are aliased in pygame 2 (things like break/pause, lsuper/lmeta and rsuper/rmeta) cannot be properly compat-ed.

To illustrate and test the changes, I generated a table of key-constant and name pairs on pygame 1.9.6, pygame 2.1.2 and this PR (pygame 2.1.3dev5).
The places where you see a `(fail)` in the table, it means that the name returned by `key.name` was not recognised by `key.key_code`

<details>
<summary>Table for key constant and name pairs</summary>

|                   | 1.9.6 (use_compat=None)   | 2.1.2 (use_compat=None)   | 2.1.3.dev5 (use_compat=True)   | 2.1.3.dev5 (use_compat=False)   |
|-------------------|---------------------------|---------------------------|--------------------------------|---------------------------------|
| K_0               | 0                         | 0                         | 0                              | 0                               |
| K_1               | 1                         | 1                         | 1                              | 1                               |
| K_2               | 2                         | 2                         | 2                              | 2                               |
| K_3               | 3                         | 3                         | 3                              | 3                               |
| K_4               | 4                         | 4                         | 4                              | 4                               |
| K_5               | 5                         | 5                         | 5                              | 5                               |
| K_6               | 6                         | 6                         | 6                              | 6                               |
| K_7               | 7                         | 7                         | 7                              | 7                               |
| K_8               | 8                         | 8                         | 8                              | 8                               |
| K_9               | 9                         | 9                         | 9                              | 9                               |
| K_AC_BACK         | N.A                       | AC Back                   | AC Back                        | AC Back                         |
| K_AMPERSAND       | &                         | &                         | &                              | &                               |
| K_ASTERISK        | *                         | *                         | *                              | *                               |
| K_AT              | @                         | @                         | @                              | @                               |
| K_BACKQUOTE       | `                         | `                         | `                              | `                               |
| K_BACKSLASH       | \                         | \                         | \                              | \                               |
[...]

PR #3426, opened 2022-08-23
Modify key compat table to store key-name pairs (fixes a rare bug)

ankith26:
This is a follow up PR for #3312

So I was testing pygame on Ubuntu 22.04 WSL today, and I found that the tests I added in the linked PR was failing. I could also reproduce this fail on my rpi, so IG this bug is not very rare? This same test passed on windows, mac and linux (centos under docker).

The problem is: That PR makes use of the SDL function `SDL_GetKeyFromScancode`, and for some reason the keymap on WSL keeps changing frequently (and incorrectly too as I suspect). I don't know what other platforms this happens on, but it does under WSL

In this PR I'm trying a new, and hopefully more robust strategy (which should give more consistent results). The older implementation hardcoded scancode-name mapping for compat, and there was an intermediate step of keycode-scancode conversions. Instead of having that intermediate step, I've added a direct keycode-name table (since public pygame API does not deal with scancodes ATM)

### DIFF
--- a/src_c/key.c
+++ b/src_c/key.c
@@ -145,477 +145,174 @@ key_get_pressed(PyObject *self, PyObject *_null)
     return ret_obj;
 }
 
-/* Keep our own table for backwards compatibility. This table is directly taken
- * from SDL2 source (but some SDL1 names are patched in it at runtime).
- * This has to be kept updated (only new things can be added, existing names in
- * this must not be changed) */
-static const char *SDL1_scancode_names[SDL_NUM_SCANCODES] = {
-    NULL,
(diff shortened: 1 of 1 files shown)


zoldalma999 (approved):
Looks good to me!

MyreMylar (approved):
LGTM 👍 

Hopefully that finally puts this to bed.

ankith26:
Force pushed to fix a merge conflict in here.
I wouldn't say this fix is critical (because it does not affect most of the commonly used keys), so I'll push this to 2.1.4 milestone

PR #3428, opened 2022-08-26
Relax SDL minor version exact match check

### LINKED ISSUE #3427: pygame assumes that SDL minor version bump implies ABI break, but this is no longer true
**Environment:**

- **Operating system**: Linux (Debian)
- **Python version**: 3.10
- **SDL version**: 2.24
- **PyGame version**: 2.1.2
- **Relevant hardware**: n/a

**Current behavior:**

The version of pygame currently in Debian was compiled against SDL 2.0.22. When run after upgrading SDL to 2.24.0, it crashes with this message:

```
RuntimeError: SDL compiled with version 2.0.22, linked to 2.24.0
```

**Expected behavior:**

pygame should either not second-guess the library dependency system at all, or only check the major version.

SDL 2.x has always been backwards-compatible as far as SDL 2.0.0, which made the previous interpretation of the version number fairly useless: the minor version was stuck at zero, and the micro version was used for "feature" updates, leaving no way to make bugfix-only point releases.

Since SDL 2.24.0, there is a [new versioning scheme](https://github.com/libsdl-org/SDL/blob/main/docs/README-versions.md): the minor version is used for "feature" updates and to indicate development vs. stable status, leaving the micro version available to indicate point releases. This is the same versioning scheme used by GLib, GTK, Flatpak and older versions of the Linux kernel.

**Steps to reproduce:**

1. Build pygame against an older version of SDL. The one currently in Debian testing was compiled with 2.0.22.
2. Upgrade SDL. The version currently in Debian unstable is 2.24.0.
3. Run any pygame game. The bug reporter in Debian used `bambam`.

**Test code**

```
import pygame
pygame.init()
```

```
  File "/usr/games/bambam", line 433, in <module>
    main()
  File "/usr/games/bambam", line 426, in main
    bambam.run()
  File "/usr/games/bambam", line 343, in run
    pygame.init()
RuntimeError: SDL compiled with version 2.0.22, linked to 2.24.0
```

ankith26:
fixes #3427

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -173,8 +173,9 @@ pg_SetDefaultWindowSurface(pgSurfaceObject *);
 static char *
 pg_EnvShouldBlendAlphaSDL2(void);
 
+/* compare compiled to linked, raise python error on incompatibility */
 static int
-pg_CheckSDLVersions(void) /*compare compiled to linked*/
+pg_CheckSDLVersions(void)
 {
     SDL_version compiled;
     SDL_version linked;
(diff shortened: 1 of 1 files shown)


src_c/base.c
```diff
@@ -173,43 +173,32 @@ pg_SetDefaultWindowSurface(pgSurfaceObject *);
...
-        /* SDL 2.0.14 replaces some macros with symbols, see
-         * https://github.com/libsdl-org/SDL/commit/316ff3847b4d9d87d9b0aab15321461db0e8ae0b
-         */
-        PyErr_Format(PyExc_RuntimeError,
-                     "Known SDL incompatibility detected! (compiled with "
-                     "version %d.%d.%d, linked to %d.%d.%d)",
-                     compiled.major, compiled.minor, compiled.patch,
-                     linked.major, linked.minor, linked.patch);
-        return 0;
-    }
```
MyreMylar, line 207: Is this no longer the case? 

I had a skim through the linked SDL discussion. Have they re-released SDL 2.0.14 as  SDL 2.14.0 now?
It is a bit confusing.

MyreMylar:
How does this change to the versioning system handle old versions of SDL still out in the wild?

src_c/base.c
ankith26: They have not re-released older versions, they only adopted a new versioning scheme starting from 2.24.0 (which is the stable version released after 2.0.22)

Because the versioning system changed like this, this check is is not robust enough anymore (for instance, this case would execute if we have a pygame build compiled with a hypothetical SDL 2.24.14 which linked with 2.0.12, which should still error, but it is not the "known SDL incompatibility" this message tells)

I just completely removed this block because it was merely a special case of the below generic error reporting block, and this did not specify more helpful information anyways. (in both cases, a fresh compile/install would be needed)

smcv: 2.0.14 is still 2.0.14, nothing has changed for past versions (we don't have a time machine, we couldn't change past versions even if we wanted to).

The block that special-cased 2.0.14 (if compiled version >= 2.0.14 and 2.0.14 > linked version then error) looks redundant with the block below it (if compiled version > linked version then error), so it seems valid to remove.

smcv:
> How does this change to the versioning system handle old versions of SDL still out in the wild?

No change. All the stable versions of SDL2 released before 2.24.0 were of the form 2.0.x, where 2.0.x is backwards-compatible with 2.0.y, for any x >= y. It is still true that 2.0.x is backwards-compatible with 2.0.y for any x >= y.

What has changed is that before 2.24, SDL upstream didn't specify whether 2.x.a is compatible with 2.y.b, where x != y. Before this MR, pygame pessimistically assumed that a minor version change is a compatibility break, similar to the way SDL 1 worked (SDL 1.2 was incompatible with SDL 1.3).

Since 2.24, SDL upstream is declaring that all (stable) versions of `-lSDL2` (`sdl2.pc`, `libSDL2.so` on Linux, `SDL2.lib`/`SDL2.dll` on Windows, etc.) are going to be backwards-compatible with older stable versions of `-lSDL2`. Next time there is an intentionally incompatible version, SDL upstream is now promising that it will be renamed to SDL 3 (`sdl3.pc`, linked using `-lSDL3`, `libSDL3.so` on Linux, `SDL3.lib`/`SDL3.dll` on Windows, etc.) and generally behave like a separate library, similar to what happened at the jump between SDL 1 and SDL 2.

smcv:
I see pygame uses `sdl2-config` on Unix. Another consequence of what I said in my previous comment is: SDL upstream is now saying that `sdl2-config` will always find a version of SDL 2 that is backwards-compatible with all older stable versions of SDL 2. Next time there is an intentional incompatible change, the library will be renamed to SDL 3, meaning `sdl2-config` will either be renamed to `sdl3-config` or removed in favour of `pkg-config sdl3` (see #3429).

src_c/base.c
robertpfeiffer: It was deliberately special-cased because 1. the original PR only had this case and not the general case and 2. we wanted a special error MSG for this case

robertpfeiffer (approved):

ankith26:
I've tested this patch on windows (by replacing `SDL.dll` with an older one) and it works as intended

smcv:
I'm not a pygame maintainer or contributor, but this looks good to me.

MyreMylar (approved):
OK, after reading the discussion and on a second look though this all seems reasonable. I'm approving. 👍

Starbuck5 (approved):
Looks good to me. Thanks for handling this Ankith, and thanks for the report smcv!

illume (approved):
👍

Issue #3430, opened 2022-08-28
SDL 2.0.22, SDL_mixer 2.6.2, SDL_ttf 2.20.1, SDL_image 2.0.5


illume:
The SDL libs need to be updated to latest versions.

- [ ] windows
- [ ] linux
- [ ] mac (also with manylinux scripts)

Ok, so it seems windows has these versions:
- SDL2-devel-2.0.22-VC.zip (old ahead of manylinux)
- SDL2_image-devel-2.0.5-VC.zip (old same as manylinux)
- SDL2_ttf-devel-2.20.0-VC.zip (latest major but there is a minor update, ahead of manylinux)
- SDL2_mixer-devel-2.6.0-VC.zip (latest major but there is a minor update, ahead of manylinux)

manylinux:
- SDL2-2.0.20.tar.gz
- SDL2_image-2.0.5.tar.gz
- SDL2_ttf-2.0.18.tar.gz
- SDL2_mixer-2.0.4.tar.gz

ankith26:
I will try to take this on in the coming week

ankith26:
I just tested an update to SDL 2.24.0 and there is still a huge performance regression that I had also observed when I tried to upgrade to SDL 2.0.22 (SDL 2.0.20 works fine)

I use `aliens` example for testing and use Ubuntu 22.04 running under WSL (I mention this because I'm still not sure whether it's a general linux issue, or is specific to my platform)

Some other details that I think are useful for debugging
- `pygame.display.get_driver()` returns `x11` for me, which means `xwayland` is being used under the hood
- Display info gives (this looks aight to me)
```py
>>> pygame.display.Info()
<VideoInfo(hw = 0, wm = 1,video_mem = 0
         blit_hw = 0, blit_hw_CC = 0, blit_hw_A = 0,
         blit_sw = 0, blit_sw_CC = 0, blit_sw_A = 0,
         bitsize  = 32, bytesize = 4,
         masks =  (0x00ff0000, 0x0000ff00, 0x000000ff, 0x00000000),
         shifts = (16, 8, 0, 0),
         losses =  (0, 0, 0, 8),
         current_w = 300, current_h = 300
         pixel_format = SDL_PIXELFORMAT_RGB888)
```
For now I'll be testing whether I can somehow compile wayland support in in our manylinux base images, and see if that works as a fix for this issue as well. Later I may make an SDL issue on this once I can collect some more info

Starbuck5:
[...]

PR #3287, opened 2022-07-09
Update SDL_ttf, SDL_mixer on Windows

Starbuck5:
SDL_ttf 2.20.0 was simple, it replaces our custom prebuilt made by me to avoid the segfault problem with SDL_ttf 2.0.19 and random fonts.

SDL_mixer 2.6.0 gave me some grief until I discovered our version `#if` statements had failed on the versioning change, so I had to update the C code for that.

PR #3433, opened 2022-08-29
Update and sync `SDL_mixer` and `SDL_ttf` to 2.6.2 and 2.20.1 respectively

ankith26:
for #3430

In this PR I'm taking on only two of the four things we have to update (to make reviewing easier)

The `SDL_ttf` bump is pretty straightforward on both windows and mac/manylinux, the bulk of this PR is for the `SDL_mixer`

I fixed the windows buildconfig to accomodate for the fact that `SDL_mixer` 2.6+ has lesser DLLs, and some of the DLLs have also moved into a nested `optional` directory. This means that #3287 introduces a regression in music loading for some music formats like opus and the many formats handled by modplug (which this PR fixes)

On the mac/manylinux front: 
`libflac`, `mpg123` and `ogg/vorbis` are not used directly by `SDL_mixer` by default anymore. `SDL_mixer` prefers vendored copies of alternate implementations.
I also noticed while testing that trying to load `opus` music fails on linux (but works on windows), and I realised this is because we missed a dependency `opusfile` even though we have `libopus` around. `opusfile` depends on `libopus` and also `libogg` for loading music from files.

### Testing

I tested this PR on Windows and Ubuntu (WSL), I used the `playmus` example to test start/stop/pause/rewind etc on as many music formats as I could easily find

- flac
- mp3
- ogg
- wav
- mid (midi file format, handled by fluidsynth, timidity or SDL "nativemidi")
- opus
- xm (one of the many formats handled by modplug)

On windows all of these work. On linux I got all of these except `mid` working out of the box. I later found out that midi needs more config from the user side, as the SDL_mixer docs mention:
> If you have built with FluidSynth support, you'll need to set the SDL_SOUNDFONTS environment variable to a Sound Font 2 (.sf2) file containing the musical instruments you want to use for MIDI playback. (On some Linux distributions you can install the fluid-soundfont-gm package)
> To play MIDI files using Timidity, you'll need to get a complete set of GUS patches from: [timidity.tar.gz](https://www.libsdl.org/projects/mixer/timidity/timidity.tar.gz) and unpack them in /usr/local/lib under UNIX, and C:\ under Win32.

Which I suppose is up to the user to choose which of these they want to use at runtime. This may need documenting and more testing, but is out of scope of this PR

I've also tested a few font examples and applications, and I did not see any regressions due to the `SDL_ttf` update.

I've not tested this PR on macs because of not having hardware, but I would appreciate it if this PR is reviewed and tested by a mac user too

### To discuss

The latest commit in this PR is put up for discussion, and can be reverted without breaking anything, if needed

- [x] Should we still be compiling `libflac`, `mpg123` and `vorbis`? **yes**
- [x] If yes, should we also use that backend instead of the `SDL_mixer` vendored alternate implementation? **yes**


illume:
> Should we still be compiling libflac, mpg123 and vorbis?

The smallest change is to leave them in there for now. I think once we have some more experience with them keeping their vendored stuff up to date we can switch to the vendored libraries.

Can you remove that commit? 

(I reran the failing ubuntu job because it had unrelated apt-get issues.)

illume (approved):
🎉 thanks for doing this.

ankith26:
> The smallest change is to leave them in there for now. I think once we have some more experience with them keeping their vendored stuff up to date we can switch to the vendored libraries.

I'm also fine with continuing to stick with the existing ones (considering libflac and libmpg123 are still being updated and maintained rn), but there's also the thing that these alternate libraries will be used on windows and this may be problematic for inter-platform consistency later. Music related bugs on windows would have a higher chance of not being reproducible on macs/linux, and vice-versa.

There's also an issue of the increasing wheelsize on manylinux due to all the recent updates, and doing this may slightly increase wheelsize (just a hunch, untested). However I've got another WIP branch with some promising results, if that works then the current increase would be quite insignificant even if it is a thing.

I will move the latest commit into a separate branch for archiving, and change the `configure` flags for `SDL_mixer` if the minor inter-platform inconsistency is okay.

illume:
> I'm also fine with continuing to stick with the existing ones (considering libflac and libmpg123 are still being updated and maintained rn), but there's also the thing that these alternate libraries will be used on windows and this may be problematic for inter-platform consistency later. Music related bugs on windows would have a higher chance of not being reproducible on macs/linux, and vice-versa.

Good point. I wonder about cross version compatibility as well.

>I will move the latest commit into a separate branch for archiving, and change the configure flags for SDL_mixer if the minor inter-platform inconsistency is okay.

Ok, let's look into this later.

illume (approved):
🎉 thanks :)

PR #3439, opened 2022-09-10
pygame wasm test ci optimization

pmp-p:
that affects only CI. not manual builds
should avoid sdl2 local port rebuilding ( which is already done in python-wasm-sdk )
it also avoids having a random port version being rebuilt and used under the hood ( CI logs are not that easy to read )


illume (approved):
Nice speed up. From about 2mins to about 53secs.

.github/workflows/build-emsdk.yml
```diff
@@ -67,13 +68,17 @@ jobs:
     - name: Install python-wasm-sdk
       run: |
         sudo apt-get install lz4
-        mkdir python-wasm-sdk && cd python-wasm-sdk
-        curl -sL --retry 5 https://github.com/pygame-web/python-wasm-sdk/releases/download/$SDK_VERSION/$SDK_ARCHIVE | tar xvP --use-compress-program=lz4
+        echo https://github.com/pygame-web/python-wasm-sdk/releases/download/$SDK_VERSION/$SDK_ARCHIVE
+        #mkdir python-wasm-sdk && cd python-wasm-sdk
+        # ??? curl -sL --retry 5 https://github.com/pygame-web/python-wasm-sdk/releases/download/$SDK_VERSION/$SDK_ARCHIVE | tar xvP --use-compress-program=lz4
+        wget -O- https://github.com/pygame-web/python-wasm-sdk/releases/download/3.1.19.0/python3.11-wasm-sdk-Ubuntu-22.04.tar.lz4 | tar xvP --use-compress-program=lz4
```
ankith26, line 74: I don't get why this change is being made. Hardcoding the URL would mean that the previously set enviroment variables are redundant.

I'd prefer if this entire diff is reverted (except for the rm SDL1 part)

buildconfig/config_emsdk.py
```diff
@@ -7,85 +7,115 @@
...
 from distutils.sysconfig import get_python_inc
 
-configcommand = os.environ.get('SDL_CONFIG', 'sdl-config',)
-configcommand = configcommand + ' --version --cflags --libs'
-localbase = os.environ.get('LOCALBASE', '')
-if os.environ.get('PYGAME_EXTRA_BASE'):
-    extrabases = os.environ['PYGAME_EXTRA_BASE'].split(':')
+configcommand = os.environ.get(
+    "SDL_CONFIG",
+    "sdl-config",
```
ankith26, line 12: Shouldn't this be `sdl2-config`? That is the case on linux atleast and I'm not sure whether it's the the same on emscripten

buildconfig/config_emsdk.py
```diff
@@ -224,6 +266,8 @@ def main(auto_config=False):
...
 
 
-if __name__ == '__main__':
-    print("""This is the configuration subscript for Unix.
-Please run "config.py" for full configuration.""")
+if __name__ == "__main__":
+    print(
+        """This is the configuration subscript for Unix.
+Please run "config.py" for full configuration."""
+    )
```
ankith26, line 273: to simplify diffs on this file, I'd prefer if black formatting is not applied to this file for now

ankith26 (changes requested):
Thanks for the PR. I left a few review comments

I'd also love if this PR was rebased to have only the needed/cleaned commits

.github/workflows/build-emsdk.yml
pmp-p: i saw a string bug in CI execution, and i thought curl line was problematic, when i switched to wget it worked and i just forgot about it. That's why reviewing is so important :)

buildconfig/config_emsdk.py
pmp-p: indeed though it probably does not have any effect. that config file could use a big cleanup i copied too much things from unix cfg.

buildconfig/config_emsdk.py
pmp-p: changed the whole message.

MyreMylar (approved):
LGTM. Slight shame about the merging of a formatting pass with some genuine changes, makign it harder to review but eh, we've all done it. 

I second the appeal from @ankith26 for a rebase if you can figure out how to do it, as this has ended up a long string of small commits, but if it is too much of a pain we can roll it in.

pmp-p:
cleaned up PR is here https://github.com/pygame/pygame/pull/3592

@MyreMylar for some reason it cannot build for me without adding `#pragma once` to https://github.com/pygame/pygame/blob/main/src_c/_blit_info.h

pmp-p:
cleaned up PR is here https://github.com/pygame/pygame/pull/3592

@MyreMylar for some reason it cannot build for me without adding `#pragma once` to https://github.com/pygame/pygame/blob/main/src_c/_blit_info.h

pmp-p:
merged in https://github.com/pygame/pygame/pull/3593

PR #3440, opened 2022-09-10
Delete rect subscripts should error not segfault

### LINKED ISSUE #3438: Segfault when trying to delete item in pygame.Rect instance
Sorry if there's already an existing issue for this problem. I searched for it but couldn't find any.

As always, thank you for your awesome selfless work.

**Environment:**

- **Operating system**: Linux (xubuntu 22.04 64bits)
- **Python version**: 3.10.4
- **SDL version**: 2.0.20
- **PyGame version**: 2.1.2

**Steps to reproduce:**

`python3 -c "from pygame import Rect; r = Rect(1, 2, 3, 4); del r[0]`

**Current behavior:**
A segmentation fault.

**Expected behavior:**
I know deleting a rect item is not a natural thing to do, but it should raise an exception nonetheless, not cause a segfault.

This is the output:

```
Fatal Python error: pygame_parachute: (pygame parachute) Segmentation Fault

Current thread 0x00007fc4a2c64000 (most recent call first):
  File "<string>", line 1 in <module>

Extension modules: pygame.base, pygame.constants, pygame.rect, pygame.rwobject, pygame.surflock, pygame.color, pygame.bufferproxy, pygame.math, pygame.surface, pygame.display, pygame.draw, pygame.event, pygame.imageext, pygame.image, pygame.joystick, pygame.key, pygame.mouse, pygame.time, pygame.mask, pygame.pixelcopy, pygame.transform, pygame.font, pygame.mixer_music, pygame.mixer, pygame.scrap, numpy.core._multiarray_umath, numpy.core._multiarray_tests, numpy.linalg._umath_linalg, numpy.fft._pocketfft_internal, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._mt19937, numpy.random.mtrand, numpy.random._philox, numpy.random._pcg64, numpy.random._sfc64, numpy.random._generator, pygame._freetype (total: 39)
Aborted (core dumped)
```

ankith26:
Simple PR to fix #3438

### DIFF
--- a/src_c/rect.c
+++ b/src_c/rect.c
@@ -1330,6 +1330,11 @@ pg_rect_ass_item(pgRectObject *self, Py_ssize_t i, PyObject *v)
     int val = 0;
     int *data = (int *)&self->r;
 
+    if (!v) {
+        PyErr_SetString(PyExc_TypeError, "item deletion is not supported");
+        return -1;
+    }
+
     if (i < 0 || i > 3) {
         if (i > -5 && i < 0) {
             i += 4;
@@ -1407,6 +1412,11 @@ pg_rect_subscript(pgRectObject *self, PyObject *op)
 static int
 pg_rect_ass_subscript(pgRectObject *self, PyObject *op, PyObject *value)
 {
+    if (!value) {
+        PyErr_SetString(PyExc_TypeError, "item deletion is not supported");
+        return -1;
+    }
+
     if (PyIndex_Check(op)) {
         PyObject *index;
         Py_ssize_t i;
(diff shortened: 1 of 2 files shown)


Starbuck5 (approved):
Looks good to me.

I see that this is almost exactly how Vectors do it, so this code isn't some crazy new thing. 😄

Starbuck5:
I see some other places use `AttributeError` instead of type error, what do you think? Like the other deletion code in Rect, and the deletion code in Color.

ankith26:
I'm not very happy with `TypeError` either, but it seems like the best option

`AttributeError` is definitely wrong, it can be used for something like `del rect_obj.topleft`, but not while subscripting

illume (approved):
🎉 looks great to me too.

Issue #3406, opened 2022-08-20
Example boilerplate in the docs


Starbuck5:
New users often struggle with the basic structure of a pygame program, and it's understandable. There is boilerplate, it's easy to mess up, and it's not provided.

This is my personal base script:
```py
import pygame

pygame.init()

screen = pygame.display.set_mode((1280, 720))

clock = pygame.time.Clock()

while True:
    screen.fill("purple")

    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit

    pygame.display.flip()
    
    clock.tick(144)
```

Perhaps something like this should be included for convenience in the documentation. I'm not sure where it should live. Figuring out a thoughtful and handy way to include this is the hardest part of this issue.

DatDerpyDude81:
You should change the `clock.tick` to 60, more common in my opinion. Also maybe add a sprite or something if it doesn't get too complicated

DatDerpyDude81:
I have put a `boilerplate.py` in the examples folder, made a pr for it :)

PR #3441, opened 2022-09-11
added boilerplate.py

DatDerpyDude81:
Credit to @Starbuck5 for the idea.

### DIFF
--- a/examples/boilerplate.py
+++ b/examples/boilerplate.py
@@ -1,24 +1,24 @@
-import pygame
+import pygame as pg
 
-pygame.init()
+pg.init()
 # define variables
-screen = pygame.display.set_mode((1280, 720))
+screen = pg.display.set_mode((1280, 720))
 
-clock = pygame.time.Clock()
+clock = pg.time.Clock()
 
 
 # main loop
 while True:
     # fill the screen with the color purple
     screen.fill("purple")
-# get pygame events
-    for event in pygame.event.get():
+    # get pygame events
+    for event in pg.event.get():
         # if pygame is quitted, end the program
-        if event.type == pygame.QUIT:
-            pygame.quit()
+        if event.type == pg.QUIT:
+            pg.quit()
             raise SystemExit
-# update the screen
-    pygame.display.flip()
+    # update the screen
+    pg.display.flip()
 
-# make the game go at 60 fps
+    # make the game go at 60 fps
     clock.tick(60)
--- a/examples/boilerplate.py
+++ b/examples/boilerplate.py
@@ -1,0 +1,24 @@
+import pygame
+
+pygame.init()
+# define variables
+screen = pygame.display.set_mode((1280, 720))
+
+clock = pygame.time.Clock()
+
+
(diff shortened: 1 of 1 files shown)


illume:
Against, because we have a full folder of examples showing this isn't the only or good way of doing things.

examples/boilerplate.py
```diff
@@ -0,0 +1,24 @@
...
+screen = pygame.display.set_mode((1280, 720))
+
+clock = pygame.time.Clock()
+
+
+# main loop
+while True:
+    # fill the screen with the color purple
+    screen.fill("purple")
+# get pygame events
```
novialriptide, line 14: Comments should be on the same indent level as the while loop

examples/boilerplate.py
```diff
@@ -0,0 +1,24 @@
+import pygame
```
novialriptide, line 1: You should be importing `pygame` as `pg` here to keep consistency with the other examples.

novialriptide (changes requested):
```python
if __name__ == "__main__":
    main()
```

You should put this at the bottom and move everything in the while loop into a `main()` function as shown here: https://github.com/pygame/pygame/blob/main/examples/sound.py

Starbuck5:
Issue link: https://github.com/pygame/pygame/issues/3406

Thanks for making a PR.

This isn't exactly what I had in mind. I've seen in other libraries documentation "quickstart" sections that just get people what they need to get started, and that inspired me to think about the bare minimum people need to get started with pygame. One thing that is non obvious is how to set up a game loop. The challenge is how to incorporate this information into the documentation so it's consistent and easily discoverable by new users.

In my opinion, merely putting into `examples/` won't be helpful, because newbies (and the vast majority of pygame users) wouldn't think to look for it there.

DatDerpyDude81:
I guess we could mention it in the README, 

> Issue link: #3406
> 
> Thanks for making a PR.
> 
> This isn't exactly what I had in mind. I've seen in other libraries documentation "quickstart" sections that just get people what they need to get started, and that inspired me to think about the bare minimum people need to get started with pygame. One thing that is non obvious is how to set up a game loop. The challenge is how to incorporate this information into the documentation so it's consistent and easily discoverable by new users.
> 
> In my opinion, merely putting into `examples/` won't be helpful, because newbies (and the vast majority of pygame users) wouldn't think to look for it there.

I guess we could the file in the README, or we could put this in the pygame docs website

novialriptide (changes requested):

MyreMylar:
What about adding a short 'Getting Started.rst' file under https://github.com/pygame/pygame/tree/main/docs somewhere that is then directly linked to from the main readme on github, and maybe included in the docs website as well?

.rst files work pretty well when linked to from github and can include code chunks. Could have the canonical basic 'open an empty window that won't crash' game loop in it - with a little explanation of each line and why it is needed. I'm actually surprised this doesn't exist somewhere already. Maybe it does and we just need to link it up better from the main readme?

PR #3445, opened 2022-09-13
Compile freetype with harfbuzz, reduce manylinux wheelsize

ankith26:
Kind of a follow up on #3430, when I was updating `SDL_ttf` in there I realised that `SDL_ttf` statically linked against a vendored freetype+harfbuzz, which meant that our wheels are actually including freetype twice. 

Without this PR, a 2.1.3 manylinux wheel will easily shoot past the 30 MB mark. And the biggest culprit is harfbuzz. I've noticed that an uncompressed and unstripped harfbuzz shared library is almost 31 MB, and this is almost half of the entire pygame install by itself. Since harfbuzz is such a large dependency, I thought that my wheelsize reducing patch should go along in this PR, even though it's technically an independent change. The rest of the bundled `so` files in the manylinux wheel together constitute a large fraction of the total wheel size. My "fix" for this is basically running `strip` on all these so files (by first consolidating all of these in `/usr/local/lib`). This reduces the wheelsize by about 60% in my initial draft (from 31.8 MB to 13 MB)

And ofc, the added benefit of this PR is that `pygame.freetype` gets harfbuzz support. Nothing much should change on the `pygame.font` end because here harfbuzz support was already introduced along with the SDL_ttf 2.0.18 update

This PR is a draft while I'm doing more testing, not a 2.1.3 blocker


ankith26:
Okay I've done some testing, and this PR should be done from my end

illume (approved):
thank-you-very-cool! 🎉

PR #3449, opened 2022-09-14
Bump SDL_image to 2.6.2

ankith26:
For #3430

In this PR I'm also replacing our vendored PNG saving function with the SDL_image equivalent. Also removed `jpeg` and `png` as direct pygame dependencies (they are still indirect dependencies)

### NOTE: This PR is blocked

There are new failing tests on pngs with palettes which are a result of `SDL_image` preferring `stb_image` (relevant issue on the SDL side: https://github.com/libsdl-org/SDL_image/issues/298). Since we compile SDL_image ourselves on mac/manylinux and can disable the usage of `stb_image`, this is a windows only regression.

I can add test skips if we need this PR in time for 2.1.3, but I think this is a non-trivial regression so this PR is not suitable for a release yet. This PR will only be ready if there's a fix on the SDL_image side for this, or if we start compiling dependencies on windows ourselves.

### Testing

Apart from the above blocker, I tested other image related things and they all seem to work. Image saving (bmp, tga, jpeg, png) work as before, and all our unittests for those are still passing.

I wrote a script to test that we can load all the images that we document supporting, and it works fine.

I first downloaded [a free image from pixabay](https://pixabay.com/illustrations/abstract-art-texture-mandala-1248169/) and downscaled it to 200x200. Then I used many online converters to convert this jpeg into all documented formats (except SVG for obvious reasons, and LBM because I could not easily find an online image converter for this).

Those images are consolidated in this [zip file](https://github.com/pygame/pygame/files/9567679/images.zip).

<details>
  <summary>My test script (which needs the attached images.zip)</summary>

```py
import zipfile

import pygame

IMAGE_SIZE = 200
SCREEN_RATIO = (5, 3)

pygame.init()

screen = pygame.display.set_mode([i * IMAGE_SIZE for i in SCREEN_RATIO])

image_zip = zipfile.Path("images.zip")
loaded_images = [
    (file.name, pygame.image.load(file.open("rb"), file.name))
    for file in image_zip.iterdir()
]

font = pygame.font.Font(None, 24)

all_looped = False
count = 0
for i in range(SCREEN_RATIO[0]):
    for j in range(SCREEN_RATIO[1]):
        if count >= len(loaded_images):
            print("INFO: Ran out of new images to use, re-using older images")
            count = 0

        name, surf = loaded_images[count]
        pos = pygame.Vector2(i, j) * IMAGE_SIZE
        screen.blit(surf, pos)
        count += 1

        ext = name.split(".")[-1].upper()
        if ext == "PBM":
            info = "Monochrome"
        elif ext == "PGM":
            info = "Greyscale"
        else:
            info = "Full Colour"

        screen.blit(font.render(f"{ext} ({info})", True, "black", "green"), pos)
        if count == (len(loaded_images) - 1):
            all_looped = True


if not all_looped:
    print("WARNING: Did not draw all loaded images!")
    print(" ".join(i[0] for i in loaded_images[count:]))

pygame.display.update()
clock = pygame.time.Clock()
while True:
    clock.tick(60)
        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit()
```

</details>

And this gives 
image


illume:
Because the SDL_image is out in the wild, we'll have to support this version with the different configs for png. There's no `@potentialExpectedFailureIf`, and skip isn't a good idea because we still want to run the test... so maybe a try/catch in there with a skip if SDL_image is the 2.6.2.

There's plenty of other improvements in this SDL image on windows. Hrmm. But are they worth it? Do we have anything blocking users, or some massive benefit for the new SDL_image?

I don't know the reasoning behind choosing the stb image libraries; considering they are slower, less featureful and buggier. I expect there are other regressions we are not testing for.

I'd prefer if we kept libjpeg(turbo), libpng, etc and disable stbimage in our builds.

What are our options?
- hold SDL_image back on all platforms? Hrmm.
- keep with old SDL_image just on windows
- compile our own
- use another packaged one
- use new one, and ship on windows with this regression, add something like a skip/expectedFailure

I guess eventually we will have to compile our own. I think for now, and 2.1.3... I think holding it on windows is the way to go.

Let's hold off on SDL_image for 2.1.3?

ankith26:
I agree with holding this entire PR for future.

I think the regressions outweigh any new fixes on windows so releasing with libsdl.org 2.6.2 prebuilts is not an option I think.

2.6.2 does have a lot of cool new stuff like qoi/avif/jxl support (of which qoi is built-in), scaled svg loading and animated gif loading; the latter two needing new API on pygame end (so those are for post 2.1.3 anyways).

The biggest bugfix I see listed in the release notes is "Fixed XCF regression introduced in 2.0.5". But `pygame.image` docs never documented supporting this format, so I think this fix is not critical for us and it is also not a widely used format.

illume:
> I agree with holding this entire PR for future.

ok, cool

PR #3453, opened 2022-09-16
Freshen up arraydemo.py

Starbuck5:
Just a few minor things I noticed

- `import numpy as N` -> `import numpy as np` (it's what the people do!)
- removed some unused code
- removed a call to a deprecated pygame function

### DIFF
--- a/examples/arraydemo.py
+++ b/examples/arraydemo.py
@@ -34,16 +34,6 @@ def surfdemo_show(array_img, name):
         if e.type == pg.MOUSEBUTTONUP and e.button == pg.BUTTON_LEFT:
             break
         elif e.type == pg.KEYDOWN and e.key == pg.K_s:
-            # pg.image.save(screen, name+'.bmp')
-            # s = pg.Surface(screen.get_size(), 0, 32)
-            # s = s.convert_alpha()
-            # s.fill((0,0,0,255))
-            # s.blit(screen, (0,0))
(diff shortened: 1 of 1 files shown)


examples/arraydemo.py
```diff
@@ -34,16 +34,6 @@ def surfdemo_show(array_img, name):
...
-            # s = pg.Surface(screen.get_size(), 0, 32)
-            # s = s.convert_alpha()
-            # s.fill((0,0,0,255))
-            # s.blit(screen, (0,0))
-            # s.fill((222,0,0,50), (0,0,40,40))
-            # pg.image.save_extended(s, name+'.png')
-            # pg.image.save(s, name+'.png')
-            # pg.image.save(screen, name+'_screen.png')
-            # pg.image.save(s, name+'.tga')
             pg.image.save(screen, name + ".png")
```
ankith26, line 37: This could probably be `f"{name}.png"` as that looks more readable to me... but oh well I think there are many places where this kind of pattern is used, and it's not a super big difference.
This does not have to be changed in this PR, I just happened to notice this here

ankith26 (approved):
LGTM, thanks 🍰

illume (approved):
thanks! 😄

examples/arraydemo.py
Starbuck5: You're right, but I'll leave it as it is for now.

PR #3455, opened 2022-09-16
Refactor `examples/mask.py`

pycoinfu:
I made `pygame.examples.mask` use `pygame.Vector2`s, follow Python conventions (currently it's using camelCase for object naming) and simplified the code by removing some of the functions that weren't necessary (`Sprite.setPos`, `Sprite.setVelocity`, etc.)
After asking some people on Discord, I also removed the Python/C performance comparison, because it was calling the exact same method, just the python one was wrapped in another function

### DIFF
--- a/examples/mask.py
+++ b/examples/mask.py
@@ -37,16 +37,16 @@ This module can also be run as a stand-alone program, excepting
 one or more image file names as command line arguments.
 """
 
-import sys
 import os
 import random
+import sys
 import timeit
 
 import pygame as pg
 
 
 class Sprite:
-    def __init__(self, pos, vel, surface: pg.Surface, mask: pg.mask.Mask=None):
+    def __init__(self, pos, vel, surface: pg.Surface, mask: pg.mask.Mask = None):
         self.surface = surface
         self.width, self.height = self.surface.get_size()
         if mask is not None:
@@ -76,7 +76,7 @@ class Sprite:
         ny = self.mask.overlap_area(
             sprite.mask, (offset[0], offset[1] + 1)
         ) - self.mask.overlap_area(sprite.mask, (offset[0], offset[1] - 1))
-        
+
         if nx == 0 and ny == 0:
             # One sprite is inside another
             return
@@ -119,7 +119,7 @@ def main(*args):
     screen_width, screen_height = 640, 480
     screen = pg.display.set_mode((screen_width, screen_height))
     clock = pg.time.Clock()
-    
+
     images = []
     masks = []
     for image_path in args:
@@ -135,7 +135,7 @@ def main(*args):
                 random.uniform(0, screen_width),
             ),
             vel=(
-                random.uniform(-5, 5), 
+                random.uniform(-5, 5),
                 random.uniform(-5, 5),
             ),
             surface=images[j],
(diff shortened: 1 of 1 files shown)


illume:
thanks-very-cool

2x combo points if you could run the python black code formatter on it? 
8x combo points if pylint passes too :)

pycoinfu:
> thanks-very-cool
> 
> 2x combo points if you could run the python black code formatter on it? 8x combo points if pylint passes too :)

Thanks for checking out my PR! 😄 
I added missing docstrings and changed the variable names that pylint was complaining about
The code was already `black`-ified, is there some setting I should use?

examples/mask.py
```diff
@@ -204,8 +195,8 @@ def main(*args):
         print("Let many copies of IMAGE(s) bounce against each other")
         print("Press any key to quit")
         main_dir = os.path.split(os.path.abspath(__file__))[0]
-        imagename = os.path.join(main_dir, "data", "alien1.png")
-        main(imagename)
+        image_name = os.path.join(main_dir, "data", "alien1.png")
```
novialriptide, line 198: Should this be condensed into one line?

examples/mask.py
```diff
@@ -132,70 +128,65 @@ def main(*args):
...
-                    s.pos[1] = screen.get_height()
-                elif s.pos[1] > screen.get_height() + 3:
-                    s.pos[1] = -s.surface.get_height()
-                screen.blit(s.surface, s.pos)
-            pg.display.update()
-        elif event.type == pg.KEYDOWN:
-            return
+        for event in pg.event.get():
+            if event.type in (pg.QUIT, pg.KEYDOWN):
+                return
```
novialriptide, line 163: Other pygame examples use `break` instead of `return`, this should be changed here to stay consistent.

pycoinfu: That would just break out of the event loop, and using `pygame.event.wait` in this case isn't possible because that would cause the sprites to stop moving

illume (approved):
🎉 🎈

examples/mask.py
```diff
@@ -36,92 +37,87 @@
...
-    return x[0] * y[0] + x[1] * y[1]
-
-
 class Sprite:
-    def __init__(self, surface, mask=None):
+    """
+    Moving Sprite demonstrating pixel-perfect collisions between pg.mask.Mask objects
+    """
+
+    def __init__(self, pos, vel, surface: pg.Surface, mask: pg.mask.Mask = None):
```
ankith26, line 52: I think this is inconsistent typehinting because some args are typed and some are not, it should be all or none IMHO (I'm personally leaning on the latter for now because most of our other examples don't have type hints, but I'll let you take the call)

If you do decide to go with the typehinting, there is a small issue here: The `mask` argument should be typed like `Optional[pg.mask.Mask]`

ankith26 (changes requested):

examples/mask.py
pycoinfu: Thanks for checking out my PR! 🥳 
Decided to remove the typehints

ankith26 (approved):
LGTM! Thanks for the PR disa 🚀 🍕

novialriptide:
🥳

pycoinfu:
Thanks everyone! 💜

Issue #2529, opened 2021-03-18
No native wayland support in manylinux wheels


robertpfeiffer:
If I install pygame via pip and set the driver with `export SDL_VIDEODRIVER=wayland`, this happens:

```python
>>> pygame.display.set_mode((640,480))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
pygame.error: wayland not available
```

It works with SDL and pygame compiled from source, or with the manylinux wheels through XWayland.

robertpfeiffer:
There is no ~~easy~~ obviously correct way to do this. This touches manylinux (specified by Python devs), debian release schedule, SDL2, KDE, and Gnome. This is way more complicated than you'd think @Starbuck5.

PR #3458, opened 2022-09-18
buildconfig: Bump SDL version to 2.24.0 on windows

illume:
For https://github.com/pygame/pygame/issues/3430

### DIFF
--- a/buildconfig/download_win_prebuilt.py
+++ b/buildconfig/download_win_prebuilt.py
@@ -78,8 +78,8 @@ def get_urls(x86=True, x64=True):
     url_sha1 = []
     url_sha1.extend([
         [
-        'https://www.libsdl.org/release/SDL2-devel-2.24.0-VC.zip',
-        '8a54459189e88c30ba024ee5f18ce4b1a5d1d9e7',
+        'https://www.libsdl.org/release/SDL2-devel-2.24.2-VC.zip',
+        '19b1e9f5bcf700a402bb924c0194c96596283c32',
         ],
         [
         'https://www.libsdl.org/projects/SDL_image/release/SDL2_image-devel-2.0.5-VC.zip',
@@ -234,12 +234,12 @@ def place_downloaded_prebuilts(temp_dir, move_to_dir, x86=True, x64=True):
         copy(
             os.path.join(
                 temp_dir,
-                'SDL2-devel-2.24.0-VC/SDL2-2.24.0'
+                'SDL2-devel-2.24.2-VC/SDL2-2.24.2'
             ),
             os.path.join(
                 move_to_dir,
                 prebuilt_dir,
-                'SDL2-2.24.0'
+                'SDL2-2.24.2'
             )
         )
 
--- a/buildconfig/download_win_prebuilt.py
+++ b/buildconfig/download_win_prebuilt.py
@@ -78,8 +78,8 @@ def get_urls(x86=True, x64=True):
     url_sha1 = []
     url_sha1.extend([
         [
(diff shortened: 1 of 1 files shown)


ankith26:
I built this branch locally, and I could reproduce the CI segfault only when I set `SDL_VIDEODRIVER="dummy"` (like the CI)

Then I ran the tests with the python `-X dev` option which gave a traceback pointing me to the line in python code where it is happening, and it is this

https://github.com/pygame/pygame/blob/7e6f48306f624debac9cb2731bb9adc5a762a5c6/test/display_test.py#L563-L565

Turns out, the issue is because of the `pygame.SCALED` usage (if I remove it, there is no segfault on that line, but in a later line where again - `SCALED` is used)

illume:
Thanks for tracking down the issue.

**shakes fist at SCALED**

What do you think about 2.1.3 without this SDL 2.24.0? There's undoubtedly bug fixes in it I guess. But we can always try again with 2.1.4... hopefully sooner. If we go with SDL 2.24.0, I'd like to do another dev release with it for at least another week. That's why I'm leaning towards going with what we have now.

ankith26:
I'm fine with postponing this PR to 2.1.4 and not having to do more dev
releases. Also given our manylinux SDL situation (where we are still stuck
on 2.0.20), having windows on 2.0.22 seems fine for this release (too much
divergence is probably not good, but is one version difference acceptable?)

Been experimenting with compiling Wayland support in, but I doubt it could
help much for this specific issue, given sdl still defaults to x11 if it's
available (and xwayland usually is these days, when older xlib one is
missing)

illume:
@ankith26 What's the issue with 2.0.22 on linux/macos? 

The tests seem to pass: PR for 2.0.22 on manylinux... https://github.com/pygame/pygame/pull/3475

ps. oops. I forgot we were still on 2.0.20 on manylinux. I think probably we should go with that on manylinux/macos.

ankith26:
Oh, I had mentioned the issue I faced earlier on [here](https://github.com/pygame/pygame/issues/3430#issuecomment-1229813334)

I didn't see any failing tests locally either, but somehow runtime performance is massively impacted

illume:
> runtime performance is massively impacted

I'm thinking we still do a 2.1.3.dev8 release with 2.0.22 on mac/linux. Probably we can have a note about known issues, and for people to use 2.1.3.dev6 in the meantime.

Then the following:
- [ ] make an SDL performance on WSL reduced issue
- [ ] make wayland issue, and try and get wayland working
- [ ] make a SCALED with 2.24.0 on windows issue

By the way, Wayland has always been at least 50% of the speed for me even without X involved. A lot is to do with them copying the image data twice. The rest is less optimized graphics drivers.

ankith26:
> make wayland issue, and try and get wayland working

There is already an issue open for it here #2529

> I'm thinking we still do a 2.1.3.dev8 release with 2.0.22 on mac/linux. Probably we can have a note about known issues, and for people to use 2.1.3.dev6 in the meantime.

Yeah, this seems like a good idea. I just found out (thanks @avaxar for testing this) that 2.0.22 works aight on a real linux machine (running latest linux mint) and is infact, faster that 2.1.2 by a significiant margin (tested the `testsprite` example)

*screams at WSL* 😩

ankith26:
Hrmmm, this PR fixes one known issue to me already. BTW I have tracked down the test fail from here and made an SDL bug report (this bug is already fixed on `main` and set to make its way to SDL 2.26.0, and if the SDL devs agree this fix may also be backported to a 2.24.2)

PR #3470, opened 2022-09-30
Fix new mypy-0.981 fails

ankith26:
Simple PR to fix the latest fails on the stubtest builds (due to the updated mypy)


buildconfig/stubs/pygame/constants.pyi
```diff
@@ -560,5 +560,3 @@ WINDOWRESTORED: int
 WINDOWSHOWN: int
 WINDOWSIZECHANGED: int
 WINDOWTAKEFOCUS: int
-
-__all__: List[str]
```
MightyJosip, line 564: You have to edit this [file](https://github.com/pygame/pygame/blob/main/buildconfig/stubs/gen_stubs.py) to remove that line

MightyJosip (changes requested):

buildconfig/stubs/pygame/constants.pyi
ankith26: Oh right, good catch. Thanks!

illume (approved):
Thanks.

PR #3471, opened 2022-10-01
2.1.3.dev6 and 2.1.3.dev7

illume:


ankith26:
I'm just curious why glib is being added, since the base images have glib (albeit an old version of it) already... Was there an issue with the older version? (since the one we build ourselves is not super new either)

illume:
> I'm just curious why glib is being added, since the base images have glib (albeit an old version of it) already... Was there an issue with the older version? (since the one we build ourselves is not super new either)

Yes, fluidsynth wasn't building.

PR #3473, opened 2022-10-02
Fix es folder location and fix some sphinx issues in ChimpanceLineaporLinea.rst

### LINKED ISSUE #3472: Spanish translation files are in the wrong location
As pointed out by @avaxar, the Spanish translation files are in the wrong location.
https://github.com/pygame/pygame/tree/main/es

I found that it came from this PR: https://github.com/pygame/pygame/pull/3421/files

illume:
Also, fix some docs/tut/es/ChimpanceLineaporLinea.rst issues sphinx warns about.
 
- So it's listed from the index page. Eventually it should probably have its own index in es/index.
- There was a weird problem with the common.txt putting sphinx into an infinite loop, so had to include that in the es.
- There's no sub folders in es yet, so the include will need to change when sub folders are added.

Fixes #3472


ankith26 (approved):
LGTM, thankies! 🍰 😎

Issue #3479, opened 2022-10-05
Add a sprite.Sprite.handle_event() hook 


python-ast-person:
**Description**

I propose to add a hook to `sprite.Sprite` and `sprite.AbstractGroup` to allow sprites to handle their own events instead of the event handling code be in the main loop.

**Proposed API**
This enhacement would add two methods and one attribute to `sprite.Sprite`:

- `Sprite.handle_event()` ,like `Sprite.update`, is a by defalt empty method designed to be overriden in a subclass. The main

Matiiss:
I fail to see the purpose of this, you can already do something like, also I don't think you should be actually changing the already existing classes, create a new one specifically for this purpose but again, I don't see a purpose here.
```py
class MySprite(pygame.sprite.Sprite):
    def update(self, events):
        for event in events:
            # handle event


some_group = pygame.sprite.Group()
some_group.add(MySprite())

# mainloop
    events = pygame.event.get()
    for event in events:
        # handle event
    
    some_group.update(events)
```

PR #3480, opened 2022-10-06
added the handle_event and listerning_events

python-ast-person:
Pull request implementing #3479 .

### DIFF
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -89,9 +89,10 @@ from warnings import warn
 
 import pygame
 
-from pygame.rect import Rect
-from pygame.time import get_ticks
-from pygame.mask import from_surface
+from pygame.rect   import Rect
+from pygame.time   import get_ticks
+from pygame.mask   import from_surface
+from pygame.event  import NUMEVENTS, EventType
 
 
 class Sprite:
@@ -109,8 +110,11 @@ class Sprite:
 
     """
 
+    listening_events = frozenset()
+
     def __init__(self, *groups):
         self.__g = {}  # The groups the sprite is in
+        self._old_listening_events = frozenset(self.listening_events)
         if groups:
             self.add(*groups)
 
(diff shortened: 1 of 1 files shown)


src_py/sprite.py
```diff
@@ -89,9 +89,10 @@
 
 import pygame
 
-from pygame.rect import Rect
-from pygame.time import get_ticks
-from pygame.mask import from_surface
+from pygame.rect   import Rect
```
novialriptide, line 92: Please remove the extra spaces.

src_py/sprite.py
```diff
@@ -245,6 +249,35 @@ def layer(self, value):
...
+
+        Sprite.handle_event(event):
+
+        The default implementation of this method does nothing; it's just a
+        convenient "hook" that you can override. This method is called by
+        Group.handle_event(event) with the event you called it with only if
+        it's type is in listening_events.
+
+        Note that if you change listening_events, the changes will not be 
+        acknoleged untill you call Sprite.update_listening_events()
```
novialriptide, line 263: There are a couple of spelling mistakes here. Please fix.

novialriptide (changes requested):
I fail to understand what this PR is actually doing, but I left some reviews. I tested this on my local machine and it fails tests.

src_py/sprite.py
```diff
@@ -611,6 +651,43 @@ def __len__(self):
...
+        """Handles multiple events at once
+        
+        AbstractGroup.handle_events(*events)
+        
+        Calls event handlers in the same manor as AbstractGroup.handle_event
+        but work on more than on event at a time.
+        Each argument should be eather an event or an iterable yields events.
+        """
+        for event in events:
+            if isinstant(event,EventType):
```
novialriptide, line 687: This isn't a valid function.

src_py/sprite.py
```diff
@@ -245,6 +249,35 @@ def layer(self, value):
...
+        method by the same name in the Group class.
+        """
+
+    def update_listening_events(self):
+        for event_type in self.listening_events-self._old_listening_event:
+            #The previous line only interates over new event types
+            for group in self.__g:
+                group._register_event(event_type,self)
+        for event_type in self._old_listening_event-self.listening_events:
+            #The previous line only interates over old event types
```
novialriptide, line 275: Btw, the comments should have a space between the hastag and the actual message content

novialriptide (changes requested):

MyreMylar:
Sorry, but I think it is a very bad idea for us to change the interface of a long-lived python class that we have been encouraging others to subclass from for years, in fact it is specifically built to be sub-classed. 

Why? Because what this will do is cause a method name clash with anyone who has sub-classed Sprite with a function called `handle_event()` that doesn't have the exact same parameters as you have used.

The way to do this with our existing implementations would be yet another new subclass of sprite (like `DirtySprite`). Though I'm not sure it is a great idea for us to just keep adding new sprite subclasses with different features either, I think we arguably already have so many sprite groups that it is confusing.

I think it is fine fore people to handle events for sprites in this way (I do it in a similar way myself) but a tutorial might be the easiest way forward with this idea.

PR #3482, opened 2022-10-07
Add wayland and kmsdrm support to manylinux wheels

### LINKED ISSUE #2529: No native wayland support in manylinux wheels

ankith26:
PR to fix #2529

So this PR just basically adds wayland (with libdecor builds) and mesa to our manylinux base images, so SDL finds this at compile time and at runtime pygame would have support for kmsdrm and wayland videodrivers. Just like X11, drm/gbm and wayland shared libraries are not actually included in the manylinux wheels. SDL loads whatever is available from the users system (X11 is still the default for SDL).

Is this change going to impact the average pygame linux user for now? Probably not. xwayland is generally available when xlib is not.

I've only tested this PR on manylinux2014_x86_64, this PR definitely needs a wider range of testing, so I think it can be postponed to an early dev release of 2.1.4 (and get better public testing that way)


MyreMylar:
> Is this change going to impact the average pygame linux user for now? Probably not. xwayland is generally available when xlib is not.

Now I know who this PR is not for, but who is it for? :)

I don't have a wayland capable system so am not much use for testing but the build scripts overall seem fine. I notice it was all being compiled with python 3.10 - should we upgrade this to python 3.11 now before merging?

ankith26:
The plan is to be more future ready, SDL3 will probably start using wayland by default and it's not that far away. It would be good if we can provide optional wayland support right away in our wheels for wider testing (by people well versed in this stuff, obviously we can't expect everyone to do this which is why I said this change doesn't impact the average pygame user) and gradual adoption (without changing default behaviour)

MyreMylar (approved):
Happy to approve this for wider public testing, it seems to pass the test runners and none of our active reviewers have the hardware to test it.

PR #3484, opened 2022-10-08
Python 3.11 on mac/manylinux (with updated cibuildwheel)

ankith26:
Updates cibuildwheel so that we can release the next dev release with 3.11 (rc2) support on mac/manylinux. This is a release candidate, but since it is intended to be ABI compatible with the final release, python encourages library devs to release with these.

~~Also removes manylinux1_i686 and manylinux2010_i686. The rationale is that manylinux1 and manylinux2010 wheels are needed by (probably) less than 5% of our users, and 32bit linux looks like it is almost non-existent these days. Besides, even if we have a user on a 32bit linux and they come with a bug report, we can always recommend them to update pip so that it picks up the manylinux2014_i686 wheel (which we should still keep around for some time IMHO)~~

~~Also does some cleanups and bugfixes (the pull targets were fixed, and manylinux1 was uncommented out) to some manylinux buildconfig code~~
EDIT: the latter have been deferred to future PRs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,6 @@ jobs:
     environment:
       # these environment variables will be passed to the docker container
       - CIBW_ENVIRONMENT: PIP_CONFIG_FILE=buildconfig/pip_config.ini PORTMIDI_INC_PORTTIME=1 SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=disk
-      - CIBW_BUILD: cp* pp*
       - CIBW_ARCHS: aarch64
       - CIBW_SKIP: '*-musllinux_*'
       - CIBW_MANYLINUX_AARCH64_IMAGE: pygame/manylinux2014_base_aarch64
@@ -28,7 +27,7 @@ jobs:
       - run:
           name: Build the Linux wheels.
           command: |
-            pip3 install --user cibuildwheel==2.5.0
+            pip3 install --user cibuildwheel==2.10.2
             PATH="$HOME/.local/bin:$PATH" cibuildwheel --output-dir wheelhouse
 
       - store_artifacts:
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -75,36 +75,39 @@ jobs:
       fail-fast: false  # if a particular matrix build fails, don't skip the rest
       matrix:
         # Split job into 5 matrix builds, because GH actions provides 5 concurrent
-        # builds on macOS
+        # builds on macOS. This needs to be manually kept updated so that each
+        # of these builds take roughly the same time
         include:
           - { 
-            name: "x86_64 (cp36-cp310)",
+            name: "x86_64 (CPython 3.6 3.10 and above)",
             macarch: x86_64,
-            pyversions: cp36-* cp310-*
+            # pattern matches 6 or any 2 digit number
+            pyversions: "cp3{6,[1-9][0-9]}-*"
           }
 
           - { 
-            name: "x86_64 (cp37-pp37)",
+            name: "x86_64 (Python 3.7)",
             macarch: x86_64,
-            pyversions: cp37-* pp37-*
+            pyversions: "?p37-*"
           }
 
           - { 
-            name: "x86_64 (cp38-pp38)",
+            name: "x86_64 (Python 3.8)",
             macarch: x86_64,
-            pyversions: cp38-* pp38-*
+            pyversions: "?p38-*"
           }
 
           - { 
-            name: "x86_64 (cp39-pp39)",
+            name: "x86_64 (Python 3.9)",
             macarch: x86_64,
-            pyversions: cp39-* pp39-*
+            pyversions: "?p39-*"
           }
 
           - { 
-            name: "arm64 (cp38-cp39-cp310)",
+            name: "arm64 (CPython 3.8 and above)",
             macarch: arm64,
-            pyversions: cp38-* cp39-* cp310-*
+            # pattern matches any number from 8 to 99
+            pyversions: "cp3{8,9,[1-9][0-9]}-*"
           }
 
     env:
@@ -151,7 +154,7 @@ jobs:
           key: macdep-${{ hashFiles('buildconfig/manylinux-build/**') }}-${{ hashFiles('buildconfig/macdependencies/*.sh') }}-${{ matrix.macarch }}
 
       - name: Build and test wheels
-        uses: pypa/cibuildwheel@v2.5.0
+        uses: pypa/cibuildwheel@v2.10.2
 
       - uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/build-manylinux.yml
+++ b/.github/workflows/build-manylinux.yml
@@ -32,24 +32,29 @@ on:
 
 jobs:
   build:
+    name: ${{ matrix.image }} [${{ matrix.arch }}]
     runs-on: ubuntu-20.04
     strategy:
       fail-fast: false  # if a particular matrix build fails, don't skip the rest
       matrix:
         # Split job into many matrix builds, because GH actions provides 20
-        # concurrent builds on ubuntu. 10 are used here
+        # concurrent builds on ubuntu. 6 are used here
         include:
           # no pypy and cpython >= cp310 on manylinux1
-          - { image: manylinux1, arch: x86_64, pyversions: cp36-* cp37-* cp38-* cp39-* }
-          - { image: manylinux1, arch: i686, pyversions: cp36-* cp37-* cp38-* cp39-* }
+          - { image: manylinux1, arch: x86_64, pyversions: "cp3[6-9]-*" }
+          - { image: manylinux1, arch: i686, pyversions: "cp3[6-9]-*" }
 
-          # no pypy 3.9 on manylinux2010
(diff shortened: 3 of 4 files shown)


.github/workflows/build-macos.yml
```diff
@@ -75,36 +75,37 @@ jobs:
...
-            pyversions: cp39-* pp39-*
+            pyversions: "?p39-*"
           }
 
           - { 
-            name: "arm64 (cp38-cp39-cp310)",
+            name: "arm64 (CPython 3.8 and above)",
             macarch: arm64,
-            pyversions: cp38-* cp39-* cp310-*
+            pyversions: "cp3{8,9,[1-9][0-9]}-*"
```
Starbuck5, line 108: Using globs seems like it makes it less clear what exactly does what.

ankith26: Well, I think the more compact representation is better. Besides, there's a human readable name just a few lines above each of those globs anyways

Starbuck5: compact representation? I expected you to talk about forwards compatibility or something. Explicit is better than implicit?

buildconfig/manylinux-build/build-wheels.sh
```diff
@@ -1,17 +1,12 @@
 #!/bin/bash
 set -e -x
 
+# This file is not used by CI anymore. This only exists incase someone wants to
+# generate manylinux wheels locally for testing
```
Starbuck5, line 5: What's the point of all the changes to this file if it's not being used anymore in CI? Would anyone actually use this locally? Have you tested these changes locally?

.github/workflows/build-macos.yml
ankith26: > I expected you to talk about forwards compatibility or something.

Yeah well that's another thing, but as you know in a year when we have to add 3.12 here a lot more things could change (like having to accommodate a new pypy version, or drop an older one) and we would probably have to move around stuff here a bit anyways.

Also I don't see how this is less clear, but this is subjective anyways. I see this pattern being used in many other projects using cibuildwheel (and also used in cibuildwheel docs) so I assumed people won't have much of a problem with it. Also this is not really an "implicit" thing.

buildconfig/manylinux-build/build-wheels.sh
ankith26: Can't speak for others, but I do use this file locally, and often with a lot of custom changes ;)
I've also tested these changes

.github/workflows/build-macos.yml
Starbuck5: I'm not going to be a complete stick in the mud about this from the perspective of approving this PR, but this is absolutely an implicit thing.

Explicit: `cp38-* cp39-* cp310-*`
Implicit: `cp3{8,9,[1-9][0-9]}-*`

ankith26: Well, there's still a `*` at the end of what you call "explicit" so I don't really see how this is an implicit vs explicit thing
Explicit is better than implicit, but practicality beats purity :wink:

Starbuck5: All I'm saying is that to actually understand the new style, you have to translate it in your head into the old style, and then you can think about it.

.github/workflows/build-manylinux.yml
```diff
@@ -32,24 +32,23 @@ on:
...
-          - { image: manylinux2010, arch: i686, pyversions: cp* pp37-* pp38-* }
+          # no pypy >= 3.9 and cpython >= cp311 on manylinux2010
+          - { image: manylinux2010, arch: x86_64, pyversions: "cp3{[6-9],10}-* pp3{7,8}-*" }
 
-          # all cpython and pypy versions should support this
-          - { image: manylinux2014, arch: x86_64, pyversions: cp* pp* }
-          - { image: manylinux2014, arch: i686, pyversions: cp* pp* }
+          # all cpython and pypy versions should support this for now
+          - { image: manylinux2014, arch: x86_64, pyversions: "*" }
+          - { image: manylinux2014, arch: i686, pyversions: "*" }
```
Starbuck5, line 51: ALL?

You mean all valid versions from the perspective of cibuildwheel? What does that mean?

.github/workflows/build-macos.yml
```diff
@@ -75,36 +75,37 @@ jobs:
       fail-fast: false  # if a particular matrix build fails, don't skip the rest
       matrix:
         # Split job into 5 matrix builds, because GH actions provides 5 concurrent
-        # builds on macOS
+        # builds on macOS. This needs to be manually kept updated so that each
+        # of these builds take roughly the same time
         include:
           - { 
-            name: "x86_64 (cp36-cp310)",
+            name: "x86_64 (CPython 3.6 3.10 and above)",
```
Starbuck5, line 82: Isn't this CPython 3.6 and above?

Right now it sounds like: 3.6, 3.10, and 3.11+

Starbuck5: Ah geez this is 3.6 or 3.[double digit number].

That's hard to parse for me.

Starbuck5:
Can you explain what gettext is and why it's changing?

.circleci/config.yml
```diff
@@ -14,7 +14,6 @@ jobs:
     environment:
       # these environment variables will be passed to the docker container
       - CIBW_ENVIRONMENT: PIP_CONFIG_FILE=buildconfig/pip_config.ini PORTMIDI_INC_PORTTIME=1 SDL_VIDEODRIVER=dummy SDL_AUDIODRIVER=disk
-      - CIBW_BUILD: cp* pp*
```
Starbuck5, line 17: I don't understand why this change is here.

.github/workflows/build-manylinux.yml
ankith26: All versions the specific cibuildwheel version supports

.github/workflows/build-macos.yml
ankith26: It is what it sounds

.circleci/config.yml
ankith26: That line was basically redundant because cibuildwheel default is to just do that

ankith26:
Basically, `gettext` is an low-level library for handling text things used by glib, which in turn is used by many of our dependencies (most notably by fluidsynth, which is in turn used by SDL_mixer)

What changed when I bumped cibuildwheel, is that the python it uses internally started depending on `/usr/local/opt/gettext`. But that was being removed as part of our mac dep builds because we install our own gettext on arm64 (in `/usr/local/gettext`, which system python can't use). So cibuildwheel started failing on startup, so I removed the line that was purging `/usr/local/opt/gettext`. Now we have both gettexts peacefully coexisting, the system python picks the one in opt (intel), and our builds pick what we install (arm64)

.github/workflows/build-manylinux.yml
Starbuck5: So right now that happens to line up with what we support 3.6, but it won't always? I know that we have a pinned dependency version, but I'm still thinking about that.

ankith26: Till someone goes ahead and bumps the cibuildwheel version, nothing is going to change here. Even when cibuildwheel drops 3.6 in a version, we won't be affected till we bump it on our side (which is why pinned dependencies are ideal)

Also in this particular instance `cp* pp*` is basically equivalent to `*` but IMHO more readable as it is immediately clear that this tells cibuildwheel to use everything it's got

ankith26: Till someone goes ahead and bumps the cibuildwheel version, nothing is going to change here. Even when cibuildwheel drops 3.6 in a version, we won't be affected till we bump it on our side (which is why pinned dependencies are ideal)

Also in this particular instance `cp* pp*` is basically equivalent to `*` but IMHO more readable as it is immediately clear that this tells cibuildwheel to use everything it's got

Starbuck5: Leave a comment though just for future people?

Starbuck5 (approved):
This PR was hard to review, because it's stuff I'm not familiar with, and because there are a lot of changes in this PR besides just adding 3.11 support. I get that everything is in the same vein, but it would've been easier perhaps as smaller PRs that just did the 1 thing? Idk.

Anyways, I asked a bunch of questions and got a bunch of answers, so I'll go ahead and approve. Thanks for working on this Ankith.

illume:
I'd prefer if we keep this for after 2.1.3. Cool?

ankith26:
The old 32bit manylinux wheels drop, I definitely don't mind postponing it because it's more of a cleanup thing anyways. But I would like 2.1.3 to have python 3.11 support on all platforms (and we have it on windows already, so this PR is meant to be a follow-up on mac/manylinux)

I could open a separate, minimal PR for that if that's okay with you

ankith26:
(this is the relevant windows PR #3486)

illume:
Yeah, cool thanks. One without dropping would be wonderful.

ankith26:
Now this PR is doing only the 3.11 bump

I also removed some of the minor docker file changes I did here, but the issue was that some manylinux1 stuff was commented out (which affected dev6, manylinux1 is still lagging behind in there)

illume (approved):
Awesome sauce. Thanks :)

PR #3486, opened 2022-10-09
Add 3.11.0.rc2 to Appveyor

Starbuck5:
The Python 3.11 full release is coming soon, and `rc2` is going to be ABI compatible with 3.11.0, so this puts it in CI so we can generate Windows Python 3.11 wheels.


ankith26 (approved):
LGTM, thank youss ⭐

MyreMylar (approved):
Yes, this seems important. Otherwise we'll  get waves of 'why doesn't pygame work' issues again after 3.11.

PR #3487, opened 2022-10-09
Internalize "SDL"_UCS4ToUTF8

### LINKED ISSUE #3417: We should never define our own "SDL_" functions
We should never define our own "SDL_" functions. This has bitten us in the past. Instead we should create pg_XXX functions which inside may call the SDL_XXX function if it's available, or include the function body. This way if someone links something with the SDL_XXX function in the future we have no problems.

Discussed here: https://github.com/pygame/pygame/pull/3349/files#diff-d28dfba3b4d41a0b5734aed1829c6878de9ef71a82f160f8ec176827298c7f48R449

If we compile with an SDL that doesn't support an SDL_XXX function, but then a future SDL dll is linked with a SDL_XXX function then it's broken. If instead we had made our pg_XXX function that internally checks if SDL_XXX is there, it's likely to work in this forward compatible way... and also when statically linking.

Starbuck5:
Fix #3417

### DIFF
--- a/src_c/key.c
+++ b/src_c/key.c
@@ -439,14 +439,9 @@ static const char *SDL1_scancode_names[SDL_NUM_SCANCODES] = {
     "AudioFastForward",
 };
 
-#if defined(BUILD_STATIC)
-// from SDL_keyboard.c
-extern char *
-SDL_UCS4ToUTF8(Uint32 ch, char *dst);
-#else
 /* Taken from SDL_iconv() */
 char *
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
Thanks!

PS: #3426 also fixes the bug this is fixing by completely removing the need for this function. It also fixes failing unit tests and makes the behaviour deternministic, but it's a larger change

Starbuck5:
Ah, I did not realize that.

illume (approved):
👍 thanks!

PR #3490, opened 2022-10-10
Cleanup of symbian code and references to bitbucket/motherhamster bugzilla

### LINKED ISSUE #3372: Update comment in `math.c` describing "bug 52"
The following comment and code is at the top of `math.c`:
```
/* Adjust gcc 4.4 optimization for floating point on x86-32 PCs running Linux.
 * This addresses bug 52:
 * http://pygame.motherhamster.org/bugzilla/show_bug.cgi?id=52
 */
#if defined(__GNUC__) && defined(__linux__) && defined(__i386__) && \
    __SIZEOF_POINTER__ == 4 && __GNUC__ == 4 && __GNUC_MINOR__ >= 4
#pragma GCC optimize("float-store")
#endif
```

The link is dead however and so there's no way to see what "bug 52" was. Is there another record of that bug somewhere that could be linked to?

ankith26:
fixes #3372 

Minor PR to clean up some old stuffs in the codebase. The rationale is to hopefully make it a little easier for potential new contributors peeking into the codebase. We have had bits of symbian support (and I suspect it is very incomplete), but SDL2 has dropped support for it since about a decade now and I don't think it's coming back anytime soon. I also can't find any doc about python ever having proper support for it (was it a python fork?)

Also updates issue references around the codebase to github, because bitbucket and motherhamster links are dead (it's a nice thing that the old issues were ported with the same numbers 😄)


test/surface_test.py
```diff
@@ -3848,8 +3848,8 @@ def test_overlap_check(self):
     # https://github.com/pygame/pygame/issues/370#issuecomment-364625291
     @unittest.skipIf("ppc64le" in platform.uname(), "known ppc64le issue")
     def test_colorkey(self):
-        # Check a workaround for an SDL 1.2.13 surface self-blit problem
```
Starbuck5, line 3851: I like the retargeted issue link, but I feel we should keep the specificity around the problem offered by these exact details.

src_py/__init__.py
```diff
@@ -178,59 +178,6 @@ def Cursor(*args):  # pylint: disable=unused-argument
     pixelcopy = MissingModule("pixelcopy", urgent=1)
 
 
-def warn_unwanted_files():
```
Starbuck5, line 181: I tracked this down in git, and it was added sometime between 1.8.1 and 1.9.0, and has been disabled since 1.9.1. So this all went down in 2008/2009, seems safe to remove in the present day.

test/surface_test.py
ankith26: I would generally agree, but this specificity is IMHO useless if the version in question is 15 years old ;) If it's a SDL 2.x.x version, it definitely makes sense to have a mention of it

src_py/__init__.py
Starbuck5: Safe to remove once the `del` of it is removed at the end of `__init__.py`, that is 😄

ankith26: Oh yeah, good catch

ankith26: And I thought I could get away with not testing a PR as "minor" as this. Good thing CI exists

test/surface_test.py
Starbuck5: I've had to go deep into things with SDL_ttf, I appreciated that the pygame comments mentioned exact versions, even if they are presently obsolete. Having the extra bit of information in the comments does no harm.

ankith26: I've reverted that change

Starbuck5 (approved):
Looks good, thanks!

zoldalma999 (approved):
Thanks!

Issue #3419, opened 2022-08-21
Segfault when creating a child class of pygame.mixer.Sound


novialriptide:
Here's the code and output:

```python
import pygame
pygame.init()

class Amogus(pygame.mixer.Sound):
    def __init__(self):
        pass

a = Amogus()
a.get_volume()
```
```
PS C:\Users\novia\documents\github\pygame> python test.py
pygame 2.1.2 (SDL 2.0.18, Python 3.9.0)
Hello from the pygame community. https://www.pygame.org/contribute.html
Fatal Python error: pygame_parachute: (pygame parachute) Segmentation Fault
Python runtime state: initialized

Current thread 0x0000c344 (most recent call first):
  File "C:\Users\novia\documents\github\pygame\test.py", line 9 in <module>
```

Starbuck5:
Perhaps this is because the expected `__init__` sets internal class state that is referenced in `get_volume()`.

PurityLake:
Seems the issue in the C code is that `snd_get_volume` requests a chunk which returns NULL then calling `Mix_VolumeChunk` on that null chunk.

Should this return something like None? Or should we raise an error?

novialriptide:
> Seems the issue in the C code is that `snd_get_volume` requests a `Mix_Chunk` which returns NULL then calling `Mix_VolumeChunk` on that null chunk.
> 
> Should this return something like None? Or should we raise an error?
> 
> Mainly the issue here anyway is incorrectly not called `super().__init__(*args, **kwargs)` or whatever is the case for Sound.
> 
> Also, ya it shouldn't be subclassable it's tp_flags includes `Py_TPFLAGS_BASETYPE`

I'm on my phone right now so I can't really rest this, but if we do mimic the behavior when we do `pygame.mixer.Sound(None)`

PurityLake:
> > Seems the issue in the C code is that `snd_get_volume` requests a `Mix_Chunk` which returns NULL then calling `Mix_VolumeChunk` on that null chunk.
> > Should this return something like None? Or should we raise an error?
> > Mainly the issue here anyway is incorrectly not called `super().__init__(*args, **kwargs)` or whatever is the case for Sound.
> > Also, ya it shouldn't be subclassable it's tp_flags includes `Py_TPFLAGS_BASETYPE`
> 
> I'm on my phone right now so I can't really test this, but we should mimic the behavior when we do `pygame.mixer.Sound(None)`

[...]

PR #3492, opened 2022-10-10
Fixed segault in subclass of Sound

PurityLake:
Refers to #3419 

Despite Sound being set as a base type, it can still be subclassed, this is a fix for if  `super().__init__()` is not called and the sound object is not initialized correctly. I have this as a draft to see if someone might know a way to enforce Sound not being subclassed.


PurityLake:
I feel this, may not be the resolution to this issue, it seems using Pygame 2.1.2 causes the crash but the latest pygame 2.1.3.dev7 the issue is not there.

ankith26 (commented):
This is definitely a step in the (imho) right direction, thanks! 🎉 

You seem to have already covered most places, but you missed the chunk handling in `chan_queue` and `chan_play` that needs similar guards. There's also `endsound_callback`, but I don't think it needs the guard because it is just using the sound object that `chan_queue` is setting
For extra minor style points, you could also make this check a macro

ankith26:
> Despite Sound being set as a base type, it can still be subclassed

The very fact that it has the `Py_TPFLAGS_BASETYPE` flag set makes it subclassable. Without that flag, subclassing it would raise errors. This is not something I'd like to change, I'd instead prefer the fix this PR is doing

PurityLake:
Made those changes now. I should have all cases checked for now

src_c/mixer.c
```diff
@@ -34,6 +34,13 @@
 
 #define PyBUF_HAS_FLAG(f, F) (((f) & (F)) == (F))
 
+#define CHECK_CHUNK_VALID(CHUNK, RET)                              \
+    if (CHUNK == NULL) {                                           \
+        PyErr_SetString(PyExc_RuntimeError,                        \
+                        "Sound object not initalised correctly."); \
+        return RET;                                                \
```
ankith26, line 41: I'd recommend wrapping `CHUNK` and `RET` in parenthesis. This is done because macro callers often expect the stuff they pass to be handled as normal expressions, and macro variable substitution can sometimes mess logic.

Not something that is a problem right now with the current uses of this macro, but could be in the future uses of this macro. 

Also while you are at it, it would also be cool if you could elaborate a bit more on that error message. Something mentioning that `__init__` was not called, so that people dealing with this error know exactly what to fix

ankith26 (approved):
left two minor suggestions, but otherwise this PR LGTM! 💎

src_c/mixer.c
PurityLake: Those changers have been made. Hope the error message is better now.

src_c/mixer.c
```diff
@@ -34,6 +34,16 @@
 
 #define PyBUF_HAS_FLAG(f, F) (((f) & (F)) == (F))
 
+static const char *sound_not_initialised_errmsg =
```
zoldalma999, line 37: Not sure why this needs a variable defined, since this is only used once. You could just put the string in the `PyErr_SetString` call. Clang will format it nicely for you as well.

zoldalma999 (commented):
It would be nice if you could add a test for subclassing `Sound`s (as well as subclassing without `super().__init__()`)

src_c/mixer.c
PurityLake: Good point, that has been changed

PurityLake:
@zoldalma999 I have added those tests now, thanks for your input

src_c/mixer.c
```diff
@@ -34,6 +34,14 @@
...
 #define PyBUF_HAS_FLAG(f, F) (((f) & (F)) == (F))
 
+#define CHECK_CHUNK_VALID(CHUNK, RET)                                      \
+    if ((CHUNK) == NULL) {                                                 \
+        PyErr_SetString(PyExc_RuntimeError,                                \
+                        "__init__() was not called on Sound object so it " \
+                        "failed to setup correctly.");                     \
+        return (RET);                                                      \
+    }
+
```
nthykier, line 44: Nit: I think this should have used the `do { ... } while(0)` trick.  However, since we do not use this with `if (...) CHECK_CHUNK_VALID(...); else ...` then I do not think it matters and I think having this merged sooner is a bigger net-benefit then pedantic correctness of a macro.

nthykier (approved):
LGTM

zoldalma999 (approved):
Looks good to me as well!

Do you mind if we squash and merge it (instead of a normal merge)?

src_c/mixer.c
PurityLake: Ya I feel it isn't completely necessary. Unless someone else chimes in

PurityLake:
@zoldalma999 I'd be fine with that, I did make a lot of commute

novialriptide:
Thanks for the fix!

PR #3493, opened 2022-10-11
Fixed incorrect volume when chunk is NULL

PurityLake:
Slight bug fix to `get_volume` in the Sound class.

`Mix_VolumeChunk` when passed `NULL` returns -1 but it was mistakenly divided by 128 instead of returning -1.

### DIFF
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -725,6 +725,9 @@ snd_get_volume(PyObject *self, PyObject *_null)
     MIXER_INIT_CHECK();
 
     volume = Mix_VolumeChunk(chunk, -1);
+    if (volume == -1) {
+        return PyLong_FromLong(volume);
+    }
     return PyFloat_FromDouble(volume / 128.0);
 }
 


src_c/mixer.c
```diff
@@ -725,6 +725,9 @@ snd_get_volume(PyObject *self, PyObject *_null)
     MIXER_INIT_CHECK();
 
     volume = Mix_VolumeChunk(chunk, -1);
+    if (volume == -1) {
+        return PyLong_FromLong(volume);
+    }
```
ankith26, line 730: Hmm... `Mix_VolumeChunk` returns `-1` when `chunk` is `NULL`. I feel like that's the real problem here. Why is a python `Sound` class associated with a `NULL` `Mix_Chunk`?

It'd also be nice to not forward things like `-1`s into python land. It's neater to raise python exceptions

PurityLake: I see what you mean, guess I've been using low-level languages too long these days. I feel then this PR is not needed, but I reopened a PR that does these checks.

Issue #3097, opened 2022-03-28
test_src_alpha_sdl2_blitter should not test for exact pixel perfect match (and instead allow a small delta)


swt2c:
When running the unit tests for pygame 2.1.2 on Debian Unstable on mips64el, mipsel, ppc64el architectures, I'm seeing a bunch of assertion failures in `test_src_alpha_sdl2_blitter`.  This is with both Python 3.9 and Python 3.10.

Some examples of the failures (full logs attached).  The numbers are close though - so are these floating point rounding errors or something?

```
======================================================================
FAIL: test_src_alpha_sdl2_blitter (pygame.tests.surface_test.GeneralSurfaceTests) (src_r=65, src_b=199, src_a=65, dest_r=0, dest_b=255, dest_a=255)
Checking that the BLEND_ALPHA_SDL2 flag works - this feature
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.9_pygame/build/pygame/tests/surface_test.py", line 1581, in test_src_alpha_sdl2_blitter
    self.assertEqual(results[key], results_expected[key])
AssertionError: (16, 255, 240, 254) != (16, 253, 239, 253)

======================================================================
FAIL: test_src_alpha_sdl2_blitter (pygame.tests.surface_test.GeneralSurfaceTests) (src_r=126, src_b=127, src_a=126, dest_r=0, dest_b=255, dest_a=255)
Checking that the BLEND_ALPHA_SDL2 flag works - this feature
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.9_pygame/build/pygame/tests/surface_test.py", line 1581, in test_src_alpha_sdl2_blitter
    self.assertEqual(results[key], results_expected[key])
AssertionError: (62, 255, 192, 254) != (62, 253, 190, 253)

======================================================================
FAIL: test_src_alpha_sdl2_blitter (pygame.tests.surface_test.GeneralSurfaceTests) (src_r=127, src_b=126, src_a=127, dest_r=0, dest_b=255, dest_a=255)
Checking that the BLEND_ALPHA_SDL2 flag works - this feature
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #3494, opened 2022-10-11
Relax strict pixel match tests in test_src_alpha_sdl2_blitter by allowing a small delta

Temmie3754:
Resolves the small issue in a test raised by #3097 
Allows for a delta of up to 2 which resolves all errors presented in the original problem.

### DIFF
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1588,7 +1588,9 @@ class GeneralSurfaceTests(AssertRaisesRegexMixin, unittest.TestCase):
                     key = ((dst_r, dst_b, dst_a), (src_r, src_b, src_a))
                     results[key] = tuple(dest_surf.get_at((65, 33)))
                     for i in range(4):
-                        self.assertAlmostEqual(results[key][i], results_expected[key][i], delta=4)
+                        self.assertAlmostEqual(
+                            results[key][i], results_expected[key][i], delta=4
+                        )
 
         # print("(dest_r, dest_b, dest_a), (src_r, src_b, src_a): color")
         # pprint(results)
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1588,7 +1588,7 @@ class GeneralSurfaceTests(AssertRaisesRegexMixin, unittest.TestCase):
                     key = ((dst_r, dst_b, dst_a), (src_r, src_b, src_a))
                     results[key] = tuple(dest_surf.get_at((65, 33)))
                     for i in range(4):
-                        self.assertAlmostEqual(results[key][i], results_expected[key][i], None, None, 4)
+                        self.assertAlmostEqual(results[key][i], results_expected[key][i], delta=4)
 
         # print("(dest_r, dest_b, dest_a), (src_r, src_b, src_a): color")
         # pprint(results)
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1586,12 +1586,13 @@ class GeneralSurfaceTests(AssertRaisesRegexMixin, unittest.TestCase):
                         src_surf, (0, 0), special_flags=pygame.BLEND_ALPHA_SDL2
                     )
                     key = ((dst_r, dst_b, dst_a), (src_r, src_b, src_a))
-                    results[key] = dest_surf.get_at((65, 33))
-                    self.assertEqual(results[key], results_expected[key])
+                    results[key] = tuple(dest_surf.get_at((65, 33)))
+                    for i in range(4):
+                        self.assertLessEqual(
+                            abs(results[key][i] - results_expected[key][i]), 2)
 
         # print("(dest_r, dest_b, dest_a), (src_r, src_b, src_a): color")
         # pprint(results)
-        self.assertEqual(results, results_expected)
 
     def test_opaque_destination_blit_with_set_alpha(self):
         # no set_alpha()
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -1588,8 +1588,7 @@ class GeneralSurfaceTests(AssertRaisesRegexMixin, unittest.TestCase):
                     key = ((dst_r, dst_b, dst_a), (src_r, src_b, src_a))
                     results[key] = tuple(dest_surf.get_at((65, 33)))
                     for i in range(4):
-                        self.assertLessEqual(
-                            abs(results[key][i] - results_expected[key][i]), 2)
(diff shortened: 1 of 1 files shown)


ankith26 (commented):
Thanks for the PR! 🎉 🎈 

Left a minor review

test/surface_test.py
```diff
@@ -1586,12 +1586,13 @@ def test_src_alpha_sdl2_blitter(self):
                         src_surf, (0, 0), special_flags=pygame.BLEND_ALPHA_SDL2
                     )
                     key = ((dst_r, dst_b, dst_a), (src_r, src_b, src_a))
-                    results[key] = dest_surf.get_at((65, 33))
-                    self.assertEqual(results[key], results_expected[key])
+                    results[key] = tuple(dest_surf.get_at((65, 33)))
+                    for i in range(4):
+                        self.assertLessEqual(
+                            abs(results[key][i] - results_expected[key][i]), 2)
```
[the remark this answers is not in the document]
Temmie3754: Alright, have changed to reflect your suggestions. Thanks for the help.

test/surface_test.py
```diff
@@ -1586,12 +1586,12 @@ def test_src_alpha_sdl2_blitter(self):
                         src_surf, (0, 0), special_flags=pygame.BLEND_ALPHA_SDL2
                     )
                     key = ((dst_r, dst_b, dst_a), (src_r, src_b, src_a))
-                    results[key] = dest_surf.get_at((65, 33))
-                    self.assertEqual(results[key], results_expected[key])
+                    results[key] = tuple(dest_surf.get_at((65, 33)))
+                    for i in range(4):
+                        self.assertAlmostEqual(results[key][i], results_expected[key][i], None, None, 4)
```
ankith26, line 1591: Instead of passing two `None`s like that (which would not work for the 3rd argument of this function which needs an integer), you could instead pass the `4` as a keyword argument like so
```suggestion
                        self.assertAlmostEqual(results[key][i], results_expected[key][i], delta=4)
```

Temmie3754: Resolved that as well now

ankith26 (approved):
LGTM! 🎉 😎 

(There is a minor CI fail in here, could you please run `black` on the python file you changed in this PR to fix the formatting?)

Temmie3754:
Formatting should hopefully be correct finally

Starbuck5:
The source of this problem is very mysterious. I think it's SIMD related, but why? SIMD doesn't do approximation, it just does acceleration.

I don't think the solution is to just add a delta to the test. I would prefer it to test that the result is EXACTLY X or Y, rather than testing whether it's within 4 of X.

ankith26:
I am of the view that as long as something blitting related does not change much visually, it shouldn't be considered a break. 
The underlying formula is an implementation detail, especially while using the SDL blitter, we should not treat things as a break when they make minor changes on their side. So to me having a delta makes sense, but I don't mind changing the exact number specified in the delta

Starbuck5:
Blitting should be 100% deterministic. Input A goes to output B, every time, 100% accurate. The fact that it isn't (on the SDL2 blitter) is extremely unsettling.

I believe our current situation is that Input A goes to output B or output C, based on SDL version / architecture. Therefore I'd like to test the output against B and C, and only fail if it is neither.

I don't want it to drift again without our noticing, and that's what a delta does.

Temmie3754:
Alright I'll look into changing the implementation to use fixed outputs as specified by the original error but I will need someone else who has those architectures to test that it resolves the issue.

ankith26:
> Blitting should be 100% deterministic. Input A goes to output B, every time, 100% accurate. The fact that it isn't (on the SDL2 blitter) is extremely unsettling.

Blitting should be deterministic for a given SDL version? yes. The *exact same* on all platforms across all SDL versions of the past and future? Probably too much to ask, even if desirable. SDL can make algorithmic changes, and since the exact algorithm is not a documented API detail, I don't see why they shouldn't be making tiny changes to it if it helps with performance, even though the exact pixel colours change a tiny bit. I can understand if there's a big change that has visual affect, it could be considered a breaking change.

If I had to guess, I'd say this PR changed things https://github.com/libsdl-org/SDL/pull/5430 (I tested on windows 2.24.1) but I'm not very sure. Atleast I didn't find anything else blit related that changed.

> I don't want it to drift again without our noticing, and that's what a delta does.

I'm saying we shouldn't care of minor drifts. I can't tell the difference between `(255, 10, 100)` and `(253, 12, 98)` when they are in the same image side by side. Visually I can't see the line of separation and it looks all pink to me, even though I'm sure the lines technically there. Am I colour blind? Probably not 😄 

I also don't want to add the maintenance burden of having to keep updating our tests everytime SDL does a minor change like now. pygame users who resort to using the SDL blitter are the same target group of people who care about performance and not exact match with the pygame blitter.

Temmie3754:
Created an implementation that checks if its either the normal outputs or the ones provided in the original issue, just need to come to a decision on which version would be a better choice now. Can view the differences here in the meantime https://github.com/Temmie3754/pygame/tree/surface-test-fixes.

MyreMylar (approved):
I mentioned this on the discord too but I am generally in favour of this change for the SDL2 alpha blitter tests. This isn't the first time, or even the second time that the SDL formula has changed slightly - in fact these changes are the whole reason we ended up creating the default pygame alpha blitter. Ultimately these test failing is outside of our control as we can't stop SDL slightly tweaking their alpha blitting formula.

For most users, they will never use the SDL2 alpha blitter as you have to go and seek it out and specifically flag to use it, and for those that do so a tiny colour discrepancy (likely less than 0.5%) between SDL versions on some platforms when alpha blitting - that is not normally visually noticeable will be unlikely to cause issues.

I do think we should create a second issue to update the docs for the `BLEND_ALPHA_SDL2` in 2.1.4 after all the performance changes around the alpha blitters that are floating around have made it in. I suspect the original description of being 'sometimes faster' will be increasingly inaccurate by that point - especially if we have full AVX2 acceleration on the pygame alpha blitters. Then at that point we can also add a note in the docs that the SDL2 blitting formula differs between different versions of SDL.

ankith26:
I'll merge it in now, it's a non-risky change and as said before should better future-proof this test. If the need arises in the future, we can consider a stricter but more-maintenance-ey test then. 
Definitely the docs around `BLEND_ALPHA_SDL2` needs updating to reflect the latest pygame blitter improvements, and the inter-SDL-version minor differences, but as MyreMylar says that can happen in a future PR, lets get this one in 🥳

Issue #3437, opened 2022-09-04
Remove __iter__ from pgScancodeWrapper


Dreamsorcerer:
From what I can tell, pgScancodeWrapper was introduced for SDL 2 to convert values when indexing.

However, this has a regression when using `__iter__`. An old example like this no longer works correctly (from Beginning Game Developement with Python and Pygame):

```
for key_constant, pressed in enumerate(pygame.key.get_pressed()):
    if pressed:
        key_name = pygame.key.name(key_constant)
```

In this example, pressing the space key actually gets the index of 44 (where K_SPACE is 32) and results in key_name being ",".
Clearly, iterating over this object provides no useful information, and confuses people using old examples.
So, we should just remove the `__iter__` capabilities from this object.

ankith26:
I can't think of a way we can fix this without a break, the main problem here is that the keycode values drastically changed in SDL2.
I agree, the best we can do at this stage is to make ` __iter__` explicitly error here.

PR #3495, opened 2022-10-11
pgScancodeWrapper now raises exception when trying to use __iter__

PurityLake:
Refers to #3437 

Added a `tp_iter` and `tp_iternext` field to pgScancodeWrapper.

Made a custom exception also in `key.c` called `pgExc_IterNotSupportedError`


PurityLake:
Ok pypy is complaining here. It seems to be causing a segfault on calling `pygame.key.get_pressed` or even `pygame.key.ScancodeWrapper`

Working to fix this problem currently

src_c/key.c
```diff
@@ -102,10 +105,20 @@ pg_scancodewrapper_repr(pgScancodeWrapper *self)
...
+                    "Iterating over key states is no long supported");
+    return NULL;
+}
+
 static PyTypeObject pgScancodeWrapper_Type = {
     PyVarObject_HEAD_INIT(NULL, 0).tp_name = _PG_SCANCODEWRAPPER_TYPE_FULLNAME,
     .tp_repr = (reprfunc)pg_scancodewrapper_repr,
     .tp_as_mapping = &pg_scancodewrapper_mapping,
+    .tp_iter = &pg_iter_raise,
+    .tp_iternext = &pg_iter_raise,
```
ankith26, line 121: The `&` in both these lines is not needed: `tp_iter` and `tp_iternext` need function pointers, and not function pointer to pointers

ankith26 (commented):
Thanks for the PR!

This is just a minor thing, but I think you can use the already existing `pygame.error` (`pgExc_SDLError` at the C level) directly instead of a custom error. Or alternatively make the custom error a subclass of `pygame.error`

ankith26:
A testcase to make sure this error is correctly raised would be nice too

PurityLake:
Ok I believe I have fixed the pypy crashes and bugs, so all is good

Starbuck5 (commented):
I'm still unconvinced that this is the right course of action, but I may come around.

I don't want to do a custom exception in this way. Why not like a TypeError? It's what other objects do.

```py
>>> a = 5
>>> for b in a:
...     print(b)
...
TypeError: 'int' object is not iterable
```

ankith26:
> I'm still unconvinced that this is the right course of action

Yeah I agree this is not ideal, but I also can't think of a better solution? Do you have something in mind not already considered?

zoldalma999:
I am with Starbuck, `TypeError` would work better here imo

PurityLake:
@zoldalma999 I have changed that

docs/reST/ref/key.rst
```diff
@@ -266,6 +266,13 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
       translate these pushed keys into a fully translated character value. See
       the ``pygame.KEYDOWN`` events on the :mod:`pygame.event` queue for this
       functionality.
+   
+   .. versionadded:: 2.1.4
+      The collection of bools returned by ``get_pressed`` can not be iterated
+      over. Previously when iterating over the collection using builtin
+      ``enumerate``, if the space bar is is pressed the value received from
+      ``enumerate`` is 44 but ``K_SPACE`` is actually 32. In this case calling
+      :func:`pygame.key.name()` returns a comma character, ",".
```
ankith26, line 275: This is some unneeded info IMHO and a bit to specific. Probably something more like `The collection of bools returned by ``get_pressed`` can not be iterated over because the indexes of the internal tuple do not correspond to the keycodes`

src_c/key.c
```diff
@@ -106,8 +150,12 @@ static PyTypeObject pgScancodeWrapper_Type = {
...
     .tp_repr = (reprfunc)pg_scancodewrapper_repr,
     .tp_as_mapping = &pg_scancodewrapper_mapping,
-    .tp_flags =
-        Py_TPFLAGS_DEFAULT | Py_TPFLAGS_TUPLE_SUBCLASS | Py_TPFLAGS_BASETYPE,
+    .tp_iter = (getiterfunc)pg_iter_raise,
+    .tp_iternext = (iternextfunc)pg_iter_raise,
+#ifdef PYPY_VERSION
+    .tp_new = pg_scancodewrapper_new,
+#endif
+    .tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
```
ankith26, line 158: hmmm, removing the `Py_TPFLAGS_TUPLE_SUBCLASS` flag from here is probably not a good idea. This wrapper could be expected by user code to be a subclass tuple.

An alternate solution that I can think of, is making `tp_new` also error (because I don't think users would want to instantiate a scancodewrapper themselves) and instead create an internal new function at the C level that `key_get_pressed` can use

PurityLake: Oops, that was accidental, good eye

MyreMylar (approved):
LGTM 👍

nthykier (approved):

illume (approved):
👍 thanks

Issue #3098, opened 2022-03-30
Test failures with 2.1.2 on big endian architectures


swt2c:
When running the unit tests for pygame 2.1.2 on Debian Unstable on big endian architectures (e.g., s390x, ppc64, etc), I'm seeing a few test failures. This is with both Python 3.9 and Python 3.10.

```
======================================================================
FAIL: testSaveJPG (pygame.tests.image_test.ImageModuleTest)
JPG equivalent to issue #211 - color channel swapping
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.9_pygame/build/pygame/tests/image_test.py", line 151, in testSaveJPG
    self.assertEqual(approx(jpg_surf.get_at(posn)), approx(color))
AssertionError: (0, 0, 252, 255) != (252, 0, 0, 255)

======================================================================
FAIL: test_load_extended (pygame.tests.image_test.ImageModuleTest) [Test loading a svg] (filename='examples/data/teal.svg', expected_color=(0, 128, 128, 255))
can load different format images.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.9_pygame/build/pygame/tests/image_test.py", line 1051, in test_load_extended
    self.assertEqual(surf.get_at((0, 0)), expected_color)
AssertionError: (255, 128, 128, 0) != (0, 128, 128, 255)

======================================================================
FAIL: test_get_init__returns_exact_values_used_for_init (pygame.tests.mixer_test.MixerModuleTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/<<PKGBUILDDIR>>/.pybuild/cpython3_3.9_pygame/build/pygame/tests/mixer_test.py", line 103, in test_get_init__returns_exact_values_used_for_init
    self.assertEqual(tuple(init_conf.values()), mixer_conf)
AssertionError: Tuples differ: (11025, 16, 1) != (11025, -16, 1)

First differing element 1:
16
-16

- (11025, 16, 1)
+ (11025, -16, 1)
?         +

======================================================================
FAIL: test_aapolygon (__main__.GfxdrawDefaultTest)
aapolygon(surface, points, color): return None
----------------------------------------------------------------------
Traceback (most recent call last):
[...]

PR #3496, opened 2022-10-12
Add debian s390x and ppc64le tests to CI

### LINKED ISSUE #1291: ARM, and Big endian testing
We don't do testing on PRs for Big endian or ARM CPUs.

There's some docker images which use qemu emulation, which (hopefully) make this fairly easy. See this comment: https://github.com/travis-ci/travis-ci/issues/8869#issuecomment-528986808

ankith26:
PR to close #1291, and progress on #3098

This is an experimental PR to add exotic architectures s390x (which is big endian) and ppc64le.

## Why?

We probably don't have many actual users running on these architectures, but having source code support for these (with no official releases) probably makes things easier for debian packaging folks, who occasionally report that big endian fail, or weird architecture specific fail. We already have aarch64 support now, so it make sense if these two follow. Hopefully this PR makes things easier for debian packagers to adopt new pygame releases faster, when they don't have to worry about this.

This should probably be least prioritized among other CI runs, but atleast having this CI in the background would make it easier to track which commit broke what if someone needs to look at history.

## Technical details

This is implemented internally with QEMU and docker (but abstracted away by a github action). Naturally there is some QEMU slowness, but the action can cache dependencies in docker base images (which is lives in github packages, but I think it needs some perms set up for it?)

### DIFF
--- a/.github/workflows/build-debian-multiarch.yml
+++ b/.github/workflows/build-debian-multiarch.yml
@@ -1,0 +1,93 @@
+# Tests pygame on more exotic architectures. This is not something that is
+# actively supported, but source code support for this is nice to have. We
+# don't do any releases from here.
+
+name: Debian Multiarch
+
+# Run CI only when a release is created, on changes to main branch, or any PR 
+# to main. Do not run CI on any other branch. Also, skip any non-source changes 
+# from running on CI
+on:
+  release:
+    types: [created]
+  push:
+    branches: main
+    paths-ignore:
+      - 'docs/**'
+      - 'examples/**'
+      - '.gitignore'
+      - '*.rst'
+      - '*.md'
+      - '.github/workflows/*.yml'
+      # re-include current file to not be excluded
+      - '!.github/workflows/build-debian-multiarch.yml'
+
+  pull_request:
+    branches: main
+    paths-ignore:
+      - 'docs/**'
+      - 'examples/**'
+      - '.gitignore'
+      - '*.rst'
+      - '*.md'
+      - '.github/workflows/*.yml'
+      # re-include current file to not be excluded
+      - '!.github/workflows/build-debian-multiarch.yml'
+
+jobs:
+  build-multiarch:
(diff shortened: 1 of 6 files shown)


ankith26:
This PR is done from my end. This PR depends on #3498 (from which it already has a cherry-picked commit for testing purposes) and #3494 being merged first.

I've also made two big endian fixes in this PR, skipped two image tests on big endian (if using an old SDL_image) and made buildconfig improvements so we handle exotic architectures generically

illume:
Adding to the why: Debian supports these architectures and it's easier to get updated versions of pygame on there if there's no test failures on those platforms (which is different to no bugs... skip/known failure are our friends).

illume:
I'm considering this one for 2.1.3 since it fixes things for s390x and ppc64le (and other architectures on debian perhaps?).

illume (approved):
thanks 👍

ankith26:
I'm fine with this going in for 2.1.3

PS: There's an optional CI build timing optimisation that can be done here, if github actions has permissions to upload packages, it will cache the docker images with dependencies in there (SDL stuff and python stuff). You can take the final call on this and change permissions accordingly (RN the permissions are disabled and multiarch builds take 14-18 mins, with the permissions enabled we will gain some time but probably not too much)

illume:
> I'm fine with this going in for 2.1.3

I think it's worth the risk if it makes things go more smoothly with a Debian package update.

*fingers crossed!*

illume:
> PS: There's an optional CI build timing optimisation that can be done here, if github actions has permissions to upload packages, it will cache the docker images with dependencies in there (SDL stuff and python stuff). You can take the final call on this and change permissions accordingly (RN the permissions are disabled and multiarch builds take 14-18 mins, with the permissions enabled we will gain some time but probably not too much)

Yeah, let's discuss in an issue? I don't understand the implications of opening up package upload permissions. I know there are docker caching facilities with gh actions though.

Issue #3107, opened 2022-03-31
Resize doesn't work after fullscreen


MainBreezy:
Windows - Python 3.10.2 :
- **Pygame 2.1.2

**Current behavior:**
After exiting from fullscreen the ability to resize the window disappears

**Expected behavior:**
I wanted the window to be able to be resized after exiting fullscreen

**Screenshots**

Expected:
![image](https://user-images.githubusercontent.com/102602813/161113782-6e48a3d9-be96-427b-b7ff-aebff489fd81.png)

Current:
![image](https://user-images.githubusercontent.com/102602813/161113934-d0e4fb4a-7f8b-4376-b62a-3ad53252273f.png)

**Steps to reproduce:**
Code below

**Test code**

If possible add a simple test program that shows the problem described in this report.
```
import pygame
import tkinter as tk
import sys

pygame.init()

root = tk.Tk()
screenWidth = root.winfo_screenwidth()
screenHeight = root.winfo_screenheight()
mainClock = pygame.time.Clock()
pygameIcon = pygame.image.load(r"C:/Users/Chris/OneDrive/Desktop/MyPython/Sprites/Reeses Puffs.png")

fullscreen = False
running = True

pygame.display.set_caption("Test Game")
pygame.display.set_icon(pygameIcon)
screen = pygame.display.set_mode((screenWidth-100, screenHeight-100), pygame.RESIZABLE)


def menu():
    global running
    global screen
    global fullscreen

    while running:
        for event in pygame.event.get():
            if event.type == pygame.VIDEORESIZE:
                if not fullscreen:
                    screen = pygame.display.set_mode((event.w, event.h), pygame.RESIZABLE)
            if event.type == pygame.KEYDOWN:
                if event.key == pygame.K_F11:
                    fullscreen = not fullscreen
                    if fullscreen:
                        pygame.display.set_mode((screen.get_width(), screen.get_height()), pygame.FULLSCREEN)
                    else:
                        pygame.display.set_mode((screen.get_width(), screen.get_height()), pygame.RESIZABLE)
                elif event.key == pygame.K_ESCAPE:
                    quitAll()
            if event.type == pygame.QUIT:
                quitAll()

        pygame.display.flip()
        pygame.display.update()


def quitAll():
    global running
    running = False


menu()
```

MyreMylar:
Confirmed that this occurs on Windows, in pretty recent builds. Made a slightly clearer reproduction case:

```
import pygame


class App:
    def __init__(self):
[...]

Issue #2698, opened 2021-08-26
Window loses frame toggling back and forth between fullscreen and windowed


frankpsy:
**Environment:**

You can get some of this info from the text that pops up in the console when you run a pygame program.

Windows 10
Python 3.9.6
SDL 2.0.14
pygame 2.0.1

Window loses frame (window caption, min/max/close/resize controls) after switching back from windowed to fullscreen one or two times. Originally wrote this code using pygame.display.set_mode(realScreenSize, pygame.FULLSCREEN) but would lose the window frame switching back to windowed just every time, tried using pygame.display.toggle_fullscreen() like below and it would keep the frame on the first switch back but on subsequent toggles it loses the window frame. 

**Expected behavior:**

Expected behavior is that it should always keep the frame in windowed mode unless pygame.NOFRAME is passed, which is not used once in the code below.

**Screenshots**

**Steps to reproduce:**

Use F to toggle back and forth between fullscreen and windowed, will lose window frame after 2-3 times.

```
#!/usr/bin/env python3
import os, sys
import random
import pygame
import pygame.gfxdraw

screenSize = 640, 360
realScreenSize = 1280, 720
prevRealScreenSize = 0, 0
fullscreen = False

pygame.init()

realScreen = pygame.display.set_mode(realScreenSize, pygame.RESIZABLE)
pygame.display.set_caption("Random Boxes")
screen = pygame.Surface(screenSize)
random.seed()

while True:
    z = 0
    for x in range(0, screenSize[0]):
        for y in range(0, screenSize[1]):
            if z % 2 == 0:
                screen.set_at((x,y), pygame.Color(128,64,0))
            else:
                screen.set_at((x,y), pygame.Color(0,0,0))
            z = z + 1
        z = z + 1
    for i in range(0, 20):
        x = random.randint(0, screenSize[0]-1)
        y = random.randint(0, screenSize[1]-1)
        xx = random.randint(0, screenSize[0]-x)
        yy = random.randint(0, screenSize[1]-y)
        r = random.randint(0, 255)
        g = random.randint(0, 255)
        b = random.randint(0, 255)
        pygame.gfxdraw.box(screen, pygame.Rect(x, y, xx, yy), (r, g, b))
    pygame.transform.scale(screen, realScreenSize, realScreen)
    pygame.display.update()
    for i in range(0,200):
        for event in pygame.event.get():
            if event.type == pygame.VIDEORESIZE:
[...]

PR #3497, opened 2022-10-12
Re-order display fullscreen handling and ensure toggle_fullscreen works for maximized window

### LINKED ISSUE #2380: display.toggle_fullscreen does not work when toggling a maximized window to fullscreen

### LINKED ISSUE #2711: Pygame 2.0.2.dev2 -> NOFRAME and RESIZABLE flags not working when switching from fullscreen back to windowed mode
**Environment:**

- **Operating system** Windows 10
- **Python version** Python 3.9.5
- **SDL version** SDL 2.0.16
- **PyGame version** Pygame 2.0.2.dev2

**Current behavior:**

The NOFRAME and RESIZABLE flags do not work when switching from fullscreen mode back to windowed mode, and a non-resizable 'framed' window is set, but the SCALED flag works fine for all cases. This issue can be overriden when the window is set at least 2 times after reverting from fullscreen mode. 

**Expected behavior:**

When window is switched to windowed mode from fullscreen to windowed mode with the NOFRAME or RESIZABLE flags (or both of them), then the window should have no title bar, and be resizable respectively.

**Screenshots**

Before fullscreen toggle:
image

In fullscreen mode:
image

Switching to windowed mode after fullscreen mode with NOFRAME flag:
image

**Steps to reproduce:**

1. open command prompt or powershell
2. pip install pygame==2.0.2.dev2
3. press any key twice after running the example code

**Test code**

```
import pygame

pygame.init()

fullscreen = False
size = (1280, 720)

full_flags = pygame.FULLSCREEN | pygame.SCALED
normal_flags = pygame.NOFRAME | pygame.RESIZABLE | pygame.SCALED


def toggle_fullscreen(f):
    if f:
        return pygame.display.set_mode(size, full_flags)
    # pygame.display.set_mode(size, normal_flags)  # uncomment this to see issue being fixed as a workaround
    return pygame.display.set_mode(size, normal_flags)


screen = pygame.display.set_mode(size, normal_flags)

clock = pygame.time.Clock()

while True:
    screen.fill("purple")

        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit

            fullscreen = not fullscreen
            screen = toggle_fullscreen(fullscreen)

    pygame.display.flip()
    clock.tick(144)  # framerate

```

Temmie3754:
This PR resolves #2711 and #3107 by changing the order that changing the fullscreen flag is handled so that it sets the window to be not fullscreen before handling the borderless and resizeable flag as when in fullscreen SDL ignores these changes.
It also partly fixes #2380 as it succeeds at creating a fullscreen window and when exiting retains the resize flag of the window but it still results in a warning message as it has to recreate the window. The toggle function could likely still be improved in this regard but for now this PR fixes the main issues found with it.


ankith26 (approved):
Wow, I'm quite impressed!
This seemingly simple fix fixes 3 really complex-looking issues! 🥳 🎉 
I could reproduce all three issues being fixed (the first 2 were spot on, the third `toggle_fullscreen` one was a bit problematic and I came across other minor and unrelated bugs, but the main issue in there is probably fixed)

We have other issues to be tracking `toggle_fullscreen` problems, and this PR does good progress already, thanks a ton! 🍰 🍕 

(due to the nature of what this is fixing, I'd like if we could get more than 3 reviewers, and from my side I'm supporting this PR to quickly make it to the next dev release for wider testing)

src_c/display.c
```diff
@@ -1112,6 +1112,12 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                  this invalidates the display surface*/
                 SDL_SetWindowTitle(win, title);
                 SDL_SetWindowSize(win, w_1, h_1);
+                if (0 !=
+                    SDL_SetWindowFullscreen(
+                        win, sdl_flags & (SDL_WINDOW_FULLSCREEN |
+                                          SDL_WINDOW_FULLSCREEN_DESKTOP))) {
+                    return RAISE(pgExc_SDLError, SDL_GetError());
+                }
```
ankith26, line 1120: To avoid potential future regressions, a comment above this explaining it has to come before setting other flags like resize/bordered/hidden etc would be nice

Temmie3754:
I've updated it so that it now has the correct formatting and contains a comment for that specified section. I also wrote some further changes for the toggle_fullscreen so that it preserves these flags and the position as well. Also related issue #2698 is fixed by this PR. These changes will require further review to check that no issues are brought about by them.

ankith26 (commented):
I think the `fullscreen_backup_x`/`fullscreen_backup_y` stuff are redundant now. Those were added as part of #2460 to workaround a similar issue (and I suspect most of #2460 could now be reverted), because this fix of handling fullscreen first and then doing other display stuff also fixes the issue that PR was fixing

Temmie3754:
Unfortunately the backup positions are still required since those get modified when going into fullscreen as it changes the position to (0, 0) when entering fullscreen. Fullscreen was already handled before position anyway so it isn't affected by this change. This does however implement the same fix #2460 used for toggle_fullscreen now to enable position saving since that was left out when the PR was made.

src_c/display.c
```diff
@@ -2096,7 +2100,7 @@ pg_toggle_fullscreen(PyObject *self, PyObject *_null)
     if (!win)
         return RAISE(pgExc_SDLError, "No open window");
 
-    flags = SDL_GetWindowFlags(win) & SDL_WINDOW_FULLSCREEN_DESKTOP;
+    flags = SDL_GetWindowFlags(win);
     /* SDL_WINDOW_FULLSCREEN_DESKTOP includes SDL_WINDOW_FULLSCREEN */
```
ankith26, line 2104: Now that the line above is changed, the comment just below it seems a bit out of place. Should it be moved elsewhere?

src_c/display.c
```diff
@@ -2273,7 +2278,8 @@ pg_toggle_fullscreen(PyObject *self, PyObject *_null)
                              1) != 0) {
                 return NULL;
             }
-            win = SDL_CreateWindow(state->title, wx, wy, w, h, 0);
+            flags &= ~SDL_WINDOW_FULLSCREEN_DESKTOP;
+            win = SDL_CreateWindow(state->title, wx, wy, w, h, flags);
             if (win == NULL) {
                 return RAISE(pgExc_SDLError, SDL_GetError());
             }
```
ankith26, line 2285: The comment in this block of code says it's a hack for SDL 2.0.8 (which is a rather old version)... not sure whether it's relevant to this day? Would need someone running pygame on linux with x11 to check though

MyreMylar (approved):
This fixes the two main linked issues for me, with the slight caveat that one of them also seems to want a resize able window without a frame which isn't a thing in the Windows platform where you resize and move windows by dragging the frame around.

Temmie3754:
Alright fixed that comment so it's now where the flag is actually mentioned, looks like everything has been resolved.

illume (approved):
Thanks. Works really well.

PR #3501, opened 2022-10-16
Warn when SIMD is usable but not compiled

ankith26:
Commit was initially a part of #3498 but is now separated for better clarity.

That PR is fixing the compile error (and is therefore, more urgent than this PR)

This PR is just adding warnings for when there's runtime support for X (can be neon/sse2/avx2) but was skipped at compile time


MyreMylar (approved):
LGTM 👍 (I reviewed this when looking at the original PR)

Starbuck5 (approved):
This looks good, thanks!

It would be a pretty rare situation for this to actually fire, but maybe it will help someone in the future.

Issue #3408, opened 2022-08-20
2.1.3.dev6 discussion


illume:
Some things before 2.1.3...

- [x] I'm reviewing/editing/labelling from the bottom/newest (mostly).
- [x] I'd like we keep the SDL* lib versions the same on different platforms if we can. See https://github.com/pygame/pygame/pull/3042#issuecomment-1229426713 I added an issue for SDL 2.24.0 update here: https://github.com/pygame/pygame/issues/3430 since I think it's easier to update the manylinux scripts than revert the windows changes.
- [x] mark all new APIs as experimental in the docs? For a release or two until more people try them and perhaps uncover issues? [New API closed PRs](https://github.com/pygame/pygame/pulls?q=is%3Apr+is%3Aclosed+label%3A%22New+API%22)
- [x] Get settings path function in a context module needs a rethink: https://github.com/pygame/pygame/pull/3100 Will remove it before 2.3.dev6 is released. Also here: https://github.com/pygame/pygame/pull/3038
- [ ]  Fix issues with newbie guide changes: https://github.com/pygame/pygame/pull/3118  The parts about dirty rects being useless are incorrect.
- [x] there were some SDL_ functions defined, which breaks forwards compatibility. https://github.com/pygame/pygame/issues/3417 should be fixed before 2.1.3, but not the dev release.

Some things before 2.1.3.dev6...

illume:
Need to find out if this API change for math.clamp is acceptable: https://github.com/pygame/pygame/pull/3361 and if not, perhaps revert it.

illume:
Get settings path function in a context module needs a rethink: https://github.com/pygame/pygame/pull/3100 Will remove it before 2.3.dev6 is released.

illume:
Hrmm. With the module labels renamed to "Submodule: display" rather than "display" older links and queries are broken.
https://github.com/pygame/pygame/labels/display doesn't work for example.

The URLs are kind of weird now too:
- `https://github.com/pygame/pygame/labels/Submodule%3A%20display`
- `https://github.com/pygame/pygame/labels/display`

Probably they need to be renamed, and perhaps the description (rather than name) can stay as "Submodule: x".

illume:
This gcc specific error should probably block the release, because it breaks clang at least (and other compilers potentially). https://github.com/pygame/pygame/pull/3381

illume:
[...]

Issue #3443, opened 2022-09-11
feature flags import from pygame.future feature?


illume:
There was a discussion on this started here: https://github.com/pygame/pygame/pull/3396#discussion_r950738414

illume:
It's a good point that zoldalma999 made: people just using the function may not notice it's experimental, because they don't read the docs... they just use the function.

For this release I'll try something fairly low effort, because we want to get the release out soonish.

- Mark all new APIs as experimental. (luckily, they are labeled in the PRs)
- For each new API, I'll reopen the original issue for feedback... or create an issue if there wasn't one, linking to the PR.
- Link to the issue in the docs, and in the release notes.

This way:
- people can leave feedback
- the people who contributed to the new API will be able to get that feedback

ankith26:
> There was another good comment on this topic somewhere else... but I forgot to link it and now can't find it.

Are you referring to [this](https://github.com/pygame/pygame/pull/3100#issuecomment-1221273545)?

About the `pygame.experimental` namespace, it works well when we are adding new experimental modules and functions, but not when we are adding new methods to pre-existing classes.

Anyways, I agree with all the points you mentioned above, and ATM our main problem is getting feedback from a wider range of users. The best way to get real user feedback is to actually release new API in the wilds while making their experimental nature clear enough. My idea for doing this is slightly different from feature flags imports, it is using warnings. We would have a new `pygame.ExperimentalAPIWarning` (or something of that sort) which would be a subclass of python `Warning`.

When we mark an API as experimental, we would also have to raise a descriptive warning to let users know it. Users may still choose to silence it, but that needs more conscious effort by the user and this way we are sure that the user is almost guaranteed to see the message atleast once.

Starbuck5:
I don't think we need to implement such solutions for this problem.

The best way to manage API mistakes, in my opinion, would be on a case by case heuristic basis, with frequent releases.

Here are a few examples

I added a function `mouse.set_system_cursor`, and then superseded that work within a small amount of time (2 months?) with a more robust `mouse.set_cursor` API. That new API was promptly released, and the `set_system_cursor` function was deprecated. The biggest obstacle to the success of this maneuver was that the documentation wasn't updated for about a year, so people were greeted with deprecation warnings with no alternate path in the documentation.

[...]

PR #3508, opened 2022-10-22
Remove pygame.context for later renaming

illume:
For https://github.com/pygame/pygame/issues/3408


ankith26:
It's been discussed before in a few previous PRs/issues, but since this PR is already implementing a revert, I'll summerise a few things here.

Personally I will agree that at some point we should change the name of this module (though I know Starbuck5 doesn't agree), but I don't think reverting the whole change is the best way. Apart from adding unneeded delays, it also means that this new API does not actually go in any form to real users (outside of the regular contributors group), whose feedback we very much want. I just don't want this to be stuck with no good and objective way to decide on a good name, and be in a separate branch (which almost all pygame users will never know about, so we will have to again make do with the feedback of only regular contributors) for a long time and ultimately get abandoned. The more user feedback we get, the better, and we can decide faster without much back-and-forth among ourselves. 

Yes, the experimental API route does add its own delays, issues and such, but I think it's a good compromise between either reverting the change fully, or directly adding the change as permanent public API

If we release this as experimental API (and there are many ideas we discussed for this in #3443), there need to be a few more changes for extra clarity of this.
First step is to remove `pygame.context` auto-import. Then we can prefix it with an underscore maybe, or put it under a dedicated experimental namespace. Raise a warning everytime someone does `import pygame.experimental.context`/`import pygame._context` and request for feedback by linking a relevant github issue. I personally think this is the best way of getting input directly from people who need to use this, better than a contributors-only discussion and also better than a social media-only poll (though we could optionally do this as an additional step after doing the "real users RFC issue" first, and linking this issue from social media too for added visibility)

If there are any reasons to not be doing what I suggest, I'm very much open to major or minor tweaks to the proposal. If we do decide to go the revert-full-change way, I'd hope we can decide on a more concrete real-user-feedback plan before merging this PR :)

illume:
There's not an agreement at this stage, so a revert is better than leaving it in. There was plenty of time for a renaming discussion to happen for this release but it wasn't done.

The non-controversial place to introduce this function was under _sdl2.

PR #3510, opened 2022-10-23
Deprecating old messy scrap functions and implementing shiny, new SDL-based scrap function

### LINKED ISSUE #3402: New scrap functions
The current scrap API is in place for historical reasons. Copying and pasting images and other media besides text uses bespoke C code that has not been maintained in over a decade, and is no longer even built with pygame, mostly.

We should consolidate functionality with what SDL provides, and is what most people want to do: copy and paste text.

And we should have an API that doesn't require encoding or decoding bytes, text in and out as strings.

These new APIs should live alongside the current scrap functions, which should be deprecated.
ALTERNATIVELY
We could put new copy/paste APIs into the new `pygame.context` module, and deprecate the entire scrap module.

See: https://wiki.libsdl.org/CategoryClipboard

oddbookworm:
Closes #3402. I added deprecations to all of the old, messy scrap stuff and implemented newer scrap functions based on what SDL2 currently has available (https://wiki.libsdl.org/CategoryClipboard). I know tests still have to be written, but I would like that to be one of the last things I do, so I'm looking for feedback.


Starbuck5:
All the functions in scrap.c call into functions defined in the old platform specific ones, so the deprecation warnings should only need to be put in scrap.c. 

Mac seems to always use scrap_sdl2.c, so that is true for it as well, I believe, despite the ifdefs.

oddbookworm:
@Starbuck5 I addressed your comments and made the necessary corrections

Starbuck5:
Looks like you need the changes from #3529

You could rebase your branch or merge main back into this branch.

MyreMylar (commented):
This seems like a good clean up to me. Code looks good and it will work on all SDL platforms as opposed to the patchy, platform specific support from the old days. I will probably transition to using this in Pygame GUI in the future.

I believe the old scrap functions haven't been working correctly for a while now? Meanwhile looking at the SDL clipboard header it has just recently got a new function so perhaps it will be the future of clipboard interaction for pygame.

I'll mark it for a good testing this weekend before final approval, as I want to check the old scrap functionality versus the new on Windows.

src_c/scrap.c
```diff
@@ -355,6 +408,76 @@ _scrap_set_mode(PyObject *self, PyObject *args)
...
+
+    if (!PyArg_ParseTuple(args, "s", &text)) {
+        return NULL;
+    }
+
+    if (SDL_SetClipboardText(text)) {
+        const char *err = SDL_GetError();
+        if (err) {
+            return RAISE(pgExc_SDLError, "video mode not initialized");
+        }
```
Starbuck5, line 457: What? Why not use SDL_GetError's error message in the raise?

src_c/scrap.c
```diff
@@ -355,6 +408,76 @@ _scrap_set_mode(PyObject *self, PyObject *args)
...
+    if (!text) {
+        const char *err = SDL_GetError();
+        if (err) {
+            SDL_free(text);
+            PyErr_SetString(PyExc_RuntimeError, err);
+            return NULL;
+        }
+    }
+
+    PyObject *returnValue = Py_BuildValue("s", text);
```
Starbuck5, line 431: Nitpick- but this seems like a strange way of doing this.

This would work iirc: https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FromString

src_c/scrap.c
```diff
@@ -355,6 +408,76 @@ _scrap_set_mode(PyObject *self, PyObject *args)
...
+{
+    char *text = SDL_GetClipboardText();
+
+    if (!text) {
+        const char *err = SDL_GetError();
+        if (err) {
+            SDL_free(text);
+            PyErr_SetString(PyExc_RuntimeError, err);
+            return NULL;
+        }
```
Starbuck5, line 428: Hmmm-
so SDL_GetClipboardText says it sends back an empty string on failure, so is that the same as returning NULL? I don't think so? The char* would be a valid pointer to a 1 long array just containing NULL?

And then I have the same question about the point of `if (err)` like before.

Why is it a runtime error instead of a pygame.SDLerror?

src_c/scrap.c
```diff
@@ -355,6 +408,77 @@ _scrap_set_mode(PyObject *self, PyObject *args)
...
+    char *text;
+
+    if (!PyArg_ParseTuple(args, "s", &text)) {
+        return NULL;
+    }
+
+    if (SDL_SetClipboardText(text)) {
+        const char *err = SDL_GetError();
+        if (err) {
+            return RAISE(pgExc_SDLError, err);
```
Starbuck5, line 457: These three lines just need to be `return RAISE(pgExc_SDLError, SDL_GetError());`

src_c/scrap.c
```diff
@@ -27,6 +27,8 @@
 
 #include "SDL_syswm.h"
 
+#include "SDL_clipboard.h"
+
```
Starbuck5, line 31: Is this needed? SDL.h is included somewhere down the tree of pygame.h, and SDL.h seems like it would include SDL_clipboard.h.

src_c/scrap.c
```diff
@@ -355,6 +408,77 @@ _scrap_set_mode(PyObject *self, PyObject *args)
 
 #endif /* !defined(MAC_SCRAP) */
 
+/**
+ * @brief Fetches a python string from the SDL clipboard. If
+ *        there is nothing in the clipboard, it will return empty
+ *
+ * @return PyObject*
+ */
```
Starbuck5, line 416: For me this method raises a pygame.error with no message if the clipboard is empty.

oddbookworm: Fixed this by checking if `err` is empty

MyreMylar (approved):
OK, I've given the old scrap and this new scrap a good tinker with:

**OLD SCRAP**

- It does work on Windows as described.
- How it works is a bit python 2 era. E.g. text data is copied and returned as bytes rather than python strings. This is just annoying and messy.
- You can apparently copy and paste wav data but only on windows - but I'm not quite sure how you would do this? 
- The main useful thing we are losing on windows from old scrap is the ability to copy directly from MS paint (CTRL +C with an image open in paint) and paste directly into pygame and then use it immediately (after a quick `pygame.image.load(paste_here, ".bmp")`) as a Surface.
- None of this non-text stuff works on macs.
- The bmp pasting may also work on X11, but probably doesn't work on wayland from what I can see.
- The initialisation of scrap is weird as heck because you have to do it after making a window.
- There is also lots of busy work functions in the old API, as well as the initialisation, multiple ways to find the type of data in the clipboard and some x11 only stuff - I expect to handle the extra middle mouse 'primary selection' clipboard that exists only on Linux. 

**NEW SCRAP**

- Simple three function API. Same on all platforms.
- Returns python strings which is what you want.
- No support for bmp data.

As far as I can see the new scrap is better in all ways except the ability to paste in bmp data on windows (and probably X11) which is probably occasionally useful.

On balance I think we should make this change as it is leveraging SDLs improved text clipboard support - and then investigate just supporting bmp/image data pasting in pygame as sadly this is not in SDL clipboard.

Issue #3483, opened 2022-10-07
tp_name of pygame types should be fully qualified (for better pydoc support)


miezebieze:
ohay
i'm used to use `pydoc` before searching the web and today i wanted to try my hand at sounds for the first time. after finding pygame.mixer i was confused what this `Sound` object was that the docstring was blabbing on about.

after consulting the online docs, i found out that for some reason [see title]. just to make sure, i tried `pydoc -k Sound` (does about the same as executing `help (Sound)` in the interpreter), which couldn't find it either.

`pydoc pygame.mixer.Sound` works but without knowing where to look it's no help.

Starbuck5:
I don't understand how we would resolve this?

pydoc / help looks at the object's members to generate the docs?

```
>>> dir(pygame.mixer)
['Channel', 'ChannelType', 'Sound', 'SoundType', '_PYGAME_C_API', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__spec__', '_internal_mod_init', 'fadeout', 'find_channel', 'get_busy', 'get_init', 'get_num_channels', 'get_sdl_mixer_version', 'init', 'music', 'pause', 'pre_init', 'quit', 'set_num_channels', 'set_reserved', 'stop', 'unpause']
```

The Sound class is a member of the pygame.mixer module object, so I really have no idea why it wouldn't be picked up.

miezebieze:
and there's more!
`ChannelType` and `SoundType` are two more that i could identify without even consulting pydoc.

in `pygame`, `Vector2` and `Vector3` don't show up, but `Color`, which is imported in the same way, does. at least the Vectors can be found in another module, although `pydoc -k Vector` doesn't find them either.

Starbuck5:
> ChannelType and SoundType are two more that i could identify without even consulting pydoc.

What?
You could identify them without pydoc?
As in you couldn't find them with pydoc?

Again, I'm not sure how your points could actually be resolved. Would you like to submit a pull request?

miezebieze:
> What? You could identify them without pydoc? As in you couldn't find them with pydoc?
i meant that because i just looked at it (as stated in my previous post) i knew it for a fact. of course i checked if it really was the case.

i know the basics of C but am unable to extract meaning from the source so i don't know how to resolve the problem either.
i was hoping someone else would know what's up and fix it.

zoldalma999:
[...]

PR #3515, opened 2022-10-24
Fixed inconsistency in `tp_name`s, changed all to `pygame.module.Type`

Matiiss:
May fix https://github.com/pygame/pygame/issues/3483


ankith26 (commented):
This is definitely improving things, thanks!

But you didn't cover all places, there's `freetype.Font`, some bufferproxy and newbuffer code too that needs changes like this.

Matiiss:
@ankith26 About those some places, they seem to have some constants in place, for example, in `bufferproxy.c` there's `PROXY_TYPE_FULLNAME`(https://github.com/pygame/pygame/blob/main/src_c/bufferproxy.c#L528), would I just replace that with `"pygame.bufferproxy.BufferProxy"` or rather change the definition of `PROXY_TYPE_FULLNAME` (to the previously mentioned string (guess it's an array in C)): https://github.com/pygame/pygame/blob/main/src_c/bufferproxy.c#L48?

Such constants are also in place for `key.c`, `newbuffer.c` and `_freetype.c`
image

ankith26:
I think for consistency, it's better if we use strings directly everywhere. There's no need to have a macro constant that is only used in one place

Matiiss:
So, should I remove all `FULLNAME` constants (their definitions) from the code too?

Matiiss:
Should I use `"pygame._freetype.Font"` or `"pygame.freetype.Font"` because it's in `_freetype.c` but from a practical standpoint I don't think anyone's going to search for that module using an underscore

ankith26:
> So, should I remove all `FULLNAME` constants (their definitions) from the code too?

If nothing else needs it, I'd say yes; go ahead.

> Should I use "pygame._freetype.Font" or "pygame.freetype.Font" because it's in _freetype.c but from a practical standpoint I don't think anyone's going to search for that module using an underscore

Hmm, that's a tough one and TBH I'm not sure either. This is not just limited to freetype, but also things like `Surface` (where most people generally use the `pygame.Surface` alias but the actual implementation is in `pygame.surface.Surface`)

Matiiss:
Well, `pydoc` seems able to find the docs for `Surface` in both cases: `pydoc pygame.Surface` and `pygame.surface.Surface`

Matiiss:
And I will need to revert that commit, right?

Starbuck5:
The build failure is fixed on main now, so you can pull that change in by rebasing or merging main back into your branch.

MyreMylar:
Looks like the type checking CI is failing here:

```
error: pygame.event.EventType is not present in stub
Stub: in file /home/runner/work/pygame/pygame/buildconfig/stubs/pygame/event.pyi
MISSING
Runtime:
<class 'pygame.event.Event'>

error: pygame.rect.RectType is not present in stub
Stub: in file /home/runner/work/pygame/pygame/buildconfig/stubs/pygame/rect.pyi
MISSING
Runtime:
<class 'pygame.rect.Rect'>

Found 2 errors (checked [21](https://github.com/pygame/pygame/actions/runs/3359828731/jobs/5568287362#step:6:22)1 modules)
```

I assume it is related to these type name changes.

Matiiss:
@MyreMylar I noticed it failing but wasn't (still am not) sure what to do exactly, since it didn't fail anywhere else. So, do I just also change the type stub files to include the full names?

MyreMylar:
> @MyreMylar I noticed it failing but wasn't (still am not) sure what to do exactly, since it didn't fail anywhere else. So, do I just also change the type stub files to include the full names? At least in those two files.

That particular CI build is specifically for checking the type stubs so it always tends to fail on its own when something is messed up with the stubs.

If you look at the bottom of `surface.pyi` here:

https://github.com/pygame/pygame/blob/9532f679c8563c952f009d8be4f7fc0483a7cc35/buildconfig/stubs/pygame/surface.pyi#L148

I believe that is the approach you'll need to take for Event and Rect. Could be wrong though.

MyreMylar (approved):
LGTM 👍

Starbuck5 (approved):

Starbuck5:
Thanks Matt!

PR #3525, opened 2022-10-28
Fix segmentation fault on uninitialized Surface

### LINKED ISSUE #3519: Segmentation fault when converting an uninitialized surface with argument
Environment:
- Windows 10
- Python 3.10.6
- `pygame` 2.1.3.dev9 (could reproduce on 2.1.2 and 2.1.0 too)

MRE:
```py
import pygame


pygame.display.set_mode((1, 1))

pygame.Surface.__new__(pygame.Surface).convert(24)
```

It does not really matter what the argument is as long as it's a python object (except for `pygame.Surface`), it raises this Segmentation fault:
```none
pygame 2.1.3.dev9 (SDL 2.0.22, Python 3.10.6)
Fatal Python error: pygame_parachute: (pygame parachute) Segmentation Fault

Current thread 0x000017c8 (most recent call first):
  File "C:\Users\Matiiss\PycharmProjects\Something\for_so.py", line 6 in <module>

Extension modules: pygame.base, pygame.constants, pygame.rect, pygame.rwobject, pygame.surflock, pygame.bufferproxy, pygame.math, pygame.surface, pygame.display, pygame.draw, pygame.event, pygame.imageext, pygame.image, pygame.joystick, pygame.key, pygame.mouse, pygame.time, pygame.mask, pygame.pixelcopy, pygame.transform, pygame.font, pygame.mixer_music, pygame.mixer, pygame.scrap, numpy.core._multiarray_umath, numpy.core._multiarray_tests, numpy.linalg._umath_linalg, numpy.fft._pocketfft_internal, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._mt19937, numpy.random.mtrand, numpy.random._philox, numpy.random._pcg64, numpy.random._sfc64, numpy.random._generator, win32api, _win32sysloader, pygame._freetype (total: 40)
```

The issue stems from this line https://github.com/pygame/pygame/blob/main/src_c/surface.c#L1449 where all it does is simply check if there is any argument given and then it checks if it is a surface and if it fails that, it tries to `memcpy` `format` which doesn't exist if the surface hasn't been initialized (at least how I understand it).

If it's nothing too complicated I'm down to fix this, I'd just like to know what the suggested way to do so would be.

Also I suppose the expected output should be the same as when doing it without an argument (or when it's a surface):
```py
import pygame


pygame.display.set_mode((1, 1))

pygame.Surface.__new__(pygame.Surface).convert()
```
```
pygame 2.1.3.dev9 (SDL 2.0.22, Python 3.10.6)
  File "C:\Users\Matiiss\PycharmProjects\Something\for_so.py", line 6, in <module>
    pygame.Surface.__new__(pygame.Surface).convert()
pygame.error: Parameter 'surface' is invalid
```

MightyJosip:
Fix #3519. Some surface methods didn't check if the Surface was actually initialized so this PR should fix that.
Methods changed:
- lock()
- unlock()
- convert()
- mustlock()
- get_locks()
- get_locked()
- get_blendmode()

Some of those methods also raised segmentation fault, and others worked. Now all of them raise pygame error "display Surface quit" like other methods (for example blit)

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -969,6 +969,8 @@ surf_unmap_rgb(PyObject *self, PyObject *arg)
 static PyObject *
 surf_lock(PyObject *self, PyObject *_null)
 {
+    if (!pgSurface_AsSurface(self))
+        return RAISE(pgExc_SDLError, "display Surface quit");
     if (!pgSurface_Lock((pgSurfaceObject *)self))
         return NULL;
     Py_RETURN_NONE;
@@ -977,6 +979,8 @@ surf_lock(PyObject *self, PyObject *_null)
 static PyObject *
 surf_unlock(PyObject *self, PyObject *_null)
 {
+    if (!pgSurface_AsSurface(self))
+        return RAISE(pgExc_SDLError, "display Surface quit");
     pgSurface_Unlock((pgSurfaceObject *)self);
     Py_RETURN_NONE;
 }
@@ -985,6 +989,8 @@ static PyObject *
 surf_mustlock(PyObject *self, PyObject *_null)
 {
     SDL_Surface *surf = pgSurface_AsSurface(self);
+    if (!surf)
+        return RAISE(pgExc_SDLError, "display Surface quit");
     return PyBool_FromLong(SDL_MUSTLOCK(surf) ||
                            ((pgSurfaceObject *)self)->subsurface);
 }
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
LGTM 👍

Starbuck5:
Two questions-

Is display surface quit a good error message? Shouldn't it say something about the surface being uninitialized instead?

Do you think it would be a good to put this check in a macro so it can be pasted wherever needed? Like `VIDEO_INIT_CHECK`, for example.

Starbuck5:
I reopened this because I didn't understand why it had been closed.

PR #3424, opened 2022-08-22
mouse.get_pos(), mouse.get_rel() optimizations

itzpr3d4t0r:
## Info
Analogue to #3401, this one implements the same optimization but for `mouse.get_pos()` and `mouse.get_rel()`, bringing a 30-40% improvement overall.
## Performance
Here are performance results and some quick graphs:

**OLD**
```
mean: 11.233, stdev: 0.549
mean: 9.143, stdev: 0.276
------
mean: 10.354, stdev: 0.218
mean: 8.479, stdev: 0.227
```
myplot
___

**NEW**
```
mean: 8.893, stdev: 0.581
mean: 7.033, stdev: 0.376
------
mean: 7.941, stdev: 0.231
mean: 6.13, stdev: 0.228
```

myplot

### Code

```Python
from statistics import pstdev, fmean
from timeit import repeat

import pygame as pg
from pygame import mouse

from plotting import create_plot


def adjust(x: float) -> float:
    # adjust to milliseconds and round
    return round(x * 1000, 3)


win = pg.display.set_mode((500, 500))

GLOB = {"pg": pg, "mouse": mouse}

REPEATS = 200
NUM = 100000

data = repeat("pg.mouse.get_pos()", globals=GLOB, number=NUM, repeat=REPEATS)
data_1 = repeat("mouse.get_pos()", globals=GLOB, number=NUM, repeat=REPEATS)
data2 = repeat("pg.mouse.get_rel()", globals=GLOB, number=NUM, repeat=REPEATS)
data2_1 = repeat("mouse.get_rel()", globals=GLOB, number=NUM, repeat=REPEATS)

print(f"mean: {adjust(fmean(data))}, stdev: {adjust(pstdev(data))}")
print(f"mean: {adjust(fmean(data_1))}, stdev: {adjust(pstdev(data_1))}")
print("------")
print(f"mean: {adjust(fmean(data2))}, stdev: {adjust(pstdev(data2))}")
print(f"mean: {adjust(fmean(data2_1))}, stdev: {adjust(pstdev(data2_1))}")
```

PR #3530, opened 2022-11-01
Move pg_tuple_from_values_int into header, rename & add triple variant

MyreMylar:
Related to the suggestion in #3424 by @Starbuck5 to reduce code duplication by having copies of this function each time it gets used.

As far as I can tell there is no obvious reason _not_ to have inline functions in header files with Pygame 3. The existing #define style 'functions' in `include/_pygame.h` appear to have been designed that way for something called `CObject`s in Python 2 (according to the header comment in the file) which were then removed in Python 3.

Happy to hear wisdom otherwise, or create a new file instead if we think that is better, but the existing #define macro functions in this header are the most similar to the purpose of this function. Also, this header is already included everywhere.

**Testing work:** does this change the performance in any way versus having it directly in `rect.c`. It shouldn't but hey, compilers are weird.

### DIFF
--- a/src_c/include/_pygame.h
+++ b/src_c/include/_pygame.h
@@ -483,7 +483,7 @@ PYGAMEAPI_EXTERN_SLOTS(math);
  *  functions in Python 3.
  */
 
-PG_INLINE PyObject *
+static PG_INLINE PyObject *
 pg_tuple_from_values_int(int val1, int val2)
 {
     PyObject *tup = PyTuple_New(2);
--- a/src_c/include/_pygame.h
+++ b/src_c/include/_pygame.h
@@ -484,8 +484,12 @@ PYGAMEAPI_EXTERN_SLOTS(math);
  */
 
 static PG_INLINE PyObject *
-pg_tuple_from_values_int(int val1, int val2)
+pg_tuple_couple_from_values_int(int val1, int val2)
 {
+    /* This function turns two input integers into a python tuple object.
+     * Currently, 5th November 2022, this is faster than using Py_BuildValue
+     * to do the same thing.
+     */
     PyObject *tup = PyTuple_New(2);
     if (!tup) {
         return NULL;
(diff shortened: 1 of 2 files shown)


MyreMylar:
I found this accepted PEP on removing macros from python in favour of inline functions:

https://peps.python.org/pep-0670/

Starbuck5:
It would be great to have a comment with the function explaining its function and purpose.

iirc:
Converts two integers to a python tuple faster than the general purpose PyBuild_Value way.

MyreMylar:
> It would be great to have a comment with the function explaining its function and purpose.
> 
> iirc: Converts two integers to a python tuple faster than the general purpose PyBuild_Value way.

Yeah, I think we will also want to rename this to "couple" and create a second version "triple" for three values into a tuple.

I can rejig this PR - or we could keep this one as a simple move and do the rename, rejig an comments in a new PR? Dunno what would be preferred for ease of review.

Starbuck5:
I’d prefer comment now and rejig later if at all.

That puts us in a good spot for readability now and potentially later.

MyreMylar:
Decided to just do the rename now and triple variant now as well. 

The triple will likely be used for the `color.rgb` setter in that PR

itzpr3d4t0r:
In our pygame_geometry project we settled on a different name for the function that i think is a bit better than this one, though i'll just throw it out there, no forcing:
`pg_TupleFromIntPair`

MyreMylar:
> In our pygame_geometry project we settled on a different name for the function that i think is a bit better than this one, though i'll just throw it out there, no forcing: `pg_TupleFromIntPair`

Now I'm just annoyed about the rapid switching between CamelCase and underscores in function names all over pygame...

Anyway, I'm happy to switch to whatever name or case arrangement as long as we have a logical way to distinguish between the couple/pair tuple and the triple/trio and are consistent.

ankith26 (approved):
LGTM! :+1: 
Definitely nice to reduce duplication and have functions like these in common files.

gresm (approved):

Issue #3531, opened 2022-11-02
Memory leak when using Vector2 distance_to()


derpsticks:
**Environment:**

You can get some of this info from the text that pops up in the console when you run a pygame program.

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Windows 10
- **Python version** (e.g. 3.7.9, 3.8.5) : 3.11
- **SDL version** (e.g. SDL 2.0.12): 2.0.22
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.1.3.dev8

**Current behavior:**

Using the distance_to() function of the Vector2 class to check entity distances in order to trigger an NPC aggro. I was seeing a large (a few megs a second) amount of memory being consumed as the checks were performed each frame. This memory consumption was not present prior to upgrading to python 3.11 / dev version of Pygame. This bug is causing a memory leak using Python 3.11 and the latest dev version of PyGame 2.1.3.dev8 at the time of writing. This error was not present using Python 3.10 and PyGame 2.1.2.

**Expected behavior:**

Not to consume memory after calling the function.

**Steps to reproduce:**

Please explain the steps required to duplicate the issue, especially if you are able to provide a sample application.
if the bug is caused by a specific file (image, font, sound, level, please upload it as an attachment

Wrote this code to find distance between two entities (works fine in python 3.10 and pygame 2.1.2):

    def distance(self, entity):
        return Vector2(self.cart_rect.center).distance_to(entity.cart_rect.center)

Where self.cart_rect is a simple Rect object with square dimensions of length/width not more than 10-15.

I was able to fix the issue by using the math libraries dist() function instead like so:

    def distance(self, entity):
        return math.dist(self.cart_rect.center, entity.cart_rect.center)

PR #3532, opened 2022-11-02
Fix memory leak for distance_to()

novialriptide:
Fixes: https://github.com/pygame/pygame/issues/3531


ankith26 (approved):
LGTM! :tada: :cake: 

I know that the 2.1.3 PR is already open atm, but this would be a good fix to get in there (since it also happens to be a regression)

PR #3538, opened 2022-11-05
Add `wraplength` to font.render, add `font.align` property

Starbuck5:
Raises an error if you try to set a font align outside of the 3 constant values, raises an error if you are on pygame compiled with old SDL_ttf and try to set align.

wraplength will be silently ignored if SDL_ttf < 2.0.18.

There's something weird font height changes, see the change I made to the font height unit test.


src_c/font.c
```diff
@@ -425,9 +483,10 @@ font_render(PyObject *self, PyObject *args)
     Uint8 rgba[] = {0, 0, 0, 0};
     SDL_Surface *surf;
     const char *astring = "";
+    int wraplength = 0;
```
novialriptide, line 486: We should add an error handler for negatives. No segmentation fault happens if we do, but we get this error: `pygame.error: Invalid parameter 'wrapLength'` which I think is ugly.

novialriptide (changes requested):
Align seems to work just fine for me, just one slight problem..
```python
a = font.render("hi starbuck", True, (255, 255, 255), wraplength=10)
```

This doesn't actually work, returning this:
```
PS C:\Users\novia\Documents\GitHub\pygame> python tester.py
pygame 2.1.4.dev1 (SDL 2.0.22, Python 3.11.0)
  File "C:\Users\novia\Documents\GitHub\pygame\tester.py", line 14, in <module>
    a = font.render("hi starbuck", True, (255, 255, 255), wraplength=10)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Font.render() takes no keyword arguments
```

We need to do this instead.
```python
a = font.render("hi starbuck", True, (255, 255, 255), (0, 0, 0, 0), 10)
```

The documentation is misleading, as it suggests that it would accept keyword arguments by showing its default values.

Instead of using `PyArg_ParseTuple()`, should we use `PyArg_ParseTupleAndKeywords()`? It'll be a performance loss, but the method that works is a messy implementation in my opinion.

## Example Code
```python
import pygame
import sys
pygame.init()

font = pygame.font.Font(None, 25)

while True:
    for e in pygame.event.get():
        if e.type == pygame.QUIT:
            sys.exit()
    
    screen.fill((0, 0, 0))
    a = font.render("hi starbuck", True, (255, 255, 255), (0, 0, 0, 0), 10)
    screen.blit(a, (0, 0))
    pygame.display.flip()
```

docs/reST/ref/font.rst
```diff
@@ -266,6 +281,10 @@ solves no longer exists, it will likely be removed in the future.
       to use for the text background. If no background is passed the area
       outside the text will be transparent.
 
+      Starting in `2.1.4`, there is a `wraplength` argument which describes the
```
novialriptide, line 284: I don't know if "Starting in `2.1.4`," is needed as it's been stated at the bottom already.

docs/reST/ref/font.rst
```diff
@@ -296,6 +312,9 @@ solves no longer exists, it will likely be removed in the future.
         pygame supports rendering UCS4 unicode including more languages and
         emoji.
 
+      .. versionchanged:: 2.1.4 newline characters now will break text into
```
novialriptide, line 315: Should specify `wraplength` keyword

novialriptide (changes requested):

novialriptide:
Testcases for `wraplength` seem to be missing as well.. although I don't know if that was intentional or not since we're just implementing SDL's wraplength.

PR #3540, opened 2022-11-05
Add an AVX2 version of BLIT_PREMULTIPLIED - Take 2

MyreMylar:
Original PR here #3268.

Apparently I pushed into a new branch on the pygame repository rather than my own copy after rebasing and updating it from @Starbuck5's review comments.

Anyway, it should have a lot more explanatory comments on the blending and shuffling now to help future folk (and my own future self) understand what the heck is going on with this SIMD stuff.


MyreMylar:
I've cleared out the old mmx version of blend pre-multiplied as it is worthless and will never be used with the SSE2 version around.

Starbuck5:
This PR should be merged first imo: #3276

MyreMylar:
I'm currently having doubts about whether this is properly adjusting for different pixel formats placement of the alpha channel.

MyreMylar:
The test I'm checking in should fail because the original AVX code is assuming the alpha is in a specific position, which it wouldn't be with some unusual pixel formats.

Unusual pixel formats won't generally come up on AVX2 compatible platforms, but I'd like to cover all bases.

MyreMylar:
And hopefully this new 'alpha index' concept should allow us to cope with the alpha being in any position in a 32 bit pixel format so ARGB or RGBA or BGRA.

MyreMylar:
Could not make the SSE2 blitter work with alpha not in the first position for some reason. I tried a macro to deal with the `imm8` issue but it just kept seg faulting when the alpha was not at the normal position. Perhaps I overlooked something, but for now I've just shunted the weird pixel formats down to the non-SIMD blitter on non-AVX2 supporting platforms.

MyreMylar:
Hooray, back to working and now supporting all 32bit pixel formats (in AVX2 mode and with fallbacks for no-avx2 platforms).

Previously we didn't support them with SIMD at all so that is an improvement, even if weird pixel format surfaces are pretty rare in pygame programs.

MyreMylar:
I will need to rework this after @Starbuck5's PR #3588

MyreMylar:
> I will need to rework this after @Starbuck5's PR #3588

Done.

Starbuck5:
We've previously discussed the difficulty of getting surfaces with unusual pixelformats. I've figured out how to do it, so maybe this will help increase the thoroughness of the tests:

```py
abgr = pygame.Surface((64, 64), masks=(0xFF, 0xFF00, 0xFF0000, 0xFF000000), depth=32)
rgba = pygame.Surface((64, 64), masks=(0xFF000000, 0xFF0000, 0xFF00, 0xFF), depth=32)
bgra = pygame.Surface((64, 64), masks=(0xFF00, 0xFF0000, 0xFF000000, 0xFF), depth=32)
argb = pygame.Surface((64, 64), masks=(0xFF0000, 0xFF00, 0xFF, 0xFF000000), depth=32)
```

I've suspected for a while that some of our alpha blitters wouldn't work on non-default 32 bit pixelformats, but I haven't really tested yet.

Starbuck5 (approved):
There are lots of things to do to polish up, improve, and verify the alpha blitters, but that shouldn't hold up this PR. The way to go is forward,

I couldn't find anything wrong with this, and I tried. Nice work Myre!

PR #3541, opened 2022-11-06
Add quickstart section to docs front page

### LINKED ISSUE #3406: Example boilerplate in the docs

Starbuck5:
Fixes #3406

Would happen instead of #3441

### DIFF
--- a/docs/reST/index.rst
+++ b/docs/reST/index.rst
@@ -15,6 +15,28 @@ Pygame Front Page
    filepaths
    logos
 
+Quick start
+-----------
+
+Welcome to pygame! Once you've got pygame installed (:code:`pip install pygame` or
+:code:`pip3 install pygame` for most people), the next question is how to get a game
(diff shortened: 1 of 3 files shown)


MyreMylar (approved):
Seems fine to me 👍

MightyJosip (approved):

PR #3547, opened 2022-11-06
Update macosx.py

reknahs:
I imported the specific packages from os rather than importing all of them. This reduces the amount of clashing package names in different libraries and improves the code readability and functionality.

### DIFF
--- a/src_py/macosx.py
+++ b/src_py/macosx.py
@@ -1,5 +1,5 @@
 import platform
-import os
+from os import getcwd, chdir, path
 import sys
 
 __all__ = ["Video_AutoInit"]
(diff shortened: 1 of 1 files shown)


Starbuck5:
Thanks for sending in a PR!

I'm -1 on this change. It doesn't help functionality and it raises the probability of name clashes. Most people wouldn't make a variable named `os`, but plenty of people would make a variable named `path` (for example).

Plus I'm not even sure this file is still in use.

MyreMylar:
> Plus I'm not even sure this file is still in use.

It's only function is called from `display.init()` on darwin macs.

I agree this doesn't add that much for us. However I don't think there is a real risk of name confusion either way here.

To accidentally overwrite the `os.path()` function here after importing it from macosx you would need a function at the following level:

```
import pygame.macosx.os.path
```

So right now to get a name collision for path at the pygame user level you'd need to create your own package called `pygame`, containing a subpackage called `macosx` that contained a file called `os` that had a function called `path` and then import that.

If we made this change it would reduce the number of functions available at:

```
pygame.macosx.os
```

but make the three os functions used in macosx.py available at:

```
pygame.macosx
```

instead.

---------------

I tested this import behaviour with a simple three file project:

_import_test.py:_
```
def do_stuff():
    print("doing stuff")
```

_other_import_test.py:_
```
import import_test


def do_stuff():
    print("doing some other stuff")
```

_final_import_test.py:_
```
import other_import_test

other_import_test.import_test.do_stuff()
other_import_test.do_stuff()
```

The two `do_stuff` function here do not collide.

PR #3548, opened 2022-11-07
Convert surface to texture's pixel format in `Texture.update`

### LINKED ISSUE #2974: Colors come out weird with _sdl2.Texture.update()
**Environment:**

- **Python version**: 3.9.6
- **SDL version**: 2.0.18
- **PyGame version**: 2.1.2
- **Relevant hardware**: nVidia GeForce RTX 2080 Ti

**Current behavior:**

The colours of certain images come out different than expected when using `_sdl2.Texture.update`. Possibly something to do with image formats. This is using the _sdl2 module.

**Expected behavior:**

Correctly rendered colours for the image.

**Screenshots**

The left image is the incorrect one, while the right is correct.
The left is using `texture.update()` to add a surface to a texture, the right is using the `texture.from_surface`
image

The original image:
lildragon01

**Steps to reproduce:**

1. Load png image as surface with `pygame.image.load`
2. Create sdl2 Texture
3. Draw surface onto texture with Texture.update()

**Test code**

```
import sys, pygame
from pygame import _sdl2


window = _sdl2.Window(size=(1048, 512))
renderer = _sdl2.Renderer(window)

surf = pygame.image.load("lildragon01.png")
# surf = pygame.image.load("lildragon01.bmp") # bmps also come out garbled

# The texture comes out weird with this
texture = _sdl2.Texture(renderer, size=(512, 512))
texture.update(surf, area=surf.get_rect())

# The colours come out correctly with this
texture2 = _sdl2.Texture.from_surface(renderer, surf)


while True:

        if event.type == pygame.QUIT:
            sys.exit()

        if event.type == pygame.KEYUP and event.key == pygame.K_ESCAPE:
            sys.exit()

    renderer.clear()
    texture.draw(dstrect=surf.get_rect())
    texture2.draw(dstrect=(512,0, surf.get_width(), surf.get_height()))

    renderer.present()
```

Zip with the source, BMP and PNG if you'd like to test:

[wrongcoloursdl.zip](https://github.com/pygame/pygame/files/7829787/wrongcoloursdl.zip)

Additionally BMPs comes out like this as well:
image

zoldalma999:
Should fix #2974

I feel like `surface.convert` should have an option to convert to a texture's format, as creating a new surface every time is not the best thing to do. But `Texture.update` should not be called a lot anyway, so it is fine for now.

Also, while it does set the converted surface's blend mode to the original, it does not set the texture's blend mode, meaning images may come out with a white outline, like this
image
but it is fixable with `texture.blend_mode = pygame.BLENDMODE_BLEND` (the other image uses `Texture.from_surface`, which does copy the blend mode)

### DIFF
--- a/src_c/_sdl2/video.c
+++ b/src_c/_sdl2/video.c
@@ -1071,7 +1071,7 @@ struct __pyx_obj_6pygame_5_sdl2_5video_Image {
 };
 
 
-/* "pygame/_sdl2/video.pyx":75
+/* "pygame/_sdl2/video.pyx":81
  *         )
  * 
  * def get_drivers():             # <<<<<<<<<<<<<<
@@ -1089,7 +1089,7 @@ struct __pyx_obj_6pygame_5_sdl2_5video___pyx_scope_struct__get_drivers {
 
 
 
-/* "pygame/_sdl2/video.pyx":922
+/* "pygame/_sdl2/video.pyx":957
  * 
  * 
  * cdef class Renderer:             # <<<<<<<<<<<<<<
@@ -1104,7 +1104,7 @@ struct __pyx_vtabstruct_6pygame_5_sdl2_5video_Renderer {
 static struct __pyx_vtabstruct_6pygame_5_sdl2_5video_Renderer *__pyx_vtabptr_6pygame_5_sdl2_5video_Renderer;
 
 
-/* "pygame/_sdl2/video.pyx":553
+/* "pygame/_sdl2/video.pyx":559
  * 
  * 
  * cdef class Texture:             # <<<<<<<<<<<<<<
@@ -1119,7 +1119,7 @@ struct __pyx_vtabstruct_6pygame_5_sdl2_5video_Texture {
 static struct __pyx_vtabstruct_6pygame_5_sdl2_5video_Texture *__pyx_vtabptr_6pygame_5_sdl2_5video_Texture;
 
 
-/* "pygame/_sdl2/video.pyx":799
+/* "pygame/_sdl2/video.pyx":834
  *             raise error()
  * 
  * cdef class Image:             # <<<<<<<<<<<<<<
@@ -2557,7 +2557,7 @@ static PyObject *__pyx_codeobj__45;
 static PyObject *__pyx_codeobj__47;
 /* Late includes */
 
-/* "pygame/_sdl2/video.pyx":65
+/* "pygame/_sdl2/video.pyx":71
  * 
  * class RendererDriverInfo:
  *     def __repr__(self):             # <<<<<<<<<<<<<<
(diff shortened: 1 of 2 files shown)


MyreMylar:
Assuming that converting a surface and then updating the texture is slower than just updating it with matching formats - is it worth also sending out a warning for the programmer so they know that information and might be able to do something about it? Though, actually, I guess sending the warning itself could be costly on a potentially performance sensitive function. 

I guess most usefully this could also be documented for the update function once this SDL video submodule actually has documentation. Perhaps we could update the function comment now as I bet the function comments get used as the basis for eventual documentation in the future. Something like "This function will be even slower if the pixel formats of the surface and texture do not match as a conversion has to be performed."

Starbuck5:
@MyreMylar your comment brings up how we don't have a good format system for surfaces and textures right now. SDL has the pixelformat enum, which is great, but we never ported it to pygame.

And we might have to switch it up b/c I believe we handle endianness different than they do. Our RGBA might be their ABGR.

MyreMylar:
> @MyreMylar your comment brings up how we don't have a good format system for surfaces and textures right now. SDL has the pixelformat enum, which is great, but we never ported it to pygame.
> 
> And we might have to switch it up b/c I believe we handle endianness different than they do. Our RGBA might be their ABGR.

Oh good.

I'm going to work up a draft PR for surfaces at least to print the SDL pixel format just so I can see what comes out and we can argue about it on that.

zoldalma999:
> Assuming that converting a surface and then updating the texture is slower than just updating it with matching formats - is it worth also sending out a warning for the programmer so they know that information and might be able to do something about it? Though, actually, I guess sending the warning itself could be costly on a potentially performance sensitive function.

Not sure honestly. In theory, one should not use this function a lot anyway (but instead [`SDL_LockTexture`](https://wiki.libsdl.org/SDL_LockTexture) and `UnlockTexture` should be used, something we did not yet expose). The bigger issue is that there is no easy way to convert a surface to a texture's format right now (as Starbuck mentioned). I imagine, once we start porting _sdl2 stuff to all the other (non-_sdl2) functions and modules, `surface.convert(texture)` will be valid.

> Perhaps we could update the function comment now as I bet the function comments get used as the basis for eventual documentation in the future. Something like "This function will be even slower if the pixel formats of the surface and texture do not match as a conversion has to be performed."

Yep, good point. Will do that.

Starbuck5 (approved):
I've been meaning to look at this for a while!

Everything checks out to me in this.

Might want to rebase and regen cython before merge just if anything has changed independently of this in video.pyx since this PR was made.

The SDL blend modes concepts are still very new to me, I had to do some digging around the codebase to start gaining some understanding. It seems like all the pygame internals manage the SDL-blend-mode of surfaces (NONE or BLEND) where appropriate. And of course the blit blend mode flags are completely implemented by us. I wonder if we'll eventually manage those  internals of textures too in a more polished API.

MyreMylar (approved):
LGTM 👍

PR #3549, opened 2022-11-07
docs: fix links

Linkid:


docs/reST/index.rst
```diff
@@ -188,12 +188,8 @@ Reference
 :ref:`search`
   Search pygame documents by keyword.
 
-.. _Readme: ../wiki/about
+.. _Readme: https://www.pygame.org/wiki/about
 
-.. _Install: ../wiki/GettingStarted#Pygame%20Installation
-
-.. _File Path Function Arguments: filepaths.html
+.. _Install: https://www.pygame.org/wiki/GettingStarted#Pygame%20Installation
```
Starbuck5, line 193: These don't need to be changed. They don't work on local docs copies sure, but that seems fine. If someone was to clone the pygame website they'd want the links to not be hard coded to the old one. Yeah? I'm open to other opinions.

Linkid: Done

Starbuck5 (approved):
Thanks!

PR #3550, opened 2022-11-08
Performance improvement for draw.circle

MightyJosip:
This should improve circles drawn with thickness, for example `pygame.draw.circle(screen, "YELLOW", (200, 200), 100, 40)` (and a few percent improvement to some other draw functions, but that is not relevant)

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1043,6 +1043,23 @@ add_pixel_to_drawn_list(int x, int y, int *pts)
     }
 }
 
+static void
+add_line_to_drawn_list(int x1, int y1, int x2, int *pts)
+{
+    if (x1 < pts[0]) {
+        pts[0] = x1;
+    }
+    if (y1 < pts[1]) {
+        pts[1] = y1;
+    }
+    if (x2 > pts[2]) {
+        pts[2] = x2;
+    }
+    if (y1 > pts[3]) {
+        pts[3] = y1;
+    }
+}
+
 static int
 clip_line(SDL_Surface *surf, int *x1, int *y1, int *x2, int *y2)
 {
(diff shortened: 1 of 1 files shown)


Starbuck5:
Does this also fix https://github.com/pygame/pygame/issues/3143?

Starbuck5 (approved):
I tested this with varying widths and everything still seems to work.

This is not pixel perfect to the old circle drawing algorithm, but it's very close.

I did some performance testing--
```py
start = time.time()
for _ in range(1000000):
    pygame.draw.circle(screen, "red", (200, 200), 78.4, 20)
print("elapsed=", time.time() - start)

# OLD: elapsed= 34.11649465560913
# NEW: elapsed= 7.675969123840332
```

That's a 344% performance improvement on this test case, very impressive.

I don't understand the implementation changes, but with my testing I'm confident enough to approve regardless.

MightyJosip:
> Does this also fix #3143?

It does not. We have supposed fix in #3151. My fix would be adding 
```
if (posx <= -2147483648 + radius || posx >= 2147483648 - radius ||
    posy <= -2147483648 + radius || posx >= 2147483648 - radius) {
        return pgRect_New4(posx, posy, 0, 0);
    }
```
here https://github.com/pygame/pygame/blob/51a9dc04d86b3d4abbc5fbe31193bb742ff78c45/src_c/draw.c#L679. I tested it and it works (actually it fails around 2**31, depending on the radius, and with the code above it would work) But it gives me signed/unsigned int comparison warning. Anyways I would leave that maybe to another PR

robertpfeiffer:
Would you be willing to merge a patch first that improves perf by ~100%?

MyreMylar (approved):
Looks good. Tested it on windows and thick circles still look like thick circles, and all the tests pass.

Issue #3526, opened 2022-10-29
VideoInfo.pixel_format: remove SDL_ prefix


illume:
For https://github.com/pygame/pygame/pull/2964

Probably would be good to remove the `SDL_` prefix to remain consistent.

PR #3551, opened 2022-11-08
Add a function to print surface's SDL pixel format

MyreMylar:
This is just a little experiment to see what results we get back from printing SDL's pixel format for pygame surface's after various manipulations.

My first test was this:

```
import pygame

argb_surf = pygame.Surface((64, 64), flags=pygame.SRCALPHA)
print(argb_surf.get_pixel_format())
byte_surf = pygame.image.tobytes(argb_surf, "ARGB")
rgba_surf = pygame.image.frombytes(byte_surf, (64, 64), "ARGB")
print(rgba_surf.get_pixel_format())
``` 

Which produced this output:
```
SDL_PIXELFORMAT_ARGB8888
SDL_PIXELFORMAT_BGRA8888
```

And I was immediately annoyed.

Are there any other good ways of converting surfaces to different pixel formats in pygame?


ankith26:
Perhaps this should strip the initial `"SDL_"` part of the string? (related issue: #3526) Or maybe we could even strip the entire `"SDL_PIXELFORMAT_"` prefix and get away with only returning something like `"ARGB8888"` or whatever

ankith26:
Of course, the underlying problem is that we never properly exposed pixelformat API, but IDK if it's a good idea to expose it now? Maybe it was intentionally left out to reduce some complexity for the users, but probably this info is useful for some advanced usages?

Starbuck5:
> Of course, the underlying problem is that we never properly exposed pixelformat API, but IDK if it's a good idea to expose it now? Maybe it was intentionally left out to reduce some complexity for the users, but probably this info is useful for some advanced usages?

Thoughts:
- convert/convert_alpha could take a `format` kwarg
- surface/texture/renderer could have a `get_format()`
- surface creation could have a `format` kwarg
- display.list_modes things have modes with different formats, that needs to be exposed
- SDL identifiers and our identifiers (image.fromstring things) disagree about how to apply endianness
- We could be consist with image.fromstring things and convert SDL pixelformats to a set of literal strings

PR #3552, opened 2022-11-08
Swap Py_BuildValue with pg_tuple_from_values_int

MightyJosip:
Connected: #3530,  #3424, #3401

### DIFF
--- a/src_c/color.c
+++ b/src_c/color.c
@@ -1841,10 +1841,10 @@ _color_slice(register pgColorObject *a, register Py_ssize_t ilow,
         return Py_BuildValue("(iiii)", c1, c2, c3, c4);
     }
     else if (len == 3) {
-        return Py_BuildValue("(iii)", c1, c2, c3);
+        return pg_tuple_triple_from_values_int(c1, c2, c3);
     }
     else if (len == 2) {
-        return Py_BuildValue("(ii)", c1, c2);
+        return pg_tuple_couple_from_values_int(c1, c2);
     }
     else if (len == 1) {
         return Py_BuildValue("(i)", c1);
--- a/src_c/_camera.c
+++ b/src_c/_camera.c
@@ -341,7 +341,7 @@ PyObject *
 camera_get_size(pgCameraObject *self, PyObject *_null)
 {
 #if defined(__unix__) || defined(PYGAME_WINDOWS_CAMERA)
-    return Py_BuildValue("(ii)", self->width, self->height);
+    return pg_tuple_couple_from_values_int(self->width, self->height);
 #endif
     Py_RETURN_NONE;
 }
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -353,7 +353,7 @@ pg_init(PyObject *self, PyObject *_null)
     }
 
     pg_is_init = 1;
-    return Py_BuildValue("(ii)", success, fail);
+    return pg_tuple_couple_from_values_int(success, fail);
 }
 
 static void
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1389,7 +1389,7 @@ pg_window_size(PyObject *self, PyObject *_null)
     if (!win)
         return RAISE(pgExc_SDLError, "No open window");
     SDL_GetWindowSize(win, &w, &h);
-    return Py_BuildValue("(ii)", w, h);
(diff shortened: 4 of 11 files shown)


itzpr3d4t0r:
Approved!

src_c/color.c
```diff
@@ -1841,10 +1841,10 @@ _color_slice(register pgColorObject *a, register Py_ssize_t ilow,
         return Py_BuildValue("(iiii)", c1, c2, c3, c4);
     }
     else if (len == 3) {
-        return Py_BuildValue("(iii)", c1, c2, c3);
+        return pg_tuple_triple_from_values_int(c1, c2, c3);
     }
     else if (len == 2) {
-        return Py_BuildValue("(ii)", c1, c2);
+        return pg_tuple_couple_from_values_int(c1, c2);
```
itzpr3d4t0r, line 1847: These should be cast to ints

ankith26 (commented):
Thanks for the PR! The lint CI is failing so we need a round of `setup.py format` here. Also some places probably need explicit `int` casts, because otherwise it would be implicit.

src_c/color.c
Starbuck5: Yes-- this is why the appveyor builds are failing.

Starbuck5:
I did some performance testing on the change of method itself.

```py
import pygame
import time

a = time.time()
for i in range(10):
    pygame.transform.test()
print(time.time() - a)
```

```c
static PyObject*
testtest(PyObject *self, PyObject *args)
{
    int x = 15;
    int y = 772;
    int z = 12;
    PyObject* result;

    for (int i=0; i<500000; i++) {
        //result = Py_BuildValue("(ii)", x, y);
        //result = pg_tuple_couple_from_values_int(x, y);
        //result = Py_BuildValue("(iii)", x, y, z);
        result = pg_tuple_triple_from_values_int(x, y, z);
    }

    Py_RETURN_NONE;
}

...
{"test", (PyCFunction)testtest, METH_NOARGS, NULL},
```

**Results**
```
# 500k*10 tuple II 
# BuildValue: 0.33 seconds
# custom: 0.22 seconds

# 500k*10 tuple III
# BuildValue: 0.35 seconds
# custom: 0.24 seconds
```

So 5 million calls translates into a savings of only 0.11 seconds. Therefore it would take around 45 million calls to something using this to notice 1 second of runtime improvement. This is very much a micro-optimization. But I don't see it adding maintenance burden for us, so I'm pro.

itzpr3d4t0r:
I think your results are a bit incorrect, firstly because the C code has a memory leak ( you didn't DECREF result) and second because by doing 
```Python
a = time.time()
for i in range(10):
    pygame.transform.test()
print(time.time() - a)
```
You are including the creation time of the range object, the looping and access to the module and in reality when this function is called internally in C you don't do all this stuff, so i came up with a slightly different measuring strategy:

- First i fixed the memry leak which was actually changing the results apparently
- Second i changed the Python code to the much more precise timeit, excluding
- I also made the measurements 5 times in order to check whether the first call add extra overhead:
```Python
from pygame.transform import test
from timeit import timeit

for _ in range(5):
    print(timeit(test, number=10))
```
And got these results:

```
TRIPLE

old
0.24411700000018755
0.2521020999997745
0.24876799999992727
0.2405932000001485
0.2455758999999489

new
0.12007300000004761
0.12075110000023415
0.12277529999983017
0.11842769999975644
0.12147270000014032

=====
DOUBLE

old
0.2224092000001292
0.22611889999961932
0.2230251999999382
0.22691179999992528
0.22417660000019168

new
0.11334439999973256
0.1101702999999361
0.11015529999986029
0.11212010000008377
0.11333589999958349
```

So from my measurements i get roughly 96% faster for the double and 106% faster for the triple.

Starbuck5:
@MightyJosip Lets move this forward.

Ankith and itzpr pointed out small things that need fixing. (casts + format run)

Once you do that the builds will probably start failing because of a sphinx issue, you'll need to rebase to get the fix for that.

Starbuck5:
MightyJosip on discord:

> About your comment, I am gonna do it next month, I don't have time now, next 3 weeks should hopefully be the last weeks of my university :hype: (not counting writing the thesis), so I will have more time

PR #3554, opened 2022-11-09
Minor typo in the "pygame.Color" docs

pgattic:
"hsva" doesn't have an "L" property, its supposed to be "V"


Starbuck5 (approved):
Thanks!

MightyJosip (approved):
Good catch. Thanks.

Issue #3478, opened 2022-10-05
pygame.transform.grayscale()


novialriptide:
Pretty self-explanatory; it's a new function that supports grayscaling a `pygame.Surface`.

The design I'm thinking of is pretty simple as well, just this:
```python
pygame.transform.grayscale(surface: pygame.Surface) -> pygame.Surface
```

## Image Example
![image](https://user-images.githubusercontent.com/35881688/194006308-0125d823-35f1-4c29-9d2f-ac72a80fdebc.png)

Starbuck5:
See https://docs.opencv.org/4.x/de/d25/imgproc_color_conversions.html#color_convert_rgb_gray

Jam-Burger:
I would like to work on this issue

PR #3556, opened 2022-11-11
grayscale for Surfaces and Colors

ScriptLineStudios:
As @novialriptide suggested in #3478

```python
pygame.transform.grayscale(surface: Surface, dest_surface: Optional[Surface] = None) -> Surface
```

And as @MightyJosip suggested below
```python
pygame.Color.grayscale() -> Color
```


MyreMylar:
Looks like some good work - I want to give it a full test at the weekend.

I also want to highlight that we should probably make a SIMD version of this (not in this PR, but quickly after it). Should be able to work something up based on existing AVX2 special blend modes for the multiplying part and gotta split everything into red, green and blue with shuffles for the addition bit and bitshift part (at 16 bit).

novialriptide:
I think we should also have an image grayscale for unit tests as well instead of a red surface.

src_c/transform.c
```diff
@@ -570,6 +570,96 @@ scale_to(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int width,
...
+            "Destination surface must be the same size as source surface."));
+    }
+
+    if (src->format->BytesPerPixel != newsurf->format->BytesPerPixel) {
+        return (SDL_Surface *)(RAISE(
+            PyExc_ValueError,
+            "Source and destination surfaces need the same format."));
+    }
+
+    Uint32 *pixels = (Uint32 *)src->pixels;
```
Starbuck5, line 600: This assumes the bit depth is 32 pixels.

If you give it a 24 bit surface or any other depth, it messes up.

32 bit:
output32

24 bit:
output24

src_c/transform.c
```diff
@@ -570,6 +570,96 @@ scale_to(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int width,
...
+
+    SDL_LockSurface(newsurf);
+
+    int i;
+    for (i = 0; i < src->w * src->h; i++) {
```
Starbuck5, line 605: This assumes every pixel is laid out one after the other directly in memory. A->B->C->D

Some surfaces are not this way, and each row of pixels is aligned on some boundary (like a multiple of 64, for example).
So all surfaces have a "stride/pitch" measurement to determine how to iterate through their pixels.

src_c/transform.c
```diff
@@ -570,6 +570,96 @@ scale_to(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int width,
...
+        SDL_GetColorKey(src, &transparent_pixel);
+
+        int x = i % newsurf->w;
+        int y = i / newsurf->w;
+        Uint32 *const target_pixel =
+            (Uint32 *)((Uint8 *)newsurf->pixels + y * newsurf->pitch +
+                       x * newsurf->format->BytesPerPixel);
+        if (pixels[i] != transparent_pixel) {
+            Uint32 new_pixel = (0xFF << 24) | (grayscale_pixel << 16) |
+                               (grayscale_pixel << 8) | grayscale_pixel;
```
Starbuck5, line 621: This assumes that the target surface is laid out as ARGB or ABGR (alpha first in memory).

With a surface with the alpha channel in a different location it generates the wrong result--
output_wrongchannels

The alpha info went into blue and the greyscale luminance went into the alpha channel.

To translate an image into that surface format, I used this code (I couldn't get the masks argument of pygame.Surface to work)--
```py
base_bytes = bytes(image.get_width() * 4 * image.get_height())
image_base = pygame.image.frombytes(base_bytes, image.get_size(), "ARGB")
image_base.blit(image, (0,0))
```

ScriptLineStudios: Okay I have updated this to use SDL_MapRGBA

src_c/transform.c
MightyJosip: In draw.c you can find the example of how to iterate pixels for single y value: https://github.com/pygame/pygame/blob/main/src_c/draw.c#L1357

ScriptLineStudios:
Okay I have made a few changes. Here is a demo of some images with different color depths being rendered. 
image

src_c/transform.c
```diff
@@ -570,6 +609,110 @@ scale_to(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int width,
     return retsurf;
 }
 
+SDL_Surface *
+grayscale(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj)
+{
+    SDL_Surface *src = pgSurface_AsSurface(srcobj);
+    SDL_Surface *newsurf;
```
Matiiss, line 616: Should check if the surface has been initialized before proceeding. Related to PR https://github.com/pygame/pygame/pull/3525 and issue https://github.com/pygame/pygame/issues/3519

Starbuck5:
This segfaults for me on a 24 bit depth surface. You can create a 24 bit depth surface with `pygame.Surface(size, depth=24)`. I believe surfaces are 32 bit RGBX-like by default.

src_c/transform.c
```diff
@@ -570,6 +609,110 @@ scale_to(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int width,
...
+                SDL_GetRGBA(pixel_8[i], newsurf->format, &r, &g, &b, &a);
+            }
+            else if (pixel_16) {
+                SDL_GetRGBA(pixel_16[i], newsurf->format, &r, &g, &b, &a);
+            }
+            else if (pixel_32) {
+                SDL_GetRGBA(pixel_32[i], newsurf->format, &r, &g, &b, &a);
+            }
+            else {
+                SDL_GetRGBA(pixel_32[i], newsurf->format, &r, &g, &b, &a);
```
Starbuck5, line 669: It seems like it should be using the src surface's format, because that's what it is reading from.

src_c/transform.c
```diff
@@ -570,6 +609,110 @@ scale_to(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int width,
...
+            }
+            Uint32 grayscale_pixel =
+                0.212671 * r + 0.715160 * g + 0.072169 * b;
+            Uint32 new_pixel =
+                SDL_MapRGBA(src->format, grayscale_pixel, grayscale_pixel,
+                            grayscale_pixel, a);
```
Starbuck5, line 675: Same here-- shouldn't this be the newsurf's format being used here?

docs/reST/ref/transform.rst
```diff
@@ -243,6 +243,18 @@ Instead, always begin with the original image and scale to the desired size.)
...
 
+.. function:: grayscale
+
+   | :sl:`grayscale a surface`
+   | :sg:`grayscale(surface, dest_surface=None) -> Surface`
+
+   Grayscales the surface.
+
+   An optional destination surface can be passed which is faster than creating a new Surface.
+   However this destination surface must have the same dimensions (width, height) as the source Surface.
```
Starbuck5, line 254: If you're going to have it only work between surfaces of the same bit depth, that should be documented here.

Matiiss:
Maybe should have a `greyscale` alias, since the colour dictionary has that.

MightyJosip:
When we are doing this maybe you could write `pygame.Color.grayscale()` as well

docs/reST/ref/transform.rst
```diff
@@ -243,6 +243,18 @@ Instead, always begin with the original image and scale to the desired size.)
 
    .. ## pygame.transform.average_color ##
 
+.. function:: grayscale
+
+   | :sl:`grayscale a surface`
+   | :sg:`grayscale(surface, dest_surface=None) -> Surface`
+
+   Grayscales the surface.
```
Starbuck5, line 251: Well, technically it returns a new surface that's been greyscaled from the original. But it would be also good to briefly describe the luminance formula you're using.

src_c/transform.c
```diff
@@ -2110,6 +2110,85 @@ clamp_4
...
+    int x, y;
+    for (y = 0; y < src->h; y++) {
+        for (x = 0; x < src->w; x++) {
+            Uint32 pixel;
+            Uint8 *pix;
+            SURF_GET_AT(pixel, src, x, y, (Uint8 *)src->pixels, src->format,
+                        pix);
+            Uint8 r, g, b, a;
+            SDL_GetRGBA(pixel, src->format, &r, &g, &b, &a);
+            Uint8 grayscale_pixel = (Uint8)(0.212671 * r + 0.715160 * g + 0.072169 * b);
```
Starbuck5, line 2149: I didn't notice this before, but this formula is not right.

https://docs.opencv.org/4.x/de/d25/imgproc_color_conversions.html#color_convert_rgb_gray
`Y←0.299⋅R+0.587⋅G+0.114⋅B`

It looks like you lifted something from `RGB ↔ CIE XYZ.Rec 709 with D65 white point` instead.

MyreMylar:
looks like this needs a typestub:

```
error: pygame.color.Color.grayscale is not present in stub
Stub: in file /home/runner/work/pygame/pygame/buildconfig/stubs/pygame/color.pyi
```

MyreMylar:
Also:
```
  FAIL: test_grayscale (pygame.tests.transform_test.TransformModuleTest)
    File "/tmp/tmp.7Xv1lUYSX7/venv/lib/python3.6/site-packages/pygame/tests/transform_test.py", line 186, in test_grayscale
      self.assertEqual(pygame.transform.average_color(dest)[0], 123)
  AssertionError: 122 != 123
  
```
Is failing in the linux builds.

MyreMylar (changes requested):
As above.

illume:
It’s failing here on Linux:
https://github.com/pygame/pygame/actions/runs/4212126223/jobs/7313061820#step:3:5599

"/tmp/tmp.pajFQLSxtU/venv/lib/python3.6/site-packages/pygame/tests/transform_test.py", line 186, in test_grayscale
[5590](https://github.com/pygame/pygame/actions/runs/4212126223/jobs/7313061820#step:3:5600)
      self.assertEqual(pygame.transform.average_color(dest)[0], 123)
[5591](https://github.com/pygame/pygame/actions/runs/4212126223/jobs/7313061820#step:3:5601)
  AssertionError: 122 != 123

Feel free to just comment out that part of the test for now, and we can merge it.

illume (approved):
👍 thanks

Issue #3523, opened 2022-10-27
`__round__` method for `pygame.Vector2` and `pygame.Vector3`


Matiiss:
Adding `__round__` method to the mentioned objects so that it is possible to do this `round(vector_instance, ndigits)` and it would return a new vector with all of its values rounded to the specified `ndigits`.

Currently doing so raises an exception
![image](https://user-images.githubusercontent.com/83066658/198416293-f9bc9634-003b-4383-90aa-5569b22b7fa2.png)

Reasoning:
Currently to achieve the same result (rounding all components) you have to do this (or something similar)
```py
vec = pygame.Vector2(1, 1)
new_vec = pygame.Vector2(round(vec.x), round(vec.y))
```

Additionally it could support specifying which components and to how many digits to round by passing a tuple of length 2 or 3 depending on the vector as the `ndigits` argument. (Supposedly either or both could be None which would default to 1 as it is with number values).

I can try my hand at implementing this if the enhancement gets accepted.

ankith26:
+1 from my side for this suggestion

Matiiss:
Alright, I would like to tackle this issue then. I suppose I just follow how other functions have been written there (declaration at top and this goes under generic vector functions) and I guess this will need tests too.

novialriptide:
+1

PR #3559, opened 2022-11-13
Added `__round__` method for vectors

Matiiss:
\+ tests (taken from https://github.com/python/cpython/blob/c32bc1bffd9d63ede0d0505abab983247a3ad0c6/Lib/test/test_builtin.py#L1477)

Fixes [#3523](https://github.com/pygame/pygame/issues/3523)

Added type stubs

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -266,6 +266,8 @@ static PyObject *
 vector_clamp_magnitude_ip(pgVector *self, PyObject *const *args,
                           Py_ssize_t nargs);
 PG_DECLARE_FASTCALL_FUNC(vector_clamp_magnitude_ip, pgVector);
+static PyObject *
+vector___round__(pgVector *self, PyObject *args);
 
 /*
 static Py_ssize_t vector_readbuffer(pgVector *self, Py_ssize_t segment, void
@@ -2082,6 +2084,47 @@ static PyBufferProcs vector_as_buffer = {
 };
 #endif
 
+static PyObject *
+vector___round__(pgVector *self, PyObject *args)
+{
+    Py_ssize_t i, ndigits;
+    PyObject *o_ndigits = NULL;
+
+    pgVector *ret = _vector_subtype_new(self);
+    if (ret == NULL) {
+        return NULL;
+    }
+
+    if (!PyArg_ParseTuple(args, "|O", &o_ndigits)) {
+        Py_DECREF(ret);
+        return NULL;
+    }
+
+    memcpy(ret->coords, self->coords, sizeof(ret->coords[0]) * ret->dim);
+
+    if (o_ndigits == NULL || o_ndigits == Py_None) {
+        for (i = 0; i < ret->dim; ++i)
+            ret->coords[i] = round(ret->coords[i]);
+    }
+    else if (RealNumber_Check(o_ndigits)) {
+        ndigits = PyNumber_AsSsize_t(o_ndigits, NULL);
+        if (PyErr_Occurred()) {
+            Py_DECREF(ret);
+            return NULL;
+        }
+        for (i = 0; i < ret->dim; ++i)
+            ret->coords[i] = round(ret->coords[i] * pow(10, (double)ndigits)) /
+                             pow(10, (double)ndigits);
+    }
+    else {
+        PyErr_SetString(PyExc_TypeError, "Argument must be an integer");
+        Py_DECREF(ret);
+        return NULL;
+    }
+
+    return (PyObject *)ret;
+}
+
 /*********************************************************************
  * vector2 specific functions
  *********************************************************************/
@@ -2473,6 +2516,7 @@ static PyMethodDef vector2_methods[] = {
     {"__safe_for_unpickling__", (PyCFunction)vector_getsafepickle, METH_NOARGS,
      NULL},
     {"__reduce__", (PyCFunction)vector2_reduce, METH_NOARGS, NULL},
+    {"__round__", (PyCFunction)vector___round__, METH_VARARGS, NULL},
 
     {NULL} /* Sentinel */
 };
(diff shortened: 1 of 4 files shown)


Matiiss:
Why is this failing on a font test?

Matiiss:
What? Those tests didn't fail this time.

Starbuck5:
This will need a note in the documentation that this was added in pygame 2.1.4.

Matiiss:
> This will need a note in the documentation that this was added in pygame 2.1.4.

That's what I thought, the question is where in the docs should this be added? I suppose it's not exactly a separate method, because it's not intended to be called like `vector_instance.__round__(ndigits)`, so how and where? A note under this: _Changed in pygame 1.9.4: [pygame.mathpygame module for vector classes](https://www.pygame.org/docs/ref/math.html#module-pygame.math) required import. More convenient pygame.Vector2 and pygame.Vector3._?

Matiiss:
Should this support kwargs?

Starbuck5:
> Should this support keyword arguments?

It should try to emulate a normal numerical `__round__`. So if they do, we should.

The docs addition you bring up makes sense.

Matiiss:
`__round__` for floats doesn't support keyword arguments, `round` on the other hand does already support keyword arguments but it's a built-in function so those keyword arguments can't really be changed, either way, it seems to work fine with keyword arguments too

oddbookworm:
> Should the docs say that this is new in version 2.1.4dev1
I think it'd be fine saying just 2.1.4

src_c/math.c
```diff
@@ -2082,6 +2084,40 @@ static PyBufferProcs vector_as_buffer = {
...
+static PyObject *
+vector___round__(pgVector *self, PyObject *args)
+{
+    Py_ssize_t i, ndigits;
+    PyObject *o_ndigits = NULL;
+
+    pgVector *ret = _vector_subtype_new(self);
+
+    if (ret == NULL || !PyArg_ParseTuple(args, "|O", &o_ndigits)) {
+        return NULL;
```
Starbuck5, line 2096: I see what you're doing here, and it does the make the code shorter, but does it make it harder to read? I don't have a strong opinion just was thinking.

Starbuck5 (commented):
math.rst line 16 has a description of all the numerical operations Vectors support, it would be good to add support for `round` to that list.

Starbuck5 (approved):
Left a couple more comments for consideration before merge, but the code and functionality looks great. Good job!

src_c/math.c
ankith26: I would very much prefer if this was done on 2 separate `if` blocks instead of one. The reason being that in this code, `PyArg_ParseTuple` will be executed regardless of the first check. But if `ret` is `NULL` we are supposed to exit immediately as it is most likely a memory error without calling more python c api functions

Matiiss: @ankith26 should it simply `return NULL;` if `ret` is `NULL` or also set an error message?

src_c/math.c
```diff
@@ -2082,6 +2084,40 @@ static PyBufferProcs vector_as_buffer = {
...
+
+    memcpy(ret->coords, self->coords, sizeof(ret->coords[0]) * ret->dim);
+
+    if (o_ndigits == NULL || o_ndigits == Py_None) {
+        for (i = 0; i < ret->dim; ++i)
+            ret->coords[i] = round(ret->coords[i]);
+    }
+    else if (RealNumber_Check(o_ndigits)) {
+        ndigits = PyNumber_AsSsize_t(o_ndigits, NULL);
+        if (ndigits == -1 && PyErr_Occurred())
```
Matiiss, line 2107: Does it need that `ndigits == -1` part?

Matiiss:
Did I just...

src_c/math.c
Starbuck5: Both PyArg_Parse and vector_subtype_new will set the error themselves, so no need for you to set the error.

src_c/math.c
Starbuck5: I'm not sure why you're asking this. That's what the function says it will do on error, seems reasonable to check that exactly-- https://docs.python.org/3/c-api/number.html#c.PyNumber_AsSsize_t.

src_c/math.c
```diff
@@ -2082,6 +2084,43 @@ static PyBufferProcs vector_as_buffer = {
...
+    Py_ssize_t i, ndigits;
+    PyObject *o_ndigits = NULL;
+
+    pgVector *ret = _vector_subtype_new(self);
+    if (ret == NULL) {
+        return NULL;
+    }
+
+    if (!PyArg_ParseTuple(args, "|O", &o_ndigits)) {
+        return NULL;
```
MyreMylar, line 2099: needs a:

```
Py_DECREF(ret);
```

see: #3590

src_c/math.c
```diff
@@ -2082,6 +2084,43 @@ static PyBufferProcs vector_as_buffer = {
...
+    memcpy(ret->coords, self->coords, sizeof(ret->coords[0]) * ret->dim);
+
+    if (o_ndigits == NULL || o_ndigits == Py_None) {
+        for (i = 0; i < ret->dim; ++i)
+            ret->coords[i] = round(ret->coords[i]);
+    }
+    else if (RealNumber_Check(o_ndigits)) {
+        ndigits = PyNumber_AsSsize_t(o_ndigits, NULL);
+        if (PyErr_Occurred())
+            return NULL;
```
MyreMylar, line 2111: also needs a:

```
Py_DECREF(ret);
```

src_c/math.c
```diff
@@ -2082,6 +2084,43 @@ static PyBufferProcs vector_as_buffer = {
...
+        for (i = 0; i < ret->dim; ++i)
+            ret->coords[i] = round(ret->coords[i] * pow(10, (double)ndigits)) /
+                             pow(10, (double)ndigits);
+    }
+    else {
+        PyErr_SetString(PyExc_TypeError, "Argument must be an integer");
+        return NULL;
```
MyreMylar, line 2118: and again:

```
Py_DECREF(ret);
```

MyreMylar (changes requested):
needs a few memory leak fixes. See #3590 for more details.

Matiiss:
@MyreMylar I have made the requested changes, it would be great if you could review them, thanks!

MyreMylar (approved):
LGTM 👍

PR #2964, opened 2021-12-29
Add pixel format name into display info

MyreMylar:
Helps make it easier to spot when display surfaces are different from what we are expecting. This came up from discussion of display surfaces being RGB on windows and linux, but ARGB on mac.

**Test program**:
```python
import pygame

pygame.init()
print(pygame.display.Info())
```

**Before:**
```
<VideoInfo(hw = 0, wm = 1,video_mem = 0
         blit_hw = 0, blit_hw_CC = 0, blit_hw_A = 0,
         blit_sw = 0, blit_sw_CC = 0, blit_sw_A = 0,
         bitsize  = 32, bytesize = 4,
         masks =  (16711680, 65280, 255, 0),
         shifts = (16, 8, 0, 0),
         losses =  (0, 0, 0, 8),
         current_w = 800, current_h = 600
>
```

**After:**
```
<VideoInfo(hw = 0, wm = 1,video_mem = 0
         blit_hw = 0, blit_hw_CC = 0, blit_hw_A = 0,
         blit_sw = 0, blit_sw_CC = 0, blit_sw_A = 0,
         bitsize  = 32, bytesize = 4,
         masks =  (16711680, 65280, 255, 0),
         shifts = (16, 8, 0, 0),
         losses =  (0, 0, 0, 8),
         current_w = 800, current_h = 600
         pixel_format = SDL_PIXELFORMAT_RGB888
>
```

PR #3560, opened 2022-11-13
Cut SDL_PIXELFORMAT_ prefix off in vidinfo

### LINKED ISSUE #3526: VideoInfo.pixel_format: remove SDL_ prefix

zoldalma999:
should fix #3526 
I also removed the `PIXELFORMAT_` part, since that also feels redundant and added `pixel_format` as an attribute (before it was only in its str method)

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -308,6 +308,7 @@ required).
                  is called. (current_h, current_w are available since
                  SDL 1.2.10, and pygame 1.8.0). They are -1 on error, or if
                  an old SDL is being used.
+     pixel_format: A string representing the format of each pixel's color
 
    .. ## pygame.display.Info ##
 
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -323,6 +323,11 @@ pg_vidinfo_getattr(PyObject *self, char *name)
         return PyLong_FromLong(current_h);
     else if (!strcmp(name, "current_w"))
         return PyLong_FromLong(current_w);
+    else if (!strcmp(name, "pixel_format")) {
+        const char *name = SDL_GetPixelFormatName(info->vfmt->format);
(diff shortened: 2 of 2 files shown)


ankith26 (commented):
Is it a coincidence that you implemented something along the lines of what I suggested in #3551? Or did you see that PR before?

Probably better to not have two PRs doing very similar things and still need discussion. I'm inclined to continue the discussion here because the other one looked like a temporary draft test thing ;)

Anyways, this PR might need more thought because `pixel_format` would be something completely new in the pygame world. I'm not sure how much of it we really need to expose given most python side users don't do low level surface manipulations I guess?  Whatever we implement, it should still probably be consistent with `fromstring`/`frombytes`/`frombuffer` (and these might need extending)

zoldalma999:
> Is it a coincidence that you implemented something along the lines of what I suggested in https://github.com/pygame/pygame/pull/3551? Or did you see that PR before?

I opened this pr to fix the linked issue

> Probably better to not have two PRs doing very similar things and still need discussion. I'm inclined to continue the discussion here because the other one looked like a temporary draft test thing ;)

The pr you linked and this one does different things. While the other one adds a new function to see what a surface's pixel format is, mine just polishes an existing behaviour added in #2964 (and aims to fix #3526 as well). Previously, pixel_format was only in the `repr`/`str` method of VideoInfo, now it is an attribute as well (just like all others).

> Anyways, this PR might need more thought because pixel_format would be something completely new in the pygame world.

As it is already part of VideoInfo, so I don't see an issue with the changes I made. Either this gets merged in or pixel_format should be removed as a whole IMO.

MyreMylar (approved):
LGTM 👍

Issue #3545, opened 2022-11-06
“Infinite loop” test


Koswu:
When I open this project in VS Code, and switch to the “testing” tab. It seems loading infinitely, and CPU usage is unreasonably high.
![image](https://user-images.githubusercontent.com/10678681/200184413-aa9c74e0-7717-4cea-9ded-19bce94a6e79.png)

Then I notice this process is running in the background, and use much CPU: 
```
/bin/python /home/user/.vscode/extensions/ms-python.python-2022.18.1/pythonFiles/testing_tools/unittest_discovery.py . *test*.py
```

I think it will run all python files which contain “test” in the filename, then I found this file:
[test/test_test_.py](https://github.com/pygame/pygame/blob/main/test/test_test_.py)

This file just run an infinite loop, and I don't know what it purpose for. 
But after I deleted this file and reopen VS Code, everything went back to normal.

Could someone let me know what the file's purpose? Or we can just delete it safely?

ExtraMobs:
I think you write your problem in the wrong repository, here is the correct one:

https://github.com/microsoft/vscode

Koswu:
> why does this module exist in pygame examples?

Did I miss something? I just confuse about why this file's exists. It just run loop forever.
https://github.com/pygame/pygame/blob/main/test/test_test_.py

I don't think it is VS Code's issue, because it's just run this file via pytest.

ExtraMobs:
just delete the file to see if it solves the problem.

I personally think that as much as this might be a VSCode plugin issue, it still doesn't make sense that this module exists

Koswu:
> just delete the file to see if it solves the problem.

Yes, I mentioned it before.

Alternatively, if you just run 
```
pytest --collect-only
```
in the project directory

It will get stuck too.
I think the problem is the file should be deleted, but I don't know is it safe to delete ?

ExtraMobs:
Delete the file. It doesn't seem to have anything essential in it, test it, anything you reinstall pygame, and it's already registered so they can analyze and delete it in the future, regardless of the result.

Koswu:
Ok, I will make a pull request then.

ankith26:
Hi! The PR you made isn't made in the main pygame repository and is instead made to your own fork, perhaps this was accidental?

Anyways, I don't know the history behind why the file exists in the first place (probably something that verifies that the custom test machinery we have handles things correctly?) but I am okay with removing the file personally.

Koswu:
@ankith26 
[...]

PR #3563, opened 2022-11-14
Delete test_test_.py

Koswu:
Delete unused test which cause `pytest --collect-only` infinite loop
see https://github.com/pygame/pygame/issues/3545

### DIFF
--- a/test/test_test_.py
+++ b/test/test_test_.py
@@ -1,2 +1,0 @@
-while True:
-    pass


MyreMylar:
I personally have no idea why this file exists and am happy for it to be deleted. Doesn't seem to be testing anything useful to me.

MyreMylar (approved):

Starbuck5:
I agree with MyreMylar, this seems fine to delete.

Starbuck5 (approved):

Issue #3564, opened 2022-11-14
segfault with pygame.freetype.get_version


oddbookworm:
**Environment:**
- **Operating system** Windows 11
- **Python version** 3.11.0
- **SDL version** 2.0.20
- **PyGame version** 2.1.3dev8 and 2.1.4dev1
- **Relevant hardware** N/A

**Current behavior:**

Calling `pygame.freetype.get_version()` before initialization causes a segmentation fault

**Expected behavior:**

No segfault, should just raise an exception

**Test code**

```
import pygame
import pygame.freetype
pygame.freetype.get_version()
```

**Stack trace/error output/other error logs**

**WSL**
ignore that I didn't import freetype directly, that's a separate thing that I borked in my own branch
```
Python 3.11.0 (main, Oct 24 2022, 19:56:13) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pygame
/home/andrew/venv/lib/python3.11/site-packages/pygame-2.1.4.dev1-py3.11-linux-x86_64.egg/pygame/pkgdata.py:25: UserWarning: Module pygame was already imported from /home/andrew/venv/lib/python3.11/site-packages/pygame-2.1.4.dev1-py3.11-linux-x86_64.egg/pygame/__init__.py, but /home/andrew/pygame is being added to sys.path
  from pkg_resources import resource_stream, resource_exists
pygame 2.1.4.dev1 (SDL 2.0.20, Python 3.11.0)
Hello from the pygame community. https://www.pygame.org/contribute.html
>>> pygame.freetype.get_version()
Fatal Python error: pygame_parachute: (pygame parachute) Segmentation Fault
Python runtime state: initialized

Current thread 0x00007f0f376a2000 (most recent call first):
  File "<stdin>", line 1 in <module>

Extension modules: pygame.base, pygame.constants, pygame.rect, pygame.rwobject, pygame.surflock, pygame.bufferproxy, pygame.math, pygame.surface, pygame.display, pygame.draw, pygame.event, pygame.imageext, pygame.image, pygame.joystick, pygame.key, pygame.mouse, pygame.time, pygame.mask, pygame.pixelcopy, pygame.transform, pygame.font, pygame.mixer_music, pygame.mixer, pygame.scrap, pygame._freetype (total: 25)
[1]    12946 IOT instruction  python
```
**WSL gdb backtrace**
```
>>> pygame.freetype.get_version()

[...]

PR #3567, opened 2022-11-15
Segfault fix in freetype.get_version

oddbookworm:
Fixed the segfault reported in #3564

### DIFF
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -2046,8 +2046,17 @@ _ft_get_version(PyObject *self, PyObject *args, PyObject *kwargs)
     }
 
     if (linked) {
+        FT_Library lib;
+        int err = FT_Init_FreeType(&lib);
+        if (err) {
+            // Not sure what the best exception to raise here is
+            // it's not an SDL exception, but I don't know of any
+            // pygame FreeType exceptions
+            PyErr_SetString(pgExc_SDLError, "FreeType could not be initialized");
+            return NULL;
+        }
         FT_Int major, minor, patch;
-        FT_Library_Version(inst->library, &major, &minor, &patch);
+        FT_Library_Version(lib, &major, &minor, &patch);
         return Py_BuildValue("iii", major, minor, patch);
     }
     else {
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -2046,6 +2046,18 @@ _ft_get_version(PyObject *self, PyObject *args, PyObject *kwargs)
     }
 
     if (linked) {
+        /*
+         * The FreeType library is being initialized here separately from the
+         * initialization of the `pygame.freetype` module so that the linked
+         * version can always be obtained. This does not affect the initialization
+         * state of `pygame.freetype` itself.
+         * 
+         * The reason this is needed is because if freetype has not been initialized,
+         * then a segmentation fault can happen. The alternative would be to return
+         * something predefined to mean something akin to "Unknown", but as this 
+         * function is meant for debugging purposes, it seems like a good idea to
+         * always be able to retrieve the linked FreeType version.
+        */
         FT_Library lib;
         int err = FT_Init_FreeType(&lib);
         if (err) {
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -2049,15 +2049,16 @@ _ft_get_version(PyObject *self, PyObject *args, PyObject *kwargs)
         /*
          * The FreeType library is being initialized here separately from the
          * initialization of the `pygame.freetype` module so that the linked
-         * version can always be obtained. This does not affect the initialization
-         * state of `pygame.freetype` itself.
-         * 
-         * The reason this is needed is because if freetype has not been initialized,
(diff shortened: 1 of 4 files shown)


Matiiss:
I don't know about this, until now it could show the version even if it wasn't initialized which kind of makes sense, so following that this should also just show the current version instead of raising an exception. However, this makes sense too, it just doesn't conform to the previous versions.

oddbookworm:
@Matiiss here is what `_ft_get_version` looked like before #3379:
```
static PyObject *
_ft_get_version(PyObject *self, PyObject *_null)
{
    /* Return the linked FreeType2 version */
    return Py_BuildValue("iii", FREETYPE_MAJOR, FREETYPE_MINOR,
                         FREETYPE_PATCH);
}
```
Although, I'm pretty sure that comment is wrong and those macros are the compiled version. That code still exists after the PR merge from above, but the linked version requires a `FreeTypeInstance`, and the one initialized in pygame doesn't get filled until `pygame.freetype.init()` is called. So, the compiled version can be returned without raising an exception, but the linked version cannot currently.

oddbookworm:
In fact, here's the declaration of those macros in the `freetype.h` header in the FreeType2 lib files:
```
  /**************************************************************************
   *
   * @enum:
   *   FREETYPE_XXX
   *
   * @description:
   *   These three macros identify the FreeType source code version.  Use
   *   @FT_Library_Version to access them at runtime.
   *
   * @values:
   *   FREETYPE_MAJOR ::
   *     The major version number.
   *   FREETYPE_MINOR ::
   *     The minor version number.
   *   FREETYPE_PATCH ::
   *     The patch level.
   *
   * @note:
   *   The version number of FreeType if built as a dynamic link library with
   *   the 'libtool' package is _not_ controlled by these three macros.
   *
   */
   ```

**Comment by oddbookworm** [comment]
But I think I have actually found a solution, so I'll add another commit

File: src_c/_freetype.c [hunk 1]
```diff
@@ -2046,8 +2046,18 @@ _ft_get_version(PyObject *self, PyObject *args, PyObject *kwargs)
...
 
     if (linked) {
+        FT_Library lib;
+        int err = FT_Init_FreeType(&lib);
+        if (err) {
+            // Not sure what the best exception to raise here is
+            // it's not an SDL exception, but I don't know of any
+            // pygame FreeType exceptions
+            PyErr_SetString(pgExc_SDLError,
+                            "FreeType could not be initialized");
```
MyreMylar, line 2056: how about `PyExc_RuntimeError` ?

MyreMylar:
Would it not be better just to do:

```
if (linked) {
    if (!inst) {
            PyErr_SetString(pgExc_SDLError,
                            "FreeType is not be initialized");
        return NULL;
    }
```

Right now it seems like this would be leaking memory from initialising two versions of freetype and the one in `get_version()` not being cleaned up?

oddbookworm:
@MyreMylar I see two routes:
1) The way I have done it (with fixes to make sure freetype gets cleaned up properly in `get_version`)
2) If `pygame.freetype` is not initialized, either throw an error or return something like `(0, 0, 0)`

With option 1, I understand that it's not the most ideal solution because of the extra initialization of freetype. But, I also would like to see this method behave similarly to the other getters, who don't need module initialization within pygame to work. Which is why I opted to pursue that route (and then forgetting to clean up after myself). Also considering that this method is intended to help diagnose problems with pygame, I think it would be bad form to throw an error or return a dummy value just because a module isn't initialized.

I'd like to hear your thoughts about it. In the meantime, I'll go ahead and push a commit hopefully cleaning up properly (for some reason, FreeType2's docs are really hard to read for me)

src_c/_freetype.c
```diff
@@ -2046,8 +2046,21 @@ _ft_get_version(PyObject *self, PyObject *args, PyObject *kwargs)
     }
 
     if (linked) {
```
Starbuck5, line 2048: Just coming back to this PR, sorry it's been a while.

I'd like you to add a comment in the code explaining why you're instantiating a whole freetype library here. I believe the answer is so you can query version independently of the pygame-freetype module being initialized.

I just think it could be helpful to someone in the future wondering why this is written this way.

Starbuck5:
Should we add a test that freetype.get_version() can be called before pygame.freetype.init?

test/freetype_test.py
```diff
@@ -1794,6 +1794,14 @@ def test_get_error(self):
 
         self.assertIsNone(error_msg)
 
+    def test_get_version(self):
+        # Test that get_version() can be called before init()
+        # Also tests that the compiled and linked versions are
+        # the same
+        ft.quit()
+
+        self.assertEquals(ft.get_version(linked=False), ft.get_version(linked=True))
```
Starbuck5, line 1803: The compiled and linked versions could potentially be different, and that would be fine. We’ve seen this with SDL at least— people have had different compiled and linked versions, as long as the linked version is more recent.

Starbuck5 (approved):
Good stuff, approved.

MyreMylar (approved):
LGTM 👍

PR #3568, opened 2022-11-16
Add "window" attribute to DROPFILE and DROPTEXT event

yunline:
image

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -1253,6 +1253,8 @@ dict_from_event(SDL_Event *event)
         case PGE_WINDOWDISPLAYCHANGED:
         case SDL_TEXTEDITING:
         case SDL_TEXTINPUT:
+        case SDL_DROPTEXT:
+        case SDL_DROPFILE:
         case SDL_MOUSEWHEEL:
         case SDL_KEYDOWN:
         case SDL_KEYUP:
@@ -1260,18 +1262,11 @@ dict_from_event(SDL_Event *event)
         case SDL_MOUSEBUTTONDOWN:
         case SDL_MOUSEBUTTONUP: {
             SDL_Window *window = SDL_GetWindowFromID(event->window.windowID);
-            PyObject *pgWindow;
-            if (!window ||
-                !(pgWindow = SDL_GetWindowData(window, "pg_window"))) {
-                pgWindow = Py_None;
-            }
-            Py_INCREF(pgWindow);
-            _pg_insobj(dict, "window", pgWindow);
-            break;
-        }
-        case SDL_DROPTEXT:
-        case SDL_DROPFILE:{
-            SDL_Window *window = SDL_GetWindowFromID(event->drop.windowID);
+            if (event->type==SDL_DROPTEXT ||
+                event->type==SDL_DROPFILE){
+                window = SDL_GetWindowFromID(event->drop.windowID);
+                }
+                
             PyObject *pgWindow;
             if (!window ||
                 !(pgWindow = SDL_GetWindowData(window, "pg_window"))) {
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -1262,11 +1262,10 @@ dict_from_event(SDL_Event *event)
         case SDL_MOUSEBUTTONDOWN:
         case SDL_MOUSEBUTTONUP: {
             SDL_Window *window = SDL_GetWindowFromID(event->window.windowID);
-            if (event->type==SDL_DROPTEXT ||
-                event->type==SDL_DROPFILE){
+            if (event->type == SDL_DROPTEXT || event->type == SDL_DROPFILE) {
                 window = SDL_GetWindowFromID(event->drop.windowID);
-                }
-                
+            }
+
             PyObject *pgWindow;
             if (!window ||
                 !(pgWindow = SDL_GetWindowData(window, "pg_window"))) {
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -1262,6 +1262,8 @@ dict_from_event(SDL_Event *event)
             window = SDL_GetWindowFromID(event->text.windowID);
             break;
         }
+        case SDL_DROPBEGIN:
+        case SDL_DROPCOMPLETE:
         case SDL_DROPTEXT:
         case SDL_DROPFILE: {
             window = SDL_GetWindowFromID(event->drop.windowID);
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -1231,7 +1231,7 @@ dict_from_event(SDL_Event *event)
     }  /* switch (event->type) */
     /* Events that dont have any attributes are not handled in switch
      * statement */
-
+    SDL_Window *window;
     switch (event->type) {
         case PGE_WINDOWSHOWN:
         case PGE_WINDOWHIDDEN:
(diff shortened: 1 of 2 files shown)


src_c/event.c
```diff
@@ -1269,6 +1269,18 @@ dict_from_event(SDL_Event *event)
...
+            SDL_Window *window = SDL_GetWindowFromID(event->drop.windowID);
+            PyObject *pgWindow;
+            if (!window ||
+                !(pgWindow = SDL_GetWindowData(window, "pg_window"))) {
+                pgWindow = Py_None;
+            }
+            Py_INCREF(pgWindow);
+            _pg_insobj(dict, "window", pgWindow);
+            break;
+        }
```
Matiiss, line 1283: This chunk of code is a duplicate of code like 2 lines above, you could just add those new cases above the existing piece instead of duplicating it.

Matiiss (changes requested):

src_c/event.c
Matiiss, line 1283: This chunk of code (minus the new `case`) is a duplicate of code like 2 lines above, you could just add those new cases above the existing piece instead of duplicating it.

Matiiss (changes requested):

src_c/event.c
```diff
@@ -1253,25 +1253,20 @@ dict_from_event(SDL_Event *event)
         case PGE_WINDOWDISPLAYCHANGED:
         case SDL_TEXTEDITING:
         case SDL_TEXTINPUT:
+        case SDL_DROPTEXT:
+        case SDL_DROPFILE:
```
Matiiss, line 1257: I'm fairly certain that adding only these two lines is all you had to do.

Matiiss (changes requested):

src_c/event.c
yunline: image  
If this if statement is not added, the "window" attribute will be always "None".
image

yunline: @Matiiss

Matiiss: Yeah, I understand, well, I'm not qualified enough to evaluate this, sorry, besides it would just be a suggestion anyways.

MyreMylar (approved):
Looks pretty good to me. 

[Relevant SDL event documentation page](https://wiki.libsdl.org/SDL_DropEvent). @ankith26 should review though as they are the `event.c` expert.

MyreMylar:
Looks like the linter is complaining:

```
'clang-format' failed. Please run: python setup.py format Do you have the latest version of clang-format?
```

if you have `clang-format` installed ([Install guide](https://www.dynamsoft.com/codepool/vscode-format-c-code-windows-linux.html#:~:text=How%20to%20Install%20Clang%2DFormat,Visual%20Studio%20Code%20for%20Windows.)) (python package: https://pypi.org/project/clang-format/) it should just be a quick:

```
clang-format src_c/event.c -i

```

from the pygame directory in the console.

ankith26 (commented):
Hi, thanks for the PR! :balloon: 

This PR looks almost good, but for the sake of better robustness I think more changes need doing around these bits of code. If you are interested, it'd be nice if you could do these changes in this PR. Otherwise we could also merge this PR first, and me or someone else could do a follow up PR later.

The issue here is that `event->window.windowID` is only guaranteed to be the correct way for accessing `windowID` attributes of `WINDOW*` events. It just so happens that this also works for some more events because of how the internal union within `SDL_Event` is defined. And as you figured out, this way of accessing is wrong for drop events (and one must not attempt to do this either, because this is basically undefined behavior and you will get a garbage id because the wrong memory location is being accessed)

MyreMylar:
> Hi, thanks for the PR! 🎈
> 
> This PR looks almost good, but for the sake of better robustness I think more changes need doing around these bits of code. If you are interested, it'd be nice if you could do these changes in this PR. Otherwise we could also merge this PR first, and me or someone else could do a follow up PR later.
> 
> The issue here is that `event->window.windowID` is only guaranteed to be the correct way for accessing `windowID` attributes of `WINDOW*` events. It just so happens that this also works for some more events because of how the internal union within `SDL_Event` is defined. And as you figured out, this way of accessing is wrong for drop events (and one must not attempt to do this either, because this is basically undefined behavior and you will get a garbage id because the wrong memory location is being accessed)

Initialise window to 0 then slap `SDL_GetWindowFromID(event->window.windowID);` into an else block?

yunline:
> Hi, thanks for the PR! 🎈
> 
> This PR looks almost good, but for the sake of better robustness I think more changes need doing around these bits of code. If you are interested, it'd be nice if you could do these changes in this PR. Otherwise we could also merge this PR first, and me or someone else could do a follow up PR later.
> 
> The issue here is that `event->window.windowID` is only guaranteed to be the correct way for accessing `windowID` attributes of `WINDOW*` events. It just so happens that this also works for some more events because of how the internal union within `SDL_Event` is defined. And as you figured out, this way of accessing is wrong for drop events (and one must not attempt to do this either, because this is basically undefined behavior and you will get a garbage id because the wrong memory location is being accessed)

I add 2 lines here  
image  
And I get this   
image   
Maybe there is a bug in SDL library.

yunline:
破案了，这就去和sdl对线

MyreMylar:
I mean the code should be like this:

```
SDL_Window *window = 0;
if (event->type == SDL_DROPTEXT || event->type == SDL_DROPFILE) {
    window = SDL_GetWindowFromID(event->drop.windowID);
}
else{
    window = SDL_GetWindowFromID(event->window.windowID);
}

PyObject *pgWindow;
if (!window || !(pgWindow = SDL_GetWindowData(window, "pg_window"))) {
    pgWindow = Py_None;
}
Py_INCREF(pgWindow);
_pg_insobj(dict, "window", pgWindow);
break;
```

ankith26:
> Maybe there is a bug in SDL library.

This is not a bug, rather it's a side effect of how the underlying `union` is implemented in the event struct. With unions, you can have multiple names to access the same memory block.

> I mean the code should be like this

Yeah, the code you suggested will do for now, but for robustness I'd like the splitting of this large switch statement into the correct sub-categories, something like (untested code).
```c
    SDL_Window *window;
    switch (event->type) {
        case PGE_WINDOWSHOWN:
        case PGE_WINDOWHIDDEN:
        case PGE_WINDOWEXPOSED:
        case PGE_WINDOWMOVED:
        case PGE_WINDOWRESIZED:
        case PGE_WINDOWSIZECHANGED:
        case PGE_WINDOWMINIMIZED:
        case PGE_WINDOWMAXIMIZED:
        case PGE_WINDOWRESTORED:
        case PGE_WINDOWENTER:
        case PGE_WINDOWLEAVE:
        case PGE_WINDOWFOCUSGAINED:
        case PGE_WINDOWFOCUSLOST:
        case PGE_WINDOWCLOSE:
        case PGE_WINDOWTAKEFOCUS:
        case PGE_WINDOWHITTEST:
        case PGE_WINDOWICCPROFCHANGED:
        case PGE_WINDOWDISPLAYCHANGED:
            window = SDL_GetWindowFromID(event->window.windowID);
            break;
        case SDL_TEXTEDITING:
        case SDL_TEXTINPUT:
            window = SDL_GetWindowFromID(event->edit.windowID);
            break;
        case SDL_DROPTEXT:
        case SDL_DROPFILE:
            window = SDL_GetWindowFromID(event->drop.windowID);
            break;
        case SDL_MOUSEWHEEL:
            window = SDL_GetWindowFromID(event->wheel.windowID);
            break;
        case SDL_KEYDOWN:
        case SDL_KEYUP:
            window = SDL_GetWindowFromID(event->key.windowID);
            break;
        case SDL_MOUSEMOTION:
            window = SDL_GetWindowFromID(event->motion.windowID);
            break;
        case SDL_MOUSEBUTTONDOWN:
        case SDL_MOUSEBUTTONUP: 
            window = SDL_GetWindowFromID(event->button.windowID);
            break;
        default:
            /* Event doesn't have a "window" attribute. Can exit the function early */
            return dict;
    }
    /* Handle "window" here */
    PyObject *pgWindow;
    f (!window ||
        !(pgWindow = SDL_GetWindowData(window, "pg_window"))) {
        pgWindow = Py_None;
    }
    Py_INCREF(pgWindow);
    _pg_insobj(dict, "window", pgWindow);
    return dict;
```

yunline:
> > I mean the code should be like this
> 
> Yeah, the code you suggested will do for now, but for robustness I'd like the splitting of this large switch statement into the correct sub-categories, something like (untested code).
> 
> ```c
>     SDL_Window *window;
>     switch (event->type) {
>         case PGE_WINDOWSHOWN:
>         case PGE_WINDOWHIDDEN:
>         case PGE_WINDOWEXPOSED:
>         case PGE_WINDOWMOVED:
>         case PGE_WINDOWRESIZED:
>         case PGE_WINDOWSIZECHANGED:
>         case PGE_WINDOWMINIMIZED:
>         case PGE_WINDOWMAXIMIZED:
>         case PGE_WINDOWRESTORED:
>         case PGE_WINDOWENTER:
>         case PGE_WINDOWLEAVE:
>         case PGE_WINDOWFOCUSGAINED:
>         case PGE_WINDOWFOCUSLOST:
>         case PGE_WINDOWCLOSE:
>         case PGE_WINDOWTAKEFOCUS:
>         case PGE_WINDOWHITTEST:
>         case PGE_WINDOWICCPROFCHANGED:
>         case PGE_WINDOWDISPLAYCHANGED:
>             window = SDL_GetWindowFromID(event->window.windowID);
>             break;
>         case SDL_TEXTEDITING:
>         case SDL_TEXTINPUT:
>             window = SDL_GetWindowFromID(event->edit.windowID);
>             break;
>         case SDL_DROPTEXT:
>         case SDL_DROPFILE:
>             window = SDL_GetWindowFromID(event->drop.windowID);
>             break;
>         case SDL_MOUSEWHEEL:
>             window = SDL_GetWindowFromID(event->wheel.windowID);
>             break;
>         case SDL_KEYDOWN:
>         case SDL_KEYUP:
>             window = SDL_GetWindowFromID(event->key.windowID);
>             break;
>         case SDL_MOUSEMOTION:
>             window = SDL_GetWindowFromID(event->motion.windowID);
>             break;
>         case SDL_MOUSEBUTTONDOWN:
>         case SDL_MOUSEBUTTONUP: 
>             window = SDL_GetWindowFromID(event->button.windowID);
>             break;
>         default:
>             /* Event doesn't have a "window" attribute. Can exit the function early */
>             return dict;
>     }
>     /* Handle "window" here */
>     PyObject *pgWindow;
>     if (!window ||
>         !(pgWindow = SDL_GetWindowData(window, "pg_window"))) {
>         pgWindow = Py_None;
>     }
>     Py_INCREF(pgWindow);
>     _pg_insobj(dict, "window", pgWindow);
>     return dict;
> ```

Thank you. The code is modified now.
I also added `SDL_FINGERMOTION`, `SDL_FINGERDOWN` and `SDL_FINGERUP` into the `case`.
```c
case SDL_FINGERMOTION:
case SDL_FINGERDOWN:
case SDL_FINGERUP: {
    window = SDL_GetWindowFromID(event->tfinger.windowID);
    break;
}
```

yunline:
Why? I don't understand. Is it a problem about wrong SDL version? But the [sdl wiki](https://wiki.libsdl.org/SDL_TouchFingerEvent) indicates that `'This structure is available since SDL 2.0.0.'`

MyreMylar:
Looks like that build uses SDL 2.0.8 released in 2018, and `windowID` was added in 2019.

see:

https://discourse.libsdl.org/t/sdl-2-0-8-released/23957

and:

https://github.com/libsdl-org/SDL/commit/d5ec735a3342beaf7a5f89f1e7e5eb618cf57f53
https://github.com/libsdl-org/SDL/commit/2fb71ac52d3bda3e48a921567e793dca9106842b
https://github.com/libsdl-org/SDL/commit/109cbd6e8bd621ed4d7df59d70071cf1ba2d4e42

src_c/event.c
```diff
@@ -1250,26 +1250,59 @@ dict_from_event(SDL_Event *event)
...
+        case SDL_TEXTEDITING: {
+            window = SDL_GetWindowFromID(event->edit.windowID);
+            break;
+        }
+        case SDL_TEXTINPUT: {
+            window = SDL_GetWindowFromID(event->text.windowID);
+            break;
+        }
+        case SDL_DROPTEXT:
+        case SDL_DROPFILE: {
```
ankith26, line 1266: Would be nice if you also added `SDL_DROPBEGIN` and `SDL_DROPCOMPLETE` here for completeness sake

ankith26 (approved):
Thanks again for the PR! I have left a minor addition suggestion, apart from that this PR is good to go! :tada: :cake:

src_c/event.c
yunline: Thanks. I forgot that.

yunline: image

Starbuck5:
Thanks for making this PR! I notice that the `events` documentation describes what attributes each type of event should have, so that should be updated for this.

Edit here: https://github.com/pygame/pygame/blob/main/docs/reST/ref/event.rst
It would also be good to put in a "versionchanged" tag for this in the documentation there.

ankith26:
Not sure about the latest commit that documents this should be a part of this PR. This is a feature that has been left undocumented so far because it has to do with `_sdl2` which isn't public and stable.

Starbuck5:
>  Not sure about the latest commit that documents this should be a part of this PR. This is a feature that has been left undocumented so far because it has to do with _sdl2 which isn't public and stable.

Lets just go ahead and put it out. The underlying SDL event provides the window data, so when we're all caught up with SDL this will be like this anyways.

You okay with that @ankith26 ?

Starbuck5:
Ok @yunline, I talked to Ankith about it and he was still against documenting this.
So I went ahead and reverted the commit on your PR branch, it can be merged when the tests pass.

Nice work!

PR #3570, opened 2022-11-17
test arc

yujyuj:
added testing logic for test_arc


novialriptide:
It seems as if the `todo_test_arc()` function was created as a placeholder during its development and was forgotten to be removed.

MyreMylar:
> It seems as if the `todo_test_arc()` function was created as a placeholder during its development and was forgotten to be removed.

All the `todo_` prefixed tests were originally created as reminders to make a good test for this later, however in the case of arc, I agree that this function already looks to be well tested by the existing entire class of tests.

I don't think this test is adding anything new to the arc testing, and it's also saving a file for no obvious reason.

This PR could be repurposed to jut delete the redundant `todo_test_` chunk instead if you like @yujyuj

test/draw_test.py
```diff
@@ -6186,9 +6186,42 @@ def test_arc__invalid_color_formats(self):
             with self.assertRaises(TypeError):
                 bounds_rect = self.draw_arc(**kwargs)
 
-    def todo_test_arc(self):
+    def test_arc(self):
         """Ensure draw arc works correctly."""
-        self.fail()
```
MyreMylar, line 6191: Instead, just remove:

```
def todo_test_arc(self):
     self.fail()
```

MyreMylar (changes requested):
I think just removing the `todo_test_arc(self)` statement entirely is the way to go here. `draw.arc()` has a large number of tests in the class this test is being added to.

test/draw_test.py
Starbuck5: Isn't that what they did?

test/draw_test.py
```diff
@@ -6186,9 +6186,42 @@ def test_arc__invalid_color_formats(self):
...
+        surface.fill(black)
+
+        # rectangle that contains for the ellipse arc.
+        # 0 pixel from left, 0 pixel from top
+        # 80 pixels wide, 40 pixels high
+        rect = (0, 0, 80, 40)
+
+        # angle of the arc in radians
+        start_angle = 0.0
+        stop_angle = 3.14
```
novialriptide, line 6206: There should be a builtin function that defines pi.

test/draw_test.py
MyreMylar: they also added a test here - but draw arc is already extensively tested.

Issue #3573, opened 2022-11-18
pygame.Vector2() does not specify that it returns a Vector with the values (0, 0) in the documentation


novialriptide:
I think this should be clarified in the documentation somewhere.

PR #3574, opened 2022-11-18
Add vector zero constructor clarification

novialriptide:
Fixes: https://github.com/pygame/pygame/issues/3573

I am not entirely sure if this is a good solution or not since I cannot find any other documentation methods similar to this. But since `Vector2()` will consistently always return `Vector2(0, 0)`, I think this is a good implementation.

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -66,7 +66,7 @@ Multiple coordinates can be set using slices or swizzling
 .. class:: Vector2
 
    | :sl:`a 2-Dimensional Vector`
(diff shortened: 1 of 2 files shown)


ankith26 (commented):
This is just a personal preference thing, but I think this can be clarified in the doc body instead. But I'm also fine with what this PR does so I'll go ahead and approve, thankies :partying_face:

ankith26 (approved):

MyreMylar (approved):
Seems fairly clear to me, so I think we can add this.

PR #3577, opened 2022-11-20
Adjust surface masks to create a proper 24 bit Surface

Starbuck5:
The SDL 2.26.0 pre-release rejects a 24 bit Surface created with this mask, which is the right call, as it needs 32 bits of space. Older SDL went ahead and created a 32 bit surface anyway, which it probably shouldn't have.


ankith26 (approved):
This change makes sense, though I'm slightly concerned since this is behaviour changing. But I don't think it would bread much "real code" so I'll go ahead and approve. Thanks :tada: 

Also unrelated to this PR, but these mask numbers would probably make more visual sense when written in hex.

MyreMylar (approved):
LGTM 👍

Issue #3580, opened 2022-11-20
Docs: minor improvements to `pygame.math` docs


Matiiss:
On line 16 (in math.rst) and few lines below the examples showing different numerical operations don't have a space between the operator and the operand:
![image](https://user-images.githubusercontent.com/83066658/202904272-b9bb0e86-2d0a-4d56-b7a0-ea92fc0a72cb.png)

Even though these are just docs, I feel like PEP 8 still sort of applies to them, so could just add a space there, for example, `vec + vec` instead of the current `vec+vec`

Also, they support some operations with a sequence like `vec - sequence`, perhaps, that should be added too.

There's also this rather confusing line
![image](https://user-images.githubusercontent.com/83066658/202905309-8d0b8a4f-f9dc-4c0b-ab30-d82804c5f55e.png)
Was the thing that changed a required import that is not required anymore or was the thing that changed this import that is now required (which it's not)?

Matiiss:
Also, they support `vec - sequence`, perhaps, that should be added too

MyreMylar:
> Was the thing that changed a required import that is not required anymore or was the thing that changed this import that is now required (which it's not)?

You used to have to create vectors like:

```
pygame.math.Vector2(x, y)
```


but now you can do:

```
pygame.Vector2(x, y)
```

because they were imported into the root pygame namespace.

Matiiss:
I get that, but all the other _Changed in <version>:_ use a sort of additive language (what I'd describe it as), so for example there's _Changed in pygame 1.9.4: Removed experimental notice._, it doesn't say _Changed in pygame 1.9.4: Experimental notice exists._ although that could also be valid since the thing that change in the version was the existence of that notice and so it does not exist anymore. Similarly it says that what changed in pygame 1.9.4 was the required import meaning it's not required anymore, however, since the other versions list changes as something that is as of that change this _pygame.math required import_ could be interpreted as the thing that's new, so you must `import pygame.math` now. 

Instead I suggest changing it to _Changed in pygame 1.9.4: pygame.math import not required. More convenient pygame.Vector2 and pygame.Vector3._

MyreMylar:
Sure, I thought you didn't know what had changed at that time, so was just clarifying that. The language of the docs could happily be updated.

PR #3581, opened 2022-11-20
Minor `pygame.math` doc enhancements

Matiiss:
Fixes https://github.com/pygame/pygame/issues/3580

Not sure if 
> Also, they support some operations with a sequence like `vec - sequence`, perhaps, that should be added too.

should be "fixed" too.


MyreMylar (approved):
LGTM 👍

ankith26 (approved):
LGTM! Thanks :taco:

PR #3582, opened 2022-11-22
Run the teyit lint tool on our unittests for consistency + latest black version linting updates

ankith26:
While scrolling through the python 3.12 notes I came across a mention of this tool that helps make unit tests more consistent. By making use of `unittest` API instead of pure python equivalents, we get better error reports for failing tests.


MyreMylar:
This is failing because teyit requires python 3.9 and the version of ubuntu out linter CI uses is on python 3.8.

ankith26:
Good catch. I'm now wondering whether teyit should be a part of `setup.py lint/format` or if it should be an external standalone thing.

Probably updating this teyit test to run on ubuntu 22.04 instead of 20.04 will fix the CI fail

MyreMylar:
> Good catch. I'm now wondering whether teyit should be a part of `setup.py lint/format` or if it should be an external standalone thing.
> 
> Probably updating this teyit test to run on ubuntu 22.04 instead of 20.04 will fix the CI fail

I think we should just update the linting test runner to ubuntu 22.04. I'll make a simple PR for that.

MyreMylar:
Related:
https://github.com/pygame/pygame/pull/3689

MyreMylar:
Toggling this to re-run teyint with maybe upgraded main? May need to merge this branch with main to actually pick up that update.

MyreMylar:
Hooray, looks like this worked. 

Now we just need to fix all the issues with the tests.

MyreMylar (approved):
OK. All changes I added here are just from running the latest black & pylint on the code and now teyit works too 👍

PR #3583, opened 2022-11-23
Fix build errors on win_arm64

cgohlke:
This PR disables x86/AMD64 specific SIMD functions when building for Windows on ARM64.

### DIFF
--- a/src_c/scale.h
+++ b/src_c/scale.h
@@ -29,10 +29,11 @@
 #if !defined(SCALE_HEADER)
 #define SCALE_HEADER
 
-#if (defined(__GNUC__) &&                                      \
-     ((defined(__x86_64__) && !defined(_NO_MMX_FOR_X86_64)) || \
-      defined(__i386__))) ||                                   \
-    (defined(MS_WIN32) && !((defined(_M_X64) || defined(_M_ARM64)) && defined(_NO_MMX_FOR_X86_64)))
+#if (defined(__GNUC__) &&                                             \
+     ((defined(__x86_64__) && !defined(_NO_MMX_FOR_X86_64)) ||        \
+      defined(__i386__))) ||                                          \
+    (defined(MS_WIN32) && !((defined(_M_X64) || defined(_M_ARM64)) && \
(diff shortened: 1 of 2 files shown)


src_c/scale.h
```diff
@@ -29,10 +29,11 @@
...
 
-#if (defined(__GNUC__) &&                                      \
-     ((defined(__x86_64__) && !defined(_NO_MMX_FOR_X86_64)) || \
-      defined(__i386__))) ||                                   \
-    (defined(MS_WIN32) && !(defined(_M_X64) && defined(_NO_MMX_FOR_X86_64)))
+#if (defined(__GNUC__) &&                                             \
+     ((defined(__x86_64__) && !defined(_NO_MMX_FOR_X86_64)) ||        \
+      defined(__i386__))) ||                                          \
+    (defined(MS_WIN32) && !((defined(_M_X64) || defined(_M_ARM64)) && \
+                            defined(_NO_MMX_FOR_X86_64)))
```
Starbuck5, line 36: Sorry for the delay in handling this @cgohlke 

I'm having a hard time understanding what this guard is doing.

With some bubble pushing, the last bit of logic becomes
`(!defined(_M_X64) && !defined(_M_ARM64)) || !defined(_NO_MMX_FOR_X86_64)`
And if that is True and its on Windows, it will include the mmx support.

So if _NO_MMX_FOR_X86_64 is undefined, it will always be True and therefore the MMX stuff will always be included?

cgohlke: The patch doesn't change the logic much, just added Windows on ARM64 to the list of platforms for which, when `_NO_MMX_FOR_X86_64` is defined, MMX will not be used. So yes, `-D_NO_MMX_FOR_X86_64` will need to be included in `Setup`, but at least with this patch the definition does what is expected. Without this patch, MMX will always be included and of course fail on Windows on ARM64.

Starbuck5: Okay, that makes sense. Thanks.

Starbuck5 (approved):
Thanks!

MyreMylar (approved):
LGTM 👍

PR #3587, opened 2022-11-24
Add VScode? _headers/ folder to gitignore

Starbuck5:
VS code keeps putting a "_headers" directory at my local pygame top level with a bunch of header files in it, and I haven't been able to figure out what's doing it-- this is practically unsearchable.

So could we just add it to the gitignore please? 😄

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ dist
 *.egg-info/
 *.so
 __pycache__
+_headers/*


ankith26 (approved):
Back when I had windows (le good ol days) VSC did the same for me, after I installed some extension, probably an rst doc file extension.

Definitely a good idea to add it to gitignore, thankss :pizza:

PR #3588, opened 2022-11-24
Consolidate SSE2 and NEON code paths

Starbuck5:
Uses macros / functions / defines to eliminate the large amount of code duplication between SSE2 and NEON paths in the big blit switch statement.

* ~~it compiles and passes tests locally, haven't given it a full test because it's getting late and I just wanted to get it out into the world~~
* TESTED: does as advertises, the dispatching still works. This is not a huge performance regression in disguise. (Although I could only test the SSE2 dispatch on my hardware)

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -45,7 +45,7 @@
 #endif
 #endif
 
-#ifdef PG_ENABLE_ARM_NEON
+#if PG_ENABLE_ARM_NEON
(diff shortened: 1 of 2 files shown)


MyreMylar (approved):
LGTM! Nice clean up, I like it best when we can delete a bunch of code.

ankith26 (approved):
Nice cleanups, thankies! :cake:

PR #3594, opened 2022-11-26
Fix `list_cameras` being empty on linux

ankith26:
This is a quickfix of a linux-only 2.1.3 regression I introduced in an older commit.

`close()` should have been only called on success of a camera existing. Otherwise it errors and the function exits early and stops looking for more cameras

### DIFF
--- a/src_c/camera_v4l2.c
+++ b/src_c/camera_v4l2.c
@@ -56,20 +56,21 @@ v4l2_list_cameras(int *num_devices)
             }
         }
 
-        fd = open(device, O_RDONLY);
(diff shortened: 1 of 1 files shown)


Starbuck5:
@oddbookworm Could you test this?

oddbookworm:
Works for me on Manjaro with Python 3.10.8. I tried to test on 3.11, but for whatever reason 3.11 seems broken for me because of some SSL problem. Still does not work for me on WSL with Ubuntu, but I'm going to chalk that down to the problem that my graphics driver has been having

Starbuck5 (approved):

MyreMylar (approved):
OK 👍

PR #3595, opened 2022-11-26
Removing redudancy in regex

TheVroum:
Fixing square brackets class redudancy : \w class is strictly containing a-z0-9 in itself (also contain A-Z and _).

Classes equivalences were also tested on regex101.com.


MyreMylar (approved):
LGTM 👍

nthykier (approved):

MyreMylar (approved):

Issue #3596, opened 2022-11-27
`font` and `freetype` fail to import on MacOS < 12


seungjec:
**Environment:**

You can get some of this info from the text that pops up in the console when you run a pygame program.

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Mac, Catalina(10.15.7), Intel
- **Python version** (e.g. 3.7.9, 3.8.5) : Python 3.11.0
- **SDL version** (e.g. SDL 2.0.12): SDL 2.0.22
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): pygame 2.1.3.dev8
- **Relevant hardware** (e.g. if reporting a bug about a controller, tell us the brand & name of it):

**Current behavior:**

Describe how the bug manifests.

When I try to use the font module in pygame, I get an error as below.
All other modules of pygame are loaded normally after import pygame, but the font module is not loaded.
However, In Windows under the same conditions, it works normally.

**[Code]**
import pygame
pygame.init()
pygame.font.get_fonts()

**[Error Message]**
pygame 2.1.3.dev8 (SDL 2.0.22, Python 3.11.0)
Hello from the pygame community. https://www.pygame.org/contribute.html
/Code/shooting/main.py:13: RuntimeWarning: use font: dlopen(/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pygame/font.cpython-311-darwin.so, 2): Library not loaded: @loader_path/libfreetype.6.dylib
  Referenced from: /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pygame/.dylibs/libfreetype.6.dylib
  Reason: image not found
(ImportError: dlopen(/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pygame/font.cpython-311-darwin.so, 2): Library not loaded: @loader_path/libfreetype.6.dylib
  Referenced from: /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/pygame/.dylibs/libfreetype.6.dylib
  Reason: image not found)
  pygame.font.get_fonts()

**Screenshots**

If applicable, copy and paste screenshots to help explain your problem.
![image](https://user-images.githubusercontent.com/16385675/204134168-80538bb1-985e-4c1c-acea-f2a71e9c8831.png)

Starbuck5:
[...]

PR #3601, opened 2022-11-30
Workaround macos < 12 font issue: revert harfbuzz on macos

ankith26:
Opened to put a non-ideal but hopefully working fix for #3596


MyreMylar (approved):
Approved for now, unless a better fix comes along

nthykier (approved):

PR #3602, opened 2022-12-01
draw mask: Added unit tests for test_arc and test_angle

yujyuj:
Added unit test for 3 functions:

1. `test_arc()` in test/draw_test.py
2. `test_convert()` in test/surface_test.py
3. `test_angel()` in test/mask_test.py


novialriptide:
Why not add more commits to the already existing PR?
https://github.com/pygame/pygame/pull/3570

test/surface_test.py
```diff
@@ -1613,32 +1613,28 @@ def test_opaque_destination_blit_with_set_alpha(self):
...
+        flags = 0
+        depth = 32
+        pygame.display.init()
 
-        # __doc__ (as of 2008-08-02) for pygame.surface.Surface.convert:
+        try:
+            convert_surface = pygame.display.set_mode(size)
+            surface = pygame.surface.Surface.convert(convert_surface)
+            self.assertIsNot(surface, convert_surface)
+            self.assertNotEqual(surface.get_size(), size)
```
MyreMylar, line 1629: This check does not work, it is failing in the test runners.

Converting a surface does not change the pixel size of it. I'm also a little unsure why you are starting with a display surface. The most frequent purpose of `convert()` is to change a regular surface from one that it is unlike the display surface to one that is much closer to it for faster blitting. 

I would probably start with a 16 or 8 bit surface and verify that it had been changed to have the same depth as the display in a call to convert with no arguments. Then try it with a surface argument and see if the new depths match etc.

MyreMylar (changes requested):
These tests currently don't pass the test runners, or even work locally.

illume (approved):
👍 thanks

illume:
Merged in https://github.com/pygame/pygame/pull/3753

PR #3605, opened 2022-12-01
Add support for ppc64le

mgiessing:
Enable manylinux2014 builds for ppc64le. This is no change content-wise.

Might discuss the possibility to build wheels using github actions and qemu

Signed-off-by: mgiessing <marvin.giessing@gmail.com>

### DIFF
--- a/buildconfig/manylinux-build/Makefile
+++ b/buildconfig/manylinux-build/Makefile
@@ -21,7 +21,10 @@ wheels-manylinux2014-x86:
 wheels-manylinux2014-aarch64:
 	docker run --rm -v ${REPO_ROOT}:/io pygame/manylinux2014_base_aarch64 /io/buildconfig/manylinux-build/build-wheels.sh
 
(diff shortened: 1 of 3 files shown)


ankith26:
Hello! First of all, thanks for the PR!

This is a good patch, but I don't think we should be merging this as of now. pygame does have source code support for ppc64le, and the main purpose is to help with downstream debian packaging. I think adding ppc64le "official wheels support" is going to increase maintenance burden in the long term. Another reason to be not merging this as of now is that most of python ecosystem doesn't have wheels support for ppc64le (most notably numpy doesn't, which happens to be one of pygames soft dependencies). So ppc64le users would generally have to rely on distro packages to get their stuff, and it is more likely they also get pygame the same way.

We can surely revisit this PR if the python ecosystem at large starts shipping ppc64le wheels.

illume (approved):
Thanks. I’ve been interested in ppc stuff in the past and worked on previous patches for it in pygame. Wheels for it would be a nice addition and save some compilation time too!

> Might discuss the possibility to build wheels using github actions and qemu

I would be appreciative of any help you could provide with this, and on any other ppc related topics.

Thanks again! 🎉

PR #3609, opened 2022-12-06
Don't set adaptive vsync by default

fladd:
Adaptive vsync should only be set when the "vsync" parameter of display.set_mode() is -1. If it is 1, normal vsync should be set.

See also: https://github.com/pygame/pygame/issues/735#issuecomment-1331435781

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1168,18 +1168,22 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                where the driver lies to us because the user has configured
                vsync to be always on or always off, or vsync is on by default
                for the whole desktop because of wayland GL compositing. */
-            if (vsync) {
+            if (vsync == -1) {
                 if (SDL_GL_SetSwapInterval(-1) != 0) {
                     if (PyErr_WarnEx(PyExc_Warning,
                                      "adaptive vsync for OpenGL not "
-                                     "available, trying regular",
+                                     "available",
                                      1) != 0) {
                         _display_state_cleanup(state);
                         goto DESTROY_WINDOW;
                     }
+                }
+            }
+            else {
+                if (vsync == 1) {
                     if (SDL_GL_SetSwapInterval(1) != 0) {
                         if (PyErr_WarnEx(PyExc_Warning,
-                                         "regular vsync for OpenGL *also* not "
+                                         "regular vsync for OpenGL not "
                                          "available",
                                          1) != 0) {
                             _display_state_cleanup(state);
@@ -1187,9 +1191,9 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                         }
                     }
                 }
-            }
-            else {
-                SDL_GL_SetSwapInterval(0);
+                else {
+                    SDL_GL_SetSwapInterval(0);
+                }
             }
         }
         else {


Starbuck5:
Hello,

I hadn't heard of adaptive vsync before, reading up on it [here](https://www.khronos.org/opengl/wiki/Swap_Interval#Adaptive_Vsync), it seems like a reasonable default. Why do you think it shouldn't?

> Looking at the code in display.c, it seems that when using an OpenGL context, currently vsync=1 always tries to set adaptive vsync first, and uses this if successful. This, however, should not be the default behaviour! It is currently not clear to me how to force normal vsync (i.e. have the the same behaviour as in Pygame 1).

I thought vsync was not a thing at all in pygame 1. I don't see any calls to SDL_GL_SetSwapInterval in [display.c from 1.9.6](https://github.com/pygame/pygame/blob/1.9.6/src_c/display.c), as well as the `vsync` parameter being newer than that.

I guess you must be writing a game in OpenGL with pygame windowing?

fladd:
With normal vsync, you can block on the vertical retrace with OpenGL. That is if you ask the video card to flip buffers, then write something on the back buffer, and then make a call to `GLFinish`, then this call will only return once the back buffer is ready for writing, hence when the flip has just happened. This is a crucial mechanism for knowing when something you plotted actually ended up being drawn on the screen.

Adaptive vsync breaks this important feature, which is commonly needed in an experimental research setting (Psychology, Neuroscience), where you want to have total control over the exact time when something is being presented to a participant in a study.

In Pygame 1, you could do `pygame.display.gl_set_attribute(pygame.GL_SWAP_CONTROL, 1)` to ask for proper vsync and hence many large research software projects used Pygame as their base (Expyriment, PsychoPy, OpenSesame, to just name a few).

In Pygame 2 you cannot ask for proper vsync, but proper vsync is always a fallback from adaptive vsync. Hence, when you run this on a video card that supports adaptive vsync you will get that, which is not what you want, and which will break proper functioning of the research software mentioned above.

MyreMylar:
I think adaptive vsync is still a good default setting, but we should have an option to set the vsync interval. This is the relevant code chunk below:

```
if (vsync) {
                if (SDL_GL_SetSwapInterval(-1) != 0) {
                    if (PyErr_WarnEx(PyExc_Warning,
                                     "adaptive vsync for OpenGL not "
                                     "available, trying regular",
                                     1) != 0) {
                        _display_state_cleanup(state);
                        goto DESTROY_WINDOW;
                    }
                    if (SDL_GL_SetSwapInterval(1) != 0) {
                        if (PyErr_WarnEx(PyExc_Warning,
                                         "regular vsync for OpenGL *also* not "
                                         "available",
                                         1) != 0) {
                            _display_state_cleanup(state);
                            goto DESTROY_WINDOW;
                        }
                    }
                }
            }
```

What we would need is a variable for `vsync_interval` instead of the hard coded `-1` and `1`. Then we could either have it as another parameter for display, or perhaps better as a hint like `PYGAME_FORCE_SCALE`. I would call it something like  `PYGAME_GL_VSYNC_SWAP_INTERVAL` and document it as having the three possible values (0, 1, -1) from the SDL docs here: https://wiki.libsdl.org/SDL2/SDL_GL_SetSwapInterval.

fladd:
Seeing that the new Pygame version has still not been released, it would be great to have this still included before release, as this is a crucial feature for timing-accurate applications.

MyreMylar (changes requested):
I think the issue with this patch is that it takes an existing parameter and changes it's behaviour. This will just break a lot of existing users code.

I agree that we should let people have the option of preferring different vsync modes when creating a display - but it needs to be implemented in a away that preserves the original default in pygame 2, which is adaptive vsync when the vsync paramter is set to True (aka as 1).

I think a hint/environment variable would probably be the best way forward here. You can see the current environment variables for pygame here:
https://www.pygame.org/docs/ref/pygame.html

Along with how they are set.  Something like:

```
// At top of set mode function
int swap_interval = -1;
char* swap_interval_env;

...

// in the vsync part
swap_interval_env = SDL_getenv("PYGAME_GL_VSYNC_SWAP_INTERVAL");
if (swap_interval_env != NULL) {
    swap_interval = SDL_atoi(swap_interval_env);
}
if (vsync) {
    if (SDL_GL_SetSwapInterval(swap_interval) != 0) {
        if (PyErr_WarnEx(PyExc_Warning,
                         "preferred vsync for openGL not "
                         "available, trying regular vsync",
                         1) != 0) {
            _display_state_cleanup(state);
            goto DESTROY_WINDOW;
        }
        if (SDL_GL_SetSwapInterval(1) != 0) {
            if (PyErr_WarnEx(PyExc_Warning,
                             "regular vsync for OpenGL *also* not "
                             "available",
                             1) != 0) {
                _display_state_cleanup(state);
                goto DESTROY_WINDOW;
            }
        }
    }
}
else {
    SDL_GL_SetSwapInterval(0);
}
```

Then you would try it by doing:

```
os.environ['PYGAME_GL_VSYNC_SWAP_INTERVAL'] = '1'
display_surf = pygame.display.set_mode((800, 600), vsync=True, flags=pygame.OPENGL)
```

Does that make sense?

Let me know if you would rather somebody else finish this off.

fladd:
@MyreMylar That's also fine by me, as long as there is the option to get the old Pygame1 behaviour back.

MyreMylar:
> @MyreMylar That's also fine by me, as long as there is the option to get the old Pygame1 behaviour back.

Yes, setting that environment variable I propose to "1" should do exactly that. 

Now, do you want to adapt this PR as I propose above - or would you prefer handing it off to someone else to finish?

illume:
This PR seems to fix a backwards compatibility issue so I’ll merge it.

@fladd can you please confirm… With this does your code work with no changes compared to pygame 1?

fladd:
@illume In theory yes. Is there maybe a new pre-build with this change so I could actually test this in practice?

illume:
What OS and python versions? There’s wheel files generated I could point you at.

If that, or compiling from source doesn’t work for you… I’ll ping you when the next pre-release is out.

fladd:
I can test on MacOs and Windows 10.

illume:
Each PR has wheels you can download and install with pip.

For windows, you can click here: https://ci.appveyor.com/project/pygame/pygame/builds/46219822
- Then click on the one for your python version and architecture, eg python 3.10 and x64
- Then click the architecture tab.
- Then click the wheel file to download

For Mac: download this https://github.com/pygame/pygame/suites/9698575591/artifacts/469479281, unzip it, and find the appropriate wheel file to `python -m pip install aaa.whl`

Otherwise, a new dev release will be out shortly.

PR #3610, opened 2022-12-11
Replace fast scaling algorithm with SDL_SoftStretch

Starbuck5:
For reduced code footprint and increased performance.

My testing shows a 40% speedup for normal scales as well as scales with destination surfaces.

The scale output does change, but I can only notice the difference when I switch between them in at the same location onscreen in an image editor.

For example:
normal
stretch
(For the record, the first one is the existing scale product, the second is one the new result)

Benchmark script:
```py
import pygame
import os
import time

pygame.init()

os.chdir("../../official_pygame/examples/data")
im1 = pygame.image.load("fist.png").convert_alpha()

start = time.time()
for _ in range(100000): #100000
    im2 = pygame.transform.scale_by(im1, 0.5)
print(time.time() - start)

pygame.quit()
raise SystemExit
```

I also fixed a bug where the destination surface could be the right bit depth but wrong pixel format. Before, it would stretch the image onto the dest surface even though the color information would be all messed up. Now, I use the error handling of SDL_SoftStretch to catch the format errors. This does change the error from a ValueError to a pygame.error, but I think this behavior is so rare that it doesn't need to be emulated.


Starbuck5:
Here is a compilation of files scaled in certain ways by the existing algorithm (normal) and the new algorithm (stretch):
[scale_normal.zip](https://github.com/pygame/pygame/files/10266464/scale_normal.zip)
[scale_stretch.zip](https://github.com/pygame/pygame/files/10266465/scale_stretch.zip)

MyreMylar:
One thing that concerns me about this is the comment on the SDL docs here:
https://wiki.libsdl.org/SDL2/SDL_SoftStretch

> **Remarks**
Please use [SDL_BlitScaled](https://wiki.libsdl.org/SDL2/SDL_BlitScaled)() instead.

Starbuck5:
> One thing that concerns me about this is the comment on the SDL docs here

I tried BlitScaled first, on that recommendation, but it does not do what we want. We want a copy, rather than a blit. I don't remember the exact details of why it didn't work, but it was related to the alpha of the scaled surface interacting with the alpha of the freshly created surface.

MyreMylar (approved):
I didn't quite see a 40% speedup - on my PC it was more like 10% but it was consistently faster across a range of sizes and pixel formats.

I also couldn't visually detect a change in output.

NB. I also had a peek at the more recent SDL_SoftStretchLinear for smoothscaling. Unfortunately it uses a different algorithm to our smooth scale so the results are often very different.

PR #3611, opened 2022-12-11
Use Python 3.11.0 rather than 3.11rc2

Starbuck5:
Just a little post-3.11-launch cleanup

No need to continue using 3.11rc2 when we could use 3.11.0.


MyreMylar (approved):
LGTM 👍

MightyJosip (approved):

PR #3613, opened 2022-12-12
Fixing docs for `pygame.MOUSEBUTTONDOWN` and `pygame.MOUSEBUTTONUP` events

### LINKED ISSUE #3612: `pygame.MOUSEBUTTONUP` and `pygame.MOUSEBUTTONDOWN` docs don't fully describe the behavior of those events when the mouse wheel is scrolled
Discovered by @lucysir\#3139 (https://github.com/kadir014) in discord. When the mouse wheel is scrolled fast, the docs only say this:

> The mouse wheel will generate pygame.MOUSEBUTTONDOWN and pygame.MOUSEBUTTONUP events when rolled. The button will be set to 4 when the wheel is rolled up, and to button 5 when the wheel is rolled down.

If the wheel is scrolled quickly enough, multiple events are created, and each one has an `event.button` attribute that is 2 higher than the previous.

sample code:
```
import pygame

pygame.init()

running = True
while running:
    print("===")
        if event.type == pygame.QUIT:
            running = False

        if event.type == pygame.MOUSEBUTTONUP:
            print(event.button)

    pygame.time.delay(100)
```
sample output
```
===
4
6
===
4
6
===
4
===
===
===
===
===
===
4
===
4
6
8
10
12
===
```

I plan on making a PR soon to update the documentation.
(ignore that github automatically linked something in the tag for sirlucy. That was not intentional)

oddbookworm:
This is a PR to fix #3612

### DIFF
--- a/docs/reST/ref/mouse.rst
+++ b/docs/reST/ref/mouse.rst
@@ -24,6 +24,16 @@ is moving many motion events will be placed on the queue. Mouse motion events
 that are not properly cleaned from the event queue are the primary reason the 
 event queue fills up.
 
+When using ``pygame.MOUSEBUTTONDOWN`` or ``pygame.MOUSEBUTTONUP`` events
+to track the scrolling of the mousewheel, if multiple events are registered
+in a frame, 2 will be added to each successive event's button attribute. So,
+if 3 events are registered during one frame when you scroll up, the first
+``pygame.MOUSEBUTTONDOWN`` event will have a button attribute of 4, the second
+will have a button attribute of 6, and the third will have a button attribute
+of 8. The same goes for ``pygame.MOUSEBUTTONUP`` events.
+During scrolling, both types of events will be generated, so to track
+scrolling, only handle one of these two types of events.
+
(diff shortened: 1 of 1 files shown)


ankith26:
Uhhhh, I'm not sure whether this is intended behaviour. It could be a bug

oddbookworm:
> Uhhhh, I'm not sure whether this is intended behaviour. It could be a bug

I thought the same, but the behavior is present as far back as `2.0.3`, so I was figuring that at this point fixing it might break someone's project. So I thought "let's at least document it"

oddbookworm:
Plus, with `pygame.MOUSEWHEEL` a thing now, isn't this functionality only being kept around for compat?

MyreMylar:
Should we close this now in favour of the fix?

oddbookworm:
> Should we close this now in favour of the fix?

I think so, so I'll close it now

oddbookworm:
Relevant fix https://github.com/pygame/pygame/pull/3642

PR #3617, opened 2022-12-15
Remove bit of MIR code

Starbuck5:
Just a tiny cleanup.

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -542,16 +542,6 @@ pg_get_wm_info(PyObject *self, PyObject *_null)
     PyDict_SetItemString(dict, "shell_surface", tmp);
     Py_DECREF(tmp);
 #endif
-#if defined(SDL_VIDEO_DRIVER_MIR) /* no longer available, left for API/ABI \
-                                     compatibility. Remove in 2.1! */
-    tmp = PyCapsule_New(info.info.mir.connection, "connection", NULL);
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
LGTM 👍

ankith26 (commented):
Seems like MIR was dropped in SDL 2.0.10, and 2.0.8 is still a "pygame supported SDL version". But IDK the specifics here so IG we could wait on this PR? (since it's a legacy code cleanup thing)

Starbuck5:
> Seems like MIR was dropped in SDL 2.0.10, and 2.0.8 is still a "pygame supported SDL version". But IDK the specifics here so IG we could wait on this PR? (since it's a legacy code cleanup thing)

Let's just drop it. The comment says to drop it in 2.1, pygame 2.1, which has already passed.
The only other mention of MIR within pygame calls it an untested and unsupported platform:
https://github.com/pygame/pygame/blob/403887dff6a3f5c607fabd6a1f64afcd87e451f3/src_c/display.c#L2146-L2147

So this is not a PR that removes MIR support. This is a PR that cleans up some dead code that's already unsupported.

ankith26:
I believe the 2.1 refers to the SDL version, given the code was likely copied from the SDL source with changes. OFC, SDL 2.1 never happened but this API/ABI is being fully removed in SDL3 instead

nthykier (approved):

PR #3618, opened 2022-12-15
Raise minimum SDL version to 2.0.8

Starbuck5:
Pygame works poorly (if at all) on these now outdated SDL versions, which I found when testing for #3239 

SDL 2.0.8 is a good target because it's the earliest that one sees in the wild these days, because it's default in an Ubuntu version (18.04?). SDL 2.0.8 was released 4.75 years ago.

So we might as well get rid of some ifdef clutter as well as be more realistic with what we say we support in the README.

pgcompat.c looks goofy now, because it's just empty. I'm leaving it in because we might use it to host SDL compat functions for _sdl2.video again in the future.


MyreMylar (approved):
Works for me. If you still wish to use an older version of SDL with pygame, older versions of pygame are still available.

ankith26 (approved):
LGTM! ty for the PR :dancers:

PR #3620, opened 2022-12-17
Correct set_(shifts/masks) docs and cleanup code

Starbuck5:
I was looking through Surface.c and display.c with Matt and I found some stuff to do.

- The docs were wrong-- the function raises a TypeError, not an AttributeError
- The docs should have the "deprecated" label
- The C code doesn't need to carefully parse args before doing an unconditional TypeError

I was tempted to do the full removal of these functions (they've been deprecated for a long time, probably not many people used them ever), but I decided to hold off on that. In the future, yes they will be removed.

### DIFF
--- a/docs/reST/ref/surface.rst
+++ b/docs/reST/ref/surface.rst
@@ -769,8 +769,10 @@
 
       This is not needed for normal pygame usage.
 
-      .. note:: In SDL2, the masks are read-only and accordingly this method will raise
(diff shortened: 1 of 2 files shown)


MyreMylar (approved):
LGTM 👍

ankith26 (approved):
LGTM! thanks :star2:

PR #3622, opened 2022-12-20
Deprecate gamma and _sdl2 window brightness functionality

### LINKED ISSUE #3600: Deprecate gamma functionality
SDL has removed it in SDL3 development already-- https://github.com/libsdl-org/SDL/pull/6617

We have stuff in display.c using `SDL_SetWindowGammaRamp`, those methods should be changed to emit deprecation warnings.

We have things using `SDL_SetWindowBrightness` and `SDL_GetWindowBrightness` in _sdl2.video, those should be removed.

SSS-Says-Snek:
Fixes #3600

### DIFF
--- a/buildconfig/stubs/pygame/_sdl2/video.pyi
+++ b/buildconfig/stubs/pygame/_sdl2/video.pyi
@@ -66,7 +66,6 @@ class Window:
     size: Iterable[int]
     position: Union[int, Iterable[int]]
     opacity: float
-    brightness: float
     display_index: int
     def set_modal_for(self, Window) -> None: ...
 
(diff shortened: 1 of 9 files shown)


Starbuck5 (commented):
You also need to remove from the video types and docs.

SSS-Says-Snek:
oops, I forgot, lemme do those things real quick

Starbuck5:
The lint is complaining that the docs changes caused a header file to generate differently.

Run `python setup.py docs` and commit the changes.

Starbuck5 (approved):
Thanks!

MyreMylar (approved):
LGTM. 

I don't image many people are using per application gamma control anymore anyway. From what I recall it doesn't work very well in SDL2 either.

EDIT: reading the linked SDL issue it seems like the only option these days is to change the gamma for the whole system - so it is no longer per application at the moment anyway.

PR #3624, opened 2022-12-21
Update __init__.py with different code reformatting

SmashedFrenzy16:
This commit reformats the print statements and reformats them, making it easier to read.

### DIFF
--- a/src_py/threads/__init__.py
+++ b/src_py/threads/__init__.py
@@ -103,14 +103,14 @@ def benchmark_workers(a_bench_func=None, the_data=None):
         wq = WorkerQueue(num_workers)
         t1 = time.time()
         for _ in range(20):
-            print(f"active count:{threading.active_count()}")
+            print("active count: {}").format(threading.active_count())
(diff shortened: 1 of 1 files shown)


oddbookworm:
You've got the `.format` on the wrong part of the statement. Also, f-strings are generally the "easier" way to read, rather than trying to parse the older `str.format` syntax. f-strings are also far faster than `str.format` (or any other string formatting/concatenation technique). Because pygame's minimal supported version also supports f-strings, I think this reversion to an antiquated, slower format is unnecessary.

SmashedFrenzy16:
Oh ok, thanks for the feedback @oddbookworm.

SmashedFrenzy16:
I have updated the .format() to be on the right side of the print statement now.

oddbookworm:
What I said still stands about using `str.format` rather than f-strings. I still am of the belief that going back to using the older, slower `str.format` syntax is not the appropriate move to make. It's generally harder to read, and it's slower than f-strings.

MyreMylar:
Agreed. This change would be a backwards step.

Issue #3632, opened 2022-12-26
Renaming (to|from)string C functions to (to|from)bytes


Matiiss:
Exactly what the title says, this entails renaming `tostring` and `fromstring` C functions in `image.c` to `tobytes` and `frombytes` since the string ones might get deprecated in the future on pygame Python's end so just to keep consistency with that.

PR #3633, opened 2022-12-26
Change all (to|from)string to (to|from)bytes in image.c

Matiiss:
Since (to|from)bytes is the preferred function it makes sense that C functions also use this name.

Fixes https://github.com/pygame/pygame/issues/3632

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -269,7 +269,7 @@ _debug_print128_num(__m128i var, const char *msg)
 
 /*
  * Generates an SSE vector useful for reordering a SSE vector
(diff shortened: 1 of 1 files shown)


Starbuck5:
I thought you would just rename the top level function, but if you're going this far you might as well go a little further

I see mentions of the old string style on lines 272, 840, 930, 1021 (of your branch).
Also the internal bytes PyObject for both functions is still named "string", that could be changed.

Starbuck5 (approved):
Thanks!

ankith26 (approved):
LGTM! ty for the PR :sunglasses:

PR #3637, opened 2022-12-28
Update __init__.py with better code formatting

SmashedFrenzy16:
This change makes the `print` statements easier to read.


MyreMylar (approved):
This seems fine, obviously it doesn't really add much - I believe this code file has never really been used in pygame and haa probably been up for deletion a few times.

In the interests of just closing off some PRs I will approve.

Starbuck5:
I'm going to close this, as Myre says this code isn't really used, so best not to touch it?

Also a lot of builds are failing because of a docs build issue, this PR can be brought up to not failing on that, but I believe it will be more effort than the benefit this PR brings.

PR #3639, opened 2022-12-28
Update __init__.py

SmashedFrenzy16:

Issue #3357, opened 2022-07-31
Remove obsolete vidcapture camera backend


Starbuck5:
Pygame's camera module, (src_py/camera.py) uses different strategies to try to get a working webcam backend.

At one point, videocapture / vidcapture was usable and used, another python library that could provide webcam functionality for Windows. However, that package is now obsolete (last updated in 2012), and we have a new camera backend for Windows now.

I'd like someone to remove this backend and relevant code.

- [ ] update documentation ( camera docs, documented env variable in "pygame" doc? )
- [ ] remove _camera_vidcapture.py source
- [ ] remove vidcapture support from camera.py
- [ ] figure out if there's something else I'm forgetting in this list that should be done

MyreMylar:
What version of Windows does the new version of the camera backend support? IIRC it was 8 or was it 8.1?

Googling online suggests Windows 8.1 is EOL from Jan 2023.

Starbuck5:
> What version of Windows does the new version of the camera backend support? IIRC it was 8 or was it 8.1?

It was unclear to me as I wrote it. It's been tested by Matiis, who runs some iteration of 8.

MyreMylar:
Yeah I got it working a while back, I think I just downloaded binary versions from that random python windows binary website on a university server - that is now being shut down. 

Since then I've updated to a new PC and then Windows 11 and the new backend has been working fine on both Windows 10 and Windows 11 so I'm pretty sure it is safe to deprecate the old backend for new versions. 

I expect the intersection of people who run pre-windows 8, managed to download the binary vidcapture backend before the website was shut down (or built it themselves from source) and still want to keep up to date on the latest versions of pygame is a very small number.

Starbuck5:
>  Hi. I would like to start contributing and this seems like a good first issue. Mind if I take it?

Sure, go ahead. I feel like a shill because I point this out a lot, but we have a `#contributing` channel on the pygame discord, if you want to ask questions in a more conversational format: https://discord.com/invite/ZuB2RySPRJ.

SmashedFrenzy16:
> At one point, videocapture / vidcapture was usable and used, another python library that could provide webcam functionality for Windows. However, that package is now obsolete (last updated in 2012), and we have a new camera backend for Windows now.

Should't we keep it for legacy computers? Many people still use operating systems like windows 7.

MyreMylar:
> > I'd like someone to remove this backend and relevant code.
> 
[...]

PR #3640, opened 2022-12-28
Update __init__.py

CodeMaster7000:


SmashedFrenzy16:
This is copying some of the changes that I made in PR #3639

SmashedFrenzy16:
So this PR is not necessary

SmashedFrenzy16:
At least, for the `__init__.py`, as that is not needed.

CodeMaster7000:
@SmashedFrenzy16 The phrasing is different from that in your PR. An additional message is also included.

SmashedFrenzy16:
@CodeMaster7000 you could have told me to update my PR instead. Anyways, you could always delete that file.

CodeMaster7000:
@SmashedFrenzy16 I honestly think it is fine for now. Once the change has been made, Pygame will most likely close or merge the PRs anyways.

SmashedFrenzy16:
@CodeMaster7000 is cannot be fine that you copy other people's pull requests to only get your merged. The best solution would be to delete the `__init__.py` file because you made unecessary extra commits in that change.

SmashedFrenzy16 (changes requested):
Delete `__init__.py` from this PR as it is not necessary.

novialriptide:
> Delete `__init__.py` as it is not necessary.

Deleting the file would break pygame

SmashedFrenzy16:
No I mean delete it from this PR

src_py/__init__.py
```diff
@@ -339,7 +339,7 @@ def __color_reduce(c):
             ver, *get_sdl_version() + sys.version_info[0:3]
         )
     )
-    print("Hello from the pygame community. https://www.pygame.org/contribute.html")
+    print("Hello from the Pygame community! To contribute, visit https://www.pygame.org/contribute.html. We appreciate you taking the time to do so.")
```
novialriptide, line 342: Lowercase p is usually preferred when referring to "pygame" as per PEP8
https://peps.python.org/pep-0008/#package-and-module-names
image

oddbookworm:
@SmashedFrenzy16 @CodeMaster7000 the comments of a pull request is not an appropriate place for you two to argue about whose pull request should contain which changes. If you have a disagreement about something, please sort that out in private. Nobody reviewing your PR's wants to sort through your argument to see if there's any important information there. Also, if someone reviews your PR and it's clear changes to the PR are needed, you won't get a merge unless you fix them. Same goes for failing actions (like the lint check, which means you need to run `python setup.py format`)

CodeMaster7000:
I understand and I apologise. It might have got a bit out of hand with 2 PRs having overlapping changes.

SmashedFrenzy16:
I also understand and apologise.

Starbuck5:
Hey @SmashedFrenzy16 and @CodeMaster7000, I believe these changes make the content slightly worse. I talked to some other contributors and I believe we're in agreement about this, so I'm going to close this and #3639 (similar content)

You seem like you want to help out here, which is admirable. If you'd like to tackle an issue (perhaps you can work together on this, you seem to know each other), #3357 could be good. Alternatively, you could try to enhance the docs. I think one promising area would be function names that are written out instead of linked. If some function description references another function it should be possible to click on that and go to the other function, but I don't believe this is the case in all cases.

Issue #3635, opened 2022-12-28
Allow mixer.music.set_pos() for sounds created from a sound array 


ArztKlein:
**Description**
If you create a sound array, convert it to a sound then try to use set_pos() on it, it will raise an error stating `pygame.error: set_pos unsupported for this codec`.

The documentation states that `MP3`, `OGG` and `MOD` are supported, however there is no mention of sounds created from sound arrays.
> Other file formats are unsupported. Newer versions of SDL_mixer have better positioning support than earlier ones. An SDLError is raised if a particular format does not support positioning.

Matiiss:
The issue you are talking about seems to originate from setting position without having loaded a music file.
```py
import pygame

pygame.mixer.init()
pygame.mixer.music.set_pos(0.5)
```

`pygame.Sound` objects don't have a `set_pos` attribute at all, so if you make a sound object from `pygame.sndarray` you can't set its position, you'd have to probably convert it to a file-like object and then load that file-like object using `pygame.mixer.music.load`

Other than that you have found a bug in pygame's code, where you get this exception while nothing is even loaded.

Starbuck5:
> you'd have to probably convert it to a file-like object and then load that file-like object

I'm not sure that would be possible.

The file-like object would have to be encoded somehow, mixer.music doesn't seem to have any way of accepting raw data. And encoding raw sound into WAV / mp3 / whatever is not something we have the means to support.

Matiiss:
@Starbuck5 does that mean there is no built-in way to (`pygame.mixer.music.`)`play` music obtained from a sound array? It could maybe be suggested that `pygame.mixer.Sound` has a `set_pos` method? Does SDL support that?

PR #3641, opened 2022-12-29
Replace `music.set_pos` current error message with `SDL_GetError()`

Matiiss:
Fixes https://github.com/pygame/pygame/issues/3635

Essentially provides a clearer error message when calling `pygame.mixer.music.set_pos` without having loaded any music (file(s))

### DIFF
--- a/src_c/music.c
+++ b/src_c/music.c
@@ -256,7 +256,7 @@ music_set_pos(PyObject *self, PyObject *arg)
     Py_END_ALLOW_THREADS;
 
     if (position_set == -1)
-        return RAISE(pgExc_SDLError, "set_pos unsupported for this codec");
+        return RAISE(pgExc_SDLError, SDL_GetError());
 
     Py_RETURN_NONE;
 }


Starbuck5 (commented):
I wonder if `Mix_SetMusicPosition` sets an error itself that we should just use instead?

Matiiss:
@Starbuck5 Using `return RAISE(pgExc_SDLError, SDL_GetError());` the error message is `pygame.error: Music isn't playing`, but I don't know what it would say if the codec was actually invalid.

Matiiss:
So, I tried with all available audio file formats that I could convert to using https://cloudconvert.com/ and it either couldn't load the file (raising a rather confusing message: `pygame.error: ModPlug_Load failed`) or it could set the position of the file. I don't know about the specifics of those formats but it seemed that `.aiff` could be both loaded and set position, however, it's not documented here: https://wiki.libsdl.org/SDL_mixer/Mix_SetMusicPosition, I also came to the conclusion that pygame docs about this function should be updated I guess, other than that I suppose the current message could be just whatever `SDL_GetError()` returns.

Matiiss:
Does this require documentation updates?

Starbuck5 (approved):
Looks good, thanks!

MyreMylar (approved):
Works for me.

PR #3642, opened 2022-12-29
Fix event button attr for fast mousewheel

ankith26:
PR to fix the underlying issue of #3612 
I think we should be fixing this issue instead of documenting broken legacy behaviour

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -569,19 +569,19 @@ pg_event_filter(void *_, SDL_Event *event)
         newupevent.button.clicks = 1;
         newupevent.button.which = event->button.which;
 
-        if (event->wheel.y > 0) {
-            newdownevent.button.button = PGM_BUTTON_WHEELUP | PGM_BUTTON_KEEP;
-            newupevent.button.button = PGM_BUTTON_WHEELUP | PGM_BUTTON_KEEP;
-        }
-        else {
-            newdownevent.button.button =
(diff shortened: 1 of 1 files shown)


Starbuck5:
Tested, works great.

Should we add a note to the docs about this changed behavior?

MyreMylar:
> Tested, works great.
> 
> Should we add a note to the docs about this changed behavior?

I don't think we should add it to the docs in this case, as the wheel up generating specifically a 4 and 5 `.button` value in the event is mainly backwards compatibility behaviour with SDL1 and neither the historic numbers (4 & 5), the recent broken ones ( 4,6,8,10,etc & 5,7,9,11 etc), or the now returned to normal ones (4 & 5) have ever been in the docs. The recent behaviour with adding 2 is also not consistent with historic pygame behaviour.

 The 'correct' modern behaviour now for handling mouse wheels is the MOUSEWHEEL event and if you are using the supported `pygame.BUTTON_WHEELUP` & `BUTTON_WHEELDOWN` constants those have always been set to 4 & 5. 

I think this just qualifies as a straight up bug for 99% of users and documenting it is just going to cause confusion for people who had no idea about the strange behaviour somewhat 'behind the scenes'.

Starbuck5:
> I don't think we should add it to the docs in this case

Ok.

Starbuck5 (approved):
Thanks for tracking this down!

PR #3648, opened 2022-12-31
Update sprite.rst

Librichard:


oddbookworm:
This is an inaccurate change. It's talking about `pygame.sprite.Group`, not `pygame.sprite.Sprite`

MyreMylar:
> This is an inaccurate change. It's talking about `pygame.sprite.Group`, not `pygame.sprite.Sprite`

I'd disagree. It is the Group classes that do that actual blitting here. See:

https://github.com/pygame/pygame/blob/f477c2a0ac893367ed0c7cd32de2e6b593e3c09d/src_py/sprite.py#L540-L559

Matiiss:
I would also change
> requires that each Sprite have a `Surface.image` attribute and a `Surface.rect`. 

to
> requires that each Sprite has an `image` and a `rect` attribute.

because what's written there currently doesn't make much sense.

docs/reST/ref/sprite.rst
```diff
@@ -19,7 +19,7 @@ of objects in the game. There is also a base Group class that simply stores
 sprites. A game could create new types of Group classes that operate on
 specially customized Sprite instances they contain.
 
-The basic Sprite class can draw the Sprites it contains to a Surface. The
+The basic Group class can draw the Sprites it contains to a Surface. The
 ``Group.draw()`` method requires that each Sprite have a ``Surface.image``
 attribute and a ``Surface.rect``. The ``Group.clear()`` method requires these
```
Matiiss, line 24: I would also change
> requires that each Sprite have a `Surface.image` attribute and a `Surface.rect`. 

to
> requires that each Sprite has an `image` and a `rect` attribute.

because what's written there currently doesn't make much sense.

Matiiss, line 24: I would also change
> requires that each Sprite have a `Surface.image` attribute and a `Surface.rect`. 

to
> requires that each Sprite has an `image` and a `rect` attribute.

because what's written there currently doesn't make much sense.

Matiiss (changes requested):
I would also change the wording about attributes.

More of a general thing but parentheses should be removed from method names, otherwise although the intent is clear, adding parentheses means calling and thus means it's not quite correct to call them methods anymore since they have been "called" and unless they return methods themselves it doesn't make sense.
That doesn't apply to signatures and when they are referenced on their own.

docs/reST/ref/sprite.rst
MyreMylar: agreed.

MyreMylar:
> More of a general thing but parentheses should be removed from method names...

I don't understand this comment. The official Python docs always add seem to add parentheses after method names - see: https://docs.python.org/3/library/array.html#array.array for an example. I think it is a generally helpful signal to identify things that are method calls versus attributes. 

Perhaps you could explain what you mean?

Matiiss:
@MyreMylar I suppose that's more of a preference but when reading code I'm mentally executing it, so if I see a name followed by the verb _call_ I understand that the name references some function. Since docs use parentheses everywhere for functions and it can be understood that _call `name()`_ means that `name` itself is a function not call the returned value of `name()` I guess there's not much to be added but rather removed from my review :)

illume (approved):
It’s a good fix. Thanks.

illume:
The conversation on this PR is a good example of how not to review a single word fix. It’s almost comical :)

Sorry you had to go through that @Librichard

PR #3649, opened 2023-01-01
Fix the memory leak

### LINKED ISSUE #3621: Memory leak when calling pygame.image.load() per frame.
**Environment:**

Windows 10 (Ameliorated Edition, Professional Edition)
Python 3.9.6
SDL 2.0.22
Pygame 2.1.3.dev8

**Current behavior:**

When calling pygame.image.load() every frame, even if not assigned to a variable, pygame leaks roughly 36 bytes of RAM every frame.
It attempts to regulate itself, but after running for roughly 4 hours the program gives up and lets it go up.

**Expected behavior:**

When calling pygame.image.load(), it should probably be expected that if the value is not assigned to a variable, it should not take up any RAM. 

**Screenshots**

memleak
image

**Steps to reproduce:**

1) import pygame
2) make a while loop that runs pygame.image.load() and then an image value

**Test code**

```
import pygame
while True:
    pygame.image.load("potato.png")
```

Starbuck5:
Use static memory for temporary file extension storage, so it doesn't memory leak. This fine because of the GIL.

Without this patch every loader in pygame (besides pygame.freetype.Font) leaks some memory to store the file extension.

I introduced this memory leak a while back because I assumed `pgRWops_ReleaseObject` actually was called to release an SDL_RWop that we create. However, this is not the case. I'd like to get rid of `pgRWops_ReleaseObject` entirely, it's only called in one place, and could be replaced by an instance of `SDL_RWclose`. To keep this PR simple I will keep that change out of this.

Fixes #3621


maxqxam:
Nice job, fast debugging and all.
Best regards

ankith26 (commented):
Another idea: remove `pgRWops_GetFileExtension` completely and change the signature of `pgRWops_FromFileObject` to something like 
`static SDL_RWops *pgRWops_FromFileObject(PyObject *obj, char **ext)`

Starbuck5:
> Another idea: remove pgRWops_GetFileExtension completely and change the signature of pgRWops_FromFileObject to something like
static SDL_RWops *pgRWops_FromFileObject(PyObject *obj, char **ext)

This was my previous idea.
However, it requires a good bit of code change so I didn't favor it.

MyreMylar:
Maybe I am going insane, but shouldn't we be doing something like wrapping `SDL_RWclose()` in a `pg_RWclose()` and clearing the 'hidden' file extension data in that if it exists? I'm not sure if all RWops are using the hidden data for file extensions, but this would seem to be the approved method for using the hidden data and then clearing it up, would it not?

Something like:

```
int pg_RWclose(SDL_RWops *context){
    if (rw->hidden.unknown.data1 != 0){
         free rw->hidden.unknown.data1;
         rw->hidden.unknown.data1 = 0;
    }
    return SDL_RWclose();
}
```

Starbuck5:
> Maybe I am going insane, but shouldn't we be doing something like wrapping SDL_RWclose() in a pg_RWclose() and clearing the 'hidden' file extension data in that if it exists? I'm not sure if all RWops are using the hidden data for file extensions, but this would seem to be the approved method for using the hidden data and then clearing it up, would it not?

We don't call SDL_RWclose to close the majority of our RWops. Instead, we use the `freesrc` parameter, which uses the `_rw_close` function pointer in the struct. HOWEVER, we can't change that function pointer because you're only supposed to do that when creating a custom thing, we're using the non-custom "File RWop". I tried changing it, and SDL called bloody murder.

MyreMylar:
OK, so the underlying issue is that we are adding custom data to our RWOp after a call to:

```
rw = SDL_RWFromFile(encoded, "rb");
```

which you aren't supposed to do - but we still need the file extension.

Looking at the two places the file extension is used, it appears to be used immediately and is then no longer required. There seems to be no point in us storing it attached to the RWOp at all as the extension string is created and then used immediately within the two functions where `pgRWops_GetFileExtension()` is called.

Would it work to just return the extension string from the getter using data from the python object and not involve the RW at all? Something like:

```
char *
pgRWops_GetFileExtension(PyObject *obj)
{
    PyObject *oencoded;
    char *encoded = NULL;
    if (!obj) {
        // forward any errors
        return NULL;
    }

    oencoded = pg_EncodeString(obj, "UTF-8", NULL, NULL);
    if (!oencoded || oencoded == Py_None) {
        /* if oencoded is NULL, we are forwarding an error. If it is None, the
         * object passed was not a bytes/string/pathlib object so handling of
         * that is done after this function, exit early here */
        Py_XDECREF(oencoded);
        return NULL;
    }

    encoded = PyBytes_AS_STRING(oencoded);
    if (encoded ) {
        char *extension = NULL;
        char *ext = strrchr(encoded, '.');
        if (ext && strlen(ext) > 1) {
            ext++;
            extension = malloc(strlen(ext) + 1);
            if (!extension) {
                return (SDL_RWops *)PyErr_NoMemory();
            }
            strcpy(extension, ext);
        }
        Py_DECREF(oencoded);
        return extension;
    }
```

Then each call:

```
ext = pgRWops_GetFileExtension(obj);

....

// after 'ext' is used in each function
free(ext);
```

Or we could have `pgRWops_GetFileExtension()` calls require a char[] buffer passed in of a suitable size like you have done with the static data.

Sorry to keep banging on - it's partly me not having a good understanding of the RWOps code, and partly me having a strong dislike of little bits of static memory squirreled about the place :)

Starbuck5:
I'd rather not duplicate the string processing code in another method. 

One of my earlier plans was to make a variant of pgRWops_FromObject that would accept a char buffer, similar to what you're suggesting. However, then you need to have another parameter for char length and another out parameter to read whether the RWop had a file extension to show you.

I think I'm coming around to Ankith's idea
`static SDL_RWops *pgRWops_FromFileObject(PyObject *obj, char **ext)`

It's simpler than my idea for adding parameters:
- Pass NULL if you don't care about the extension
- It can set NULL or a valid char* (discriminating whether an extension was found or not)
- Just have to free it afterwards
- It allows the deletion of pgRWops_GetFileExtension (simplify the API?)

Starbuck5:
Alright, implemented a new solution from your feedback.

Test run with allocating and deallocating extension memory printed out:
```py
>>> pygame.image.load("Capture.PNG")
ALLOCATING
DEALLOCATING
<Surface(126x70x32 SW)>
>>> xdp = "official_pygame/examples/data"
>>> pygame.mixer.music.load(xdp + "/house_lo.mp3")
ALLOCATING
DEALLOCATING
>>> pygame.mixer.Sound(xdp + "/house_lo.mp3")
<pygame.mixer.Sound object at 0x00000257690DA0F0>
>>> pygame.font.Font(None, 32)
<pygame.font.Font object at 0x000002577FCCCDE0>
>>> import pygame
>>> pygame.freetype.Font(None)
Font('C:\Users\charl\AppData\Local\Programs\Python\Python39\lib\site-packages\pygame\freesansbold.ttf')
```

MyreMylar (approved):
Yes, I like this much better. Well done 👍

src_c/music.c
```diff
@@ -391,16 +391,20 @@ _load_music(PyObject *obj, char *namehint)
         return NULL;
```
nthykier, line 391: I think `ext` is now leaked here as a consequence of this change.

src_c/rwobject.c
```diff
@@ -694,21 +688,20 @@ _rwops_from_pystr(PyObject *obj)
...
-                return (SDL_RWops *)PyErr_NoMemory();
+        /* If a valid extptr has been passed, populate it with a dynamically
+         * allocated field for the file extension. */
+        if (extptr) {
+            char *ext = strrchr(encoded, '.');
+            if (ext && strlen(ext) > 1) {
+                ext++;
+                *extptr = malloc(strlen(ext) + 1);
+                if (!(*extptr)) {
+                    return (SDL_RWops *)PyErr_NoMemory();
```
nthykier, line 699: This return does not seem to have a `Py_DECREF(oencoded);`, but there is one in the happy path.  I suspect this is a leak-on-error case

nthykier: Also, shouldn't we clean up `rw` as well in case of error?

src_c/music.c
Starbuck5: `ext` will always be left as NULL if the `rw` is NULL, because `ext` is only allocated in RWobject when it's returning a valid non-NULL `RWop`.

src_c/rwobject.c
Starbuck5: Good spot! Pushed a commit to resolve this.

Starbuck5:
Hey @nthykier, cool to see you around. Maybe I've said this to you before, but the pygame contributors have a chat on the pygame community discord (invite link: https://discord.com/invite/pygame), so I should invite you if you haven't heard about that. Come say hi in the contributing channel if you want.

Issue #3650, opened 2023-01-01
Port SDL_mixer metadata functionality


Starbuck5:
SDL_mixer 2.6 introduces some new metadata getter features for music.

See:
https://wiki.libsdl.org/SDL_mixer/Mix_GetMusicAlbumTag
https://wiki.libsdl.org/SDL_mixer/Mix_GetMusicArtistTag
https://wiki.libsdl.org/SDL_mixer/Mix_GetMusicCopyrightTag
https://wiki.libsdl.org/SDL_mixer/Mix_GetMusicTitle || https://wiki.libsdl.org/SDL_mixer/Mix_GetMusicTitleTag

I believe we're on SDL_mixer 2.6 on our platforms, so this functionality could be cool to see. Imagine writing a music player and getting this sort of metadata from the actual files, rather than having to store them someplace else.

These could be ported as individual functions, or maybe they be one ~`get_metadata`  function.

Matiiss:
I'll do this

PR #3652, opened 2023-01-02
Implementing `pygame.mixer.music.get_metadata`

Matiiss:
Fixes https://github.com/pygame/pygame/issues/3650


Matiiss:
Should docs mention the different behavior due to unsupported(?) `SDL_mixer` versions?

Starbuck5:
>  Should docs mention the different behavior due to unsupported(?) SDL_mixer versions?

Yes.

Matiiss (commented):
LGTM 👍

MyreMylar (approved):
LGTM 👍  works on my machine and matches the metadata I see in my file browser.

src_c/music.c
```diff
@@ -496,6 +496,66 @@ music_queue(PyObject *self, PyObject *args, PyObject *keywds)
...
+
+    if (obj) {
+        music = _load_music(obj, namehint);
+        if (!music) {
+            return NULL;
+        }
+    }
+    else if (namehint) {
+        PyErr_SetString(
+            PyExc_RuntimeError,
```
Starbuck5, line 523: Runtime error is strange here. Maybe a pygame.error or a ValueError?

docs/reST/ref/music.rst
```diff
@@ -272,3 +272,29 @@ MP3 in most cases.
...
+   
+   Refer to the :func:`pygame.mixer.music.load` function for arguments regarding specifying a file or a file-like object 
+   whose metadata you want to retrieve. For this function all arguments are optional, 
+   however, specifying only the ``namehint`` will raise an exception.
+   
+   Since the underlying functionality was introduced in version 2.6.0 of SDL_mixer,
+   calling this function with an older version of SDL_mixer will return a dictionary
+   with all values being set to empty strings. You can find your version of SDL_mixer
+   by using :func:`pygame.mixer.get_sdl_mixer_version`.
+
```
Starbuck5, line 297: Maybe it would be nice to have an example metadata dictionary shown in the docs so people can easily see the format?

src_c/music.c
```diff
@@ -496,6 +496,66 @@ music_queue(PyObject *self, PyObject *args, PyObject *keywds)
...
+    if (!music) {
+        return RAISE(pgExc_SDLError, "music not loaded");
+    }
+
+    meta_dict = Py_BuildValue("{ss ss ss ss}", "title", title, "album", album,
+                              "artist", artist, "copyright", copyright);
+
+    if (!current_music) {
+        Mix_FreeMusic(music);
+    }
```
Starbuck5, line 554: Won't this leak memory if there is music playing and the user still specifies a file to look at?

Matiiss: Without freeing music it will certainly leak memory, so freeing it is necessary. If current music is playing it won't touch it, it will only free the loaded music, I could do a bit more testing though.

docs/reST/ref/music.rst
Matiiss: Alright, will do

src_c/music.c
Matiiss: I wasn't sure what type to use since the issue is that one of the optional arguments is required to use the other one, however, it's possible to pass only the other one since they're both optional. A `pygame.error` could cover this case I guess.

src_c/music.c
Starbuck5: I'm not sure you get what I'm saying.

Environment:
Music is playing (so current_music is not NULL)
User asks for pygame.mixer.music.get_metadata("another_file.mp3")

The `music` in get_metadata is only freed if current_music is NULL, so the function leaks memory when called in this configuration.

PR #3655, opened 2023-01-04
Change casting to `list` to using `copy` in `pygame.sprite`

Matiiss:
Remove `list(` and the closing `)` where the argument is already a list

### DIFF
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -688,7 +688,7 @@ class OrderedUpdates(RenderUpdates):
         RenderUpdates.__init__(self, *sprites)
 
     def sprites(self):
-        return list(self._spritelist)
+        return self._spritelist
 
     def add_internal(self, sprite, layer=None):
         RenderUpdates.add_internal(self, sprite)
@@ -834,7 +834,7 @@ class LayeredUpdates(AbstractGroup):
         LayeredUpdates.sprites(): return sprites
 
         """
-        return list(self._spritelist)
+        return self._spritelist
 
     def draw(self, surface):
         """draw all sprites in the right order onto the passed surface
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -688,7 +688,7 @@ class OrderedUpdates(RenderUpdates):
         RenderUpdates.__init__(self, *sprites)
 
     def sprites(self):
-        return self._spritelist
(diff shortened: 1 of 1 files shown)


Starbuck5:
The existing code made a copy of the list, this does not, so I think this could manifest some edge case?
Regardless, it's not the same semantic.

```py
>>> a = []
>>> id(a)
2502817516544
>>> b = list(a)
>>> id(b)
2502821134784
```

Matiiss:
Alright, in that case I will change them to use `copy` method instead as that seems to be faster:

```py
import time


def timer(func):
    def inner(*args, **kwargs):
        start = time.perf_counter()
        func(*args, **kwargs)
        print(f"{func.__name__}: {time.perf_counter() - start}")
    return inner


N = 10_000_000
lst = [1, 2, 3]


@timer
def type_cast():
    for _ in range(N):
        list(lst)


@timer
def copy():
    for _ in range(N):
        lst.copy()


type_cast()
copy()
```

Results:
```
type_cast: 1.2166718000080436
copy: 1.03610479994677
```

MyreMylar:
OK, one thing about this. OrderedUpdates is, I believe, an out of date class added to work around the behaviour of dictionaries in Python 3.5.

Since Python 3.6 dictionaries are kept ordered anyway so the entire class could be reduced to just:

```
class OrderedUpdates(RenderUpdates):
    pass
```

And it should behave, functionally, the same but use less memory.

I'm going to guess that this would speed up everything except calls to `sprites()` as it will be doing less work.

Honestly, the data structures in this whole thing are a mess. a basic group should not be storing sprites in a dictionary - it literally says in AbstractGroup that they are in a dictionary in case the group supports layers, then in the LayeredGroup they store the sprites in an extra list as well anyway, so....

In the principle of deleting code is better than adding I would change this to try deleting most of OrderedUpdates and verify it still passes the tests now that python dictionaries are always ordered. 

Then we could do a second pass that tries to replace AbstractGroup's `self.spritedict` with a list data structure and moves the dictionary part down the hierarchy to LayeredUpdates where it actually gets used.

Eventually I'd like to move this whole file to C, but it would be easier to do so if it made more sense.

Starbuck5 (approved):
Sure.

MyreMylar (approved):

PR #3656, opened 2023-01-04
Re-add and re-name context module to `system` module

Starbuck5:
See #3508


MyreMylar:
This name works for me, seem pretty consistent with python having a 'sys' module:

https://docs.python.org/3/library/sys.html

...for system specific information. 

I agree with the take that generic 'misc' modules are a bad idea and tend to accumulate junk - but I think we do need a place from within pygame to get common 'game domain' information that is specific to the system the application is running on, but where we also don't need additional depth. 

e.g. 
 - the path to stick saved games in that is going to make this system happy.
 - the preferred system language (useful to boot a game up in the right language).
 - current battery power (useful to potentially throttle game performance to improve battery life).
 - functions that might be only relevant to linux or only relevant on windows etc - but are none the less useful for games.

these things don't fit easily in any of the more defined submodules, so the only other alternative would be to stick them under the 'pygame' root module which is probably the current 'misc' module for pygame, or make tiny specific submodules for each like 'power' or 'filesystem'.

In conclusion I think a system submodule is OK.

MyreMylar (approved):
LGTM :thumbs

nextlevelgamer:
Can I ask why was this closed?

ankith26:
Deleting branch on request of owner Starbuck5

PR #3657, opened 2023-01-05
Fixed typo in `pygame.image.tobytes` docs

oddbookworm:
`pygame.image.tobytes` said it was an alias for `pygame.image.fromstring`. Changed it to be an alias for `pygame.image.tostring`


Starbuck5 (approved):
Thanks!

MightyJosip (approved):
Good change 👍

PR #3663, opened 2023-01-10
Fixup deprecation error on MacOS 11

Tobias-Fischer:


h-vetinari:
Note that -- in combination with `-Werror` -- this leads to failures on any OSX version newer than 11.0, à la:
```
   src_c/sdlmain_osx.m:211:69: error: 'NSStringPboardType' is deprecated: first deprecated in macOS 10.14 [-Werror,-Wdeprecated-declarations]
       NSString *info = [[NSPasteboard generalPasteboard]stringForType:NSStringPboardType];
                                                                       ^~~~~~~~~~~~~~~~~~
                                                                       NSPasteboardTypeString
```

Starbuck5:
Thanks for submitting a PR!

The apple docs on this make me nervous: https://developer.apple.com/documentation/appkit/nspasteboardtypestring
> Apps that adopt App Sandbox cannot access files identified using the NSPasteboardTypeString pasteboard type. Instead, use an [NSURL](https://developer.apple.com/documentation/foundation/nsurl) object, a bookmark, or a filename pasteboard type.

Apparently any apps on the Mac app store need to have app sandbox enabled? So by making this change we also disable pygame scrap on things that try to distribute via the app store?

Pygame is moving towards deleting our own native scrap implementations in favor of the ones provided by SDL2, so hopefully this won't be a problem for a long time more.

MyreMylar:
This code is still being used on mac in the current implementation of scrap.

Starbuck5:
> This code is still being used on mac in the current implementation of scrap.

I don't think it is.
https://github.com/pygame/pygame/blob/main/src_c/scrap.c#L64-L84

The first `#if` sets it to use the SDL2 scrap if it's not on Windows.

MyreMylar:
> > This code is still being used on mac in the current implementation of scrap.
> 
> I don't think it is. https://github.com/pygame/pygame/blob/main/src_c/scrap.c#L64-L84
> 
> The first `#if` sets it to use the SDL2 scrap if it's not on Windows.


Oh, weird! 

That set of #if #elifs is a big mess, it looks like two of the conditions are unreachable and like to all the x11 and mac specific code so we should definitely delete everything they are pointing to and clean up that #elif.

On the plus side the mac/x11 wrapper for the old style scrap stuff is already built.

MyreMylar:
Related: https://github.com/pygame/pygame/pull/3671

illume (approved):
👍 thanks

Issue #3665, opened 2023-01-10
Segfault in `PixelArray` when trying to access surface after closing the PixelArray


oddbookworm:
If you try to access the `surface` attribute of a `PixelArray` after you closed it, then pygame segfaults. Obviously you shouldn't grab the surface from a closed `PixelArray`, but it shouldn't segfault either.

MRE
```
import pygame
import pygame.pixelarray

screen = pygame.display.set_mode((500, 500))

surf = pygame.Surface((50, 50))
surf.fill("red")

def do_stuff():
    pixels = pygame.pixelarray.PixelArray(surf)
    pixels.close()
    screen.blit(pixels.surface, (0, 0))

running = True
while running:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            running = False
    
    screen.fill("black")
    print("crashes now")
    do_stuff()
    pygame.display.flip()
```

(Windows 11, Python 3.11.1, pygame 2.1.3dev8)
(Windows 11, Python 3.11.1, pygame 2.1.4dev1)

PR #3666, opened 2023-01-10
fixed segfault

oddbookworm:
Fixes https://github.com/pygame/pygame/issues/3665 by raising a `ValueError` if the surface access is after closing the `PixelArray`

### DIFF
--- a/src_c/pixelarray.c
+++ b/src_c/pixelarray.c
@@ -421,6 +421,10 @@ _pxarray_get_dict(pgPixelArrayObject *self, void *closure)
 static pgSurfaceObject *
 _pxarray_get_surface(pgPixelArrayObject *self, void *closure)
 {
+    if (self->surface == NULL) {
+        PyErr_SetString(PyExc_ValueError, "Operation on closed PixelArray.");
+        return NULL;
+    }
     Py_INCREF(self->surface);
     return self->surface;
 }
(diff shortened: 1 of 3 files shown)


Matiiss (changes requested):
I'm fairly certain it would be important to add this check to other places where it hasn't been added:
https://github.com/pygame/pygame/blob/25aac95a17ea925567046440e655f5e577d47506/src_c/pixelarray.c#L432-L438

and I'm not sure about this but if `shape` also gets set to NULL (even if that doesn't happen, I suppose these should then still check if `self->surface == NULL`):
https://github.com/pygame/pygame/blob/25aac95a17ea925567046440e655f5e577d47506/src_c/pixelarray.c#L443-L450
https://github.com/pygame/pygame/blob/25aac95a17ea925567046440e655f5e577d47506/src_c/pixelarray.c#L455-L462
https://github.com/pygame/pygame/blob/25aac95a17ea925567046440e655f5e577d47506/src_c/pixelarray.c#L467-L471

And I'm fairly certain there are a couple more functions that need this check, perhaps, also in `pixelarray_methods.c`

MyreMylar:
From my testing these are the PixelArray public functions and members that segfault after a first call to `.close()`:

- `.surface`
- `.itemsize`
- `.close()` - a second call to it.
- `transpose()`

The others already say:

```
ValueError: Operation on closed PixelArray.
```

Or return some valid data. I didn't bother testing all the dunder methods because this is already a fairly dumb thing for somebody to be doing.

MyreMylar (changes requested):
Please add a similar check to clear the other segfaults outlined above.

MyreMylar (approved):
LGTM 👍

Starbuck5:
This looks reasonable to me, thanks Andrew!

Starbuck5 (approved):

Issue #1633, opened 2020-04-11
Possible memory leak in `_set_targets` of  `scrap_x11.c`


nthykier:
Hi,

The _set_targets in `scrap_x11.c` has a memory leak on error conditions under Python3 using the following path:

 * The malloc of `Atom *targets` returns successfully
 * In one iteration, `chars = PyUnicode_AsASCIIString(PyList_GetItem(list, i));` causes chars to be NULL, which trips an immediate return without a free of `targets` nor passing on `targets` on to `XChangeProperty`.

I got two questions:

 * Should the atoms returned by `_convert_format` be free'ed somehow? As I understand https://tronche.com/gui/x/xlib/window-information/XInternAtom.html, the answer is "probably not" but I do not think it is clear enough.
 * How do we want to handle this error case?  Should we free(targets) and signal an error to the caller (which should then pass it on to its caller) or just complete the call to `XChangeProperty` with however many items we managed to complete (possibly 0).

The code in question for reference.

```C

static void
_set_targets(PyObject *data, Display *display, Window window, Atom property)
{
    int i;
    char *format;
    PyObject *list = PyDict_Keys(data);
#if PY3
    PyObject *chars;
#endif
    int amount = PyList_Size(list);
    /* All types plus the TARGETS and a TIMESTAMP atom. */
    Atom *targets = malloc((amount + 2) * sizeof(Atom));
    if (targets == NULL)
        return;
    memset(targets, 0, (amount + 2) * sizeof(Atom));
    targets[0] = _atom_TARGETS;
    targets[1] = _atom_TIMESTAMP;
    for (i = 0; i < amount; i++) {
#if PY3
        chars = PyUnicode_AsASCIIString(PyList_GetItem(list, i));
        if (!chars) {
            return;
        }
        format = PyBytes_AsString(chars);
#else
        format = PyString_AsString(PyList_GetItem(list, i));
#endif
        targets[i + 2] = _convert_format(format);
#if PY3
        Py_DECREF(chars);
#endif
    }
    XChangeProperty(display, window, property, XA_ATOM, 32, PropModeReplace,
                    (unsigned char *)targets, amount + 2);
}

```

oddbookworm:
Could this issue be closed if #3510 gets merged because all of the old scrap would then be deprecated anyway?

PR #3671, opened 2023-01-15
Remove unused old module scrap code

MyreMylar:
This code was no longer being used for the old scrap module (which is about to be deprecated anyway in another PR).

Closes: #1633 (by removing the affected code)


MyreMylar:
This could use a quick test on mac and linux to double check you can still use scrap to paste on those platforms. 

You should be able to as this code was defined out - but a double check never hurts.

nthykier:
Linked the PR to #1633.

PR #3672, opened 2023-01-16
Draft: Idea for gradually deprecating gfxdraw - share C code with draw

MyreMylar:
This is one idea I have had to start solving the 'too much duplicated functionality' problem in pygame.

In this case I'm looking at trying to eventually remove the 'gfxdraw' module because it is too similar to the 'draw' module. I have similar issues with the 'freetype' module duplicating functionality in font.

Ideally we can take any good things from these similar modules and bring them into one final superior module that does it all.

---

In this draft I'm investigating clearing out code that does the exact same thing in two different places e.g. we can draw an outlined rectangle with `draw.rect` or with `gfxdraw.rectangle()`. 

At the minute this PR literally does the minimal work to replace gfxdraw's simple rectangle drawing with the main function in draw.

**Potential impacts:**

 - Code performance.
 - Code size (some of the same code will be compiled into two different modules - but we also will be able to get rid of some code so it is probably a wash.


MyreMylar:
I'll revisit these ideas at a later point I think. It was an interesting experiment though!

PR #3676, opened 2023-01-17
Draft: Move sprite dictionary storage further down inheritance tree

MyreMylar:
While looking at another PR I noted that all sprite Group's base data structure was a dictionary, which doesn't make much sense given that most of the time you want to be speedily iterating over their contents for drawing/updating.

Also, several of the Group types lower down the inheritance family tree were storing sprites in lists as well as the dictionary. I decided to try moving stuff around so that a list is the base data structure of an `AbstractGroup`, mostly to see if it would work.

So far it seems to pass the tests.

Benefits
---------

- Performance seems to improve by about 10-15% when draw speed is constrained by looping speed (i.e. lots of small blits, I was testing with 10,000 32x32 and 16x16 pixel sized sprites).
- Potential for another 10-15% speed up if a group draw uses the 'fblits' method from #3297
- Once we switch to fblits, you can also pass in a single blend flag to draw - and thus take advantage of the faster blend modes like `BLEND_PREMULTIPLIED`.
- I'm actually not sure if there is much benefit in pulling this any more into C code, perhaps we could have a C for loop instead of the draw loop, but most other stuff in sprite is too tightly tied to user code.

Remaining To Dos
---------------------

- [x] Can we restore a `.clear()` to the current base group based on the drawlist rectangles? does it make any difference using rectangles we are going to draw with versus rectangles returned from blits? Does it matter much (I'm thinking blits might clip rectangle to surface edges/clips but since they are to be used for future blitting anyway in `.clear()` they will get clipped there all the same.
- [x] Ideally I'd like Group to just _be_ the `FastGroup` not adding a dumbly name class, which was handy for quick testing.
- [ ] See if we can extend any of these drawlist changes further down the hierarchy of Group types to speed up layered Groups too.
- [ ] Handle any formerly public, but not documented, attributes so they work - but are deprecated.
- [ ] remove dictionary from sprite, replace with set.
- [ ] Final task is to split this into a few smaller, more digestible, non-draft PRs with this as the final target state.


src_py/sprite.py
```diff
@@ -111,9 +113,31 @@ class Sprite:
 
     def __init__(self, *groups):
         self.__g = {}  # The groups the sprite is in
```
nthykier, line 115: While we are at it, isn't `__g` basically just a poor-man's `set` ?  At least I see no usage of `__g` in the base class that warrants it being a `dict`.

(I know you did not introduce that change, but I was confused because I thought the Group would be in the value rather than the key section of the `dict`)

src_py/sprite.py
```diff
@@ -416,7 +450,7 @@ def copy(self):
 
         """
         return self.__class__(  # noqa pylint: disable=too-many-function-args
-            self.sprites()  # Needed because copy() won't work on AbstractGroup
+            self._spritelist  # Needed because copy() won't work on AbstractGroup
```
nthykier, line 453: Copy says we need a copy, but the change seems to contradict that (`self.spites()` would trigger a copy AFAICT)

nthykier: ... actually, I am not sure what the comment is trying to say, but it does say `copy()` which smells like it wanted a copy.  If the comment is outdated (or could be worded better and you have an idea what it means - I do not), then I think it should be updated as well.

src_py/sprite.py
MyreMylar: I'm not _as_ concerned about this one because generally the `self.__g` dict is not looped over regularly, but yes only the keys of the dictionary here appear to be used which does add confusion.

src_py/sprite.py
MyreMylar: Yeah I'm not sure what that comment it is about. 

AFAICT things seem to work fine passing the raw list in here, you get a new Group with a new list containing the same sprites. 

I'll tinker around with copy a bit more I think, but either way it isn't important to change this anyway so I can revert it to calling `sprites()`.

itzpr3d4t0r:
Would be cool to have some benchmarks

MyreMylar:
> Would be cool to have some benchmarks

This is my current super basic benchmark:

```
import pygame
import random


pygame.init()
clock = pygame.time.Clock()

fps_font = pygame.font.Font(None, 32)

fast_group_test = pygame.sprite.Group()

test_surf = pygame.Surface((16, 16), pygame.SRCALPHA)
test_surf.fill(pygame.Color(200, 150, 50, 80))

for _ in range(0, 30000):
    new_sprite = pygame.sprite.Sprite(fast_group_test)
    new_sprite.image = test_surf
    new_sprite.rect = pygame.Rect((random.randint(0, 784), random.randint(0, 584)), (16, 16))


fps_average = []

while True:
    screen.fill("black")

        if event.type == pygame.QUIT:
            pygame.quit()

    fast_group_test.draw(screen)

    frame_time = max(clock.tick() / 1000.0, 0.0000000000000001)

    if len(fps_average) >= 1000:
        fps_average.pop(0)

    fps_average.append(frame_time)
    fps_text = fps_font.render(f"{(1.0 / (sum(fps_average)/len(fps_average))):.2f}", True, pygame.Color('#FFFFFF'))

    screen.blit(fps_text, pygame.Rect((700, 40), fps_text.get_size()))

    pygame.display.flip()
```

MyreMylar:
I noticed that flblits also lets us pass in a blend flag, so if you combo this, the fblits PR and the latest BLEND_PREMULTIPLIED improvements you can get some doubling of performance in some, fairly contrived, 10,000 static sprite drawing scenarios.

I can get 10,000 moving, alpha-blended blitted rectangle sprites running at about 38 FPS:

image

versus 24 FPS:

image

with current main sprite module and normal alpha blending. Most of the time is lost in the updating to move the sprites.

MyreMylar:
I think this was a relatively successful test of the concept of refactoring sprite a bit. This PR is however too large and unwieldy at the minute. For now I am closing it down, and may revisit it's ideas in smaller chunks in the future.

PR #3680, opened 2023-01-22
fixed lots of potential segfaults

### LINKED ISSUE #3679: Segfault when trying to draw to screen after calling pygame.quit
**Environment:**

- **PyGame version** (pygame 2.1.3.dev8):
- **Python version** (Python 3.10.0)
- **SDL version** (SDL 2.0.22)

**Current behavior:**

If there's a pygame.draw.rect() line in the main game loop and pygame.quit() got called in e.g. "if event.type == pygame.QUIT" without the main loop stopping/breaking, there will be a pretty long segfault message printed to the user. (see screenshot below)

**Screenshots**

image

**Getting the segfault**

```
import pygame

pygame.quit()

pygame.draw.rect(screen, "blue", pygame.Rect(0, 0, 1, 1))

### PR DESCRIPTION
**Pull request by oddbookworm**
fixes #3679 

I'm pretty sure that I got most of them, but I probably missed a few

### CODE REVIEW

**Comment by Starbuck5** [comment]
So I’ve been thinking about this as Matt has brought it up. Pygame.display.quit turns the screen SDL_surface into NULL— but should it? Is there any (current) reason the screen surface can’t continue in software after the display behind it has quit? Why try and error on everything a surface could do, what if we could instead error on pygame.event.get or pygame.display.flip.

**Comment by MightyJosip** [comment]
You can probably include stuff from #3525 in your PR. There is also this comment that may be relevant https://github.com/pygame/pygame/pull/3525#issuecomment-1304607993

**Comment by oddbookworm** [comment]
> So I’ve been thinking about this as Matt has brought it up. Pygame.display.quit turns the screen SDL_surface into NULL— but should it? Is there any (current) reason the screen surface can’t continue in software after the display behind it has quit? Why try and error on everything a surface could do, what if we could instead error on pygame.event.get or pygame.display.flip.

This is fair, but I still think that my PR is still relevant. If someone somehow still would end up with a weird, broken surface it would still potentially segfault

**Comment by oddbookworm** [comment]
> You can probably include stuff from #3525 in your PR. There is also this comment that may be relevant [#3525 (comment)](https://github.com/pygame/pygame/pull/3525#issuecomment-1304607993)

Are you suggesting that the two PR's should be effectively merged into one? Since the same code is repeated in different places in both, I think that would be a good idea. I also like the idea of a macro. Maybe `SURF_INIT_CHECK` instead of `VIDEO_INIT_CHECK` like @Starbuck5 suggested in your PR?

Maybe something like this?
```
#define SURF_INIT_CHECK(SDL_SurfObj) if (!SDL_SurfObj) {return RAISE(pgExc_SDL_Error, "Surface is not initialized");}
```

PR #3685, opened 2023-01-27
Add "from_existing_window" function to the _sdl2.Window

yunline:
```python
from_existing_window(hwnd: int) -> Window
```
With this function, pygame window could be inserted into another window (Tk/Qt/MFC/...) .
Here is an example based on Tkinter：

```python
import tkinter
import threading

root=tkinter.Tk()
root.geometry("640x480")
btn=tkinter.Button(root,text="HelloWorld\n你好世界")
btn.pack()
cv=tkinter.Canvas(root)
cv.pack()

off=0
def pg_loop(hwnd):
    import pygame
    from pygame._sdl2 import video

    pygame.display.init()

    win=video.Window.from_existing_window(hwnd)
    
    win.focus()         # focus/restore is available
    #win.grab=1         # grab is available
    #win.opacity=0.5    # opacity is available
    #win.hide()         # hide/show is availabe
    #win.resizable=1    # available but works weirdly

    ren=video.Renderer(win)

    pos=(20,20)
    while not off:
        ren.draw_color=(0,0,0,0)
        ren.clear()
        ren.draw_color=(255,0,0,0)
        ren.fill_rect((*pos,10,10))
        ren.present()

            if event.type==1024:#MOUSEMOTION
                pos=event.pos
            else:
                print(event)


if __name__ == "__main__":
    #hwnd=int(root.frame(),16)  # get the handle of the root window
    hwnd=cv.winfo_id()          # get the handle of the canvas widget

    p=threading.Thread(target=pg_loop,args=(hwnd,))

    p.start()
    root.mainloop()
    off=1
```
image


MyreMylar:
Related Stale PR: https://github.com/pygame/pygame/pull/2981

To reference when reviewing

yunline:
Since the environment variable `SDL_WINDOWID ` was no longer used in SDL2, I added a new `hwnd` argument to the `set_mode()` function to embed the pygame window to another window. I also wrote an example here
```python
import tkinter
import threading

root=tkinter.Tk()
root.geometry("640x480")
cv=tkinter.Canvas(root,width=400,height=400)
cv.pack()

off=0
def pg_loop(hwnd):
    import pygame
    pygame.display.init()

    sf=pygame.display.set_mode(hwnd=hwnd)

    while not off:
        sf.fill((0,0,0))
        pygame.draw.rect(sf,(255,0,0),(10,10,10,10))
        pygame.display.update()

threading.Thread(target=pg_loop,args=(cv.winfo_id(),)).start()
root.mainloop()
off=1
```
It seems this feature is not compatible with OpenGL. When I try to embed the glcube into a tkinter window, it raises an exception like this: `pygame.error: The specified window isn't an OpenGL window`. I don't know if this is a bug.

yunline:
> ```python
> import pygame
> from pygame._sdl2 import video
> import tkinter
> 
> import multiprocessing as mp
> 
> root=tkinter.Tk()
> root.geometry("640x480")
> btn=tkinter.Button(root,text="HelloWorld\n你好世界")
> btn.pack()
> cv=tkinter.Canvas(root)
> cv.pack()
> 
> def pg_loop(hwnd):
>     pygame.display.init()
> 
>     win=video.Window.from_existing_window(hwnd)
> 
>     #win.opacity=0.5 # opacity is available
>     #win.hide() # hide/show is availabe
> 
>     ren=video.Renderer(win)
>     ren.draw_color=(0,0,0,0)
>     ren.clear()
>     ren.draw_color=(255,0,0,0)
>     while 1:
>         ren.fill_rect((10,10,100,100))
>         ren.present()
> 
> if __name__ == "__main__":
>     #p=mp.Process(target=pg_loop,args=(int(root.frame(),16),))
>     p=mp.Process(target=pg_loop,args=(cv.winfo_id(),))
> 
>     p.start()
>     root.mainloop()
>     p.terminate()
> ```
When bundled by Pyinstaller, this script doesn't work properly.
However, When using the display module with `set_mode(hwnd=hwnd)`, the exe runs well.
Maybe there is a bug somewhere (?

MyreMylar:
> This script runs well. However, it seems this feature is not compatible with OpenGL. When I try to embed the glcube into a tkinter window, it raises an exception like this: `pygame.error: The specified window isn't an OpenGL window`. I’m not sure if it is a bug.

It is not a bug. You need an OpenGL enabled window to do any openGL rendering. You can create an OpenGL tkinter window with this library I believe:

https://pypi.org/project/pyopengltk/

There is also the `SDL_HINT_VIDEO_WINDOW_SHARE_PIXEL_FORMAT` to consider.

yunline:
[This](https://github.com/juyanith/Win32Host/blob/422833d3b77e58135ae9d493964902690774844f/samples/Sdl2Example/Sdl2Window.cs) might be a reference.
image

MyreMylar:
I think that the new `hwnd` parameter could use a little bit of documentation in the display module docs for `set_mode()`.
This file:
https://github.com/pygame/pygame/blob/main/docs/reST/ref/display.rst

It is possible we might want to re-create the old `SDL_WINDOWID` variable for backwards compatability (we could potentially add it even though it doesn't exist in SDL anymore, it would just be a little confusing) - but I'm not 100% convinced by that logic. We have now gone a pretty long time with SDL2 without recreating it, and honestly, it was a kind of clunky way to do it in Pygame 1. I personally prefer your new API with a parameter for `set_mode()` to achieve the same thing in a cleaner way.

yunline:
I think there should be an API for opening an OpenGL window in _sdl2.video module. But that might be another pull request.

yunline:
I merged some code from #3695 to pass the check.

chimosky:
Tested this in sugar and it works as expected, the test activity starts and runs like it should.

Mega-JC:
It's finally great to have this kind of functionality in pygame again.

One small remark: `OpenGL_Window` is a pretty non-standard way of naming classes in Python (mixing PascalCase with snake_case), especially if exposed to the end user. `OpenGLWindow` is just as clear and doesn't have that issue.

docs/reST/ref/display.rst
```diff
@@ -172,6 +152,32 @@ required).
...
+   On high resolution displays(4k, 1080p) and tiny graphics games (640x480)
+   show up very small so that they are unplayable. SCALED scales up the window
+   for you. The game thinks it's a 640x480 window, but really it can be bigger.
+   Mouse events are scaled for you, so your game doesn't need to do it. Note
+   that SCALED is considered an experimental API and may change in future
+   releases.
+
+   The ``depth`` argument represents the number of bits to use for color. It is 
+   usually best to not pass the depth argument. It will default to the
+   best and fastest color depth for the system. If your game requires a
```
oddbookworm, line 169: I think documentation overhauls like this are better suited to a separate PR. IMO, each PR should be fairly focused on what it's purpose is and not cover too many changes outside of that particular scope.

docs/reST/ref/display.rst
```diff
@@ -189,26 +195,12 @@ required).
 
    .. versionadded:: 2.0.0 ``vsync``
 
-   Basic example:
+   The ``hwnd`` argument is an integer contains the window id (or handle) of
+   a foreign window. By setting this parameter, pygame display can be embedded
```
oddbookworm, line 199: of course, keep the part you're adding documentation for new arguments and new API

examples/aliens.py
```diff
@@ -304,7 +304,6 @@ def main(winstyle=0):
 
     # Run our main loop whilst the player is alive.
     while player.alive():
-
```
oddbookworm, line 307: There's enough of these random empty line removals/additions that I think deserves another PR, but do not belong in this one because they are completely unrelated to `_sdl2.Window`

src_py/midi.py
```diff
@@ -68,8 +68,7 @@ def _module_init(state=None):
         _module_init.value
     except AttributeError:
         return False
-    else:
```
oddbookworm, line 71: Again, I think this should be in a separate PR. It's completely unrelated to `_sdl2.Window`

src_py/threads/__init__.py
```diff
@@ -250,15 +249,15 @@ def tmap(f, seq_args, num_workers=20, worker_queue=None, wait=True, stop_on_erro
         # print("after wait")
         # print("queue size:%s" % wq.queue.qsize())
         if wq.queue.qsize():
-            raise Exception("buggy threadmap")
+            raise RuntimeError("buggy threadmap")
```
oddbookworm, line 252: Another PR IMO

src_py/threads/__init__.py
```diff
@@ -250,15 +249,15 @@ def tmap(f, seq_args, num_workers=20, worker_queue=None, wait=True, stop_on_erro
...
         # if we created a worker queue, we need to stop it.
         if not worker_queue and not _wq:
             # print("stopping")
             wq.stop()
             if wq.queue.qsize():
                 um = wq.queue.get()
                 if not um is STOP:
-                    raise Exception("buggy threadmap")
+                    raise RuntimeError("buggy threadmap")
```
oddbookworm, line 260: Another PR IMO

test/pixelarray_test.py
```diff
@@ -698,35 +698,25 @@ def test_subscript(self):
             # Test simple slicing
             self.assertEqual(len(ar[:, :]), 6)
             self.assertEqual(
-                len(
-                    ar[
-                        :,
-                    ]
-                ),
+                len(ar[:,]),
```
oddbookworm, line 701: Another PR

test/pixelarray_test.py
```diff
@@ -698,35 +698,25 @@ def test_subscript(self):
...
                 6,
             )
             self.assertEqual(len(ar[1, :]), 8)
             self.assertEqual(len(ar[:, 2]), 6)
             # Empty slices
             self.assertEqual(
-                ar[
-                    4:4,
-                ],
```
oddbookworm, line 714: Another PR

test/pixelarray_test.py
```diff
@@ -698,35 +698,25 @@ def test_subscript(self):
...
+                ar[4:4,],
                 None,
             )
             self.assertEqual(ar[4:4, ...], None)
             self.assertEqual(ar[4:4, 2:2], None)
             self.assertEqual(ar[4:4, 1:4], None)
             self.assertEqual(
-                ar[
-                    4:4:2,
```
oddbookworm, line 722: Another PR

test/pixelarray_test.py
```diff
@@ -698,35 +698,25 @@ def test_subscript(self):
...
-                ],
+                ar[4:4:2,],
                 None,
             )
             self.assertEqual(
-                ar[
-                    4:4:-2,
```
oddbookworm, line 728: Another PR

test/pixelarray_test.py
```diff
@@ -857,9 +847,7 @@ def test_ass_subscript(self):
             self.assertEqual(ar[0, 0], 0)
             self.assertEqual(ar[1, 0], 0)
             self.assertEqual(ar[-1, -1], 0)
-            ar[
-                ...,
```
oddbookworm, line 861: Another PR

docs/reST/ref/display.rst
```diff
@@ -78,12 +78,6 @@ required).
       Windows : windib, directx
       Unix    : x11, dga, fbcon, directfb, ggi, vgl, svgalib, aalib
 
-   On some platforms it is possible to embed the pygame display into an already
-   existing window. To do this, the environment variable ``SDL_WINDOWID`` must
-   be set to a string containing the window id or handle. The environment
```
oddbookworm, line 83: I think rather than deleting this, it should be updated to say that this is only for pygame 1.x and no longer applies to pygame 2.x due to the change from SDL 1 to SDL 2.

oddbookworm (changes requested):

docs/reST/ref/display.rst
yunline: Thank you for your review. I will fix that.

Rybec:
Question about this feature:  Is there a way to use it to connect to an external window?  XScreenSaver runs screensavers by creating a window itself and then passing a window ID to the screensaver program, which it then is expected to draw to.  If this feature would allow Pygame to draw to the window created by XScreenSaver, that would allow screensavers for XScreenSaver to be written in Python, which would be totally awesome!

PR #3688, opened 2023-01-29
Implement get_pointsize() and set_pointsize() via TTF_SetFontSize

novialriptide:
Depends on: https://github.com/pygame/pygame/pull/3695 for test purposes

## Working example
lmao

```python
import sys
import pygame

pygame.init()
font = pygame.font.Font(None, 25)
print(font.get_pointsize())

while True:
    
    for e in pygame.event.get():
        if e.type == pygame.QUIT:
            sys.exit()
        if e.type == pygame.KEYUP:
            if e.key == pygame.K_UP:
                font.set_pointsize(font.get_pointsize() + 1)
                print(font.get_pointsize())
            if e.key == pygame.K_DOWN:
                font.set_pointsize(font.get_pointsize() - 1)
                print(font.get_pointsize())

    screen.fill((0, 0, 0))
    a = font.render("lmao", True, (255, 255, 255))
    screen.blit(a, (250, 250))
    
    pygame.display.flip()
```


docs/reST/ref/font.rst
```diff
@@ -472,6 +472,20 @@ solves no longer exists, it will likely be removed in the future.
 
       .. ## Font.get_descent ##
 
+   .. method:: set_pointsize
+
+      | :sl:`set the point size of the font`
+      | :sg:`set_pointsize(bool) -> None`
+
```
zoldalma999, line 479: Not sure if intentional, but there is no description/body here

MightyJosip:
Why point_size if SDL function is font_size

novialriptide:
> Why point_size if SDL function is font_size

In the documentation for SDL_ttf (https://wiki.libsdl.org/SDL_ttf/TTF_SetFontSize), the parameters suggest that we're actually changing the point size, plus `pygame.font.Font.size` already exists.

docs/reST/ref/font.rst
novialriptide: I felt like the small description was enough, but I can add more to it if needed.

Issue #3659, opened 2023-01-07
pygame.transform.blur


ScriptLineStudios:
```python
pygame.transform.blur(surface: pygame.Surface, kernel_size: int) -> pygame.Surface
```

pretty self explanatory allows you to blur an image, passing in an (optional?) kernel size for the blur

theRealProHacker:
I think the reason this is/should/will (choose for yourself) not be implemented is that there are libraries that already do these kind of things, which you can use through the `surfarray` api. Implementing this in pygame would just reinvent the wheel and means more code to maintain, that might not even give a performance boost over using `numpy` arrays. 

And if you look into the [pygame docs](https://www.pygame.org/wiki/about), you see that pygames simplicity is a feature

> **Small amount of code**. It does not have hundreds of thousands of lines of code for things you won't use anyway. The core is kept simple, and extra things like GUI libraries, and effects are developed separately outside of pygame.

illume:
Yeah. There's examples of implementing a blur with surfarray or PixelArray. It would be extra code to download size and to maintain this as a separate bunch of C code. Additionally there's PIL and cv2 for anyone who wants some advanced options for this. Then there's shaders for anyone who is using a gpu based renderer.

Closing this to keep pygame lean.

PR #3698, opened 2023-02-04
pygame.transform.blur

yunline:
# Add `blur` to transform module.
```python
def box_blur(
    surface: Surface,
    radius: int,
    repeat_edge_pixels: Optional[bool] = True,
    dest_surface: Optional[Surface] = None
) -> Surface: ...

def gaussian_blur( ... ) -> Surface: ...
def bokeh_blur( ... ) -> Surface: ...
```
Inspired by #3659.

image
☑ Box Blur
☑ Gaussian Blur
☐ Bokeh Blur

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -2967,7 +2967,7 @@ surf_average_color(PyObject *self, PyObject *args, PyObject *kwargs)
 }
 
 static void
-box_blur(SDL_Surface *src, SDL_Surface *dst, int radius)
+box_blur(SDL_Surface *src, SDL_Surface *dst, int radius, SDL_bool repeat)
 {
     // Reference : https://blog.csdn.net/blogshinelee/article/details/80997324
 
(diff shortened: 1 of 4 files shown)


novialriptide:
Looks like you need to run `python setup.py format` to pass the linting test.

src_c/transform.c
```diff
@@ -2966,6 +2966,152 @@ surf_average_color(PyObject *self, PyObject *args, PyObject *kwargs)
     return Py_BuildValue("(bbbb)", r, g, b, a);
 }
 
+static void
+box_blur(SDL_Surface *src, SDL_Surface *dst, int radius)
```
yunline, line 2970: It takes about 100ms to process a image using this algorithm.

Image: 2048x2048 RGBA
CPU: AMD Ryzen7 6800H

src_c/transform.c
```diff
@@ -2966,6 +2966,152 @@ surf_average_color(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    new_surf = blur(src_surf_obj, dst_surf_obj, radius);
+    if (!new_surf) {
+        return NULL;
+    }
+
+    if (dst_surf_obj) {
+        Py_INCREF(dst_surf_obj);
+        return (PyObject *)dst_surf_obj;
+    }
+    else {
```
novialriptide, line 3110: Code quality nitpick: Could you remove the else statement and put the return new surface line at the end of the function?

novialriptide (changes requested):

src_c/transform.c
```diff
@@ -2966,6 +2966,152 @@ surf_average_color(PyObject *self, PyObject *args, PyObject *kwargs)
...
+}
+
+static SDL_Surface *
+blur(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int radius)
+{
+    SDL_Surface *src = NULL;
+    SDL_Surface *retsurf = NULL;
+
+    if (radius < 0) {
+        return RAISE(PyExc_ValueError, "Radius less than zero.");
```
novialriptide, line 3045: phrasing could be better like "the radius should not be less than zero"

novialriptide (changes requested):

novialriptide:
Blurring Better Call Saul is giving me weird results here... I'm not sure if this is normal

### Result
image

### Original Image
maxresdefault

novialriptide:
You can never go wrong with Saul's good looks.

yunline:
> It takes about 220ms to process an image using this algorithm.

I optimized the algorithm and now it takes 75ms to process one image.

Image: 2048x2048 RGBA
CPU: AMD Ryzen7 6800H

itzpr3d4t0r:
I think that a blur function should be named depending on the blur algorithm it uses, so gaussian, bokeh etc.

yunline:
New parameter `repeat_edge_pixels`
image

yunline:
Gaussian blur √
image

PR #3703, opened 2023-02-08
Fix the fault of stubtest

yunline:
# Fixed the fault of stubtest.
[https://github.com/pygame/pygame/actions/runs/4120284878/jobs/7114857489](https://github.com/pygame/pygame/actions/runs/4120284878/jobs/7114857489)

a


src_py/sysfont.py
```diff
@@ -474,7 +474,7 @@ def get_fonts():
     return list(Sysfonts)
 
 
-def match_font(name, bold=0, italic=0):
+def match_font(name, bold=False, italic=False):
     """pygame.font.match_font(name, bold=0, italic=0) -> name
```
illume, line 478: Probably the doc string needs to be updated to be `bold=False, italic=False`.

illume (approved):
👍

src_py/sysfont.py
illume: I can fix this later.

PR #3708, opened 2023-02-12
Transform.blur

yunline:


illume:
Hello.

The appveyor windows builds are failing because the warnings are being treated as errors. See the warnings here: https://ci.appveyor.com/project/pygame/pygame/builds/46209091/job/fpht4tu2qgk8uxf3#L707

PR #3722, opened 2023-02-19
Add `special_flags` argument to `Group.draw`

### LINKED ISSUE #3719: Allow passing `special_flags` to `Group.draw`
**Description**
Add a `special_flags` argument to `Group.draw` (and all the other group classes) and this can be passed to `Surface.blit`.
This would be useful as with this, I could write code like this:
```python
group.draw(surface, special_flags=pygame.BLEND_RGBA_MAX)
```
instead of
```python
for s in group.sprites():
    surface.blit(s.image, s.rect, special_flags=pygame.BLEND_RGBA_MAX)
```
Also, for more complicated groups (eg. `RenderUpdates`), you have to copy the entire code of the `.draw` functions and change just the line where the `.blit` function is called which is quite annoying.

MarcellPerger1:
I've added the `special_flags` argument to all the `Group`s except `DirtyLayered` to fix #3719.

I'm not sure whether to add the `special_flags` argument to `DirtyLayered` as `DirtyLayered` already partially handles this with the `blendmode` attribute on the `DirtySprite` objects.


illume:
Thanks :)

In my humble opinion, it would make sense for the DirtyLayered to support this, since it is a subclass. Probably all the subclasses that come with pygame should support it?

illume:
Looks like the file needs formatting with the black formatter.

illume:
There’s a type issue here:
https://github.com/pygame/pygame/actions/runs/4218255639/jobs/7323102489

pygame/sprite.pyi:208: error: Argument 2 of "draw" is incompatible with supertype "AbstractGroup"; supertype defines the argument type as "int"  [override]
[20](https://github.com/pygame/pygame/actions/runs/4218255639/jobs/7323102489#step:6:21)
pygame/sprite.pyi:208: note: This violates the Liskov substitution principle
[21](https://github.com/pygame/pygame/actions/runs/4218255639/jobs/7323102489#step:6:22)
pygame/sprite.pyi:208: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides

MarcellPerger1:
The problem is that `AbstractGroup` (and most other groups) have the draw method being
```python
def draw(self, surface: Surface, special_flags: int = 0) -> List[Rect]: ...
```
but `LayeredDirty` has the draw method being
```python
def draw(self, surface: Surface, bgd: Optional[Surface] = None, special_flags: int = 0) -> List[Rect]: ...
```
so code that takes any instance of `AbstractGroup` and passes the `special_flags` as the 2nd argument would actually be passing it as the `bgd` argument if that instance turned out to be a `LayeredDirty`.

I can thing of 2 solutions:
1. Make the `special_flags` argument keyword-only so that it can't accidentally be passed as the `bgd` argument to `LayeredDirty`
2. Add an unused argument to the other draw methods so that `special_flags` is always the 3rd argument.

gresm:
@MarcellPerger1 in my opinion the option 1 is better.

illume:
> 2. Add an unused argument to the other draw methods so that special_flags is always the 3rd argument.

I'm a fan of your second solution (except rename bgd to `background` or `bgsurf`). Because then there is a consistency with each of the classes draw methods.

Do you see any drawbacks with this solution 2?

(I couldn't find any code using a bgd keyword argument, so I think it's safe to rename)



Q: Is a blendmode(special_flags) attribute on sprite would be better here?  IMHO, No. I think there's a use case where we want the special_flags to be used for the whole group. It's more convenient when you want everything to use the same

MarcellPerger1:
I agree that special_flags should be used for the whole group. Because then it will be consistent with all the other groups as the special_flags argument will always apply to everything in the group (and that won't depend on if its a normal `Group` or a `LayeredDirty`)

illume:
Looks like there's a couple of pylint issues:

```
src_py/sprite.py:539:28: W0613: Unused argument 'bgsurf' (unused-argument)
src_py/sprite.py:1169:16: W0612: Unused variable 'flags' (unused-variable)
```

illume (approved):
👍 thanks

PR #3727, opened 2023-02-20
Fix of inverted boolean of borderless window property

### LINKED ISSUE #3707: `pygame._sdl2.video.Window` reversed boolean attribute `borderless`
**Environment:**

- **Operating system**: Windows
- **Python version**: 3.8.0
- **SDL version**: 2.0.18
- **PyGame version**: 2.1.2

**Current behavior:**

When setting the attribute `borderless` of an instance of `pygame._sdl2.video.Window` to `True` it does not make it *borderless* and when `False` it makes it *borderless*.

**Expected behavior:**

When setting the attribute `borderless` of an instance of `pygame._sdl2.video.Window` to `True` it makes it *borderless* and when `False` it does not make it *borderless*.

**Test code**

```
from pygame import _sdl2 as sdl

window = sdl.video.Window()
window.borderless = False # Which makes the window borderless somehow
window.borderless = True # Which makes the window not borderless
# THEY ARE THE EXACT OPPOSITE!
```

joaquin30:
Fixes #3707 by changing a line in `video.pyx` and recompiling into `video.c`.


illume (approved):
Thanks 👍

PR #3732, opened 2023-02-24
Remove `:return` from `sysfont._parse_font_entry_win`

MarcellPerger1:
Remove `:return` from `sysfont._parse_font_entry_win` as it doesn't actually return anything.


illume (approved):
👍 thanks

Issue #3734, opened 2023-02-24
Add `Vector2.create_from_polar()`


MarcellPerger1:
**Description**

Add a `Vector2.create_from_polar()` classmethod as currently there is only `from_polar` which sets the x and y of an existing vector, making it a bit harder to pass a vector created from polar coordinates into a function.
For example you currently have to do something like this:
```python
v = pygame.Vector2()
v.from_polar((length, angle))
some_function(v)
```
instead of being able to write
```python
some_function(pygame.Vector2.create_from_polar(length, angle))
```
This is especially annoying is you only use the created vector for passing to the function and don't use it anywhere else. I imagine this could also get quite annoying in list comprehensions.

KyleWynne:
Hey, i'd love to help with this issue. I'm new to this project in which file are these functions defined?

MarcellPerger1:
They are defined in the `src_c/math.c` file.

illume:
As a class method like you suggested is indeed nice.

How about using from_polar to return a vector?

KyleWynne:
That's also a great idea i'm going to start playing around with both and see what I can come up with.

joaquin30:
Sorry, I should have mentioned that I was working on this, but I started when there were no comments. The draft I made is a possible implementation of `create_from_polar`, but I think what @ilume says makes a lot more sense. Besides that we should have a similar method for `from_spherical`.

illume:
> Sorry, I should have mentioned that I was working on this, but I started when there were no comments. The draft I made is a possible implementation of create_from_polar, but I think what @illume says makes a lot more sense. Besides that we should have a similar method for from_spherical.

Cool. If @MarcellPerger1 is also agreeing about "a from_polar class method returning the new vector" I think we should do that.

MarcellPerger1:
Yes, I agree that a from_polar classmethod returning the vector is a good idea.

joaquin30:
OK, i need help :face_exhaling:. I don't now how to have a class method and a instance method with the same name. I was thinking on doing something like:
~~~py
class vec:
  def __init__(self, x, y):
     self.x = x
     self.y = y
     self.from_polar = self._from_polar_instance
  @classmethod
  def from_polar(cls, r, phi):
    return cls(r*cos(phi), r*sin(phi))
  def _from_polar_instance(self, r, phi):
    self.x = r*cos(phi)
    self.y = r*sin(phi)
~~~
[...]

PR #3744, opened 2023-03-05
Implementation of Vector2.from_polar and ClassObjectMethod descriptor

joaquin30:
Another possible implementation of #3734.


illume:
The lint error is because of clang-format. If you install that, and then format it should be good. Then you can run before checking:

`python3 setup.py lint`

illume:
The windows compilation error can be seen here: https://ci.appveyor.com/project/pygame/pygame/builds/46412174/job/bhnaqycrhpa09l20#L763

> src_c/math.c(4345): error C2099: initializer is not a constant

joaquin30:
For math.pyi I checked some code with mypy and it seems to output the correct errors, except when using `Vector2.from_polar(Vector2(), (1,1))` (shows no errors), but it gives a syntax error using python. The problem is the test "Ubuntu sdist / build (ubuntu-22.04) (pull_request)", but if I use `@classmethod` or `@staticmethod` it will fail further tests, and mypy doesn't work correctly. Maybe we should just leave it as it is?

illume:
Nice progress.

There seems to be an issue with the mypy stubtest.
https://github.com/pygame/pygame/actions/runs/4355791723/jobs/7612896124#step:6:19

```
  cd buildconfig/stubs
  python3 -m pip install mypy==0.982
  python3 -m mypy.stubtest pygame --allowlist mypy_allow_list.txt
```

```
pygame/math.pyi:223: error: Self argument missing for a non-static method (or an invalid type for self)
pygame/math.pyi:233: error: A function returning TypeVar should receive at least one argument containing the same Typevar
pygame/math.pyi:233: note: Consider using the upper bound "_GenericVector" instead
pygame/math.pyi:325: error: Self argument missing for a non-static method (or an invalid type for self)
pygame/math.pyi:325: error: A function returning TypeVar should receive at least one argument containing the same Typevar
pygame/math.pyi:325: note: Consider using the upper bound "_GenericVector" instead
```

illume:
> For math.pyi I checked some code with mypy and it seems to output the correct errors, except when using Vector2.from_polar(Vector2(), (1,1)) (shows no errors), but it gives a syntax error using python. The problem is the test "Ubuntu sdist / build (ubuntu-22.04) (pull_request)", but if I use @classmethod or @staticmethod it will fail further tests, and mypy doesn't work correctly. Maybe we should just leave it as it is?

Did you try using cls as the first argument for the class method? https://docs.python.org/3/library/functions.html?highlight=classmethod#classmethod

joaquin30:
OK, I think i have a workaround:

~~~pyi
from typing_extensions import Protocol

# Not defined in code, only for type checking from_polar ClassObjectMethod
class _from_polar_protocol(Protocol):
    def __call__(self, value: Tuple[float, float]) -> Optional[_TVec]: ...

class Vector2(_GenericVector):
    from_polar: _from_polar_protocol
~~~

I tried the test and type-checking some code and seems that it works. Also, i can't use `@classmethod` or `@staticmethod` because it shows an error that the method is not a classmethod at runtime.

illume (approved):
Thanks 🎈 🎉

PR #3747, opened 2023-03-06
docs: fix move images tutorial final example

SalimBAhmed:
fixing the tutorial for moving images, where in the last example the player image is not blitted in the screen.


illume (approved):
👍 thanks a lot

PR #3762, opened 2023-03-14
Fix docs for `cursors.Cursor.copy()`

MarcellPerger1:
Previously, the docstring for `cursors.Cursor.copy()` was empty and it wasn't formatted correctly on the website (https://www.pygame.org/docs/ref/cursors.html#pygame.cursors.Cursor.copy):
image


illume (approved):
🎈 🎉 thanks

PR #3764, opened 2023-03-17
Be more careful when installing pygame parachutes for fatal signals

timfel:
Using signal(2) is discouraged for various reasons, but the reason for us here is that in a multi-threaded environment (e.g., one without a GIL), installing the pygame_parachute function as signal handler eagerly and then later resetting to the old handler if one was installed is dangerous. Instead, we should first get the current handler and only install the pygame_parachute function as handler if the current one is SIG_DFL. This is still racy, but better than before.

This is an issue currently for the nogil fork of CPython, when running multiple subinterpreters of CPython in the same process, as well as running on runtimes like GraalPy which also support multiple interpreters in one process.

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -1,6 +1,7 @@
 /*
   pygame - Python Game Library
   Copyright (C) 2000-2001  Pete Shinners
+  Copyright (C) 2023  Oracle and/or its affiliates
 
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Library General Public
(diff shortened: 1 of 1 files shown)


illume:
Hello!

Very cool.

This is something I was looking into last year in order to avoid clobbering things provided by another API... but then I got distracted wondering if we can remove the pygame parachute entirely now. The parachute used to be required for cleanup with SDL1, but now I'm not so sure the pygame_parachute is required.

I'll do some testing on removing it, and get back to you. https://github.com/pygame/pygame/pull/3768

timfel:
Looks like removing it merged ok, I'll close this :)

PR #3766, opened 2023-03-17
Replaced dict with weakset in Sprites groups

### LINKED ISSUE #3763: Memory leak when a Group has no references but isn't empty
**Environment:**

- **Operating system** : macOS Monterey v12.6.3
- **Python version** : Python 3.11.2
- **PyGame version** : 2.3.0
- **Relevant hardware** : N/A

**Current behavior:**

When a Group is no longer referenced (except by its members) it will continue to exist in memory. If a Group is created on-the-fly for some reason and isn't emptied, it will never be deallocated, even after leaving its scope.

**Expected behavior:**

Groups should deallocate themselves when only being referenced by their own sprites, or at least have a bool property to toggle the suggested behavior.

**Steps to reproduce:**

1. create a Group with members in a function
2. do not have any references to it (except from the members)
3. use something like `len(member.groups())` to view the amount of groups increase every time the function is called

**Test code**

```
import pygame
pygame.init()


class Obj(pygame.sprite.Sprite):
    def __init__(self, x, y):
        super().__init__()
        self.rect = pygame.rect.Rect(x, y, 1, 1)


def colliding(obj: Obj) -> bool:
    if pygame.sprite.spritecollide(obj, pygame.sprite.Group([s for s in objects if s != obj]), False):
        return False
    return True


testObj = Obj(5, 4)
objects = pygame.sprite.Group(testObj, Obj(9, 1), Obj(12, 0))
while True:
    for obj in objects:
        print(len(obj.groups()))
    print(" ")
    colliding(testObj)

```
I am aware that not casting the list comprehension to a Group is valid, but this is just for example purposes.

theRealProHacker:
Fixes #3763 

We simply replace the `__g` `dict` with a `WeakSet`

Not tested but should work.


src_py/sprite.py
```diff
@@ -110,7 +111,7 @@ class Sprite:
     """
 
     def __init__(self, *groups):
-        self.__g = {}  # The groups the sprite is in
+        self.__g = WeakSet()  # The groups the sprite is in
```
illume, line 114: Maybe WeakKeyDictionary?

theRealProHacker: Why? Isn't `__g` a set?

theRealProHacker: Bruh, I didn't even think about the fact that `{}` is a `dict`. I just assumed it was a set.

But now I am asking myself why this isn't a set?

theRealProHacker:
Another update: I saw that there is also a Cython implementation. I have no idea how that works with `__weakrefs__` and generally I haven't ever written any Cython (only read a little).

src_py/sprite.py
illume: Sets weren't included with python back then, they were an external extension. So people mostly implemented sets with dicts.

illume:
The Cython version is mainly an experiment, and can be ignored. I seem to recall Cython supports weakrefs, so it should be possible.

src_py/sprite.py
theRealProHacker: I expected that but I also thought that the code would have been updated since then.

theRealProHacker:
So, should we then take https://github.com/pygame/pygame/commit/5a535e068e64facd3c24ed5e4e3552f411f295cc ?

If I understood correctly, the tests all passed. 
For the other stuff, I'll make a separate PR.

FirTreeMan:
Adding a WeakSet in place of a Set stops a Group existing with no references, right? I think having it be optional would be a good idea. The method I would like to implement does not include WeakRefs, so I think I should be making a separate fork.

FirTreeMan:
Have to eat my words a bit here, turns out I do indeed need WeakRef for my method.

FirTreeMan:
https://github.com/pygame/pygame/pull/3767

I'm very new to open-source contributing. Is this pull request similar enough that it should just be part of your fork? Please forgive me if so.

theRealProHacker:
> I think having it be optional would be a good idea. 

@FirTreeMan  Good point. I forgot about that. But if we want to change the behaviour in this way, we should probably subclass the default group with a `WeakGroup` that then automatically uses a `WeakSet`. IMO that makes a lot more sense. 

But I would still definitely use a set instead of a dictionary in both groups. If someone is relying on it being a dictionary, it's his own fault really. 

@illume I think it would be really cool if you could tell us what you think

illume:
@FirTreeMan 
> I'm very new to open-source contributing. Is this pull request similar enough that it should just be part of your fork? Please forgive me if so.

You can work together on this if you like. We can add a co author tag when we merge so you're both listed.

illume:
> I think having it be optional would be a good idea.
>
> @FirTreeMan Good point. I forgot about that. But if we want to change the behaviour in this way, we should probably subclass the default group with a WeakGroup that then automatically uses a WeakSet. IMO that makes a lot more sense.
>
> But I would still definitely use a set instead of a dictionary in both groups. If someone is relying on it being a dictionary, it's his own fault really.

In my humble opinion... we need to do two types of testing first before deciding if it should be optional:
1. add some unit tests
2. test with some older games to see if they're still working

theRealProHacker:
What do you think of this? It's totally backwards compatible, so we don't have to worry about crashing someone's code because he relied on it. 

Also, we could start a slow deprecation process and then make `WeakSprite` the default in some future version, if that's desirable.

theRealProHacker:
Okay, so it builds, but the tests fail that assert the "weak" behavior. I'll look into this, but I have problems building pygame on Windows, so the development is a bit difficult.

theRealProHacker:
Basically, I just tried to rewrite the tests. However, I am not sure whether this will even change anything. I'll have to do more when I have more time and can isolate all the relevant code and test it on my local machine.

theRealProHacker:
@illume if you have no special remarks, you can merge this know. The tests simply needed a `gc.collect()` to run successfully.

src_py/sprite.py
```diff
@@ -341,6 +348,10 @@ def __repr__(self):
         )
 
 
+class WeakDirtySprite(WeakSprite, DirtySprite):
```
illume, line 351: lol'd at the name :) Love it.

test/weak_test.py
```diff
@@ -0,0 +1,470 @@
+from weakref import WeakSet
```
illume, line 1: I'm not sure if we need this weak_test.py in here?

There seems to be enough tests in sprite_test.py for WeakSprite and such.

illume (commented):
Very cool. Thanks.

- I think you can remove the test/weak_test.py file now? I guess it was for debugging?
- there's some pylint issues mentioned in https://github.com/pygame/pygame/actions/runs/4532521201/jobs/8092649581?pr=3766#step:4:6
  - You can pip install pylint and then run `python setup.py lint` in the pygame repo.
- The docs would need updating to add the new Weak* classes. You can see the file is here: https://github.com/pygame/pygame/blob/main/docs/reST/ref/sprite.rst

illume:
https://github.com/pygame/pygame/actions/runs/4778373738/jobs/8616342876?pr=3766
Seems like the code needs to be formatted with `clang-format` by using `python setup.py format`.

illume (approved):
Thanks for this :)

theRealProHacker:
> Thanks for this :)

Thanks for formatting the code. That really helped, even though it was such a small thing to do.

PR #3767, opened 2023-03-18
Deletion of Group if Group has no references besides members

FirTreeMan:
Toggle behavior for new Groups with `AbstractGroup.terminate_if_unref` (defaults to disabled).


illume:
Very cool.

ps. I wrote in the other PR... but will repeat some of it here in case you missed it.

In my humble opinion... we need to do two types of testing first before deciding if it should be optional:

1. add some unit tests
2. test with some older games to see if they're still working

illume:
The test failure is saying it needs the `black` code formatter run on it.

```
python -m pip install black --user
python -m black src_py
```

FirTreeMan:
New commit is black formatted.

illume:
This https://github.com/pygame/pygame/issues/3763 is done now.

PR #3783, opened 2023-04-01
fixed self.pos of move method in GameObject in Moving Multiple Images

saif-11bit:
Error: Instead of moving the object horizontally it's moving vertically
`self.pos = self.pos.move(0, self.speed)`

Fix: Move the object horizontally
`self.pos = self.pos.move(self.speed, 0)`


illume (approved):
Thanks 👍

PR #3797, opened 2023-04-11
Add documentation for DualShock 5

Aashutosh-922:
Outline changes in buttons, axes and hat for PlayStation 5 official controller - DualShock 5 for pygame 2.x


illume (approved):
👍 thanks

PR #3807, opened 2023-04-21
Fix the signature of midi.get_device_info

SpeedyOrc-C:
When I try to retrieve the information of a connected MIDI device, it returns a byte string for device’s name, but this function’s signature says it returns a string.


illume (approved):
👍 thanks

PR #3823, opened 2023-04-29
Update ImportInit.rst

AKNation:


illume:
Hello. What was the idea here?

illume:
Probably this could be put somewhere else.  Maybe on your github account instead?
(I don't think it was meant to be put where the existing ImportInit.rst was.)

PR #3843, opened 2023-05-10
Add support for keyboard grab without mouse grab (expose SDL_GetWindowKeyboardGrab)

Yannik:
Currently, only [SDL_SetWindowGrab](https://wiki.libsdl.org/SDL2/SDL_SetWindowGrab) is exposed in pygame, which grabs the mouse, and, optionally, the keyboard.

In situation where you only need to grab the keyboard, always being forced to grab the mouse may be undesirable. (Atleast it is in an application I am developing).

To fix this, I am exposing the [SDL_GetWindowKeyboardGrab](https://wiki.libsdl.org/SDL2/SDL_GetWindowKeyboardGrab) method as `pygame.event.set_keyboard_grab(bool)` and `pygame.event.get_keyboard_grab()` in this PR.


Yannik:
@illume Would you be able to give feedback on why the CI builds are still failing? I fixed the obvious problems (e.g. clang linter), but don't know what else to do now. Locally, this builds and works for me.

illume:
Hey,

Click on the details links for the checks that are failing. There's a bunch of noise, but in there you should be able to see what is happening.

#### https://ci.appveyor.com/project/pygame/pygame/builds/47008471/job/el4cuexts8uj4oqq

```c
src_c/event.c(2255): error C2065: 'DOC_PYGAMEEVENTSETKEYBOARDGRAB': undeclared identifier
   Creating library build\temp.win-amd64-3.9\Release\src_c\key.cp39-win_amd64.lib and object build\temp.win-amd64-3.9\Release\src_c\key.cp39-win_amd64.exp
Generating code
src_c/event.c(2257): error C2065: 'DOC_PYGAMEEVENTGETKEYBOARDGRAB': undeclared identifier
Finished generating code
src_c/event.c(2254): error C2099: initializer is not a constantFinished generating code
building 'pygame.rect' extension
```



#### https://github.com/pygame/pygame/actions/runs/4937250903/jobs/8831461651?pr=3843#step:5:327

```
Generating docs caused changes. Please check them in.
You may need to run: python3 setup.py docs --fullgeneration
 M src_c/doc/event_doc.h
```

This one means docs/reST/ref/key.rst needs to have the docs put in there. Then running `python setup.py docs` will generate the .h file.



#### https://github.com/pygame/pygame/actions/runs/4937250895/jobs/8831457863?pr=3843#step:3:1167

```c
    src_c/event.c:2255:6: error: ‘DOC_PYGAMEEVENTSETKEYBOARDGRAB’ undeclared here (not in a function); did you mean ‘DOC_PYGAMEEVENTSETGRAB’?
     2255 |      DOC_PYGAMEEVENTSETKEYBOARDGRAB},
          |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          |      DOC_PYGAMEEVENTSETGRAB
    src_c/event.c:2257:6: error: ‘DOC_PYGAMEEVENTGETKEYBOARDGRAB’ undeclared here (not in a function); did you mean ‘DOC_PYGAMEEVENTGETGRAB’?
     2257 |      DOC_PYGAMEEVENTGETKEYBOARDGRAB},
          |      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
          |      DOC_PYGAMEEVENTGETGRAB
```

src_c/event.c
```diff
@@ -1591,6 +1591,41 @@ get_grab(PyObject *self, PyObject *_null)
...
+    int doit = PyObject_IsTrue(arg);
+    if (doit == -1)
+        return NULL;
+
+    VIDEO_INIT_CHECK();
+
+    SDL_Window *win = pg_GetDefaultWindow();
+    if (win) {
+        if (doit) {
+            SDL_SetWindowKeyboardGrab(win, SDL_TRUE);
```
illume, line 1606: There needs to be a guard around SDL_SetWindowKeyboardGrab and SDL_GetWindowKeyboardGrab in case it's not available in older versions of SDL. When it's not available just do nothing I think.

```c
#if SDL_VERSION_ATLEAST(2, 0, 14)
    SDL_Window *win = pg_GetDefaultWindow();
    if (win) {
    ...
#endif
Py_RETURN_NONE;
```

Yannik:
@illume Ah, I see. I did not know that `event_doc.h` was autogenerated from reST. Thanks for the explanation.
I have fixed that and added the SDL version check.

illume:
Hi again.

Probably we need to add a couple more things here to finish it off.

- `.. versionadded:: 2.5.0` to the docs. See other docs.
- A basic unit test that calls the functions. It should only run if at least SDL 2.0.16
- Bonus points for an interactive test. (optional)

The test needs to be marked with this:
```python
    @unittest.skipIf(pygame.get_sdl_version() < (2, 0, 16) , "Needs at least SDL 2.0.16")
```

Yannik:
Hi @illume 

I have updated my PR with `versionadded` and a test. It tests both set and get in a single test (calling set and then verifying the state with get), which made the most sense to me.

Yannik:
I cannot figure out why the test is failing.

I have tried this locally:
```
import pygame

pygame.init()

surf = pygame.display.set_mode((10, 10))
pygame.event.set_keyboard_grab(True)
print(pygame.event.get_keyboard_grab())
pygame.event.set_keyboard_grab(False)
print(pygame.event.get_keyboard_grab())
```

and it results in the correct
```
True
False
```

@illume Could you have a look?

illume:
I think it's because the 'dummy' video driver doesn't implement it correctly. It seems to just ignore it.

So we can skip it when running under the dummy driver:

```python
    @unittest.skipIf(
        "requires the SDL_VIDEODRIVER to be a non dummy value",
    )
```

Yannik:
@illume Nope, this does not work. Maybe it suffers from the same problems as the `test_get_grab` test (which is always skipped)?

PR #3849, opened 2023-05-15
Update sprite.py

sarthaknimbalkar:
Here are the changes that I've made 
From line 1212 on-wards: 
-     spr.dirty = 0: This line tells the sprite (spr) that it doesn't need to be redrawn or updated for now.
-     if self._time_threshold > 0: This checks if a certain time threshold is greater than zero. 
-     pygame.display.flip(): If the time threshold is greater than zero, this line updates the entire display on the screen. It makes all the changes visible at once.
-     If the time threshold is not greater than zero, the code continues here:
-     dirty = [spr.rect for spr in local_sprites if spr.dirty]: This line creates a list called dirty. It goes through each sprite in a group of sprites called local_sprites and adds their rectangles to the dirty list if they need to be redrawn or updated (if spr.dirty is true).
-     if local_bgd is not None and local_bgd.dirty:: This checks if there is a background object (local_bgd) and if it needs to be redrawn or updated.
-     dirty.append(local_bgd.rect): If the background object satisfies the condition above, its rectangle is added to the dirty list.
-     pygame.display.update(dirty): Finally, the update() function is called with the dirty list. This function updates only the specified areas of the screen that need to be redrawn or updated, based on the rectangles in the dirty list.

### DIFF
--- a/src_py/sprite.py
+++ b/src_py/sprite.py
@@ -85,13 +85,15 @@ Sprites are not thread safe, so lock them yourself if using threads.
 # specific ones that aren't quite so general but fit into common
 # specialized cases.
 
-from warnings import warn
 from weakref import WeakSet
+from warnings import warn
+
 
 import pygame
-from pygame.mask import from_surface
+from pygame import display
 from pygame.rect import Rect
 from pygame.time import get_ticks
+from pygame.mask import from_surface
 
 
 class Sprite:
@@ -393,8 +395,7 @@ class AbstractGroup:
     def add_internal(
         self,
         sprite,
-        layer=None,
-        # noqa pylint: disable=unused-argument; supporting legacy derived classes that override in non-pythonic way
+        layer=None,  # noqa pylint: disable=unused-argument; supporting legacy derived classes that override in non-pythonic way
     ):
         """
         For adding a sprite to this group internally.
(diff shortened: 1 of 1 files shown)


illume:
Hey hey. It seems it doesn't like draw or update running inside the draw method. You can see the specific error by clicking on the details and seeing the traceback. https://github.com/pygame/pygame/actions/runs/4983576559/jobs/8965797484?pr=3849#step:3:5668

```
  ERROR: test_repaint_rect_with_clip (pygame.tests.sprite_test.LayeredDirtyTypeTest__DirtySprite)
    File "/tmp/tmp.OsaUYlhnDI/venv/lib/python3.6/site-packages/pygame/tests/sprite_test.py", line 1105, in test_repaint_rect_with_clip
      group.draw(surface)
    File "/tmp/tmp.OsaUYlhnDI/venv/lib/python3.6/site-packages/pygame/sprite.py", line 1214, in draw
      pygame.display.flip()  # Use flip mode
```

SDL_VIDEODRIVER=dummy python -m pygame.tests

Is there a way to fix it without calling update/display in there?

sarthaknimbalkar:
Okay I'll something else, re-test and come back.  
Thanks.

sarthaknimbalkar:
Hey I removed the part where display was being used and it seems to be working when I explicitly set sprite=0. Give it a try now.
Thanks

PR #3907, opened 2023-06-07
Dropevent test branch

amirnajafiosu:
A unittest for dropevent.py in the examples folder. Thank you.


illume (changes requested):
Hello. Thanks for this.

A few notes...

- Would you mind putting the test into test/event_test.py?
- Also, we have an example for the drop events already.
- run the code formatter on it. `python -m pip install black && python -m black test/event_test.py`


cheers!

illume:
Hello.

I read this, but it doesn't make sense and it doesn't run.

Was this made with chatgpt or similar?

Feel free to open another PR if you'd like to try again.

PR #3911, opened 2023-06-11
Refactor `chimp.py` to be more object-oriented

veganaize:
I've refactored the `chimp.py` example to make it more consistently object-oriented.

_I'd imagine this change would necessitate updating the tutorial, as well._

Please feel free to slam this idea, as appropriate...

P.S.  One of the benefits -- you can add a second chimp by simply uncommenting line 203 `#Game.chimp2 = Chimp((90,10))`
I explain it all in a video here: https://youtu.be/10XZVvOrarw

### DIFF
--- a/examples/chimp.py
+++ b/examples/chimp.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 """ pygame.examples.chimp
-
 This simple example is used for the line-by-line tutorial
 that comes with pygame. It is based on a 'popular' web banner.
 Note there are comments here, but for the full explanation,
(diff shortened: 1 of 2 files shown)


examples/chimp.py
```diff
@@ -1,203 +1,234 @@
...
-
-# this calls the 'main' function when this script is executed
+    def get_punched(self, obj):
+        """Cause monkey to spin"""
+        self.move = self._spin
+
+        
+class Game:
+
+    def initialize():
```
illume, line 167: The methods should have self as their first argument.

illume (commented):
Thanks.

I'm unsure if we should make these changes. Because (as you say) there are tutorials based on this file.

I do like that you've made a video about refactoring it.

examples/chimp.py
veganaize: It's effectively a static class.  So it's a static method, as well.

I couldn't justify making `Game` instantiable on this iteration but if you think it's worth the added complexity we can refactor it.

veganaize: _I'd be willing to refactor the tutorial, as well, as part of this PR._

illume:
Hey hey.

What do you think about making this a new refactoring tutorial titled like this PR... "Refactoring chimp.py to be more object-oriented"? Like video I guess.

It's kind of the path that many people take going from functions to classes. People often make a Game class, or a Scene class like in here. That it shows the benefit of OOing by adding the second Chimp is really nice.

I feel the static methods are a bit much for an intro-level tutorial like chimp, but would be brilliant in a "Refactoring chimp.py to be more object-oriented" style tutorial.

Issue #3737, opened 2023-03-01
Documentation is unclear on several issues


theRealProHacker:
1. For me, the biggest problem are the event attributes. While most are relatively obvious, there are many that are not really. For example
    - Keyboard events: What is the difference between `key`, `mod`, `unicode` and `scancode` and how can I use them?
    - ACTIVEEVENT attributes and meaning?
    - MOUSEWHEEL: which, flipped, precise_x, precise_y?
    - TEXTEDITING: When is this triggered?
    - AUDIODEVICE*: What is `which`? What can I do with it? Why is it there?
    - FINGER* and MULTIGESTURE?
    - DROP*: When (and in which order) are these events emitted?
    - MIDI*: What is the difference to AUDIODEVICE*?
    - Many (if not all) events also have a "window" attribute. What does that mean? What should I do with it?
2. I will add more here when I find more

I don't think "These questions are answered in tutorials and examples" is neither totally factual (afaik) nor a good response. In my opinion, this information should be contained in the reference.

- [ ] Unrelated: "The following is a list event types with their specific attributes." in https://www.pygame.org/docs/ref/event.html misses an "of" before "event"
- [ ] Add the above-mentioned unclear things to the documentation

illume:
Some good points. Thanks.

theRealProHacker:
I would like to make a PR. But honestly, a lot of this I don't know or would have to experiment with. Maybe you could help me a bit on that. 

For fonts, I found a good source for understanding some of the things mentioned above: https://fonts.google.com/knowledge/choosing_type/exploring_x_height_the_em_square

jack5790:
I'm going to try to answer everything at once here, if there's any confusion, just leave a reply and I'll look into it.
1. A: pygame.key is a module that holds all functions related to the keyboard, not keyboard events themselves. All pygame events are queued in the pygame.event module.
    B: A "key mod" is any key that is used to "modify" other keys. CTRL and Shift keys are obvious examples of modifying keys.
    C: The unicode and scancode attributes are gathered with the event KEYDOWN. The unicode attribute will hold the unicode value of whichever key was pressed, (this may be useful for getting letters pressed with the shift key, but will include letters pressed with caps lock on.) Scan code will return the platform specific key. Here's how a StackOverflow user describes it.

[...]

PR #3912, opened 2023-06-11
Update documentation on several events

RyanNicoletti:
Issue: [Documentation is unclear on several issues #3737](https://github.com/pygame/pygame/issues/3737)

Added requested documentation for events and attributes listed in issue #3737 

I did not include additional documentation for Fonts, the methods (ie: get_linesize) do a good job (I think) of explaining what the listed attributes (height, bold, etc...) do. I could not find any reference as to how doing things like boldening font would affect height/linesize, etc..


illume (approved):
Great. Thanks for that. 🎉

PR #3918, opened 2023-06-14
Updating "Python Pygame Introduction"

Ronoc21:
Updated the resolution size of the code snippet to 1920 x 1080. The previous resolution was extremely small and caused the ball to bounce too fast. Added some clarification about making sure the intro_ball.gif file can be found by the code snippet. The snippet itself was not displaying properly either. Added an image file to display it instead of formatting in github


docs/reST/tut/PygameIntro.rst
```diff
@@ -63,43 +63,18 @@ version of that same thing. This should be simple enough to follow
...
-       pygame.display.flip()
+.. image:: introsnippet.PNG
 
 This is as simple as you can get for a bouncing animation.
 First we see importing and initializing pygame is nothing noteworthy.
 The ``import pygame`` imports the package with all the available
 pygame modules.
 The call to ``pygame.init()`` initializes each of these modules.
-
+Make sure the gif file of the bouncing ball is in the same folder
```
illume, line 73: This is a nice change, thanks.

docs/reST/tut/PygameIntro.rst
```diff
@@ -63,43 +63,18 @@ version of that same thing. This should be simple enough to follow
...
-       ballrect = ballrect.move(speed)
-       if ballrect.left < 0 or ballrect.right > width:
-           speed[0] = -speed[0]
-       if ballrect.top < 0 or ballrect.bottom > height:
-           speed[1] = -speed[1]
-
-       screen.fill(black)
-       screen.blit(ball, ballrect)
-       pygame.display.flip()
+.. image:: introsnippet.PNG
```
illume, line 66: I don't think the code listing should be removed for an image. One reason is the code listing can be copy/pasted, but the image not.

Can you please remove this change?

illume (changes requested):
Thanks.

I left a request about adding the code listing back.

docs/reST/tut/PygameIntro.rst
Ronoc21: let me know if the new edit is ok, the reasoning is, the code block doesn't show up at all on the github page, but I don't want to negatively impact https://www.pygame.org/docs/tut/PygameIntro.html, i can remove the change altogether and go back to how it was if needed

illume: It seems removing the linenos fixed it so it displays in github again. Thanks.

PR #3926, opened 2023-06-18
key: Allow list(get_pressed()) again

illume:
So it doesn't crash apps that used to work.


Dreamsorcerer:
Personally, I'm still not sure this is the right move. We know that using the behaviour following pygame 1 code breaks the app in non-obvious ways, and nobody has managed to provide an actual use case for it to exist yet.

PR #3963, opened 2023-07-21
Improvement

OldBoxingKing:
This pull request is set to fixed the issue pygame#3930.The issue pygame#3930 doesn't seem to have any problem at all.I test it on my own laptop and it works out fine.However,I think there is some space for improvement here.

# Draft
There is no window set up to receive keyboard events.
-You need to add some display settings before the game loop, such as using pygame.display.set_ mode() function which is used to create a window.
-You need to update the screen display during the game loop, such as using the pygame.display.flip() or pygame.display.update() functions.
-Otherwise, the window may lose focus and fail to respond to keyboard events

There are other ways to improve these codes.
-You can use pygame.key.get_pressed () function is used to obtain the status of all buttons, rather than using a list to record them.This can simplify the code and also support more buttons.
-You can use the pygame.draw.rect() function to draw blocks instead of concatenating them with line segments.It can make the blocks clearer and also conveniently set the position and size of the blocks.
-You can use the pygame.sprite module to manage blocks instead of using loops to draw.It can make the code more modular and also conveniently add block movement and collision detection functions.
-You can use the pygame.mixer module to load and play audio files, rather than just silent games.
-You can use the pygame.font module to display some text information, such as scores, time, prompts, etc.It can make the game more complete and increase interaction

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,4 @@
+Hello World
 .. image:: https://raw.githubusercontent.com/pygame/pygame/main/docs/reST/_static/pygame_logo.svg
   :alt: pygame
   :target: https://www.pygame.org/


OldBoxingKing:
I defined some constants to represent colors and buttons, which can improve the readability and maintainability of the code.
Also I simplify the logic of event processing by using a loop to check the status of buttons instead of multiple "if" statements.
我定义了一些常量来表示颜色和按键，这样可以提高代码的可读性和可维护性。
并且简化了事件处理的逻辑，使用一个循环来检查按键的状态，而不是多个if语句。

As regards to the issue you mention,it may be caused by a change in the draw.line method in Pygame version 2.3.0. Before Pygame version 2.1.1, the draw.line method caused half of the width of the lines to exceed the rectangular area. After Pygame version 2.1.1, the draw.line method correctly draws the width of lines within a rectangular area, which may affect the position and effect of the lines. You can try adjusting the coordinates and width parameters in your code
至于你遇到的问题可能是由于 pygame 2.3.0 版本中 draw.line 方法的变化导致的。在 pygame 2.1.1 版本之前，draw.line 方法绘制线条会导致线条的一半宽度超出矩形区域。在 pygame 2.1.1 版本之后，draw.line 方法会正确地在矩形区域内绘制线条的宽度，这可能会影响线条的位置和效果。你可以尝试调整您的代码中的坐标和宽度参数

Here are the codes I improved.下面是我优化的代码：

import pygame

screen = pygame.display.set_mode([400, 720])
pygame.display.set_caption("Piano Block")
clock = pygame.time.Clock()

WHITE = [255, 255, 255]
BLACK = [0, 0, 0]
BROWN = [237, 189, 101]
LIGHT_BROWN = [255, 222, 173]
KEYS = [pygame.K_d, pygame.K_f, pygame.K_j, pygame.K_k]

keydown = [False] * 4

keepgoing = True

while keepgoing:
# Handle the events
# Quit if the user closes the window or presses Q
if event.type == pygame.QUIT or (event.type == pygame.KEYDOWN and event.key == pygame.K_q):
keepgoing = False
break
# Update the key states if the user presses or releases a key
elif event.type == pygame.KEYDOWN or event.type == pygame.KEYUP:
for i in range(4):
if event.key == KEYS[i]:
keydown[i] = (event.type == pygame.KEYDOWN)
break
# Draw the screen
if keepgoing:
screen.fill(WHITE)
for i in range(4):
color = LIGHT_BROWN if keydown[i] else BROWN
pygame.draw.line(screen, color, [(i * 100) + 5, 650], [(i * 100) + 95, 650], 5)
# Draw three vertical lines to separate the keys
for i in range(1, 4):
pygame.draw.line(screen, BLACK, [i * 100, 0], [i * 100, 720], 3)
# Update the display
pygame.display.update()
clock.tick(30)

illume:
Hello. It seems like you are just testing with PR? It just has “Hello world” in there.

I’ll close it, but feel free to let me know if you were planning to continue something here. We can open it back up if that is the case :)

PR #3973, opened 2023-08-06
Removed the unused global score variable

alroyer:
I removed the reference to the global variable 'score' which does not exist and was not used.


illume (approved):
👍 thanks

Issue #3914, opened 2023-06-14
Function Comment For list_cameras


dpatters550:
**Description**

I am completely new to this project and contributing to open source in general, and I wanted to make a small contribution in the form of a function comment. I noticed that the list_cameras function on _camera_opencv.py file is missing a function comment. I also noticed that a description is in the docs for this project. I was thinking of adding the description from the docs into the Python file itself. This is a fairly small change, so I did not know if this needed an issue or not.

PR #3988, opened 2023-08-27
Update some comments in _camera_opencv.py

N7z:
Updated some comments in _camera_opency.py file.
In response to issue #3914


illume:
Hello.

See there is some documentation for list_cameras in: docs/reST/ref/camera.rst This is closer to what is expected to be in there.

Issue #3922, opened 2023-06-15
Missing import information in camera and sdl2 controller modules documentation


kkatsuro:
To initialize sdl2 controller and camera modules it is necessary to import it first, but you can use joystick module without importing. I have not found any information about need of importing it in documentation.

https://www.pygame.org/docs/ref/sdl2_controller.html
https://www.pygame.org/docs/ref/camera.html
https://www.pygame.org/docs/ref/joystick.html

illume:
Hello. Good point.

If someone wants to make a PR for adding a line something like the following to the docs for this:
"Use import pygame.camera before using this module."

- https://github.com/pygame/pygame/blob/main/docs/reST/ref/sdl2_controller.rst
- https://github.com/pygame/pygame/blob/main/docs/reST/ref/joystick.rst
- https://github.com/pygame/pygame/blob/main/docs/reST/ref/camera.rst

jjosecodes:
is this closed ?
i can do it quickly if needed

aguirre-lucas:
Hi, if this hasn't been closed, I would like to work on it!

Wiz-Rayyan:
> Hi, if this hasn't been closed, I would like to work on it!

Apparently it's closed. Good to do it.

Mandru-JashwanthKumar:
is it closed

kkatsuro:
I just tested with pygame 2.6.1 and `pygame._sdl2.controller.init()` worked for me with importing just pygame, but didn't test if camera module works too.

PR #4001, opened 2023-09-16
Added import information in camera and sdl2 controller modules docs

AsaadAreeb:
#3922  have added a Note in the documentation to import the modules before using them.

Capture1
Capture


illume (approved):
🎉🎈
thanks

PR #4023, opened 2023-09-30
Fix support for OpenBSD and NetBSD

aX0Rb:
Hi.  I wasn't able to get Pygame to compile on OpenBSD due to the OS not having the <linux/videodev2.h> header and instead using <sys/videoio.h>.  This is my first ever actual pull request to a project so if someone could check to make sure my change doesn't break anything that would be greatly appreciated.  Thank you.


illume (approved):
👍 thank you

PR #4029, opened 2023-10-05
Added a minor check if config is correctly initialised

blackmambaza:
Made a minor update to the build config for unix to validate "config" and to throw an error if config is not initialised correctly


illume (approved):
🎉🎈 thanks

PR #4031, opened 2023-10-05
Updated the reST/themes/classic/elements html documention

blackmambaza:
Added alt text and modernised some of the html tags in the documentation


illume (approved):
👍 thanks for this.

PR #4038, opened 2023-10-07
Minor improvements to tests

blackmambaza:


blackmambaza:
I think that we should add a CI step to run python3 setup.py format before the lint. What do you think?

PR #4043, opened 2023-10-08
Updated the pygame.font.SysFont function so that it can accept keyword arguments

atif5:
A trivial change to the `pygame.font.SysFont` function, so that it accepts keyword arguments. 
Also added a unit test to check whether the function can accept keyword arguments.

Originally, it would raise a `TypeError` if one would try to call the function with any keyword arguments.

Related to #808.

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -389,8 +389,8 @@ def font_constructor(fontpath, size, bold, italic):
 # the exported functions
 
 
-def SysFont(name, size, bold=False, italic=False, constructor=None):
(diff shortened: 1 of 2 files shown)


illume:
Hello. 

It seems the original issue was not so clear. It was not about allowing things to accept any keyword params, but allowing functions which did not accept keywords for **the existing params**.

My apologies it was not clear.

atif5:
Actually, my apologies for not understanding the issue clearly. Will close the PR now.

Issue #4046, opened 2023-10-13
A Python 3.12 Build


llindstrom:
Python 3.12 has just been released. This is a request to have pygame available for the newest Python release.

#### The platform
- old PC with Intel Pentium multi-core processor
- Intel integrated graphics
- Linux Mint 21.2 Cinnamon (i86_64) (Ubuntu 22.4) using X11
- Python3.12  3.12.0-1+jammy1 apt package from the deadsnakes repository
- all pygame dependencies loaded as standard apt packages
- numpy 1.26.0 and PyOpenGL 3.1.7 from PyPI 
#### Build and Install
The **distutils** package was removed from Python 3.12. It can be restored by installing **setuptools**. The `python3.12 setup.py -config -auto` and `python3 setup.py install --user` incantation works, but gives a deprecation warning message and lots of compiler warnings regarding C pragmas.
#### Testing
The basic unit tests pass (`python3.12 -m pygame.tests`), Of the examples tried, **go_over_there** raises an exception regarding passing a float to **random.randint**. Example **glcube** raises the exception:

    py3.12-linux-x86_64.egg/pygame/examples/glcube.py", line 541, in main
        pg.display.set_mode(display_size, pg.OPENGL | pg.DOUBLEBUF | pg.RESIZABLE)
    pygame.error: Could not create GL context: GLXBadFBConfig

This could be due to the old hardware used. Example **scrap_clipboard** has issues unrelated to Python 3.12. Examples **audiocapture**. **camera**, **joystick**, and **midi** were skipped.
### Conclusion
For the most part pygame works with Python 3.12 as is. A few small changes to the Python code, such as with **pygame.examples.randint**, may be need. The build process needs updating to whatever has replaced **distutils**. Also the pesky C warnings during compilation should be checked into.

llindstrom:
### Further Notes 
- pygame built from revision 9bb56587 cloned from repository.
- The Python **pip** package installer looks for a _pyproject.toml_ file described in [PEP 518](https://peps.python.org/pep-0518/), [PEP 517](https://peps.python.org/pep-0517/), [PEP 621](https://peps.python.org/pep-0621/) and [PEP 660](https://peps.python.org/pep-0660/). This file lists the build dependencies, such as [setuptools](https://pypi.org/project/setuptools/), needed to build the package. These dependencies are automatically fetched by **pip** during a build.

sebastic:
[...]

PR #4048, opened 2023-10-14
Fix **go_over_there** example bug in Python 3.12 as per issue #4046

llindstrom:
For Python 3.12, **pygame.examples.go_over_there** fails with exception:

    [snip]
      File "[...]/pygame/examples/go_over_there.py", line 43, in reset
        random.randint(0, SCREEN_SIZE.x), random.randint(0, SCREEN_SIZE.y)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        [snip]
    TypeError: 'float' object cannot be interpreted as an integer

It is a documented Python 3.12 change that `randint` now only accepts integer arguments. This pull request fixes the problem and is backward compatible.

### DIFF
--- a/examples/go_over_there.py
+++ b/examples/go_over_there.py
@@ -40,7 +40,8 @@ def reset():
     balls = []
     for x in range(MAX_BALLS):
         pos = pg.Vector2(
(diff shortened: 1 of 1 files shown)


illume (approved):
👍 thanks

Issue #3976, opened 2023-08-07
Different Clock.tick() method


RaffaeleFiorillo:
**Description**

Hi, I have been using pygame for quite a few projects and, for games where I want to use a frame-independant approach, I use this trick:

delta = clock.tick(FPS)/1000

delta gives me the time between frames, which I can later use for objects movement. 

I would sugest creating a new method that directly returns the value of delta without having to make the division every frame.

illume:
Hello.

So it seems the next step is to come up with a proposal on how this could be done exactly.


> delta = clock.tick(FPS)/1000

So this gives the results in seconds. Note, for any other readers... Clock.tick already gives the amount of time since the last tick() call. But it returns the results in milliseconds not seconds.

I see two possibilities here:

1) method argument to request seconds

> clock.tick(framerate=30, seconds=True)

2) Clock setting

Then all the methods return seconds instead.

> clock = Clock(seconds=True)
> delta_seconds = clock.tick()

The benefit of the first one is that you don't need to supply or change all the methods with a separate argument.


What do you think?

RaffaeleFiorillo:
I think the first option is good. It also makes clear what the function is doing. In cases where the clock instance is created in a different place (module, file) you wouldn't be confused during debugging.

prathamroy:
I also agree that the first proposal is a better idea because of the aforementioned reasons.

krobinson395:
Hello,
I took the suggestions from the first proposal by @illume and executed it in PR #4162. I'm open for any and all feedback for any revisions I should make if this is still an interest!

PR #4053, opened 2023-10-19
Implemented a method for frame-independent time calculations.

prathamroy:
Implemented the method as requested in issue - Different Clock.tick() method #3976.

### DIFF
--- a/src_py/delta_time.py
+++ b/src_py/delta_time.py
@@ -1,0 +1,12 @@
+import pygame
+
+def get_delta_time(clock, fps):
+    return clock.tick(fps) / 1000
+
+# Sample usage. The value of FPS can be modified as per requirement.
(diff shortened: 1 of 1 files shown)


illume:
Thank you for this. 

I left some discussion on https://github.com/pygame/pygame/issues/3976 I think the next step there is coming up with a proposal for how this should be solved exactly.

prathamroy:
Yeah! Based on the discussion in #3976, I believe using the method argument to request seconds, would be the ideal next step.

prathamroy:
Just wondering what the next step for my implementation would look like. I have already created an additional method using delta_time.py and I'm unsure of where I go from here to implement the above.

illume:
Would you mind keeping the joystick documentation as a separate pull request?

PR #4055, opened 2023-10-20
Update PygameIntro.rst

kcivin:
if you write black on screen.fill(black) line its shows an arrow screen.fill("black") it worked like this in my computer


illume (approved):
Thanks 👍

Issue #3989, opened 2023-09-01
pygame.draw.polygon() not working properly with big coordinates


Mostercoder:
**Environment:**

- **Operating Windows (11)**
- **Python 3.11.5** 
- **SDL 2.28.2**
- **PyGame 2.5.1**

**Current behavior:**

You will see two triangles: a red one only consisting of the wireframe and one partially filled in green.

**Expected behavior:**

The green triangle should fill the red one out, as they are the same triangle with the only diffrence being, that the first one should get filled.

**Screenshots**

![image](https://github.com/pygame/pygame/assets/84497752/65d88ded-6bbe-4cfe-990d-166adc13241a)

**Steps to reproduce:**

1. Open code editor
2. Paste in code
3. Run

**Test code**

```
import pygame

pygame.init()

screen_width = 2560
screen_height = 1440

screen = pygame.display.set_mode((screen_width, screen_height))

white = (255, 255, 255)
black = (0, 255, 0)
red = (255,0,0)

while True:

    screen.fill(white)

    # Draw the triangle twice once with lines once without
    pygame.draw.polygon(screen, black, ((713+ screen_width/2-1000,screen_height/2 - 471+1000),(440+ screen_width/2-1000,screen_height/2 - 934+1000),(58356+ screen_width/2-1000,screen_height/2 - 38726+1000)))
    pygame.draw.polygon(screen, red, ((713+ screen_width/2-1000,screen_height/2 - 471+1000),(440+ screen_width/2-1000,screen_height/2 - 934+1000),(58356+ screen_width/2-1000,screen_height/2 - 38726+1000)),2)

    pygame.display.flip()
```

illume:
Hello. Thank you for reporting this!

Here's some next steps for anyone interested in working on this:
- Add a unit test to tests/draw_test.py that fails showing the property that it should hold
- Fix the polygon function in src_c/draw.c

u7156704:
Hi there, I am an ANU student and I would love to work on and be assigned to this issue.
I have the ability to replicate the environment with the operating system and believe I can complete this issue.

u7156704:
I'm trying to make a push with test files onto github, but I do not have permission. Is it possible I could gain access so I can continue work on this issue and have my pushes saved on the github for a possible peer review

Zireael07:
@u7156704 you need to fork pygame and push to your fork

u7156704:
I've forked pygame, but when I attempt to push my branch, this pops up:

[...]

PR #4056, opened 2023-10-21
Polygon large coordinate test cases for issue#3989

u7156704:
Issue: #3989 
I was unable to bug fix the large coordinate polygon issue, but was able to write test cases if someone wanted to fix this issue.
I also manage to narrow down the problem for large negative y-axis points for polygons are producing the bug.
I was unsure if this was a memory issue with the larger integers or something affecting the y-axis point value.
The test cases in draw_test.py should assist in fixing this issue.

### DIFF
--- a/test/draw_test.py
+++ b/test/draw_test.py
@@ -4281,7 +4281,6 @@ class DrawPolygonMixin:
     def test_polygon_large_coords(self):
         """
         Ensures draw polygon works correctly with large points.
-
         Testing the drawings of filled polygons
         """
         point_a = (600, 50)
@@ -4343,7 +4342,6 @@ class DrawPolygonMixin:
         self.draw_polygon(surface, green, (point_a, point_b, extreme_negative_y))
         extreme_y_pass(self)
         extreme_y_fail(self)
-
 
 class DrawPolygonTest(DrawPolygonMixin, DrawTestCase):
     """Test draw module function polygon.
--- a/test/draw_test.py
+++ b/test/draw_test.py
@@ -4278,70 +4278,74 @@ class DrawPolygonMixin:
                     self.assertEqual(surface.get_at(pt), expected_color, pt)
 
                 surface.unlock()
-    def test_polygon_large_coords(self):
-        """
-        Ensures draw polygon works correctly with large points.
-        Testing the drawings of filled polygons
-        """
-        point_a = (600, 50)
-        point_b = (50, 600)
-        extreme_points_coords = (58000, 100000)
-        extreme_negative_coords = (-58000, -100000)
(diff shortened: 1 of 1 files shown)


u7156704:
PyObject * Polygon (lines 775 – 890)
Draw_fillPoly (lines 2322 - 2428)

This functions in src_c/draw_c is a good starting place to fix this bug from my findings

illume:
Nice one! 🎉

Some notes...

- You can mark the test with `@unittest.expectedFailure `  This way the test runners succeed. It means "we know this one will fail".
- There's something weird with the first commit. Maybe try creating a new branch from main and copy the code in? First delete your local branch, then create a new branch with the same name off main, finally copy your code in and make a new commit.

u7156704:
I manage to change the test cases to include @unittest.expectedFailure, however I couldn't manage to fix the build. If someone could try fix it for me, I can create a new issue to extend based of the issue and this PR.

illume:
Hi,

you can see the test failures on the github actions by clicking on the checks.
I've pasted the failure below.


It might be better to put just one expectedFailure decorator on the whole test.

```
@unittest.expectedFailure
def test_polygon_large_coords(self):
```



```
FAIL: test_polygon_large_coords (pygame.tests.draw_test.DrawPolygonTest.test_polygon_large_coords)
Ensures draw polygon works correctly with large points.
  File "C:\Python312\Lib\site-packages\pygame\tests\draw_test.py", line 4339, in test_polygon_large_coords
    extreme_x_pass(self)
  File "C:\Python312\Lib\site-packages\pygame\tests\draw_test.py", line 4314, in extreme_x_pass
    self.assertEqual(surface.get_at((600, 600)), white)
AssertionError: (0, 255, 0, 255) != (0, 0, 0, 255)
Ran 217 tests in 4.557s
```

u7156704:
I can divide the larger test case into two separate but similar functions, one for expected passing tests and one for expecting failing tests. The way it was written to test different shapes to ensure the bug does occur and is properly fixed, leading to a larger quantity of tests, but I can try simplify it down or comment out some test cases.

u7156704:
I've commented out of the tests, for a future issue

illume:
Ok. Let me know if you want some help continuing this?

u7156704:
I believe commenting out the code was the main solution, as the required checks pass now. This would probably be the best way at the moment while the bug exists. I recon if this PR get approved, I'll reference this PR and the previous issue in a new issue that addresses the bug more directly with the commented out code as test cases for it

illume:
> I believe commenting out the code was the main solution

Good point.

illume (approved):
thanks 🎈🎉

Issue #4015, opened 2023-09-19
Empty Destination Palette when copying a text surface


Sapphirewhale:
**Environment:**

You can get some of this info from the text that pops up in the console when you run a pygame program.

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Windows 11
- **Python version** (e.g. 3.7.9, 3.8.5) : 3.11.4
- **SDL version** (e.g. SDL 2.0.12): 2.28.2
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.5.1
- **Relevant hardware** (e.g. if reporting a bug about a controller, tell us the brand & name of it):

**Current behavior:**

When attempting to copy a surface created by pygame.font.SysFont().render with both the color and background values set to white (specifically (255,255,255)), a pygame.error is thrown with message "Empty destination palette"

**Expected behavior:**

Either for the (admittedly useless) pure white surface to be returned, or for a more specific error to be thrown at the pygame.font.SysFont().render call explaining that the color and background parameters cannot both be set to white

**Screenshots**

![image](https://github.com/pygame/pygame/assets/49212921/681487f9-79d1-43ae-b8a0-17a0d20f0e55)

**Steps to reproduce:**

1. Open python console with pygame installed
2. Import and init pygame
3. Run the following command: `pygame.font.SysFont("Arial", 20).render("test", True, (255, 255, 255), (255, 255, 255)).copy()`

**Test code**

If possible add a simple test program that shows the problem described in this report.
```
import pygame
pygame.init()
image = pygame.font.SysFont("Arial", 20).render("test", True, (255, 255, 255), (255, 255, 255))

image.copy() # Throws pygame.error: Empty destination palette
```

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
pygame.error: Empty destination palette
```


Wanted to note that obviously there's an element of user error here, writing white text on a white background isn't a great way to get yourself an empty white surface, but the presentation of the error being removed from the root cause makes it very difficult to troubleshoot as a user (probably took me about 90 minutes).
In an instance (like what happened to me) where the values are incorrectly assigned, but the surface is not used immediately, an error at the render call is all it would take to save a lot of troubleshooting. That said I can also image situations where the colors here could be parameterized (for example from user input) in which case having an error occur in edge cases would be especially undesirable.

nkSunUow:
I'd like to take the ticket.

nkSunUow:
[...]

PR #4057, opened 2023-10-21
Add comment to surf_copy

nkSunUow:
Issue Addressed
#4015 

PR Summary

add a comment to surf_copy so such constraints on surface in order to copy can be more visible.


illume (approved):
👍

PR #4071, opened 2023-11-02
Update README.rst

Thunderborne:
Capitalized some proper names, e.g. "Pygame," "Python," and "GitHub" for consistency

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,14 +12,14 @@ It uses the `Simple DirectMedia Layer library`_ and several other
 popular libraries to abstract the most common functions, making writing
 these programs a more intuitive task.
 
-`We need your help`_ to make Pygame the best it can be!
+`We need your help`_ to make pygame the best it can be!
 New contributors are welcome.
 
 
 Installation
 ------------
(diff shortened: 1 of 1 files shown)


illume:
hi. I keep pygame as all lower because the package is lower case, and it's an error to `Import PyGame`.

Thunderborne:
> hi. I keep pygame as all lower because the package is lower case, and it's an error to `Import PyGame`.

Oh I see. Thank you very much for the feedback!

illume:
> Oh I see. Thank you very much for the feedback!

Thank you for your understanding. Do you want to update it with the other changes (GitHub and Python)?

Thunderborne:
> Thank you for your understanding. Do you want to update it with the other changes (GitHub and Python)?

Definitely. Thank you. This is my first time trying to contribute to a FOSS project so thank you for the review. I went ahead and changed back "Pygame" to "pygame" in the README.rst.

illume (approved):
🎉🎈 great, thank you

PR #4084, opened 2023-11-07
mixer_test: mixer_music_test: Add test for sound_from_array and rewind

romgird2:
Hello,

We are a group of software engineering students learning to work on open source projects, and as such, we wish to contribute to PyGame.

Thus, we created a unit test for creating sound from an array.

Thank you for your consideration

### DIFF
--- a/test/mixer_test.py
+++ b/test/mixer_test.py
@@ -4,7 +4,7 @@ import unittest
 import pathlib
 import platform
 import time
-import numpy
+from array import array
 
 from pygame.tests.test_utils import example_path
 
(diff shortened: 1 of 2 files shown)


test/mixer_test.py
```diff
@@ -1059,9 +1060,11 @@ def todo_test_sound__from_buffer(self):
         """Ensure Sound() creation with a buffer works."""
         self.fail()
 
-    def todo_test_sound__from_array(self):
+    def test_sound__from_array(self):
         """Ensure Sound() creation with an array works."""
-        self.fail()
+        array1 = sndarray.array(example_path(os.path.join("data", "house_lo.wav")))
```
illume, line 1065: I wonder if this should be using pythons array.array rather than the pygame.sndarray? This way it would not depend on sndarray, but only what is in the python standard library.

qrabilloud:
Hello there,

I'm from the group of software engineering students. I added another test in the pull request regarding the rewind() function. Tell us if you think it's correct so that we can merge after the review.

Thanks for your help, best regards.

ricky074game (approved):

illume (approved):
🎉🎈

Thanks.

PR #4087, opened 2023-11-09
Remove duplicated block of code from documentation

douglasdcm:
Hi, how are you? I was reading the tutorials recently and I noticed that this one has a lot of duplicated code that is not related to the main purpose of the tutorial (maybe a copy and paste issue). I'm opening this Pull Request to remove these blocks of code, so that the reading is smoother and clear especially in mobile phones or other devices with a small screen

I didn't open a issue to it as it is a simple change, despite of its size. Let me know if an issue is necessary, so I can reference it here.

I tested all the pages using `python setup.py docs` and all of them looks fine. Thanks a lot!

image


illume:
Hi. Good find. Seems like a good idea to me.

illume (approved):
👍

PR #4095, opened 2023-11-24
Remove references to Numeric in the documentation

rpep:
Numeric was the predecessor of NumPy, but NumPy has been established for a long time, and so references to Numeric are now quite dated. This pull request removes references to Numeric, since nowadays they're just likely to confuse people.

There is one remaining reference in `examples/sound_array_demos.py`, but in this case, it is part of the changelog. The code it talks about (looking for Numeric instead of NumPy) was removed in 2019 so it may make more sense just to remove the changelog (which hasn't been kept updated), but for now I have left this as-is.


illume (approved):
Thanks 🎈🎉

PR #4096, opened 2023-11-28
Adding test_queue in mixer tests

Phegerr:
Hello,

We have added "test_queue" in the mixer test file. Additionally, we have removed a TODO in the "test_getQueue" because the functionality it was referring to has been implemented in the meantime.

We were also wondering if you have any recommendations for issues we can work on, as most of the listed issues seem to be already in progress, and we want to avoid duplicating efforts.

Thank you for your consideration.


illume (approved):
🎉 🎈 

Thanks.

PR #4102, opened 2023-12-04
docs: DisplayModes: Fix description of width, and height default with set_mode and document display

chpauline:
The size (width and height) is no longer a required argument for pygame.display.set_mode(): https://www.pygame.org/docs/ref/display.html#pygame.display.set_mode.


illume (approved):
🎉🎈

Thanks.

PR #4105, opened 2023-12-05
Refactored code to make file easier to read, understand, and maintain

JacobOSU:
Split cursor class into subclasses
Changed code to use subclasses
Removed deletion of methods
Changed variable names to use snakecase
Changed variable names to be more descriptive
Refactored load_xbm function to rid of duplicate code

### DIFF
--- a/src_py/cursors.py
+++ b/src_py/cursors.py
@@ -28,11 +28,11 @@ Here is a list of available cursors:
     arrow, diamond, ball, broken_x, tri_left, tri_right
 
 There is also a sample string cursor named 'thickarrow_strings'.
-The compile() function can convert these string cursors into cursor byte data that can be used to
-create Cursor objects.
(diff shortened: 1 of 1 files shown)


Shidoengie (changes requested):
The subclass split was rather unecesserary and imo bad, everything else is alright

illume:
Hello. Thanks for this, it's a good idea.

Looks like there are some test failures though.

Do you want to continue this? If so, it would be best to make a change and commit to smaller parts. Perhaps only change and commit one function at a time and push. That way the tests should run each time.

We now use the pre-commit tool for running code formatters/linters/etc.

```
python3 -m pip install pre-commit black ruff pylint
pre-commit run -a
```

To run the tests for cursor...
```
python3 -m pip install .
python3 test/cursors_test.py
```

PR #4106, opened 2023-12-06
Add kwargs to surface.map_rgb and surface.unmap_rgb

Cdog1526:
Continuing work on #808 by adding keyword arguments to Surface.map_rgb and Surface.unmap_rgb. Updated the unit tests and passed all of those, and updated documentation as well. Follows same standards as previous commits for this issue.


illume (approved):
👍 

thanks

PR #4112, opened 2023-12-09
simple test case for camera

Alihyp:
I added a simple test case for camera just to check if it works well

### DIFF
--- a/test/camera_test.py
+++ b/test/camera_test.py
@@ -1,5 +1,31 @@
 import unittest
+import pygame
+import pygame.camera
 
 
 class CameraModuleTest(unittest.TestCase):
     pass
+
+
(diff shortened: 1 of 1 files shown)


test/camera_test.py
```diff
@@ -1,5 +1,31 @@
...
+
+
+class TestPygameCamera(unittest.TestCase):
+    def setUp(self):
+        pygame.init()
+        pygame.camera.init()
+
+    def test_camera(self):
+        cameras = pygame.camera.list_cameras()
+        self.assertTrue(len(cameras) > 0, "No cameras found")
```
illume, line 17: If there's no cameras, it's probably better the test is skipped instead.

This way the test should work on more computers too.

PR #4114, opened 2023-12-10
Surface: Add key word arguments to .set_at()

RarachelLuo:
Continuing work on https://github.com/pygame/pygame/issues/808 by adding keyword arguments to Surface.set_at(). Updated the unit tests, and updated documentation as well. Follows same standards as previous commits for this issue.


illume (approved):
🎉 🎈 

Thanks.

PR #4117, opened 2023-12-12
updated documentation for fix of rect clipline

nobodyg0n3:


illume:
Thanks for this. But I don't really see how this is useful information in the example here.

PR #4118, opened 2023-12-12
Grammatical changes in README files

justin-ngo-jn:
Identified and fixed grammatical and spelling issues that I found in the README files


purpleysound (approved):

robotsnh (commented):
@justin-ngo-jn I think that this works, but do you think that it would be better to merge lines 5 & 6?

Issue #4100, opened 2023-12-01
Generic `Group`


611h:
**Description**

The return type of `Group.sprites()` is List<Unknown>. While iterating through `Group`, we cannot get help from the IDE since the type of the element is `Unknown`.

How about adding generics to `Group` class so that we can annotate any custom sprite types to the `Group`.

```
class Group(Generic[T]):
    ...
```

PR #4119, opened 2023-12-12
Added generics to abstract group for IDE completion

Lukasgrat:
Made in response to issue #4100


illume (approved):
👍 thanks.

So with this you could do something like:
```python
alien_group = Group[AlienSprite]()
```

Issue #4086, opened 2023-11-08
Build fails "cc1: fatal error: src_c/pypm.c: No such file or directory"


jgf-osu:
**Environment:**

- Linux Mint 21 (Cinnamon)
- Python 3.10.12
- PyGame: just forked and cloned main branch of git repo

**Report:**

I followed [the procedure under PYGAME 2 here](https://www.pygame.org/wiki/CompileUbuntu). No problems with apt installing the dependencies. The command `setup.py -config -auto` produced this, which looks OK to me:

```
Skipping Cython compilation


WARNING, No "Setup" File Exists, Running "buildconfig/config.py"
Using UNIX configuration...


Hunting dependencies...
SDL     : found 2.0.20
FONT    : found
IMAGE   : found
MIXER   : found
PNG     : found
JPEG    : found
SCRAP   : found
PORTMIDI: found
PORTTIME: found
FREETYPE: found 24.1.18

If you get compiler errors during install, double-check
the compiler flags in the "Setup" file.

```

However, `setup.py install --user` fails:

```
Skipping Cython compilation
running install
/usr/lib/python3/dist-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
/usr/lib/python3/dist-packages/setuptools/command/easy_install.py:158: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
/usr/lib/python3/dist-packages/pkg_resources/__init__.py:116: PkgResourcesDeprecationWarning: -VERSION- is an invalid version and will not be supported in a future release
  warnings.warn(
running bdist_egg
running egg_info
creating pygame.egg-info
writing pygame.egg-info/PKG-INFO
writing dependency_links to pygame.egg-info/dependency_links.txt
writing entry points to pygame.egg-info/entry_points.txt
writing top-level names to pygame.egg-info/top_level.txt
writing manifest file 'pygame.egg-info/SOURCES.txt'
package init file 'docs/__init__.py' not found (or not a regular file)
reading manifest file 'pygame.egg-info/SOURCES.txt'
writing manifest file 'pygame.egg-info/SOURCES.txt'
installing library code to build/bdist.linux-x86_64/egg
running install_lib
running build_py
creating build
creating build/lib.linux-x86_64-3.10
creating build/lib.linux-x86_64-3.10/pygame
[...]

PR #4126, opened 2023-12-13
add BUILD.rst to provide build instructions

jgf-osu:
This pull request includes two changes:

1. Add a new file named **BUILD.rst** based on the [Compilation page](https://www.pygame.org/wiki/Compilation) of the Wiki.  By moving this information into the git repo, the community can more easily assist in maintaining its accuracy. In this initial commit, the BUILD.rst file includes an update to the Ubuntu build guide to close [issue number 4086](https://github.com/pygame/pygame/issues/4086) and a completely reworked Fedora build guide tested in Fedora 39. The Windows and MacOS build guides are direct copies of the current guides on the wiki. The PyPy section is taken directly from the wiki.

2. Update README.rst to link to BUILD.rst where it currently links to the Compilation page of the Wiki.

### DIFF
--- a/BUILD.rst
+++ b/BUILD.rst
@@ -1,0 +1,225 @@
+.. image:: https://raw.githubusercontent.com/pygame/pygame/main/docs/reST/_static/pygame_logo.svg
+  :alt: pygame
+  :target: https://www.pygame.org/
+
+
+|AppVeyorBuild| |PyPiVersion| |PyPiLicense|
+|Python3| |GithubCommits| |BlackFormatBadge|
+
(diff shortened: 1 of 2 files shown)


illume:
Thanks for this.  Much appreciated, this will be very helpful for folks.

I’ve put the fedora section in the wiki for now…
https://www.pygame.org/wiki/CompilationFedora also linked from other pages.

We should also adjust the build config to show a link there if the build fails on fedora.

The Ubuntu change still needs to be done, and the bug in #4096 still needs to be fixed.

Probably I’d prefer a compilation guide in docs/ and I know someone else already started writing one for there. Will give them a nudge to see where they’re up to.

PR #4143, opened 2024-01-06
platform.linux_distribution() deprecated python >= 3.7

dslackw:
Since python 3.7, platform.linux_distribution() deprecated, so I suggest distro package instead.

As I am maintainer for the Slackware distribution, I added the Slackware in the list, if you don't mind.

Thanks

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,6 +7,7 @@
 #     python setup.py install
 
 import glob
+import distro
 import platform
 import sysconfig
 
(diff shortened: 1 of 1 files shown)


illume:
Nice one. Thank you.

We have some failures because the `distro` module is not there. Can this be made to be an optional dependency? Where you check if it can import and only use it then?

dslackw:
> Nice one. Thank you.
> 
> We have some failures because the `distro` module is not there. Can this be made to be an optional dependency? Where you check if it can import and only use it then?

Yes, but `compilation_help() ` it's not working anyway. I believe more distributions use python >= 3.9.

illume:
This was continued / merged with the fallback if distro isn't there in here: https://github.com/pygame/pygame/pull/4205

PR #4162, opened 2024-02-21
Add additional argument seconds argument for clock.tick

krobinson395:
### Issue Addressed
#3976 

### PR Summary
The goal of this PR was to add the additional functionality requested in #3976, for an additional optional argument to be added to the clock.tick method to allow for the user to request the return in seconds rather than milliseconds as is the current default behavior. I have added this additional optional argument, it should not affect any existing code as the default behavior is left unchanged but rather should only add additional options if needed. I have also added an additional set of tests, modelling the original clock.tick tests, to test the behavior when the seconds argument is enabled.

### DIFF
--- a/src_c/time.c
+++ b/src_c/time.c
@@ -413,15 +413,10 @@ clock_tick_base(PyObject *self, PyObject *arg, int use_accurate_delay)
     PyClockObject *_clock = (PyClockObject *)self;
     float framerate = 0.0f;
     int nowtime;
-    int return_in_seconds = 0;
-    PyObject *seconds = NULL;
+    int seconds = 0;
 
-    if (!PyArg_ParseTuple(arg, "|f0", &framerate, &seconds))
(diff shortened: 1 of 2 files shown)


illume:
@krobinson395 catching up on my long backlog... apologies I didn't notice this until now.

illume (changes requested):
The appveyor windows builds found some issues. Looking at the logs when you click through the actions you can see these.

```c
src_c/time.c(512): error C2220: warning treated as error - no 'object' file generated
src_c/time.c(512): warning C4113: 'PyObject *(__cdecl *)(PyObject *,PyObject *,PyObject *)' differs in parameter lists from 'PyCFunction'
src_c/time.c(516): warning C4113: 'PyObject *(__cdecl *)(PyObject *,PyObject *,PyObject *)' differs in parameter lists from 'PyCFunction'
```

Let me know if you want to continue the change?

PR #4166, opened 2024-02-29
Denested Code

Shidoengie:
Denested code blocks in various files in order to improve code readability


illume:
Hello. Thanks for this, it's a good idea.

Looks like there are some test failures though.

Do you want to continue this? If so, it would be best to make a change and commit to smaller parts. Perhaps only change and commit one function at a time and push. That way the tests should run each time.

We now use the pre-commit tool for running code formatters/linters/etc.

```
python3 -m pip install pre-commit black ruff pylint
pre-commit run -a
```

PR #4178, opened 2024-03-22
Keep GitHub Actions up to date with GitHub's Dependabot

cclauss:
Fixes the Node.js 16 warning at the bottom right of https://github.com/pygame/pygame/actions/runs/8384244178

Consolidates all GitHub Actions upgrades so that there can only be a single PR at a time.

* https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot
* https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem


illume:
Thanks for this.

Having separate pull requests makes it easier to see if one of them caused something wrong. Would it be possible to move it away from a single pull request?

illume (approved):
🎉 thanks

PR #4187, opened 2024-03-30
build-ubuntu-sdist.yml: apt-update but not upgrade

cclauss:
This is a useful change because it allows the tests to continue further.  I Do not know have the failing tests in a later step.
* Before: https://github.com/pygame/pygame/actions/runs/8490728204/job/23262095318
* After: https://github.com/pygame/pygame/actions/runs/8490938867/job/23262459187


illume (approved):
Great! Thanks a lot.

Adding a retry here has been in the back of my mind for some time.

PR #4190, opened 2024-03-31
buildconfig/macdependencies glib: Fix build

illume:
- This should supply distutils which new versions of python broke.
- Newer glib broke their configure / make style build.
- The new glib build system encounters a meson bug which doesn't overwrite dangling symlinks. So have to make sure we remove all of those files first.
- the mac build didn't use mypy / pytype / pyright, and at least one of them was failing to install on some versions of python


illume:
@cclauss would you mind having a look at this?

illume:
Oh, I broke it again. One of those cleanup scripts I commented out are needed after all.

buildconfig/manylinux-build/docker_base/glib/glib.sha512
```diff
@@ -1 +1 @@
-280a46c2af13283a08c15ff0b4f5492659c2884521930600ad45310ed181c44a878ad8f9b36bae68ed6e7d92db6f1630f7bf015148c513dc317d25807f13abb0  glib-2.56.4.tar.xz
+1514d62aeb4c4a1a1048ae0f84f7db7f0dbf355772b2dadf6a34ec547045b163a5e28331b096e7616fe3c9c19bed98025a0202b05073f5d7ee901d0efaffe143  glib-2.80.0.tar.xz
```
cclauss, line 1:
```suggestion
1514d62aeb4c4a1a1048ae0f84f7db7f0dbf355772b2dadf6a34ec547045b163a5e28331b096e7616fe3c9c19bed98025a0202b05073f5d7ee901d0efaffe143  glib-2.80.0.tar.xz

```

cclauss: GitHub puts a 🚫 on files that do not end with a single blank line.

cclauss (approved):

buildconfig/manylinux-build/docker_base/glib/glib.sha512
illume: Thanks. That's fixed now.

PR #4201, opened 2024-04-07
Remove lint logic from setup.py

cclauss:
~**After** #4200 is merged~ remove lint logic from setup.py
* #4200

To replace the commands that were removed from `setup.py`, the following will work...
* `pre-commit run black`
* `pre-commit run clang-format`
* `pre-commit run --all-files`

### DIFF
--- a/.github/workflows/format-lint.yml
+++ b/.github/workflows/format-lint.yml
@@ -28,6 +28,8 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3.0.2
+      - name: Install deps
+        run: python3 -m pip install sphinx
       - name: Check docs changes are checked in
         run: |
           python3 setup.py docs
--- a/.github/workflows/format-lint.yml
+++ b/.github/workflows/format-lint.yml
@@ -28,6 +28,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3.0.2
-
-      # This job will be completly be replaced by the pre-commit job in #4200
-
+      - name: Check docs changes are checked in
+        run: |
+          python3 setup.py docs
+          if [[ `git status --porcelain` ]]; then
+            echo "Generating docs caused changes. Please check them in."
+            echo "You may need to run: python3 setup.py docs --fullgeneration"
+            # Run git status again, so people can see what changed.
+            git status --porcelain
+            exit 1
+          fi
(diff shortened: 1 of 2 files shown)


.github/workflows/format-lint.yml
```diff
@@ -29,19 +29,5 @@ jobs:
...
       - uses: actions/checkout@v3.0.2
 
-      - name: Install deps
-        run: python3 -m pip install pylint==2.15.5 black clang-format sphinx
+      # This job will be completly be replaced by the pre-commit job in #4200
 
-      - name: Check code Formatting and Linting
-        run: python3 setup.py lint
-
-      - name: Check docs changes are checked in
```
illume, line 38: This docs check will need to remain.

illume (commented):
Hi, Thanks for this.

I think we still need a place to put commands that can be run during development. But I'm not sure if there even is a new place in whatever the latest python packaging way is to do this? In old Makefile projects this might be in `make check`. npm projects have a package.json scripts. Rust projects have task runners like cargo make. Seems tasks aren't standardised in python yet... except in distutils/setuptools commands.

I left a note about the docs check in CI probably still needing to be there. I guess that will move to precommit too eventually?

cclauss:
> we still need a place to put commands that can be run during development

The following will work...
* `pre-commit run black`
* `pre-commit run clang-format`

.github/workflows/format-lint.yml
cclauss: Done.

setup.py
```diff
@@ -305,7 +305,7 @@ def consume_arg(name):
     if cython_only:
         sys.exit(0)
 
-no_compilation = any(x in ['lint', 'format', 'docs'] for x in sys.argv)
+no_compilation = any(x in ('docs') for x in sys.argv)
```
cclauss, line 308:
```suggestion
no_compilation = 'docs' in sys.argv
```

.github/workflows/format-lint.yml
illume: thanks

illume (approved):
👍 thanks

illume:
Not related to this PR exactly...

I noticed the pre-commit that comes with Ubuntu 22.04 is a bit too old for our config. There also aren't any install instructions for pre-commit with Ubuntu and other OS that work for me on the install page https://pre-commit.com/#install 
(I installed with pip into my pyenv bin folder, but I guess we will need better pre-commit install instructions for folks on windows/Ubuntu)

PR #4209, opened 2024-04-14
Minor optimizations by using comprehensions

cclauss:
% `pipx install ruff`
% `ruff check --select=C4,PERF --statistics`
```
10	C406   	[*] Unnecessary `tuple` literal (rewrite as a `dict` literal)
10	C417   	[*] Unnecessary `map` usage (rewrite using a generator expression)
 9	C408   	[*] Unnecessary `dict` call (rewrite as a literal)
 9	PERF401	[ ] Use a list comprehension to create a transformed list
 2	C419   	[*] Unnecessary list comprehension
 2	PERF203	[ ] `try`-`except` within a loop incurs performance overhead
 2	PERF402	[ ] Use `list` or `list.copy` to create a copy of a list
```
% `ruff check --select=C4,PERF --fix --unsafe-fixes`
```
Found 44 errors (31 fixed, 13 remaining).
```
% `ruff rule C408`
# unnecessary-collection-call (C408)

Derived from the **flake8-comprehensions** linter.

Fix is always available.

## What it does
Checks for unnecessary `dict`, `list` or `tuple` calls that can be
rewritten as empty literals.

## Why is this bad?
It's unnecessary to call, e.g., `dict()` as opposed to using an empty
literal (`{}`). The former is slower because the name `dict` must be
looked up in the global scope in case it has been rebound.

## Examples
```python
dict()
dict(a=1, b=2)
list()
tuple()
```

Use instead:
```python
{}
{"a": 1, "b": 2}
[]
()
```

## Fix safety
This rule's fix is marked as unsafe, as it may occasionally drop comments
when rewriting the call. In most cases, though, comments will be preserved.

## Options
- `lint.flake8-comprehensions.allow-dict-calls-with-keyword-arguments`


illume (approved):
🎉🎈

Thanks.

PR #4211, opened 2024-04-17
buildconfig module cleanup

henserobbie:
**Description**
With the deprecation of `distutils` in Python 3.10, my senior design team thought it would be helpful to bring pygame's build system up to the PEP 517 standards. Additionally, with a push to clean up pygame, we are looking to remove unused files. This PR is focused on the `buildconfig` module. In this pull request, we replaced usages of `distutils` and removed unused python scripts.

**Changes**
The following files use `distutils.sysconfig.get_python_inc()` to fetch the python include directory:
`buildconfig/config_darwin.py`
`buildconfig/config_msys2.py`
`buildconfig/config_unix.py`
`buildconfig/config_win.py`
The python include directories are now obtained through
`sysconfig.get_path('include')`.

The following file is no longer used, so it has been deleted.
`buildconfig/config_emskd.py`

**Tests**
The testing of the proposed changes will be done automatically through Appveyor. The buildconfig module generates a setup makefile containing the c modules, dependency libraries, and data files to be used by the setup.py script to build pygame. If the Appveyor targets build successfully and pass all tests, the changes have successfully maintained functionality while removing distutils usage.

PR #4212, opened 2024-04-17
Compliance with PEP517 and PEP632

henserobbie:
**Description**
This PR is an extension of PR #4211 with the goal of continuing to remove deprecated libraries, simplifying the build process and bringing the Pygame build configuration into compliance with PEP-517 and PEP-632. This PR is focused on the refractor of the `setup.py` file, the introduction of `pyproject.toml` and the removal of `disutils` from `setup.py`. The end goal is a more simplified and intuitive build system that uses `pip install` as the primary build command rather than running `setup.py install`.

**Changes**
* `pyproject.toml`: This was added and metadata information and build dependencies that were previously defined in `setup.py` have been moved here.
    * The addition of the `pyproject.toml` file brings the build structure in line with PEP-517 allowing for build dependencies and the `setuptools` backend to be specified within.
* `setup.cfg`: Added duplicate metadata to allow builds on Python 3.6.
    * Prior to Python 3.7, `pip` looked for metadata within the `setup.cfg` rather than the `pyproject.toml`. Duplicating the metadata was a simple fix to ensure continued support for Python 3.6.
* `setup.py`: Refactored for organization, `distutils` was removed and replaced with `setuptools`, and metadata was removed.
    * The overall refactoring intent was to allow future maintainers to more easily make changes to smaller parts of the build system without having to parse through the entire process. This should hopefully help future-proof the process by allowing for deprecated dependencies to more easily be swapped out.
    * `disutils` was removed in most places in compliance with PEP-632, usages of `distutils` have been replaced by the appropriate counterparts in `setuptools`. If no appropriate counterpart was found a custom implementation of the utility was created.
    * Metadata removed from the `setup.py` file in favor of being moved to `pyproject.toml` and `setup.cfg` as described above.
* `buildconfig/appveyor.yml`: Appveyor build changed to use `pip` instead of calling `setup.py` directly.
    * mscv analyze option is now specified via environment variable, `USE_ANALYZE`, which gets detected by the `setup.py` script similar to `PYGAME_DETECT_AVX2`. This is because you cannot pass arguments to the `setup.py` when building with `pip`.

**Tests**
These changes will be tested through Appveyor to ensure that the build process works similar to PR #4211. If the Appveyor targets record successful builds with all tests passing then that will be our indication that the updated build system functions as intended and works within the legacy code base.

### DIFF
--- a/buildconfig/appveyor.yml
+++ b/buildconfig/appveyor.yml
@@ -168,8 +168,7 @@ install:
   - "%PYTHON% -m pip install Sphinx==4.5.0 sphinxcontrib-applehelp==1.0.2 pip install sphinxcontrib-devhelp==1.0.2 sphinxcontrib-htmlhelp==2.0.0 sphinxcontrib-qthelp==1.0.3 sphinxcontrib-serializinghtml==1.1.5"
   - "%PYTHON% setup.py docs"
   - "%WITH_COMPILER% %PYTHON% -m buildconfig"
-  - "%WITH_COMPILER% %PYTHON% setup.py build -j4 %USE_ANALYZE%"
-  - "%WITH_COMPILER% %PYTHON% setup.py -setuptools %DISTRIBUTIONS%"
+  - "%WITH_COMPILER% %PYTHON% -m pip wheel . --wheel-dir=dist"
   - ps: "ls dist"
 
   # Install the wheel to test it
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,51 @@
+[project]
+name = "pygame"
+authors = [{ name = "A community project.", email = "pygame@pygame.org" }]
+description = "Python Game Development"
+readme = "README.rst"  # Assuming long_description is in README.rst; adjust filename as needed
+license = { text = "LGPL" }
+classifiers = [
+    "Development Status :: 6 - Mature",
+    "License :: OSI Approved :: GNU Library or Lesser General Public License (LGPL)",
+    "Programming Language :: Assembly",
+    "Programming Language :: C",
+    "Programming Language :: Cython",
+    "Programming Language :: Objective C",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3.6",
+    "Programming Language :: Python :: 3.7",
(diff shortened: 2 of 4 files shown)


illume (changes requested):
Hi,

thanks for this cleanup work. Nice improvements.

I have a proposal below for merging this in pieces... What do you think?

These two commits could be merged in already if they were in a separate PR:

- [build: setup.cfg: added metadata](https://github.com/pygame/pygame/pull/4212/commits/8f1e5cec959669b2adc92765aa85c23a545e2c2c)
- [buildconfig: appveyor: changed to use pip wheel](https://github.com/pygame/pygame/pull/4212/commits/764fad619d0aef7e2b6a644cab4b64354e2b1027)

For the reorder commit it would be good if it could be smaller atomic commits. So it's easier to see what is changing, and also so that we can bisect more easily if there is a bug and we need to track it down later.

The pyproject.toml one is breaking things, so I think that one would need more work. Best keep it aside for now and let's get the other pieces merged in?

buildconfig/appveyor.yml
```diff
@@ -168,8 +168,7 @@ install:
   - "%PYTHON% -m pip install Sphinx==4.5.0 sphinxcontrib-applehelp==1.0.2 pip install sphinxcontrib-devhelp==1.0.2 sphinxcontrib-htmlhelp==2.0.0 sphinxcontrib-qthelp==1.0.3 sphinxcontrib-serializinghtml==1.1.5"
   - "%PYTHON% setup.py docs"
   - "%WITH_COMPILER% %PYTHON% -m buildconfig"
-  - "%WITH_COMPILER% %PYTHON% setup.py build -j4 %USE_ANALYZE%"
-  - "%WITH_COMPILER% %PYTHON% setup.py -setuptools %DISTRIBUTIONS%"
+  - "%WITH_COMPILER% %PYTHON% -m pip wheel . --wheel-dir=dist"
```
illume, line 171: It would be better if the "-j4" part was passed in. However, it's not the end of the world, because it doesn't seem to help as much as it used to help. Now it only saves 2-4 minutes total off the build time.

illume:
Going to close this to tidy up the PR list.

If you want to continue, please feel free to submit smaller PRs as mentioned in my previous comment.

PR #4261, opened 2024-06-11
updating type hinting for SysFont

JovialKnoll:
Howdy,

As far as I can tell, it's valid to pass in `None` as the name argument when using SysFont, but the type hinting doesn't reflect this. I think this change would fix the type hinting, but let me know if I'm missing something.

Thanks!


illume (approved):
You're right.

Thanks 👍

Issue #4099, opened 2023-11-29
Update to new C API for Python 3.13 


limburgher:
src_c/math.c: In function ‘com_init’:
src_c/math.c:4425:10: error: implicit declaration of function ‘_PyArg_NoKeywords’ [-Werror=implicit-function-declaration]
 4425 |     if (!_PyArg_NoKeywords("ClassObjectMethod", kwds))
      |          ^~~~~~~~~~~~~~~~~
cc1: some warnings being treated as errors

https://bugzilla.redhat.com/show_bug.cgi?id=2250670

limburgher:
Any updates?

hroncok:
This is the function:

https://github.com/python/cpython/blob/v3.13.0b2/Python/getargs.c#L2839-L861

```c
/* For type constructors that don't take keyword args
 *
 * Sets a TypeError and returns 0 if the args/kwargs is
 * not empty, returns 1 otherwise
 */
int
_PyArg_NoKeywords(const char *funcname, PyObject *kwargs)
{
    if (kwargs == NULL) {
        return 1;
    }
    if (!PyDict_CheckExact(kwargs)) {
        PyErr_BadInternalCall();
        return 0;
    }
    if (PyDict_GET_SIZE(kwargs) == 0) {
        return 1;
    }

    PyErr_Format(PyExc_TypeError, "%.200s() takes no keyword arguments",
                    funcname);
    return 0;
}
```

And this is the usage:

https://github.com/pygame/pygame/blob/00f0c5793ddf22e6e3990d60e4e9f8a31f0258a0/src_c/math.c#L4425-L4426

In worst case scenario, the function can be inlined.

PR #4269, opened 2024-06-18
Inline _PyArg_NoKeywords from Python 3.13 in case it is missing

hroncok:
Python 3.13 made this private:
https://github.com/python/cpython/pull/110966

Fixes https://github.com/pygame/pygame/issues/4099

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -4413,6 +4413,34 @@ com_descr_get(PyObject *self, PyObject *obj, PyObject *type)
     return PyMethod_New(com->obj_callable, obj);
 }
 
+#ifndef _PyArg_NoKeywords
+/* For type constructors that don't take keyword args
+ *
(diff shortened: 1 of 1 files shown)


illume:
Hi @hroncok,

thanks for this.

For fixing the formatting...
```
python3 -m pip install pre-commit
pre-commit install
pre-commit run -a
```

The python 3.6 builds are failing. I guess a version guard would be suitable.
```C
/* python > 3.12 */
#if PY_VERSION_HEX > 0x030C0000
#ifndef _PyArg_NoKeywords

#endif
#endif
```

hroncok:
I won't be able to dedicate more time to this, sorry. Feel free to amend any changes necessary.

illume (approved):
👍

illume:
Cool. Thanks again. I'll finish it off.

PR #4270, opened 2024-06-19
Python the 13th

illume:
Continued on from https://github.com/pygame/pygame/pull/4269 by @hroncok

- **Inline _PyArg_NoKeywords from Python 3.13 in case it is missing**
- **math: Fix _PyArg_NoKeywords ifdef for python 3.13**

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -4413,6 +4413,8 @@ com_descr_get(PyObject *self, PyObject *obj, PyObject *type)
     return PyMethod_New(com->obj_callable, obj);
 }
 
+/* python > 3.12 */
+#if PY_VERSION_HEX > 0x030C0000
 #ifndef _PyArg_NoKeywords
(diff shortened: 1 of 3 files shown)


src_c/math.c
```diff
@@ -4413,6 +4413,37 @@ com_descr_get(PyObject *self, PyObject *obj, PyObject *type)
     return PyMethod_New(com->obj_callable, obj);
 }
 
+/* python > 3.12 */
+#if PY_VERSION_HEX > 0x030C0000
+#ifndef _PyArg_NoKeywords
```
hroncok, line 4418: I guess this `#ifndef` is no longer necessary. The idea was that it would prevent the code from existing on Python < 3.13, but apparently, it did not work properly everywhere.

illume: Yeah. It can't hurt I don't think.

PR #4273, opened 2024-06-25
2.6.0 -> 2.6.1.dev1

illume:


AUeno2025:
Thank you very much for the information on the version of pygame. I tried to use pygame 2.6.1 with Python 3.14.0 on macOS (Tahoe 26.0.1.). Although I had to install several tools and libraries (XCode, wheel etc..), pygame 2.6.1 worked very well with Python 3.14.0. Thank you very much again.

PR #4320, opened 2024-08-29
Fix scale by rounding issue

chitvs:
## Description

This pull request addresses a rounding issue in the `pygame.transform.scale_by` function. Previously, the function would truncate scaling results to integers, leading to inaccuracies in the dimensions of scaled surfaces. This fix uses `round` instead of casting to `int` to ensure that dimensions are correctly calculated.

### Changes
- **Updated** `transform.c` to use `round` for scaling calculations.
- **Updated** `transform_test.py` to include tests that verify the correct behavior of the updated scaling function.

### Testing
- Ran all existing tests to ensure no regressions.
- Added new tests to validate the fix for the rounding issue.

Please review the changes and let me know if there are any further modifications needed.


illume:
Hello. Is there something I can help you with?

PR #4358, opened 2024-10-07
Updating transform.c to use round on transform.scale_by

### LINKED ISSUE #4294: transform.scale_by rounding issue
**Environment:**

- **Operating system**: Linux(Ubuntu)
- **Python version** : 3.12.2
- **SDL version**: 2.28.5+dfsg-2+22.04+jammy+release+build2
- **PyGame version**: 2.6.0
- **Relevant hardware** : not relevant

**Current behavior:**

I have a picture that is 1468 pixels wide. I want to resize it to be 240 wide. I wanted to use `scale_by` so that the vertical dimension is automatically scaled by the same factor. What I get is a factor of 0.16348773841961853, which, when fed into `scale_by`, results in a 239 pixel wide `Surface`.

**Expected behavior:**

Manually multiplying `0.16348773841961853 * 1468` in Python is `240.0`, so I would expect the final `Surface` to have the same width (i.e. 240).

**Test code**

```Python
import pygame
pygame.transform.scale_by(pygame.Surface((1468, 200)), 240 / 1468)
# Returns <Surface(239x32x32 SW)>
```

**Possible solution**

[Line 630 in transform.c](https://github.com/pygame/pygame/blob/8d929587ec246e214408763e778ec2a1813c9ff7/src_c/transform.c#L630) should probably read `round(surf->w * scalex)` instead of `(int)(surf->w * scalex)`.

devfernandoa:
Closes #4294 
I used the proposed fix on the pull request and it seemed to work just fine

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -627,8 +627,8 @@ surf_scale_by(PyObject *self, PyObject *args, PyObject *kwargs)
     surf = pgSurface_AsSurface(surfobj);
     if (!surf)
         return RAISE(pgExc_SDLError, "display Surface quit");
-    newsurf = scale_to(surfobj, surfobj2, (int)(surf->w * scalex),
-                       (int)(surf->h * scaley));
+    newsurf = scale_to(surfobj, surfobj2, round(surf->w * scalex),
+                       round(surf->h * scaley));
     if (!newsurf) {
         return NULL;
     }


devfernandoa:
not sure why the app build is failing, anyone know why?

src_c/transform.c
```diff
@@ -627,8 +627,8 @@ surf_scale_by(PyObject *self, PyObject *args, PyObject *kwargs)
     surf = pgSurface_AsSurface(surfobj);
     if (!surf)
         return RAISE(pgExc_SDLError, "display Surface quit");
-    newsurf = scale_to(surfobj, surfobj2, (int)(surf->w * scalex),
-                       (int)(surf->h * scaley));
+    newsurf = scale_to(surfobj, surfobj2, round(surf->w * scalex),
```
tgandor, line 630: Hey, maybe the build is failing, because this is C and not Python?
This should be `(int)(surf->w * scalex + 0.5)` and same for `scaley` below. This will do rounding instead of truncating.

tgandor: Or maybe not? I hear C99 has `round()`. Not sure how it is across compilers (like on MacOS...)

PR #4424, opened 2025-01-03
add explanations for less-clear aspects of joystick example

jacob-thompson:
This commit adds two comments to the joystick module example source code, and fixes a typo.

I feel that this change is appropriate since this example tends to explain methods that are rather intuitive:

https://github.com/pygame/pygame/blob/79807da84c9bacf8df5a177763e14c924e3b15e2/examples/joystick.py#L84-L85

While failing to explain methods that are not as easily understandable:

https://github.com/pygame/pygame/blob/79807da84c9bacf8df5a177763e14c924e3b15e2/examples/joystick.py#L101-L105


jacob-thompson:
closing as this repo appears to be dead

PR #4472, opened 2025-03-20
Updated base script to close pygame gracefully

lettuceDestroyer:
Updated the file `base_script.py` and `base_script_example.py` to close pygame gracefully.

### DIFF
--- a/docs/reST/ref/code_examples/base_script.py
+++ b/docs/reST/ref/code_examples/base_script.py
@@ -1,18 +1,19 @@
 # Example file showing a basic pygame "game loop"
 import pygame
+import sys
 
 # pygame setup
 pygame.init()
 screen = pygame.display.set_mode((1280, 720))
 clock = pygame.time.Clock()
-running = True
 
-while running:
+while True:
     # poll for events
     # pygame.QUIT event means the user clicked X to close your window
     for event in pygame.event.get():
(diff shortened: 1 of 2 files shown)


docs/reST/ref/code_examples/base_script.py
```diff
@@ -1,18 +1,19 @@
...
+import sys
 
 # pygame setup
 pygame.init()
 screen = pygame.display.set_mode((1280, 720))
 clock = pygame.time.Clock()
-running = True
 
-while running:
+while True:
```
rick2047, line 10: looks like you forgot to change it back to running, since its set to true in line 8 its functionally equivalent but more understandable.

docs/reST/ref/code_examples/base_script_example.py
```diff
@@ -1,21 +1,22 @@
...
 pygame.init()
 screen = pygame.display.set_mode((1280, 720))
 clock = pygame.time.Clock()
-running = True
 dt = 0
 
 player_pos = pygame.Vector2(screen.get_width() / 2, screen.get_height() / 2)
 
-while running:
+while True:
```
rick2047, line 13: should revert back to running

rick2047 (changes requested):

PR #4507, opened 2025-06-03
Update pkgdata.py

Locked-chess-official:
Change the deprecated api "pkg_resources"


Locked-chess-official:
Now another programmer is doing it too #4511, This closed.

PR #4533, opened 2025-07-23
Fix scaling tests to match new rounding logic

GrandBIRDLizard:


GrandBIRDLizard (commented):
Closes #4294 fixes rounding logic in transform.c and updates transform_test to pass

samsepro:
I tested this fix locally and it works as expected.

Before: (int)(surf->w * scalex) caused incorrect truncation
After: (int)(surf->w * scalex + 0.5) correctly fixes the rounding

Tested with:
- Surface: (1468, 200)
- Factor: 240 / 1468
- Result before fix: 239
- Result after fix: 240

The added test cases for scale, scale_by and smoothscale_by are correct and consistent with the fix.

Issue #4537, opened 2025-07-25
[Windows] Pygame window in created behind Command Prompt after CLI interaction


konbraphat51:
### Description
After user input happend in CLI and then pygame started, the pygame window always created **behind** the CLI window (ex: Command Pront / VSCode)

### Reproducible Code
```python
import pygame as pg

# CLI interaction
input("press enter: ")

# start pygame
pg.init()
screen = pg.display.set_mode((640, 480))
screen.fill((0, 0, 0))
pg.display.flip()

# wait for user closing pygame window
running = True
clock = pg.time.Clock()
while True:
    for event in pg.event.get():
        if event.type == pg.QUIT:
            pg.quit()
    clock.tick(60)
```

results in...

<img width="1298" height="808" alt="Image" src="https://github.com/user-attachments/assets/fc96b341-7cfc-40cf-837c-072c3ad20745" />

### Environment
```
OS Name	Microsoft Windows 11 Pro
Version	10.0.26100 Build 26100
Other OS Description 	Not Available
OS Manufacturer	Microsoft Corporation
Processor	13th Gen Intel(R) Core(TM) i7-13700F, 2100 Mhz, 16 Core(s), 24 Logical Processor(s)
```

PR #4539, opened 2025-07-26
Add `pygame.display.focus()` for Windows OS

konbraphat51:
## Problem
Pygame window in created behind Command Prompt after CLI interaction: https://github.com/pygame/pygame/issues/4537

## What is added
- `pygame.display.focus()`  forces to activate (focus) the pygame window **only for Windows OS**
- and its test code (<- passed!)

## Quick Check code
```python
import pygame as pg
import subprocess

# CLI interaction
input("press enter: ")

# start pygame
pg.init()
screen = pg.display.set_mode((640, 480))
screen.fill((0, 0, 0))
pg.display.flip()

pg.time.wait(1000)

subprocess.Popen("notepad.exe")

pg.time.wait(1000)

pg.display.focus()

# wait for user closing pygame window
running = True
clock = pg.time.Clock()
while True:
    for event in pg.event.get():
        if event.type == pg.QUIT:
            pg.quit()
    clock.tick(60)
```


konbraphat51:
@dusu233 
umm, sorry, I could not get what you are trying to tell me through that code...

PR #4583, opened 2025-09-14
Replace deprecated pkg_resources with importlib.resources

### LINKED ISSUE #4313: Deprecation warning when testing with pygame imported
**Environment:**

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Windows 11 
- **Python version** (e.g. 3.7.9, 3.8.5) : 3.12.4
- **SDL version** (e.g. SDL 2.0.12):
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.6.0
- **Relevant hardware** (e.g. if reporting a bug about a controller, tell us the brand & name of it):

**Current behavior:**
When running tests with pytest with pygame imported the following warning appears
``` 
platform win32 -- Python 3.12.4, pytest-8.3.2, pluggy-1.5.0

C:\Users\***\AppData\Roaming\Python\Python312\site-packages\pygame\pkgdata.py:25
  C:\Users\***\AppData\Roaming\Python\Python312\site-packages\pygame\pkgdata.py:25: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    from pkg_resources import resource_stream, resource_exists
```
**Expected behavior:**

No deprecation warning appears. 

**Screenshots**

If applicable, copy and paste screenshots to help explain your problem.

**Steps to reproduce:**
Running Pytest on the following code 

**Test code**
```import pygame

def test_passing():
    pass
```

rchaput:
Because setuptools' `pkg_resources` module is deprecated, importing Pygame raises a warning:

>>> pygame\pkgdata.py:25: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  from pkg_resources import resource_stream, resource_exists

The [`importlib.resources`](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources) is the officially recommended module to use as a replacement for `pkg_resources`; this PR proposes to:

- Drop support for Python < 3.9, because `importlib.resources` is only available since Python 3.7; within this module, the functions that we want to use (and are not deprecated themselves) are only available since 3.9. The older Python versions are no longer supported (even Python 3.9 is in its end-of-life state), so this should be an acceptable change. Users that want to keep using older versions of Python could pin their version of Pygame to an older one as well.
- Replace `pkg_resources` with `importlib.resources`; because this new module has a slightly different API, 2 wrapper functions are also created to offer a 1-to-1 mapping with the previous `pkg_resources` API, so that the remaining Pygam code does not require any update.

Fixes #4313

### DIFF
--- a/src_py/pkgdata.py
+++ b/src_py/pkgdata.py
@@ -22,7 +22,14 @@ import sys
 import os
 
 try:
-    from pkg_resources import resource_stream, resource_exists
+    from importlib.resources import files
+
+    def resource_exists(_package, _resource_name):
+        return files(_package).joinpath(_resource_name).is_file()
+
+    def resource_stream(_package, _resource_name):
+        return files(_package).joinpath(_resource_name).open("rb")
+
 except ImportError:
 
     def resource_exists(_package_or_requirement, _resource_name):


rchaput:
Glad it helped!

Sep 14, 2025 21:51:19 Zane Bradley ***@***.***>:

PR #4604, opened 2025-10-09
make faster for users not useing pip

ghost:


ghost:
this helps because i encounter a errore where pip wouldent install a depence so this will fix ity in case

PR #4610, opened 2025-10-14
Surface.set_at and Surface.get_at now accept keyword arguments

atif5:
Changes done to Surface.set_at and Surface.get_at functions. 
They now accept keyword arguments! 

Related to #808.


atif5:
@illume sorry for the ping but can you look at this? 🥲

PR #4653, opened 2025-11-09
fix: correct typos in examples/README.rst

Albez0-An7h:
- Fix filename reference from aacircles.py to aacircle.py to match the filename.
- Fix grammar: "to drawing" -> "to draw"


Albez0-An7h:
@illume when you get time, can you please review this?

PR #4676, opened 2025-11-25
Fix test_magic NameError by renaming helper function

NGANAMODEIJunior:
When we use pytest to test image_test.py, test_magic is considered like a test. so we rename it into  _test_magic


NGANAMODEIJunior:
the small fail in the CI is an error unrelated to these changes

PR #4693, opened 2025-12-20
Fixed an error where no cameras would give None instead of []

robotsnh:
Formerly: if len(cameras)==0:
New: if len(cameras)==0 or cameras==None:


robotsnh:
didn't write a unit test; will do that next

PR #4694, opened 2025-12-20
Fixed an error occuring when no cameras are connected

robotsnh:
Added a section where, if there are zero video cameras connected, therefore cameras being None instead of [], it will skip the camera unit test.


robotsnh:
need to create a pull request for some of the frozen runpy appveyor stuff

PR #4726, opened 2026-01-17
Rename docs/reST/index.rst to import pygame print(pygame.version.ver)

soni5656shiva-ship-it:


robotsnh (changes requested):
I don't see the rationale behind this change. This was a perfectly fine, working reStructuredText documentation file that has been renamed to a piece of Pygame code with no file extension whatsoever.

PR #4762, opened 2026-02-14
I've added a completely new game

Devanch-hub:
This game adds a new look to this repository by adding a chance factor. Although this game doesn't require  the  player's action, it would be great to bet one and enjoying

### DIFF
--- a/examples/custom_file_format.py
+++ b/examples/custom_file_format.py
@@ -1,0 +1,98 @@
+import struct
+import pygame
+import io
+import sys
+
+class DaychIO:
+    """
+    A utility class for the .daych custom image format.
+    Designed for high-performance Pygame surface manipulation.
+    """
+    HEADER_FORMAT = "!5sBHHB"  # Sig(5), Ver(1), W(2), H(2), BPP(1)
+    HEADER_SIZE = struct.calcsize(HEADER_FORMAT)
+    SIGNATURE = b"DAYCH"
+
+    @staticmethod
+    def load_surface(source):
+        """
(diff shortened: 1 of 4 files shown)


Devanch-hub:
sxcsdesd

robotsnh (changes requested):
I don't see the reason for your change in `examples/midi.py`. As well as this, I don't see why you've changed the resolution and FPS in `examples/setmodescale.py`. In addition to this, Pygame is... well, Pygame. It is not turtle, so this example is not appreciated and I doubt it will be by the community. In addition, files in the `examples/` directory follow an already-established naming convention, which, in your case, has been completely flouted. Therefore, I'd suggest you close this PR as off-topic.

robotsnh:
See above

Devanch-hub:
Is the old github pr still there and cant my new pr bee seen just because
of that, because im quite new in this and due to the internet restrictions
in my country im quite getting trouble to do whatever i want at any time.
Ill have ear on for your earliest response

robotsnh:
@Devanch-hub You have to create a new branch to have a different PR. If you have limited internet access, I'd recommend using GitHub Desktop or the `git` command line tool to work offline and "pushing" your commits back here when you have access; I'm sure I could even write a script that automatically pushes when you're online.

Devanch-hub:
Wow thanks a lot for all these these clarification and help! I will start
working on and hope those will work at the end

Devanch-hub:
And about script, i think that wouldnt work as i live in a country where
there are everything accesed via vpn (not North Korea), though i always
have connection on the wall just doesnt let me pass it. And also its really
hard to find vpn so....

PR #4770, opened 2026-02-22
fifa

quangkhang6677-cmd:


robotsnh (changes requested):
Python is not written in Unity or C#, and that is a reStructuredText documentation file. In addition, this has nothing to do with FIFA. I'd suggest that you close this PR as off-topic.

PR #4779, opened 2026-03-02
Update index.rst

alvaromorata280213-a11y:


robotsnh:
@alvaromorata280213-a11y Why are you pasting a Python program into a documentation page?

PR #4832, opened 2026-04-21
Remove greeting message from __init__.py

ghost:
Removed greeting message from the pygame community. We don't need this.


abelgarza (commented):
trash pr

robotsnh:
@benjaminnautilus I believe that the greeting message is used to test whether the package is importing correctly, but a second opinion would help here.

Arminh388:
Id say its not needed

TheHashiramaSenju (commented):
I checked the file and felt that it also might act as an indication script, So removing that is not advisable

PR #4607, opened 2025-10-13
[Fix] Move Font import

ninjaguardian:
Fixes circular import from Font.
`font.py` imports `sysfont.py`, which imports `font.py`

Sysfont's `from pygame.font import font` was moved to the function where it is needed.

Fixes #4170

PR #4876, opened 2026-07-04
Fix circular import in font fallback

Qeya:
When the C extension for font fails to build (or isn't available yet, such as in Python 3.14 environments), pygame falls back to a pure python implementation. This fallback suffers from a circular import where `pygame.font` imports `pygame.sysfont`, which in turn imports `Font` from `pygame.font` at the module level. Moving the `Font` import to be lazy in `sysfont.font_constructor` resolves the issue and allows the fallback to work properly.


Qeya:
#4607 just fix it already ;-;

